PR #302, opened 2019-12-07
Remove extra whitespace hiding inside d2l-link elements

j3parker:
This gap was really noticable when a link was beside punctuation (e.g. commas, periods, etc.)

I noticed this when [switching the blog to use Brightspace UI](https://github.com/Brightspace/cdo-blog/pull/179). Here's the before:

d2l-link-wc-inline

and after:

d2l-link-wc-inline

I didn't look closely at how visual-diff works; it looked like something else was maybe supposed to push the goldens to GitHub so I just left it alone. If this fails CI or there is more to do I can check it out on Monday!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ components/*/test/screenshots/golden/
 mixins/*/test/screenshots/current/
 mixins/*/test/screenshots/golden/
 test/sass.output.css
+.DS_Store
--- a/components/link/link.js
+++ b/components/link/link.js
@@ -41,15 +41,13 @@ class Link extends LitElement {
 	}
 
 	render() {
-		return html`
-			<a class="d2l-link"
+		return html`<a class="d2l-link"
 				aria-label="${ifDefined(this.ariaLabel)}"
 				?download="${this.download}"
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ components/*/test/screenshots/golden/
 mixins/*/test/screenshots/current/
 mixins/*/test/screenshots/golden/
 test/sass.output.css
+.DS_Store
```
j3parker: Sorry; this is not strictly related. Viewing the golden images in Finder in MacOS dirties the repo with these files.

dlockhart: I just have a global ignore rule locally on my machine... otherwise we end up with `.DS_Store`, plus Visual Studio user files, plus Atom user files etc. etc.

dlockhart:
Can you re-run this from a branch instead of a fork? Otherwise the visual diff tests don't run (they need access to encrypted environment variables) and since this will actually produce a new golden we need them to run.

j3parker:
Ah ok. You'll need to give me write permission and/or add me to the org then :)

j3parker:
I'll re-open this as a branch.

PR #305, opened 2019-12-09
input label styles, fieldset support and text labels

dlockhart:
The goal here is to provide shared styles and components such that folks can easily visually label  the various types of inputs we support today and in the future. The tricky part is that labels cannot span shadow boundaries. That means that a label with `for` and the input with `id` must be in the same shadow root, AND a `<label>` element that wraps an input must also be in the same shadow root.

This resulting PR adds a bunch of new features.

**1) Shared label styles for several different ways to label inputs, including using `<fieldset>`s to contain groups of related inputs**

These techniques require that input styles be imported into the LitElement in which this is being used. Furthermore, a label can gain the "required" asterix by either adding the `d2l-input-label-required` CSS class to the label node OR by surfacing a `required` attribute on the host custom element.

a) Linking labels with inputs using `for` and `id`. Simpler DOM but requires wire-up of the IDs.

```html
<label for="myInput" class="d2l-input-label [d2l-input-label-required]">Name</label>
<input id="myInput" type="text">
```

b) Wrapping a label around an input. Avoids having to link up IDs but requires an extra node so that the space between the label text and the input can be applied (and also so the required asterix can go in the right place).

```html
<label>
    <span class="d2l-input-label [d2l-input-label-required]">Name</span>
    <input type="text">
</label>
```

c) Wrapping a `<fieldset>` and `<legend>` around a group of related inputs.

Screen Shot 2019-12-09 at 4 41 57 PM

```html
<fieldset class="d2l-input-label-fieldset">
    <legend class="d2l-input-label [d2l-input-label-required]">Bun</legend>
    <d2l-input-checkbox checked>Whole wheat</d2l-input-checkbox>
    <d2l-input-checkbox>White</d2l-input-checkbox>
</fieldset>
```

**2) Baked in support for a label on the `<d2l-input-text>` web component**

Previously folks had to label this using only `aria-label`. I've retired that attribute (while still maintaining support for it) and replaced it with `label` and `label-hidden` (boolean).

Screen Shot 2019-12-09 at 4 41 49 PM

```html
<!-- visual label -->
<d2l-input-text label="First Name">
<!-- hidden label (goes on aria-label) -->
<d2l-input-text label="Last Name" label-hidden>
<!-- this still works, but isn't documented anymore -->
<d2l-input-text aria-label="Other thing">
```

I considered making this a mixin (I even wrote the code), but there currently aren't any other web components that need it so I left it in text-input. If the need arises, we can refactor it out.

**3) Fieldset helper element**

I also created a `<d2l-input-fieldset>` helper element to accomplish the same thing as using the native fieldset + legend + shared styles:

```html
<d2l-input-fieldset label="Bun" required>
    <d2l-input-checkbox checked>Whole wheat</d2l-input-checkbox>
    <d2l-input-checkbox>White</d2l-input-checkbox>
</d2l-input-fieldset>
```

**Next Steps**

If this looks good, I'll add in some README docs and merge. Then I want to create a couple other elements to handle form layout -- getting the spacing right for groups of fields that are positioned next to each other horizontally and vertically.


components/inputs/README.md
```diff
@@ -9,6 +9,87 @@ There are various input components available:
...
 
+## Labelling Inputs
+
+All inputs *must* have a label. Web component-based inputs like `<d2l-input-checkbox>`, `<d2l-input-search>` and `<d2l-input-text>` come with built-in labels. For the rest, labelling is accomplished visually using a `<label>` element or with a hidden label via the `aria-label` attribute.
+
+Groups of inputs (like checkboxes or radios) should be wrapped in a `<fieldset>` which can have label styles applied to it.
+
+### Visible labels using the `<label>` element
+
+![example screenshot of input label](./screenshots/label.gif?raw=true)
```
margaree: It looks like the screenshots in this PR are actually png so these are broken.

dlockhart: Good catch -- fixed!

components/inputs/input-fieldset.js
```diff
@@ -0,0 +1,41 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { labelStyles } from './input-label-styles.js';
+
+class InputFieldset extends LitElement {
```
dbatiste: Does this need the RTLMixin, since it's pulling in the label styles that includes some RTL specific stuff?

dlockhart: Good catch -- yes! Fixed.

PR #306, opened 2019-12-10
prevent the search button from drifting sometimes when the container grows

jstefaniuk-d2l:
Capture


dlockhart:
Weird... so what's actually happening here? Like why is the container growing but the host element (which has `position: relative`) isn't?

jstefaniuk-d2l:
the thing wrapping the search box has `display: flex` so I think that throws it off

dlockhart:
Since the container now has `position: relative`, I've removed it from the host -- which was the source of most of these issues. Also added some visual diff tests to catch this in the future.

The lesson here is that we shouldn't be relying on the host for relative/absolute positioning of its children, since the most can grow/shrink based on external factors.

PR #307, opened 2019-12-11
removing box-shadow transition on buttons

dlockhart:
None of the other inputs have transitions on the focus styles anymore, and this is causing a lot of grief in our visual diff tests.


components/button/test/button-icon.visual-diff.js
```diff
@@ -40,7 +40,7 @@ describe('d2l-button-icon', function() {
 						if (entry.category === 'translucent-enabled') {
 							await focus(page, '#translucent-enabled > d2l-button-icon');
 						} else {
-							await focus(page, `#${entry.category}`);
+							await page.$eval(`#${entry.category}`, (elem) => elem.focus());
```
dlockhart: Had to leave `focus()` around here for transluscent buttons.

PR #308, opened 2019-12-12
adding a method to simulate clicks on the checkbox

dlockhart:
This is so that folks can simulate a user interaction with the checkbox in their tests without needing to reach into the shadow root and click on the underlying `input` element.

### DIFF
--- a/components/inputs/input-checkbox.js
+++ b/components/inputs/input-checkbox.js
@@ -107,6 +107,15 @@ class InputCheckbox extends RtlMixin(LitElement) {
 		if (elem) elem.focus();
 	}
 
+	simulateClick() {
+		this.checked = !this.checked;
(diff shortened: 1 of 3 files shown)


components/inputs/input-checkbox.js
```diff
@@ -107,6 +107,15 @@ class InputCheckbox extends RtlMixin(LitElement) {
...
+	simulateClick() {
+		this.checked = !this.checked;
+		this.indeterminate = false;
+		this.dispatchEvent(new CustomEvent(
+			'change',
+			{bubbles: true, composed: false}
+		));
+	}
+
 	_handleChange(e) {
```
dlockhart: `_handleChange` could just call into `simulateClick` as the code is the same... but I felt weird calling it from there.

PR #310, opened 2019-12-18
Bind to the checked property of the checkbox, rather than attribute, …

tylergee:
…as described here:

https://github.com/Polymer/lit-element/issues/601
This fixes an issue whereby the checkbox could not have its checked property changed programatically, after it had been physically clicked.


dlockhart:
Re-created as #311, closing this one.

PR #311, opened 2019-12-18
binding to input property to fix user interaction bug

dlockhart:
Re-created #310 (FYI @tylergee) as a branch so that visual diff tests could run.

This issue [is described here](Polymer/lit-element#601). Essentially what's happening is that by binding to the underlying input's attribute instead of its property, as long as all interaction happens with `<d2l-input-checkbox>`, all is well. But when the user interacts directly with the checkbox (as is bound to happen!), then subsequent manipulation of `<d2l-input-checkbox>` does nothing. Once interacted with, native checkbox inputs don't seem to care about whether or not they have a `checked` attribute -- they respect their property instead.

### DIFF
--- a/components/inputs/test/input-checkbox.html
+++ b/components/inputs/test/input-checkbox.html
@@ -113,9 +113,30 @@
 						expect(input.indeterminate).to.be.true;
 					});
 
+					it('should bind "checked" attribute to input property', async() => {
+						elem.setAttribute('checked', 'checked');
+						await elem.updateComplete;
+						expect(input.checked).to.be.true;
+					});
+
+					it('should bind "checked" property to input property', async() => {
+						elem.checked = true;
+						await elem.updateComplete;
+						expect(input.checked).to.be.true;
+					});
+
+					it('should continue to bind "checked" property after interaction', async() => {
+						input.click();
+						expect(input.checked).to.be.true;
+						expect(elem.checked).to.be.true;
+						await elem.updateComplete;
+						elem.checked = false; // eslint-disable-line require-atomic-updates
+						await elem.updateComplete;
+						expect(input.checked).to.be.false;
+					});
+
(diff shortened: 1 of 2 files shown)


components/inputs/test/input-checkbox.html
```diff
@@ -113,9 +113,30 @@
...
+						await elem.updateComplete;
+						expect(input.checked).to.be.true;
+					});
+
+					it('should continue to bind "checked" property after interaction', async() => {
+						input.click();
+						expect(input.checked).to.be.true;
+						expect(elem.checked).to.be.true;
+						await elem.updateComplete;
+						elem.checked = false;
```
dlockhart: Linting is failing on this with a [require-atomic-updates](https://eslint.org/docs/rules/require-atomic-updates) error. I'm confused by it, but looking into it...

components/inputs/test/input-checkbox.html
```diff
@@ -113,9 +113,30 @@
...
+
+					it('should continue to bind "checked" property after interaction', async() => {
+						input.click();
+						expect(input.checked).to.be.true;
+						expect(elem.checked).to.be.true;
+						await elem.updateComplete;
+						elem.checked = false; // eslint-disable-line require-atomic-updates
+						await elem.updateComplete;
+						expect(input.checked).to.be.false;
+					});
```
dlockhart: @tylergee I simplified the test you had a bit to just replicate the issue once. If you noticed a good reason to flip it back and forth a couple times I can make it more complex, but this seemed to do the trick.

tylergee: No good reason to leave the flip-flop in there, nope. I just did that to fully reassure myself that things were working properly in both directions. For testing that the bug is fixed, just the simple case is good.

PR #289, opened 2019-11-28
checkbox should use RtlMixin

dlockhart:
This should have always been there.

Interestingly it's hard to catch these cases, since `d2l-demo-snippet` will put `dir="rtl"` on all the elements inside its shadow root when you press the button, which triggers the RTL CSS. But if the page itself has `dir="rtl"`, then you notice the problem.

PR #312, opened 2019-12-20
Initial Commit for Primary Secondary Page Template

wongvincent:
The demo looks like this
image


templates/template-primary-secondary.js
```diff
@@ -0,0 +1,59 @@
...
+
+class TemplatePrimarySecondary extends LitElement {
+
+	static get properties() {
+		return {};
+	}
+
+	static get styles() {
+		return css`
+			.container {
```
dlockhart: Is the `<div>` container even needed, or could you apply these styles to the host itself using `:host`?

templates/template-primary-secondary.js
```diff
@@ -0,0 +1,59 @@
...
+				display: grid;
+				grid-template-columns: 2fr auto 1fr;
+				grid-template-rows: auto auto auto;
+				grid-template-areas:
+				"header header header"
+				"primary divider secondary"
+				"footer footer footer";
+			}
+			#header {
```
dlockhart: We have to be a bit careful with IDs, since for browsers who don't yet support Shadow DOM (coughEdgecough), all of these IDs will be thrown into the global namespace and could collide with others on the page or other CSS on the page.

For that reason, I'd recommend going with CSS classes instead, and namespace the class names with the component name. So they'd become `d2l-template-primary-secondary-header`, `d2l-template-primary-secondary-footer`, and so on. It's a bit of a mouthful, but GZIP to the rescue!

templates/demo/template-primary-secondary.html
```diff
@@ -0,0 +1,51 @@
...
+							I'm in the PRIMARY slot!<br /><br />Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque euismod pellentesque dui, ac porta sapien laoreet et. Suspendisse tempor purus nec arcu finibus aliquam. Quisque eu erat blandit, luctus nulla nec, tempus velit. Morbi malesuada eu mauris volutpat feugiat. Duis eget hendrerit ex. Fusce vitae sem pretium, gravida velit eget, dapibus odio. Cras a maximus risus, ut sagittis augue.<br /><br />This is the end of the PRIMARY slot!
+						</div>
+						<div slot="secondary">
+							I'm in the SECONDARY slot!<br /><br />Duis efficitur rutrum ipsum rhoncus tristique. Pellentesque nulla elit, porta in auctor id, interdum eu dui. Suspendisse turpis urna, malesuada vel luctus quis, scelerisque sagittis justo. Quisque ac nulla nec dui varius posuere in nec est.<br /><br />This is the end of the SECONDARY slot!
+						</div>
+						<div slot="footer">I'm in the FOOTER slot!</div>
+					</d2l-template-primary-secondary>
+					<style>
+						d2l-template-primary-secondary div[slot=header] {
+							background-color: #888;
```
dlockhart: I'd probably do this with just inline styles, i.e. `<div slot="primary" style="background-color: #888">`.

templates/template-primary-secondary.js
```diff
@@ -0,0 +1,59 @@
...
+		return css`
+			.container {
+				display: grid;
+				grid-template-columns: 2fr auto 1fr;
+				grid-template-rows: auto auto auto;
+				grid-template-areas:
+				"header header header"
+				"primary divider secondary"
+				"footer footer footer";
+			}
```
dlockhart: I think the design implied that the template should always be the full height of the viewport. From the demo screenshot, I'm not sure it's doing that?

templates/template-primary-secondary.js
dbatiste: I sometimes find myself on the fence about this.  If they are on the host, technically the consumer can override them, and in this case if they change the `display` the rest of the properties (and potentially others for nested elements) don't make sense and the layout breaks.  On the flip-side, sometimes giving the consumer the power is valid and opens up the API for them... but they ultimately take responsibility for it.

dlockhart: Right, that's a really good point. Haha so which case does this fall under? The only trade-off I can think of is 1 extra element, which for something like buttons is a big deal but for a page-level template is not. So I'd be good keeping the extra `<div>` for that safety.

svanherk:
Is there somewhere we can see the current state of the design?  I haven't been in the latest meetings, I'm interested in this naming convention... what are we imagining for potential future templates?

dlockhart:
Sent you a link offline to the design. I don't think design has put too much thought into other templates yet, but you can probably imagine them based on how our existing pages are laid out.

templates/template-primary-secondary.js
dbatiste: Yeah I'm not sure it really matters in this case.  I think the pro for using `host` in this case is that it avoids a potentially unnecessary element, and the goal isn't to expose any aspect of the layout... is that true?  

`d2l-card` and `d2l-dialog` are two other good examples - they have a specific internal layout they are trying to provide, and if the consumer could override it, things are going to go haywire.  Card is especially interesting, because we expect the user might set `display` on it, in order to render with other cards in grid-like layout.  

I think `display` is one of those muddy properties that, depending on it's value, can affect both the layout behavior of the element in its parent context (ex. `inline-block`, `inline`, et.al.), and also the layout behavior of children within it (ex. `flex`, `grid`).  In some cases (like `d2l-card`), we need the extra element to separate these two concerns, but in other cases it's not so clear.

Anyway, not to belabor the topic.  Just wanted to air it as a point of non-clarity for many elements, at least for me... 🤔

svanherk:
> Sent you a link offline to the design. I don't think design has put too much thought into other templates yet, but you can probably imagine them based on how our existing pages are laid out.

Thanks, that helps a lot, "primary" and "secondary" are well defined and used throughout the doc, but it doesn't imply all templates would follow this kind of name if it doesn't make sense (like in a single panel and three panel layout).  I think this looks like a good start!

templates/template-primary-secondary.js
dlockhart: Totally, and since this is relying on `display: grid` I think it makes a lot of sense to hide that away in this container. So ignore my initial comment @wongvincent, keep the container. :)

Having said all that, you should probably set an initial display value for the host since by default it'll be an inline element which probably isn't what you want. And then as soon as you do that, you'll need to provide the `[hidden]` selector:

```css
:host { display: block; }
:host([hidden]) { display: none; }
```

templates/demo/template-primary-secondary.html
```diff
@@ -0,0 +1,51 @@
...
+		<meta charset="UTF-8">
+		<link rel="stylesheet" href="../../components/demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '../../components/demo/demo-page.js';
+			import '../template-primary-secondary.js';
+		</script>
+	</head>
+	<body>
+		<d2l-demo-page page-title="Primary Secondary Template">
```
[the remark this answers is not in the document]
svanherk: The RTL toggle is a nice feature and feels like it would be important to test for this component, although hopefully grid will handle almost everything for us.  Should we also have an `index-rtl.html`?

templates/template-primary-secondary.js
wongvincent: Yeah, it isn't right now. We discussed this briefly in the meeting about how we will need to calculate the height of the header and footer and subtract that from the viewport height to determine the height of the primary and secondary panels. Haven't gotten to figuring out how to do that yet.

templates/demo/template-primary-secondary.html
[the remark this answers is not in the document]
dlockhart: TBQH I think the RTL button can be misleading -- it sets `dir="rtl"` directly on the component, which can cause things to look fine even if the component is missing the `RtlMixin`.

For page templates, I think relying on a visual diff that does RTL testing is probably good enough, but yeah we could do an `index-rtl.html` if we wanted as well (although that duplicates a bunch of demo code).

templates/template-primary-secondary.js
dlockhart: There might be other techniques we could use as well that don't involve quite as much constant re-calculating and observing. I can help with that when you get to it.

templates/demo/template-primary-secondary.html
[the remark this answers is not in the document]
svanherk: Oh _interesting_... you're right that is misleading, because I completely didn't think about that (even though it can only really be done that way, to work for each individual demo.  I wonder if it should be moved to just a higher-level toggle that would toggle the whole page and act more like a real-life scenario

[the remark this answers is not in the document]
svanherk: Ah, seems like this was already discussed here: https://github.com/BrightspaceUI/core/pull/289
Sorry, missed this!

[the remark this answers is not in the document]
wongvincent: I've removed the `d2l-demo-page` and `d2l-demo-snippet` wrappers. Moved the demo folder into the `primary-secondary` folder and renamed this demo file to `index.html`.

templates/template-primary-secondary.js
dbatiste: `flex-direction: column` worked out pretty well for both `d2l-card` and `d2l-dialog`.

wongvincent: I've fixed it in this commit by setting `height: 100vh` on the container and fixing an error I had in the grid-template-row height. https://github.com/BrightspaceUI/core/pull/312/commits/4bc74b573d7b0c58f22c3d7e7d0de1b860f44b34

templates/primary-secondary/primary-secondary.js
```diff
@@ -0,0 +1,67 @@
...
+		return css`
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.container {
+				display: grid;
+				grid-template-columns: 2fr auto 1fr;
```
dlockhart: Should this be `2fr 1px 1fr`?

templates/primary-secondary/primary-secondary.js
```diff
@@ -0,0 +1,67 @@
...
+			}
+			.d2l-template-primary-secondary-divider {
+				grid-area: divider;
+				background-color: var(--d2l-color-mica);
+				width: 1px;
+				height: 100%;
+			}
+			.d2l-template-primary-secondary-secondary {
+				grid-area: secondary;
+				min-width: 320px;
```
dlockhart: I think you can set this from the grid definition [using `minmax`](https://css-tricks.com/snippets/css/complete-guide-grid/#article-header-id-4). So something like `grid-template-columns: 2fr auto minmax(320px, 1fr);`

templates/primary-secondary/primary-secondary.js
```diff
@@ -0,0 +1,67 @@
...
+			.d2l-template-primary-secondary-primary {
+				grid-area: primary;
+				min-width: 320px;
+				overflow-y: auto;
+			}
+			.d2l-template-primary-secondary-divider {
+				grid-area: divider;
+				background-color: var(--d2l-color-mica);
+				width: 1px;
+				height: 100%;
```
dlockhart: I think you can use `align-self: stretch` (or `align-items: stretch` if you want ALL the cells to do this) so make the container fill the entire height  & width of the grid cell.

wongvincent: Actually `height: 100%`, isn't necessary anymore since I fixed the row to fill up the remaining height on the page. So this divider cell already has the full height.

PR #314, opened 2019-12-30
Vwong/US112517 Primary Secondary Template Visual Diff Tests

wongvincent:


templates/primary-secondary/test/primary-secondary.visual-diff.html
```diff
@@ -0,0 +1,62 @@
...
+		import '../../../components/typography/typography.js';
+		import '../primary-secondary.js';
+	</script>
+	<title>d2l-template-primary-secondary</title>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+</head>
+<body class="d2l-typography">
+	<div class="visual-diff">
```
dlockhart: The `visual-diff` CSS class just puts some margins below the `<div>` so that when we're testing many things on a page they don't get bunched up -- I think it can be excluded.

Also `styles.css` puts some padding on the body that you likely don't want, so I think the whole CSS file can be left out.

I'm wondering since these templates should only be used one-per-page, if this should be broken out into one file per visual diff test? Otherwise They could be interfering with each other, especially once we start doing responsive stuff with the panels.

templates/primary-secondary/test/primary-secondary.visual-diff.js
```diff
@@ -0,0 +1,33 @@
...
+		});
+
+		[
+			{ name: 'normal', selector: '#normal', fileName: 'primary-secondary-desktop.visual-diff.html' },
+			{ name: 'larger-than-viewport-height', selector: '#larger-than-viewport-height', fileName: 'primary-secondary-desktop-large.visual-diff.html' }
+		].forEach((entry) => {
+			it(entry.name, async function() {
+				await page.goto(`${visualDiff.getBaseUrl()}/templates/primary-secondary/test/${entry.fileName}`, {waitUntil: ['networkidle0', 'load']});
+				await page.bringToFront();
+				const rect = await visualDiff.getRect(page, entry.selector);
```
dlockhart: Nit: since the pages only have the one element in them you could remove the `id` and simplify this to:
`visualDiff.getRect(page, 'd2l-template-primary-secondary')` (using an element selector).

templates/primary-secondary/test/primary-secondary.visual-diff.js
```diff
@@ -0,0 +1,33 @@
...
+		});
+
+		[
+			{ name: 'normal', fileName: 'primary-secondary-desktop.visual-diff.html' },
+			{ name: 'larger-than-viewport-height', fileName: 'primary-secondary-desktop-large.visual-diff.html' }
+		].forEach((entry) => {
+			it(entry.name, async function() {
+				await page.goto(`${visualDiff.getBaseUrl()}/templates/primary-secondary/test/${entry.fileName}`, {waitUntil: ['networkidle0', 'load']});
+				await page.bringToFront();
+				const rect = await visualDiff.getRect(page, 'd2l-template-primary-secondary');
```
dbatiste: Note: the API for `getRect` accepts `(page, selector, margin)`, the `margin` being optional and defaulting to `10px`. The reason for the default margin is that many of our components use `box-shadow` for various purposes, but `box-shadow` falls outside of the bounding rect as measured by the browser. In order for us to reliably capture the shadow, we default to taking a screenshot slightly larger than the component, but this may be overridden as necessary. This is also why we try to separate the fixtures (when there are multiple on a single page)... so that a screenshot does not include the bottom/top of adjacent fixtures.  Anyway, that's the background... long story short, we _might_ want `0` for the margin in this case. Not sure if it makes a difference in this case - it might, but we might not have noticed.

PR #315, opened 2019-12-31
switching to use Intl v3

dlockhart:
This adopts the breaking [changes to our Intl library](https://github.com/Brightspace/intl/pull/31). Won't be merged until Intl is published at its new location.


mixins/localize-mixin.js
```diff
@@ -1,14 +1,14 @@
-import {
-	addListener, formatDate, formatDateTime, formatFileSize, formatNumber,
-	formatTime, getDocumentLanguage, getDocumentLanguageFallback, getTimezone,
-	localize, parseDate, parseNumber, parseTime, removeListener
-} from '../helpers/localization.js';
+import {formatDate, formatDateTime, formatTime, parseDate, parseTime} from '@brightspace-ui/intl/lib/dateTime.js';
+import {formatNumber, parseNumber} from '@brightspace-ui/intl/lib/number.js';
+import {formatFileSize} from '@brightspace-ui/intl/lib/fileSize.js';
```
dlockhart: Separately, I'd like to go through places that are using these and switch them to just use `intl` directly, leaving this mixin solely for the `localize()` method. Otherwise, any component that just needs to localize a piece of text gets all the date/time/number formatting and parsing included without needing it.

mixins/test/localize-mixin.html
```diff
@@ -54,142 +49,18 @@
 					});
 				}
 
-				describe('initial load', () => {
```
dlockhart: These deleted tests moved into `intl`.

package.json
```diff
@@ -62,9 +62,9 @@
     "whatwg-fetch": "^3"
   },
   "dependencies": {
+    "@brightspace-ui/intl": "Brightspace/intl#v3",
```
dlockhart: This still needs to be updated to the actual publish location on NPM when it's there.

mixins/localize-mixin.js
svanherk: Sounds good, should we update the README know to try to stop future use?

dlockhart: I'd like to actually go through and remove this from here if it ends up being easy to clean up places that are using it, making a README update unnecessary. If it ends up being too much work and I need to create a new slimmer localize mixin, then yeah I'll deprecate and add a note.

PR #316, opened 2020-01-02
Added a max width to the meter linear, so if the value is larger than…

TomMichielsD2L:
… the max, the meter does not grow too large


dlockhart:
@TomMichielsD2L, would it be possible to add a visual diff test that exercises this use case? Just so we don't cause a regression in the future.

TomMichielsD2L:
> @TomMichielsD2L, would it be possible to add a visual diff test that exercises this use case? Just so we don't cause a regression in the future.

@dlockhart I do not know how to do that :)

TomMichielsD2L:
@dlockhart `npm run test:diff:golden:commit` did not work for me, so I manually added the generated images to the ci folder. Did I miss anything else?

m6jones:
What is the case for going over? Should this be handled better?

Maybe when calculating the values?

m6jones:
Refreshing what has been done. This seems reasonable.

TomMichielsD2L:
> What is the case for going over? Should this be handled better?
> 
> Maybe when calculating the values?

The use case is that we want to show the user that the completed x/y hours of work, where x can be larger than y. I figured that since the radial meter handled this, that the linear meter should as well.

PR #319, opened 2020-01-02
Add optional footer to d2l-primary-secondary template

wongvincent:


templates/primary-secondary/primary-secondary.js
```diff
@@ -4,7 +4,9 @@ import { css, html, LitElement } from 'lit-element/lit-element';
 class TemplatePrimarySecondary extends LitElement {
 
 	static get properties() {
-		return {};
+		return {
+			hideFooter: { type: Boolean }
```
dbatiste: If we go this route and we want to set this from an attribute, I think we would want:
```
hideFooter: { type: Boolean, attribute: 'hide-footer' }
```
We might actually want it to be an adjective rather than a verb, same as native hidden attributes:
```
hiddenFooter: { type: Boolean, attribute: 'hidden-footer' }
```
However, this may also be accomplished without a **public** property/attribute by using the `slotchange` event and checking for assigned nodes, and then setting a private property.

```
static get properties() {
    return {
        _hasFooter: { type: Boolean }
    };
}

render() {
    return html`
        ...
        <div ?hidden="${this._hasFooter}><slot @slotchange="${this._handleFooterSlotChange}"></slot></div>
        ...
    `;
}

_handleFooterSlotChange(e) {
    const nodes = e.target.assignedNodes();
    this._hasFooter = (nodes.length !== 0);
}
```
Providing the public property along with the slot sorta results in two sources of truth, unless there are cases where there would be content in the slot but the footer is still meant to be hidden (a scenario that comes up with Polymer `dom-if`, but generally avoided with Lit conditional rendering). The `slotchange` approach naturally handles changes to the slot alone.  The downside to the `slotchange` approach is that `render` is potentially triggered twice, but I would consider it very marginal cost for a single component.

Food for thought.

wongvincent: I like the last option you suggested with the `slotchange`. But it seems like it isn't supported in Edge?

dbatiste: It's not supported natively in Edge, as web components and slots aren't supported in Edge either, but the web components Polyfill takes care of it.  We have a few components that make use of it.  I did a quick test with some console logs with one of them in Edge to confirm. Hopefully it's a viable option - it would be preferred for the user to just not put anything in the slot.

Note that `assignedNodes()` will return text nodes as well as elements.  There is an `assignedElements()` API that would exclude those, but I'm not sure if Safari supports it.  If non-element nodes are a problem when using `assignedNodes()`, one just needs to ignore them by filtering on `nodeType`.

templates/primary-secondary/test/primary-secondary.visual-diff.js
```diff
@@ -20,12 +20,13 @@ describe('d2l-template-primary-secondary', function() {
 
 		[
 			{ name: 'normal', fileName: 'primary-secondary-desktop.visual-diff.html' },
-			{ name: 'larger-than-viewport-height', fileName: 'primary-secondary-desktop-large.visual-diff.html' }
+			{ name: 'larger-than-viewport-height', fileName: 'primary-secondary-desktop-large.visual-diff.html' },
+			{ name: 'hideFooter', fileName: 'primary-secondary-desktop-hide-footer.visual-diff.html' }
```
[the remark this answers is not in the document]
wongvincent: Sorry, it is included now

templates/primary-secondary/primary-secondary.js
```diff
@@ -43,6 +45,8 @@ class TemplatePrimarySecondary extends LitElement {
 			}
 			.d2l-template-primary-secondary-footer {
 				grid-area: footer;
+				box-shadow: 0 -2px 4px rgba(86, 90, 92, .2);
+				height: 78px;
```
dlockhart: I know the design says the height of this is `78px`, but I'm not sure if it actually should be. Our buttons are actually not a consistent height -- they grow/shrink based on the user's font size preference. So they're `42px` by default, but can be taller or shorter. Even with the largest font size, the buttons could probably still fit in `78px` (just with less top/bottom space), but it might be worth checking with design what the actual intent is?

templates/primary-secondary/primary-secondary.js
dlockhart: Great suggestion Dave, I like where this ended up.

templates/primary-secondary/primary-secondary.js
wongvincent: I checked with Daniel and it doesn't need to be a consistent 78px.

PR #321, opened 2020-01-03
Updated all the meters to have the same behaviour when the max is zero

TomMichielsD2L:
The linear meter is empty if the max value is zero, but the other meters were not. After confirming with the design team it was determined that if the max is zero the meter should be empty. 

This change updates the circle and radial meter to be consistent with the expectations and linear meter.


components/meter/test/meter-circle.visual-diff.html
```diff
@@ -26,6 +26,9 @@
 	<div class="visual-diff" style="width: 90px;">
 		<d2l-meter-circle id="round-to-zero" value="0.004" max="100"></d2l-meter-circle>
 	</div>
+	<div class="visual-diff" style="width: 90px;">
+		<d2l-meter-circle id="zero-max" value="10" max="0"></d2l-meter-circle>
+	</div>
```
dlockhart: So there's already a "max-zero" test where the value is also 0. Could you rename it to "max-zero-value-zero" and name this one something like "max-zero-with-value" just to make it clear what the difference is? You should be able to just rename the existing visual diff file(s) and it'll "just work" when the diff runs again.

TomMichielsD2L: That makes sense to me, I have renamed the tests as per your suggestion. When I checked the diffs looked as expected, I will double check when the visual diff test runs again

components/meter/test/meter-linear.visual-diff.js
```diff
@@ -21,9 +21,10 @@ describe('d2l-meter-linear', function() {
 		{ title: 'no-progress', fixture: '#no-progress'},
 		{ title: 'has-progress', fixture: '#has-progress'},
 		{ title: 'completed', fixture: '#completed'},
-		{ title: 'max-zero', fixture: '#max-zero'},
+		{ title: 'max-zero-value-zero', fixture: '#max-zero-value-zero'},
```
dlockhart: Since this test was renamed, you'll want to rename the diff file that's checked in at `test/screenshots/ci/meter-linear/d2l-meter-linear-max-zero.png`.

TomMichielsD2L: It must be too early in the morning since you told me to do this :D

PR #326, opened 2020-01-09
Add dialog async handling

dbatiste:
Enabled dialogs to detect async content state changes and show/hide loading spinner accordingly.

* adds `async` property to `d2l-dialog` so it knows to expect async state changes
* adds a `d2l-loading-spinner` is `async`
* hides spinner when async state is com

I started to create visual-diff tests for this, but then realized it was not very doable since the loading spinner has the endless animation.


components/demo/code-view.js
```diff
@@ -40,7 +40,8 @@ class CodeView extends LitElement {
 			/* Current, non-relative imports don't appear to work with Polymer dev server
 			for FF, Edge, IE11.  Use of non-default languages is limited to dev with this approach.
 			https://github.com/Polymer/tools/issues/3402 */
-			this._dependenciesPromise = import(`/node_modules/prismjs/components/prism-${language}.min.js`);
```
dbatiste: Very strangely, I started to get an eslint parsing error on this line. Extracting the path from the import statement resolved the issue.

components/dialog/dialog.js
```diff
@@ -90,6 +97,19 @@ class Dialog extends LocalizeStaticMixin(DialogMixin(LitElement)) {
...
 
 	render() {
+
+		let content;
+		if (this.async && this.asyncState !== asyncStates.complete) {
+			content = html`
+				<div class="d2l-dialog-content-loading">
+					<d2l-loading-spinner size="100"></d2l-loading-spinner>
+				</div>
+				<div style="display: none;"><slot></slot></div>
```
dbatiste: The slotted async content is still rendered into the DOM so that we can listen for it's async events, however we set `display: none` so that it does not add any extra height/whitespace below the spinner.

components/dialog/demo/dialog-async-content.js
```diff
@@ -0,0 +1,64 @@
...
+		return html`${runAsync(this.href, (href) => this._getContent(href), {
+			success: (content) => content
+		})}`;
+	}
+
+	_getContent(href) {
+		return new Promise((resolve) => {
+			if (!href) {
+				throw new InitialStateError();
+				//reject();
```
dlockhart: Nit: comment

components/dialog/demo/dialog-async.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script>
+			//window.D2L = { DialogMixin: { preferNative: false } };
```
dlockhart: Commented out code

components/dialog/demo/dialog-async.html
```diff
@@ -0,0 +1,48 @@
...
+					<d2l-button id="open">Show Dialog</d2l-button>
+					<d2l-dialog id="dialog" title-text="Dialog Title" async>
+						<d2l-dialog-demo-async-content></d2l-dialog-demo-async-content>
+						<d2l-button slot="footer" primary dialog-action="ok">Click Me!</d2l-button>
+						<d2l-button slot="footer" dialog-action>Cancel</d2l-button>
+					</d2l-dialog>
+					<script>
+						document.querySelector('#open').addEventListener('click', () => {
+							document.querySelector('#dialog').opened = true;
+							//document.querySelector('#dialog').resetAsyncState();
```
dlockhart: Commented out code

components/dialog/demo/dialog-async.html
```diff
@@ -0,0 +1,48 @@
...
+							document.querySelector('d2l-dialog-demo-async-content').href = 'some-href';
```
dlockhart: I don't quite get the purpose of the href property in this demo component...

components/dialog/dialog.js
dlockhart: Nit: seems like there could be a slight optimization here somehow so that the `<div><slot></slot></div>` doesn't need to be repeated in both branches of the `if`-statement.

Maybe using a CSS class and `classMap` to hide the `<div>`?

```
let asyncContent;
let contentClasses = {};
if (this.async && this.asyncState !== asyncStates.complete) {
    contentClasses = {
			'd2l-dialog-content-hidden': true
		};
    asyncContent = html`<div class="d2l-dialog-content-loading><d2l-loading-spinner>...`;
}
```

```

dbatiste: Yeah I can pull that out.  I think each `${some-other-template}` might might result in start/end comment nodes generated by Lit (it uses them) so I was sorta just avoiding those.

components/dialog/demo/dialog-async.html
dbatiste: The runAsync directive used by the async component requires a key (typically it would be a property value), and when it changes, `render` gets rerun, and the async function get kicked off (returning a result after 5s using `setTimeout`). If I define this in advance or hard-code the key inside the component, the async state will be complete before the user opens the dialog - a valid scenario but it doesn't really demonstrate the state change.

So the purpose of the demo, I want to trigger the async function when the dialog opens so that we see the spinner and then the result.  I've tried to configure the demo component to match what I expect consumers to do (aside from the `setTimeout`).

components/dialog/dialog.js
dbatiste: I've refactored this to remove the duplication.  These are the comments I was referring to... 🤷‍♂ ...
image

PR #327, opened 2020-01-16
Rename export const for input-label-styles.js

cbelan-d2l:
* It had the same name as the style.js in typography.
* Also renamed the uses of it


dlockhart:
Looks like linting is failing due to alphabetical order of imports.

PR #328, opened 2020-01-17
Accounting for when footer children aren't in a slot.

cbusmand2l:
The children of the footer slot aren't being grabbed in Edge because the `.d2l-dialog-footer slot` selector doesn't match anything (they're direct children of the dialog, rather than in a slot), resulting in a null reference.

### DIFF
--- a/components/dialog/dialog-confirm.js
+++ b/components/dialog/dialog-confirm.js
@@ -72,7 +72,16 @@ class DialogConfirm extends DialogMixin(LitElement) {
 
 	_focusInitial() {
 		const footer = this.shadowRoot.querySelector('.d2l-dialog-footer slot');
-		const nodes = footer.assignedNodes();
+		let nodes;
+		if (footer) {
+			nodes = footer.assignedNodes();
+		} else {
+			//For Edge.
+			nodes = [...(this.childNodes.length ? this.childNodes : [])];
+			nodes = nodes.filter(
+				node => node.slot === 'footer'
+			);
+		}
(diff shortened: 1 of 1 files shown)


components/dialog/dialog-confirm.js
```diff
@@ -72,7 +72,16 @@ class DialogConfirm extends DialogMixin(LitElement) {
...
+		let nodes;
+		if (footer) {
+			nodes = footer.assignedNodes();
+		} else {
+			//For Edge.
+			nodes = [...(this.childNodes.length ? this.childNodes : [])];
+			nodes = nodes.filter(
+				node => node.slot === 'footer'
+			);
+		}
```
dlockhart: To simplify and only have one code path that works for both, would something like this work?

```
const footerNodes = this.shadowRoow.querySelector('.d2l-dialog-footer [slot="footer"]');
```

cbusmand2l: I don't think so - I can't seem to find a selector/combination of selectors that gets me down to the buttons in the slot. If we want to have the same code path either way, I think we'd have to use the new one using `childNodes` for all cases.

dlockhart: I played around with this a bit more, and was able to get a single selector working if we put the class name on the slot element. So I put `<slot class="d2l-dialog-footer-slot">` and then changed the querySelector to be: `const footer = this.shadowRoot.querySelector('.d2l-dialog-footer-slot')`. Seemed to work in at least Chrome and Edge.

PR #329, opened 2020-01-20
Calendar - Initial Work

margaree:
This PR is for general initial work on calendar mostly mimicking our existing calendar.
The keyboard navigation behaviour is based off of https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/datepicker-dialog.html
I separated out a calendar day in order to keep `calendar.js` a bit cleaner.

Current functionality:
- Displays calendar of dates that shows a `selected-value`; if no selected value then today is the selected value
- Navigation to previous and next month via selecting the appropriate subtle icon button or by using arrow keys (note that the previous or next month is navigated to once a date from that month becomes the focus date, even if that date is visible in the current month. This is how the ARIA example works)
- Other keyboard navigation via arrow keys and HOME and END buttons
- Switching selected date via clicking or enter or space
- Event (`d2l-calendar-selected`) triggered when date selected which contains object containing month/date/year of selected date
- The focus date (e.g., the 12th) stays constant when months are switched (e.g., if Jan 12th is currently the focus date (the date that would receive focus styling when keyboard navigation is resumed) and next month button is clicked, Feb 12th would be the focus date. This is consistent with ARIA example but this might get changed with our designs and is a spot to keep in mind).

To Do:
- [x] Any style updates once design is complete (places to watch for include hover, focus, hover on selected, focus on selected, what happens when different month is navigated to, today)
-  ~What happens with weekend days~ (I will deal with this separately if needed)
- [x] Page Up and Page Down keyboard functionality
- [x] There is a bug where navigating to previous month via arrow keys when 1 day in previous month is visible in current month doesn't focus on correct date when previous month is shown which is "fixed" by https://github.com/BrightspaceUI/core/blob/5aad37573f4b30d97ce60a9a2f266e8c40ccf3ab/components/calendar/calendar.js#L157 a second `this.updateComplete` which doesn't seem great so look into this.


components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldVal, prop) => {
+			if (prop === '_selectedDate') {
+				this._generateDaysInMonth();
+			}
+			if (prop === '_shownMonth') {
+				this._onMonthChange();
+				this._generateDaysInMonth();
+				this.updateComplete.then(() => {
```
dbatiste: Yeah this definitely seems odd... especially since there doesn't appear to be any change between these calls.

Aside, it would be valid to make this method `async` (for the `forEach`) and use `await` to avoid the nesting here.

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+						</tr>
+					</thead>
+					<tbody>
+						<tr>
+							${weekdayHeaders}
+						</tr>
+						${dayRows}
+					</tbody>
+				</table>
+				<slot></slot>
```
dbatiste: Is the slot used for anything?

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
+import '../button/button-icon.js';
+import './calendar-date.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { getDateTimeDescriptor, parseDate } from '@brightspace-ui/intl/lib/dateTime.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
+import { heading4Styles } from '../typography/styles.js';
+import { LocalizeStaticMixin } from '../../mixins/localize-static-mixin.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
```
dbatiste: Is `RtlMixin` used (or will it be)?

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+			PAGEUP: 33,
+			PAGEDOWN: 34,
+			SPACE: 32,
+			RIGHT: 39,
+			UP: 38
+		};
+
+		let preventDefault = false;
+
+		switch (e.keyCode) {
```
dbatiste: Question: when initially focusing on the  calendar via keyboard, where should focus go, say for focusing forwards and also focusing backwards? If tabbing backwards, should focus go to the date, or should it go to the same place as when focusing forwards?

components/calendar/calendar-date.js
```diff
@@ -0,0 +1,137 @@
...
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((old, prop) => {
+			if (prop === 'month' || prop === 'year') {
+				const todayDate = new Date();
+				this._today = (this.year === todayDate.getFullYear() && this.month === todayDate.getMonth() && this.date === todayDate.getDate());
+			}
+			if (prop === 'focused') {
```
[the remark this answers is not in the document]
dlockhart: I actually noticed a bug related to this: if you keyboard around, then click a different date, then continue keyboarding, the focus continues on from where the keyboard left off instead of where you clicked.

components/calendar/calendar-date.js
```diff
@@ -0,0 +1,137 @@
...
+			month: this.month,
+			date: this.date,
+			year: this.year
+		};
+		const eventDetails = {
+			bubbles: true,
+			composed: true,
+			detail: { date: fullDate }
+		};
+		this.dispatchEvent(new CustomEvent('d2l-calendar-selected', eventDetails));
```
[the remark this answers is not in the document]
dlockhart: I don't think it should, as that's not normally expected from JavaScript APIs -- only when the user interacts with the component. In the past, we've added JS methods that take a boolean as to whether to fire the event or not to facilitate testing.

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+			this._arrangedWeekdays.push(day);
+		}
+
+		this._dialogLabelId = getUniqueId();
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this.addEventListener('keydown', this._onKeyDown);
```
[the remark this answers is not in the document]
dlockhart: I think so, or `keypress`.

components/calendar/calendar.js
dlockhart: I think it will be used in the future when we need to add the "Set to Today" and "Clear" buttons, but that could wait until a future PR as it's a separate story. Either way!

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+			date: date,
+			month: month,
+			year: year,
+			otherMonth: prevMonth || nextMonth,
+			selected: this._checkIfDatesEqual(month, date, year, this._selectedDate),
+			focused: this._checkIfDatesEqual(month, date, year, this._focusDate)
+		};
+	}
+
+	_generateDaysInMonth() {
```
dlockhart: This and `_getNumberOfDaysInMonth` seem like they could just be utility functions outside of the class. The more we can move out of the class itself, the smaller the classes are that get instantiated over and over by the browser.

components/calendar/calendar.js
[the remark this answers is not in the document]
margaree: So I think we do need to use `keydown` here. With `keyup` I don't think we can prevent scrolling with the arrow keys (though we could also have a `keydown` listener that does the prevent default for those cases and then use `keyup` for this portion?), and then `keypress` doesn't work with the arrow keys (at least for chrome).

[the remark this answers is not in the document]
dbatiste: Yep, I think that's what we want.  😄

components/calendar/calendar.js
margaree: I think it will be - right now the keyboard arrow key navigation is backwards in rtl so I think that will have to be fixed (assuming that the expectation of rtl is pressing the left arrow key would focus on the date to the left etc.)

components/calendar/calendar.js
```diff
@@ -0,0 +1,474 @@
...
+					return html`
+						<td>
+							<d2l-calendar-date
+								date=${day.date}
+								month=${day.month}
+								year=${day.year}
+								?focused=${day.focused}
+								?other-month=${day.otherMonth}
+								?selected=${day.selected}>
+							</d2l-calendar-date>
```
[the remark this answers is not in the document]
margaree: It was because this component was getting large, but I have moved things back into one component now and I think it has actually ended up simplifying things.

components/calendar/calendar.js
dbatiste: I think so - our arrowkeys mixin has some similar logic for RTL.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+		const year = selectedDate.getAttribute('data-year');
+		const month = selectedDate.getAttribute('data-month');
+		const date = selectedDate.getAttribute('data-date');
+		this.selectedValue = new Date(year, month, date);
+		this._focusDate = new Date(year, month, date);
+
+		const eventDetails = {
+			bubbles: true,
+			composed: true,
+			detail: { date: `${year}-${parseInt(month) + 1}-${date}` }
```
dbatiste: Looks like the format here doesn't match the format in the readme for this event.  Also, should this date property be an actual `Date` or should we leave it as a `String`?

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this.selectedValue = this.selectedValue ? new Date(this.selectedValue) : null;
+		this._today = new Date();
+		this._focusDate = this.selectedValue ? new Date(this.selectedValue.getFullYear(), this.selectedValue.getMonth(), this.selectedValue.getDate()) : new Date(this._today.getFullYear(), this._today.getMonth(), 1);
+		this._shownMonth = this.selectedValue ? this.selectedValue.getMonth() : this._today.getMonth();
+		this._shownYear = this.selectedValue ? this.selectedValue.getFullYear() : this._today.getFullYear();
+	}
+
```
dbatiste: If consumers try to call `focus` on the component, I'm not sure it will do what we expect.  We _might_ need to implement `focus` and apply to the correct internal element.  There has been much discussion on focus behavior in web components, but it may still not be handled automatically.

components/calendar/README.md
```diff
@@ -0,0 +1,25 @@
...
+<!-- Today is February 21, 2020 -->
+<d2l-calendar
+	selected-value="2020-02-09"
+	summary="Click on a day to select it as the assignment due date.">
+</d2l-calendar>
+```
+
+**Properties:**
+
+- `selected-value` (String): Currently selected date. This should be in ISO 8601 Extended Format (`YYYY-MM-DDTHH:mm:ss.sssZ`).
```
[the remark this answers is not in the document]
dlockhart: Yeah I think we should be a little flexible here and handle a string that contains the date and time or just the date. Similarly, I think the time input should handle the date and time or just the time -- they maybe should share a parse/validation utility function.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	render() {
+		const weekdayHeaders = this._daysOfWeekIndex.map((index) => html`
+			<th role="columnheader" abbr="${this._descriptor.calendar.days.long[index]}">
+				<abbr class="d2l-body-small" title="${this._descriptor.calendar.days.long[index]}">
+					${this._descriptor.calendar.days.short[index]}
+				</abbr>
+			</th>
+		`);
```
dlockhart: Nice usages of `map()`!

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+							data-year=${day.getFullYear()}
+							tabindex=${focused ? '0' : '-1'}>
+							${day.getDate()}
+						</div>
+					</td>`;
+			});
+
+			return html`
+				<tr>${weekHtml}</tr>
+			`;
```
dlockhart: Super-nit: I think for really small snippets like this we could just put it all on one line

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+						<div
+							@click="${this._onDateSelected}"
+							@keydown="${this._onKeyDown}"
+							class=${classMap(classes)}
+							data-date=${day.getDate()}
+							data-month=${day.getMonth()}
+							data-year=${day.getFullYear()}
+							tabindex=${focused ? '0' : '-1'}>
+							${day.getDate()}
+						</div>
```
dlockhart: Is the `<div>` needed or could all of this have gone on the `td`?

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+									text="${this._computeText(getNextMonth(this._shownMonth))}"
+									icon="tier1:chevron-right">
+								</d2l-button-icon>
+							</td>
+						</tr>
+					</thead>
+					<tbody>
+						<tr>
+							${weekdayHeaders}
+						</tr>
```
dlockhart: Nit: collapse to 1 line to save whitespace/text nodes.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+				<slot></slot>
+			</div>
+		`;
+	}
+
+	_computeText(month) {
+		return this.localize('show', {month: this._descriptor.calendar.months.long[month]});
+	}
+
+	_focusDateAddFocus() {
```
dlockhart: Another option to all this code that's removing/adding tabindex to the currently focused date would be to move `_focusDate` to be a private property being tracked by Lit. That way whenever it changes, it would call render again which already handles setting the `tabindex` on the focussed date.

components/calendar/calendar.js
dlockhart: As mentioned above, I think `selectedValue` should always be that same ISO string (and not be converted to a Date) at boot-up time. Then the consumer can always get the ISO value by accessing it. That would make the event details less useful, but we could still include the ISO value and the Date value in the details for simplicity.

components/calendar/test/calendar.test.js
```diff
@@ -0,0 +1,61 @@
...
+
+const normalFixture = html`<d2l-calendar selected-value="2015-09-02T12:00Z"></d2l-calendar>`;
+
+describe('d2l-calendar', () => {
+
+	it('passes all axe tests', async() => {
+		const calendar = await fixture(normalFixture);
+		await expect(calendar).to.be.accessible();
+	});
+
```
dlockhart: I think there's a good opportunity here to write actual unit tests for all the utility methods that you have outside of the component class. e.g. `getDatesInMonthArray`, etc.

components/calendar/test/calendar.test.js
```diff
@@ -0,0 +1,61 @@
...
+	it('dispatches event when space key pressed on date', async() => {
+		const calendar = await fixture(normalFixture);
+		const el = calendar.shadowRoot.querySelector('div[data-date="2"]');
+		setTimeout(() => dispatchKeyEvent(el, 32));
+		const { detail } = await oneEvent(calendar, 'd2l-calendar-selected');
+		expect(detail.date).to.equal('2015-9-2');
+	});
+
```
dlockhart: Recommend having some tests here to handle some of the more complex stuff, such as:
- Ensuring `selected-date` value is kept up-to-date as user interacts
- Ensuring focus goes where you expect as the user arrows around or moves from month to month

components/calendar/test/calendar.visual-diff.html
```diff
@@ -0,0 +1,38 @@
...
+		import '../calendar.js';
+
+		const oldDate = Date;
+
+		// eslint-disable-next-line no-global-assign
+		Date = function(input1, input2, input3) {
+			if (!input1) return new oldDate('2018-02-12T12:00Z');
+			return (input2 !== 0 && !input2) ? new oldDate(input1) : new oldDate(input1, input2, input3);
+		};
+		Date.prototype = oldDate.prototype;
```
dlockhart: Are you overriding `Date` so that you can control when "now" is? I think [Sinon useFakeTimers](https://sinonjs.org/releases/v1.17.6/fake-timers/) lets you do that.

components/calendar/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+The `d2l-calendar` component can be used to display a responsively sized calendar that allows for date selection. It indicates a currently selected date if a `selected-value` specified, or if the user selects a date.
+
+![Calendar](./screenshots/calendar.png?raw=true)
+
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/calendar/calendar.js';
+</script>
+<!-- Today is February 21, 2020 -->
```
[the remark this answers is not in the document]
margaree: It's because in the screenshot Feb 21st has a larger font size and is bold, indicating that it is today, so the comment is specifically calling that out so the user knows that's why the 21st looks like that.

components/calendar/calendar-styles.js
```diff
@@ -0,0 +1,83 @@
...
+	.d2l-calendar-date.d2l-calendar-date-hover {
+		background-color: var(--d2l-color-gypsum);
+	}
+
+	.d2l-calendar-date:focus {
+		border: 2px solid var(--d2l-color-celestine);
+		outline: none;
+	}
+
+	.d2l-calendar-date-first-row {
```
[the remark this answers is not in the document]
margaree: I don't know that that would work given the way it's currently implemented since technically the first `tr` is the row with the days of the week (this one would more accurately be something like `d2l-calendar-date-first-date-row`)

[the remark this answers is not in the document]
dlockhart: Hmm, what about something along the lines of `tbody > tr:first` though? I'm assuming the days of the week are in the `thead`.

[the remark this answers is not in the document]
margaree: They are not currently in `thead` (I had just implemented the structure of what was in the monolith, which has them in `tbody`) but they definitely feel like they should be.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	if (firstDayOfMonth !== firstDayOfWeek) {
+		numDaysFromLastMonthToShowThisMonth = firstDayOfMonth - firstDayOfWeek;
+		if (numDaysFromLastMonthToShowThisMonth < 0) {
+			numDaysFromLastMonthToShowThisMonth += daysInWeek;
+		}
+	}
+	return numDaysFromLastMonthToShowThisMonth;
+}
+
+function getNumberOfDaysInMonth(month, year) {
```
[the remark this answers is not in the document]
margaree: ooo neat!

components/calendar/calendar-styles.js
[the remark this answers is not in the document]
dlockhart: Yeah seems like a worthwhile change.

components/calendar/calendar.js
margaree: I had found the easiest way to style the date to be a square within a scalable calendar was to have a div within the `td`, so it's potentially possible - I'll look into it.

components/calendar/test/calendar.test.js
margaree: I had looked into that but I think they would have to not be private in order for me to do that (though I could be missing/forgetting about a way to do this).

components/calendar/calendar.js
dlockhart: OK not a big deal if it makes things harder, just trying to reduce unnecessary elements when possible.

components/calendar/test/calendar.test.js
dlockhart: Yeah you'd need to export them. Not sure if we care that folks could import and use them... seems like an OK trade-off when it enables better testing.

margaree: ok cool, yeah seems like a good call to me then.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+			this._daysOfWeekIndex.push(i % daysInWeek);
+		}
+
+		this._dialogLabelId = getUniqueId();
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this.selectedValue = this.selectedValue ? new Date(this.selectedValue) : null;
```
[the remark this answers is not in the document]
margaree: Agreed. We could have a separate `_selectedValueDate` that's the `Date`-ified version of `selectedValue` (since it's useful for internal logic) and keep `this.selectedValue` as the string value?

components/calendar/calendar.js
margaree: This seems more ideal. `this._focusDate.setDate` doesn't actually trigger a re-render which is a bit of a pain, but it works if it's changed with something like `this._focusDate = new Date...`

components/calendar/calendar.js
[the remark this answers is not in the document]
dlockhart: Yeah you could, although then you need to always keep those 2 things in sync whenever something internal or external updates the property. In the time-input, we just created a parse function that converts the string to a `Date` and we just call it whenever we need the object version.

components/calendar/calendar.js
dlockhart: Cool, I think it'll end up with a more logical flow as well. It'll also prevent situations where the focus moves and updates tabindexes but then a re-render happens right away anyway.

components/calendar/calendar.js
margaree: Good point - it does not seem to be handled automatically. My thought is that the element to receive focus in that case would be the left `d2l-button-icon`, since that's the way it is if you tab in. Does that make sense to you?

components/calendar/calendar.js
margaree: For focusing forwards, focus initially goes to the previous month arrow. I'm not sure about focusing backwards. I think it would make sense to go to the same place as when focusing forwards.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	const firstDayOfMonth = new Date(year, month, 1).getDay();
+	const hasDaysFromPrevMonth = (firstDayOfMonth !== firstDayOfWeek);
+	return daysInWeek * (numWeeksPrevMonth - ((hasDaysFromPrevMonth) ? 1 : 0));
+}
+
+export function getPrevMonth(month) {
+	return (month === 0) ? 11 : (month - 1);
+}
+
+export function parseDate(val) {
```
margaree: I feel like this isn't great and I should go about this differently and should incorporate `convertUTCToLocalDateTime` to determine what date the user should actually see

components/calendar/calendar.js
dbatiste: Yep, I _think_ so.

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	if (!val) return null;
+
```
dlockhart: Wondering if maybe a regex might make this a bit easier to validate and then extract the various bits using groups? That's what the time input is doing and was pretty straightforward.

Naive attempt:
`/^([0-9]{4})-([0-9]{2})-([0-9]{2})/`

That way it'll match whether the time is there or not and you can use the groups + `parseInt` to grab the year, month and day.

components/calendar/calendar.js
dlockhart: Hmm, yeah worth having a discussion about as this will impact the calendar, time and date inputs. These ISO formats are always UTC offset zero, so yeah... maybe we should indeed be converting to local (if available) before rendering and then back to UTC when the value changes. I like that...

components/calendar/calendar.js
dlockhart: We could also use [Date.parse()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)... but it supports all kinds of crazy formats including those with timezone offsets which we maybe don't want if we want to force this into UTC.

components/calendar/calendar.js
dlockhart: If this is the case, then we always need the time as well, ya?

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+			'es': { show: 'Show {month} ' },
+			'fr': { show: 'Show {month} ' },
+			'ja': { show: 'Show {month} ' },
+			'ko': { show: 'Show {month} ' },
+			'nl': { show: 'Show {month} ' },
+			'pt': { show: 'Show {month} ' },
+			'sv': { show: 'Show {month} ' },
+			'tr': { show: 'Show {month} ' },
+			'zh': { show: 'Show {month} ' },
+			'zh-tw': { show: 'Show {month} ' }
```
dlockhart: We now support German (`de`) and Danish (`da`) so should include those translations as well. Can you get the actual translated values from the LMS somewhere?

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+	constructor() {
+		super();
+
+		this._labelId = getUniqueId();
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this._today = new Date();
```
dlockhart: This could be a potential source of bugs as this will be in the user computer local timezone. Not sure what this should be... UTC "now" then converted to local?

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+		if (this.selectedValue) {
+			const selectedValueDate = parseDate(this.selectedValue);
+			this._focusDate = new Date(selectedValueDate.getFullYear(), selectedValueDate.getMonth(), selectedValueDate.getDate());
+			this._shownMonth = selectedValueDate.getMonth();
+			this._shownYear = selectedValueDate.getFullYear();
+		} else {
+			this._focusDate = new Date(this._today.getFullYear(), this._today.getMonth(), 1);
+			this._shownMonth = this._today.getMonth();
+			this._shownYear = this._today.getFullYear();
+		}
```
dlockhart: Nit: this could be re-written as:
```
const date = this.selectedValue ? parseDate(this.selectedValue) : this._today;
this._focusDate = new Date(date.getFullYear(), date.getMonth(), 1);
this._shownMonth = date.getMonth();
this._shownYear = date.getFullYear();
```

components/calendar/calendar.js
```diff
@@ -0,0 +1,480 @@
...
+					if (numDaysChange > 6) {
+						numDaysChange -= daysInWeek;
+					}
+				}
+				preventDefault = true;
+				break;
+			} case keyCodes.PAGEUP: {
+				if (e.shiftKey) {
+					// Changes the grid of dates to the previous Year.
+					// Sets focus on the same day of the same week. If that day does not exist, then moves focus to the same day of the previous or next week.
```
dlockhart: Is the page up/down stuff TODO?

margaree: It is, though it's in the design as a nice to have so it might be better to leave it out of the code completely and then add it if/when it's getting implemented.

components/calendar/calendar.js
margaree: So I _think_ this should now work with other timezones. It will now parse the ISO date time string, convert to local, and then use local as `selectedValue`. I think the rest of the code should be fine. For example, If ISO string is `2019-08-01T20:00:00Z` and timezone is in Australia, then we would expect selected value to be August 2 (which seems to work), so we have `parseDate` return that, and the focus date becomes that, etc. When a value is selected, it would then `selectedValue` to convert from local date to UTC.

The weird thing I'm not sure about here is the affect of time on output/changing `selectedValue`. I have `calendar.js` currently using 12 as the hour when creating a date to keep things consistent (though maybe 0 is a better choice) but this would potentially be affected by time picker as well. Not sure if this should be considered as part of this PR or in a subsequent one when the date/time work starts to combine.

Hopefully this all made sense. Timezones seem to get confusing (at least to me!)

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+}
+
+export function getPrevMonth(month) {
+	return (month === 0) ? 11 : (month - 1);
+}
+
+export function getToday() {
+	const val = new Date().toISOString();
+	const re = /([0-9]{4})-([0-9]{2})-([0-9]{2})T([0-9]{1,2}):([0-9]{1,2}):([0-9]{1,2})/;
+	const match = val.match(re);
```
dlockhart: Why do you need the regular expression here to extra the different date parts? Couldn't you just use the `Date` methods? Oh I guess it's because it'll be in local time but `toISOString` converts it to UTC?

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+		hours: parseInt(match[4]),
+		minutes: parseInt(match[5]),
+		seconds: parseInt(match[6])
+	};
+
+	const valInLocalDateTime = convertUTCToLocalDateTime(dateTime);
+	return new Date(valInLocalDateTime.year, valInLocalDateTime.month - 1, valInLocalDateTime.date);
+}
+
+export function parseDate(val) {
```
dlockhart: Nit: I might call this `parseDateValue` to distinguish it from Intl's `parseDate`.

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+	if (!val) return null;
+	const re = /([0-9]{4})-([0-9]{2})-([0-9]{2})/;
+	const match = val.match(re);
+	if (!match || match.length !== 4) {
+		throw new Error('Invalid selected-value date input: Expected format is YYYY-MM-DD');
```
dlockhart: Since this can throw, does the code that calls this need to handle the exceptions at all, or is it mostly dev-user error cases where you want them to see the exceptions? Basically I want to confirm that there's no way for end users to cause errors in the console, only developers.

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+	_getFocusDateElement() {
+		return this.shadowRoot.querySelector(`div[data-date="${this._focusDate.getDate()}"][data-month="${this._focusDate.getMonth()}"][data-year="${this._focusDate.getFullYear()}"]`);
+	}
+
+	_onDateSelected(e) {
+		const selectedDate = e.composedPath()[0];
+		const year = selectedDate.getAttribute('data-year');
+		let month = selectedDate.getAttribute('data-month');
+		let date = selectedDate.getAttribute('data-date');
+		this._focusDate = new Date(year, month, date);
```
dlockhart: Do these need to be parsed before going into `Date`? I guess it can handle them haha.

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+		month = (parseInt(month) + 1).toString();
+		if (month.length < 2) {
+			month = `0${month}`;
+		}
+		if (date.length < 2) {
+			date = `0${date}`;
+		}
+
+		this.selectedValue = [year, month, date].join('-');
```
dlockhart: Nit: Could be handy to centralize this in a `formatSelectedValue()` that takes a `Date` (or a month/year/day). Basically the opposite if your `parseDate`.

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+			this._focusDate = new Date(selectedValueDate.getFullYear(), selectedValueDate.getMonth(), selectedValueDate.getDate());
+		} else {
+			this._focusDate = new Date(this._shownYear, this._shownMonth, 1);
+		}
+	}
+
+	_updateFocusDateOnKeyDown(numDays) {
+		const oldFocusDate = new Date(this._focusDate);
+		this._focusDate.setDate(this._focusDate.getDate() + numDays);
+		this._focusDate = new Date(this._focusDate.getFullYear(), this._focusDate.getMonth(), this._focusDate.getDate());
```
dlockhart: Nit: I might combine these 2 lines so you're not changing the date value and then replacing the object one after the other:

    this._focusDate = new Date(
        this._focusDate.getFullYear(),
        this._focusDate.getMonth(),
        this._focusDate.getDate() + numDays
    );

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,272 @@
...
+			});
+
+		});
+	});
+
+	const tabToDates = async function() {
+		await page.keyboard.press('Tab');
+		await page.keyboard.press('Tab');
+		await page.keyboard.press('Tab');
+	};
```
dlockhart: Normally I don't think it's super valuable to do visual diffs in every language, but in this case since there's so much rendering that depends on the locale settings I think it might be worth it? Not every permutation in every language, but just a standard calendar view in each one maybe?

components/calendar/calendar-styles.js
```diff
@@ -0,0 +1,83 @@
+import '../colors/colors.js';
+import { css } from 'lit-element/lit-element.js';
+
```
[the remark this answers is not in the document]
margaree: Nope it's just for code organization. I'll move it in.

components/calendar/calendar.js
```diff
@@ -0,0 +1,488 @@
...
+}
+
+export function checkIfDatesEqual(date1, date2) {
+	if (!date1 || !date2) return false;
+	return date1.getTime() === date2.getTime();
+}
+
+export function getDatesInMonthArray(shownMonth, shownYear) {
+	// checking for undefined because month can be 0
+	if (shownMonth === undefined || !shownYear) return [];
```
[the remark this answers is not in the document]
margaree: It was because when `render` is initially run, `_shownMonth` and `_shownYear` are undefined

components/calendar/calendar.js
margaree: My thought was that it should only be the dev-user error cases. Potentially we could deal with them, though I think it would be important for devs to know that they did something wrong.

components/calendar/calendar.js
margaree: We could - it seems to have worked so far

components/calendar/calendar.js
margaree: Yeah exactly

components/calendar/calendar.js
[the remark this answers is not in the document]
dlockhart: Hmm, I guess then could `render()` better handle that? Like if either of those are `undefined` then it probably wants to render nothing?

components/calendar/calendar.js
dlockhart: Cool I'm good with throwing then if it's only dev-user error :)

components/calendar/test/calendar.visual-diff.js
margaree: I'm struggling to get this to work. Have you seen an example of this working? Basically I just can't seem to be able to update the language and then have that be what's reflected on the visual-diff page (it still shows english). The only way that seems like it would currently work (unless I've missed something) is to have a new `.visual-diff.html` page for each `lang` which has the corresponding `lang` set on `html` and that doesn't seem ideal (though it there isn't a better way then that is fairly straightforward to do).

dlockhart: I would expect something like this to work:

```
import { getDocumentLocaleSettings, supportedLocales } from '@brightspace-ui/intl/lib/common.js';

describe('Test Suite', () => {

	const documentLocaleSettings = getDocumentLocaleSettings();
	afterEach(() => documentLocaleSettings.reset();

	supportedLocales.forEach((lang) => {
		it(`should work for ${lang}`, () => {
			documentLocaleSettings.language = lang;
			await oneEvent(elem, 'd2l-localize-behavior-language-changed');
			// take visual diff
		});
	});

});
```

Not sure if you need the `await oneEvent(...)` or not.

margaree: I don't think we can currently use `import` with `visual-diff` so I don't think we can use changing `documentLocaleSettings`. I have mostly tried different variants of `await page.evaluate(() =>document.querySelector('html').setAttribute('lang', 'ar'));` and some other options I have found online but haven't had luck so far.

dlockhart: Ah, right. What about combining the `setAttribute` to change the language with the `await` for the `d2l-localize-behavior-language-changed` event?

dlockhart: If you can get this working, this is probably something we could refactor into the visual-diff code to make it easier for others. @dbatiste FYI.

margaree: Ooo I figured it out. Updating `lang` in this way does actually trigger all the right property updates etc. but because `calendarData` is already instantiated for `en`, it does not get updated. I can listen for that event in `calendar.js` and update `calendarData`

dlockhart: Nice! 🎉

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,302 @@
...
+		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+	});
+
+	describe('localization', () => {
+
+		after(async() => {
+			await page.evaluate(() => document.querySelector('html').setAttribute('lang', 'en'));
+		});
+
+		[
```
dbatiste: What's the rationale for screenshots of all the languages? Is this already covered by localization tests?

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,302 @@
...
+			'ko',
+			'nl',
+			'pt',
+			'sv',
+			'tr',
+			'zh',
+			'zh-tw'
+		].forEach((lang) => {
+			it(`works correctly in ${lang}`, async function() {
+				await page.evaluate(lang => document.querySelector('html').setAttribute('lang', lang), lang);
```
dbatiste: This looks fairly straightforward, but is this what you were suggesting possibly moving to visual-diff as a helper? If so, sounds reasonable to me.

components/calendar/test/calendar.visual-diff.js
dlockhart: Mostly that the calendar changes dramatically based on the language -- day headers, first-day-of-the-week most notably. Normally we wouldn't care about multiple languages, but I feel like in calendar's case it's important.

.travis.yml
```diff
@@ -1,5 +1,23 @@
...
+      - chinese*
+      # Korean
+      - language-pack-ko
+      - korean*
+      # Fonts
+      - fonts-arphic-ukai
+      - fonts-arphic-uming
+      - fonts-ipafont-mincho
+      - fonts-ipafont-gothic
+      - fonts-unfonts-core
```
dlockhart: What are these?

components/calendar/test/calendar.test.js
```diff
@@ -0,0 +1,528 @@
...
+			});
+
+			it('should return expected day in America/Toronto timezone', () => {
+				expect(getToday()).to.deep.equal(new Date(2018, 1, 12));
+			});
+
+			it('should return expected day in Australia/Eucla timezone', () => {
+				documentLocaleSettings.timezone.identifier = 'Australia/Eucla';
+				expect(getToday()).to.deep.equal(new Date(2018, 1, 13));
+				documentLocaleSettings.timezone.identifier = 'America/Toronto';
```
dlockhart: Shouldn't need to set this here since you're doing it in `afterEach`.

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,302 @@
...
+
+		it('has correct style when today selected', async function() {
+			const rect = await visualDiff.getRect(page, '#today-selected');
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+
+		it('has correct hover style on non-selected-value', async function() {
+			await page.$eval(firstCalendarOfPage, (calendar) => {
+				const date = calendar.shadowRoot.querySelector('div[data-date="20"]');
+				date.classList.add('d2l-calendar-date-hover');
```
dbatiste: I guess you were having some trouble getting hover style to show properly without manually applying the class?

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,302 @@
...
+			'es',
+			'fr',
+			'ja',
+			'ko',
+			'nl',
+			'pt',
+			'sv',
+			'tr',
+			'zh',
+			'zh-tw'
```
dlockhart: The `common.js` in Intl exports an array of all the languages that you could use here instead.

components/calendar/test/calendar.visual-diff.js
```diff
@@ -0,0 +1,302 @@
...
+				await page.evaluate(lang => document.querySelector('html').setAttribute('lang', lang), lang);
+				const rect = await visualDiff.getRect(page, firstCalendarOfPage);
+				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+			});
+		});
+	});
+
+	describe('style', () => {
+		afterEach(async() => {
+			await page.reload();
```
dbatiste: Reloading the page is generally the safest approach, but is probably also the slowest.  If the time of the tests really blows up, one option is to reset the component between tests, by reseting focus and removing such classes - I've done that elsewhere... it's not bullet proof and may trip people up if they make changes to the tests, but a compromise in effort to reduce testing time.

components/calendar/test/calendar.visual-diff.js
margaree: yup exactly

.travis.yml
margaree: the OS travis uses does not have asian language fonts, so the visual diffs just had empty square for these languages. This was what google said needed to be added.

components/calendar/test/calendar.visual-diff.js
margaree: I don't think this will work because of import not working in visual-diff

components/calendar/test/calendar.visual-diff.js
margaree: I think the helper was more for if it ended up being a complicated solution, though it turned out not to be, but yeah I can add this as a helper.

dbatiste: Up to you... pretty simple, not sure how often this will come up.

PR #331, opened 2020-01-20
Update dialog mixin to attempt connection to ifrau.

dbatiste:
* connect to ifrau
* if connection to ifrau, then show/remove backdrop in outer frame as needed
* get outer frame layout context to position dialog


dlockhart:
Not sure how feasible it is, but it might be nice to have some unit tests that exercise some of this new ifrau code... you could mock out ifrau and then assert at least that the dialog opens in the place you expect and doesn't error out. I could see us accidentally breaking this pretty easily in the future.

dbatiste:
Sounds good re test - am going to try and do this visual-diff tests.

dbatiste:
I was able to capture the bulk of the behavior through a mock of the ifrau service.

AllanKerr:
The visual diff screenshot names prevent pulling `master` on Windows since `>` and `<` are invalid filename characters on Windows.

dbatiste:
Gross - thanks.  I'll put together a change.

PR #330, opened 2020-01-20
adding a "future enhancement" section to the component READMEs

dlockhart:
As discussed in our retro.

We're hoping that calling out the things we know we want to add in the future will encourage folks to contribute those enhancements instead of building them outside of here.

PR #334, opened 2020-01-21
Port d2l-status-indicator from Polymer to Lit

AllanKerr:
# Changes
- Add support for `bold` status-indicators with white text and colored backgrounds.
bold-indicators


components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+			state: {
+				type: String,
+				reflect: true
+			},
+			/**
+			 * Text for status indicator.
+			 */
+			text: {
+				type: String,
+				value: null
```
dlockhart: `value` isn't a thing in Lit -- you set the default value if required in the constructor. In this case though, I think text can just be left `undefined`.

components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+			},
+			/**
+			 * Use when the status is very important and needs to have a lot of prominence.
+			 */
+			bold: {
+				type: Boolean,
+				value: false
```
dlockhart: Similarly, the `value` should be set in the constructor. But this should be reflected since you're using CSS attribute selectors that require it.

components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+			:host([bold][state="success"]) {
+				background-color: var(--d2l-color-olivine-minus-1);
+			}
+			:host([bold][state="alert"]) {
+				background-color: var(--d2l-color-cinnabar);
+			}
+			:host([bold][state="none"]),
+			:host([bold][state="null"]) {
+				background-color: var(--d2l-color-ferrite);
+			}
```
dlockhart: Love how easy attribute selectors make this!

components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+		`;
+	}
+
+	constructor() {
+		super();
+		this.state = 'default';
```
dlockhart: Here you can default `bold` to `false`.

components/status-indicator/test/status-indicator.html
```diff
@@ -0,0 +1,100 @@
...
+
+			describe('subtle', () => {
+
+				beforeEach(async() => {
+					statusIndicator = fixture('subtle');
+					await statusIndicator.updateComplete;
+				});
+
+				it('passes all axe tests', async() => {
+					await runAxe(statusIndicator);
```
dlockhart: I think we'll want to run the aXe tests for each possible state + bold value, just to make sure all the colour contrast passes. Should be able to do it in a single test though using some fancy `forEach`-ing.

components/status-indicator/test/status-indicator.html
```diff
@@ -0,0 +1,100 @@
...
+					await statusIndicator.updateComplete;
+					expect(statusIndicator.getAttribute('state')).to.equal('none');
+				});
+
+				it('should be subtle', () => {
+					expect(statusIndicator.hasAttribute('bold')).to.be.false;
+				});
+
+			});
+
```
dlockhart: I know most of these tests are holdovers from the old components... except for the aXe ones, I question the value of the others now that we have visual diff tests.

components/status-indicator/test/status-indicator.visual-diff.js
```diff
@@ -0,0 +1,42 @@
...
+		{ title: 'no-state-subtle', fixture: '#no-state-subtle' },
+		{ title: 'default-subtle', fixture: '#default-subtle' },
+		{ title: 'success-subtle', fixture: '#success-subtle' },
+		{ title: 'alert-subtle', fixture: '#alert-subtle' },
+		{ title: 'none-subtle', fixture: '#none-subtle' },
+		{ title: 'no-state-bold', fixture: '#no-state-bold' },
+		{ title: 'default-bold', fixture: '#default-bold' },
+		{ title: 'success-bold', fixture: '#success-bold' },
+		{ title: 'alert-bold', fixture: '#alert-bold' },
+		{ title: 'none-bold', fixture: '#none-bold' },
```
dlockhart: Nit: looks like the input here could be simplified since the ID is just the title prepended by a `#` in all cases.

components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+		return {
+			/**
+			 * State of status indicator to display. Valid values are: 'default', 'success', 'alert', and 'none'.
+			 */
+			state: {
+				type: String,
+				reflect: true
+			},
+			/**
+			 * Text for status indicator.
```
dbatiste: So far we've been avoiding adding these comments in the code unless they are absolutely needed.  Usually they end up bloating the code unnecessarily.

components/status-indicator/status-indicator.js
```diff
@@ -0,0 +1,99 @@
...
+			:host([state="success"]) {
+				border-color: var(--d2l-color-olivine-minus-1);
+				color: var(--d2l-color-olivine-minus-1);
+			}
+			:host([state="alert"]) {
+				border-color: var(--d2l-color-cinnabar);
+				color: var(--d2l-color-cinnabar);
+			}
+			:host([state="none"]),
+			:host([state="null"]) {
```
dbatiste: I don't think we should have `state="null"`... ?

components/status-indicator/test/status-indicator.visual-diff.js
```diff
@@ -0,0 +1,42 @@
...
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await browser.newPage();
+		await page.setViewport({width: 800, height: 800, deviceScaleFactor: 2});
+		await page.goto(`${visualDiff.getBaseUrl()}/components/status-indicator/test/status-indicator.visual-diff.html`, {waitUntil: ['networkidle0', 'load']});
+		await page.bringToFront();
+	});
+
+	beforeEach(async() => {
+		await visualDiff.resetFocus(page);
```
dbatiste: I don't _think_ you should need this `resetFocus`, since focus is not being applied to the element being diff'd. This method is helpful in cases where the test applies focus to the element, but we're not refreshing the page, and we need a convenient way to remove focus from the element for subsequent tests.

components/status-indicator/status-indicator.js
AllanKerr: I did consider removing it. I only left it to guarantee backwards compatibility. I'd be happy to remove it if we think it'd be safe to do so.

dlockhart: It's there because it used to be a valid state and was removed. If we can somehow confirm that no one's using it, we could remove.

components/status-indicator/README.md
```diff
@@ -0,0 +1,49 @@
...
+<d2l-status-indicator state="success" text="complete" bold></d2l-status-indicator>
+<d2l-status-indicator state="alert" text="overdue" bold></d2l-status-indicator>
+<d2l-status-indicator state="none" text="closed" bold></d2l-status-indicator>
+```
+
+**Properties:**
+
+- `state` (String, default: `'default'`): State of status indicator to display. Can be one of  `default`, `success`, `alert` , `none`
+- `text` (required, String): The text that is displayed within the status indicator
+- `bold` (Boolean): Use when the status is very important and needs to have a lot of prominence
```
dlockhart: Could you throw a "Future Enhancements" section at the bottom similar to what's [in this PR](#330)?

PR #335, opened 2020-01-21
Update test names for vdiff

dbatiste:
Updating since `<` and `>` are invalid in filenames on Windows.


dlockhart:
Next time if you want to bypass the vdiff commit stage for changes like this, you can just rename the files yourself and push those changes -- the vdiff will pass first time around.

PR #337, opened 2020-01-22
Fix issue with menu-items created through dom-repeat not working

margaree:


components/menu/menu.js
```diff
@@ -67,6 +67,10 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 		this._labelChanged();
 
 		this._onMenuItemsChanged();
+		const slot = this.shadowRoot.querySelector('slot');
+		slot.addEventListener('dom-change', () => {
```
dlockhart: You're interested in finding out when the distributed nodes change? I think `dom-change` is a Polymer thing... does the `slotchange` standard event work?

components/menu/menu.js
```diff
@@ -156,7 +160,7 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 			items.unshift(returnItem);
 		}
 		return items.filter((item) => {
-			const role = item.role;
+			const role = item.role || item.getAttribute('role');
```
dlockhart: After mucking around a bit, it doesn't seem like the ARIA role attribute reflects to a property -- although we've set up our Lit element so that the property reflects to the attribute. That's likely why you're seeing this. I wonder if we should just always use `getAttribute('role')` in this code then?

components/menu/menu.js
margaree: Yup - looks like `slotchange` will work!

components/menu/menu.js
margaree: In theory I think we could use just `getAttribute('role')` but with the template-menu demo (https://github.com/BrightspaceUI/menu/blob/master/demo/template-menu.html) `getAttribute('role')` only seems to be working on the first menu item, and the rest return `null`, where as `item.role` works (but then there are other cases like in `user-switcher` where `item.role` does not work and we need `item.getAttribute`). Not sure why that's happening - digging into it a bit.

dlockhart: That's very weird... maybe a Polymer-y dom-repeat-y thing?

margaree: Yeah I'm thinking so

margaree: Adding a `requestAnimationFrame` inside of the `slotchange` event listener seems to fix that so we can just use `getAttribute('role')`

dbatiste: Re the `requestAnimationFrame()`... maybe the role is not reflected to the attribute immediately upon `slotchange`... Lit needs a change to run its code to do that.  

You could also try `queueMicrotask()`, since that's what what Lit element uses for a bunch. You might not need to wait a full frame. It's not supported by Edge or IE11, but if you decide to try it and use it, there is a [super simple polyfil (few lines of code)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask#When_queueMicrotask_isnt_available) for it. It uses a Promise that resolves immediately (since that is microtask), and falls back to setTimeout (full task) when Promises aren't supported). If you do include it, I'd put it in our helpers next to `requestIdleCallback`. It's something I've been wanting to try and use, in effort to tighten up timing of some code I've written/writing.

components/menu/menu.js
```diff
@@ -67,6 +67,12 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 		this._labelChanged();
 
 		this._onMenuItemsChanged();
+		const slot = this.shadowRoot.querySelector('slot');
```
dbatiste: Be sure to check this in Edge - recently a similar query (for a slot element) was failing for dialog footer in Edge, and I think we ended up having to query using a class on the slot instead of the tag name.  Strange... anyway, definitely worth checking if you haven't already.

margaree: Ah yup - so edge works without the `requestAnimationFrame` in there (but then we need to add back the `role = item.role || item.getAttribute('role')` and so no longer works with the most recent change.

components/menu/menu.js
margaree: `queueMicrotask` actually works nicely! Thanks!

components/menu/menu.js
dlockhart: Querying by tag name alone should be fine, the dialog case was `.some-class slot` which I'm guessing the shady DOM polyfill wasn't able to handle.

helpers/README.md
```diff
@@ -77,6 +77,17 @@ requestIdleCallback((deadline) => {
 }, { timeout: 1000 });
 ```
 
+## queueMicrotask
+
+A polyfill for [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask).
```
dbatiste: Looks good.  The only think I might add is a link to an article that provides a bit more info on microtasks (such as the [mozilla one](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide)). Not a big deal and devs can go searching if they are curious to know more, but for most, they are this mysterious thing that presenters mention at conferences.

components/menu/menu.js
margaree: So unfortunately after more cross-browser testing it seems this solution doesn't actually work for Safari. I have been able to get this solution working for all browsers if I replace `queueMicrotask` with `requestAnimationFrame` in `_onMenuItemsChanged`.

dlockhart: Well it was a valiant effort nonetheless!

PR #344, opened 2020-01-30
Use open-wc tools

dlockhart:
This PR uses the [open-wc](https://open-wc.org/)'s [es-dev-server](https://open-wc.org/developing/es-dev-server.html#getting-started) which replaces `polymer serve` and their [testing tools](https://open-wc.org/testing/testing-karma.html#getting-started), which will replace `polymer test`.

I only converted one test file over just to give you an idea of what things will be like -- much simpler IMO. The open-wc Karma tests are set up to run alongside the old Polymer ones for now, and then once we convert everything over we can remove the Polymer stuff for good.

You'll notice that the open-wc tools come with their own Chai-based a11y testing (which uses aXe), so we'll be able to simplify our stuff once everything's converted.


mixins/localize-mixin.js
```diff
@@ -1,6 +1,5 @@
 import {getDocumentLocaleSettings} from '@brightspace-ui/intl/lib/common.js';
-import IntlMessageFormat from 'intl-messageformat/src/main.js';
-window.IntlMessageFormat = IntlMessageFormat;
+import IntlMessageFormat from 'intl-messageformat';
```
dlockhart: This shouldn't be in here, will fix.

README.md
```diff
@@ -70,18 +70,12 @@ After cloning the repo, run `npm install` to install dependencies.
...
-If you don't have it already, install the [Polymer CLI](https://www.polymer-project.org/3.0/docs/tools/polymer-cli) globally:
-
-```shell
-npm install -g polymer-cli
-```
-
 ### Running the demos
 
-To start a [local web server](https://www.polymer-project.org/3.0/docs/tools/polymer-cli-commands#serve) that hosts the demo page and tests:
+To start a local web server that hosts the demo page and tests:
```
Scy-D2L: Would it still be helpful to have [a link to the server](https://open-wc.org/developing/es-dev-server.html#startserver) in here?

components/button/test/button-icon.test.js
```diff
@@ -0,0 +1,31 @@
+import '../button-icon.js';
+import { expect, fixture, oneEvent } from '@open-wc/testing';
+
+const basicFixture = '<d2l-button-icon icon="d2l-tier1:gear" text="Icon Button"></d2l-button-icon>';
+const disabledFixture = '<d2l-button-icon disabled icon="d2l-tier1:gear" text="Icon Button"></d2l-button-icon>';
+
+describe('d2l-button-icon', () => {
+
+	it('dispatches click event when clicked', async() => {
+		const el = await fixture(basicFixture);
```
svanherk: Does this handle the waiting for `updateComplete` for us?  That's super nice

dlockhart: It does indeed! Although I need to find a way to handle components using the localize mixin, as they need to wait a bit longer.

README.md
dlockhart: Good call, done.

karma.conf.sauce.js
```diff
@@ -0,0 +1,50 @@
...
+	safari: {
+		base: 'SauceLabs',
+		browserName: 'safari',
+		platform: 'OS X 10.13'
+	}
+};
+
+module.exports = config => {
+	config.set(
+		merge(createDefaultConfig(config), {
```
dbatiste: Is it possible to use `Object.assign()`?  I think nodejs has supported it for a while.

dlockhart: I can give it a try. This is just following [their recommendation](https://open-wc.org/testing/testing-karma.html#manual).

components/button/test/button-icon.test.js
dbatiste: Could wait for a localize event.  For composite/app components, they would not know how many events to wait for though.  The async container mixin pattern comes to mind, though it was not written with localization specifically in mind, though it is applicable.

karma.conf.sauce.js
dbatiste: Cool... yeah I think it would be nice to avoid if possible.

dlockhart: I don't think they're equivalent... deepmerge does a nested merge vs. `Object.assing()` only goes one level deep.

dbatiste: Ah 😞 .  I think those open-wc guys are pretty sharp, I was worried about that.

karma.conf.sauce.js
```diff
@@ -0,0 +1,55 @@
...
+		platform: 'OS X 10.13'
+	},
+	safari: {
+		base: 'SauceLabs',
+		browserName: 'safari',
+		platform: 'OS X 10.13'
+	},
+	edge: {
+		base: 'SauceLabs',
+		browserName: 'microsoftedge',
```
dlockhart: Oh I was able to enable old-Edge testing for the Karma stuff! 🎉

PR #346, opened 2020-01-30
Initial Dropdown Polymer to Lit Conversion

AllanKerr:
This pull request includes the conversion of the following files from Polymer to Lit:

# Components
- [d2l-dropdown](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown.js)
- [d2l-dropdown-opener-behavior](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-opener-behavior.js)
- [d2l-dropdown-content](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-content.js)
- [d2l-dropdown-content-behavior](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-content-behavior.js)

# Unit Tests
- [dropdown-content](https://github.com/BrightspaceUI/dropdown/blob/master/test/dropdown-content.html)


components/dropdown/test/dropdown-content.html
```diff
@@ -0,0 +1,303 @@
...
+					});
+					content.addEventListener('d2l-dropdown-close', () => {
+						expect(content.opened).to.be.false;
+						done();
+					});
+					content.setAttribute('opened', true);
+				});
+
+			});
+
```
AllanKerr: The [width](https://github.com/BrightspaceUI/dropdown/blob/master/test/dropdown-content.html) tests were removed and converted to visual diff tests.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+
+		if (newValue) {
+			if (!this.renderContent) {
+				await this.forceRender();
+			} else {
+				await this.updateComplete;
+			}
+
+			doOpen();
+
```
AllanKerr: [The mutation observer](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-content-behavior.js#L554) will be re-added in the pull request that converts the `d2l-dropdown-tabs` and `d2l-dropdown-menu` components.

components/dropdown/dropdown-content-styles.js
```diff
@@ -0,0 +1,153 @@
+import { css } from 'lit-element/lit-element.js';
+
+export const dropdownContentStyles = css`
+			:host {
```
dbatiste: Nit: looks like these styles can be indented a couple stops.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,145 @@
+export const DropdownOpenerMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			isDropdownOpener: {
+				type: Boolean
+			},
+			noAutoOpen: {
```
dbatiste: We may need to initialize this to `false` in the constructor.  I think there are certain instances where the linter warn/error if referencing uninitialized booleans.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,145 @@
+export const DropdownOpenerMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			isDropdownOpener: {
```
dbatiste: I think we should follow the naming convention of omitting the `is` off of this boolean.

components/dropdown/dropdown.js
```diff
@@ -0,0 +1,30 @@
...
+import { dropdownOpenerStyles } from './dropdown-opener-styles.js';
+
+class Dropdown extends DropdownOpenerMixin(LitElement) {
+
+	static get properties() {
+		return {};
+	}
+
+	static get styles() {
+		return [dropdownOpenerStyles];
```
dbatiste: Nit: I don't think this needs to be an array.  I think we can just return `dropdownOpenerStyles` in this case.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	__getContentElement() {
+		return this.shadowRoot.querySelector('slot').assignedNodes().find(
+			node => node.nodeType === 1 && node.nodeName === 'D2L-DROPDOWN-CONTENT'
+		);
+	}
+
+	__onKeyPress(e) {
+		if (e.keyCode !== 13) {
```
dbatiste: Nit: feel free to condense this if you wish:
```
if (e.keyCode !== 13) return;
if (this.noAutoOpen) return;
```
or 
```
if (e.keyCode !== 13 || this.noAutoOpen) return;
```

components/dropdown/dropdown-content-styles.js
```diff
@@ -0,0 +1,153 @@
...
+
+			:host .d2l-dropdown-content-top-scroll {
+				box-shadow: 0 3px 3px 0 rgba(0, 0, 0, 0.05);
+			}
+
+			:host .d2l-dropdown-content-bottom-scroll {
+				box-shadow: 0 -3px 3px 0 rgba(0, 0, 0, 0.05);
+			}
+
+			:host-context([dir="rtl"]) {
```
dbatiste: These `:host-context` and `:host[:dir(rtl)]` selectors should be replaced with the following.  The `RtlMixin` should be extended in the content classes in order to sprout this attribute on the host.
```
:host([dir="rtl"]) {
    ...
}
```
The `:host-context` is not supported going forward, and the `:dir` is a Polymer 3 thing that required a special import.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
+import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
+import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
+import { getComposedActiveElement, getFirstFocusableDescendant, getPreviousFocusableAncestor } from '../../helpers/focus.js';
+import { html } from 'lit-element/lit-element.js';
+import { styleMap } from 'lit-html/directives/style-map.js';
+
+export const DropdownContentMixin = superclass => class extends superclass {
```
dbatiste: Will need to import and extend the `RtlMixin` here (as per other comment in the styles file).

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+				type: Object,
+			},
+			opened: {
+				type: Boolean,
+				reflect: true
+			},
+			/**
+			 * Private.
+			 */
+			openedAbove: {
```
dbatiste: I think this should become `_openedAbove` since it is private.  I'd probably also reflect as `_opened-above` so there's no confusion if anyone if looking at the HTML. If updating the attribute, don't forget to update the styles that reference it.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+	}
+
+	async __openedChanged(newValue) {
+
+		this.__previousFocusableAncestor =
+			newValue === true
+				? getPreviousFocusableAncestor(this, false, false)
+				: null;
+
+		this.__isRTL = (getComputedStyle(this).direction === 'rtl');
```
dbatiste: I think the `__isRTL` should no longer be needed.  We'll need to use the [RtlMixin](https://github.com/BrightspaceUI/core/blob/master/mixins/rtl-mixin.js) to properly handle the styles, and it will sprout `dir` on the host.  So we can avoid the `getComputedStyle` call (which can sometimes be costly) by using `this.getAttribute('dir')`. Referencing the property may also be an option.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+			}
+		};
+	}
+
+	constructor() {
+		super();
+
+		this.__opened = false;
+		this.__content = null;
+		this.__isRTL = false;
```
dbatiste: We'll be able to get rid of this.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+
+		window.addEventListener('resize', this.__onResize);
+		document.body.addEventListener('focus', this.__onAutoCloseFocus, true);
+		document.body.addEventListener('click', this.__onAutoCloseClick, true);
+
+		this.addEventListener('d2l-dropdown-close', this.__onClose);
```
dbatiste: It looks like we could move wire-up to `d2l-dropdown-close` and `d2l-dropdown-position` to `firstUpdated`, thereby avoiding the explicit cleanup. Ya?

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+		this.__opened = val;
+		this.requestUpdate('opened', oldVal);
+		this.__openedChanged(val);
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this.__content = this.__getContentContainer();
+		this.__content.addEventListener('scroll', this.__toggleScrollStyles);
```
dbatiste: Looks like this `scroll` handler could be wired up down in `_renderContentContainer` in the template... `@scroll...`.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		window.removeEventListener('resize', this.__onResize);
+		document.body.removeEventListener('focus', this.__onAutoCloseFocus, true);
+		document.body.removeEventListener('click', this.__onAutoCloseClick, true);
+		this.removeEventListener('d2l-dropdown-close', this.__onClose);
+		this.removeEventListener('d2l-dropdown-position', this.__toggleScrollStyles);
+		this.__content.removeEventListener('scroll', this.__toggleScrollStyles);
```
dbatiste: Note: there would be no need to remove this handler here since it is currently wired up in `firstUpdated` and would be cleaned up when the node is garbage collected since it is attached to a node in this element's tree.  If the component were disconnected and then reconnected (ex. moved in the DOM) this would currently result in broken scroll behavior because the handler would not be re-wired in the connected callback.  However, per previous comment, the wire-up can be moved into the handler... either way, we should be able to remove this line.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+			this.close();
+		} else {
+			this.open(applyFocus);
+		}
+	}
+
+	/**
+	 * Private.
+	 */
+	scrollTo(scrollTop) {
```
dbatiste: If these `scrollTo` and `height` methods are really private, we should look at renaming them the be `_scrollTo()` and `height()`.

AllanKerr: I never actually could find anywhere that these two where being used. The only reason I added them in was due to being nervous that they were being used somewhere in the LMS that I missed and removing them would unintentionally result in broken behavior.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+				clearDismissible(this.__dismissibleId);
+				this.__dismissibleId = null;
+			}
+
+			this.dispatchEvent(new CustomEvent('d2l-dropdown-close', { bubbles: true, composed: true }));
+
+		}
+	}
+
+	__position(ignoreVertical, contentRect) {
```
dbatiste: I think we need to refactor a bit here so that we're not modifying the DOM directly - basically we should never really update the DOM outside of `render` (or it's helpers).  I think we should create private properties (or a collection of them in an object), and set them from here.  The `render` method should apply those style properties.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,556 @@
...
+
+		if (isOverflowing || this.__content.scrollHeight > maxHeight) {
+			this.__content.style.overflowY = 'auto';
+		} else {
+			/* needed for IE */
+			this.__content.style.overflowY = 'hidden';
+		}
+	}
+
+	__toggleScrollStyles() {
```
dbatiste: Same thing for this method - I think we need to refactor slightly so that we update properties, and then these classes get added/removed via `render`.  Lit will batch these updates and they will all happen in its render phase.

components/dropdown/dropdown-content-mixin.js
AllanKerr: Sounds good to me, I debated a bit on reducing risk by keeping it as close to the original as possible versus adding additional properties and moving it into the private `_renderContentContainer` and `_renderWidthContainer` functions.

components/dropdown/dropdown-content-mixin.js
AllanKerr: Talked with @dbatiste , `opened-above` is referenced in the LMS context menu CSS so going to leave this name the same to prevent introducing a breaking change.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,582 @@
...
+
+			await this.__position();
+
+			if (!this.noAutoFocus && this.__applyFocus) {
+				const focusable = getFirstFocusableDescendant(this);
+				if (focusable) {
+					// bumping this to the next frame is required to prevent IE/Edge from crazily invoking click on the focused element
+					requestAnimationFrame(() => {
+						focusable.focus();
+					});
```
dlockhart: Nit: can be simplified to `requestAnimationFrame(() => focusable.focus());`

components/dropdown/dropdown-content-styles.js
```diff
@@ -0,0 +1,148 @@
+import { css } from 'lit-element/lit-element.js';
```
dlockhart: Is this needed by more than 1 thing? If not, we should pull it into `dropdown-content.js`. Otherwise when we finally get unbundled builds working this will be an unnecessary extra request.

components/dropdown/dropdown-content-styles.js
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	:host([opened-above]) {
+		bottom: calc(100% + var(--d2l-dropdown-verticaloffset, 20px));
+		top: auto;
+		-webkit-animation: d2l-dropdown-above-animation 300ms ease;
+		animation: d2l-dropdown-above-animation 300ms ease;
+	}
+
+	:host .d2l-dropdown-content-pointer {
```
dlockhart: The plain `:host` prefixes shouldn't be needed -- the style is already scoped to inside the host by shadow DOM.

components/dropdown/dropdown-content.js
```diff
@@ -0,0 +1,25 @@
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { DropdownContentMixin } from './dropdown-content-mixin.js';
+import { dropdownContentStyles } from './dropdown-content-styles.js';
+
+class DropdownContent extends DropdownContentMixin(LitElement) {
+
+	static get properties() {
+		return {};
```
dlockhart: Could we just exclude `properties` altogether?

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,130 @@
...
+		this.addEventListener('d2l-dropdown-close', this.__onClosed);
+
+		const opener = this.getOpenerElement();
+		const content = this.__getContentElement();
+		if (!opener) {
+			return;
+		}
+		opener.setAttribute('aria-haspopup', 'true');
+		opener.addEventListener('keypress', this.__onKeyPress);
+		opener.addEventListener('mouseup', this.__onMouseUp);
```
dlockhart: Should these listeners be removed at some point?

components/dropdown/dropdown-content-mixin.js
AllanKerr: Looks like `scrollTo` and `height` are used inside `d2l-time-picker`: https://github.com/BrightspaceUI/time-picker/blob/master/d2l-time-picker.js#L292
I couldn't find any other uses but its possible I missed them since they're difficult functions to search for.

components/dropdown/test/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "brightspace/wct-lit-config"
+}
```
dlockhart: Sorry as I know this stuff wasn't ready when you opened the PR -- and feel free to do it in a follow-up, but could you convert these to the new open-wc based tests? Use the button tests as an example.

components/dropdown/dropdown-content-styles.js
AllanKerr: At the moment, no. But it will be pulled in to [`dropdown-menu`](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-menu.js#L21) and [`dropdown-tabs`](https://github.com/BrightspaceUI/dropdown/blob/master/d2l-dropdown-tabs.js#L8) when they get added.

Given how similar they are, it could likely be converted to a component rather than a mixin but for the initial iteration I went keeping it as close to the original as possible to make it easier to review and reduce the risk of accidentally breaking something.

dlockhart: Cool makes sense. If it's needed, it's needed!

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,582 @@
...
+
+export const DropdownContentMixin = superclass => class extends RtlMixin(superclass) {
+
+	static get properties() {
+		return {
+			minWidth: {
+				type: Number,
+				reflect: true,
+				attribute: 'min-width'
+			},
```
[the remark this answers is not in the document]
AllanKerr: I'm a bit nervous that there are places in the LMS that either access them to do additional styling or are really on `getAttribute()`; that being said, I can try and go through and do an audit to see where that could be an issue.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,130 @@
...
+		opener.setAttribute('aria-expanded', (content && content.opened || false).toString());
+	}
+
+	/**
```
dbatiste: Feel free to clean up/remove these comments that are stating the obvious.  :)

components/dropdown/dropdown-opener-mixin.js
dbatiste: The opener _should_ be in this elements composed sub-tree, so this _should_ get cleaned up automatically.  However, it could be in the light-DOM (as is the case when using `d2l-dropdown`), so technically the opener element is subject to any craziness user-land code might do.  For the unpredictable case, we'd need to wire up un `connectedCallback`, and remove in `disconnectedCallback`.

AllanKerr: This was reason I moved it from `firstUpdated` to the `connectedCallback` and added explicit `removeEventListener` calls in `disconnectedCallback`.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -0,0 +1,259 @@
...
+			expect(content.opened).to.be.true;
+		});
+
+		it('closes dropdown when open', async() => {
+			content.toggleOpen();
+			await oneEvent(content, 'd2l-dropdown-open');
+
+			const dropdownClose = oneEvent(content, 'd2l-dropdown-close');
+			content.toggleOpen();
+			await dropdownClose;
```
dlockhart: For cases like this where you need the Promise but can't `await` it until you do something else, I've seen the open-wc examples do this:

    setTimeout(() => content.toggleOpen());
    await oneEvent(content, 'd2l-dropdown-close');

That has the effect of calling `toggleOpen` _after_ the `await` line.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -0,0 +1,540 @@
...
+		changedProperties.forEach((_, propName) => {
+			if (propName === 'verticalOffset') {
+				// for IE11
+				if (window.ShadyCSS) window.ShadyCSS.styleSubtree(this, { '--d2l-dropdown-verticaloffset': `${this.verticalOffset}px` });
+				else this.style.setProperty('--d2l-dropdown-verticaloffset', `${this.verticalOffset}px`);
+			}
+		});
+	}
+
+	_renderContent() {
```
dbatiste: Nit: looks like some method are out of order in here. I am surprised that the [linting rules](https://github.com/Brightspace/eslint-config-brightspace/blob/master/sort-member-config.js) aren't causing failures for it.  I wonder if they are actually still working?

AllanKerr: I did get a linter error for constructor placement that it auto-fixed but other than that I haven't seem any other errors related to method order.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -0,0 +1,132 @@
...
+			return;
+		}
+		if (this.noAutoOpen) {
+			return;
+		}
+		this.toggleOpen(true);
+	}
+
+	__onMouseUp() {
+		if (this.noAutoOpen) {
```
dbatiste: Super minor... if it were me I'd start to tidy up some of these to be one-liners. i.e..
```
if (this.noAutoOpen) return;
```
The original code was written at a time before we relaxed this practice a bit.  Don't feel like you have to, but if I were in here I probably would, simply just to make more code visible without costing readability.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -0,0 +1,259 @@
...
+		});
+
+	});
+
+	describe('toggleOpen', () => {
+
+		it('opens dropdown when closed', async() => {
+			content.toggleOpen();
+			await oneEvent(content, 'd2l-dropdown-open');
+			expect(content.opened).to.be.true;
```
dbatiste: Is this captured by a visual-diff test?

components/dropdown/test/dropdown-content.test.js
```diff
@@ -0,0 +1,259 @@
...
+		});
+
+		it('closes dropdown when open', async() => {
```
dbatiste: Same with this one... is this captured by visual-diff already?

components/dropdown/test/dropdown-content.visual-diff.js
```diff
@@ -0,0 +1,56 @@
...
+
+	after(() => browser.close());
+
+	[
+		'default-width',
+		'min-width',
+		'default-max-width',
+		'max-width',
+		'scroll-bottom-shadow',
+		'top-middle',
```
dbatiste: I think there may be a permutation of positioning that art not covered.  For positioning, I think we should have:

* centered with target (dropdown content is wider than opener)
* centered with target (dropdown content is narrower than opener)
* slide to right (not enough space to center/left edge of page)
* slide to left (not enough space to center/right edge of page)
* ^ all the above with RTL
* open upwards (when not enough space below, but there is above)

I think I would name them according to the behavior.  So for example, I think `top-right` should be named something like `slide-left`.

I'm not quite as familiar with the boundary usage, but I _think_ they are covered here.

components/dropdown/test/dropdown-helper.js
```diff
@@ -0,0 +1,29 @@
...
+		return getEvent(page, selector, 'd2l-dropdown-open');
+	},
+
+	async open(page, selector) {
+		const openEvent = this.getOpenEvent(page, selector);
+		await page.$eval(selector, (dropdown) => {
+			dropdown.toggleOpen();
+		});
+		// 300ms fade-in animation on open
+		await asyncSetTimeout(300);
```
dbatiste: Normally I would handle the animation by waiting for the animation:
```
await page.$eval(selector, (dropdown) => {
        dropdown.querySelector('d2l-dropdown-content').addEventListener('animationend', () => resolve(), { once: true });
        dropdown.toggleOpen();
    });
});
```
Aside, if this approach is used with `transitionend` events, one must be careful to wait for all the properties to animate... the event will fire once for each property being transitioned, which can obviously have different durations.

components/dropdown/test/dropdown-content.visual-diff.js
dbatiste: Just taking a second look at this.  Good job on the images/fixture setup by the way!  

I think the only scenario(s) not covered are the "dropdown content is narrower than opener", and some RTL. 

While the RTL cases are visually opposite of the LTR, the calculation for the RTL is done in a different code path since the property assignment is different, and calc for slide lift/right is flipped, so for complete position coverage we would want RTL captures for those scenarios.

AllanKerr:
> Not sure if DL has any more feedback, but this is look'n good. If there's any other minor things we uncover as the rest of the components are converted, we can always make further adjustments then, given that we're not quite to the point of swapping out the Polymer impl just yet.

I'm expecting little changes here and there to come as I port over the other components that use the opener and content mixins so that sounds good to me!

PR #347, opened 2020-01-31
Polyfill Intl.PluralRules

dlockhart:
With the new `intl-messageformat`, we need to include a polyfill for the `Intl.PluralRules` stuff, which doesn't exist in Edge or older versions of Safari.

### DIFF
--- a/wct.conf.json
+++ b/wct.conf.json
@@ -16,7 +16,7 @@
 				{
 					"browserName": "safari",
 					"platform": "OS X 10.13",
-					"version": "11.1"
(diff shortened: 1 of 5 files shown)


wct.conf.json
```diff
@@ -16,7 +16,7 @@
 				{
 					"browserName": "safari",
 					"platform": "OS X 10.13",
-					"version": ""
+					"version": "11.1"
```
dlockhart: This will be removed once I see Sauce pass with it.

Issue #232, opened 2019-09-20
Try to cache language resources


dlockhart:
Each time `getLocalizeResources` is called in `LocalizeMixin` and `LocalizeStaticMixin`, some work occurs. This work is repeated for every instance of that component, when it only needs to happen once.

We should investigate caching the results of the call for a given set of input languages.

PR #350, opened 2020-01-31
delaying updateComplete promise until localized resources are loaded

dlockhart:
The `open-wc` testing tools will automatically wait for the `updateComplete` Promise on components to resolve before starting the tests. This is super convenient, and will allow us to remove a lot of our `beforeEach` boilerplate code.

Unfortunately, as @mdgbayly noticed, for components using the `LocalizeMixin`, it isn't guaranteed that the localized resources have loaded and the component has rendered localized before `updateComplete` resolves.

Luckily, the Lit docs have as section on [overriding updateComplete](https://lit-element.polymer-project.org/guide/lifecycle#updatecomplete), and even mention using it to wait for child components -- a similar case. So that's what this does -- it delays the `updateComplete` Promise until the localized resources are ready.


components/inputs/input-search.js
```diff
@@ -109,12 +109,6 @@ class InputSearch extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 	}
 
 	render() {
-		requestAnimationFrame(
-			() => this.dispatchEvent(new CustomEvent('d2l-input-search-render', {
-				bubbles: false,
-				composed: false
-			}))
-		);
```
dlockhart: This change meant we don't need this event anymore.

mixins/demo/localize-test-async.js
```diff
@@ -0,0 +1,72 @@
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { LocalizeMixin } from '../../mixins/localize-mixin.js';
```
dlockhart: I broke the demo components into separate static and async versions to simplify things.

mixins/test/localize-mixin.html
```diff
@@ -10,7 +10,7 @@
 		<script src="/node_modules/@polymer/test-fixture/test-fixture.js"></script>
 		<script src="/node_modules/wct-mocha/wct-mocha.js"></script>
 		<script type="module" src="../demo/localize-test.js"></script>
-		<script type="module" src="../demo/localize-test-no-resources.js"></script>
```
dlockhart: This component/fixture weren't actually used, and don't really make sense now that the mixin ONLY does localization.

mixins/test/localize-mixin.html
```diff
@@ -61,74 +112,41 @@
 					});
 				});
 
-				describe('localize', () => {
```
dlockhart: I kept all these tests, but just put them inside the `forEach` that runs them for both static and async versions.

mixins/test/localize-mixin.html
```diff
@@ -61,74 +112,41 @@
...
-					it('should localize term using plurals', () => {
-						const val = elem.localize('plural', {itemCount: 1});
-						expect(val).to.equal('You have 1 item.');
-					});
-				});
 				describe('shouldUpdate tracking', () => {
 
 					it('should pass all changed properties to updated()', (done) => {
-						const elem = fixture('async');
+						const container = fixture('container');
```
dlockhart: Needed to do a bit of hackery here and not use `fixture` -- by creating the component ourselves we can register for events that would normally fire before our fixture gets called.

components/inputs/test/input-search.html
```diff
@@ -37,13 +37,9 @@
 				let elem, input;
 
 				async function getFixture(name) {
-					return new Promise((resolve) => {
```
dbatiste: Aside - the old code didn't look quite right - the `async` was originally not needed.

mixins/localize-mixin.js
```diff
@@ -15,30 +15,42 @@ export const LocalizeMixin = superclass => class extends superclass {
...
-					}
-					this.__language = res.language;
-					this.__resources = res.resources;
-					if (first) {
-						first = false;
-					} else {
-						this._languageChange();
-					}
-				});
+		this.__resourcesLoadedPromise = new Promise((resolve) => {
```
dbatiste: I know this is sort of separate concern, but do you have any appetite for updating the mixin so that the work only gets done once per instance of a component by either storing it on the class (not instance) or in a weak map outside of the class?

dlockhart: I think that's captured in #232?

dbatiste: I think so yep - wasn't sure if we wanted to tackle it while refactoring this...?

dlockhart: Haha I personally do not, but I can take a crack at it in a follow-up.

PR #360, opened 2020-02-05
Us112805/time input textbox

Scy-D2L:
Created the text box for time input per US112805
- Takes a “H:mm:ss” 24-hour clock string input value and renders it in the user’s locale. Does not render the seconds.
- Defaults to 12AM if no value is provided
- Always exposes the current value as “H:mm:ss” string property
- Has a getTime() method that returns an object containing `hour` `minute` and `second`
- When the user changes the value in the input, it’s re-parsed (by our Intl library), the value is set and a "d2l-time-input-changed" event is fired. If it can’t be parsed, the old value is retained (and no event is fired).
- A label must be provided and is rendered above the input. It can be hidden using an attribute.
- Can be disabled

Includes unit/visual-difference tests


components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+			if (isNaN(second) || second < 0 || second > 59) {
+				second = 0;
+			}
+		}
+	}
+	const time = new Date(TODAY.getFullYear(), TODAY.getMonth(), TODAY.getDate(), hour, minute, second);
+	return time;
+}
+
+class InputTime extends RtlMixin(LitElement) {
```
dlockhart: Shouldn't need `RtlMixin`, at least for the time being?

components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+
+	static get properties() {
+		return {
+			disabled: { type: Boolean, reflect: true },
```
dlockhart: Don't think this needs to be reflected.

components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+			minute: time.getMinutes(),
+			second: time.getSeconds()
+		};
+	}
+
+	_getAriaLabel() {
+		if (this.label && this.labelHidden) {
+			return this.label;
+		}
+		if (this.hasAttribute('aria-label')) {
```
dlockhart: Supporting `aria-label` was a thing that text-input had to do for backwards compatibility, but since time input is new, it does not. Can you remove this bit? It only needs to understand `labelHidden` and if that's true, `this.label` is the `aria-label` instead of a visible one.

components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+			return this.getAttribute('aria-label');
+		}
+		return undefined;
+	}
+
+	async _handleChange(e) {
```
dlockhart: Just want to confirm that this `change` event doesn't actually bubble up further right? Like we're handling it and then firing our own, but the original one is swallowed by the shadow DOM? (oooh, sounds scary)

components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+		const time = parseTime(value);
+		this._formattedValue = value;
+		await this.updateComplete;
+		if (time === null) {
+			this._formattedValue = formatTime(parseValue(this.value));
+		} else {
+			this.value = formatValue(time);
+			// Change events aren't composed, so we need to re-dispatch
+			this.dispatchEvent(new CustomEvent(
+				'd2l-time-input-changed',
```
dlockhart: Let's just call this event `change`, similar to text inputs.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+		});
+
+		it('passes all axe tests when disabled', () => {
+			elem.setAttribute('disabled', 'disabled');
+			expect(elem).to.be.accessible();
+		});
+
+		it('passes all axe tests when focused', async() => {
+			setTimeout(() => getInput(elem).focus());
+			await oneEvent(elem, 'focus');
```
dlockhart: Missing an `await expect(elem).to.be.accessible();`?

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+import { aTimeout, expect, fixture, oneEvent } from '@open-wc/testing';
+
+const basicFixture = '<d2l-input-time label="label text"></d2l-input-time>';
+
+describe('d2l-input-time', () => {
+
+	let elem;
+
+	beforeEach(async() => {
+		elem = await fixture(basicFixture);
```
dlockhart: Now that it's super easy to create fixtures in the open-wc stuff, I'm actually more a fan of not reusing the same fixture for everything and manipulating in JS. Like have one default fixture, but then when you need something slightly different, create it, like this:

```
it('passes all axe tests when label is hidden', async() => {
    const elem = await fixture(html`<d2l-input-time label="label" label-hidden></d2l-input-time>`);
    await expect(elem).to.be.accessible();
});
```

There are several cases down below where you're using JS to manipulate state before asserting things where you could just use a different fixture initially.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+	});
+
+	function getInput() {
+		return elem.shadowRoot.querySelector('.d2l-input');
+	}
+
+	describe('accessibility', () => {
+
+		it('passes all axe tests', () => {
+			expect(elem).to.be.accessible();
```
dlockhart: You actually need an `await` before the expect for all these -- otherwise the test returns and passes.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+		});
+
+		it('should put hidden label on "aria-label"', async() => {
+			elem.setAttribute('label-hidden', 'label-hidden');
+			await elem.updateComplete;
+			expect(getLabel()).to.be.null;
+			expect(getInput().getAttribute('aria-label')).to.equal('label text');
+		});
+
+		it('should fall back to using "aria-label" for backwards compatibility', async() => {
```
dlockhart: This test can go away based on other comment.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+		});
+
+		it('passes all axe tests when focused', async() => {
+			setTimeout(() => getInput(elem).focus());
+			await oneEvent(elem, 'focus');
+		});
+	});
+
+	describe('attribute reflection', () => {
+		it('should reflect disabled property to attribute', async() => {
```
dlockhart: This test can go away based on other comment.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+			await elem.updateComplete;
+			expect(getLabel()).to.be.null;
+			expect(getInput().getAttribute('aria-label')).to.equal('new label');
+		});
+
+	});
+
+	describe('value', () => {
+
+		function dispatchEvent(eventType, composed) {
```
dlockhart: Can you move this outside to the top of the file? Just trying to stay consistent with what I've been doing in other tests.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+describe('d2l-input-time', () => {
+
+	let elem;
+
+	beforeEach(async() => {
+		elem = await fixture(basicFixture);
+	});
+
+	function getInput() {
```
dlockhart: Can you move this to the top of the file to stay consistent with what I've been doing elsewhere? It'll need to take `elem` as a parameter.

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+		});
+
+		it('should not fire "d2l-time-input-changed" event when input value is invalid', async() => {
+			let fired = false;
+			elem.addEventListener('d2l-time-input-changed', () => {
+				fired = true;
+			});
+			getInput().value = 'invalid input text';
+			dispatchEvent('change', false);
+			await aTimeout(500);
```
dlockhart: 500 seems like overkill and will slow our tests down a lot. Can you play with different values to see how small it can be before it stops failing?

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+		});
+
+		it('should change "value" property when input value changes', async() => {
+			getInput().value = '11:22AM';
+			dispatchEvent('change', false);
+			await oneEvent(elem, 'd2l-time-input-changed');
+			expect(elem.value).to.equal('11:22:00');
+		});
+
+		it('should provide a time object with hour, minute and second', async() => {
```
dlockhart: If all we want to test here is that `getTime()` works properly, we could simplify to:

```
it('should provide a time object with hour, minute and second', async() => {
    const elem = await fixutre(html`<d2l-time-input value="11:22 AM"></d2l-time-input>`);
    expect(elem.getTime()).to.deep.equal({ hour: 11, minute: 22, second: 0 });
}
```

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+			getInput().value = '11:22 AM';
+			dispatchEvent('change', false);
+			await oneEvent(elem, 'd2l-time-input-changed');
+			expect(elem.getTime()).to.deep.equal({ hour: 11, minute: 22, second: 0 });
+		});
+
+		it('should not save input seconds', async() => {
+			//The seconds property is exclusely for the end-of-day value
+			getInput().value = '11:22:33 AM';
```
dlockhart: Hmm, I actually think this should work -- otherwise how is the date input going to tell it to go to "end of day"? I think I'd change this to something like:

```
it('should not save input seconds after time changes', async() => {
    const elem = await fixutre(html`<d2l-time-input value="11:22:33 AM"></d2l-time-input>`);
    expect(elem.getTime().second).to.equal(33);
    getInput().value = '11:45 AM';
    dispatchEvent('change', false);
    await oneEvent(elem, 'd2l-time-input-changed');
    expect(elem.value).to.equal('11:45:00');
}
```

Essentially, I think we want to support seconds when the `value` gets set, we just don't surface it in the input's UI. And then if the value changes, we obviously need to forget about them.

components/inputs/test/input-time.visual-diff.js
```diff
@@ -0,0 +1,37 @@
...
+		'basic',
+		'disabled',
+		'labelled',
+		'label-hidden',
+	].forEach((name) => {
+		it(name, async function() {
+			const rect = await visualDiff.getRect(page, `#${name}`);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+		it(`${name}-focus`, async function() {
```
dlockhart: I know text inputs likely did this, but because time input just uses text input, I feel like just having a single focus visual test is likely enough -- not one for each of the different states.

components/inputs/input-time.js
```diff
@@ -0,0 +1,156 @@
...
+				{bubbles: true, composed: false}
```
dbatiste: Is this meant to be composed?

components/inputs/test/input-time.test.js
```diff
@@ -0,0 +1,130 @@
...
+	});
+
+	describe('labelling', () => {
+
+		function getLabel() {
+			return elem.shadowRoot.querySelector('.d2l-input-label');
+		}
+
+		it('should display visible label', async() => {
+			expect(getLabel().innerText).to.equal('label text');
```
dbatiste: Is this captured in visual-diff already?

components/inputs/input-time.js
dlockhart: The text and search inputs both don't use composed events... guess it depends whether we want these crossing shadow DOM boundaries. Not sure if that decision was arbitrary or not, but I think we should be consistent at least, unless there's a good reason to deviate.

Scy-D2L: If I remember correctly, we eventually plan on having a full `date-time` component that contains the individual `date` and `time` input components, right? Is there any reason we would want events from the sub-components to be read at a higher level there, or are we just lumping both into a changed event for the whole thing? 🤔

dlockhart: I think we actually would want those events to specifically NOT bubble up -- the date-time component would want them trapped and it would send its own `change` event that contains the combined date and time.

dlockhart: i.e. I'm leaning towards leaving this as not composed.

dbatiste: Cool - I wasn't sure what to think of it, based on the comment which made me think we meant to make it composed.

dlockhart: That's a good point though -- Scy can you remove the comment as it was intended for text-inputs which do just need to re-send the original event. In our case, we trap the text-input event and send our own intentionally.

components/inputs/test/input-time.test.js
Scy-D2L: There are specific tests for `basic` and for `label-hidden`, but the visual-diff doesn't do anything to make sure they're different or that the label exists afaik. I mean going forward now that I've generated the goldens with them it should, but that doesn't seem like the intended purpose of visual-diff 🤔 I may be wrong.

components/inputs/test/input-time.test.js
dlockhart: Still missing `await`s before each of the `expect`s.

components/inputs/test/input-time.visual-diff.html
```diff
@@ -0,0 +1,32 @@
...
+	</head>
+	<body class="d2l-typography">
+		<div style="width: 300px;">
+			<div class="visual-diff">
+				<d2l-input-time id="basic" value="text"></d2l-input-time>
+			</div>
+			<div class="visual-diff">
+				<d2l-input-time id="labelled" label="Name" value="text"></d2l-input-time>
+			</div>
+			<div class="visual-diff">
```
dlockhart: Super nit: the label for these isn't really "typical", i.e. something more appropriate would be "Start Time"

components/inputs/test/input-time.test.js
Scy-D2L: Recap of elsewhere conversation: For some reason VScode warns that adding `await` will have no effect, but actual testing and the [Chai A11y test docs for open-wc](https://open-wc.org/testing/testing-chai-a11y-axe.html) prove it does. Making this change in next commit 👍

components/inputs/test/input-time.visual-diff.html
Scy-D2L: Just realized that these should be using times as values too - updating.

PR #361, opened 2020-02-05
using open-wc for menu tests

dlockhart:
This ended up being a larger refactoring of the tests than some of the others. I tried to only test the functionality introduced by the component/mixin in the associated tests vs. there was some cross-testing of behaviour in the old ones.


components/menu/menu-item-selectable-mixin.js
```diff
@@ -9,6 +9,11 @@ export const MenuItemSelectableMixin = superclass => class extends MenuItemMixin
 		};
 	}
 
+	constructor() {
+		super();
+		this.selected = false;
```
dlockhart: @margaree @dbatiste I made this change to the actual component. Otherwise `this.selected` is `undefined` unless it's set.

components/menu/test/menu-item-mixin.test.js
```diff
@@ -0,0 +1,62 @@
+import { defineCE, expect, fixture, oneEvent } from '@open-wc/testing';
+import { LitElement } from 'lit-element/lit-element.js';
+import { MenuItemMixin } from '../menu-item-mixin.js';
+
+const tag = defineCE(
+	class extends MenuItemMixin(LitElement) {}
```
dlockhart: Was able to kill all of the test elements and replace them with simple classes like this. 🚀

components/menu/test/menu-checkbox.visual-diff.js
```diff
@@ -47,7 +47,7 @@ describe('d2l-menu checkbox', function() {
 	});
 
 	describe('selection behavior', () => {
-		this.afterEach(async() => {
+		afterEach(async() => {
```
dlockhart: No idea why this was needed, but visual diff tests were failing with `this.afterEach is not a function`.

components/menu/menu-item-selectable-mixin.js
dbatiste: I've noticed a couple instances where the linter will complain about uninitialized booleans in certain statements.

components/menu/test/menu-checkbox.visual-diff.js
dbatiste: Interesting - I guess since `afterEach` is global for Mocha, it was not needed.  Were they always failing, or did they start failing when the outermost `describe` callback was changed to an arrow function?

dlockhart: Yeah you're right, that likely explains it.

PR #362, opened 2020-02-05
using open-wc for meter tests

dlockhart:
@m6jones FYI.


test/index.html
```diff
@@ -16,7 +16,6 @@
 				'../components/menu/test/menu-items/menu-item-radio-mixin.html',
 				'../components/menu/test/menu-items/menu-item-selectable-mixin.html',
 				'../components/menu/test/menu-items/menu-item-separator.html',
-				'../components/meter/test/meter-linear.html',
```
dlockhart: Radial and circle were missing, so were not getting run. :(

PR #372, opened 2020-02-07
moving version update into a stage

dlockhart:
The way stages work is that each one is run in sequence, and the next stage doesn't run unless the previous stage completes successfully. So I think it makes a lot of sense for the version update to run in a separate stage.


dbatiste:
Does this result in installing the dependencies a 3rd time?  Not against, but I was definitely 😞 to see dependencies being installed a second time when we switched to stages previously, since it takes a good amount of time to install them.

dlockhart:
Unfortunately yes it does, although it's actually what Dilgot recommends [in the bot docs](https://github.com/BrightspaceUI/visual-difference-bot) to avoid having to check specifically for the currently running stage like we were doing.

Separately, I'm investigating what would be involved in moving the visual diff tests into a job inside the "test" stage instead of being its own stage. That way it would run _in parallel_ with the unit tests instead of sequentially, which should save us a ton of time.

I also want to investigate a better version update/release mechanism than frau-ci, which seems... a little icky to me.

But I still think this is a good change in the meantime.

dlockhart:
Just pushed another change that moves the environment variables into the stages where they're actually needed.

.travis.yml
```diff
@@ -2,51 +2,51 @@ language: node_js
...
-  provider: npm
-  email: d2ltravisdeploy@d2l.com
-  skip_cleanup: true
-  api_key:
-    # 57c4......d343
-    secure: IitMXMxHrtsjCy3eucLs/mxja8e27DRpxT1YnA7HM6nCpKEr+SzPaHqV4tSNfX8fmKvqhyR3SRijPWnGahaZNVAPlN0l01MSEaG419o3QIKFzSgpNZLjNe5BsGTH3CFC5yPgtrU6ILW2Et1pb1+oixghXSifybxxan4aJKwBuri1DnrijmBnlEyog1Wglra1e7GzwZnEohDitHWKqO2V6uE+v1LmDtnHMzKoEjEeD/iHk1/oKdfsbZLEcZ+l6xU3gORBMskG+XT8EPgPOpfnuCiL0tqJ+ZFcP2hbNrzMMKs/k5TAmvL0DJyvmDfjlBXLik4aHKX6GmYGfhfktBCEN4hl8rmo9eBoQxJuzWWnCJq5BxVRGxaBZg7yZK5qZYR8cUwS5t241LqDgMK4rQP6J6cImCZ1Fe93t9yMd3GsurSHIByHwwQ3tuY4Xj3f6Nr/skGYrS9azqRtREWsRY4rsiudcKuTQGsDfLMpEN6ZGZQBXrYgKZpCrgIUiK5PyggbzTfOfpSrxe4oBD6IIb6bITYPPU5rXnE4EMOLY6AUn5OBxR/ACkooccFa7gHxlgVaNAy+RRIFk4sW6NAbADZXzy1U+jp8QsCzZBU0gmjNLShD0uuUfWSdv75qq3eKlwBzrlI6TpADoGmsFxoO3Yc6d0YGOmcxWVu6ZRuqeyVTZ9I=
-  on:
-    repo: BrightspaceUI/core
-    tags: true
+    - stage: tests
```
dbatiste: In the past I've thought about changing this so we have... "unit-tests" and "visual-difference-tests".  🤷‍♂

dbatiste: But if you end up moving it into a job in the same stage, then "tests" makes a lot of sense.  Didn't notice the rename before.

dlockhart: Haha well technically this stage is "lint" and "unit tests" but yeah, I'll look at the naming if I can get them into the same stage. It's going to require a change to the bot.

PR #378, opened 2020-02-10
Port d2l-dropdown-button and d2l-dropdown-button-subtle to Lit

AllanKerr:


components/dropdown/dropdown-button-subtle.js
```diff
@@ -0,0 +1,37 @@
+import '../button/button-subtle.js';
+import '../icons/icon.js';
```
dlockhart: Seems like this dependency only needs to be in `button-subtle`.

components/dropdown/dropdown-button.js
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	static get styles() {
+		return [dropdownOpenerStyles, css`
+			d2l-icon {
+				height: 0.8rem;
+				width: 0.8rem;
+			}
+			.d2l-dropdown-button-text {
+				margin-right: 0.6rem;
```
dlockhart: Hmm, I wonder why we didn't simplify this and just put the margin on the icon, removing the need for this span...

index.html
```diff
@@ -40,6 +40,8 @@ <h2 class="d2l-heading-3">Components</h2>
 			Dropdowns
 			<ul>
 				<li><a href="components/dropdown/demo/dropdown.html">d2l-dropdown</a></li>
+				<li><a href="components/dropdown/demo/dropdown-button.html">d2l-dropdown-button</a></li>
+				<li><a href="components/dropdown/demo/dropdown-button-subtle.html">d2l-dropdown-button-subtle</a></li>
```
dlockhart: Nit: I'd be fine with combining these 2 demo pages into a single "openers" or "dropdown-button" one, similar to the tests.

components/dropdown/dropdown-button-subtle.js
```diff
@@ -0,0 +1,37 @@
...
+		};
+	}
+
+	static get styles() {
+		return dropdownOpenerStyles;
+	}
+
+	render() {
+		return html`
+			<d2l-button-subtle text=${this.text} icon="tier1:chevron-down" icon-right=""></d2l-button-subtle>
```
dbatiste: `icon-right` is boolean for the component, so we should not need to assign empty string here.

components/dropdown/demo/dropdown-button.html
```diff
@@ -10,6 +10,7 @@
 	<script type="module">
```
dbatiste: Now that this contains all the openers, I'd probably rename it to `dropdown-openers.html`. Presumably the other openers (context-menu (chevron icon) and more (... icon) will be added here once converted?

AllanKerr:
> The `d2l-dropdown-context-menu` and `d2l-dropdown-more` openers would fit nicely with this PR as well, but if you want to do them as a separate PR, that's fine too.

I didn't include the `d2l-dropdown-more` since its demo uses the `d2l-dropdown-menu` and I figured I'd keep the demo the same when converting. I can add the `d2l-dropdown-context-menu` to this one and then group the `d2l-dropdown-more` and `d2l-dropdown-menu` into another PR.

components/dropdown/dropdown-context-menu.js
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Gets the "d2l-button-icon" opener element (required by dropdown-opener-mixin).
+	 * @return {HTMLElement}
+	 */
+	getOpenerElement() {
+		return this.shadowRoot.querySelector('d2l-button-icon');
+	}
+
+	_onClose() {
+		this.removeAttribute('opened');
```
dbatiste: Do you know what the purpose of these handlers is?  Looks like the "more" opener has them too.

AllanKerr: I couldn't find anywhere they're being used but I can't 100% rule out cases that might be using `getAttribute("opened")` or referencing `opened` in an attribute selector that doesn't reference the `d2l-dropdown-context-menu` element explicitly since they're somewhat difficult to search for.

dbatiste: Looks like it was used a long time ago.  The styles were removed here, and it looks like perhaps we should have removed these handlers at the same time... 🤔 

https://github.com/BrightspaceUI/dropdown/commit/2a08a6aa20ffabd4e4e5ded1812d4c894a29ebda#diff-abcef24286f2e79076d2e11d242b7528L40

I'd be keen to remove them if possible.

AllanKerr: Sounds good to me, I removed it! With having found where it was originally used and not finding anywhere else its referenced it should be relatively low risk to remove it.

PR #382, opened 2020-02-12
Convert d2l-dropdown-menu from Polymer to Lit

AllanKerr:
# Changes
1. Fix a defect in the `d2l-dropdown-button` where clicking the icon would not trigger the button's focus.
2. Modify the `d2l-dropdown-opener` mixin to look for the `d2l-dropdown-content` attribute instead of the `d2l-dropdown-content` element so that it works with both `d2l-dropdown-content` and `d2l-dropdown-menu` components.

## Note:
This PR does not fix the issue where an animation will play and the menu will resize when the menu is closed on another page and then opens on the first page. This requires changes to the `d2l-hierarchical-view` and will be added in its own PR.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -380,14 +386,15 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
-			doOpen();
+			await doOpen();
 
 		} else {
 
 			if (this.__dismissibleId) {
 				clearDismissible(this.__dismissibleId);
 				this.__dismissibleId = null;
 			}
+			await this.updateComplete;
```
AllanKerr: Adding an `await` here guarantees that the dropdown will actually be closed when the `d2l-dropdown-close` event is received.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -228,13 +234,13 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
 			/* add 2 to content width since scrollWidth does not include border */
-			width: this._width ? `${this._width + 20}px` : undefined
+			width: this._width ? `${this._width + 20}px` : ''
 		};
 
 		const containerStyle = {
 			minWidth: this.minWidth ? `${this.minWidth}px` : undefined,
 			/* set width of content in addition to width container so IE will render scroll inside border */
-			width: this._width ? `${this._width + 18}px` : undefined,
+			width: this._width ? `${this._width + 18}px` : '',
```
AllanKerr: I'm not entirely sure why this change was necessary but with `undefined`, the width of the dropdown content would infinitely expand. This does not occur with `''`.

components/dropdown/test/dropdown-menu.visual-diff.js
```diff
@@ -0,0 +1,27 @@
...
+		browser = await puppeteer.launch();
+		page = await browser.newPage();
+		await page.setViewport({ width: 800, height: 800, deviceScaleFactor: 2 });
+		await page.goto(`${visualDiff.getBaseUrl()}/components/dropdown/test/dropdown-menu.visual-diff.html`, { waitUntil: ['networkidle0', 'load'] });
+		await page.bringToFront();
+	});
+
+	after(() => browser.close());
+
+	it('first-page', async function() {
```
AllanKerr: I'm planning on adding an additional case for the second page once the `hierarchical-view` issue is fixed to ensure that the `dropdown-menu` content resizes properly when the inner menu resizes. Right now it can't be added without accessing `d2l-menu` / `d2l-hierarchical-view` implementation details or a hard-coded timeout to determine when the animation ends.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -93,6 +93,11 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			},
 			_contentOverflow: {
 				type: Boolean
+			},
+			_dropdownContent: {
```
dbatiste: Nit: normally I would try to keep the properties order alphabetically.
```
static get properties() {
	return {
		// public ordered alpha
		...
		// private order alpha
		...
	};
}
```

It might be just me, but I tend to put the property def'n on one line to save some scrolling. That being said, I might have been the source of the original mixup.  🤷‍♂

components/dropdown/dropdown-button.js
```diff
@@ -25,6 +25,7 @@ class DropdownButton extends DropdownOpenerMixin(RtlMixin(LitElement)) {
 				height: 0.8rem;
 				width: 0.8rem;
 				margin-left: 0.6rem;
+				pointer-events: none;
```
dbatiste: Curious - what prompted this?

components/dropdown/dropdown-menu.js
```diff
@@ -0,0 +1,91 @@
...
+	_onSelect(e) {
+		if (e.target.tagName !== 'D2L-MENU-ITEM') {
+			return;
+		}
+		this.close();
+	}
+
+	__getMenuElement() {
+		return this.shadowRoot.querySelector('slot')
+			.assignedNodes()
```
dbatiste: Curious if `assignedElements()` works in all browsers now?  I think Safari might have been the hold-out... but based on a quick check it seems like it works for me in one of my StackBlitz projects.  It may avoid the check for `hasAttribute`. If it works, there are other places that are using `assignedNodes()` and explicitly filtering to get elements - they could eventually be cleaned up if this API is fully supported now.  Not a big deal though.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -93,6 +93,11 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			},
 			_contentOverflow: {
 				type: Boolean
+			},
+			_dropdownContent: {
+				type: Boolean,
+				attribute: 'd2l-dropdown-content',
```
dbatiste: I think this could just be `dropdown-content`.

components/dropdown/dropdown-content-mixin.js
dbatiste: Presumably with this change, `width=""` gets rendered in the inline styles, and it's essentially the same as `someElem.style.width = '';`?  What is the difference in computed value for the `width` in the browser?

components/dropdown/dropdown-button.js
dlockhart: That was me. I noticed that in the LMS clicking the dropdown text or icon wouldn't cause focus styles to be applied, but clicking elsewhere would. Things are slightly different in the Lit version, where only clicking the icon would cause it.

components/dropdown/dropdown-menu.js
dlockhart: Technically we still support Safari 11, so we'll need to be careful.

components/dropdown/test/dropdown-menu.visual-diff.html
```diff
@@ -0,0 +1,48 @@
...
+<html lang="en">
+
+<head>
+	<link rel="stylesheet" href="../../../test/styles.css" type="text/css">
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '../../typography/typography.js';
+		import '../../menu/menu.js';
+		import '../../menu/menu-item.js';
+		import '../../menu/menu-item-radio';
```
dlockhart: Missing `.js` extension.

components/dropdown/dropdown-content-mixin.js
AllanKerr: With `''` it looks like the width of the `d2l-dropdown-content-width` is always `370px` from the `width: 100vw` and `max-width: 370px` in the `dropdown-content-styles`. With `undefined` the `d2l-dropdown-content-width` drops down to the width of the `d2l-dropdown-content-container + 2` and triggers the `d2l-menu-resize` infinitely until it reaches the max of `370px`.

Interestingly, this only seems to show if I log with it `getComputedStyle().width`. If I try and observe it in the browser the computed style shows as `370px` from the `width: 100vw` for both `''` and `undefined`.

PR #383, opened 2020-02-13
DE37329/ Remove extra spacing from menu

Scy-D2L:
This change prevents the [extra bottom padding that appears the second time you open a dropdown containing a menu.](https://rally1.rallydev.com/#/15545167705d/detail/defect/360088526908)


components/menu/menu.js
```diff
@@ -27,6 +27,7 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 			:host {
 				box-sizing: border-box;
 				display: block;
+				vertical-align: top;
```
Scy-D2L: This is the same v-align fix we used to stop it from happening in the initial load. The fix works - but `lint` hates it, because this property doesn't do anything with `display: block`. The issue is that after opening and closing the dropdown once, when you reopen it, [this is applied](https://github.com/BrightspaceUI/core/blob/master/components/hierarchical-view/hierarchical-view-mixin.js#L38) from the hierarchical view- so it's displayed as an inline-block from that point. (and the mystery padding sets in).

svanherk: Do we know why we set it to `inline-block` and not `block` in hierarchical-view, or why we use `block` here and not `inline-block`?  I remember investigating this with Margaree and we both thought it was odd to have it be `block` the first time it's opened, and `inline-block` the second time... @dbatiste do you remember if this was intentional?

dlockhart: Could you move it into that CSS rule then so the linter doesn't complain? Like:

```
:host([shown]) {
	display: inline-block;
	vertical-align: top;
}
```

Scy-D2L: If we don't anticipate any issues from that rule being applied elsewhere, for sure.

dlockhart: Let's put it there -- that way it's next to the line that's causing the problem, and the linting will pass.

dlockhart: You could even put a little comment:

```
vertical-align: top; /* DE37329: required to prevent extra spacing caused by inline-block */
```

If this is easily reproducible, it would also be great to add a visual diff test for it so that we don't accidentally regress in the future.

components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -37,6 +37,7 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 			}
 			:host([shown]) {
 				display: inline-block;
+				vertical-align:top; /* DE37329: required to prevent extra spacing caused by inline-block */
```
dbatiste: Micro-nit: like to have space after the `:`.

components/dropdown/test/dropdown-menu.visual-diff.js
```diff
@@ -1,27 +1,40 @@
...
-
-	after(() => browser.close());
-
-	it('first-page', async function() {
-		await helper.open(page, '#dropdown-menu');
-		const rect = await helper.getRect(page, '#dropdown-menu');
-		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-	});
-
-});
```
dlockhart: Is your IDE switching CRLF/LF's around?

Scy-D2L: I didn't even see that - weird. But it's only this one file?

dlockhart: Yeah I dunno. VS Code actually has a little thing in the status bar that shows you which it's using for a particular file, and clicking it lets you switch it back and forth.

Scy-D2L: TIL: The VScode line-endings button exists.
This was switched to LF (which is what we want) so there's no issue. Going to add linting for line-endings in another PR.

PR #387, opened 2020-02-14
catching errors thrown from intlmessage-format

dlockhart:
This change will prevent exceptions thrown from `intlmessage-format` from bubbling up and breaking our app, like what happened today with my-courses. If folks like this approach, I'll do the same to `d2l-localize-behavior` -- which is where the problem today originated from.

This still outputs the error to the console, so hopefully developers will still see it. But in not totally breaking the app, there is a chance that this causes devs to not notice problems. I think it's worth the trade-off though.

When this happens, the localized string without formatting applied will be used instead.


mixins/localize-mixin.js
```diff
@@ -112,7 +112,13 @@ export const LocalizeMixin = superclass => class extends superclass {
 		}
 
 		const translatedMessage = new IntlMessageFormat(translatedValue, this.__language);
-		return translatedMessage.format(params);
+		let formattedMessage = translatedValue;
+		try {
+			formattedMessage = translatedMessage.format(params);
+		} catch (e) {
+			console.error(e);
```
dbatiste: Does the resulting console statement contain enough info for us to track down what's going on?

dlockhart: Screen Shot 2020-02-14 at 5 42 39 PM

dbatiste: ... based on the test, it  like it  👀 like does.

PR #389, opened 2020-02-18
Fix defect where opener is not found when it has multiple classes

AllanKerr:


components/dropdown/dropdown.js
```diff
@@ -19,7 +19,7 @@ class Dropdown extends DropdownOpenerMixin(LitElement) {
 	getOpenerElement() {
 		return this.shadowRoot.querySelector('slot')
 			.assignedNodes()
-			.find(node => node.className === 'd2l-dropdown-opener');
+			.find(node => node.classList && node.classList.contains('d2l-dropdown-opener'));
```
dlockhart: Nit: probably safe at this point given our supported browsers to assume the `classList` API exists

AllanKerr: It seems to be necessary because `assignedNodes` can return non-element nodes. For text nodes, `classList` is undefined.

dlockhart: Ah gotcha 👍

dbatiste: Does this flatten the tree when providing the nodes?  I think the original implementation did a query that may have worked regardless of where the opener was in the userland sub-tree, but I am not certain.

AllanKerr: `assignedNodes` does not flatten by default. I'll double check to see what the behavior is on the orginal.

dbatiste: Looks like there is a `{ flatten: true }` option for the API if we need it.

AllanKerr: Double checked this; the original implementation only checks immediate children so we should be good without `flatten`.

PR #393, opened 2020-02-21
Dropdown Minor Version Bump

AllanKerr:
Bump the minor version so that all core Dropdown components are fixes are in.


AllanKerr:
Well I messed that one up... In my haste I pasted `[increment minor]` in the wrong box so it didn't actually bump the version 😑. Going to put up a README PR and do the version bump along with that.

PR #396, opened 2020-02-24
Dropdown header and footer

Scy-D2L:
- Added header and footer slots to the dropdown
- If slots aren't filled, the top and bottom caps default to the previously expected size/behaviour
- When slots are filled caps scale to accommodate
- By default padding is the same for top/bottom caps as it is for regular content (in dropdown-menu the padding is the same as it would be for a regular dropdown) but using `no-padding-header` and `no-padding-footer` sets it to 0 instead.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -424,7 +456,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			});
 
 			const spaceRequired = {
-				height: contentRect.height + 20,
+				height: contentRect.height + headerRect.height + footerRect.height + 10,
```
dlockhart: Curious what the extra `10px` is?

components/dropdown/dropdown-content-mixin.js
```diff
@@ -438,7 +470,8 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 				this._position = position;
 			}
 
-			const maxHeight = Math.floor(this.openedAbove ? spaceAround.above : spaceAround.below);
+			const headerFooterHeight = headerRect.height + footerRect.height;
```
dlockhart: Looks like the header + footer height is used above to compute `spaceRequired` -- could the declaration of this variable be moved up so both places can use it?

components/dropdown/dropdown-content-mixin.js
Scy-D2L: I'm not entirely sure, I removed 10px to account for the header/footer heights (they were 5px before and still default to it, but since we're measuring we didn't need them) and this worked with the visual diff. @AllanKerr might have a better idea where they come from?

dlockhart: My guess: `5px` and `5px` was from the old header and footer. The little triangle thing is `18px` but it's positioned `-10px` so would take up an extra `8px` of height... but inspecting it appears to be closer to `25px` high positioned up `-10px`... so I'm not really sure because that would be `15px` plus the other 10 making it `25px` instead of `20px`.

AllanKerr: That aligns with my best guess as well. I just pulled through the same value to keep it consistent with the Polymer version. It shouldn't make a huge difference since all its used to determine is whether to place the dropdown above or below the opener.

components/dropdown/demo/dropdown-menu.html
```diff
@@ -97,6 +97,37 @@ <h2>Normal</h2>
...
 
+		<h2>With Header/Footer</h2>
+		<d2l-demo-snippet>
+			<template>
+				<d2l-dropdown>
+					<button class="d2l-dropdown-opener">Open it!</button>
+					<d2l-dropdown-menu id="dropdown" no-padding-footer>
+						<div slot="header" style="font-weight: bold; width: 100%; text-align: center;">Physics Topics</div>
+						<d2l-menu>
+							<d2l-menu-item text="Introduction" id="first-item"></d2l-menu-item>
```
dbatiste: Are the `first-item` and `last-item` ids used at all?

Scy-D2L: It doesn't look like it, actually. It's in the demo page and the code examples in the `dropdown-menu` `readme.md` but the only place I see any use is in a [test for `dropdown-menu`](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/test/dropdown-menu.test.js#L51) but that's only `first-item` and not `last-item`.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -100,8 +100,8 @@ describe('d2l-dropdown', () => {
 		it('sets scrollTop to specified value', async() => {
 
 			content.boundary = {
-				below: 1,
```
Scy-D2L: This change relates to how scaling is handled. Because the header and footer are accounted for in the calculations now, at very small sizes scrolling won't initiate for the content space (think 1-10 pixels total height for the content area and there isn't room for that) but this is a non-issue because the full dropdown will still appear, it may just be off-screen and require scrolling the window to see it.

components/dropdown/demo/dropdown-menu.html
Scy-D2L: Checked with Allan, they were back in the Polymer tests but aren't used apart from the test-case, so all except those test-cases have been removed.

components/dropdown/README.md
```diff
@@ -207,7 +207,7 @@ dropdown.addEventListener('click', function() {
       </d2l-menu-item>
       <d2l-menu-item text="Stars &amp; Galaxies"></d2l-menu-item>
       <d2l-menu-item text="The Night Sky"></d2l-menu-item>
-      <d2l-menu-item text="The Universe" id="last-item"></d2l-menu-item>
+      <d2l-menu-item text="The Universe"></d2l-menu-item>
     </d2l-menu>
   </d2l-dropdown-menu>
 </d2l-dropdown-button>
```
dlockhart: Do we want to mention these new slots in the README?

Scy-D2L: I know initially we didn't want to broadcast the capability but that seems to have changed with the use-cases that came up in discussion, so I can throw something in. 😃

components/dropdown/README.md
```diff
@@ -176,6 +178,18 @@ dropdown.addEventListener('click', function() {
 * `d2l-dropdown-open`: dispatched when the dropdown is opened
 * `d2l-dropdown-close`: dispatched when the dropdown is closed
 
+**Slots:**
```
dbatiste: This might be slightly confusing or misleading since there are three slots - it might be worth mentioning the default slot. In other places, I didn't both listing the slots since the snippet already shows them.  🤷‍♂

Scy-D2L: Is it more or less confusing? I mean anything that isn't in a named slot ends up in the default content slot, which has no name and you don't make a div for 🤔 Might just be me

dbatiste: Possibly.  I guess the other thing is being consistent about whether we document the slots this way.  So far I've been opting for less text especially when the snippet already conveys the info.  I don't feel super strongly about it, but I'd rather be consistent either way.

Scy-D2L: I'm wondering if having some explicit listing like this might be useful in places where the slots are functionally optional, or if that's easy to assume from the outside? I haven't seen enough cases to know.

If we don't need them explicitly listed I can make the adjustment in another PR and check the docs in general for that consistency with that. 🙂 Just wanted to see what thoughts were (but merging this in for now so I can press on with other things in the meantime)

PR #397, opened 2020-02-25
Update colors for part 2 of the WCAG 2.1 changes

AllanKerr:


components/colors/colors.scss
```diff
@@ -10,83 +10,77 @@ $d2l-color-tungsten: #6e7376 !default;
...
+$d2l-color-citrine-plus-1: #fff9d6 !default;
 $d2l-color-citrine: #ffba59 !default;
-$d2l-color-citrine-minus-1: #a87b3b !default;
-$d2l-color-citrine-minus-2: #694b25 !default;
-
-// topaz
-$d2l-color-topaz-plus-1: #fffbb2 !default;
-$d2l-color-topaz: #f5ec5a !default;
-$d2l-color-topaz-minus-1: #a69f21 !default;
-$d2l-color-topaz-minus-2: #635f15 !default;
```
dlockhart: So the old colors repo is [referencing topaz a couple times](https://github.com/BrightspaceUI/colors/blob/2ab6adb98e3de8fd8ba9e27d6f46526f1e392aea/d2l-colors.scss#L12).

We can either: 
1. Investigate whether those references are actually used and if not, remove them as well
2. Hardcode the HEX for topaz in there

components/colors/colors.js
```diff
@@ -17,86 +17,80 @@ if (!document.head.querySelector('#d2l-colors')) {
...
+			--d2l-color-malachite-minus-2: #00635e;
 
 			/* primary accent */
 			--d2l-color-primary-accent-action: var(--d2l-color-celestine);
 			--d2l-color-primary-accent-indicator: var(--d2l-color-carnelian);
 
 			/* feedback */
 			--d2l-color-feedback-error: var(--d2l-color-cinnabar);
-			--d2l-color-feedback-warning: var(--d2l-color-citrine);
+			--d2l-color-feedback-warning: var(--d2l-color-carnelian);
```
dlockhart: In the feedback screenshot, it looks like the warning bit should probably use white text now that the background is darker.

components/colors/colors.scss
AllanKerr: Good catch, I'll investigate.

AllanKerr: There don't seem to be any references to lurite or chartronic so I'll remove those.

dlockhart: Sounds good!

PR #403, opened 2020-02-28
Dropdown scales based on header&footer content

Scy-D2L:
Dropdown width is determined by header, content or footer - whichever is widest.
Previously it only considered content.

### DIFF
--- a/components/dropdown/dropdown-content-styles.js
+++ b/components/dropdown/dropdown-content-styles.js
@@ -79,6 +79,9 @@ export const dropdownContentStyles = css`
 		max-width: 370px;
 		position: absolute;
 		width: 100vw;
+		display: flex;
+		flex-direction: column;
+		align-items: flex-start;
 	}
 
 	:host([opened-above]) .d2l-dropdown-content-width {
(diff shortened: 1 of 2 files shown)


components/dropdown/dropdown-content-styles.js
```diff
@@ -79,6 +79,9 @@ export const dropdownContentStyles = css`
 		max-width: 370px;
 		position: absolute;
 		width: 100vw;
+		display: flex;
```
dbatiste: Was the flex change need to address the width, or what this change for a different layout req?  I notice none of the children define flex properties... what's the expected behavior?

Scy-D2L: The previous behaviour had the content as an `inline-block` while the header and footer sections were `block` - this meant they automatically took up the full width and I couldn't 'measure' them to figure out what the width should be. Switching them to inline sometimes caused things to pile up, etc.

Switching to `flex` in the parent container makes header-content-footer align to one side in a column, allows for variable widths and avoids a lot of weird line-height line-spacing everything-on-the-same-line problems that came out of most of the experiments with `inline` and `inline-block`.

So the child elements are expected to show up in a column, one to a line, with no spacing between them. This seemed like a cleaner solution than adding div layers or a number of properties to negate the text-i-ness of inline things.

PR #404, opened 2020-02-29
Add optional predicate to getFirstFocusableDescendant focus helper

emil-furniss:
When an error occurs in the FACE editors, we want to focus on the first input field that is in error.
Doing this via mixins/events/etc might be tricky, fragile, and high-maintenance, due to the many different sub-editors on the page and the state management to support the new save-cancel workflow.

The alternative is to do a dom search, which is what this PR strives to help support.
Since our use case doesn't care about the _next_ focusable element but instead the next focusable element that is also in error, this PR adds an optional predicate which allows us to filter the nodes returned by `getFirstFocusableDescendant`. Now it will return the first node that is focusable _and_ fulfills the predicate.

If this is generally approved of, I can add some tests.
(And I had originally also added the predicate to `getNextFocusable`, but ended up not needing that function so I removed it.)

### DIFF
--- a/helpers/focus.js
+++ b/helpers/focus.js
@@ -99,13 +99,11 @@ export function getPreviousFocusable(node, includeHidden) {
 	return focusable;
 }
 
-export function getNextFocusable(node, includeHidden, predicate) {
+export function getNextFocusable(node, includeHidden) {
 
 	if (!node) return null;
 
 	if (includeHidden === undefined) includeHidden = false;
-
-	if (predicate === undefined) predicate = () => true;
 
 	const _getNextAncestorSibling = (node) => {
 		let parentNode = getComposedParent(node);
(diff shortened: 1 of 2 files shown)


helpers/focus.js
```diff
@@ -26,14 +26,16 @@ export function getComposedActiveElement() {
...
 	const composedChildren = getComposedChildren(node);
 
 	for (let i = 0; i < composedChildren.length; i++) {
-		if (isFocusable(composedChildren[i], includeHidden)) return composedChildren[i];
+		if (isFocusable(composedChildren[i], includeHidden) && predicate(composedChildren[i])) return composedChildren[i];
 
-		const focusable = getFirstFocusableDescendant(composedChildren[i], includeHidden);
-		if (focusable) return focusable;
+		const focusable = getFirstFocusableDescendant(composedChildren[i], includeHidden, predicate);
+		if (focusable && predicate(focusable)) return focusable;
```
dbatiste: Is the `predicate(focusable)` needed here? -it looks like this will already have been invoked inside the `getFirstFocusableDescendant` call on the previous line.  I _think_ we just need to check if an element was returned so that we don't return `null` below.

emil-furniss: You're right, thanks! The predicate will have returned true for the found descendant in line 35 and then bubbled up.

PR #406, opened 2020-03-02
Create d2l-focus-trap container element.

dbatiste:
This is a factor of `DialogMixin`'s focus trapping logic.  I will update `DialogMixin`, add/move tests as necessary if want to pursue this further.  It provides `d2l-focus-trap` - simple container element that traps the focus.


components/focus-trap/focus-trap-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+		const focusable = getNextFocusable(this.shadowRoot.querySelector('.d2l-focus-trap-start'));
+		if (focusable) focusable.focus();
+	}
+
+	_getFocusTrapContainer() {
+		return this.shadowRoot.querySelector('.d2l-focus-trap-start').parentNode;
+	}
+
+	_handleFocusTrapBodyFocus(e) {
+		const trapContainer = this._getFocusTrapContainer();
```
dbatiste: Perhaps I would check `this.trap` here as a precaution... even though this handler should not be wired up while not trapping.

components/focus-trap/focus-trap-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+				lastFocusable.focus();
+				return;
+			}
+		}
+		this._focusTrapFocusFirst();
+	}
+
+	_renderFocusTrap(inner) {
+		if (this.trap) {
+			return html`
```
dbatiste: The only difference here is the absence of `tabindex="0"`.  I'd like to find a better way to avoid this duplication.  I thought of setting `tabindex="-1"`, but that would still make it focusable via script and may be targeted by various getFocusable helpers. Omitting it is really what I want when not actively trapping.

dlockhart: Can you use the `ifDefined` thing? Like:

```
const tabindex = this.trap ? "0" : undefined;
const blah = html`<span tabindex="${ifDefined(tabindex)}">`;

File: components/focus-trap/focus-trap.js [hunk 3]
```diff
@@ -0,0 +1,20 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { FocusTrapMixin } from './focus-trap-mixin.js';
+
```
dlockhart: Wondering a bit how useful the mixin actually is... doesn't seem to be providing much value over the element.

components/focus-trap/focus-trap-mixin.js
```diff
@@ -0,0 +1,99 @@
+import { getNextFocusable, getPreviousFocusable } from '../../helpers/focus.js';
+import { html } from 'lit-element/lit-element.js';
+import { isComposedAncestor } from '../../helpers/dom.js';
+
+export const FocusTrapMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			trap: { type: Boolean, reflect: true }
```
dlockhart: What the use case for the attribute? Like folks who want a focus trap, but don't want it to always be turned on?

components/focus-trap/focus-trap-mixin.js
dbatiste: Yep I think you are right.  Not sure what I was thinking yesterday.

components/focus-trap/focus-trap.js
dbatiste: Yeah, I'd started with the mixin and then introduced the component after.  I could definitely just be a component.

components/focus-trap/focus-trap-mixin.js
dbatiste: Dialogs is a great example - if we assume it's always trapping and we have a `d2l-dialog` in the DOM, it would be wired up to capture focus events elsewhere in the body (see the body event handler, which redirects focus to the trap) even when the dialog is not open.

dlockhart: I see. So dialog would bind the `trap` attribute to its own `opened`.

dbatiste: An alternative is to have dialog conditionally render `d2l-focus-trap` based on it's open state.  Not sure how I feel about that though ... not sure I like the dialog internal DOM changing to that degree when the dialog is opened/closed.

dlockhart: Yeah agreed that sounds a little wonky. The other option would be for the focus trap to know whether or not it's actually visible in the DOM, but that also sounds complicated.

components/focus-trap/focus-trap.js
dbatiste: Dialog actually has some custom logic for determining the first focusable element when it's coming from the body - it places focus on the first focusable element after the abort button (but is also inside the trap because it is part of the cycle).  In that case, the mixin is useful because it allows the consumer to provide a custom `focusFirst` method.

Maybe I could still move the mixin code into the component though.  The `d2l-focus-trap` could dispatch an event, `d2l-focus-trap-enter`, and dialog could listen for it an place focus accordingly.  🤔

dlockhart: Yeah or you could put like a `d2l-focus-trap-skip` CSS class on elements you want it to ignore when doing its calculations.

components/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,97 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { getNextFocusable, getPreviousFocusable } from '../../helpers/focus.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { isComposedAncestor } from '../../helpers/dom.js';
+
+class FocusTrap extends LitElement {
+
+	static get properties() {
+		return {
+			trap: { type: Boolean, reflect: true }
```
dlockhart: Reflection doesn't actually seem required?

components/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,97 @@
...
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: inline-block;
+			}
```
dlockhart: Nit: include the `:host([hidden]) { display: hidden; }` in case folks want to hide it that way.

components/focus-trap/focus-trap.js
```diff
@@ -0,0 +1,97 @@
...
+		return html`
+			<span class="d2l-focus-trap-start" @focusin="${this._handleStartFocusIn}" tabindex="${ifDefined(tabindex)}"></span>
+			<slot></slot>
+			<span class="d2l-focus-trap-end" @focusin="${this._handleEndFocusIn}" tabindex="${ifDefined(tabindex)}"></span>
+		`;
+	}
+
+	_focusFirst() {
+		const focusable = getNextFocusable(this.shadowRoot.querySelector('.d2l-focus-trap-start'));
+		if (focusable) focusable.focus();
```
dlockhart: Is it possible in dialog's case that this could focus first on one element and then dialog would move it to another? More worried about the screen reader experience.

dbatiste: Yep that would happen.  I did a test with VO and it seems to be ok... in the same way that landing on the trap-start or trap-end elements redirects wraps focus to the end or beginning and only the final active element is announced.  It's a valid concern though.  I _could_ add a property that enables consumers to opt-in/out of this behavior, but so far it seems unnecessary.

dlockhart: Cool, seems fine then!

PR #407, opened 2020-03-03
Fix bug with clicking a menu-item-link with voiceover on

margaree:
This affects any voice over users using iOS 13.

The anchor just isn't receiving any click event when voice over is on. I haven't been able to track down exactly what changed to cause this. I think this fix is fairly safe but let me know if there's something I'm missing or if there is potentially a better option. The `bubbles: false` is to prevent the `d2l-menu-item-select` event from being dispatched a second time.

### DIFF
--- a/components/menu/menu-item-link.js
+++ b/components/menu/menu-item-link.js
@@ -8,7 +8,6 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 	static get properties() {
 		return {
 			href: { type: String },
-			preventDefault: { type: Boolean, reflect: true, attribute: 'prevent-default' },
 			target: { type: String }
 		};
 	}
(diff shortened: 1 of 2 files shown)


components/menu/menu-item-link.js
```diff
@@ -61,6 +61,7 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 	}
 
 	_onClick(e) {
+		this.shadowRoot.querySelector('a').dispatchEvent(new CustomEvent('click', { bubbles: false }));
```
svanherk: Should this happen even if they want to prevent default?  Since that will cancel the click event on the menu item, but not this new event we are dispatching?

dbatiste: Yeah, similar to the keydown scenario below, which comes into play when the menu-item gets the focus and the user hits enter/space. It's code makes me suspicious that there could be an issue with ghost-clicking.

components/menu/menu-item-link.js
```diff
@@ -39,7 +39,7 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 	firstUpdated() {
 		super.firstUpdated();
 		this.addEventListener('keydown', this._onKeyDown);
-		this.shadowRoot.querySelector('a').addEventListener('click', this._onClick);
+		this.addEventListener('click', this._onClick);
```
dbatiste: Won't this change cause two click events on the anchor when on desktop?

margaree: No, so it seems by using a `CustomEvent` for the click, `bubbles` is false and so only one `d2l-menu-item-select` event is dispatched (whereas if I were to do something like `this.shadowRoot.querySelector('a').click()` there are two `d2l-menu-item-select` events dispatched.

PR #411, opened 2020-03-06
Input-text: Add left and right slots for icons/button-icons

margaree:
This is a component which contains an input that contains an icon or button-icon on either the left or right. There is a slot for the icon/button-icon. 

There is a property `right` that if true, the icon is on the right, else it defaults to left. 

There is a css var `--d2l-input-container-slot-content-padding` for how far left/right the icon/button-icon is. It defaults to 0.8rem, which works for the icon, but the button-icon usage would want to set this to 0.3rem.

This would be used in `d2l-input-search` and `d2l-input-date`.

I named this `d2l-input-container` but perhaps there is a more accurate/better name (`d2l-input-icon-container`?).

Once this seems to be thought of as a good idea/on the right track, I'll add info to the readme and tests.


components/inputs/input-container.js
```diff
@@ -0,0 +1,142 @@
...
+					class="${inputClass}"
+					@change="${this._handleChange}"
+					?disabled="${this.disabled}"
+					@input="${this._handleInput}"
+					@keypress="${this._handleKeyPress}"
+					maxlength="${ifDefined(this.maxlength)}"
+					placeholder="${ifDefined(this.placeholder)}"
+					type="${ifDefined(this.type)}"
+					.value="${this.value}">
+				<slot></slot>
```
dbatiste: Is the focus order meant to be the same regardless of the visual order?  i.e. if the icon button is on the left, should the input get the focus first, or the input?

margaree:
I have updated this to put the slots in the existing `input-text`. The focus behavior is still WIP (as it does make sense to me that if a button-icon were on the left, it would be focused on first), otherwise if this seems like the right track I'll add some visual diff tests to `input-text`.

components/inputs/input-text.js
```diff
@@ -150,5 +249,32 @@ class InputText extends RtlMixin(LitElement) {
...
+	}
+
+	_onSlotChange(e) {
+		const children = e.target.assignedNodes();
+		if (children.length > 0) {
+			const slotName = e.target.name;
+			const affectedSlotName = this.dir !== 'rtl' ? slotName : (slotName === 'left' ? 'right' : 'left');
+			const slotContent = children[0];
+			const slotContentWidth = slotContent.offsetWidth;
+			const diff = (this.slotPaddingWidth * documentFontSize - slotContentWidth) / 2;
```
dlockhart: Could you leverage the `calc` above instead of needing to use the document font size? Like here we could just set `this._slotLeftWidth = e.target.offsetWidth;` and then when it renders do a `calc(0.75rem + ${this._slotLeftWidth}px)`. Would need some tweaking.

components/inputs/input-text.js
```diff
@@ -150,5 +249,32 @@ class InputText extends RtlMixin(LitElement) {
...
+
+			slotContent.style[affectedSlotName] = `${diff}px`;
+			slotContent.addEventListener('focus', this._handleFocus);
+			slotContent.addEventListener('blur', this._handleBlur);
```
dlockhart: Do you need event handlers on each of the slot children? I don't think the search input was doing this. I was kind of expecting if you just fired up to the focus & blur on the surrounding custom element, then the events from inner elements would bubble up.

components/inputs/input-text.js
```diff
@@ -56,20 +81,73 @@ class InputText extends RtlMixin(LitElement) {
...
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		const slots = this.shadowRoot.querySelectorAll('slot');
+		slots.forEach((slot) => slot.addEventListener('slotchange', this._onSlotChange.bind(this)));
+
+		this.addEventListener('blur', this._handleBlur);
+		this.addEventListener('focus', this._handleFocus);
+
```
dlockhart: Could do this with inline events on the slots themselves: `<slot @slotchange="${this._handleBlur}">`

components/inputs/input-text.js
```diff
@@ -84,17 +162,29 @@ class InputText extends RtlMixin(LitElement) {
...
 		if (this.label && !this.labelHidden) {
-			return html`
+			input = html`
 				<label>
 					<span class="d2l-input-label">${this.label}</span>
 					${input}
 				</label>`;
 		}
+
+		if (this._hasSlot) {
```
dlockhart: To simplify the rendering and reduce the number of different permutations, does it hurt anything having the container even if there aren't any slots?

components/inputs/input-text.js
```diff
@@ -135,10 +225,19 @@ class InputText extends RtlMixin(LitElement) {
 
 	_handleInput(e) {
 		this.value = e.target.value;
+		this.dispatchEvent(new CustomEvent('d2l-input-container-input', { bubbles: true, composed: true }));
```
dlockhart: What's this needed for? Previously I've wired up to the `input` event and things have worked...

components/inputs/input-text.js
```diff
@@ -44,6 +52,23 @@ class InputText extends RtlMixin(LitElement) {
...
+					text-overflow: ellipsis;
+					white-space: nowrap;
+					-webkit-appearance: textfield;
+				}
+				::slotted(*) {
+					position: absolute;
+					top: 50%;
+					transform: translateY(-50%);
+				}
+				:host([label]) ::slotted(*) {
```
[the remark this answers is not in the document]
margaree: Yup that fixes it, thanks!

components/inputs/input-text.js
margaree: The problem is that `left` or `right` is being set on the slotcontent so it makes using a `styleMap` and using `calc` in `render more complicated/maybe not possible. I'd like to simplify this though so I'm looking into workarounds.

components/inputs/input-text.js
margaree: So with search input, tabbing from input to the search icon did not trigger the blur event (it would be triggered when the whole component had focus removed), but with the button icon in the slot, tabbing from input to the button icon triggers the blur event. I'll look into a way around this.

components/inputs/input-text.js
margaree: It's probably fine. I'll run the visual-diff tests to confirm.

components/inputs/input-text.js
margaree: I think it was something I had added when working on `input-date`. I'll remove for now and try to work without it.

margaree:
Any additional feedback for this PR? The only outstanding issue is the case where a button-icon would be on the left and focus order would be unintuitive, which isn't a use case that we currently have.

components/inputs/input-text.js
```diff
@@ -58,35 +85,71 @@ class InputText extends RtlMixin(LitElement) {
...
+					placeholder="${ifDefined(this.placeholder)}"
+					?readonly="${this.readonly}"
+					size="${ifDefined(this.size)}"
+					step="${ifDefined(this.step)}"
+					style="${styleMap(inputStyles)}"
+					tabindex="${ifDefined(this.tabindex)}"
+					type="${this._getType()}"
+					.value="${this.value}">
+					<div id="first-slot"><slot name="${firstSlotName}" @slotchange="${this._onSlotChange}"></slot></div>
+					<div id="last-slot"><slot name="${lastSlotName}" @slotchange="${this._onSlotChange}"></slot></div>
```
dlockhart: Super nit: these 2 divs are indented one too many I think

components/inputs/input-text.js
```diff
@@ -133,12 +200,24 @@ class InputText extends RtlMixin(LitElement) {
...
 
 	_handleKeypress(e) {
+		const eventDetails = {
+			bubbles: true,
+			composed: true,
+			detail: {
+				keyCode: e.keyCode
+			}
+		};
+		this.dispatchEvent(new CustomEvent('d2l-input-container-keypress', eventDetails));
```
dlockhart: Is this needed as part of this PR?

PR #412, opened 2020-03-06
Dbatiste/swipe gesture

dbatiste:
This PR is the first step in moving the swipe gesture code from [d2l-polymer-behaviors-ui](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-gestures-swipe.js).  It is used by image-based nav tray. It provides basic info - distance, direction, and duration of swipe.  Next step will be to remove that implementation, and update BSI to assign `window.D2L.Gestures.Swipe = {register: registerSwipe};`.

### DIFF
--- a/components/demo/demo-snippet.js
+++ b/components/demo/demo-snippet.js
@@ -51,10 +51,10 @@ class DemoSnippet extends LitElement {
 		text = text.replace(/^[\t]*<template>[\n]*/, '').replace(/[\n]*[\t]*<\/template>$/, '');
 
 		// fix script whitespace (for some reason brower keeps <script> indent but not the rest)
-		let lines = text.replace(/\t/g, '  ').replace(/<\/script>/g, '\n</script>').replace(/<script>/g, '<script>\n').split('\n');
+		let lines = text.replace(/\t/g, '  ').replace(/<\/script>/g, '\n</script>').replace(/<script>/g, '<script>\n').replace(/<script type="module">/g, '<script type="module">\n').split('\n');
 		let scriptIndent = 0;
 		lines = lines.map((l) => {
-			if (l.indexOf('<script>') > -1) {
+			if (l.indexOf('<script') > -1) {
 				scriptIndent = l.match(/^(\s*)/)[0].length;
 				return l;
 			} else if (l.indexOf('</script>') > -1) {
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -63,6 +63,28 @@ getOffsetParent(node);
 isComposedAncestor(ancestorNode, node);
 ```
 
+## Gesture - Swipe
+
+A simple helper for swipe touch gestures, providing distance, direction, and duration.
+
+```js
+import { registerGestureSwipe } from '@brightspace-ui/core/helpers/gestures.js';
+
+// sets up event listeners for swipe gesture
+registerGestureSwipe(element);
+
+// listen for custom swipe event
+element.addEventListener('d2l-gesture-swipe', (e) => {
+    console.log(
+        e.detail.distance,             // .x/.y
+        e.detail.direction.angle,      // deg, clock-wise from y-axis
+        e.detail.direction.horizontal, // left/right
+        e.detail.direction.vertical,   // up/down
+        e.detail.duration              // ms
+    );
+});
+```
+
 ## queueMicrotask
 
 A polyfill for [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask). For more information on microtasks, read [this article from Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide).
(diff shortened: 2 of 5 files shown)


components/demo/demo-snippet.js
```diff
@@ -51,10 +51,10 @@ class DemoSnippet extends LitElement {
 		text = text.replace(/^[\t]*<template>[\n]*/, '').replace(/[\n]*[\t]*<\/template>$/, '');
 
 		// fix script whitespace (for some reason brower keeps <script> indent but not the rest)
-		let lines = text.replace(/\t/g, '  ').replace(/<\/script>/g, '\n</script>').replace(/<script>/g, '<script>\n').split('\n');
+		let lines = text.replace(/\t/g, '  ').replace(/<\/script>/g, '\n</script>').replace(/<script>/g, '<script>\n').replace(/<script type="module">/g, '<script type="module">\n').split('\n');
```
dbatiste: This is a small fix to account for `<script type="module">` in demo snippet, otherwise indentation in the formatted code is not right.  Previously it only accounted for `<script>`.

components/demo/demo-snippet.js
```diff
@@ -51,10 +51,10 @@ class DemoSnippet extends LitElement {
...
 		let scriptIndent = 0;
 		lines = lines.map((l) => {
-			if (l.indexOf('<script>') > -1) {
+			if (l.indexOf('<script') > -1) {
```
dbatiste: This is a small fix to account for `<script type="module">` in demo snippet, otherwise indentation in the formatted code is not right.  Previously it only accounted for `<script>`.

helpers/README.md
```diff
@@ -63,6 +63,17 @@ getOffsetParent(node);
 isComposedAncestor(ancestorNode, node);
 ```
 
+## queueMicrotask
```
dbatiste: I just reordered this in the readme to make things alphabetically ordered.

helpers/gestures.js
```diff
@@ -0,0 +1,109 @@
+const maxTime = 2000;
```
dbatiste: This is the same impl as before, with `var` replaced, and using arrow functions where appropriate.

helpers/gestures.js
```diff
@@ -0,0 +1,109 @@
...
+	node.addEventListener('touchstart', handleTouchStart);
+}
+
+function handleTouchStart(e) {
+
+	const node = this; /* eslint-disable-line no-invalid-this */
+
+	let tracking = {
+		start: {
+			time: new Date().getTime(),
```
dlockhart: Might want to use the [performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) high resolution time stuff? It'll give you a much more accurate timestamp.

helpers/gestures.js
```diff
@@ -0,0 +1,109 @@
+const maxTime = 2000;
+const minDistance = 30;
+
+export function registerSwipe(node) {
+	node.addEventListener('touchstart', handleTouchStart);
```
dlockhart: Does this ever need to be unregistered?

dbatiste: Good question.  Currently, it would only get removed when the element is disposed, similar to how we wire up events to elements internal to a Lit component.  The original impl provided an unregister method (that was never used) that would be useful if a consumer wants to stop listening for touch events but they still want to keep the element.  Do you think I should provide that?

dlockhart: Would it even get removed though on disposal? I guess unlike events that get wired up to the document, touch events on the element aren't going to continue to fire long after the element's been removed from the DOM.

dbatiste: Yeah, it's just the ones that get wired-up outside that scope that are of concern.  

That's the approach being used in Lit components as well - the topic comes up every once a while in the Polymer Slack, with the general recommendation being to simply wire-up in `firstUpdated` for the `host`, although I think Justin mentioned it can be delayed even further by wiring up in `constructor() { setTimeout(() => { this.addEventListener... }, 0) }`, and only using `connectedCallback`/`disconnectedCallback` when wiring up to things outside its scope.  Long story short, they also assume handlers are cleaned up when the node is cleaned up.

That all being said, the unregister _could_ be useful to stop listening for touch... in effective disabling swipe for some reason.  However the consumer can also accomplish that by removing their `d2l-swipe` handler.  🤷‍♂

helpers/gestures.js
```diff
@@ -0,0 +1,116 @@
...
+		let vertical = 'none';
+		if (Math.abs(distanceY) >= minDistance) {
+			if (theta > 295 || theta < 65) {
+				vertical = 'up';
+			} else if (theta > 115 && theta < 245) {
+				vertical = 'down';
+			}
+		}
+
+		node.dispatchEvent(new CustomEvent('d2l-swipe', {
```
dlockhart: Nit: since the event name is the only thing that falls into the "API" category, and the filename is called `gestures.js` (I'm assuming more could go in here in the future), it might make sense to call the event `d2l-gesture-swipe`. But I understand there's existing code using this event name so I'm totally fine with keeping it as-is.

dbatiste: Yeah I like that...  I'll switch it.

PR #413, opened 2020-03-09
Focus radio-menu selection

Scy-D2L:
On focus d2l-menus with d2l-item-radio children will set focus to the selected menu item instead of the first one

### DIFF
--- a/components/menu/menu.js
+++ b/components/menu/menu.js
@@ -76,7 +76,7 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 	}
 
 	focus() {
-		if (this.children[0].tagName === 'D2L-MENU-ITEM-RADIO') {
+		if (this._items[0].role === 'menuitemradio') {
 			this._focusSelected();
 		} else {
 			this._focusFirst();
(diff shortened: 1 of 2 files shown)


components/menu/menu.js
```diff
@@ -76,7 +76,11 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 	}
 
 	focus() {
-		this._focusFirst();
+		if (this.children[0].tagName === 'D2L-MENU-ITEM-RADIO') {
```
dbatiste: Does this work with Edge?  I _think_ `this.children[0]` without shadowDOM would be `<div class="d2l-menu-items d2l-hierarchical-view-content">`.  If this is an issue, it may be avoided by calling `assignedNodes()` (and filtering) or `assignedElements()` of the slot, though care must be taken with the later, since I think it was not properly supported in Safari until very recently.

dbatiste: On looking at this further, can we utilize `this._items`?  It looks like we handle a bunch of this ^.

Also, I think it would be better to check `role === 'menuitemradio'` rather than the `tagName`.

PR #414, opened 2020-03-09
Dropdown menu scrolls selected item to visibility when opened

Scy-D2L:


components/dropdown/dropdown-menu.js
```diff
@@ -65,6 +65,15 @@ class DropdownMenu extends DropdownContentMixin(LitElement) {
 				menu.focus();
 			}, 0);
 		}
+
```
dlockhart: I believe applying focus will also scroll the item into view. If so, then I think the scrolling code should be in the `else` block of the `if (this__applyFocus) {`.

components/menu/menu.js
```diff
@@ -76,13 +76,29 @@ class Menu extends HierarchicalViewMixin(LitElement) {
...
 	focus() {
-		if (this._items[0].role === 'menuitemradio') {
+		if (this.getMenuType() === 'menu-radio') {
 			this._focusSelected();
 		} else {
 			this._focusFirst();
 		}
 	}
 
+	getMenuType() {
```
dlockhart: Might be nice if this returned `'menu'` if there were no items? That way it always returns something.

components/dropdown/dropdown-menu.js
Scy-D2L: This scroll code also vertically centers the selected item in the visible space, which focus doesn't. So I can do that but it would be a minor inconsistency in the behavior.

dlockhart: Ah OK, thanks. I'm good with keeping it so it's consistent!

components/menu/menu.js
```diff
@@ -76,13 +76,31 @@ class Menu extends HierarchicalViewMixin(LitElement) {
...
+				case 'menuitemcheckbox':
+					return 'menu-checkbox';
+
+				case 'menuitem':
+				default:
+					return 'menu';
+			}
+		} else {
+			return 'menu';
+		}
```
dlockhart: Super nit: I like using `if`-statement to bail out quickly and then reduce the nesting of the rest of my blocks. In cases like this it also results in less code. For example:

```
if (this._items.length === 0) {
    return 'menu';
}
switch (this._items[0].role) {
	case 'menuitemradio':
		return 'menu-radio';
	case 'menuitemcheckbox':
		return 'menu-checkbox';
	case 'menuitem':
	default:
		return 'menu';
}
```

components/menu/test/menu.test.js
```diff
@@ -115,6 +115,20 @@ describe('d2l-menu', () => {
 			expect(document.activeElement).to.equal(elem.querySelector('#b1'));
 		});
 
+		it('focus method sets focus to selected radio menu item', async() => {
```
dlockhart: Nit: typically with the describe/it Mocha syntax, we try to make the `it` tests read like a sentence. So a different way to write the test name would be: `it('focuses on the selected radio menu item when focus() is called', ...`

Scy-D2L: Sounds more useful :) though some other tests in that file aren't following that pattern.

PR #422, opened 2020-03-11
Moving round borders to illustration slot

zinabat:
Rounded corners for images in list items are always used on demo pages and are shown on [design.d2l/components/lists](http://design.d2l/components/lists/). This change moves the rounded corners from the demo images to the illustration slot itself.

The change will provide consistency and make several related defects moot (e.g., [DE38050](https://rally1.rallydev.com/#/detail/defect/372781442368?fdp=true)).

The change can easily be overridden with custom CSS in components that require square corners.


dbatiste:
For anyone wondering why a few of the dialog goldens are removed here, it is because I removed those tests when I refactored the focus trap logic, but I forgot to regenerate the goldens to remove the golden images for them.

PR #424, opened 2020-03-13
Tooltip (1/2): Add tooltip show/hide logic

AllanKerr:
# Changes
- This pull request adds the logic to show the tooltip on `mouseenter`, `focus`, and if `force-show` is set.
- The tooltip is closed when it isn't hovered, focused, or `force-show` is set.

## Next PR
- Layout and positioning

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -1,0 +1,238 @@
+import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { bodyCompactStyles } from '../typography/styles.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class Tooltip extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			delay: { type: Number },
+			disableFocusLock: { type: Boolean, attribute: 'disable-focus-lock' },
+			for: { type: String },
+			forceShow: { type: Boolean, attribute: 'force-show' },
+			showing: { type: Boolean, reflect: true, attribute: 'showing' },
+			state: { type: String, reflect: true }, /* Valid values are: 'info' and 'error' */
+		};
+	}
+
+	static get styles() {
+		return [bodyCompactStyles, css`
+			:host {
+				--d2l-tooltip-background-color: var(--d2l-color-ferrite); /* Deprecated, use state attribute instead */
+				background-color: var(--d2l-tooltip-background-color);
+				box-sizing: border-box;
+				color: white;
+				display: none;
+				position: absolute;
+				text-align: left;
+				white-space: normal;
+				z-index: 1000; /* position on top of floating buttons */
+			}
+
+			:host([state="error"]) {
+				--d2l-tooltip-background-color: var(--d2l-color-cinnabar);
+			}
+
+			:host([dir="rtl"]) {
+				text-align: right;
+			}
+
+			:host([showing]) {
+				display: inline-block;
+			}
+		`];
+	}
+
+	constructor() {
+		super();
+
+		this._onMouseEnter = this._onMouseEnter.bind(this);
+		this._onMouseLeave = this._onMouseLeave.bind(this);
+		this._onFocus = this._onFocus.bind(this);
+		this._onBlur = this._onBlur.bind(this);
+
+		this.delay = 0;
+		this.disableFocusLock = false;
+		this.forceShow = false;
+		this.showing = false;
+		this.state = 'info';
+
+		this._isFocusing = false;
+		this._isHovering = false;
+		this._dismissibleId = null;
+	}
+
+	get for() {
+		return this._for;
+	}
+	set for(val) {
+		const oldVal = this._for;
+		if (oldVal !== val) {
+			this._for = val;
+			this.requestUpdate('for', oldVal);
+			this._updateTarget();
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+class Tooltip extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			delay: { type: Number },
+			disableFocusLock: { type: Boolean, attribute: 'disable-focus-lock' },
+			for: { type: String },
+			forceShow: { type: Boolean, attribute: 'force-show' },
+			showing: { type: Boolean, reflect: true, attribute: 'showing' },
+			state: { type: String, reflect: true }, /* Valid values are: 'info' and 'error' */
```
margaree: Does `state` being info actually do anything, or is `error` the only value that matters? Would it make sense to just have a Boolean of `error` or something like that?

dlockhart: I personally don't mind the state being a string -- it's the same as we did for `status-indicators`. It'll allow us to have other states in the future if we need them I guess?

AllanKerr: It was inspired by `state` in `status-indicators`. I added it because before it was entirely configurable with CSS variables if people needed different colors or styles. Now `state` gives a clear place to add other possible variants if people need them in the future.

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+
+class Tooltip extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			delay: { type: Number },
+			disableFocusLock: { type: Boolean, attribute: 'disable-focus-lock' },
+			for: { type: String },
+			forceShow: { type: Boolean, attribute: 'force-show' },
+			showing: { type: Boolean, reflect: true, attribute: 'showing' },
```
dlockhart: Don't think you need `attribute` here since it's the same?

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+	get for() {
+		return this._for;
+	}
+	set for(val) {
+		const oldVal = this._for;
+		if (oldVal !== val) {
+			this._for = val;
+			this.requestUpdate('for', oldVal);
+			this._updateTarget();
+		}
```
dlockhart: I wonder if we need the accessor or if just using the `attributeChangedCallback` and checking for it in there to trigger `_updateTarget` would work? Same for `forceShow` and `showing`.

AllanKerr: That'll definitely work. For my own understanding, is there a reason to prefer one versus the other?

I was leaning towards using `setters` / `getters` since they'll only run when that attribute changes versus `attributeChangedCallback` which runs whenever any attribute changes and needs to explicitly check each time to see what the name of the changed attribute was to determine which code block to run.

dlockhart: I don't have enough experience with them to have a strong opinion. I think setters/getters are obviously great if you need to validate/normalize the value before actually letting it get through. But yeah, don't feel strongly -- good with the getters/setters here.

dbatiste: I've also been contemplating this a bit.  I've also found `update(changedProperties)` to be worthwhile - it gets called right after `attributeChangedCallback`, before `render`.

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+
+	connectedCallback() {
+		super.connectedCallback();
+
+		requestAnimationFrame(() => {
+			this._updateTarget();
+		});
+	}
+
+	disconnectedCallback() {
```
dbatiste: If the tooltip is disconnected from the DOM, it looks like the listeners on the target will not be removed because they are not part of the tootip's sub-tree.  The listeners will continue to hold references to everything accessible within the handler's closure (like `d2l-tooltip`), preventing it from being cleaned up.  It is a scenario that we might not expect, but I think it could easily arise.

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+		this._isHovering = false;
+		this._isFocusing = false;
+		this._updateShowing();
+	}
+
+	show() {
+		this.showing = true;
+	}
+
+	_updateTarget() {
```
dbatiste: Nit: let's order these private methods alpha to make locating things easier.

components/tooltip/tooltip.js
AllanKerr: Good catch, I'll add `_removeListeners`

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+		if (!this._target) {
+			return;
+		}
+		this._target.removeEventListener('mouseenter', this._onMouseEnter);
+		this._target.removeEventListener('mouseleave', this._onMouseLeave);
+		this._target.removeEventListener('focus', this._onFocus);
+		this._target.removeEventListener('blur', this._onBlur);
+	}
+
+	_onMouseEnter() {
```
dbatiste: For clarity, I think I would name these handlers: `_onTarget...`.

components/tooltip/tooltip.js
```diff
@@ -0,0 +1,238 @@
...
+		this._target = target;
+		this._addListeners();
+
+		if (this.showing) {
+			// TODO: Perform layout
+		}
+	}
+
+	_findTarget() {
+		const parentNode = this.parentNode;
```
dbatiste: Nit: looks like `parentNode` is only needed in the `else` block below and so this could be moved there.

PR #425, opened 2020-03-16
Updating dialog styles for long buttons inside footer.

dbatiste:


components/dialog/dialog-styles.js
```diff
@@ -101,14 +101,15 @@ export const dialogStyles = css`
 	.d2l-dialog-footer {
 		box-sizing: border-box;
 		flex: none;
-		padding: 30px;
+		padding: 30px 30px 12px 30px; /* 18px margin below footer children */
```
dlockhart: Are we worried about cases where dialogs had footers but didn't have things in them? Their heights would change. Haha probably not though... that's a strange case.

dbatiste: Yeah that is a weird case - but it has come up. Before this change, if they had an empty footer then they would have crazy amount of whitespace (60px below their content).  The case arises when someone tries to put a component in the content slot while having its own workflow buttons, and so far we've suggested that the component should be refactored to allow buttons in the footer slot and provide consistent scrolling/sticky footer behavior. With this change, they'd see 42px whitespace.  To deal with it, I think we'd need to check for footer elements and apply special style to achieve the 30px whitespace.

dbatiste:
Not sure why Travis did not report status update.  All builds look good.

PR #426, opened 2020-03-17
S akulick/time dropdown

Scy-D2L:
Added a dropdown menu to `input-time`
Offers options between 0-23:30 in half hour increments, along with 23:59
All options displayed in the user's locale
Timezone displayed in the footer
Selecting option in dropdown updates input in textbox
Changing text in textbox alters selection when dropdown is opened (if input value matches an increment it is selected, otherwise no value is selected)

Includes unit and v-diff tests


components/inputs/input-time.js
```diff
@@ -1,12 +1,29 @@
...
 import { formatTime, parseTime } from '@brightspace-ui/intl/lib/dateTime.js';
+import { bodySmallStyles } from '../typography/styles.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
 import { inputStyles } from './input-styles.js';
 
 const VALUE_RE = /^([0-9]{1,2}):([0-9]{1,2})(:([0-9]{1,2}))?$/;
 const TODAY = new Date();
 const DEFAULT_VALUE = new Date(TODAY.getFullYear(), TODAY.getMonth(), TODAY.getDate(), 0, 0, 0);
+const INTERVALS = getIntervals();
```
dlockhart: So this code is going to run immediately in the browser when the script is included -- whether or not a time input is actually rendered on the page. Therefore, it might be better to just call `getIntervals()` directly in the `render()` method. It'll run each time the component is rendered, but won't run when it's not on the page.

Another option to get the best of both worlds would be something like:
```
let intervals = null;
function getIntervals() {
    if (intervals !== null) return intervals;
    ...
}
```

components/inputs/input-time.js
```diff
@@ -69,6 +86,16 @@ class InputTime extends LitElement {
 				label {
 					display: block;
 				}
+				.d2l-input-time-menu {
+					border-top: 1px solid var(--d2l-color-gypsum);
+					border-bottom: 1px solid var(--d2l-color-gypsum);
```
dlockhart: Hmm... ideally we're not setting borders and stuff on things like menus -- since they already have their own borders and us "knowing about them" makes the code a lot more fragile. Why is this needed?

components/inputs/input-time.js
```diff
@@ -79,6 +106,10 @@ class InputTime extends LitElement {
 		this.labelHidden = false;
 		this._value = formatValue(DEFAULT_VALUE);
 		this._formattedValue = formatTime(DEFAULT_VALUE);
+		this._timezone = formatTime(new Date(), {format: 'ZZZ'});
```
dlockhart: Super-nit: could use `TODAY` const.

components/inputs/input-time.js
```diff
@@ -92,13 +123,40 @@ class InputTime extends LitElement {
...
+						?disabled="${this.disabled}"
+						@keypress="${this._handleKeypress}"
+						.value="${this._formattedValue}">
+				</div>
+				<d2l-dropdown-menu id="dropdown" no-padding-footer min-width="195">
+					<d2l-menu
+						id="time-menu"
+						role="listbox"
+						class="d2l-input-time-menu"
+						aria-label="${ifDefined(this.label)}"
```
dlockhart: Would using `aria-labelledby` and pointing at the `<span class="d2l-input-label">` (you'll need to put an ID on it) work as well? Reduces the number of times the label text is repeated.

components/inputs/input-time.js
```diff
@@ -92,13 +123,40 @@ class InputTime extends LitElement {
...
+						aria-label="${ifDefined(this._getAriaLabel())}"
+						@change="${this._handleChange}"
+						class="d2l-input"
+						?disabled="${this.disabled}"
+						@keypress="${this._handleKeypress}"
+						.value="${this._formattedValue}">
+				</div>
+				<d2l-dropdown-menu id="dropdown" no-padding-footer min-width="195">
+					<d2l-menu
+						id="time-menu"
```
dlockhart: We don't use `ID`s a lot because for browsers who don't support shadow DOM all of this gets slammed into the same DOM tree, and then multiple instances of a control result in duplicate IDs. Not sure if that's actually a problem or not, but one strategy we've used is to just generate ids (there's a helper for it).

components/inputs/input-time.js
```diff
@@ -131,6 +189,14 @@ class InputTime extends LitElement {
 		return undefined;
 	}
 
+	__onOpened() {
+		this.shadowRoot.getElementById('combo-container').setAttribute('aria-expanded', 'true');
```
dlockhart: I would have expected dropdown to do this for you? I thought it looked for a parent with `d2l-dropdown-opener` and applied them automatically...

components/inputs/input-time.js
Scy-D2L: After checking - the bottom isn't helping, I can get rid of it, that border can come from the header and footer slots now that they exist. The top is the difference between these two: 
image that horizontal line under the top 'padding' isn't there (without this line) unless we have header content or you've scrolled a bit, so you kind of get what looks like inconsistent whitespace and a hovering scrollbar. If we're fine with that I can get rid of both of them.

components/inputs/input-time.js
Scy-D2L: I think the concern is that if we're using `label-hidden` that span doesn't exist 🤷‍♀

components/inputs/input-time.js
dlockhart: If a consistent border is something we want, I'd expect us to fix this at the dropdown-menu level, not in this instance of it. @geurts do you have opinions here? Dropdowns seem to only spawn that top border when there's a header or if they're scrolled -- are you OK with time inputs just behaving the same way?

components/inputs/input-time.js
dlockhart: Gotcha yep, sorry forgot that was conditionally rendered.

components/inputs/input-time.js
Scy-D2L: Minor input: most menus don't have a footer by default so there isn't a concern with them being 'bottom heavy' or vertically asymmetrical. In this case we have the timezone in the footer at the bottom so not having a visible top 'cap' isn't necessarily balanced. But that's an opinion, I'll defer to Jeff's judgement. :)

dlockhart: If that's the logic we'd like to apply, I think we could still put it inside dropdown/menu where the header/footer is defined. They know that there's a footer but not a header so could apply the border if that's how we want to treat it.

components/inputs/input-time.js
```diff
@@ -92,13 +123,40 @@ class InputTime extends LitElement {
...
+						@change="${this._handleChange}"
+						class="d2l-input"
+						?disabled="${this.disabled}"
+						@keypress="${this._handleKeypress}"
+						.value="${this._formattedValue}">
+				</div>
+				<d2l-dropdown-menu id="dropdown" no-padding-footer min-width="195">
+					<d2l-menu
+						id="time-menu"
+						role="listbox"
```
dbatiste: I'm definitely unsettled on overriding the role given the Aria model.  What are your thoughts on this... using this to get by in the short-term, and then introduce a real `listbox` (refactoring menu and its styles as appropriate)?

Scy-D2L: Wouldn't that involve a new component that we wanted to avoid when we were discussing the radio-menu and the aria role issues earlier?

components/inputs/test/input-time.test.js
```diff
@@ -23,24 +27,24 @@ describe('d2l-input-time', () => {
...
-		});
+		}).timeout(3000);
 
 		it('passes all axe tests when focused', async() => {
 			const elem = await fixture(basicFixture);
 			setTimeout(() => getInput(elem).focus());
 			await oneEvent(elem, 'focus');
 			await expect(elem).to.be.accessible();
-		});
+		}).timeout(5000);
```
Scy-D2L: Just a note on these (and this longest one in particular): it looks like something about `d2l-menu-item-radio` slows these tests down a lot. After chatting with Dave L. it was decided to investigate that issue separately and increase the time limits on these for now.

components/inputs/input-time.js
```diff
@@ -91,14 +115,45 @@ class InputTime extends LitElement {
 	}
 
 	render() {
+		if (INTERVALS === null) {
```
dlockhart: Can you put this check in the `getIntervals` method itself? That way it's the one gating access to the variable, instead of every place that needs the intervals (which is just render right now).

components/inputs/test/input-time.test.js
```diff
@@ -102,14 +106,30 @@ describe('d2l-input-time', () => {
...
 			dispatchEvent(elem, 'change', false);
 			await oneEvent(elem, 'change');
 			expect(elem.value).to.equal('11:45:00');
 		});
 
+		it('should update value when dropdown changes', async() => {
+			const elem = await fixture(fixtureWithValue);
+			expect(elem.getTime(elem).second).to.equal(33);
+			getFirstOption(elem).click();
+			await aTimeout(1);
```
dlockhart: Can you await the change event instead of using a timeout? Timeouts are generally kind of flaky.

components/inputs/test/input-time.test.js
```diff
@@ -102,14 +106,30 @@ describe('d2l-input-time', () => {
...
 			expect(elem.getTime().second).to.equal(33);
 			getInput(elem).value = '11:45 AM';
 			dispatchEvent(elem, 'change', false);
 			await oneEvent(elem, 'change');
 			expect(elem.value).to.equal('11:45:00');
 		});
 
+		it('should update value when dropdown changes', async() => {
+			const elem = await fixture(fixtureWithValue);
+			expect(elem.getTime(elem).second).to.equal(33);
```
dlockhart: I don't typically like asserting 2 things in a single test. Instead, I like building up my tests such that the other tests have already covered certain assumptions. i.e. have a test that asserts that when you give it a value, that the value is correctly set. Then you don't need to worry about asserting that anywhere else, and this test can just test the one thing it wants to -- that when the dropdown is clicked, the value is reflected.

Scy-D2L:
Do we have any more concerns before this gets merged in?

PR #428, opened 2020-03-18
Calendar: Accessibility Improvements

margaree:
Changes include:
- Adding `role`s
- Adding additional `aria-` attributes where necessary
- Adding full date as `aria-label` on dates
- Moving the {month} {year} and button icon heading out of the `table`

These changes are based on http://eureka2.github.io/ab-datepicker/ and https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/datepicker-dialog.html


components/calendar/calendar.js
```diff
@@ -325,57 +331,66 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
-							<td>
-								<d2l-button-icon
-									@click="${this._onNextMonthButtonClick}"
-									text="${this._computeText(getNextMonth(this._shownMonth))}"
-									icon="tier1:chevron-right">
-								</d2l-button-icon>
-							</td>
-						</tr>
-						<tr>${weekdayHeaders}</tr>
+			<div aria-labelledby="${this._tableInfoId}-heading" class="d2l-calendar">
```
dlockhart: I'm curious... since this `<div>` doesn't have a role or anything, does the label actually do anything?

components/calendar/calendar.js
```diff
@@ -325,57 +331,66 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
+					</d2l-button-icon>
+				</div>
+				<table
+					aria-activedescendant="${activeDate}"
+					aria-labelledby="${this._tableInfoId}-heading"
+					aria-readonly="true"
+					role="grid"
+					summary="${ifDefined(this.summary)}">
+					<thead aria-hidden="false" role="presentation">
+						<tr role="row">${weekdayHeaders}</tr>
```
dlockhart: I'm surprised these `tr`s and the `th`s inside them needed row/gridheader roles... is it because of the `grid` role on the table itself that's turning it into something other than a table? If so, I think I'd expect the `td`s to need `presentation` roles too.

components/calendar/calendar.js
```diff
@@ -449,13 +464,15 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 				} else {
 					numDaysChange = -1;
 				}
+				preventDefault = true;
```
dlockhart: Stops scrolling?

components/calendar/calendar.js
```diff
@@ -325,57 +331,66 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
 							@click="${this._onDateSelected}"
 							@keydown="${this._onKeyDown}"
 							class=${classMap(classes)}
-							data-date=${day.getDate()}
-							data-month=${day.getMonth()}
-							data-year=${day.getFullYear()}
+							data-date=${date}
+							data-month=${month}
+							data-year=${year}
+							headers="${this._tableInfoId}-weekday-${index}"
```
dlockhart: Typically just having things in standard table markup gets you accessible header handling for free, so I'm curious why this is necessary. Might moving the role/aria-* stuff to the `<td>` fix things?

components/calendar/calendar.js
margaree: Stops safari voiceover from behaving not expectedly - it seems to go deep into the text when left/right were pressed rather than staying at the cell level, which was causing it not to read the aria-label.

components/calendar/calendar.js
margaree: ah nope I don't think it is needed. I was debating if this one was needed. It looks like there might be some form of this needed when `calendar` is in `dropdown` but I'll take a look more at that in that context.

components/calendar/calendar.js
margaree: I was thinking of trying to restructure this part to do that so I'll give a stab at it.

components/calendar/calendar.js
dlockhart: Cool -- can you throw in a comment saying something to that effect? That way we know what it's for in the future and better judge if we can clean it up.

components/calendar/calendar.js
margaree: After some experimenting it seems the best option is to remove the `row` and `presentation` roles. They don't seem to improve anything and different combinations cause different/undesired behavior in some browsers (most problematic seemed to be chrome + voice over)

components/calendar/calendar.js
```diff
@@ -326,53 +334,58 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
-							@keydown="${this._onKeyDown}"
-							class=${classMap(classes)}
-							data-date=${day.getDate()}
-							data-month=${day.getMonth()}
-							data-year=${day.getFullYear()}
-							tabindex=${focused ? '0' : '-1'}>
-							${day.getDate()}
-						</div>
+					<td
+						aria-label="${formatDate(day, {format: 'medium'})}"
```
margaree: I used medium instead of full because the column header information adds the context of the weekday name

margaree:
Is there any additional feedback for this PR?

PR #429, opened 2020-03-18
Tooltip (2/2): Add tooltip positioning and sizing logic

AllanKerr:
# Positioning
positioning
1. `d2l-tooltip-target-position` is positioned so that it is always `offset` pixels away from the target to the `left`, `right`, `top`, or `bottom`. Everything in the tooltip is positioned relative to this. This is needed because the `for` attribute means the tooltip is usually not a child of its target.
2. `d2l-tooltip-position` is used:
    - To shift the tooltip relative to its target. It can slide `left` or `right` for tooltips that are above or below their target or `up` and `down` for tooltips that are to the left or right of their target.
    - To constrain the width of the inner `d2l-tooltip-content`. It limits the width based on the width of the available space which is based on the lesser of the boundary and view port. Height is allowed to freely grow.

# Sizing
sizing-left-right
sizing-top-bottom

To size the tooltip, the size of the four available spaces, left, right, above, and below, are computed as the space between the tooltip target and the viewport bounds. [This is the same method `d2l-dropdown-content` uses to compute its space around](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-content-mixin.js#L454). The spaces can be further constrained using the `boundary` attribute.

For each space, the `tooltip` max-width is constrained to the width of the space then the `scrollWidth` and `scrollHeight` are checked to see if they fit inside the available space. If they don't fit, the next space is tried until an available space is found. If none of the four spaces are large enough to fully contain the tooltip then the space with the largest area is chosen.


components/tooltip/tooltip.js
```diff
@@ -128,6 +308,57 @@ class Tooltip extends RtlMixin(LitElement) {
...
+		if (this.getAttribute('dir') === 'rtl') {
+			const tmp = spaces[2];
+			spaces[2] = spaces[3];
+			spaces[3] = tmp;
+		}
+		return spaces;
+	}
+
+	_computeTooltipShift(centerDelta, spaceLeft, spaceRight) {
+
```
dlockhart: A few of these methods seem like they could move out of the class itself and just live up above in the file. That way they stay hidden to consumers and we also likely get a small performance boost as the JS classes that get created over and over again are smaller.

components/tooltip/tooltip.js
```diff
@@ -98,9 +305,39 @@ class Tooltip extends RtlMixin(LitElement) {
...
+			}
+		}
+
 		return html`
-			<slot>
-			</slot>`
+			<div class="d2l-tooltip-target-position" style=${styleMap(targetPositionStyle)}>
+				<div class="d2l-tooltip-container">
+					<div class="d2l-tooltip-position" style=${styleMap(tooltipPositionStyle)}>
+						<div class=" d2l-body-small d2l-tooltip-content">
```
dlockhart: Nit: extra space

PR #430, opened 2020-03-20
Initial migration of d2l-tabs to lit-element.

dbatiste:
Lit version of `d2l-tabs`.  Has complete feature parity with the Polymer version, with the explicit exclusion of `auto-select`.

Current rendering:
image


components/tabs/tab-panel-mixin.js
```diff
@@ -0,0 +1,65 @@
...
+		if (this.id.length === 0) this.id = getUniqueId();
+	}
+
+	async attributeChangedCallback(name, oldval, newval) {
+		super.attributeChangedCallback(name, oldval, newval);
+		if (name === 'selected') {
+			if (this.selected) this._dispatchSelected();
+		} else if (name === 'text') {
+			this.setAttribute('aria-label', this.text);
+			this.dispatchEvent(new CustomEvent(
```
dbatiste: This event existed previously - it's primary purpose it to notify `d2l-tabs` so that it can update measurements / visibility of the scroll buttons.

components/tabs/tab-panel-mixin.js
```diff
@@ -0,0 +1,65 @@
...
+				'd2l-tab-panel-text-changed', { bubbles: true, composed: true, detail: { text: this.text } }
+			));
+		}
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		/*
+		if (this.selected) {
```
dbatiste: This was from original, not sure it's needed anymore.  So far it is not.

components/tabs/tab.js
```diff
@@ -0,0 +1,132 @@
...
+		this.tabIndex = -1;
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this.addEventListener('click', () => {
+			this.ariaSelected = 'true';
+		});
+		// perhaps this should be keyup?
+		this.addEventListener('keydown', (e) => {
```
dbatiste: The existing component wires up to `keydown`, but perhaps it should be `keyup`.  I'm not sure if this was missed during WCAH 2.1 work, or if there is a legit reason for it to still use `keydown`.

components/tabs/tab.js
```diff
@@ -0,0 +1,132 @@
...
+	update(changedProperties) {
+		super.update(changedProperties);
+		changedProperties.forEach((oldVal, prop) => {
+			if (prop === 'ariaSelected' && this.ariaSelected === 'true') {
+				this.dispatchEvent(new CustomEvent(
+					'd2l-tab-selected', { bubbles: true, composed: true }
+				));
+			} else if (prop === 'text') {
+				this.title = this.text;
+				// need to measure again
```
dbatiste: The size of the `d2l-tab` text affects the virtual scrolling in `d2l-tabs`.  Remains to be seen how the measurements will be done... this comment is simply a reminder that previously the measurements were updated here.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { getNextFocusable, getPreviousFocusable } from '../../helpers/focus.js';
+import { ArrowKeysMixin } from '../../mixins/arrow-keys-mixin.js';
+import { bodyCompactStyles } from '../typography/styles.js';
+import { findComposedAncestor } from '../../helpers/dom.js';
+import { LocalizeStaticMixin } from '../../mixins/localize-static-mixin.js';
+import {repeat} from 'lit-html/directives/repeat';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+//const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
```
dbatiste: This will be used if necessary in the code where normally we would wait for animation/transition events.  If reduced motion is preferred and we turn off transitions, that code will need to conditionally account for the absence of those events.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+	static get properties() {
+		return {
+			maxToShow: { type: Number },
+			_tabInfos: { type: Array }
+		};
+	}
+
+	static get styles() {
+		return [bodyCompactStyles, css`
+			:host {
```
dbatiste: All these styles are the same as the original version, with the exception of updates for `:host(dir="rtl")`.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+			}
+			.d2l-tabs-scroll-button:hover,
+			.d2l-tabs-scroll-button:focus {
+				background-color: var(--d2l-color-gypsum);
+			}
+			.d2l-tabs-scroll-button:focus {
+				border-color: var(--d2l-color-celestine);
+			}
+
+			@media (prefers-reduced-motion: reduce) {
```
dbatiste: Here, we will override `transition` as necessary to opt out of transitions.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+
+//const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
+
+//const scrollButtonWidth = 56;
+
+class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
+
+	static get properties() {
+		return {
+			maxToShow: { type: Number },
```
dbatiste: I did not include `autoSelect` on the API because it is not used - it was something that we originally developed but opted not to use for a11y reasons.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+			else
+				update the virtual scroll position so that the browser doesn't force a scroll
+
+		};
+		*/
+	}
+
+	render() {
+
+		return html`
```
dbatiste: The rendering here is exactly the same except that we update for the new `localize` API, use the `@` directive to wire-up to events in almost all cases, use the `repeat` directive to render the `d2l-tab` elements, use the `ArrowKeysMixin` template helper, and wire-up to the slot change event to observe panel changes.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+	_handleFocusStart(e) {
+		if (e.relatedTarget && e.relatedTarget.role === 'tab') {
+			const previousFocusable = getPreviousFocusable(e.target, false);
+			if (previousFocusable) previousFocusable.focus();
+		} else {
+			this._focusSelected();
+		}
+	}
+
+	_handlePanelsSlotChange(e) {
```
dbatiste: This `slotchange` handler replaces the original `observeNodes` Polymer handler.  The purpose is the same - to inspect the panels and create tabs for them, but the logic is slightly different because this event does not provide added/removed nodes, and rather than appending `d2l-tab` elements directly to the DOM, we update and array, `this._tabInfos` which is used by `render` to update the DOM.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this.arrowKeysFocusablesProvider = async() => {
+			return [...this.shadowRoot.querySelectorAll('d2l-tab')];
+		};
+		/*
+		this.arrowKeysOnBeforeFocus = async(tab) => {
```
dbatiste: This will be updated in follow-up commits.  Basically, we use this callback to update the virtual scroll position before focus is applied to prevent the browser from (automatically) natively scrolling the focused element into view.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+
+			}
+		`];
+	}
+
+	static get resources() {
+		return {
+			'ar': { 'scroll.previous': 'التمرير إلى اليسار', 'scroll.next':'التمرير إلى اليمين' },
+			'en': { 'scroll.previous': 'Scroll Backward', 'scroll.next': 'Scroll Forward' },
+			'es': { 'scroll.previous': 'Desplácese a la izquierda', 'scroll.next': 'Desplácese a la derecha' },
```
dlockhart: We should get translations for German and Danish, as they're languages we now support.

components/tabs/tabs.js
```diff
@@ -0,0 +1,388 @@
...
+	_getPanels(slot) {
+		if (!slot) return;
+		return slot.assignedNodes({flatten: true})
+			.filter((node) => node.nodeType === Node.ELEMENT_NODE && node.role === 'tabpanel');
+	}
+
+	_getTabInfo(id) {
+		for (let i = 0; i < this._tabInfos.length; i++) {
+			if (this._tabInfos[i].id === id) return this._tabInfos[i];
+		}
```
dlockhart: Nit, could use:

```
return this._tabInfos.find( (t) => t.id === id );
```

Although if it doesn't find anything, `undefined` is returned not `null`.
```

File: components/tabs/tabs.js [hunk 14]
```diff
@@ -0,0 +1,388 @@
...
+		e.stopPropagation();
+		const selectedTab = e.target;
+		const selectedPanel = this._getPanel(selectedTab.controlsPanel);
+
+		/*
+			update the virtual scroll position to bring selected tab into view
+		*/
+
+		selectedPanel.selected = true;
+		for (let i = 0; i < this._tabInfos.length; i++) {
```
dlockhart: Since you don't need `i`, a `foreach` might be a bit cleaner here.

components/tabs/tab.js
```diff
@@ -0,0 +1,132 @@
+import '../colors/colors.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
```
dlockhart: `<d2l-tab>` seems to be a "private" element -- like it's only used by `<d2l-tabs>` right? Wondering if we could do more to "hide" it so that others don't mistakenly use it....

components/tabs/tab-panel-mixin.js
```diff
@@ -0,0 +1,65 @@
+import { css } from 'lit-element/lit-element.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
+
+export const TabPanelMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			noPadding: { type: Boolean, attribute: 'no-padding', reflect: true },
+			role: { type: String, reflect: true },
```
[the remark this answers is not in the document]
dbatiste: I can probably remove this.  Without this, the `role` will not get reflected on the elements as an attribute, but that perhaps does not matter for screen readers.  Currently, we are not using the attribute for CSS selectors or for querying.

components/tabs/tab-panel.js
```diff
@@ -0,0 +1,12 @@
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { TabPanelMixin } from './tab-panel-mixin.js';
+
+class TabPanel extends TabPanelMixin(LitElement) {
```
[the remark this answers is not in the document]
dbatiste: Currently `TabPanelBehavior` is used by [facet-filter-sort](https://search.d2l.dev/xref/GitHub/BrightspaceUI/facet-filter-sort/components/d2l-filter-dropdown/d2l-filter-dropdown-category.js?r=76966fd0#15) to provide a custom panel.  I can merge it, but if we convert that component (which my understanding is that it is the go-forward component for), we'd need the mixin.  We could not provide the mixin, and wait & see?

components/tabs/tabs.js
dbatiste: Yeah I was considering that, but elected not to use it since IE11 is in limbo...?

components/tabs/tab-panel.js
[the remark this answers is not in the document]
dlockhart: Cool gotcha, I'm good with keeping it as long as there's a clear potential value.

components/tabs/tabs.js
dlockhart: Pretty sure it'll get Babelified for IE11.

components/tabs/tab.js
dbatiste: At one point I'd gone down the path of baking this into `d2l-tabs` completely, but it tends to bloat the code.  I could move the component into the `d2l-tabs.js` file so that it's not quite so "public"?  It would avoid the extra HTTP request too.  That still would not prevent someone from using it though. There might be a way to create the elements inside of `render` using the Tab class, without having to use `customElements.define('d2l-tab', Tab);`... if so, that would truly make it private.

dlockhart: Possibly yeah. Or you could register it as `d2l-tab-internal-private-do-not-use-extra-rich-and-unsafe` haha.

dbatiste: I renamed `d2l-tab` to `d2l-tab-internal`.  While not strictly private, it should make it clear that consumers should not try to use it.

I tried constructing the element from `Tab` and not using the `customElements.define` API but I didn't have much luck.  While it's simple enough to construct `Tab`, I wasn't able to figure out how to append it to the DOM without the define API... there might be a way, but it's certainly not "advertised" and probably not recommended.

dlockhart: Makes sense -- I think what you have should suffice!

components/dropdown/dropdown-tabs.js
```diff
@@ -25,17 +29,21 @@ class DropdownTabs extends DropdownContentMixin(LitElement) {
 		`;
 	}
 
-	_onMenuResize() {
```
dbatiste: Similar to `d2l-dropdown-menu`, this logic needs to use the provided dimensions.  In addition, it must account for the height of the tabs, and whitespace.  This is needed so that the dropdown can animate/position/size itself.  Otherwise, it would need to wait until the menu animation is complete, or disable the animation, especially for nested menus.

Background - hierarchical-view (and in turn menu) dispatches this event, and it provides the measured dimensions of the content as a hint, which dropdown is intended to use, so that it can position/animate/size itself immediately.

components/tabs/tab-panel-mixin.js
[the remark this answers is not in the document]
dbatiste: I've done some testing for this - so for at least VO, if the `tabpanel` role is not reflected to the attribute, the panel is not announced correctly when focus is moved into the panel.

The other thing I noticed for `d2l-tab-internal` was that axe tests would fail unless the `role="tab"` is reflected to the attribute, because it looks for children (of the element with `role="tablist"`) that have the attribute.

Result may vary depending on browser, screen reader, and perhaps role, but based on what I've found here, I'm tempted to suggest that we should always reflect the role.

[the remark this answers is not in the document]
dlockhart: Gotcha -- makes sense. Thanks for investigating.

components/tabs/tabs.js
```diff
@@ -0,0 +1,840 @@
...
+		} else {
+			this._focusSelected();
+		}
+	}
+
+	async _handlePanelsSlotChange(e) {
+
+		const panels = this._getPanels(e.target);
+		/*
+		// do fancy remove the panels
```
dlockhart, line 500: Was this meant to be left commented out?

components/tabs/tabs.js
```diff
@@ -0,0 +1,840 @@
...
+					this._state = 'anim';
+					requestAnimationFrame(() => {
+						this._state = 'shown';
+					});
+				}
+			}
+		}
+
+		let selectedTabInfo = null;
+		this._tabInfos = [];
```
dlockhart, line 538: Seems like this could use `this._tabInfos = panels.map((panel) => { ... });`?

components/tabs/test/tabs.test.js
```diff
@@ -0,0 +1,46 @@
...
+			await expect(el).to.be.accessible();
+		});
+
+	});
+
+	describe('events', () => {
+
+		it('dispatches d2l-tab-panel-selected', async() => {
+			const el = await fixture(normalFixture);
+			const panel = el.querySelectorAll('d2l-tab-panel')[1];
```
dlockhart, line 30: The pattern we've been using for this kind of thing is to `setTimeout` (borrowed from the open-wc docs):

```
const el = await fixture(normalFixture);
const panel = el.querySelectorAll('d2l-tab-panel')[1];
setTimeout(() => panel.selected = true);
await oneEvent(panel, 'd2l-tab-panel-selected');
```

components/tabs/tabs.js
dbatiste: Just about to push some changes that includes this.  :)

PR #431, opened 2020-03-24
Input-date + Calendar

margaree:
This PR is on top of the accessibility improvements to calendar in  https://github.com/BrightspaceUI/core/pull/428

This puts calendar in a dropdown that is opened when a user clicks, presses enter on, or presses the down arrow on the text input.

Note that when calendar is in this dropdown, the initial focus when calendar is open is actually the focus date and not the left icon button. This is to improve accessibility. When calendar is used on its own as a widget the initial focus remains the left icon button.


components/calendar/calendar.js
```diff
@@ -163,6 +163,7 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 
 	static get properties() {
 		return {
+			dialog: { type: Boolean },
```
margaree: Needed for accessibility, in order to add necessary attributes to wrap the calendar.

components/calendar/calendar.js
```diff
@@ -307,7 +308,7 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 		super.updated(changedProperties);
 
 		changedProperties.forEach((oldVal, prop) => {
-			if (prop === '_shownMonth') {
+			if (prop === '_shownMonth' && this._keyboardTriggeredMonthChange) {
```
margaree: Did some rearranging with `_keyboardTriggeredMonthChange` just so that it wasn't necessary to set it to true before every call to `_focusDateAddFocus`

components/calendar/calendar.js
```diff
@@ -323,6 +324,7 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 		const weekdayHeaders = calendarData.daysOfWeekIndex.map((index) => html`
 			<th
 				abbr="${calendarData.descriptor.calendar.days.long[index]}"
+				aria-label="${calendarData.descriptor.calendar.days.long[index]}"
```
margaree: In NVDA without this for some days of the week it will say the abbreviated version (e.g., `Mon` instead of `Monday`)

components/calendar/test/calendar.visual-diff.js
```diff
@@ -19,12 +19,12 @@ describe('d2l-calendar', () => {
 
 	after(() => browser.close());
 
-	it('has no selected-value when selected-value not set', async function() {
+	it('no selected value', async function() {
```
margaree: Most changes in here were to make test names smaller and make this generally more readable

components/inputs/README.md
```diff
@@ -109,6 +110,7 @@ The `<d2l-input-text>` element is a simple wrapper around the native `<input typ
 
 **Properties:**
 
+- `aria-haspopup` (String):  indicates that the input has a popup menu
```
margaree: This was added because `dropdown` is adding it to `d2l-input-text` but it isn't being brought through to the `input`.

`title` was added in order to add more context/direction for screenreader users (perhaps there was a better way to do that? I saw it used in an example)

components/inputs/input-date.js
```diff
@@ -1,11 +1,24 @@
...
 import './input-text.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
-import { formatDate, parseDate } from '@brightspace-ui/intl/lib/dateTime.js';
+import { formatDate, getDateTimeDescriptor, parseDate } from '@brightspace-ui/intl/lib/dateTime.js';
+import { getToday } from '../calendar/calendar.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeStaticMixin } from '../../mixins/localize-static-mixin.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
+let calendarDataDescriptor;
```
margaree: I'm thinking that some of these functions that overlap between `calendar` and `input-date` should be pulled out into a separate helper

components/inputs/input-date.js
```diff
@@ -57,50 +71,150 @@ class InputDate extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		this._formattedValue = this.value ? formatDate(parseISODate(this.value)) : '';
+		this._dropdown = this.shadowRoot.querySelector('d2l-dropdown-content');
 	}
 
 	render() {
-		const placeholder = this.placeholder || this.localize('chooseDate');
+		const placeholder = this.placeholder || calendarDataDescriptor.formats.dateFormats.short;
```
margaree: I changed the placeholder to be the accepted short date format (this was included as an option in the design) since at the moment the format that will be accepted by our parser is fairly limited.

components/inputs/input-date.js
```diff
@@ -109,21 +223,74 @@ class InputDate extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
+			if (prop === 'value') {
+				if (this.value) this._formattedValue = formatISODateInUserCalDescriptor(this.value);
+				else this._formattedValue = '';
+
+			}
+		});
+	}
+
+	async _handleFocusTrapEnter() {
+		this.shadowRoot.querySelector('d2l-calendar').focus(true);
```
margaree: For the case where a user opens the calendar by clicking (focus should still remain on the `input-text`) and then tabs into the calendar, the initial focus should go to the calendar `focusDate` (which is why `true` is included)

components/inputs/input-date.js
```diff
@@ -57,50 +71,150 @@ class InputDate extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			:host([disabled]) d2l-icon {
 				opacity: 0.5;
 			}
+			d2l-focus-trap {
+				margin-bottom: -0.7rem;
+				margin-top: -0.7rem;
```
dlockhart: Did you want `no-padding` on the dropdown content instead?

components/inputs/input-date.js
```diff
@@ -109,21 +223,74 @@ class InputDate extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldVal, prop) => {
+			if (prop === 'value') {
+				if (this.value) this._formattedValue = formatISODateInUserCalDescriptor(this.value);
+				else this._formattedValue = '';
+
```
dlockhart: Nit: might be more legible as:

    this.m_formattedValue = this.value ? formatISODateInUserCalDescriptor(this.value) : '';

components/calendar/calendar.js
```diff
@@ -397,9 +402,13 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 		`;
 	}
 
-	focus() {
-		const firstButton = this.shadowRoot.querySelector('d2l-button-icon');
-		if (firstButton) firstButton.focus();
+	focus(focusOnDate) {
```
dlockhart: `focus()` is a native method that exists on all elements, so I don't feel great about adding a parameter here. Could the logic be that if it's in "dialog" mode, it does one thing and if not does another?

components/calendar/calendar.js
dlockhart: Could we maybe hide this a bit more? Or could we detect it automatically somehow by seeing if it's inside a dropdown? Just don't want others to see it and set it if we can avoid it.

components/inputs/input-date.js
dlockhart: A couple things:
- This makes me question whether we should support a custom placeholder at all... it seems like the format hint is super important, and if in some places it gets replaced by a custom placeholder, then those users don't get it.
- Along the same lines, the placeholder doesn't get read to screen readers, so they're not going to get the hint. Maybe an offscreen `aria-describedby` that says "Format: YYYY-MM-DD" might help?

components/calendar/calendar.js
margaree: Agreed - I like the idea of automatic detection. I'll look into that.

components/inputs/input-date.js
margaree: Agreed, that does seem problematic. Perhaps we shouldn't include it initially and see if use cases pop up requesting it, then evaluate further.

I have found that the placeholder does get read (at least with NVDA - will confirm voice over). I like the idea of an `aria-describedby` too though in case there is a value there just to make format extra clear.

components/calendar/calendar.js
margaree: Yup that sounds like a much better idea.

components/inputs/input-date.js
dlockhart: Yeah I'd be on board with removing it entirely assuming Jeff is good with it.

margaree: Confirmed with Jeff that this would be okay.

components/inputs/input-date.js
margaree: So I think I will leave the caching related to `dateTimeDescriptor` within core for now, as it seems a bit complicated to move it into intl and be able to keep up to date when language/locale settings changes, and I'd like to do that more separately.

components/calendar/calendar.js
```diff
@@ -363,16 +365,19 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 			return html`<tr>${weekHtml}</tr>`;
 		});
 		const activeDate = `${this._tableInfoId}-${this._focusDate.getFullYear()}-${this._focusDate.getMonth()}-${this._focusDate.getDate()}`;
+		const labelId = `${this._tableInfoId}-heading`;
+		const labelledBy = this.dialog ? labelId : undefined;
 		const heading = `${calendarData.descriptor.calendar.months.long[this._shownMonth]} ${this._shownYear}`;
+		const role = this.dialog ? 'dialog' : undefined;
 		return html`
-			<div class="d2l-calendar">
+			<div aria-labelledby="${ifDefined(labelledBy)}" aria-modal="${ifDefined(this.dialog)}" class="d2l-calendar" role="${ifDefined(role)}">
```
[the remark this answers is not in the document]
dbatiste: Note: I tried using `aria-modal` with the `d2l-dialog` components and I had mixed results.  In fact, I think it might have caused some problems.  I had to exclude it.  Something to watch out for.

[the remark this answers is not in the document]
margaree: Discussed with Dave and decided that it was best to remove `aria-modal` since it isn't really applicable here and could cause problems when calendar is used with VoiceOver

PR #432, opened 2020-03-27
Remove lang duplicates

ppaskaris-d2l:
This PR removes duplicate language tags from showing up in list of possible languages. This frees the clients from needing to do this deduping themselves (either indirectly via `import` or directly like I am here).

e.g. we currently see `["en-us","en","en-us","en","en-us","en"]` for simple English pages. With this, it's just `["en-us","en"]`


ppaskaris-d2l:
Oh gosh, you're quick 😄

PR #433, opened 2020-03-27
Time Input Dropdown Fixes

Scy-D2L:
- Dropdown no longer appears when `input-time` component is disabled
- Fixed label issue for time dropdown. Combobox requires `aria-labelledby` instead of `aria-label`. Label now always exists, but is offscreen when hidden.
  - Tweaked tests related to labels based on implementation changes

### DIFF
--- a/components/inputs/input-time.js
+++ b/components/inputs/input-time.js
@@ -7,7 +7,6 @@ import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { formatTime, parseTime } from '@brightspace-ui/intl/lib/dateTime.js';
 import { bodySmallStyles } from '../typography/styles.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
-import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
 import { inputStyles } from './input-styles.js';
 import { offscreenStyles } from '../offscreen/offscreen-styles.js';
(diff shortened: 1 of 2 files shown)


components/inputs/input-time.js
```diff
@@ -122,47 +122,43 @@ class InputTime extends LitElement {
...
-								value="${formatValue(i)}"
-								?selected=${this._value === formatValue(i)}>
-							</d2l-menu-item-radio>
-						`)}
-					</d2l-menu>
-					<div class="d2l-input-time-timezone d2l-body-small" slot="footer">${this._timezone}</div>
-				</d2l-dropdown-menu>
-			</d2l-dropdown>
+			<label>
+				<span class="${this.label && !this.labelHidden ? 'd2l-input-label' : 'd2l-offscreen'}" id="${this._dropdownId}-label">${this.label}</span>
```
dlockhart: Maybe we want `label` to be required eh? Like throw a `console.warn` in if it's not provided. Because here it'll render an offscreen empty label if it doesn't exist, which is kind of pointless.

components/inputs/test/input-time.test.js
```diff
@@ -49,19 +49,14 @@ describe('d2l-input-time', () => {
...
-		}
-
 		it('should display visible label', async() => {
 			const elem = await fixture(basicFixture);
-			expect(getLabel(elem).innerText).to.equal('label text');
+			expect(elem.shadowRoot.querySelector('.d2l-input-label').innerText).to.equal('label text');
 		});
 
-		it('should put hidden label on "aria-label"', async() => {
+		it('should created offscreen label when label-hidden', async() => {
```
dlockhart: Nit: `should create offscreen label...`

PR #435, opened 2020-03-30
input-date-time component

margaree:
This PR is based on https://github.com/BrightspaceUI/core/pull/431


components/calendar/calendar.js
```diff
@@ -254,8 +254,8 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		this._today = getToday();
-		const date = this.selectedValue ? parseISODate(this.selectedValue) : this._today;
+		this._today = getDateFromDateObj(getToday());
```
margaree: I did some reorganizing of the helper functions for consistency in what they're returning, so this and a few other changes in here are a result of that

components/inputs/README.md
```diff
@@ -429,11 +430,34 @@ The `<d2l-input-date>` component consists of a text input field for typing a dat
 
 * `d2l-input-date-change`: dispatched when a date is selected or typed. `value` reflects the selected value and is in ISO 8601 calendar date format (`YYYY-MM-DD`).
 
+## Date-Time Inputs
+
+The `<d2l-input-date-time>` component consists of a `<d2l-input-date>` and a `<d2l-input-time>` component. The time input only appears once a date is selected. This component displays the `value` if one is specified, and reflects the selected value when one is selected or entered.
```
margaree: TODO: add gif

components/inputs/input-time.js
```diff
@@ -176,9 +152,9 @@ class InputTime extends LitElement {
 	getTime() {
 		const time = parseValue(this.value);
 		return {
-			hour: time.getHours(),
-			minute: time.getMinutes(),
-			second: time.getSeconds()
+			hours: time.getHours(),
```
margaree: modified this for consistency. I have been using plural in other places to be consistent with javascript `Date` which has `getHours` etc.

helpers/dateTime.js
```diff
@@ -1,43 +1,125 @@
 import { convertUTCToLocalDateTime, getDateTimeDescriptor } from '@brightspace-ui/intl/lib/dateTime.js';
 
+// val is an object containing year, month, date
```
margaree: I updated this file to have more consistency with naming and what functions return. `parse*` functions now all return objects, as does `getToday`. 2 functions return `Date`s and that is specified in their names.

components/inputs/input-date-time.js
```diff
@@ -0,0 +1,134 @@
+import './input-date.js';
+import './input-fieldset.js';
+import './input-time.js';
+import { convertLocalToUTCDateTime, convertUTCToLocalDateTime } from '@brightspace-ui/intl/lib/dateTime.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { formatDateInISO, formatTimeInISO, parseISODate, parseISODateTime, parseISOTime } from '../../helpers/dateTime.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+export function formatDateTimeInISO(val) {
```
margaree: I'm torn about if this should throw and if it should be here vs. in the `helpers/dateTime.js` file, since it's only used here.

components/inputs/input-date-time.js
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return `${formatDateInISO({year: val.year, month: val.month, date: val.date})}T${formatTimeInISO({hours: val.hours, minutes: val.minutes, seconds: val.seconds})}.000Z`;
+}
+
+class InputDateTime extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			disabled: { type: Boolean },
+			label: { type: String },
```
dlockhart: Since there's no way to properly label things from outside the component, I think we should `console.warn` if a label isn't provided -- just like time input does. Same in `date-input`.

components/inputs/input-date-time.js
```diff
@@ -0,0 +1,134 @@
...
+		this.required = false;
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		if (this.value) {
+			try {
+				const parsed = parseISODateTime(this.value);
+				const localDateTime = convertUTCToLocalDateTime(parsed);
```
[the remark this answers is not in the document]
margaree: Yup it was (unless we had changed that in a later chat and I'm forgetting which is totally possible).

components/inputs/README.md
```diff
@@ -429,11 +430,34 @@ The `<d2l-input-date>` component consists of a text input field for typing a dat
...
+  label="Start Date"
+  value="2020-11-20T12:00:00.000Z">
+</d2l-input-date-time>
+```
+
+**Properties:**
+
+- `label` (String, required): accessible label for the input
+- `disabled` (Boolean): disables the input
+- `required` (Boolean): indicates that a value is required
```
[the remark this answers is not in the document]
margaree: Jeff is creating a separate story for `required` support as there will likely be a bit of functionality to go along with the use of it.

components/inputs/input-date-time.js
```diff
@@ -0,0 +1,134 @@
...
+		return css`
+			:host {
+				display: inline-block;
+				white-space: nowrap;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			d2l-input-date {
+				max-width: 10rem;
```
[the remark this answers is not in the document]
margaree: The both end up adding a lot of space to their right and are appearing wider than in the design. I can move these changes into the components though - that might make more sense

components/inputs/input-date-time.js
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	render() {
+		const timeHidden = !this._parsedDate;
+		return html`
+			<d2l-input-fieldset label="${ifDefined(this.label)}" ?required="${this.required}">
+				<d2l-input-date
+					@d2l-input-date-change="${this._handleDateChange}"
+					?disabled="${this.disabled}"
+					label="${this.label}"
```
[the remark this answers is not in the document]
margaree: Yup that's a good idea

components/inputs/input-date-time.js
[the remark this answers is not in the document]
dlockhart: Yeah I think if they can be the correct size on their own that's ideal.

components/inputs/input-date.js
```diff
@@ -165,7 +170,7 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 					@keydown="${this._handleKeydown}"
 					label="${ifDefined(this.label)}"
 					?label-hidden="${this.labelHidden}"
-					placeholder="${this._dateTimeDescriptor.formats.dateFormats.short}"
+					placeholder="${(this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase()}"
```
margaree: Jeff requested the placeholder be uppercase (e.g., instead of "M/d/yyyy" have it be "M/D/YYYY") for a more consistent look. I'll confirm that this doesn't result in screenreader yelling.

dlockhart: Cool. The case in the actual date format strings actually means something, but in this context it should be OK to uppercase it.

margaree: yeah we were thinking/hoping users wouldn't necessarily know/care about that.

margaree: Checked NVDA with firefox and Voice over with Safari and did not experience any yelling with all caps.

dlockhart: Other option would be to use `text-transform: uppercase` in the CSS. If this work though, 🤷‍♂

PR #440, opened 2020-03-31
Tooltip README + Bug Fixes

AllanKerr:
# Changes
1. Add `d2l-tooltip` README
2. Fix tooltip when positioned right and `rtl` is enabled and add corresponding visual diff tests.
3. Rename `_layoutTooltip` to `updatePosition` for backwards compatibility where its being called manually by users of the tooltip.

Once this merges I'll be doing a minor version bump.


components/tooltip/README.md
```diff
@@ -0,0 +1,66 @@
...
+  import '@brightspace-ui/core/components/tooltip/tooltip.js';
+</script>
+
+<input type="text" placeholder="Hover for Error" id="tooltip-error">
+<d2l-tooltip for="tooltip-error" state="error" align="start" offset="10">
+	Your error message will display here
+</d2l-tooltip>
+```
+
+### Boundaries
```
dlockhart: Can you bump this down under an "Advanced Usage" heading or something? Same for any properties that we want to discourage the use of -- hide them.

components/tooltip/README.md
```diff
@@ -0,0 +1,66 @@
...
+	</d2l-tooltip>
+</div>
+```
+
+
+**Properties:**
+* `for` (required, String): provide the `id` of the tooltip's target element. If this attribute is not provided, the tooltip's parent element will be used as its target.
+* `align` (String): optionally align the tooltip with either the start or end of its target. If not set, the dropdown will attempt be centred. Valid values are: `start` and `end`.
+* `boundary` (Object) - optionally provide boundaries to where the tooltip will appear. Valid properties are `"top"`, `"bottom"`, `"left"` and `"right"`. The boundary is relative to the tooltip's [offset parent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent).
+* **DEPRECATED** `custom-target` (Object) - optionally provide a rectangle that the tooltip should position itself around. The object must provide values for `"left"`, `"top"`, `"width"`, and `"height"`. The custom target is relative to the tooltip's [offset parent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent). If a custom target is provided the user is responsible for showing and hiding the tooltip since it no longer has a target element that can be focused or hovered.
```
dlockhart: Maybe we don't even mention it?

components/tooltip/README.md
```diff
@@ -0,0 +1,66 @@
...
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/tooltip/tooltip.js';
+</script>
+
+<input type="text" placeholder="Hover for Error" id="tooltip-error">
+<d2l-tooltip for="tooltip-error" state="error" align="start" offset="10">
+	Your error message will display here
+</d2l-tooltip>
+```
```
dlockhart: Can you add a note in here about how the the `for` and corresponding `id` element need to be in the same shadow root?

components/tooltip/tooltip.js
```diff
@@ -357,6 +363,92 @@ class Tooltip extends RtlMixin(LitElement) {
 		});
 	}
 
+	async updatePosition() {
```
dlockhart: Is this now meant to be more of a "public" method that consumers would call?

components/tooltip/README.md
```diff
@@ -0,0 +1,66 @@
...
+* `offset`: (Number, default: `16.5`): adjust the size of the gap between the tooltip and its target.
+* `position` (String): optionally force the tooltip to open in a certain direction. Valid values are: `top`, `bottom`, `left` and `right`. If no position is provided, the tooltip will open in the first position that has enough space for it to open in the order: top, bottom, right, left.
+* `showing` (Boolean, default: `false`): a **readonly** only property that can be used to determine if the tooltip is currently visible
+* `state` (String, default: `info`): the style of the tooltip based on the type of information it displays. Valid values are: `info` and `error`.
+
+**Events:**
+* `d2l-tooltip-show`: dispatched when the tooltip is opened
+* `d2l-tooltip-hide`: dispatched when the tooltip is closed
+
+**CSS Variables:**
```
dlockhart: I'm good with not even mentioning things we don't want people to use.

components/tooltip/README.md
AllanKerr: The main reason I thought it might be worth mentioning is that way people don't discover it themselves and assume its good to use. It was never documented in the original `d2l-tooltip` but it still go used. That being said, I can definitely leave it out. I don't feel strongly either way.

components/tooltip/tooltip.js
AllanKerr: Its used publicly in one place that I'm hoping to go in and remove so it can go back to being private.

components/tooltip/README.md
dlockhart: Gotcha. Is this one of the things you'd like to clean up and remove afterwards? If so, then we could likely safely exclude it. Otherwise it seems OK to leave the docs but maybe with a harsher "don't use this" vibe.

AllanKerr: I think I should be able to remove it entirely once I go in and clean up the places that use it. I'll exclude it and re-evaluate if something unexpected happens and I can't remove it entirely.

components/tooltip/README.md
AllanKerr: The reasoning here was also the same reason for documenting `custom-target`. I'm hoping, possibly naively, if people see them they'll understand why we don't want people to use them and that their is a proper route to solve their problem by opening an issue and adding new `state` values.

AllanKerr: I'm going to remove the documentation for these and move the quote to the `state` property.

> If you find yourself needing a style that isn't supported by the `state` attribute please create a Github issue.

AllanKerr:
Added one more changed based on feedback from Jeff in the design meeting we just had. We both agreed it would make more sense to have the positioning order be `bottom`, `top`, `right`, `left` so `bottom` comes before `top`.

PR #441, opened 2020-03-31
S akulick/input time intervals

Scy-D2L:
Input-time now supports custom intervals with the `time-intervals` property (5, 10, 15, 20, 30, 60), defaulting to 30. The values are text-based to match the old component.
Using `enforce-time-intervals` will require values matching the intervals, rounding up when typed input doesn't conform. Using `enforce-time-intervals` will also remove the end-of-day option (11:59 PM)

Includes unit tests and slight performance improvements.


components/inputs/input-time.js
```diff
@@ -80,24 +112,33 @@ class InputTime extends LitElement {
...
+			}
+		}
 		this._value = formatValue(time);
 		this._formattedValue = formatTime(time);
 		this.requestUpdate('value', oldValue);
 	}
 
 	render() {
-		getIntervals();
+		getIntervals(this.timeInterval);
```
dlockhart, line 141: If it's not going to actually return anything, better to call this `initIntervals()` or something.

components/inputs/test/input-time.test.js
```diff
@@ -127,7 +132,32 @@ describe('d2l-input-time', () => {
...
+			expect(getNumberOfIntervals(elem)).to.equal(24);
+		});
 
+		it('should round-up to next interval when intervals are enforced', async() => {
+			const elem = await fixture(hourLongIntervalsEnforced);
+			getInput(elem).value = '2:01AM';
+			dispatchEvent(elem, 'change', false);
+			await oneEvent(elem, 'change');
+			expect(elem.value).to.equal('3:00:00');
+		});
```
dlockhart, line 160: Suggestion for another similar test:

```
it('should round-up to next interval when intervals are enforced and value is set', () => {
    const elem = await fixture(hourLongIntervalsEnforced);
    elem.value = '2:01AM';
    expect(elem.value).to.equal('3:00:00');
});
```

PR #443, opened 2020-04-02
Calendar transitions & animations

margaree:
Transitions:
- fade in when moving between months
- box-shadow when navigating between dates with arrow keys 

Animation:
- Months slide in from left/right


components/calendar/calendar.js
```diff
@@ -180,48 +180,109 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
+				box-shadow: 0 0 0 2px var(--d2l-color-celestine-plus-2), 0 0 0 4px var(--d2l-color-celestine);
 			}
 
 			div.d2l-calendar-date-today,
 			div.d2l-calendar-date-selected {
 				font-size: 1rem;
 				font-weight: 700;
 			}
+
+			@keyframes d2l-calendar-next-animation {
```
dbatiste, line 269: Wondering if this could have been done with a transition?

components/calendar/calendar.js
```diff
@@ -275,6 +336,8 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 			getCalendarData(true);
 			this.requestUpdate();
 		});
+
+		this.classList.add('d2l-calendar-initial-month');
```
dbatiste, line 340: Note: we should apply class changes in `render`.

components/calendar/calendar.js
```diff
@@ -572,11 +636,23 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 	_updateShownMonthDecrease() {
 		if (this._shownMonth === 0) this._shownYear--;
 		this._shownMonth = getPrevMonth(this._shownMonth);
+		this.classList.remove('d2l-calendar-initial-month');
```
dbatiste, line 639: What do you think of adding/removing these classes from an internal `div` rather than the host element?  It would enable them to be applied in `render` where we usually want DOM mutations to be done.  Alternatively, when I've needed to expose a "private" bit of state on the host, I've defined a private property that gets reflected, which makes it clear that it's private, where as `class` on the host is very much a "public" property.

dlockhart: I think putting it on an internal div is the "cleanest" as it hides everything, then yeah the private reflected property would be next-best.

margaree: Yup that does sound cleaner - I'll give that a try.

PR #446, opened 2020-04-02
Calendar transition improvements

margaree:


components/calendar/calendar.js
```diff
@@ -134,6 +134,9 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
 			summary: { type: String },
 			_dialog: { type: Boolean },
 			_focusDate: { type: Object },
+			_initialMonth: { type: Boolean },
+			_nextMonthNav: { type: Boolean },
+			_prevMonthNav: { type: Boolean },
```
dlockhart, line 139: Only one of these can the `true` at once I think ya? So could it just be a single property that gets set to either `initial`, `next` or `prev`?

components/calendar/calendar.js
```diff
@@ -199,32 +208,41 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
+			.d2l-calendar-prev .d2l-calendar-date div {
+				left: -10px;
+			}
+
+			.d2l-calendar-next .d2l-calendar-title .d2l-heading-4,
+			.d2l-calendar-prev .d2l-calendar-title .d2l-heading-4,
+			.d2l-calendar-initial-month .d2l-calendar-title .d2l-heading-4,
+			.d2l-calendar-next .d2l-calendar-date div,
+			.d2l-calendar-prev .d2l-calendar-date div,
+			.d2l-calendar-initial-month .d2l-calendar-date div {
```
dlockhart, line 231: Could use like an additional class like `d2l-calendar-animating` or something that's applied whenever any of them are happening -- simplify the CSS.

components/calendar/calendar.js
```diff
@@ -199,32 +208,41 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
-			:host(.d2l-calendar-next) .d2l-calendar-title .d2l-heading-4,
-			:host(.d2l-calendar-prev) .d2l-calendar-title .d2l-heading-4,
-			:host(.d2l-calendar-initial-month) .d2l-calendar-title .d2l-heading-4,
-			:host(.d2l-calendar-next) .d2l-calendar-date,
-			:host(.d2l-calendar-prev) .d2l-calendar-date,
-			:host(.d2l-calendar-initial-month) .d2l-calendar-date {
+			.d2l-calendar-next .d2l-calendar-date div {
+				left: 10px;
+			}
+
```
dlockhart, line 221: Any RTL implications for this?

components/calendar/calendar.js
```diff
@@ -636,22 +641,22 @@ class Calendar extends LocalizeStaticMixin(LitElement) {
...
 	_updateShownMonthIncrease() {
 		if (this._shownMonth === 11) this._shownYear++;
 		this._shownMonth = getNextMonth(this._shownMonth);
-		this.classList.remove('d2l-calendar-initial-month');
-		this.classList.remove('d2l-calendar-prev');
-		this.classList.remove('d2l-calendar-next');
+		this._initialMonth = false;
+		this._nextMonthNav = false;
+		this._prevMonthNav = false;
 		setTimeout(() => {
```
dbatiste, line 658: What happens in FF without the timeout?  Is the timeout still needed now that the DOM mutation has been moved to `render`?

margaree: So with this change to use the classMap etc., the `setTimeout` seems to be needed for all browsers, and then firefox seems to need it to actually have a value or else basically every other month change transition just doesn't happen.

I have tried removing the setTimeout and using other things like `requestUpdate` + `await this.updateComplete` before modifying the class, moving code around and so far I haven't had luck avoiding the `setTimeout` so definitely open to ideas to avoid this.

PR #449, opened 2020-04-06
Scroll into view fix

Scy-D2L:
Partial fix for [DE38481](https://rally1.rallydev.com/#/detail/defect/379578838464?fdp=true): _Time Input > Selected date scroll into view intermittent with Firefox and broken in Legacy Edge_ to fix Legacy Edge issues.

- Legacy Edge no longer throws error with dropdown radio-menus.
- Legacy Edge now scrolls selected item into view when dropdown is opened.
- Includes fix for scroll 'popping' in dropdown radio-menus in Chrome.
- Firefox issue in DE38481 seems to be a separate race condition to be investigated/merged in another PR


components/dropdown/dropdown-menu.js
```diff
@@ -91,8 +91,15 @@ class DropdownMenu extends DropdownContentMixin(LitElement) {
 	}
 
 	__getMenuElement() {
-		return this.shadowRoot.querySelector('.d2l-dropdown-content-container > slot')
-			.assignedNodes()
+		let slot = this.shadowRoot.querySelector('.d2l-dropdown-content-container > slot');
```
dlockhart, line 94: I think I've seen this before actually... what if you try putting a CSS class name on the slot and then change this selector to be `.d2l-dropdown-content-slot`, does it still need the Edge hack then?

components/inputs/test/input-date.test.js
```diff
@@ -52,7 +52,7 @@ describe('d2l-input-date', () => {
 			const elem = await fixture(basicFixture);
 			const dropdown = getChildElem(elem, 'd2l-dropdown');
 			dropdown.toggleOpen();
-			await aTimeout(1);
+			await oneEvent(dropdown, 'd2l-dropdown-open');
```
dlockhart, line 55: Yay less timeouts! I thought I remembered a timeout in some of the dropdown or menu auto-focusing code that was listening for `d2l-dropdown-open` event as well and then doing its own `setTimeout`.  I think that should be able to be removed?

components/dropdown/dropdown-menu.js
dbatiste: Yeah there is some weirdness here.  As Dave mentioned, I think we should be able to get this working without two different selectors.

components/inputs/test/input-date.test.js
Scy-D2L: Do you have any more context on that? I'm seeing a lot of places where we setTimeout without a clear reason, so I'm not entirely sure what I should be looking for - though you're probably right. :)

dlockhart: [This one](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-menu.js#L64) -- it was listening for that `d2l-dropdown-open` event and doing its own timeout (as is your scrollIntoView code in that same method). Now that the event is fired from inside a timeout, hopefully one or both of them aren't needed.

Scy-D2L:
Regenerating goldens - this is expected because the scrolling behavior no longer centers the selected item in the dropdown.

components/inputs/test/input-date.test.js
Scy-D2L: The one you referenced can go, it works fine without it, but the timeout for scrolling things into view is still needed. Without it, the post-resize-scroll back to the top consistently happens after our desired scroll.

PR #451, opened 2020-04-06
Fix for selecting content section of dropdown in edge

Scy-D2L:
Fix for dropdown issue in Edge that threw errors and had the potential to cause resizing bugs.


components/dropdown/dropdown-menu.js
```diff
@@ -91,8 +91,15 @@ class DropdownMenu extends DropdownContentMixin(LitElement) {
 	}
 
 	__getMenuElement() {
-		return this.shadowRoot.querySelector('.d2l-dropdown-content-container > slot')
-			.assignedNodes()
+		let slot = this.shadowRoot.querySelector('.d2l-dropdown-content-container > slot');
```
dlockhart, line 94: Same comment here as in the other pull request. I think this can be solved by simplifying the selector to not reference the `slot` element, since it doesn't exist on browsers who don't support shadow DOM.

PR #453, opened 2020-04-06
Dbatiste/dialog render fix

dbatiste:
Interestingly, when the dialog is opened via the attribute, changes to internal properties will trigger a rerender (as desired), however when the dialog is opened via property assignment, subsequent internal properties changes do not.  Bumping these updated to the next frame results in render calls as needed to update the width, followed by height measurement, etc.  This also fixes the issue where the `showing` state was not causing a rerender, which was in turn causing the backdrop not to be rendered as expected.


dbatiste:
Digging into this some more.  The reason why the backdrop is not being rendered correctly on some subsequent dialog openings is because the animation is not being completed properly on previous closings due to the backdrop parent (the dialog) not being visible.  I will create a new PR that hopefully is a little cleaner commit-wise.

PR #455, opened 2020-04-07
US114945 Constrain width of Primary-Secondary template

wongvincent:
This implements the option to specify a maximum `widthType` for the template, so that the template does not always take 100% of the viewport width.

The two `widthTypes` are `fullscreen` (default) and `normal` (1230px).

**Example**
image


templates/primary-secondary/primary-secondary.js
```diff
@@ -52,9 +58,21 @@ class TemplatePrimarySecondary extends LitElement {
...
+		const navigationStyles = {};
+		if (this.width) {
+			containerStyles.maxWidth = `${this.width}px`;
+			navigationStyles['--d2l-navigation-immersive-content-max-width'] = `${this.width}px`;
+		}
+
 		return html`
-			<div class="container">
-				<header><slot name="header"></slot></header>
+			<div class="container" style="${styleMap(containerStyles)}">
```
dbatiste, line 69: It looks like this would constrain everything, including the header and its contents.  In the above code, I _think_ it would have to be `containerStyles['max-width'] = ...`, unless the styleMap directive does some automagical conversion.

More importantly, I don't think the page template should have nav baked into it.  Rather, I think the max-width should just be applied to the non-header elements, but the consumer would have to apply similar outside of this component for `--d2l-navigation-immersive-content-max-width`. Maybe other solutions, but I don't think core should be sucking in nav.

wongvincent: The `lit-html` docs show it to be camel-case.

Yep, I've removed the navigation and will continue to pass it in the `header` `slot`.

wongvincent:
This story's PRs for constraining width of FACE template:
Primary-secondary template - https://github.com/BrightspaceUI/core/pull/455
Immersive Nav - https://github.com/BrightspaceUI/navigation/pull/95
Updating Activities LMS renderer to pass appropriate `widthType` through to Immersive Nav and Activities component - https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/10439/overview
Updating FACE Assignments to pass `widthType` to template - https://github.com/BrightspaceHypermediaComponents/activities/pull/828

wongvincent:
Changed the width of the viewport for the test so that I could test the constrained `normal` (1230px) page width, so that's why the other primary-secondary visual-diff tests also failed.

templates/primary-secondary/primary-secondary.js
```diff
@@ -26,6 +28,9 @@ class TemplatePrimarySecondary extends LitElement {
 				"primary divider secondary"
 				"footer footer footer";
 				height: 100vh;
+				margin-left: auto;
+				margin-right: auto;
+				min-width: min-content;
 			}
```
dlockhart, line 34: You should be able to use a selector just like with navigation instead of doing custom stuff in `render()`:

```css
:host([width-type="normal"]) {
    max-width: 1230px;
}
```

templates/README.md
```diff
@@ -43,10 +43,12 @@ If no nodes are assigned to the `footer` slot, the footer is hidden.
 
 **Note:** this template automatically includes `<header>`, `<main>`, `<aside>` and `<footer>` elements, so there's no need to include them inside the various slots.
 
+**Properties:**
+- `widthType` (string, default: `fullscreen`): The width that the template content will be constrained to. Accepts `fullscreen` or `normal`. Normal constrains width to `1230px`.
```
dlockhart, line 47: If the default is `fullscreen`, can you set the property to that in the component's constructor? Otherwise in JavaScript it'll be `undefined`.

templates/primary-secondary/primary-secondary.js
```diff
@@ -52,8 +57,17 @@ class TemplatePrimarySecondary extends LitElement {
 	}
```
dlockhart, line 58: I'm wondering how the width-type should affect the footer? What does design say about that? Like could we constrain the buttons and the box-shadow somehow automatically?

wongvincent: Spoke with design again, after a bit more discussion, they would like the footer to be the full-width, but the content inside the footer should be constrained to 1230px. So I'll be updating the current PR to do that!

dlockhart: Awesome, that's what I was hoping for 👍

templates/primary-secondary/primary-secondary.js
```diff
@@ -17,19 +18,33 @@ class TemplatePrimarySecondary extends LitElement {
...
-				"footer footer footer";
+				"header"
+				"content"
+				"footer";
 				height: 100vh;
 			}
 			header {
 				grid-area: header;
 			}
+			.content-container {
```
dlockhart, line 39: Can you rename this CSS class to `d2l-template-primary-secondary-content-container`? I know it's a mouthful, but we like to scope our CSS classes for browsers who don't yet support shadow DOM -- otherwise someone else could define styles for a `content-container` CSS class and they'd leak in (or vice versa).

Same for `footer-container`.

PR #457, opened 2020-04-08
Dbatiste/tabs no panels fix

dbatiste:
During some integration testing with My Courses I discovered that the tabs component may throw an error if the panel slot change even is triggered without any panels in it... this happens initially for My Courses.  This PR improves fixes that issue and improves the reliability of adding/removing tabs.

### DIFF
--- a/components/tabs/tabs.js
+++ b/components/tabs/tabs.js
@@ -574,37 +574,36 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 		}
 		this._tabInfos = newTabInfos;
 
+		if (this._tabInfos.length > 0 && !selectedTabInfo) {
+			this._tabInfos[0].selected = true;
+			selectedTabInfo = this._tabInfos[0];
+		}
+
+		await this.updateComplete;
+
+		const animPromises = [];
 		if (!this._initialized && this._tabInfos.length > 0) {
 
 			this._initialized = true;
-			if (!selectedTabInfo) {
-				this._tabInfos[0].selected = true;
-				selectedTabInfo = this._tabInfos[0];
-			}
-
-			await this.updateComplete;
-
 			await this._updateTabsContainerWidth(selectedTabInfo);
-			this._updateMeasures();
-			await this._updateScrollPosition(selectedTabInfo);
 
 		} else {
 
-			await this.updateComplete;
+			if (this._tabInfos.length > 1) {
+				this._tabInfos.forEach((info) => {
+					if (info.state === 'adding') animPromises.push(this._animateTabAddition(info));
+					else if (info.state === 'removing') animPromises.push(this._animateTabRemoval(info));
+				});
+			}
 
-			const animPromises = [];
-			this._tabInfos.forEach((info) => {
-				if (info.state === 'adding') animPromises.push(this._animateTabAddition(info));
-				else if (info.state === 'removing') animPromises.push(this._animateTabRemoval(info));
-			});
-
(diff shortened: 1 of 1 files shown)


components/tabs/tabs.js
```diff
@@ -541,6 +541,7 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 		const panels = this._getPanels(e.target);
 
 		if (this._initialized) this._updateTabListVisibility(panels);
+		else if (panels.length === 0) return;
```
dbatiste, line 544: This fixes the case where `d2l-tabs` is initially rendered without panels.

components/tabs/tabs.js
```diff
@@ -573,38 +574,37 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 		}
 		this._tabInfos = newTabInfos;
 
-		if (!this._initialized && this._tabInfos.length > 0) {
+		if (this._tabInfos.length > 0 && !selectedTabInfo) {
```
dbatiste, line 577: Previously we only defaulted to select the first tab (i.e. when no other panel is selected) on first render.  This change moves that logic out, so that it will select the first tab in the even that the selected tab is removed and the consumer has not previously selected a different tab.

components/tabs/tabs.js
```diff
@@ -573,38 +574,37 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
...
-			this._tabInfos.forEach((info) => {
-				if (info.state === 'adding') animPromises.push(this._animateTabAddition(info));
-				else if (info.state === 'removing') animPromises.push(this._animateTabRemoval(info));
-			});
-
-			Promise.all(animPromises).then(() => {
-				this._updateMeasures();
-				return this._updateScrollPosition(selectedTabInfo);
-			});
+			if (this._tabInfos.length > 1) {
```
dbatiste, line 592: The tablist is only visible when there's more than one tab, so this conditional ensure that we only animate the addition/removal of the tab when the tablist is visible... otherwise the animation promise will never resolve.

components/tabs/tabs.js
```diff
@@ -573,38 +574,37 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
...
+					if (info.state === 'adding') animPromises.push(this._animateTabAddition(info));
+					else if (info.state === 'removing') animPromises.push(this._animateTabRemoval(info));
+				});
+			}
 
+			// required for animation
 			this._updateMeasures();
 		}
 
+		Promise.all(animPromises).then(() => {
```
dbatiste, line 603: This was just moved out since it's common to both conditional branches.

components/tabs/tabs.js
```diff
@@ -722,7 +722,9 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 		this._tabInfos.forEach((tabInfo) => {
 			if (tabInfo.selected && tabInfo.id !== selectedTab.controlsPanel) {
 				tabInfo.selected = false;
-				this._getPanel(tabInfo.id).selected = false;
+				const panel = this._getPanel(tabInfo.id);
+				// panel may not exist if it's being removed
+				if (panel) panel.selected = false;
```
dbatiste, line 727: When a panel is removed, its corresponding tab removal is animated, and so it's tabinfo still exists in `this._tabinfos`... so when we set the select status of panels we cannot assume the panel exists.

PR #459, opened 2020-04-08
input-date: reset calendar on close

margaree:
The calendar should be reset to the month containing selected value or today on each opening, so reset it each time the user closes it (I tried doing this on open but focus date complicated things a bit - it's doable but this code is simpler)


components/inputs/test/input-helper.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
 				const content = dropdown.querySelector('[dropdown-content]');
 				content.scrollTo(0);
 				if (content.opened) {
-					content.addEventListener('d2l-dropdown-close', () => resolve(), { once: true });
+					content.addEventListener('d2l-dropdown-close', () => { setTimeout(resolve, 10); }, { once: true });
```
margaree, line 26: I was having problems with running `reset` followed by `open` and so adding this "fixed" it. I have tried different combinations of `updateComplete` etc. and haven't had success there, but very open to suggestions on alternatives to this.

dlockhart: This definitely seems fragile. I wonder if [the work Scy is doing](https://github.com/BrightspaceUI/core/pull/449) to delay firing the `d2l-dropdown-close` event until it's actually closed will solve this?

margaree: That does sound like something that would potentially help here

PR #462, opened 2020-04-08
Centre the tier1:arrow-collapse-small icon

ianlaidlaw:
The problem:
`arrow-collapse-small` and `arrow-expand-small` are clearly meant to be used in conjunction with each other, much like their regular counterparts are used on the accordion component. However, they are not positioned in the same area of their bounding box making transitioning between them awkward.

expand

image

image

Talking with some designers about these icons, apparently `arrow-collapse` and `arrow-expand` (the big ones) should not even exist and the small ones should be named without the `-small` suffix, but having that discussion is a separate issue.

This pr will simply centre the collapse arrow like it's corresponding `expand` icon is.


dlockhart:
Thanks. As a FRA, it could be pinned on an older version of things so who knows when it would update, but best double-check that it's not already compensating for the misalignment!

PR #463, opened 2020-04-08
d2l-list-item: find d2l-list as ancestor

omsmith:
TODO:
- [ ] Tests?

```js
render() {
  return html`<d2l-list>${this.renderItems()}</d2l-list>`;
}

renderItems() {
  return this.items.map(item => this.renderItem());
}

renderItem() {
  return html`<d2l-list-item></d2l-list-item>`
}
```

In this case `parentNode` of `d2l-list-item` is a documentFragment (which in addition to not working properly, throws because fragments don't have `getAttribute`)

### DIFF
--- a/components/list/list-item.js
+++ b/components/list/list-item.js
@@ -2,6 +2,7 @@ import '../colors/colors.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { checkboxStyles } from '../inputs/input-checkbox-styles.js';
 import { classMap} from 'lit-html/directives/class-map.js';
+import { findComposedAncestor } from '../../helpers/dom.js';
 import { getFirstFocusableDescendant } from '../../helpers/focus.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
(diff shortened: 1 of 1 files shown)


dbatiste:
Was this an issue when the items were in a different DOM scope than the list?

omsmith:
Ran into this while changing a FRA to use `d2l-list-item`'s `selectable`. Usage here:

https://github.com/Brightspace/publish-fra/blob/06d9eb7f6b9a60a270fae08738127d4a2794f1a4/src/components/select-course-list/select-course-list.js#L123-L159

Their actual list-item component before was:

```js
render() {
  return html`
    <div class="row"><d2l-list-item></d2l-list-item></div>
  `;
}
```

Which didn't blow up because `parentNode` was the `div`, but that also meant the seperator lookup wouldn't have worked.

dbatiste:
I think currently there is an assumption (and perhaps it does not suffice) that the list items are direct children.  This list component also has some code that gets the items (by role="listitem") so that it can provide selection info.

omsmith:
Yeah I think that's reasonable. The documentFragment though is an issue

omsmith:
Just added a `null` check as well. Not that its an expected use case or anything, but previously if a list-item _wasn't_ in a `d2l-list` but did have any parent node, it wouldn't throw, now it would.

m6jones:
So, the assumption is that list items are direct children. If people are wrapping them we should find out why and fix the need to do that.

PR #464, opened 2020-04-08
Add Oslo localize helper

ppaskaris-d2l:
This PR adds the Oslo localize helper. Web components will use this to fetch their language files, and in turn, will get free access to language overrides from the LMS.

## Modern mode

When `CacheStorage` is available, we run in modern mode, which means that we cache requests client-side beyond the page lifetime, we batch and pool requests for resources, and we do prompt cache invalidation when overrides are changed in the LMS.

In order to support the dynamic import methods, which we cannot intercept, and to avoid having to intercept all fetch requests, we instead create synthetic `Response` objects, then derive `Blob`s from them and `createObjectURL`s for the components to fetch. Boom, intercepted!

We operate in a two-tier cache:

1. `window` cache (the `blobs` variable): we cache the `objectURL` of each `Blob` we produce for the page
2. `CacheStorage` cache (the `cache` variable): we cache the synthetic `Response`s of each resource we fetch for all pages

For invalidation, each page load is rendered with a version marker (used as the `ETag` for the resources). When a cache entry does not match our most recent version marker, we serve it stale, and then re-validate it asynchronously.

Also, in order to minimize the number of requests we send, we queue requests with debounce and fetch them at "component mount barriers." These are the blocking points where components need their terms in order to render their children. With hypermedia and other stuff, a lot of stuff can cause rendering delays, and we want to strike a balance between spamming the LMS with requests, and waiting forever to batch them all up. We expect that the language will be requested once per user and then cached, so paying this cost on first load seems ok. We serve from cache (even stale) immediately to minimize waits on all subsequent requests.

## Legacy Mode

When CacheStorage isn't available, we run in legacy mode for Internet Explorer and bad mobile browsers. We also do this for HTTP connections that aren't to `localhost` due to the browser's security model. In legacy mode, batching is disabled, and we just forward requests for overrides to the LMS, where it pretends to be a CDN, and manages caching using HTTP cache headers.


ppaskaris-d2l:
Oh gosh that's a lot more mandatory reviewers than I expected 😅

helpers/getLocalizeResources.js
```diff
@@ -0,0 +1,355 @@
...
+		if (cachePromise === undefined) {
+			cachePromise = caches.open(CacheName);
+		}
+		cache = await cachePromise;
+	}
+
+	const cacheKey = new Request(formatCacheKey(resource));
+	const cacheValue = await cache.match(cacheKey);
+	if (cacheValue === undefined) {
+		console.log(`[Oslo] cache miss: ${resource}`);
```
dlockhart, line 157: Wonder if it's worth implementing some kind of debug mode so we can enable/disable these console logs at our leisure?

helpers/getLocalizeResources.js
```diff
@@ -0,0 +1,355 @@
...
+	} else {
+
+		for (const request of requests) {
+
+			request.reject(BatchFailedReason);
+		}
+	}
+
+	if (queue.length > 0) {
+		setTimeout(flushQueue, 0);
```
dlockhart, line 109: In this case, more requests came in while we were processing the queue... so why don't we let the normal batching process handle them? i.e. why process them immediately? I guess it's because `debounceQueue` bails if the state isn't Idle...

helpers/getLocalizeResources.js
```diff
@@ -0,0 +1,355 @@
...
+	cache = undefined;
+	cachePromise = undefined;
+}
+
+export async function getLocalizeResources(
+	possibleLanguages,
+	filterFunc,
+	formatFunc,
+	fetchFunc
+) {
```
dlockhart, line 320: How do you foresee this helper being used by a component -- could you give some sample code of what it might look like?

helpers/getLocalizeResources.js
```diff
@@ -0,0 +1,355 @@
...
+
+	for (const language of possibleLanguages) {
+
+		if (filterFunc(language)) {
+
+			if (supportedLanguage === undefined) {
+				supportedLanguage = language;
+			}
+
+			const translations = fetchFunc(formatFunc(language));
```
dlockhart, line 339: I wasn't expecting a bunch of fetches to happen... only the "best" language match. Why does it need to fetch every language?

helpers/getLocalizeResources.js
ppaskaris-d2l: Sure: https://github.com/BrightspaceHypermediaComponents/activities/compare/ppaskaris/oslo-get-overrides

helpers/getLocalizeResources.js
ppaskaris-d2l: This is actually what the model component was doing already. Presumably it does this so that, during development, fallback terms are incorporated into the "best" language match. Given that this is all permanently cached after the first hit, and that it's all fetched in parallel over HTTP/2, I didn't think it prudent to change the behaviour.

helpers/getLocalizeResources.js
ppaskaris-d2l: Yeah, I didn't want to add an artificial delay if queuing was already happening while the request was in flight. This might be tweakable after if the model component (FACE) doesn't correspond well to other usages.

helpers/getLocalizeResources.js
ppaskaris-d2l: I forgot to remove these before opening the PR. How would you suggest adding a debug mode? I like that better than just killing the lines.

dlockhart: Hmm. You could make it part of the configuration it grabs off the document. Or just a global that gets set. Or an `options` object that gets passed into the entrypoint of the API.

helpers/getLocalizeResources.js
dlockhart: Cool makes sense, thx.

helpers/getLocalizeResources.js
dlockhart: Hmm, that's still surprising to me. I would have expected for the non-static resources case for the dynamic import model inside a switch statement to only fetch the current document's language at runtime. At least in [our example](https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.md#example-2-dynamically-imported-resources) that's what we intended with the `return` statement inside the loop.

ppaskaris-d2l: hmm, I can change it to work like that then. I found multiple examples of doing it this way:

- https://github.com/BrightspaceHypermediaComponents/activities/blob/fa932e04e528af6fae09fede0026458d13053ff4/components/d2l-activity-editor/localization.js
- https://github.com/BrightspaceHypermediaComponents/activities/blob/fa932e04e528af6fae09fede0026458d13053ff4/components/d2l-activity-collection-editor/localization.js
- https://github.com/Brightspace/discovery-fra/blob/48daf747abb9db71bb602f570dc793ad605fd7e4/src/components/d2l-discover-list/localization.js
- https://github.com/Brightspace/attachment/blob/e44cab1f0cecc55dd93acf59212fabc6872c0bd3/helpers/localization.js

But I see now that there are plenty that do it the way you describe, so I'll change it to only fetch the first match.

helpers/getLocalizeResources.js
ppaskaris-d2l: I settled on making it a special export that you can opt-into during development for now.

helpers/getLocalizeResources.js
dlockhart: Wow yeah that was definitely not the intent when we made the mixin. I think it makes sense to just fetch the one though right? Like otherwise we'll be seeing ~12X the number of requests -- albeit cached, but browser caches definitely aren't infinite so evictions of valid things might happen.

ppaskaris-d2l: Yeah, I guess I picked a suboptimal component to base this off of (FACE) and then started retconning why it was written like that 😄

dlockhart: Retconning... good word.

kieranderson:
Looking at getting this in with the caching updates to come in another PR. Still some more to do in the LMS before we start switching the web components to using this new way of fetching terms.

PR #466, opened 2020-04-08
Tooltip Positioning Cleanup

AllanKerr:
# Changes
1. Update the tooltip boundary so that the right boundary is relative to its `offsetParent`'s right instead of left.
1. Remove custom-target attribute due to no longer being used anywhere


AllanKerr:
4 PRs are currently in review that are needed for the right boundary change. Waiting until those are approved so merge so I can merge all at once.

components/tooltip/tooltip.js
```diff
@@ -45,13 +45,6 @@ class Tooltip extends RtlMixin(LitElement) {
 		return {
 			align: { type: String }, /* Valid values are: 'start' and 'end' */
 			boundary: { type: Object },
-			customTarget: {
-				type: Object, attribute: 'custom-target', converter: {
-					/* required because some places are setting custom-target="" to prevent _findTarget
-					from falling back to its parent when the 'for' attribute is missing */
-					fromAttribute: value => (value === undefined ? undefined : value === '' ? '' : JSON.parse(value))
-				}
-			},
```
dlockhart, line 54: 🎉

PR #467, opened 2020-04-08
Dbatiste/backdrop improvements

dbatiste:
Improvements:

* updated to use property so flow would be same whether setting using attribute or property
* use isVisible helper instead of offsetParent since offsetParent is null whether position is fixed
* updated to honor reduce motion preference
* add the regular stuff - readme, demo
* add visual-diff test that uses reduce motion media query


components/backdrop/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+<div id="target"><button>toggle backdrop</button></div>
+<d2l-backdrop for-target="target" shown></d2l-backdrop>
+```
+
+Set the visible state of the backdrop by using the `shown` attribute/property.
+
+```javascript
+button.addEventListener('click', () => {
+  backdrop.shown = true;
```
dlockhart, line 23: If it's toggling, shouldn't this be `backdrop.shown = !backdrop.shown;`?

components/backdrop/backdrop.js
```diff
@@ -1,16 +1,19 @@
...
+
 const scrollKeys = [];
 let scrollOverflow = null;
 
 class Backdrop extends LitElement {
 
 	static get properties() {
 		return {
-			shown: { type: Boolean },
+			shown: { type: Boolean, attribute: true },
```
dlockhart, line 16: Is this needed? I thought observers were created automatically and you only needed to set this if you wanted to rename the attribute or not have an observer.

components/backdrop/test/backdrop.visual-diff.js
```diff
@@ -0,0 +1,32 @@
...
+describe('d2l-backdrop', () => {
+
+	const visualDiff = new VisualDiff('backdrop', __dirname);
+
+	let browser, page;
+
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await browser.newPage();
+		await page.emulateMediaFeatures([{ name: 'prefers-reduced-motion', value: 'reduce' }]);
```
dlockhart, line 13: Probably don't want everyone to have to set this in all their tests (they'll forget, or do it inconsistently)... is there an easier way we could expose this for everyone?

components/backdrop/backdrop.js
dbatiste: I just tried this out again and it appear to work without... will remove.  I must have added while having some trouble and thought that it was needed.

components/backdrop/test/backdrop.visual-diff.js
dbatiste: 🤔 of some options but they all have issues.  Perhaps lets discuss tomorrow if you have time.  :)

PR #469, opened 2020-04-09
Visual Diff Typography Preload

AllanKerr:
# Changes
- Add typography preload to prevent visual diff tests failing due to the font not loading at initial page load. This change is required because browsers do not load fonts until they are displayed on the page. For visual diff tests without any text initially on the page, font loading gets missed by the `networkidle0` wait.

### DIFF
--- a/test/typography-preload.js
+++ b/test/typography-preload.js
@@ -1,0 +1,15 @@
+import { fonts, importUrl } from '../components/typography/typography.js';
+
+if (!document.head.querySelector('.d2l-typography-font-face-preload')) {
+	for (const type of Object.keys(fonts)) {
+
+		const preload = document.createElement('link');
+		preload.rel = 'preload';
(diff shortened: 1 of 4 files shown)


test/typography-preload.js
```diff
@@ -0,0 +1,15 @@
...
+
+if (!document.head.querySelector('.d2l-typography-font-face-preload')) {
+	for (const type of Object.keys(fonts)) {
+
+		const preload = document.createElement('link');
+		preload.rel = 'preload';
+		preload.className = 'd2l-typography-font-face-preload';
+		preload.href = `${new URL(`${fonts[type]}.woff2`, importUrl)}`;
+		preload.as = 'font';
+		preload.crossOrigin = 'anonymous';
```
dlockhart, line 11: This reminds me... we should probably do this in the LMS. Separate issue though.

PR #471, opened 2020-04-13
input-date: mobile-related improvements

margaree:
- change max-width of dropdown content
- if calendar isn't fully visible when dropdown-content opens, scroll to show it all (used `d2l-focus-trap` because this wraps all the content better than just using `d2l-calendar`)


margaree:
This is ready for re-review if needed. Some css changes were made in order to allow the width of the dropdown content to be smaller while having the different styles (focus, hover, selected) be correctly spaced.

PR #472, opened 2020-04-13
Dbatiste/reduce motion dialog

dbatiste:


components/dialog/dialog-mixin.js
```diff
@@ -17,6 +17,7 @@ if (window.D2L.DialogMixin.preferNative === undefined) {
 	window.D2L.DialogMixin.preferNative = true;
 }
 
+const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
```
dlockhart, line 20: This means you have to reload the page after changing your OS preference, ya? Not a big deal and likely worth the optimization trade-off.

components/dialog/dialog-styles.js
```diff
@@ -119,19 +119,23 @@ export const dialogStyles = css`
...
 			padding: 0 20px;
 		}
-
 		.d2l-dialog-footer {
 			padding: 20px;
 		}
-
 	}
 
+	@media (prefers-reduced-motion: reduce) {
```
dlockhart, line 133: Does it ever make sense to inverse this? i.e. only apply the transition in the first place if the value is `no-preference`. Or does that not work in browser who don't support this (IE11).

components/dialog/dialog-mixin.js
dbatiste: That's correct.  

If we want... we could run this each time it's needed, or we could be notified when it changes, which is useful for other types of media queries where the match result is expected to change.  Unwrapped, it would look something like:
```
const query = window.matchMedia('(prefers-reduced-motion: reduce)');
let reduceMotion = query.matches;
query.addListener((e) => reduceMotion = e.matches);
```
But because we wouldn't want lots of places defining that same code, we'd probably wrap it up in an object that observes that event, which would result in an extra import, which probably isn't worth it in this case as you mentioned.  :)

I think there might also be some trickiness with IE11, depending on whether `addListener` or `addEventListener` is used, but I'd have to retest it to see where it was falling over.

components/dialog/dialog-styles.js
dbatiste: Hmmm that's a good question.  Both old-Edge and IE11 both return `false` for these media queries, regardless of `reduce` or `no-preference`, so I think in order for animations to be the default, we need to write the queries to opt out by using `reduce`.  In my testing, the other browsers do translate `no-preference` -> `true` though.

components/dialog/dialog-mixin.js
dlockhart: Probably not worth it, mostly just curious if we'd get amazing instant results like with dark mode -- which would be sweet for demos and that's about all.

PR #473, opened 2020-04-13
list-item: composed d2l-list-item-selected event

omsmith:
Same situation as https://github.com/BrightspaceUI/core/pull/463

`<d2l-list-item>` is being rendered inside a wrapping "list-item" component.

### DIFF
--- a/components/list/list-item.js
+++ b/components/list/list-item.js
@@ -323,7 +323,8 @@ class ListItem extends RtlMixin(LitElement) {
 	_dispatchSelected(value) {
 		this.dispatchEvent(new CustomEvent('d2l-list-item-selected', {
 			detail: { key: this.key, selected: value },
-			bubbles: true
+			bubbles: true,
+			composed: true
 		}));
 	}
 


components/list/list-item.js
```diff
@@ -323,7 +323,8 @@ class ListItem extends RtlMixin(LitElement) {
 	_dispatchSelected(value) {
 		this.dispatchEvent(new CustomEvent('d2l-list-item-selected', {
 			detail: { key: this.key, selected: value },
-			bubbles: true
+			bubbles: true,
+			composed: true
```
dlockhart, line 327: Hmm... I might be worried that this is now going to pass-through shadow DOM and be caught by ancestors who weren't expecting it, whereas before it wouldn't have.

omsmith: Yeah - I have to figure out if it even fixes things before opening it anyway.

The case I'm trying to use it in (replacing a custom checkbox-for-list-items) is:

```
<some-dialog>
  <course-select-list></course-select-list>

  <!-- dialog buttons -->
  <button disabled={list.selectedItems.length === 0}>Submit</button>
</some-dialog>
```

where `course-select-list` is

```
onFirstUpdate() {
   this.addEventListener('d2l-list-selection-changed', () => {
    this.selectedItems = this.listItems.filter(x => x.isSelected);
   });
}

render() {
  return html`<d2l-list>${renderListItems()}</d2l-list>`
}

renderListItems() {
  return this.courses.map(course => html`<course-select-item course={course}></course-select-item>`;);
}
``

where `course-select-list-tem` is

```
render() {
  return html`<d2l-list-item selectable></d2l-list-item>`;
}
```

but `d2l-list` never sees the `d2l-list-item-selected` and thus never first the `d2l-list-selection-changed`

:shrug:

omsmith: This doesn't seem like a terribly unreasonable way to use the list components, but things are built to assume that the `d2l-list-item`s will be rendered by the same component as the list}

dlockhart: Yeah it was definitely not built with this kind of thing in mind. I'd actually worry about the accessibility side -- the roles we're applying ("list" and "listitem") need to be in a parent-child relationship. Alternatively, if things exist in the DOM in between them, those things need a role of `presentation` so that AT knows to skip them.

But yeah, I'm surprised this was working at all.

omsmith: So, something like this change ends up at expected usage then?

dlockhart: Yeah exactly.

dbatiste: I know you closed this, but thought to chime in to add a little more info.

Yeah, as Dave mentioned, the expectation was that the list and listitem roles are parent-child.  We did have some discussion about this when it was built, and there was also the possibility of creating a list-item mixin (that would specify the role) that allows that relationship to be maintained for "custom" list-items.  @m6jones may have more to say, and I know she is planning to rejig things a bit.

omsmith: Cool. Thanks @dbatiste and @dlockhart.

Consolidating the building of the list-item in its own component doesn't seem unreasonable, especially because there's different slots so you can't really do `<d2l-list-item><my-list-item-body/></d2l-list-item>`.

But, in this case it wasn't doing anything too terribly interesting or difficult, so wasn't bad!

m6jones: So this is 100% on purpose. The parent-child give us a lot of accessibility. The different slots also have a lot of important logic tied to it to make them usable and accessible.

This also makes sure all lists in our system look and act the same. Right now we have a lot of lists and 50% work with dragon. There is no reason for that. 

I wanted to go more extreme by making `d2l-list` hide none `d2l-list-item` components. 

With all this said, I am making changes such that it will not be hard to re-skin some different looking lists which can be made in labs.

The problem that this solves is people making their own list item styles and having them all work a little different and look slightly off.

For example manager dashboard has a list item where the checkbox does put the check in the center!

PR #474, opened 2020-04-14
button: aria* information table

margaree:
I have been working in the accessibility group to figure out ways to improve dev a11y usage/experience.

One idea I had was to use a table to make it easier for devs to find what attributes they can/should use on core components and why. I'm starting out with button. If you all think this seems like a good path, I'll work on something similar for the others (though I think button is the one with the most aria* attributes/most intertwined usages)

I put it in its own document, but it's small enough where it might be okay just in the regular readme.


margaree:
Good suggestions! I'm okay with the idea of mostly not mentioning `aria-label` other than with `d2l-button-icon` and just encouraging usage of `description`

components/button/README.md
```diff
@@ -19,6 +19,16 @@ The `d2l-button` element can be used just like the native button element, but al
 - `disabled` (Boolean): disables the button
 - `primary` (Boolean): styles the button as a primary button
 
+**Accessibility:**
+
+To make your `d2l-button` accessible to screen reader users, use the following properties when applicable:
```
dlockhart, line 24: Accessibility applies to more than just screen readers, so I'd drop that bit and just say:

> To make your `d2l-button` accessible, use the following properties when applicable:

PR #476, opened 2020-04-14
Tooltip Accessibility Fixes

AllanKerr:
# Changes
1. Change `aria-describedby` to only be used for elements that are consistently supported across browsers and screenreaders.
1. Add `announce` for elements without consistent `aria-describedby` support.
1. Non-Accessibility Change: Change pointer gap to `0px` based on feedback from design.


components/tooltip/tooltip.js
```diff
@@ -14,12 +15,19 @@ const pointerOverhang = 7; /* how far the pointer extends outside the content */
...
 const pointerRotatedOverhang = ((pointerRotatedLength - pointerLength) / 2) + pointerOverhang;
 
-const pointerGap = 6; /* spacing between pointer and target */
+const pointerGap = 0; /* spacing between pointer and target */
 const defaultViewportMargin = 18;
 const contentBorderRadius = 6;
 const contentBorderSize = 1;
 const contentHorizontalPadding = 15;
 
+const interactiveElements = {
```
AllanKerr, line 24: `a` wasn't included in this list because there is inconsistent behavior across screenreaders if no `href` is set.

components/tooltip/tooltip.js
```diff
@@ -564,18 +577,30 @@ class Tooltip extends RtlMixin(LitElement) {
 		return this._openDir === 'bottom' || this._openDir === 'top';
 	}
 
+	_isInteractive(ele) {
+		if (ele.nodeType !== Node.ELEMENT_NODE) {
+			return false;
+		}
+		const nodeName = ele.nodeName.toLowerCase();
+		return !!interactiveElements[nodeName];
```
dlockhart, line 585: Could also use an array with `indexOf(nodeName) > -1`

components/tooltip/tooltip.js
dlockhart: Would it be worth detecting that as a special case inside `_isInteractive`? i.e. the tag name is `a` AND it has an `href` set?

components/tooltip/tooltip.js
```diff
@@ -14,12 +15,19 @@ const pointerOverhang = 7; /* how far the pointer extends outside the content */
...
 const defaultViewportMargin = 18;
 const contentBorderRadius = 6;
 const contentBorderSize = 1;
 const contentHorizontalPadding = 15;
 
+const interactiveElements = {
+	'input': true,
+	'select': true,
+	'textarea': true,
+	'button': true
```
dlockhart, line 28: What about `img`? I know it's not an interactive element, but I was assuming that because screen readers recognize it and is has a label (`alt`) that `aria-describedby` would work.

components/tooltip/tooltip.js
```diff
@@ -296,6 +304,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		this._dismissibleId = null;
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this.setAttribute('role', 'tooltip');
```
dlockhart, line 309: Wondering if we should NOT apply the role if `aria-describedby` isn't getting set? [The spec](https://www.w3.org/WAI/PF/aria/roles#tooltip) seems to imply that it's a requirement, but not sure if it matters.

components/tooltip/tooltip.js
```diff
@@ -564,18 +577,30 @@ class Tooltip extends RtlMixin(LitElement) {
...
 	_onTargetFocus() {
+		const prevFocusing = this._isFocusing;
 		if (this.disableFocusLock) {
 			this.showing = true;
 		} else {
 			this._isFocusing = true;
 			this._updateShowing();
 		}
+		if (!prevFocusing && !this._isInteractive(this._target)) {
+			announce(this.textContent);
```
dlockhart, line 602: Nice!

components/tooltip/tooltip.js
AllanKerr: That was how I wrote it originally but I figured I'd use an object for direct key lookup rather than iterating the array (although the perf difference is irrelevant given how small the list is)

components/tooltip/tooltip.js
AllanKerr: There likely are a few more elements that could be added to the list. I'm happy to test a bunch of different elements if we're interested in expanding it a bit.

I left it with these ones because they're ones I'm confident are consistent in all cases regardless of attributes and I also foresee them getting used in the future when tooltip gets integrated into other `core` components like `d2l-button` and `d2l-input`.

components/tooltip/tooltip.js
AllanKerr: I considered that, the reason I didn't is because the `announce` works almost as good and I thought there might be some cases where the `href` attribute is dynamically added or removed which wouldn't be accounted for. I'm on the fence so I'm certainly open to adding that as a special case.

components/tooltip/tooltip.js
AllanKerr: That sounds reasonable to me, reading into other sources it sounds like all `role="tooltip"` does is expand support for `aria-describedby` with some screen readers so its useless if `aria-describedby` isn't present.

components/tooltip/tooltip.js
dlockhart: Yeah likely doesn't matter, and there's `Map` as well but that seems super overkill given the size.

components/tooltip/tooltip.js
dlockhart: I mostly bring up `img` and `a` because of the work Margaret did in the monolith to add tooltip support to the legacy `<Image>` control -- which can associate the tooltip with both of those depending on the setup.

components/tooltip/tooltip.js
dlockhart: Cool, don't feel strongly. Could mitigate the dynamic modification by doing the `_isInteractive` call at runtime (which you're already doing I think).

components/tooltip/tooltip.js
AllanKerr: Unfortunately it looks like `img` won't work. `aria-describedby` doesn't trigger announcments for `img` tags even if they have `alt` text and a `tabindex`

components/tooltip/tooltip.js
AllanKerr: The one issue there is that I'd need to also dynamically remove `aria-describedby` since JAWS will announce anchor tags with no `href` but NVDA won't. For now I'm going to stick with `announce` for `a` tags which I think should work pretty good but if issues do arise then I'll explore adding some special conditions for `a` with and without `href`.

components/tooltip/tooltip.js
dlockhart: Oh dear, that might ruin Margaret's use case then. Does `aria-labelledby` work? Although that'll just totally replace the `alt` likely.

AllanKerr: With `img` I think the `announce` helper is the best we'll be able to do. `aria-labelledby` and `aria-describedby` are super inconsistent with static content. [I found this article which seems to be accurate for the most part](http://www.davidmacd.com/blog/does-aria-label-override-static-text.html)

dlockhart: Seems that way yeah. Cool well thanks for investigating and all the testing!

AllanKerr: With the announce helper it reads:

> `${alt text}`
graphic
`${tooltip message}`

dlockhart: Perfect. 🥇

PR #480, opened 2020-04-15
Added an optional boolean align-top attribute to d2l-checkbox

mpharoah-d2l:
Added an optional `align-top` attribute to `d2l-checkbox`  that makes the checkbox align with the top of multi-line labels instead of the middle. This alignment is more appropriate for checkboxes in a tree view.


components/inputs/input-checkbox.js
```diff
@@ -62,6 +63,10 @@ class InputCheckbox extends RtlMixin(LitElement) {
 				.d2l-input-checkbox-text-disabled {
 					opacity: 0.5;
 				}
+				input.d2l-input-checkbox.align-top {
+					vertical-align: top;
+					margin-top: 0.2em;
+				}
```
dlockhart, line 69: How about using an attribute selector instead? Should be able to accomplish the same effect with no JS changes.

```css
:host([align-top]) .d2l-input-checkbox {
    margin-top: 0.2em;
    vertical-align: top;
}
```

components/inputs/input-checkbox.js
```diff
@@ -62,6 +63,10 @@ class InputCheckbox extends RtlMixin(LitElement) {
 				.d2l-input-checkbox-text-disabled {
 					opacity: 0.5;
 				}
+				input.d2l-input-checkbox.align-top {
+					vertical-align: top;
+					margin-top: 0.2em;
```
dlockhart, line 68: `0.2em` doesn't seem like the right value here. If I overlay the text baseline, there's more of the checkbox above the text than below.

Screen Shot 2020-04-15 at 12 59 44 PM

I'd defer to a designer (@geurts any thoughts?), but the correct value to me seems to be `0`:

Screen Shot 2020-04-15 at 1 04 21 PM

mpharoah-d2l: Using `0` makes it way too high:
image

mpharoah-d2l: Actually, I think changing the alignment of the text label might fix it without the need for a margin

mpharoah-d2l: Oh, I realized what's wrong. I'm setting the styling on the wrong element.

dlockhart: Hmm so that to me indicates that there's other factors at play here. If I apply `align-top` to a single-line checkbox I end up with something similar where the alignment seems "off":

Screen Shot 2020-04-15 at 1 23 47 PM

Have you considered just using a native checkbox element and [applying our checkbox styles](https://github.com/BrightspaceUI/core/tree/master/components/inputs#applying-styles-to-native-checkboxes) to it? Then you could do whatever you like to it. I still think toying with this kind of top alignment is possibly still desire-able, just want to make sure we're doing it correctly to handle all cases.

mpharoah-d2l: With my latest changes, it now seems to be correctly centred:
image

image

mpharoah-d2l:
image
🤔

dlockhart:
After some offline discussion, we've decide to make @mpharoah-d2l's change the default behaviour. So multi-line checkboxes will now be top-aligned and there should be no impact on single-line checkboxes.

We'll hold off on merging this until after branching on Friday just to be extra safe we don't accidentally introduce a regression.

dlockhart:
@mpharoah-d2l I needed to reduce things to `0.1rem` (which is `2px` at our default font size instead of the `4px` you had) in order to get the single-line visual diff tests [to show no differences](https://s3.ca-central-1.amazonaws.com/visualdiff.gaudi.d2l/screenshots/BrightspaceUI/core/input-checkbox/2020-04-15-21.06.22.991-78918e11d0058cf54d5da748fde7cd77762ea6e9-report.html).

Hopefully that'll still work for your use case. If not, I can experiment with some different approaches other than top-margin.

dlockhart:
Actually I backtracked and tried a different approach. I noticed that because our default line-height is `1.4rem` and we don't apply the `1.2rem` until we get to the text label, the host element actually has 1.4. That's causing an extra `4px` of space at the top of the element.

That extra `4px` wasn't as noticeable when everything was vertically centred, but when the text top-aligns it becomes more obvious, hence Matt why you needed top drop the text down `4px`.

By applying the `1.2rem` line-height on the host, and top-aligning both the text and the checkbox, everything is "tight" to the top and maintains the original relative position between the text and checkbox for single-line checkboxes.

The visual diff tests aren't going to like it though!

mpharoah-d2l:
Ah, that makes sense. I had originally used 0.2rem because I saw the that text had line height 1.4rem, and half the difference of that an 1rem is 0.2rem. But since we actually want 1.2rem, not 1.4, then an 0.1 margin makes sense :+1:

mpharoah-d2l:
Oh wait, it's because of the different between the checkbox line height and the label line height, I see. So now no margin is needed? That's much better

dlockhart:
Yeah exactly. Previously the checkbox & text was 1.2rem line-height but the parent container was incorrectly 1.4rem. And when it was vertically centred that resulted in an extra 0.1rem at the top and bottom -- 2px each. Centred it wasn't super obvious, but as soon as you top-align things it became visible.

So the result of this change is an overall 4px reduction in the height of the component by having 2px less whitespace at the top and bottom. It's a little risky if folks were relying on that space for some reason, but I think it's worth it to clean things up.

PR #485, opened 2020-04-16
Add reduced motion to components

dbatiste:
I excluded `input-date/time` and the `tooltip` components since I didn't want to create conflicts - some of their respective visual-diff tests can probably be simplified.


components/button/test/floating-buttons.visual-diff.js
```diff
@@ -10,13 +10,13 @@ describe('d2l-floating-buttons', () => {
 	before(async() => {
 		browser = await puppeteer.launch();
 		page = await browser.newPage();
-		await visualDiff.disableAnimations(page);
+		page = await visualDiff.createPage(browser);
```
margaree, line 13: Are both this line and line 12 needed?

components/button/test/button-icon.visual-diff.js
```diff
@@ -52,30 +42,4 @@ describe('d2l-button-icon', () => {
...
-	const focus = (page, selector) => {
-		return page.evaluate((selector) => {
-			return new Promise((resolve) => {
-				const elem = document.querySelector(selector);
-				elem.shadowRoot.querySelector('button').addEventListener('transitionend', resolve);
-				elem.focus();
-			});
-		}, selector);
-	};
-
```
dlockhart, line 80: Nice, so much simpler.

components/menu/test/menu-rtl.visual-diff.js
```diff
@@ -40,10 +40,12 @@ describe('d2l-menu rtl', () => {
...
 			// this scenario also tests height change when going from 1 menu item to 2 within nested menu
-			const selector = '#nested-item';
-			const resize = contentResize(page, selector);
-			await page.$eval(selector, (item) => item.click());
-			await resize;
+			await page.$eval('#nested-item', item => {
+				return new Promise((resolve) => {
+					item.addEventListener('d2l-hierarchical-view-show-complete', resolve, { once: true });
+					item.click();
+				});
```
dlockhart, line 47: Would this technique work? It's more common in our tests:

```
await page.$eval('#nested-item', item => {
    setTimeout(() => item.click());
    await oneEvent(item, 'd2l-hierarchical-view-show-complete');
});
```

Or does the extra `once: true` business mess that up...

components/button/test/floating-buttons.visual-diff.js
dbatiste: Thanks!... will update.  Missed that.

components/menu/test/menu-rtl.visual-diff.js
dbatiste: Yeah we could do something like that - we'd just need to get `oneEvent` into the page with a scope that is globally accessible.  Or define our own helper to go in the page:
```
<script type="module">
    import { oneEvent } from '@open-wc/testing';
    window.oneEvent = oneEvent;
</script>
```
or our own...
```
<script>
    window.oneEvent = (elem, name) => {
            elem.addEventListener(name, resolve, { once: true });
        });
    };
</script>
```
The downside being devs need to remember to include this stuff in their HTML page, but there might a way to inject it automatically with Puppeteer.  There may be a couple other helpers that could be useful, such as sending particular key events - I think that has been repeated a couple of times.

Now that I know the `$eval` vs. `evaluate` issue, I could move this back to something like:
```
it('opens submenu on click', async function() {
    const showCompleteEvent = getShowCompleteEvent('#nested-item');
    page.$eval('#nested-item', item => item.click());
    await showCompleteEvent;
    const rect = await visualDiff.getRect(page, '#nested');
});

const getShowCompleteEvent = (selector) => {
    return page.$eval(selector, elem => {
            elem.addEventListener('d2l-hierarchical-view-show-complete', resolve, { once: true });
        });
    });
};
```
This has the advantage of keeping all the test code in once place, but we will end up with a bunch of this. 

I'll plan to merge as-is, and iterate on this further based on what we decide.

dlockhart: Ah right I forgot our visual diff tests run in a different world.

PR #489, opened 2020-04-17
Calendar/input-date: Transition tweaks

margaree:
- Calendar: only the first date receiving focus when entering the grid should have the box shadow transition; subsequent navigation with arrow keys should not cause transition
- Calendar: updated transitions depending on how month is switched:
    - up arrow: vertical transform, comes from above
    - down arrow: vertical transform, comes from below
    - page up: vertical transform: comes from above
    - page down: vertical transform: comes from below
    - remainder (left arrow, right arrow, clicking left and right buttons) are same left/right horizontal transforms
- input-date: reduced top and bottom padding on calendar in dropdown
- Implemented `prefers-reduced-motion` required changes


components/calendar/test/calendar.visual-diff.js
```diff
@@ -316,27 +298,6 @@ describe('d2l-calendar', () => {
...
-		});
-	};
-
-	const dateChangeEvent = (page, selector) => {
-		return page.$eval(selector, (elem) => {
-			return new Promise((resolve) => {
-				elem.shadowRoot.addEventListener('transitionend', resolve, { once: true });
-			});
-		});
-	};
```
dlockhart, line 338: Nice!

PR #491, opened 2020-04-20
input-date: reset calendar calendar when date typed

margaree:
Bug: when a date is typed and then calendar is opened, calendar does not open to the month containing the typed date.


components/inputs/input-date.js
```diff
@@ -226,16 +226,22 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 		// open dropdown on down arrow or enter and focus on calendar focus date
 		if (e.keyCode === 40 || e.keyCode === 13) {
 			this._dropdown.open();
-			this._calendar.focus();
+			requestAnimationFrame(async() => {
```
margaree, line 229: otherwise focus goes to the wrong date (it will focus on the date where the previous focus date was in the previously shown month)

components/inputs/input-date.js
```diff
@@ -270,8 +278,11 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 		this._dropdownOpened = true;
 	}
 
-	_handleMouseup() {
-		if (!this.disabled) this._dropdown.toggleOpen(false);
+	_handleMouseup(e) {
+		if (!this.disabled) {
+			if (!this._dropdownOpened) this._handleChange(e);
```
margaree, line 283: handle the case where date is typed then input is clicked and calendar opens

components/calendar/calendar.js
```diff
@@ -382,7 +382,9 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			if (prop === '_shownMonth' && this._keyboardTriggeredMonthChange) {
 				this._focusDateAddFocus();
 			} else if (prop === 'selectedValue') {
-				this._updateFocusDate();
+				requestAnimationFrame(async() => {
```
margaree, line 385: `dateElem` in `_updateFocusDate` might not exist yet if month hasn't changed

components/inputs/input-date.js
dlockhart: I'm slightly confused... it's listening to the `change` event on `d2l-input-text`, which is only going to fire once the input loses focus. But hitting the down key (or ENTER key) isn't going to trigger a focus loss, so how does waiting for an animation frame solve this? Maybe opening the dropdown causes focus loss?

margaree: I think the enter key is triggering the `change` event (though that's made me realize that this doesn't work for the down arrow)

dlockhart: I see. Yeah seems like we need to treat the ENTER and down arrow keys the same way as blurring the input -- trigger a parse & validate, which may result in the previous value coming back. Only after all that happens should the dropdown open and focus be applied.

PR #492, opened 2020-04-20
Fix Broken Tooltip Constructor

AllanKerr:
# Defect
- Fix defect where `document.creatElement('d2l-tooltip')` triggers an error due to attributes being set in the constructor.
- Error: `Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes`

# Fix
- Move `showing` to the `connectedCallback` so `setAttribute` isn't called in the constructor.
- I tried to create a test case that calls `createElement('d2l-tooltip')` but it doesn't seem to cause the exception to be thrown.

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -269,7 +269,6 @@ class Tooltip extends RtlMixin(LitElement) {
 		this.disableFocusLock = false;
 		this.forceShow = false;
 		this.offset = pointerRotatedOverhang + pointerGap;
-		this.showing = false;
 		this.state = 'info';
 
 		this._dismissibleId = null;
@@ -292,6 +291,7 @@ class Tooltip extends RtlMixin(LitElement) {
 
 	connectedCallback() {
 		super.connectedCallback();
+		this.showing = false;
 		window.addEventListener('resize', this._onTargetResize);
 
 		requestAnimationFrame(() => {


components/tooltip/tooltip.js
```diff
@@ -269,7 +269,6 @@ class Tooltip extends RtlMixin(LitElement) {
 		this.disableFocusLock = false;
 		this.forceShow = false;
 		this.offset = pointerRotatedOverhang + pointerGap;
-		this.showing = false;
```
dbatiste, line 272: For sake of clarity for anyone reading this later, is this because the `showing` setter gets called, which in turn calls `_showingChanged` which calls `setAttribute` for `aria-hidden`, which would not be allowed in the constructor for custom element?

AllanKerr: Yeah, the defect appeared in `v1.42.1` when the `setAttribute('aria-hidden', 'false')` call was added to `_showingChanged`. `showing` and `state` were both reflected and set in the constructor but don't cause any issues on their own.

components/tooltip/tooltip.js
```diff
@@ -292,6 +291,7 @@ class Tooltip extends RtlMixin(LitElement) {
 
 	connectedCallback() {
 		super.connectedCallback();
+		this.showing = false;
```
dlockhart, line 294: Worth adding a unit test that just creates the element using `createElement`?

AllanKerr: Going to add a test to all components in another PR

PR #493, opened 2020-04-20
Add Constructor Tests

AllanKerr:
# Changes
- Add constructor tests to all components to ensure that they can successfully be created. This test case was added because `createElement` will fail if custom elements set attributes within their constructor. These tests protect against that.

### DIFF
--- a/components/dialog/test/dialog.test.js
+++ b/components/dialog/test/dialog.test.js
@@ -1,0 +1,14 @@
+import '../dialog.js';
+import { runConstructor } from '../../../tools/constructor-test-helper.js';
+
+describe('d2l-dialog', () => {
+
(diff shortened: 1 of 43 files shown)


components/dialog/test/dialog.test.js
```diff
@@ -0,0 +1,14 @@
+import '../dialog.js';
+import { runConstructor } from '../../../tools/constructor-test-helper.js';
+
+describe('d2l-dialog', () => {
+
+	describe('constructor', () => {
+
+		it('should construct', () => {
+			runConstructor('d2l-dialog');
```
dlockhart, line 9: There's `d2l-dialog-config` as well.

PR #494, opened 2020-04-21
Input README enhancements

margaree:
Changes:
- Separate out components into separate READMEs
- Add accessibility tables where applicable
- document `d2l-input-time`


components/inputs/documentation/date-time-inputs.md
```diff
@@ -0,0 +1,109 @@
...
+| Attribute | Description |
+|--|--|
+| `label` | **REQUIRED** [Acts as a primary label on the input](https://www.w3.org/WAI/tutorials/forms/labels/). Visible unless `label-hidden` is also used. |
+| `label-hidden` | Use if label should be visually hidden but available for screen reader users |
+
+**Events:**
+
+* `d2l-input-date-change`: dispatched when a date is selected or typed. `value` reflects the selected value and is in ISO 8601 calendar date format (`YYYY-MM-DD`).
+
+## Time Inputs
```
margaree, line 39: FYI @Scy-D2L - let me know if I missed anything here/anything looks incorrect

components/inputs/README.md
```diff
@@ -2,14 +2,13 @@
...
-- [Radio Buttons](#radio-buttons)
-- [Date](#date-inputs)
-- [Date-Time](#date-time-inputs)
+- [Text](documentation/text-inputs.md)
+- [Text Areas](documentation/text-areas.md)
+- [Search](documentation/search-inputs.md)
+- [Select Lists](documentation/select-lists.md)
+- [Checkboxes](documentation/checkboxes.md)
+- [Radio Buttons](documentation/radio-buttons.md)
+- [Date & Time Inputs](documentation/date-time-inputs.md)
```
dlockhart, line 11: Nit: worth alphabetizing the list?

components/inputs/documentation/checkboxes.md
```diff
@@ -0,0 +1,81 @@
...
+|--|--|
+| `aria-label` | Use when text on checkbox does not provide enough context |
+
+**Methods:**
+
+- `simulateClick()`: useful for testing, it simulates the user clicking on the checkbox, which toggles the state of the checkbox and fires the `change` event
+
+**Events:**
+
+When the checkbox's state changes, it dispatches the `change` event:
```
dlockhart, line 38: Nit: extra space

components/inputs/README.md
```diff
@@ -2,14 +2,13 @@
...
 There are various input components available:
 
-- [Text](#text-inputs)
-- [Text Areas](#text-areas)
-- [Search](#search-inputs)
-- [Select Lists](#select-lists)
-- [Checkboxes](#checkboxes)
-- [Radio Buttons](#radio-buttons)
-- [Date](#date-inputs)
-- [Date-Time](#date-time-inputs)
```
dlockhart, line 12: Nit: I'd be OK with just having the MD files inline and named the same as the `.js` file or in `docs/` (shorter)?

components/inputs/documentation/date-time-inputs.md
```diff
@@ -0,0 +1,109 @@
...
+  value="2020-11-20">
+</d2l-input-date>
+```
+
+**Properties:**
+
+- `label` (String, required): accessible label for the input
+- `disabled` (Boolean): disables the input
+- `label-hidden` (Boolean): hides the label visually (moves it to the input's `aria-label` attribute)
+- `value` (String, default: `''`): value of the input. This should be in ISO 8601 calendar date format (`YYYY-MM-DD`) and should be localized to the user's timezone (if applicable).
```
dlockhart, line 24: Probably a good time to have a section in here on how to convert to and from the user's timezone? Then "localized to the user's timezone" could link down to that?

components/inputs/documentation/date-time-inputs.md
```diff
@@ -0,0 +1,109 @@
...
+To make your usage of `d2l-input-date` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `label` | **REQUIRED** [Acts as a primary label on the input](https://www.w3.org/WAI/tutorials/forms/labels/). Visible unless `label-hidden` is also used. |
+| `label-hidden` | Use if label should be visually hidden but available for screen reader users |
+
+**Events:**
+
+* `d2l-input-date-change`: dispatched when a date is selected or typed. `value` reflects the selected value and is in ISO 8601 calendar date format (`YYYY-MM-DD`).
```
dlockhart, line 37: Noticed this as a discrepancy between the date and time inputs -- the name of the change event. Could we just go with `change` for the name here? Then it's consistent with `d2l-input-text`, `d2l-input-checkbox` and `d2l-input-time`. Would need to change `d2l-input-date-time` as well.

components/inputs/documentation/date-time-inputs.md
```diff
@@ -0,0 +1,109 @@
...
+</script>
+<d2l-input-date-time
+  label="Start Date"
+  value="2020-11-20T12:00:00.000Z">
+</d2l-input-date-time>
+```
+
+**Properties:**
+
+- `label` (String, required): accessible label for the input
```
dlockhart, line 95: Did we decide against supporting `label-hidden` here? I'm fine with that, just curious if it was on purpose. Can always add later.

components/inputs/documentation/date-time-inputs.md
```diff
@@ -0,0 +1,109 @@
...
+
+To make your usage of `d2l-input-date-time` accessible, use the following property:
+
+| Attribute | Description |
+|--|--|
+| `label` | **REQUIRED** [Acts as a primary label on the input](https://www.w3.org/WAI/tutorials/forms/labels/) |
+
+**Events:**
+
+* `d2l-input-date-time-change`: dispatched when a change in selected date or selected time (when date is already selected). `value` reflects the selected value and is in ISO 8601 calendar date-time format (`YYYY-MM-DDTHH:mm:ss.sssZ`).
```
dlockhart, line 109: Ditto here on the name of the change event.

components/inputs/documentation/date-time-inputs.md
margaree: `input-date-time` uses a fieldset which as far as I can tell does not support a hidden label. That could be implemented if it's something we want/that is okay to implement on a fieldset.

components/inputs/documentation/date-time-inputs.md
Scy-D2L: Is there any use-case for knowing if it was the date or time specifically that changed in a date-time input, or are we assuming the complete value needs to be checked on any adjustment?

components/inputs/docs/input-date-time.md
```diff
@@ -0,0 +1,193 @@
...
+## Timezone
+
+The `input-date` and `input-time` components do not handle timezone and so require the input to be in the user's timezone, which corresponds to the timezone in the `data-timezone` attribute on the `html` element.
+
+### Convert UTC Date/Time to Local Date
+
+To convert a UTC date/time (presumably in ISO 8601 format, `YYYY-MM-DDTHH:mm:ss.sssZ`, adjust if not) to a date in the user's timezone for use in `input-date`:
+- Parse into object (`parseISODateTime` from `helpers/dateTime.js`)
+- Convert UTC object into local timezone object (`convertUTCToLocalDateTime` from the [intl library](https://github.com/BrightspaceUI/intl#datetime-conversion-based-on-user-timezone))
+- Format as ISO 8601 date (`formatDateInISO` from `helpers/dateTime.js`)
```
dlockhart, line 120: 3 steps across two libraries seems like a lot of duplicated effort. Doesn't need to happen in this PR, but I think we can improve this such that we could provide something that takes an ISO UTC date string -> ISO local date string.

margaree: Yup agreed

PR #496, opened 2020-04-21
Sakulick/dropdown max height

Scy-D2L:
Adds `max-height` property to `d2l-dropdown`
`max-height` includes header, content, and footer of dropdown


components/dropdown/dropdown-content-mixin.js
```diff
@@ -461,7 +466,8 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			});
 
 			const spaceRequired = {
-				height: contentRect.height + headerFooterHeight + 10,
+				height: this.maxHeight !== null && contentRect.height > this.maxHeight
```
dbatiste, line 469: Would you not want to check? 
```
contentRect.height + headerFooterHeight + 10 > this.maxHeight
```

components/dropdown/dropdown-content-mixin.js
```diff
@@ -477,8 +483,9 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 
 			const maxHeight = Math.floor((this.openedAbove ? spaceAround.above : spaceAround.below) - headerFooterHeight);
 			if (!this.noAutoFit && maxHeight && maxHeight > 0) {
-				this._maxHeight = maxHeight;
-				this.__toggleOverflowY(contentRect.height > maxHeight);
+				this._height = this.maxHeight !== null && maxHeight > this.maxHeight && headerFooterHeight < this.maxHeight
```
dbatiste, line 486: These conditions make me a bit nervous, and to be honest it's not immediately what cases they are.  Hopefully they are captured in visual-diff tests.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -477,8 +483,9 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 
 			const maxHeight = Math.floor((this.openedAbove ? spaceAround.above : spaceAround.below) - headerFooterHeight);
```
dbatiste, line 484: The variable names have becomes somewhat confusing.  Is my understanding correct that this `maxHeight` is effectively the maximum possible height for the **content**, as constrained by the viewport?  Whereas, `this.maxHeight` (referenced in the code changes below) is the maximum height for the entire dropdown including the header and footer that's desired by the consumer.  And, ultimately the maximum height that the component uses is the smaller of the two, factoring in the header and footer?

dbatiste:
Originally I think the intention was that the dropdown would be as tall as it needs to be without overflowing the viewport. Just curious to know the use case that requires adding the extra complexity to override... might be worth adding to the description.

dlockhart:
I don't think you were in the date/time meetings where this keeps coming up, but the use-case is the one where the date input opens the calendar dropdown down (because it's smaller), but then when you move to the time input it opens up because it's much larger, creating a jarring experience. We'd like to put a max-height on the time dropdown that's no higher than the date dropdown, ensuring that they always both open the same direction.

dbatiste:
Thanks Dave.  That makes sense.  I'm definitely a little nervous about it.

components/dropdown/dropdown-content-mixin.js
Scy-D2L: I added a comment to give some explanation and another v-diff test to cover the other case. In longer form, to sanity-check that the comment and behaviour make sense:

- We only want to apply the max height if it exists and there's enough space to need to restrict it.
- If the header and footer combined are greater than the max height, no content would show up, so we default to regular sizing to make sure content appears (the header and footer sections can't scroll so there's nothing to be done about that 🤷) Typically this shouldn't happen but it's a graceful failure in case of changes impacting header/footer content.

components/dropdown/dropdown-content-mixin.js
Scy-D2L: The +10 is padding related and needs to be added to the result either way, but good catch on the header and footer part I fixed that :)

components/dropdown/dropdown-content-mixin.js
Scy-D2L: I've updated the variable names into something that is hopefully easier to read and understand. 🤔 feedback welcome.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -461,7 +465,9 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			});
 
 			const spaceRequired = {
-				height: contentRect.height + headerFooterHeight + 10,
+				height: this.maxHeight !== null && contentRect.height + headerFooterHeight > this.maxHeight
+					? this.maxHeight + 10
+					: contentRect.height + headerFooterHeight + 10,
```
dlockhart, line 470:
```suggestion
				height: Math.min(this.maxHeight ?? Number.MAX_VALUE, contentRect.height + headerFooterHeight) + 10,
```

Scy-D2L: Cool! :)

PR #497, opened 2020-04-21
List item checkbox mixin for new list-item architecture

zinabat:
**Context**
[US115493](https://rally1.rallydev.com/#/detail/userstory/383949056036?fdp=true)
[Design Details](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#d2l-list-item-placement-marker)

Introduces the `ListItemCheckboxMixin` which will add and handle checkbox behaviour to a `list-item`. This change is in line with the new `d2l-list-item` design which will change list-items to use a grid.

**Note that this PR does not change the list item itself.** This is only precursor work to implement the mixin so that the list-item can be changed to use it later.

Screen Shot 2020-04-21 at 4 31 51 PM

**Quality**
* The list-item-sample has been added and changed to demo the mixin
* Four new unit tests have been added to ensure the mixin works as expected

To test these changes:
* Run `npm run test:headless` and inspect the `ListItemCheckboxMixin` tests
* Run `polymer serve` and visit http://127.0.0.1:8081/components/@brightspace-ui/core/components/list/demo/list-item-sample.html


components/list/demo/list-item-sample.html
```diff
@@ -0,0 +1,55 @@
...
+		import '../list-item-content.js';
+		import './list-item-sample.js';
+	</script>
+</head>
+
+<body unresolved>
+
+	<d2l-demo-page page-title="d2l-list-item-sample">
+
+		<h2>List item sample for development</h2>
```
dbatiste, line 21: To be semantically correct, the items should be placed inside of a `d2l-list` element, even for the demo.  This will allow a tester to check both the behavior of `d2l-list` with custom items as well as the behavior of the mixin.

Can we just add this demo to the existing demo page?  So far I think this could go quite nicely with the rest of the demos snippets, but we should just provide demo for the bits that are different or could be affected by overrides, and omitting what's already covered by the standard list-item demo.
```
<d2l-list>
    <d2l-list-item-sample key="foobar selectable="true" draggable="true">
       ...
    </d2l-list-item-sample>
</d2l-list>
```

Another minor note: the string values should be omitted from the boolean attributes:
```
<d2l-list-item-sample key="foobar selectable draggable disabled>...</d2l-list-item-sample>
```
The exception to that rule is aria attributes, which awkwardly expect/require string values.

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,50 @@
...
+			disabled: {type: Boolean },
+			href: { type: String },
+			key: { type: String, reflect: true },
+			role: { type: String, reflect: true },
+			_breakpoint: { type: Number }
+		};
+	}
+
+	static get styles() {
+		return [ (super.styles ? super.styles : css``), css`
```
dbatiste, line 18: I am curious if this is necessary?  I know properties of the mixin are automatically mixed with the subclass, but I am not sure about the styles.  In other places we'd placed the styles in a separate file and consumers import them, which isn't ideal from a request perspective.

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,50 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ListItemCheckboxMixin } from '../list-item-checkbox-mixin.js';
+//import { ListItemDragMixin } from '../list-item-drag-mixin.js';
+
+class ListItemSample extends ListItemCheckboxMixin(LitElement) {
+	static get properties() {
```
dbatiste, line 6: I think `disabled`, `key`, and `role` should be coming from the mixin(s).  Presumably the breakpoint stuff is specific to our built-in list-item layout and isn't going into the mixin, but I am not sure.  (@m6jones , what did you intend here?)

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,123 @@
+import { css, html } from 'lit-element/lit-element.js';
+import { checkboxStyles } from '../inputs/input-checkbox-styles.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
+
+export const ListItemCheckboxMixin = superclass => class extends superclass {
+
+	/**
```
dbatiste, line 7: Nit: let's get rid of the comments here.  We've been avoiding the comments except in cases where they are key to understanding why something is done where not obvious.

components/list/demo/list-item-sample.html
dbatiste: I see you comment in Slack now.  Hopefully, unless we need a special page, the mixin functionality can eventually be adequately demo'd on the existing page.

components/list/demo/list-item-sample.js
zinabat: The styles are not mixed automatically - this is something we're doing in other places: https://github.com/BrightspaceUI/core/blob/master/components/hierarchical-view/hierarchical-view.js#L7

It seemed cleaner than putting the styles in a separate file when they are only related to the mixin, but I could be convinced of another way

dbatiste: Ok thanks.  Yeah I see menu does this as well, with the hierarchical-view as you mention.  I think there are other places we may similarly optimize to avoid a separate HTTP request.  Something I will look into.  :)

components/list/demo/list-item-sample.js
m6jones: `breakpoint` I forgot about that stuff!

m6jones: `disabled` `key` should be part of the `checkboxMixin`

I would leave the `role` and `breakpoint` logic here for now.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,123 @@
...
+			disabled: {type: Boolean },
+			key: { type: String, reflect: true },
+			selectable: {type: Boolean },
+			selected: { type: Boolean, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [ checkboxStyles, css`
+		.list-item-checkbox {
```
m6jones, line 24: These classes will be defined in the `d2l-generic-list-item` so you just have to extend the class `control-area` and `control-action-area`

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/**
+	 * Dispatches a custom event with information on what
+	 * items are selected and their keys.
+	 * https://lit-element.polymer-project.org/guide/events
+	 *
+	 * @param {*} value New state of checkbox
+	 */
+	_dispatchSelected(value) {
```
m6jones, line 52: Method order. 

https://github.com/Brightspace/eslint-config-brightspace/blob/master/sort-member-config.js

Not sure why the linter isn't picking this up.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/**
+	 * Sets whether the item is selected. Requests an event dispatch
+	 * unless suppressEvent is true
+	 *
+	 * @param {*} selected
+	 * @param {*} suppressEvent
+	 */
+	setSelected(selected, suppressEvent) {
```
m6jones, line 66: nit: default `suppressEvent = false` so people know it is optional.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,123 @@
...
+				<input
+					id="${this._checkBoxId}"
+					class="d2l-input-checkbox"
+					@change="${this._handleCheckboxChange}"
+					type="checkbox"
+					.checked="${this.selected}"
+					?disabled="${this.disabled}">
+			</div>
+			<div class="list-item-checkbox-action" @click="${this._handleCheckboxActionClick}"></div>
+			` : '';
```
m6jones, line 121: Use the nothing render here: https://lit-html.polymer-project.org/guide/writing-templates#rendering-nothing

components/list/list-item-checkbox-mixin.js
dbatiste: Yeah I don't think that plugin has been working... at one point it was.

components/list/list-item-checkbox-mixin.js
zinabat: Didn't even know about this - thanks!

components/list/demo/list-item-sample.html
zinabat: Yeah, the sample is only for development of the mixins in the interim before we actually change the list-item itself. It will be removed and demoed properly on the existing demo page.

components/list/list-item-checkbox-mixin.js
zinabat: I'm all for consistency for the most part, but I'd like to leave the function headers in. If they can help just one person while developing, I would say it's worth it. I have a plugin that shows function documentation automatically in my IDE, which I find pretty useful.

As an aside, there's also a plugin you can install that hides all comments and function headers by default :)

components/list/list-item-checkbox-mixin.js
zinabat: I pretty much never remember to do this when it isn't caught by the linter.

components/list/demo/list-item-sample.js
dbatiste: I feel pretty strongly that this should not be specified by consumers.  Will look forward to iterating on this.

components/list/list-item-checkbox-mixin.js
dbatiste: Which IDE are you using.  Generally we've leaned toward keeping the comments lean since they are frequently redundant (as over the years we've gone down this path from time to time) and requires extra maintenance.  Happy to consider this, but would definitely like to be consistent throughout `core`.

components/list/list-item-checkbox-mixin.js
m6jones: same but it makes the coding experience so much better. I now have muscle memory of where parts of the code should be and get anxiety when it isn't.

components/list/list-item-checkbox-mixin.js
zinabat: I use VScode. Function headers could save time in the long run despite extra maintenance - if someone is changing a function such that they need to change the function header, this can be a helpful sanity check. I'm always trying to think of ways to ease development for those who are new to a code base.

If this is something that we want to add to more of `core`, I'm happy to do that extra initial maintenance. Otherwise, I'm happy to keep a small island of functions with headers maintained where needed.

components/list/demo/list-item-sample.js
m6jones: Yeah, this will come later closer to the end of of all these changes.

m6jones: I promise that a consumer will never have to touch roles. You know me I believe strongly in that too :D

zinabat: I've added a TODO note and removed role and breakpoint altogether for now to defer for later :)

components/list/list-item-checkbox-mixin.js
zinabat: I've removed function headers for now so we can discuss as a group another time :)

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,85 @@
...
+		}
+		this.setSelected(!this.selected);
+	}
+
+	_handleCheckboxChange(event) {
+		this.setSelected(event.target.checked);
+	}
+
+	_renderCheckbox() {
+		return this.selectable ? html`
```
dbatiste, line 72: Where are these slots defined?

zinabat: Good question - they are defined in a different PR for `list-item-generic`

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,47 @@
...
+			.d2l-list-item-actions {
+				grid-column: actions-start / actions-end;
+				grid-row: 1 / 2;
+			}
+		`];
+	}
+	render() {
+		return html`
+			<div class="list-grid">
+				${ this._renderDragHandle ? this._renderDragHandle() : '' }
```
dbatiste, line 38: Seems like this is future-looking for drag & drop...?  It seems like this conditional should not be needed - this helper would be provided by one of the mixins that are being extended, and we should not need to check for it.

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,47 @@
...
+			}
+			.d2l-list-item-actions {
+				grid-column: actions-start / actions-end;
+				grid-row: 1 / 2;
+			}
+		`];
+	}
+	render() {
+		return html`
+			<div class="list-grid">
```
dbatiste, line 37: Nit: normally I would say this should be `d2l-list-item-grid`.  For production components, we'd definitely need to be careful about this in cases where shadowDOM is not supported.

components/list/list-item-checkbox-mixin.js
dbatiste: Thanks, I see.

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,47 @@
...
+				${ this._renderDragHandle ? this._renderDragHandle() : '' }
+				${ this._renderCheckbox() }
+				<div class="d2l-list-item-content"></div>
```
dbatiste, line 40: Are these `div`s meant to contain elements - i.e. they would either define a slot, or other elements?

components/list/demo/list-item-sample.js
zinabat: Correct, the conditional is meant to be removed once implemented. This is only meant as a hint for developers.

components/list/demo/list-item-sample.js
zinabat: Absolutely. In fact, this was named this way intentionally because the next PR will remove it.

components/list/demo/list-item-sample.js
zinabat: These dives are meant to define slots, which currently don't exist.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,85 @@
...
+			<div slot="control">
+				<input
+					id="${this._checkBoxId}"
```
dbatiste, line 75: This should be `this._checkboxId`.  Currently the `id` is being rendered with "undefined" value.

zinabat: Good catch. Fixed

components/list/demo/list-item-sample.js
dbatiste: I see.  Currently the `div[slot="control-action"]` element from the checkbox mixin is rendered on top of these (because it has `z-index: 2;`) so if these contained something interactive, these divs/slots, or the elements within would need a `z-index > 2` to put them on top of the stack.

zinabat: This is expected behaviour according to the design document. Anything interactive should be inside of one of the `action` slots or `actions`.

dbatiste: I am a little confused by this, perhaps due to the names of the classes.  I think it might be worthwhile to see the sample with the slots and an action that works (on top of the clickable area for toggling the checkbox).

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,85 @@
...
+			key: { type: String, reflect: true },
+			selectable: {type: Boolean },
+			selected: { type: Boolean, reflect: true }
+		};
+	}
+
+	// TODO: Remove grid styles after list-item-generic is created
+	static get styles() {
+		return [ checkboxStyles, css`
+		[slot="control"] {
```
dbatiste, line 20: If I understand correctly, use of the checkbox mixin requires that the component defines the named gridlines.  I think I'd prefer it if the checkbox mixin didn't need to know anything about the layout, and perhaps this could be done by the mixin defining two render helpers: one that renders the checkbox, and one that renders the toggle area, and let the consuming component compose the layout however it needs to.
```
renderCheckbox() {
    return html`
        <input
            id="${this._checkboxId}"
            class="d2l-input-checkbox"
            @change="${this._handleCheckboxChange}"
            type="checkbox"
            .checked="${this.selected}"
            ?disabled="${this.disabled}">
    `;
}

renderCheckboxToggleContainer(inner) {
    return html`<div @click="${this._handleCheckboxActionClick}">${inner}</div>`
}
```

components/list/demo/list-item-sample.js
zinabat: Fair enough - I've altered the sample component to make this clearer and added a button to the `actions` slot.
Screen Shot 2020-04-30 at 4 42 09 PM

components/list/list-item-checkbox-mixin.js
zinabat: I agree this is a better approach - will update accordingly

components/list/demo/list-item-sample.js
```diff
@@ -0,0 +1,52 @@
...
+		`];
+	}
+	render() {
+		return html`
+			<div class="list-grid">
+				${ this._renderDragHandle ? this._renderDragHandle() : html`
+					<div>=</div>
+				` }
+				${ this._renderCheckbox() }
+				<div slot="content">
```
dbatiste, line 43: Thanks Zina.  This makes more sense to me now.  I _think_ the `slot="content"` should become `class="d2l-list-item-content"` though, since there's no named slot `content`.

zinabat: There will be a named content slot similar to the `control` and `control-action` slots. See this PR for context: https://github.com/BrightspaceUI/core/pull/506/files

I can change it back if it makes more sense to do so in this PR

components/list/list-item-checkbox-mixin.js
zinabat: Updated and changed to two separate methods so that the consumer is in control of the layout only.

components/list/demo/list-item-sample.js
dbatiste: Ok I see how this is intended to eventually be used.  I think it would have been ok if we'd included the generic "item" in this PR, but let's get this in.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,76 @@
...
+				@change="${this._handleCheckboxChange}"
+				type="checkbox"
+				.checked="${this.selected}"
+				?disabled="${this.disabled}">
+			` : nothing;
+	}
+
+	_renderCheckboxAction(inner) {
+		return this.selectable ? html`
+			<div @click="${this._handleCheckboxActionClick}" class="checkbox-action">${inner}</div>
```
dbatiste, line 73: Nit: let's make the class `d2l-checkbox-action`.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,76 @@
...
+			key: { type: String, reflect: true },
+			selectable: {type: Boolean },
+			selected: { type: Boolean, reflect: true }
+		};
+	}
+
+	// TODO: Remove grid styles after list-item-generic is created
+	static get styles() {
+		return [ checkboxStyles, css`
+			.checkbox-action { height: 100%}
```
dbatiste, line 20: Nit: missing semi-colon, whitespace.

components/list/list-item-checkbox-mixin.js
```diff
@@ -0,0 +1,76 @@
...
+	static get properties() {
+		return {
+			disabled: {type: Boolean },
+			key: { type: String, reflect: true },
+			selectable: {type: Boolean },
+			selected: { type: Boolean, reflect: true }
+		};
+	}
+
+	// TODO: Remove grid styles after list-item-generic is created
```
dbatiste, line 17: Looks like this comment is no longer needed here?

zinabat: Nope, grid styles aren't meant to live here so this is needed

components/list/demo/list-item-sample.js
zinabat: Yeah, I think in retrospect I should have done the grid PR first.

m6jones: I think the grid allows for us to use classes instead of slots. So one default slot and people can pick from the predefined classes to place things. If that make sense.

PR #498, opened 2020-04-22
a11y: add "accessibility" sections to other READMEs

margaree:


components/dropdown/dropdown-context-menu.js
```diff
@@ -34,7 +34,7 @@ class DropdownContextMenu extends DropdownOpenerMixin(VisibleOnAncestorMixin(Lit
 
 	render() {
 		return html`
-			<d2l-button-icon aria-label=${this.text} ?disabled=${this.disabled} icon="tier1:chevron-down" text=${this.text} ?translucent=${this.translucent}>
+			<d2l-button-icon ?disabled=${this.disabled} icon="tier1:chevron-down" text=${this.text} ?translucent=${this.translucent}>
```
margaree, line 37: FYI @AllanKerr I removed the `aria-label` here and in `dropdown-more` because `text` on `button-icon` also adds the `aria-label` to the `button`. Let me know if there were additional reasons why this `aria-label` was added.

components/calendar/README.md
```diff
@@ -20,6 +20,14 @@ The `d2l-calendar` component can be used to display a responsively sized calenda
 - `selected-value` (String): Currently selected date. This should be in ISO 8601 calendar date format (`YYYY-MM-DD`) and should be localized to the user's timezone (if applicable).
 - `summary` (String): Summary of the calendar for accessibility.
 
+**Accessibility:**
```
dbatiste, line 23: Do we want to switch the rest of the API doc to use tables as well?

margaree: I would be okay with that (and can make that change over the next bit) if others are interested.

PR #500, opened 2020-04-22
Button Icon Tooltip Integration

AllanKerr:
# Changes
- Update `d2l-button-icon` to show a tooltip with the `text` instead of setting the `title` attribute on the `button`

### DIFF
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -1,8 +1,10 @@
 import '../icons/icon.js';
+import '../tooltip/tooltip.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { VisibleOnAncestorMixin, visibleOnAncestorStyles } from '../../mixins/visible-on-ancestor-mixin.js';
 import { ButtonMixin } from './button-mixin.js';
 import { buttonStyles } from './button-styles.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
@@ -13,7 +15,8 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 			hAlign: { type: String, reflect: true, attribute: 'h-align' },
 			icon: { type: String, reflect: true },
 			text: { type: String, reflect: true },
-			translucent: { type: Boolean, reflect: true }
+			translucent: { type: Boolean, reflect: true },
+			_buttonId: { type: String }
 		};
 	}
 
@@ -110,12 +113,21 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 		];
 	}
 
+	constructor() {
+		super();
+		this._buttonId = getUniqueId();
+	}
+
 	render() {
+		const tooltipType = this.text && this.ariaLabel ? 'descriptor' : 'label';
+		const tooltipText = this.text || this.ariaLabel;
+		const label = this.text ? this.ariaLabel : undefined;
 		return html`
 			<button
+				id=${this._buttonId}
 				aria-expanded="${ifDefined(this.ariaExpanded)}"
 				aria-haspopup="${ifDefined(this.ariaHaspopup)}"
-				aria-label="${this.ariaLabel ? this.ariaLabel : ifDefined(this.text)}"
+				aria-label="${ifDefined(label)}"
 				?autofocus="${this.autofocus}"
 				class="d2l-label-text"
 				?disabled="${this.disabled}"
@@ -126,9 +138,9 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 				?formnovalidate="${this.formnovalidate}"
 				formtarget="${ifDefined(this.formtarget)}"
 				name="${ifDefined(this.name)}"
-				title="${ifDefined(this.text)}"
 				type="${this._getType()}">
 				<d2l-icon icon="${ifDefined(this.icon)}" class="d2l-button-icon"></d2l-icon>
+				${tooltipText ? html`<d2l-tooltip for=${this._buttonId} for-type=${tooltipType} disable-focus-lock>${tooltipText}</d2l-tooltip>` : null}
 		</button>
 		`;
 	}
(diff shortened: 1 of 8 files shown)


components/button/button-icon.js
```diff
@@ -126,9 +128,9 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 				?formnovalidate="${this.formnovalidate}"
 				formtarget="${ifDefined(this.formtarget)}"
 				name="${ifDefined(this.name)}"
-				title="${ifDefined(this.text)}"
 				type="${this._getType()}">
 				<d2l-icon icon="${ifDefined(this.icon)}" class="d2l-button-icon"></d2l-icon>
+				${this.text ? html`<d2l-tooltip for="target" disable-focus-lock>${this.text}</d2l-tooltip>` : null}
```
AllanKerr, line 133: The tooltip needs to be inside the `button`. If it isn't then the `offsetParent` changes from the `d2l-button-icon` to a different element when the VOA transition finishes causing it to be positioned incorrectly. Because the `button` is positioned `relative`, this doesn't happen. I chose this rather than positioning the `d2l-button-icon` itself `relative`.

AllanKerr, line 133: I added `disable-focus-lock` because a common pattern in the LMS involves having multiple `d2l-button-icon`s in a row, all with tooltips. `disable-focus-lock` means the tooltip will close when the user mouses off so they don't get visibly overlapped tooltips.

dbatiste, line 133: Continuing conversation re API now that we're actually doing this: the current approach limits consumers to text-only for the tooltip contents.  This would not suffice for cases like GBL (which I know has it's own style applied to native button). Is Design ok with this limitation in the core components?  We could surface the slot later, but the API starts to get pretty dirty so I tend to prefer the simplicity of the text, but it is a significant difference nonetheless.  

I anticipate this approach (of baking the tooltip into the core component) to be used elsewhere, for example, potentially the card-footer-link, and I think the API should be the same.  That is, if we go text-only here, I'd expect text only there too.  The dropdown openers are another scenario where we would make use of this concept (via consuming `d2l-button-*`), and if a slot were surfaced, it would affect it's API too. 

Anyway, just to call this out/validate that limitation before we get too far along this path.

dlockhart: Good to call out for sure. Since this is really just an extension of the button's label, and previously was stored in the `title` attribute, my preference would be for this to be text-only. I don't know what it would mean for rich HTML to be inside a button's label, but I'm pretty sure we don't want it.

components/button/button-icon.js
```diff
@@ -113,9 +114,10 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 	render() {
 		return html`
 			<button
+				id="target"
 				aria-expanded="${ifDefined(this.ariaExpanded)}"
 				aria-haspopup="${ifDefined(this.ariaHaspopup)}"
-				aria-label="${this.ariaLabel ? this.ariaLabel : ifDefined(this.text)}"
+				aria-label="${ifDefined(this.ariaLabel)}"
```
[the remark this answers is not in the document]
AllanKerr: Added `for-type` to the tooltip to change between label and descriptor. `display: none` is handled by `aria-labelledby` and `aria-describedby` so they'll automatically be visible: https://w3c.github.io/aria-practices/#naming_with_aria-labelledby
> It incorporates content from elements regardless of their visibility, i.e., it even includes content from elements with the HTML hidden attribute, CSS display: none, or CSS visibility: hidden in the calculated name string.

components/tooltip/test/tooltip.test.js
```diff
@@ -4,13 +4,20 @@ import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
 const basicFixture = html`
 	<div>
-		<div id="implicit-target" tabindex="-1">
+		<div id="implicit-target" tabindex="-1" role="button">
```
dlockhart, line 7: What's the purpose of the `role` on the implicit target?

helpers/focus.js
```diff
@@ -163,9 +163,9 @@ export function getPreviousFocusableAncestor(node, includeHidden, includeTabbabl
 	return null;
 }
 
-export function isFocusable(node, includeHidden, includeTabbablesOnly) {
+export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDisabled) {
```
dlockhart, line 166: In retrospect, we probably should have done this with an `options` parameter with smart defaults. Then you could specify one without needing to specify the rest.

components/tooltip/test/tooltip.test.js
AllanKerr: It was added without it because it generated warning logs in the test output because the tooltip was used in a non-accessible way 😅

components/dialog/dialog-mixin.js
```diff
@@ -281,6 +281,9 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		// native dialog backdrop does not prevent body scrolling
 		this._bodyScrollKey = preventBodyScroll();
 
+		if (document.activeElement) {
```
dbatiste, line 284: Does this happen with both the native dialog and custom?  Do you know what is placing the focus on the button... we should not be trying to place focus until below... but it will be one frame from this code.

dbatiste: At the top of the demo page there is a commented line of code if you wish to test with the custom dialog impl in Chrome.

AllanKerr: Its caused by the native dialog. When running the custom dialog its fine.

AllanKerr: The other option I considered here was increasing the `tabindex` of the close button so that its considered last which also fixes this issue with native dialogs.

dbatiste: I'd lean away from the `tabindex` approach since it could interfere with the cycling of the focus within the dialog.  I think the `blur` is fine, but perhaps add a comment similar to the one above... `\\ native dialog... tooltip... ` since it won't be otherwise clear why this is needed.

AllanKerr:
Closing because we are putting integration on hold as a result of the issues mentioned [here](https://github.com/BrightspaceUI/core/issues/508). The tooltip improvements have been split off into a separate PR [here](https://github.com/BrightspaceUI/core/pull/509) because they still provide value.

PR #501, opened 2020-04-23
input-date-time: Simplify utc to local and back conversion

margaree:


components/inputs/input-date-time.js
```diff
@@ -128,10 +125,8 @@ class InputDateTime extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			this.value = '';
 			this._parsedTime = '';
 		} else {
-			const time = this._parsedTime ? parseISOTime(this._parsedTime) : this.shadowRoot.querySelector('d2l-input-time').getTime();
-			const date = parseISODate(this._parsedDate);
-			const converted = convertLocalToUTCDateTime(Object.assign(date, time));
-			this.value = formatDateTimeInISO(converted);
+			const time = this._parsedTime ? this._parsedTime : this.shadowRoot.querySelector('d2l-input-time').value;
```
dlockhart, line 128: Was this just a bug where we were double-parsing the time before?

components/inputs/docs/input-date-time.md
```diff
@@ -112,82 +112,110 @@ To make your usage of `d2l-input-date-time` accessible, use the following proper
 
 The `input-date` and `input-time` components do not handle timezone and so require the input to be in the user's timezone, which corresponds to the timezone in the `data-timezone` attribute on the `html` element.
```
dlockhart, line 113: I don't think we should mention the `data-timezone` attribute... really it's an internal thing that nobody should know about except the platform components that deal with it. Instead, you could just mention "... in the user's timezone as specified in their account settings."

components/inputs/docs/input-date-time.md
```diff
@@ -112,82 +112,110 @@ To make your usage of `d2l-input-date-time` accessible, use the following proper
...
-import {convertUTCToLocalDateTime} from '@brightspace-ui/intl/lib/dateTime.js';
-import {formatDateInISO, parseISODateTime} from '@brightspace-ui/core/helpers/dateTime.js';
+import {getLocalDateFromUTCDateTime, getLocalTimeFromUTCDateTime, getUTCDateTimeFromLocalDateTime} from '@brightspace-ui/core/helpers/dateTime.js';
 
 const UTCDateTime = '2018-03-01T12:20:00.000Z';
-const UTCDateTimeObject = parseISODateTime(UTCDateTime);
-const localDateTime = convertUTCToLocalDateTime(UTCDateTimeObject);
-const value = formatDateInISO(localDateTime); // use as value in input-date
+const localTime = getLocalTimeFromUTCDateTime(UTCDateTime); // for later conversion back to UTC
+let value = getLocalDateFromUTCDateTime(UTCDateTime); // use as value in input-date
```
dlockhart, line 126: This example is interesting, where they have a time and want to preserve it because the date-input will obliterate it. I'm wondering if this actually matters in real life... where they'd have a UTC date with a time, want to throw that into a date input but then hold onto the time even if the date changes? I can't really see anyone wanting that? Seems like we could simplify the examples a lot by just not worrying about this edge case?

components/inputs/docs/input-date-time.md
```diff
@@ -112,82 +112,110 @@ To make your usage of `d2l-input-date-time` accessible, use the following proper
...
+		const isoFormattedUTCDateTime = getUTCDateTimeFromLocalDateTime(value, this._initialLocalTime);
+	}
+}
+customElements.define('d2l-my-element', MyElement);
 ```
 
-### Convert Local Date to UTC Date/Time
+### input-time: Convert UTC Date/Time to Local Timezone and Back to UTC
+
+To convert a UTC date/time (presumably in ISO 8601 format, `YYYY-MM-DDTHH:mm:ss.sssZ`, adjust if not) to a time in the user's timezone for use in `input-time`, use `getLocalTimeFromUTCDateTime` from `helpers/dateTime.js`, which will return the time in the user's timezone in the ISO 8601 date format `hh:mm:ss`.
```
dlockhart, line 171: Same comment here I think. I could be wrong, but I think in most cases where folks are dealing with only a time, they'll just have the `HH:mm:ss.sssZ` bit and even if they have the date part they don't care about preserving it.

PR #502, opened 2020-04-23
d2l-skeleton Component(?)

ChrisKraljevicD2L:
I realize that this "component" is a bunch of css applied to a div, and I completely understand if this can just exist as a sass mixin and `___-styles.js` file. But I did think that the [design.d2l specs for skeletons](http://design.d2l/patterns/skeleton-loading/) should be committed somewhere where anyone can use them.

skeleton-multi-line

(this gif is not representative of what skeletons look like in a browser, for some reason LICEcap doesn't record it well)


m6jones:
So this looks good. The styles should also not assume a div is being used. I tend to use inline svg to block out the areas. So add a fill and that would be awesome.

I am concerned that this doesn't match design.d2l. http://design.d2l/patterns/skeleton-loading/

dlockhart:
I spoke with Chris in Slack, but Gaudi's going to do some work to investigate how folks are using skeletons today, and as well look into the design and make sure whatever ends up in core meets everyone's needs. This PR is definitely a great first step!

Currently this is relatively high on our backlog, but we want to close out the date/time and validation work thats in flight first.

m6jones:
Can I be involved in the discussion as I have a lot of input on this :D
I have stories to add skeletons to the list component at the end of the drag n drop and want them to be aligned.

m6jones:
List item skeleton: https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js#L663

dlockhart:
Absolutely, yeah!

ChrisKraljevicD2L:
Hey Maya. I haven't worked with SVGs before unfortunately. I'll do my best to support svg skeletons and please let me know if what I put up needs to change.

I also see how this doesn't match design.d2l. I have the shimmer animation as a horizontal pulse instead of what's actually in the spec. I'll fix that :+1:

components/skeleton/skeleton-styles.js
```diff
@@ -0,0 +1,57 @@
...
+		display: block;
+		border-radius: 4px;
+		background-color: var(--d2l-color-sylvite);
+		fill: var(--d2l-color-sylvite);
+		overflow: hidden;
+		position: relative;
+	}
+
+	.d2l-skeleton[hidden] {
+		display: none;
```
apalaniuk, line 55: Though `!important` is something you typically don't want to use, hiding elements using `display: none` is one of those cases, since it ensures that other `display` values that are applied to the same elements don't override this.

PR #506, opened 2020-04-24
Add generic list item component to allow default grid

zinabat:
### **Context**
[US115379](https://rally1.rallydev.com/#/detail/userstory/382419013332?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#Generic-List-Item)
Introduces the `list-item-generic-layout` simple component to be used for defining default grid behaviour. The new grid includes seven slots. Of these, they are split into two groups:

**Group 1: Content**
67841605-ac02-4bf6-b86d-f585fff6d8f3

1. `outside-control` - For controls such as the drag 'n' drop handle.
2. `control` - Control defined within the component boundaries, e.g., the checkbox
3. `content` - Content, such as that defined within a `list-item-content` component
4. `actions` - Actions, such as buttons or other top-level clickable behaviour that sit above `content-action`

**Group 2: Action areas**
03766529-aa58-4f11-80fd-b21b11642273

1. `outside-control-action` (a) - The first layer of clickable area, which encapsulates the component
2. `control-action` (b) - Second layer of clickable area, which includes the length of the component up to outside-control-action
3. `content-action` (c) - Third layer of clickable area, which includes only the content and action portions

**Result**
Screen Shot 2020-05-01 at 2 49 20 PM

### **Quality**

- Manually tested
  - [x] Edge (Legacy) 
  - [x] Chrome
  - [x] Safari
  - [x] Firefox

- To test, pull the branch and run `npm start`, then visit http://localhost:8000/components/list/demo/list-item-sample.html


components/list/list-item-generic.js
```diff
@@ -0,0 +1,80 @@
...
+			::slotted([slot="control"]),
+			::slotted([slot="content"]),
+			::slotted([slot="actions"]) {
+				grid-row: 1 / 2;
+			}
+			::slotted([slot="outside-control"]) {
+				grid-column: outside-control-start / outside-control-end;
+			}
+
+			::slotted([slot="control"]) {
```
m6jones, line 27: I was thinking that this would be `::slotted(.d2l-list-item-area-control)` or something similar. So anything with that class gets place in the slot. Let me know if this would cause other problems that I do not see.

m6jones: I could see this causing the problems for focusing? Which might make slots better?

components/list/list-item-checkbox-mixin.js
```diff
@@ -71,6 +82,7 @@ export const ListItemCheckboxMixin = superclass => class extends superclass {
 	_renderCheckboxAction(inner) {
 		return this.selectable ? html`
 			<div @click="${this._handleCheckboxActionClick}" class="d2l-checkbox-action">${inner}</div>
+>>>>>>> master
```
m6jones, line 85: Merge conflict stuff got in.

components/list/demo/list-item-sample.js
```diff
@@ -10,55 +11,25 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
...
-				z-index: 4;
-			}
-		`];
+		return [ (super.styles ? super.styles : css``)];
 	}
+
 	render() {
 		return html`
-			<div class="list-grid">
+			<d2l-list-item-generic>
```
m6jones, line 20:
```
classMapRender = ['d2l-list-item-area-control': true] //default value but overridable
classMapAction = ['d2l-list-item-area-control-action': true] //default value but overridable
<d2l-list-item-generic>
      ${renderCheckbox(classMapRender, classMapAction)} 
</d2l-list-item-generic>
```
Then magic happens and it just works.

m6jones: Granted the way you are doing it has similar results.

The reason I wanted the generic list host to define the grid is so the user could override it if they wanted too. Which probably makes things difficult with focus order stuff. :D Sorry new information that might not be required.

components/list/list-item-generic.js
```diff
@@ -0,0 +1,80 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+export class ListItemGeneric extends RtlMixin(LitElement) {
```
dbatiste, line 4: I hesitate to chime before this is ready alas, earlier the better I think...  

Currently, I think this "generic" component is simply to define a generic layout.  It is not a list item, rather, it is used by a list-item implementation.  If that remains to be the case, then I think this should be `ListItemGenericLayout` / `d2l-list-item-generic-layout`.  I was mulling over the idea of whether this should actually just be a mixin, and would be happy to hear thoughts on that too.

Besides that, it looks like this class currently does not need to be exported.

components/list/list-item-generic.js
zinabat: So yeah, this was sort of my rational for doing it this way. This means that no matter what order the user defines their areas in, the focus order is always the same because the DOM structure is the same.

components/list/list-item-generic.js
zinabat: @m6jones What do you think about this?

components/list/demo/list-item-sample.js
zinabat: My thought was that if the user wants to override the default layout, they would just need to implement a new version of `list-item-generic` to do so.

I guess the question is whether we want to force focus order with slots or not. If not, I'll need to think of another solution for that problem.

m6jones: I would prefer that no one makes a new version of `list-item-generic`

components/list/list-item-generic.js
m6jones: So, the new name YES! I always hated the name and hoping that someone later would make a better one. In the design doc I avoiding naming it.

So this doesn't need to be a mixin as it can do everything it requires without being one. Which makes it easier to use and edit. So the decision which I mulled over for a while comes down to those key factors. 

The only pro for a mixin was to pre optimize by minimizing the dom depth. Didn't seem worth it. 

Side note @dbatiste, you comments are always welcome before it is ready. earlier the better is good. The goal with draft PR approval from our team then making it a real PR for your approval is so we can save your time when you are more busy as hopefully we get at least all the small things. If you have time to comment on the draft / want to take part it will just make things better.

components/list/list-item-generic.js
m6jones: Would allowing the user define the focus order be good? I'm not sure. Since it as easy as place in order of how you want focus to work it might be good. The user in this case will be new list items renderers which should be very few and far between.

components/list/demo/list-item-sample.js
zinabat: That makes sense. I think I'm not super clear on the magic that's supposed to happen in the `renderCheckbox` method. Before, I had it so that the checkbox control and action areas were rendered automagically in that method, but @dbatiste mentioned that it would be beneficial for users implementing a new list item component to have separate methods for doing that. 

Are you saying that the user would have control of the two areas via the class maps fed to the method? And if they wanted to change `d2l-list-item-area-control` to `some-custom-class` they would do that through the class map?

How is this different from a user just changing their custom component itself, e.g.,
```
<div class="some-custom-control" slot="control">${ this._renderCheckbox() }</div>
<div class="some-custom-action" slot="control-action">${ this._renderCheckboxAction() }</div>
```
or targetting with css?

components/list/list-item-generic.js
```diff
@@ -0,0 +1,73 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+
+class ListItemGeneric extends LitElement {
+
+	static get styles() {
+		return css`
+			:host {
```
m6jones, line 7: This will need `position: relative;` so that it forces the `z-index` of the elements to fall to 0 outside of this dom element.

components/list/list-item-generic.js
```diff
@@ -0,0 +1,73 @@
...
+
+			::slotted([slot="control"]) {
+				grid-column: control-start / control-end;
+			}
+
+			::slotted([slot="content"]) {
+				grid-column: content-start / content-end;
+			}
+
+			::slotted([slot="actions"])  {
```
m6jones, line 34: This one might be good to do add `position:relative` so then anything in this will have a z-index relative to the one defined here. We should make sure this is actually how it works...

zinabat: I'll double check that.

m6jones: Since this is grid z-index this might not matter :D

components/list/list-item-generic.js
m6jones: Grid might make us not have to do this :D

components/list/demo/list-item-sample.js
m6jones: I don't think it is different.

m6jones: As long as `nothing` works.

m6jones: The current way is probably more intuitive which is better.

components/list/list-item-generic.js
zinabat: Yeah after playing around with this it appears this isn't necessary. 🎉

components/list/list-item-generic.js
zinabat: If the user is able to define their own focus order, doesn't that sort of make [US115380](https://rally1.rallydev.com/#/detail/userstory/382436537132?fdp=true) a moot point?

The way we've written these, it means users can only define focus order within action areas, but the action areas themselves have a predefined focus order.

m6jones: Yup! you are right :)

zinabat: So I guess it's about how much control do we want to give here: more v less. I'm fine with either approach TBH but I think giving less forces an accessibility standard that's good, so I would argue forcing focus order is better in that case.

components/list/demo/list-item-sample.js
zinabat: I will probably always be a fan of more intuitive vs. less markup <- (someone quote this for me in 2025)

But I am totally cool with whatever is easiest for our users, in this case developers.

m6jones: More intuitive is better for devs so I say leave it the way you did it.

dbatiste:
A couple minor things in the description worth knowing:

Re the note about `:host-context`: this is not supported by Firefox either.  It was supported cross browser previously only for Polymer components, but from what I understand, there is no consensus on it, and it may never be supported.

Re the note about testing with `polymer serve`: wile devs may have the `polymer-cli` installed globally, `core` now uses `es-dev-server`.  Devs can/should just start the server with `npm start`.

components/list/list-item-checkbox-mixin.js
```diff
@@ -7,17 +7,22 @@ export const ListItemCheckboxMixin = superclass => class extends superclass {
...
 
-	// TODO: Remove grid styles after list-item-generic is created
 	static get styles() {
 		return [ checkboxStyles, css`
-			.d2l-checkbox-action { height: 100%; }
+			.d2l-checkbox-action {
+				height: 100%;
+				cursor: pointer;
+			}
+			:host-context([disabled]) .d2l-checkbox-action {
```
dbatiste, line 23: I think this should/could just be:
```
:host([disabled]) .d2l-checkbox-action {
    ...
}
```

Aside from that, I don't think we can use `:host-context` going forward.  There's a bit of discussion about in the `webcomponentsjs` repo, and while I haven't searched for the actual meeting notes on it, as far as I know, there is still no consensus or intention on Apple's part to support it.

components/list/list-item-generic-layout.js
```diff
@@ -0,0 +1,73 @@
...
+			:host {
+				display: grid;
+				grid-template-columns:
+					[start outside-control-start] minmax(0, min-content)
+					[control-start outside-control-end] minmax(0, min-content)
+					[control-end content-start] auto
+					[content-end actions-start] auto
+					[end actions-end];
+			}
+			::slotted([slot="outside-control"]),
```
dbatiste, line 16: Just one thing to be aware of here, I _think_ this selector would not work in IE11.  It would need to be combined with another selector, such as `.some-class ::slotted([slot="outside-control"])`.  Not sure the root cause, but I think it's something to do with the polyfills.

dbatiste: ^ probably don't need to worry about this at this point, but something to be aware of.

components/list/list-item-checkbox-mixin.js
zinabat: Good to know - I will just pass the disabled state down to the layout component from the parent in that case.

components/list/list-item-generic-layout.js
zinabat: So, I tested this in IE11. The selector itself works because it's a direct child. However, IE11 doesn't support `grid-column` and so the grid does not work properly. I spoke to Maya about this and the plan is to only roll this out for drag 'n' drop before June, then only look at changing the current `list-item` component after we drop IE11 support.

components/list/list-item-checkbox-mixin.js
dbatiste: I don't think it will need to be passed though, since `disabled` is a property of a mixin, and therefor on the `host`.

PR #507, opened 2020-04-27
input-text: use for/id for labeling

margaree:
This fixes of the problem of over-reading the placeholder on input.
Hover/click behavior seems to be working the same as before when hovering/clicking on the label text (checked chrome, legacy edge, firefox, safari).

The only other place I saw where we use the `label` wrapped around the `input` is `input-time` and I am hesitant to change that one due to it being a bit more complex, and it not using placeholders so it would not have the bug anyways.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -1,5 +1,6 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { classMap } from 'lit-html/directives/class-map.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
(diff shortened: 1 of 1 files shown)


components/inputs/input-text.js
```diff
@@ -167,10 +170,10 @@ class InputText extends RtlMixin(LitElement) {
 		`;
 		if (this.label && !this.labelHidden) {
 			return html`
-				<label>
+				<label for="${this._inputId}">
 					<span class="d2l-input-label">${this.label}</span>
```
dlockhart, line 174: Now that the label doesn't wrap the input, I believe the `d2l-input-label` class can be applied to the label itself, and the span can go away.

PR #510, opened 2020-04-27
input-date: add empty-state-text

margaree:
Changes:
- Add `empty-state-text` (e.g., "No Start Date") which is visible when specified and when there is no selected value and the user's focus is not within the text box. When the user's focus is the text box, the placeholder (short date format) should appear.
- Width should be based on the widest potential content of the input, and is the widest of the placeholder (short date format), the `empty-state-text`, or `2020-12-20` formatted to the language's short date format.

Notes:
- To me the width piece feels hacky and I'm wondering if there is anywhere that we've done this before that is different/better.
- Are there any other name suggestions besides `empty-state-text`? The property is meant to portray that this is text that is shown to the user when there is no value selected and the field is not required and it is okay for them to submit the form with this field left blank.


components/inputs/input-date.js
```diff
@@ -20,9 +20,12 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	static get properties() {
 		return {
 			disabled: { type: Boolean },
+			emptyStateText: { type: String, attribute: 'empty-state-text'},
```
dlockhart, line 23: I'm OK with this name, or `empty-text` which I think is a term we've used before whenever we have no search results or no items in a list.

components/inputs/input-date.js
```diff
@@ -212,12 +222,35 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
+		text.innerHTML = formatISODateInUserCalDescriptor('2020-12-20');
+		const contentWidth = text.clientWidth;
+
+		let emptyStateWidth = 0;
+		if (this.emptyStateText) {
+			text.innerHTML = this.emptyStateText;
+			emptyStateWidth = text.clientWidth;
+		}
+		const width = Math.max(placeholderWidth, contentWidth, emptyStateWidth);
+		this._contentWidth = width + 10;
```
dlockhart, line 250: What's the `+ 10` for? If you're seeing a mismatch, it could be because our inputs have `letter-spacing: 0.2rem`...

components/inputs/input-date.js
```diff
@@ -158,17 +161,24 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 	}
 
 	render() {
 		return html`
 			<d2l-dropdown ?disabled="${this.disabled}" no-auto-open>
 				<d2l-input-text
+					@blur="${this._handleInputTextBlur}"
 					@change="${this._handleChange}"
 					class="d2l-dropdown-opener"
+					content-width="${ifDefined(this._contentWidth)}"
```
dlockhart, line 178: I'm not a huge fan of having to expose `content-width` on the text input... is there a way we could just add the width of the slot to `_contentWidth` and then use `<d2l-input-text style="max-width: X">`?

components/inputs/input-date.js
```diff
@@ -212,12 +222,35 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
+		document.body.appendChild(text);
+		text.style.fontSize = '0.8rem';
+		text.style.width = 'auto';
+		text.style.position = 'absolute';
+
+		// in some languages (e.g., fr) placeholderWidth is bigger, in others (e.g., zh) contentWidth is bigger
+		text.innerHTML = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+		const placeholderWidth = text.clientWidth;
+		text.innerHTML = formatISODateInUserCalDescriptor('2020-12-20');
+		const contentWidth = text.clientWidth;
```
dlockhart, line 242: Definitely hacky. I can't think of a way that would accomplish the same thing and be as accurate. One idea I had initially would be to use the font-size & letter-spacing to come up with a formula that for a given number of characters you could guess a width. It incorrectly assumes that all characters are the same width though.

One optimization to this might be to only calculate it once for the first date-input... since having many of them on the page would add a bunch of divs, then remove them.

components/inputs/input-date.js
```diff
@@ -158,17 +161,24 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 		this.addEventListener('d2l-localize-behavior-language-changed', () => {
 			this._dateTimeDescriptor = getDateTimeDescriptorShared(true);
 			this.requestUpdate();
+			this._getContentWidth();
 		});
+
+		this._formattedValue = this.emptyStateText ? this.emptyStateText : '';
```
dlockhart, line 167: How does this work for screen readers? Do the focus & blur events actually fire when the virtual cursor is being used such that it reads the placeholder instead of the empty text?

components/inputs/input-date.js
margaree: Without it we get the `...` at the end instead of the final letters/numbers. Is there a better alternative?

components/inputs/input-date.js
margaree: The optimization would probably usually be okay, but we could run into problems if a later one is using the `empty-state-text` and it's wider

components/inputs/input-date.js
margaree: That sounds promising - I'll give it a try

components/inputs/input-date.js
```diff
@@ -212,12 +222,35 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 
+	_getContentWidth() {
+		const text = document.createElement('div');
+		document.body.appendChild(text);
+		text.style.fontSize = '0.8rem';
+		text.style.width = 'auto';
+		text.style.position = 'absolute';
+
+		// in some languages (e.g., fr) placeholderWidth is bigger, in others (e.g., zh) contentWidth is bigger
+		text.innerHTML = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
```
dlockhart, line 239: If we can use `innerText` instead of `innerHTML` it avoids possible HTML-injection attacks, which I'd especially be worried about coming from `emptyStateText`.

components/inputs/input-date.js
dlockhart: I'd just like to know the root cause and why `10` works but `8` or `9` doesn't. If the user has a larger font size, do you need a larger value here?

margaree: So it turned out to be a combination of needing to use `letter-spacing`, adding border width, and needing to do some rounding up. The mysterious value has been eliminated🎉

components/inputs/input-date.js
margaree: So it mostly doesn't get read (voiceover seems to sometimes but I don't think NVDA ever did). Talking more to Jeff about it, he'd like it included using `aria-describedby`(when `empty-text` and no value selected). This seems like it would involve adding something like `description` to `input-text` in order to add an offscreen span with the text and `aria-describedby` to the `input` in `input-text`. Thoughts or concerns or alternatives to this?

components/inputs/input-date.js
```diff
@@ -65,72 +66,72 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 			'en': {
 				clear: 'Clear',
-				openInstructions: 'Arrow down or press enter to access mini-calendar',
+				openInstructions: 'Use date format {format}. Arrow down or press enter to access mini-calendar.',
 				setToToday: 'Set to Today'
 			},
 			'es': {
 				clear: 'Borrar',
-				openInstructions: 'Arrow down or press enter to access mini-calendar',
+				openInstructions: 'Use date format {format}. Arrow down or press enter to access mini-calendar.',
```
dlockhart, line 89: Hmm, let's not forget to get these terms translated.

components/inputs/input-date.js
```diff
@@ -155,25 +159,35 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
+		this.addEventListener('blur', this._handleBlur);
 		this.addEventListener('d2l-localize-behavior-language-changed', () => {
 			this._dateTimeDescriptor = getDateTimeDescriptorShared(true);
-			this.requestUpdate();
+			this._getContentWidth();
 		});
+
+		this._formattedValue = this.emptyText ? this.emptyText : '';
+		const fontsPromise = document.fonts ? document.fonts.ready : Promise.resolve();
+		await Promise.resolve(fontsPromise);
```
dlockhart, line 170: Would this not do the same?

```suggestion
		await document.fonts ? document.fonts.ready : Promise.resolve();
```

components/inputs/input-date.js
```diff
@@ -155,25 +159,35 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 					?disabled="${this.disabled}"
+					@focus="${this._handleInputTextFocus}"
 					@keydown="${this._handleKeydown}"
 					label="${ifDefined(this.label)}"
 					?label-hidden="${this.labelHidden}"
 					@mouseup="${this._handleMouseup}"
-					placeholder="${(this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase()}"
-					title="${this.localize('openInstructions')}"
+					placeholder="${shortDateFormat}"
+					style="${styleMap({maxWidth: inputTextWidth})}"
```
dlockhart, line 189: With `max-width` set, what's actually causing it to become wide? I guess it's the text input's `width: 100%`?

components/inputs/input-date.js
```diff
@@ -211,12 +225,46 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
-				this._formattedValue = this.value ? formatISODateInUserCalDescriptor(this.value) : '';
+			if (prop === '_dateTimeDescriptor' || prop === 'value') {
+				this._setFormattedValue();
 			}
 		});
 	}
 
+	_getContentWidth() {
+		const text = document.createElement('div');
+		document.body.appendChild(text);
```
dlockhart, line 236: Nit: you may want to append it after setting all the style properties, just in case changing the style after it's in the DOM would cause a repaint.

components/inputs/input-date.js
```diff
@@ -211,12 +225,46 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
+		text.style.fontSize = '0.8rem';
+		text.style.fontWeight = '400';
+		text.style.letterSpacing = '0.02rem';
+		text.style.lineHeight = '1.4rem';
+		text.style.position = 'absolute';
+		text.style.width = 'auto';
+
+		// in some languages (e.g., fr) placeholderWidth is bigger, in others (e.g., zh) contentWidth is bigger
+		text.textContent = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+		const placeholderWidth = text.getBoundingClientRect().width;
```
dlockhart, line 247: Each `getBoundingClientRect()` call is going to force a paint (obviously). I wonder if it would be nicer performance wise to nest these inside `requestAnimationFrame` calls... @dbatiste any opinions here? This is what fastdom does under the covers.

The code would become lest clean and more nest-y though.

components/inputs/demo/input-date.html
```diff
@@ -41,6 +41,13 @@ <h2>Disabled</h2>
 					<d2l-input-date label="Name (disabled)" disabled></d2l-input-date>
 				</template>
 			</d2l-demo-snippet>
+
+			<h2>Empty State Text</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-date label="Graduation Date" empty-text="No graduation date"></d2l-input-date>
```
dbatiste, line 48: Are we sure that we need to expose `empty-text` on the API, and that we can't use some generic term.  If we can use a generic term, then the max-width stuff might be simplified.

Nit: the heading for the demo says "Empty State Text" but the attribute is `empty-text`.  Looks like there are a couple other references to "state" in the code.  May want to be consistent in naming.

margaree: We do want to expose `empty-text` - it seems the value will likely vary depending on context/usage.

components/inputs/input-date.js
dbatiste: Yeah it would be nice to avoid this.  This does seem kinda hacky.  Perhaps we could include these things in the template, with visibility: hidden to force the component to be at least that big without needing to create it elsewhere, measure it, and then update with the measured width... ?

components/inputs/demo/input-date.html
dlockhart: Plus different languages and the user's font size & zoom settings would impact its size, so I think we need to calculate no matter what.

components/inputs/input-date.js
dlockhart: Not a terrible idea... we'd need to hide them from AT as well somehow.

margaree: Thanks for the suggestion - I'll give this a go

components/inputs/input-date.js
```diff
@@ -155,25 +170,44 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 	}
 
 	render() {
+		const shortDateFormat = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+		const inputTextWidth = `calc(${this._hiddenContentWidth} + 0.75rem + 2px)`; // text and icon width + paddingRight + border width
 		return html`
+			<div class="d2l-input-date-hidden-content">
+				<div><d2l-icon icon="tier1:calendar"></d2l-icon>${formatISODateInUserCalDescriptor('2020-12-20')}</div>
+				<div><d2l-icon icon="tier1:calendar"></d2l-icon>${shortDateFormat}</div>
+				<div><d2l-icon icon="tier1:calendar"></d2l-icon>${this.emptyText}</div>
```
dlockhart, line 196: Good idea! The `visibility: hidden` for sure hides this from screen readers?

margaree: It does seem to (tested on NVDA and VoiceOver, and googling seems to indicate that it should). I'll add `aria-hidden="true"` too just in case.

PR #512, opened 2020-04-28
Improve event handling for nested tabs

dbatiste:
The original code would have worked - this change just avoid possible JS error when `d2l-tab-panel-selected` or `d2l-tab-panel-text-changed` is dispatched from a nested set of tabs.

### DIFF
--- a/components/tabs/demo/tabs.html
+++ b/components/tabs/demo/tabs.html
@@ -56,27 +56,6 @@
 				</template>
 			</d2l-demo-snippet>
 
-			<h2>Tabs (nested)</h2>
-
-			<d2l-demo-snippet>
-				<template>
-					<d2l-tabs>
-						<d2l-tab-panel text="Biology">Tab content for Biology</d2l-tab-panel>
-						<d2l-tab-panel text="Earth Sciences">
-							Nested tabs for Earth Sciences...
-							<d2l-tabs>
-								<d2l-tab-panel text="Mineralogy">Tab content for Mineralogy</d2l-tab-panel>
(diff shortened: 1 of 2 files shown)


components/tabs/demo/tabs.html
```diff
@@ -118,6 +118,27 @@ <h3>Tabs (max-to-show)</h3>
...
+								<d2l-tab-panel text="Geochemistry">Tab content for Geochemistry</d2l-tab-panel>
+								<d2l-tab-panel text="Groundwater Modeling">Tab content for Groundwater Modeling</d2l-tab-panel>
+								<d2l-tab-panel text="Stratigraphy">Tab content for Stratigraphy</d2l-tab-panel>
+							</d2l-tabs>
+						</d2l-tab-panel>
+						<d2l-tab-panel text="Physics">Tab content for Physics</d2l-tab-panel>
+					</d2l-tabs>
+				</template>
+			</d2l-demo-snippet>
+
```
dlockhart, line 141: Personally I don't like it when really weird edge cases that we don't expect to see often "in the wild" take up space in our demo pages -- which are meant to be a showcase of the component.

How hard would it be to move this into a test that exercises the error and now the solution?

dbatiste: I'm am tempted to just ignore it in that case.

svanherk: Yeah I could write one in my-courses, but I'm also working to get rid of the nesting there since it's not necessary.  And then I don't know of anywhere else that uses it... maybe if someone has put a filter dropdown inside a tab panel somewhere?

PR #515, opened 2020-04-29
PoC: Add hidden inputs to the light DOM using a FormElementMixin and use a native form

AllanKerr:
## Proof of Concept
This works really well for including custom elements in forms but falls short on validation support.

### Pros

1. Easy to implement
2. Works with native forms

### Cons

1. No support for `async` validation. `onsubmit` can prevent submission but only synchronously since input validity is either valid or error, there is no pending state. https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation
2. We need to wire up our own validation in the form submit event because `setCustomValidity` is only supported on native form elements and ignored on hidden inputs.

### DIFF
--- a/components/validation/demo/validation.html
+++ b/components/validation/demo/validation.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+	<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+	<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+			import '../../inputs/input-checkbox.js';
+			import '../../inputs/input-text.js';
+			import '../../typography/typography.js';
+			import '../../button/button.js';
+			import '../../colors/colors.js';
+	</script>
+	<style>
+		body {
+			padding: 30px;
+		}
+	</style>
+</head>
+
+<body>
+
+	<form action="/action" onsubmit="return validate()">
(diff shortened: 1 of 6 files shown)


components/validation/form-element-mixin.js
```diff
@@ -0,0 +1,35 @@
...
+		super();
+		this.__hiddenInput = document.createElement('input');
+		this.__hiddenInput.type = 'hidden';
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		requestAnimationFrame(() => {
+			const parentNode = this.parentNode;
+			const parent = parentNode.nodeType === Node.DOCUMENT_FRAGMENT_NODE ? this.getRootNode().host : parentNode;
```
dbatiste, line 13: I think this would need to be `getRootNode({composed: true})`.  Without the `composed: true` (which defaults to `false`), `getRootNode` will return the shadow root, but we'd want to append these hidden form elements to the document root.

components/validation/form-element-mixin.js
```diff
@@ -0,0 +1,35 @@
+export const FormElementMixin = superclass => class extends superclass {
+
+	constructor() {
+		super();
+		this.__hiddenInput = document.createElement('input');
```
dbatiste, line 5: Is there a reason not to defer this work until `connectedCallback`?

AllanKerr: Yeah, it'd be fine to defer it. Just threw it there to get the proof of concept functional as fast as possible. Its still pretty rough around the edges.

components/validation/demo/validation.html
```diff
@@ -0,0 +1,43 @@
...
+	<style>
+		body {
+			padding: 30px;
+		}
+	</style>
+</head>
+
+<body>
+
+	<form action="/action" onsubmit="return validate()">
```
dbatiste, line 26: I think it would be worthwhile for the PoC to wrap this up in a mock app component, so all this ends up being contained in another shadowDOM.

PR #517, opened 2020-04-29
Default value for time input

Scy-D2L:
Adds a default value option to the time input to override the default value of 00:00:00
Value can be of the format `hh:mm:ss` or one of `startOfDay`/`endOfDay`
Invalid default-value results in default of 00:00:00

Includes tests/docs


components/inputs/input-time.js
```diff
@@ -36,6 +35,18 @@ function getIntervalNumber(size) {
...
 }
 
+function getDefaultTime(time) {
+	switch (time) {
+		case 'endOfDay':
+			return END_OF_DAY;
+		case 'startOfDay':
+		case undefined:
+			return new Date(TODAY.year, TODAY.month, TODAY.date, 0, 0, 0);
+		default:
```
dlockhart, line 45: What happens if it fails to parse here? I guess `parseValue` which calls `parseISOTime` throws an exception, which is the same as what would happen with an invalid value?

components/inputs/input-time.js
```diff
@@ -194,6 +204,9 @@ class InputTime extends LitElement {
 		if (this.label === null) {
 			console.warn('d2l-input-time component requires label text');
 		}
+		if (this.value === undefined) {
+			this.value = formatValue(getDefaultTime(this.defaultValue));
```
dlockhart, line 208: And the reason we only have to do this the first time is that now this becomes the "last known good" value so it'll revert to it in the future when an invalid value is entered.

components/inputs/test/input-time.test.js
```diff
@@ -113,11 +113,21 @@ describe('d2l-input-time', () => {
 			expect(elem.value).to.equal('0:00:00');
 		});
 
+		it('should apply custom default', async() => {
+			const elem = await fixture('<d2l-input-time label="label text" default-value="02:00:00"></d2l-input-time>');
+			expect(elem.value).to.equal('2:00:00');
+		});
```
dlockhart, line 119: Can you add tests for the `startOfDay`, `endOfDay` and maybe an invalid value as well (which should throw)?

components/inputs/input-time.js
Scy-D2L: When `parseValue` is given something invalid it returns a date object with a time equivalent to `00:00:00`, `parseTime` is the one that returns null which we use with invalid input values, if I understand what you're referencing.

components/inputs/input-time.js
Scy-D2L: ~Correct. We used to assign `_value` and `_formattedValue` directly in the constructor because they relied on an constant external to the component instance, but as I understand I can't get our dynamic default until `firstUpdated` (after the regular `value` property is already assigned) we only need to overwrite it if no value was assigned.~

Found a flaw with this. If the `value` is set to something invalid in the html the value will be set to midnight instead of undefined before we get here - so if the value is junk then we end up with 12AM instead of the default 🤔 but there's presently no way to tell if what was originally passed in was 12AM or something else. I'd need to see what `value` was raw before it passed through the setter, and I'm not sure how to do that.

dlockhart: So seems like you need to apply the default  in the `value` setter? Will `oldValue` initially be `undefined` the first time it runs? Maybe you can use that somehow...

components/inputs/input-time.js
Scy-D2L: Updated this so that it will throw.

components/inputs/input-time.js
```diff
@@ -118,15 +131,19 @@ class InputTime extends LitElement {
...
 	get value() { return this._value; }
 	set value(val) {
+		if (this.value === undefined) {
+			this._givenValue = val;
+			return;
+		}
+
 		const oldValue = this.value;
-		const time = parseValue(val);
+		const time = val === '' ? getDefaultTime(this.defaultValue) : parseValue(val);
```
dlockhart, line 145: Is empty string the only case where we'd want to revert to the default? Like what happens if the value attribute is completely removed? i.e. I have this `<d2l-input-time value="09:00:00">` and then I go `elem.removeAttribute('value')` -- what gets passed into the setter here? I'm guessing `undefined`.

components/inputs/input-time.js
Scy-D2L: Because of the need to check against the default value and given value together and the way they don't both seem to exist when it's called the first time I took a different approach to figure out the initial value after the setter is called. Feedback welcome.

components/inputs/input-time.js
```diff
@@ -118,15 +131,19 @@ class InputTime extends LitElement {
...
-		this._formattedValue = formatTime(DEFAULT_VALUE);
 		this._timezone = formatTime(new Date(), {format: 'ZZZ'});
-		this._value = formatValue(DEFAULT_VALUE);
 	}
 
 	get value() { return this._value; }
 	set value(val) {
+		if (this.value === undefined) {
+			this._givenValue = val;
+			return;
```
dlockhart, line 141: Trying to figure out what this is doing... so `_givenValue` is tracking the first ever provided `value`? And then later in `firstUpdated` if that first-ever-value is undefined (or empty), it'll set `value` to default otherwise use the value they provided. Why can't we just do that here in the setter? Sorry if we covered that in the first round.

components/inputs/test/input-time.test.js
```diff
@@ -113,11 +113,56 @@ describe('d2l-input-time', () => {
...
+			const elem = await fixture('<d2l-input-time label="label text" default-value="startOfDay"></d2l-input-time>');
+			expect(elem.value).to.equal('0:00:00');
+		});
+
+		it('should apply custom default value from keyword: endOfDay', async() => {
+			const elem = await fixture('<d2l-input-time label="label text" default-value="endOfDay"></d2l-input-time>');
+			expect(elem.value).to.equal('23:59:59');
+		});
+
+		it.only('should apply default when given value is empty', async() => {
```
Scy-D2L, line 141: Missed this, will be fixed in the next commit.

components/inputs/input-time.js
Scy-D2L: It passes `null` apparently - I'll accommodate that.

helpers/test/dateTime.test.js
```diff
@@ -293,20 +293,10 @@ describe('date-time', () => {
...
-		it('should return all 0 when empty input', () => {
-			expect(parseISOTime('')).to.deep.equal({hours: 0, minutes: 0, seconds: 0});
-		});
-
-		it('should return all 0 when just date', () => {
-			expect(parseISOTime('2019-02-12')).to.deep.equal({hours: 0, minutes: 0, seconds: 0});
-		});
-
-		it('should return all 0 when no input', () => {
-			expect(parseISOTime('')).to.deep.equal({hours: 0, minutes: 0, seconds: 0});
```
dlockhart, line 309: I'd recommend still testing for a few of these edge cases -- you can do a `forEach` like this:

```javascript
['12', '', '2019-02-12'].forEach((val) => {
    expect(() => parseISOTime(val)).to.throw('Invalid input: Expected format is hh:mm:ss');
});
```

components/inputs/input-time.js
Scy-D2L: To sum up a slack convo for posterity: this made sense when `parseValue` gave `0:00:00` as a default instead of throwing (had to check for value and default value at the same time to prevent overwriting things) but now that value can be checked on its own first it can be done here. Fixed :)

PR #518, opened 2020-04-29
Visual-diff updates 

dbatiste:
The reduced motion stuff avoids having to listen for transition/animation events.  These listeners were removed from test code in previous PRs.

Ideally, other developers can interact with the core components in their app components and get correct, consistent screenshots without waiting, especially arbitrarily.  This PR attempts to explore that, and considers updating tests in core to do exactly that.

Interested in thoughts/opinions.  No rush.  Even if we keep visual-diff tests the same, this at least helps to understand what app developers should/must do if they are interacting with complex core components in their diff tests. Similar changes could be made to dropdown, tooltip, and other tests.

### DIFF
--- a/components/dialog/test/dialog.visual-diff.js
+++ b/components/dialog/test/dialog.visual-diff.js
@@ -44,34 +44,40 @@ describe('d2l-dialog', () => {
 					});
 
 					it('opened', async function() {
-						await helper.open(page, '#dialog');
+						await page.$eval('#dialog', (dialog) => dialog.opened = true);
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
 					it('scroll bottom shadow', async function() {
-						await helper.open(page, '#dialogLong');
+						await page.$eval('#dialogLong', (dialog) => dialog.opened = true);
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
 					it('scroll top shadow', async function() {
-						await helper.open(page, '#dialogLong');
-						await page.$eval('#dialogLong #bottom', (bottom) => {
-							bottom.scrollIntoView();
+						await page.$eval('#dialogLong', (dialog) => {
+							dialog.opened = true;
+							return new Promise((resolve) => {
+								requestAnimationFrame(() => {
+									dialog.querySelector('#bottom').scrollIntoView();
+									resolve();
+								});
+							});
 						});
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
 					it('rtl', async function() {
-						await helper.open(page, '#dialogRtl');
+						await page.$eval('#dialogRtl', (dialog) => dialog.opened = true);
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
 					it('resize', async function() {
-						await helper.open(page, '#dialogResize');
-						await page.$eval('#dialogResize', (dialog) => {
+						await page.$eval('#dialogResize', async(dialog) => {
+							dialog.opened = true;
 							dialog.querySelector('div').style.height = '60px';
 							dialog.width = 500;
 							dialog.resize();
+							return new Promise((resolve) => requestAnimationFrame(resolve));
 						});
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
--- a/components/menu/test/menu-rtl.visual-diff.js
+++ b/components/menu/test/menu-rtl.visual-diff.js
@@ -39,12 +39,7 @@ describe('d2l-menu rtl', () => {
 
 		it('opens submenu on click', async function() {
 			// this scenario also tests height change when going from 1 menu item to 2 within nested menu
-			await page.$eval('#nested-item', item => {
-				return new Promise((resolve) => {
-					item.addEventListener('d2l-hierarchical-view-show-complete', resolve, { once: true });
-					item.click();
-				});
-			});
+			await page.$eval('#nested-item', item => item.click());
 			const rect = await visualDiff.getRect(page, '#nested');
 			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 		});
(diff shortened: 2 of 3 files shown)


components/dialog/test/dialog.visual-diff.js
```diff
@@ -44,34 +44,40 @@ describe('d2l-dialog', () => {
 					});
 
 					it('opened', async function() {
-						await helper.open(page, '#dialog');
```
dbatiste, line 47: The helper simply waits for the `d2l-dialog-open` event, which technically is probably the safest because it ensure that the dialog has been sized and the focus applied.  So in this case I would tend to prefer waiting for the event, and unfortunately I'd suggest app devs do the same.  They should however, not need to wait an arbitrary time anymore.

components/dialog/test/dialog.visual-diff.js
```diff
@@ -44,34 +44,40 @@ describe('d2l-dialog', () => {
...
-						await helper.open(page, '#dialogLong');
+						await page.$eval('#dialogLong', (dialog) => dialog.opened = true);
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
 					it('scroll top shadow', async function() {
-						await helper.open(page, '#dialogLong');
-						await page.$eval('#dialogLong #bottom', (bottom) => {
-							bottom.scrollIntoView();
+						await page.$eval('#dialogLong', (dialog) => {
```
dbatiste, line 57: This is another example that I think is actually better waiting for the event, otherwise this code needs to wait a frame before trying to scroll.

components/menu/test/menu-rtl.visual-diff.js
```diff
@@ -56,12 +51,7 @@ describe('d2l-menu rtl', () => {
 
 		it('opens long menu item submenu on click', async function() {
 			// this scenario also tests height change going from 3 menu items to 2 within nested menu
-			await page.$eval('#nested-item-long', item => {
-				return new Promise((resolve) => {
-					item.addEventListener('d2l-hierarchical-view-show-complete', resolve, { once: true });
-					item.click();
-				});
-			});
+			await page.$eval('#nested-item-long', item => item.click());
```
dbatiste, line 54: This is an example where the test is nicely simplified, and avoiding the underlying hierarchical-view event is nice.  However, I can't always explain why it's not necessary to wait a frame before taking a screenshot (which is required in the next case with the custom nested view).

dbatiste: Based on my CI failure... the need to wait a frame may actually be more common.

components/dialog/test/dialog.visual-diff.js
dlockhart: Yeah seems like waiting for the event is best, since it's part of the dialog's public API anyway.

components/menu/test/menu-rtl.visual-diff.js
dlockhart: So waiting for the event I guess is necessary. Could this pattern be used though to simplify?

```
setTimeout(() => item.click());
await oneEvent('d2l-hierarchical-view-show-complete', item);
```

dbatiste:
Thanks Dave.  Yeah I agree, the events are cleaner/safer.  I've been thinking about the helpers a bit.  There are the two contexts where helper could be useful... in the test context, and in the client-side context.  

**Test context, scenario 1**: the app wants to open the dialog directly through code.  I haven't used the handle API before, but it could perhaps be used.

app test...
```
const dialogHandle = await page.evaluateHandle(() => document.querySelector('d2l-app').shadowRoot.querySelector('d2l-dialog'));
await openDialog(page, dialogHandle);
// screenshot, etc.
```

helper...
```
openDialog(page, handle) {
    return page.evaluate((dialog) => {
            dialog.addEventListener('d2l-dialog-open', resolve, { once: true });
            dialog.opened = true;
        });
    }, handle);
}
```
So that's pretty clean, but sometimes we expect to interact in different ways...

**Test context, scenario 2**: the app wants to open the dialog by clicking on a button.

app test...
```
const dialogHandle = await page.evaluateHandle(() => document.querySelector('d2l-app').shadowRoot.querySelector('d2l-dialog'));

const openEvent = oneEvent('d2l-dialog-open', page, dialogHandle);
await page.evaluate(() => document.querySelector('d2l-app').shadowRoot.querySelector('d2l-button').click());
await openEvent;
// screenshot, etc.
```

helper...
```
oneEvent(name, page, handle) {
    return page.evaluate((elem) => {
            elem.addEventListener(name, resolve, { once: true });
        });
    }, handle);
}
```
The `oneEvent` helper could prove useful, but the app test code is awkward.  This can perhaps be done more cleanly if all the logic is moved into a single `evaluate`...

**Client-side context:** move all the logic into `evaluate`:
 
app test...
```
await page.$eval('d2l-app', (app) => {
    setTimeout(() => app.shadowRoot.querySelector('d2l-button').click());
    return oneEvent('d2l-dialog-open', app.shadowRoot.querySelector('d2l-dialog'));
});
// screenshot, etc.
```

helper (in browser, imported or custom)...
```
oneEvent(name, elem) {
    return new Promise((resolve) => elem.addEventListener(name, resolve, { once: true }));
}
```
Which either requires the dev to add the helper script to their test page, or we leverage our custom `visualDiff.createPage()` to automatically inject it.
```
createPage(...) {
    ...
    await page.addScriptTag({
        window.D2L.TestHelpers.oneEvent = (name, elem) => {
            return new Promise((resolve) => elem.addEventListener(name, resolve, { once: true }));
        };
    });
    return page;
}
```
Which I think is basically what you'd suggested in one of your comments.  I might prefer to leave it to the dev to add the shared script to their test page, that way there is no mystery, but I am open to either.  I think there could be some other client-side helper code that could be added there... for example the key event stuff.  Would we want to version such helpers with core, or with visual-diff?

Of all of the above, I think I prefer the last, where the helper is client-side.

The other idea you mentioned was adding methods to the component API that would return a promise.  I've considered this in the past as well, and in some cases it would be helpful, however it does not necessarily reflect how the user and app are really interacting.  In the core suite, we have tests for different interaction, such as click, keyboard, etc.  Dialog also already has an `open` method that returns a promise that resolves with the dialog result when it closes, so that particular case becomes extra awkward.

PR #522, opened 2020-05-01
Fix tabs find and findIndex for IE11.

dbatiste:
Given this code will not be around for very long (famous last words), I opted to not include a full blown array polyfill, and I kept this code as close to the original as possible so that removing it is just a matter of deleting the conditional statements/else block, even though it results in a tiny bit of repeat.

### DIFF
--- a/components/tabs/tabs.js
+++ b/components/tabs/tabs.js
@@ -378,12 +378,7 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 			const handleTransitionEnd = (e) => {
 				if (e.propertyName !== 'max-width') return;
 				tab.removeEventListener('transitionend', handleTransitionEnd);
-				if (this._tabInfos.findIndex) {
-					this._tabInfos.splice(this._tabInfos.findIndex(info => info.id === tabInfo.id), 1);
-				} else {
-					// IE11
-					this._tabInfos.splice(this._getTabInfoIndex(this._tabInfos, tabInfo.id), 1);
-				}
+				this._tabInfos.splice(this._tabInfos.findIndex(info => info.id === tabInfo.id), 1);
 				this.requestUpdate();
 				resolve();
 			};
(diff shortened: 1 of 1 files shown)


components/tabs/tabs.js
```diff
@@ -378,7 +378,12 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 			const handleTransitionEnd = (e) => {
 				if (e.propertyName !== 'max-width') return;
 				tab.removeEventListener('transitionend', handleTransitionEnd);
-				this._tabInfos.splice(this._tabInfos.findIndex(info => info.id === tabInfo.id), 1);
```
dlockhart, line 381: I'm a little confused. We're already [including a findIndex](https://github.com/Brightspace/brightspace-integration/blob/master/js/polyfills.js#L2) polyfill in BSI...

components/tabs/tabs.js
```diff
@@ -514,8 +519,22 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
...
+	_getTabInfoIndex(infos, id) {
+		for (let i = 0; i < infos.length; i++) {
+			if (infos[i].id === id) return i;
+		}
+		return -1;
+	}
+
 	_getTabInfo(id) {
-		return this._tabInfos.find((t) => t.id === id);
+		if (this._tabInfos.find) {
```
dlockhart, line 531: We are not including a `find` polyfill though, so maybe that caused the issue?

components/tabs/tabs.js
dbatiste: Oh, I was testing in the core demo pages where it's not included.  Hmm.  If we want to just rely on that, it would simplify this a bit.

components/tabs/tabs.js
dbatiste: Yeah, `find` was the original issue, but then I ran into `findIndex` in core demo page while testing IE11.

dbatiste: Ok I've updated to simply rely on the polyfill for `findIndex`, but to be safe, we should technically pull that polyfill into core so `d2l-tabs` can import it, since there are few things that are relying on tabs and I'm not certain they are all in BSI.

dbatiste:
Ok, I took a super temp copy of the polyfill that this can use - aiming for minimal effort to remove in a few weeks.

PR #523, opened 2020-05-04
input-text: fix width calculation when not initially displayed

margaree:
This surfaced as a bug where the placeholder text was overlapping with the icon in input-date.
The problem was that input-date was within a div that was initially not displayed, and `offsetWidth` does not seem to work to calculate the slot content width if not displayed. This solution gets the slot content (`d2l-icon`) and calculates overall width using width and margins.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -258,9 +258,10 @@ class InputText extends RtlMixin(LitElement) {
 		if (!slotContent) return;
 		const id = e.target.parentNode.id;
 
-		this.requestUpdate(); // needed for legacy Edge
-		this.updateComplete.then(() => {
-			const slotWidth = parseFloat(getComputedStyle(slotContent).width) + parseFloat(getComputedStyle(slotContent).marginLeft) + parseFloat(getComputedStyle(slotContent).marginRight);
+		// requestUpdate needed for legacy Edge
+		this.requestUpdate().then(() => {
+			const style = getComputedStyle(slotContent);
+			const slotWidth = parseFloat(style.width) + parseFloat(style.marginLeft) + parseFloat(style.marginRight);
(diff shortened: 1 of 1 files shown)


components/inputs/input-text.js
```diff
@@ -253,11 +253,16 @@ class InputText extends RtlMixin(LitElement) {
 		this._hovered = false;
 	}
 
-	_onSlotChange() {
+	_onSlotChange(e) {
+		const slotContent = e.target.assignedNodes()[0];
+		if (!slotContent) return;
+		const id = e.target.parentNode.id;
+
 		this.requestUpdate(); // needed for legacy Edge
```
dbatiste, line 261: Note, `requestUpdate` returns the same promise that the `updateComplete` would in this scenario.  So this could be written like:
```
this.requestUpdate().then(() => {
    ...
});
```
Or...
```
async _onSlotChange(e) {
    ...
    await this.requestUpdate();
    ...
}
```

components/inputs/input-text.js
```diff
@@ -253,11 +253,16 @@ class InputText extends RtlMixin(LitElement) {
...
 		this.updateComplete.then(() => {
-			this._firstSlotWidth = this.shadowRoot.querySelector('#first-slot').offsetWidth;
-			this._lastSlotWidth = this.shadowRoot.querySelector('#last-slot').offsetWidth;
+			const slotWidth = parseFloat(getComputedStyle(slotContent).width) + parseFloat(getComputedStyle(slotContent).marginLeft) + parseFloat(getComputedStyle(slotContent).marginRight);
```
dbatiste, line 263: The browser _probably_ caches the result of the computed style call, but I'd probably just call it once.
```
const slotContentStyle = getComputedStyle(slotContent);
...
```

PR #524, opened 2020-05-04
Time Input Accessibility Fixes 

Scy-D2L:
Fix for [DE38483](https://rally1.rallydev.com/#/detail/defect/379582709596?fdp=true): Time Input > Screen readers getting "blank" tab stop when entering dropdown and Legacy Edge reads everything all the time.

- Dropdown menu wasn't correctly shifting focus to the menu when tabbing into the dropdown.
- Ledge was reading the entire dropdown as a label.


components/inputs/input-time.js
```diff
@@ -144,47 +144,47 @@ class InputTime extends LitElement {
 		const input = html`
 			<label>
 				<span class="${this.label && !this.labelHidden ? 'd2l-input-label' : 'd2l-offscreen'}" id="${this._dropdownId}-label">${this.label}</span>
```
dlockhart, line 146: Having a `<label>` that doesn't use `for` and doesn't wrap its input isn't useful, so I think the `<span>` can just become the main node here.

components/dropdown/dropdown-menu.js
```diff
@@ -80,6 +81,10 @@ class DropdownMenu extends DropdownContentMixin(LitElement) {
 		this.close();
 	}
 
+	_onFocus() {
+		this.__getMenuElement().focus();
```
dlockhart, line 85: Won't this move focus to the menu when the time selector receives focus? I would have thought we'd want to focus on the text input... but maybe I'm misunderstanding something.

components/inputs/input-time.js
Scy-D2L: Whoops - absolutely, thanks :)

components/dropdown/dropdown-menu.js
Scy-D2L: This applies to the `d2l-dropdown-menu` element (here: https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-time.js#L162) which is more like the content wrapper/container for the dropdown box and doesn't include the text input. When you tab, it moves the focus from the dropdown opener (the text input) to that wrapper, but without this code it wasn't shifting from the wrapper to the actual menu (thus the 'blank' tab stop). Hopefully that makes sense.

dlockhart: Ah gotcha, makes sense -- thanks!

dbatiste: I think the original intent of the code that creates the tab-stop on the container was - for generic (non-menu) content, we need to move the focus into the dropdown _somewhere_, so if there are no focusable elements in the content, we create that tab stop on the fly.  Not ideal... a corner case though. In the `d2l-dropdown-menu` case, this was thought to never be hit because we can theoretically expect there to be a focusable element, the menu/items.  So... I'm wondering why this usage is different?

Scy-D2L: I was able to reproduce the bug using a generic dropdown menu with enough items to require scrolling, so I think it just went under the radar because most menus aren't this long. I don't know why exactly it only kicked in when it reached scrolling length, though.

PR #525, opened 2020-05-04
d2l-alert Lit Port

AllanKerr:
# Changes
1. Port `d2l-alert` to Lit (this is only `d2l-alert`, not `d2l-alert-toast`)
1. Change the `default` color from `celestine+1` to `celestine+0` to align with design.d2l.


components/alert/README.md
```diff
@@ -0,0 +1,33 @@
+# Alerts
+
+## d2l-alert
+
+The `d2l-alert` component can be used to communicate important information relating to the state of the system and the users’ work flow.
```
dlockhart, line 5:
```suggestion
The `d2l-alert` component can be used to communicate important information relating to the state of the system and the user's workflow.
```

components/alert/alert.js
```diff
@@ -0,0 +1,156 @@
...
+import { LocalizeStaticMixin } from '../../mixins/localize-static-mixin.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
+
+	static get properties() {
+		return {
+			buttonText: { type: String, attribute: 'button-text' },
+			hasCloseButton: { type: Boolean, attribute: 'has-close-button' },
+			hidden: { type: Boolean, reflect: true },
```
dlockhart, line 15: Interesting... `hidden` is a standard DOM attribute that all nodes have. I don't _think_ we actually need to specify it? The only reason we might need to would be if reflection isn't working without it.

components/alert/alert.js
```diff
@@ -0,0 +1,156 @@
...
+				}
+				to {
+					opacity: 1;
+					transform: translate(0,0);
+				}
+			}
+
+			@media (prefers-reduced-motion: reduce) {
+				:host {
+					animation: none;
```
dlockhart, line 99: 💯

components/alert/alert.js
```diff
@@ -0,0 +1,156 @@
...
+		return [bodyCompactStyles, css`
+
+			:host {
+				animation: 600ms ease drop-in;
+				background: white;
+				border-radius: 0.3rem;
+				border: 1px solid var(--d2l-color-mica);
+				box-sizing: border-box;
+				display: block;
+				display: flex;
```
dlockhart, line 30: I don't think `display: block` is supposed to be there? I don't see it on the old version.

components/alert/test/alert.test.js
```diff
@@ -0,0 +1,60 @@
...
+	});
+
+	describe('events', () => {
+
+		it('should fire "d2l-alert-closed" event when close button is clicked', async() => {
+			const closeButton = alert.shadowRoot.querySelector('d2l-button-icon.d2l-alert-action');
+			setTimeout(() => closeButton.click());
+			await oneEvent(alert, 'd2l-alert-closed');
+			expect(alert.hasAttribute('hidden')).to.be.true;
+		});
```
dlockhart, line 50: Might want the same test for the `close()` method, which is technically part of its public API. I know the button just calls it, but that may not always be true.

components/alert/test/alert.visual-diff.html
```diff
@@ -0,0 +1,47 @@
...
+	</div>
+	<div class="visual-diff">
+		<d2l-alert id="warning" type="warning">A warning message.</d2l-alert>
+	</div>
+	<div class="visual-diff">
+		<d2l-alert id="close" type="error" has-close-button>A message.</d2l-alert>
+	</div>
+	<div class="visual-diff">
+		<d2l-alert id="button" type="call-to-action" button-text="Do it!">A message.</d2l-alert>
+	</div>
```
dlockhart, line 33: Thanks for including tests for "error" and "call-to-action", which are only supported for backwards-compatibility.

components/alert/test/alert.visual-diff.html
```diff
@@ -0,0 +1,47 @@
...
+			subtext="I am subtext. I'm here to test the wrapping capabilities of adding subtext to these alerts, as well as other styling issues. Feel free to add to me!">
+			A message.
+		</d2l-alert>
+	</div>
+	<div class="visual-diff">
+		<d2l-alert id="rtl" type="default" button-text="Do it!" has-close-button dir="rtl"
+			subtext="I am subtext. I'm here to test the wrapping capabilities of adding subtext to these alerts, as well as other styling issues. Feel free to add to me!">
+			A message.
+		</d2l-alert>
+	</div>
```
dlockhart, line 45: Weird to test for this.. but maybe a test for the `hidden` property?

components/alert/test/alert.visual-diff.js
```diff
@@ -0,0 +1,36 @@
...
+
+	[
+		'default',
+		'success',
+		'critical',
+		'warning',
+		'close',
+		'button',
+		'button-close',
+		'rtl'
```
dlockhart, line 27: Missing "error" and "call-to-action" here though :)

dlockhart:
Based on the design, I was expecting the close and action buttons to be more vertically centred when there's only a single line of text. Possible that the position is off a bit?

components/alert/test/alert.visual-diff.js
AllanKerr: I wasn't sure whether they were worth adding more test cases for since they aren't documented as part of `design.d2l` or the README. They're really only there for backwards compatibility since they share colors with `critical` and `default`.

components/alert/alert.js
dlockhart: A bit more on this: normally we like for people to manipulate attributes instead of call methods on components. i.e. `<d2l-dialog open>` instead of `open()`. So if `hidden` is the attribute/property equivalent of not needing to call the `close()` method, then I'm more OK with it. But I wish we'd called it something other than `hidden`, which is already a thing.

AllanKerr: In the Polymer version it used `setAttribute('hidden', 'hidden')` within `close`, I changed it to a reflected property so we didn't need to set the attribute ourselves.

As for the CSS style, I included it because it doesn't work in the demo pages so the close button looks broken. It does however work without the additional `css` in the LMS. I'll move the CSS so it only applies to the demo page rather than the component itself.

components/alert/test/alert.visual-diff.js
dlockhart: I think it's worth it solely for the future-case when we go into this code and are like "why is it listing `error` AND `critical` in the CSS? Must be a mistake -- DELETE DELETE DELETE" and this would catch the regression.

components/alert/alert.js
dlockhart: The CSS is definitely needed if we're changing the default `display` to anything other than `inline` -- otherwise the browser's built-in CSS for the `display` attribute don't apply.

AllanKerr:
> Based on the design, I was expecting the close and action buttons to be more vertically centred when there's only a single line of text. Possible that the position is off a bit?

There are definitely some inconsistencies between the existing alert implementation and design.d2l. I included a few of the easy ones in this PR but left any potentially breaking or more substantial changes out. I was going to create a separate story to get them in sync and put it in the backlog because it felt like that may be prioritized differently than the validation work.

components/alert/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+- `button-text` (optional, String): text that is displayed within the alert's action button. If no text is provided the button is not displayed.
+- `has-close-button` (Boolean, default: `false`) gives the alert a close button that will close the alert when clicked.
+- `hidden` (Boolean, default: `false`): Whether or not the alert is currently visible.
+- `subtext` (optional, String) The text that is displayed below the main alert message.
+- `type` (String, default: `'default'`): type of the alert being displayed. Can be one of  `default`, `critical`, `success` , `warning`
+
+**Events:**
+* `d2l-alert-closed`: dispatched when the alert's close button is clicked
+* `d2l-alert-button-pressed`: dispatched when the alert's action button is clicked
```
dbatiste, line 29: Kind of a weird event name.  Again, pain to change the API at this point.

components/alert/alert.js
```diff
@@ -0,0 +1,153 @@
...
+			}
+			.d2l-alert-subtext {
+				margin: 0.5rem 0 0;
+			}
+
+			.d2l-alert-action {
+				margin: 0.3rem;
+			}
+
+			@keyframes drop-in {
```
dbatiste, line 84: I know this was copied forward from the original.  I think think could have been done with a simple transition though.

Nit: I think the `0.0` on opacity below could just be `0`.  Another nit: the translate could have been `translateY(...)`.  Meh.

components/alert/alert.js
```diff
@@ -0,0 +1,153 @@
...
+
+	render() {
+		return html`
+			<div class="d2l-alert-highlight"></div>
+			<div class="d2l-alert-text">
+				<slot></slot>
+				${this.subtext ? html`<p class="d2l-body-compact d2l-alert-subtext">${this.subtext}</p>` : null}
+			</div>
+			${this.buttonText && this.buttonText.length > 0 ? html`<d2l-button-subtle class="d2l-alert-action" text=${this.buttonText} @click=${this._onButtonClick}></d2l-button-subtle>` : null}
+			${this.hasCloseButton ? html`<d2l-button-icon class="d2l-alert-action" icon="d2l-tier1:close-default" text="${this.localize('close')}" @click=${this.close}></d2l-button-icon>` : null}
```
dbatiste, line 148: Nit: the `d2l-` prefix is no longer needed on our icons.

components/alert/alert.js
```diff
@@ -0,0 +1,153 @@
...
+			'd2l-alert-button-pressed', { bubbles: true, composed: true }
+		));
+	}
+
+	render() {
+		return html`
+			<div class="d2l-alert-highlight"></div>
+			<div class="d2l-alert-text">
+				<slot></slot>
+				${this.subtext ? html`<p class="d2l-body-compact d2l-alert-subtext">${this.subtext}</p>` : null}
```
dbatiste, line 145: While `null` and `undefined` generally work, [the docs](https://lit-html.polymer-project.org/guide/writing-templates#rendering-nothing) suggest `''` is more consistent across browsers.  I haven't noticed anything specific, just going by what the docs say.  

The `nothing` sentinel is even better since really does render nothing at all, but requires importing it, and is only really necessary when rendering slot content where we want fallback slot content to be rendered when there is "nothing" in the slot.

components/alert/demo/alert.html
```diff
@@ -0,0 +1,78 @@
...
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+	<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '../../demo/demo-page.js';
+		import '../alert.js';
+	</script>
+	<style>
```
dbatiste, line 14: Nit: unused style element.

PR #528, opened 2020-05-06
Added down-arrow behaviour and improved focus for input-time

Scy-D2L:
Updated `input-time` focus and keyboard behaviour to match `input-date`
- When input is focused pressing `down arrow` will open and focus the dropdown
- When dropdown is closed, focus moves back to the input instead of the next element on the page

### DIFF
--- a/components/inputs/input-time.js
+++ b/components/inputs/input-time.js
@@ -168,12 +168,16 @@ class InputTime extends LitElement {
 						aria-controls="${this._dropdownId}"
 						aria-labelledby="${this._dropdownId}-label"
 						@change="${this._handleChange}"
+						@keydown="${this._handleKeydown}"
 						class="d2l-input"
 						?disabled="${this.disabled}"
-						@keypress="${this._handleKeypress}"
 						.value="${this._formattedValue}">
 				</div>
-				<d2l-dropdown-menu id="dropdown" no-padding-footer max-height="${ifDefined(this.maxHeight)}" min-width="195">
+				<d2l-dropdown-menu
+					@d2l-dropdown-close="${this.focus}"
+					no-padding-footer
+					max-height="${ifDefined(this.maxHeight)}"
+					min-width="195">
 					<d2l-menu
 						aria-describedby="${this._dropdownId}-timezone"
 						id="${this._dropdownId}"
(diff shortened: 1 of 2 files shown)


components/inputs/input-time.js
```diff
@@ -260,5 +264,16 @@ class InputTime extends LitElement {
 			{bubbles: true, composed: false}
 		));
 	}
+
+	async _handleKeydown(e) {
+		console.log(e.keyCode);
```
dlockhart, line 269: Console :)

components/inputs/input-time.js
```diff
@@ -168,12 +168,16 @@ class InputTime extends LitElement {
 						aria-controls="${this._dropdownId}"
 						aria-labelledby="${this._dropdownId}-label"
 						@change="${this._handleChange}"
+						@keydown="${this._handleKeydown}"
```
dlockhart, line 171: What was the motivation to switch from keypress -> keydown? We try to avoid keydown if we can since folks can hit it accidentally, whereas keypress only fires after a down and then an up.

components/inputs/input-time.js
```diff
@@ -168,12 +168,16 @@ class InputTime extends LitElement {
...
 						class="d2l-input"
 						?disabled="${this.disabled}"
-						@keypress="${this._handleKeypress}"
```
dlockhart, line 173: Am I missing something... not seeing `_handleKeypress` anywhere in the code...

Scy-D2L: Neither am I, and I don't remember using it for anything, which confuses me 🤔 maybe it ended up in there initially for some debug purpose and nobody noticed in previous PRs.

components/inputs/input-time.js
Scy-D2L: I used `keydown` to use the same events as `input-date` and because when I went to look up the difference between the two, I found `keypress` is apparently deprecated https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event - it's not technically a switch because apparently that `keypress` listener didn't do anything (see other conversation)

components/inputs/input-time.js
Scy-D2L: Question: there are [lint rules](https://eslint.org/docs/rules/no-console) to catch these, is there a reason we don't use them?

components/inputs/input-time.js
dlockhart: Crazy, that's my new learned thing for today.

components/inputs/input-time.js
dlockhart: I'm guessing since `this._handleKeyPress` is `undefined`, Lit is smart enough to just not wire up the handler and not throw.

components/inputs/input-time.js
dlockhart: Not sure... we use `console` a lot in our demos/tests, so that could be it. But yeah, might be worth adding at least in core.

dbatiste: I think the eslint recommended rules have changed.  We used to get errors for this, which is why there are opt-out comments in a few places.  

Perhaps we should not rely on the eslint recommended ruleset (we currently extend it, and override where we thought necessary).

PR #529, opened 2020-05-06
Port d2l-alert-toast to Lit

AllanKerr:
# Changes
- Substantial refactoring to convert this a more manageable state machine.
- Replace `autoclose` int property with `no-auto-close` boolean property
- Change width from 710px to 610px per design.d2l.
- Remove `announce-text` attribute. This is no longer necessary now that the role is on the `d2l-alert-toast` component itself rather than the container inside the shadow DOM.


components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+			}
+
+			:host([_state="opening"]) .d2l-alert-toast-container {
+				opacity: 1;
+				transform: translateY(0);
+				visibility: visible;
+			}
+
+			.d2l-alert-toast {
+				animation: none;
```
dbatiste, line 77: Looks like this is left-over from converting to transition?

AllanKerr: I haven't had a chance to convert `d2l-alert` to use transitions yet. This disables the `drop-in` animation that `d2l-alert` has since we don't want it running along with the toast's transitions.

dbatiste: Oh I see.  Hmmm.  Icky.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+				transition-duration: 250ms;
+				transition-property: transform, opacity, visibility;
+				transition-timing-function: ease-in;
+			}
+
+			:host([_state="preopening"]) .d2l-alert-toast-container,
+			:host([_state="closing"]) .d2l-alert-toast-container {
+				opacity: 0;
+				transform: translateY(0.5rem);
+				visibility: hidden;
```
dbatiste, line 67: Are you sure you need `visibility`?  I _think_ `opacity` should be enough.  We do this sort of thing elsewhere, ex.  [visible-on-ancestor](https://github.com/BrightspaceUI/core/blob/master/mixins/visible-on-ancestor-mixin.js#L8), and there might be one or two other cases.  If looking at VoA, ignore the `!important`... it is being used to override the host in event it is setting those properties with high specificity.

Also, seems like there might be more states than what you really need, but whatever is most understandable.  :)

components/alert/alert-toast.js
dbatiste: The class name sorta threw me off.  I think this would be clearer if is were:
```
d2l-alert {
    animation: none;
}
```
Then the class name could be removed from that element as well.

components/alert/alert-toast.js
AllanKerr: `opacity` is likely enough, `visibility` was carried through from the Polymer version. I'll give it a go.

Regarding redundant states, I could combine `opening` and `open`.
I experimented with it a bit and found that:
1. Made the code slightly more difficult to understand because `closing` had no corresponding `opening` state
1. There was no easy way to detect when the alert finished opening to do things like change focus, announce text, or start the auto-close timer.
1. I had to replace it with the `reduce-motion` media query anyway.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+		await this.updateComplete;
+
+		if (newOpen) {
+			if (this._state === states.CLOSING) {
+				this._state = states.OPENING;
+			} else if (this._state === states.CLOSED) {
+				if (!reduceMotion) {
+					this._state = states.PREOPENING;
+					// pause before running the opening animation because transitions won't run when changing from 'diplay: none' to 'display: block'
+					this._preopenFrame = requestAnimationFrame(() => {
```
dbatiste, line 149: Ignoring the reduceMotion for the moment, would it be enough to:
```
this._state = states.PREOPENING;
await this.requestUpdate();
this._state = states.OPENING;
```
Or, if necessary to wait a frame, then:
```
this._state = states.PREOPENING;
await this.requestUpdate();
requestAnimationFrame(() => this._state = states.OPENING);
```
Lit will update the state as a micro-task.  Do we need to wait two frames?

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+
+	static get properties() {
+		return {
+			buttonText: { type: String, attribute: 'button-text' },
+			hideCloseButton: { type: Boolean, attribute: 'hide-close-button' },
+			noAutoClose: { type: Boolean, attribute: 'no-auto-close' },
+			open: { type: Boolean, reflect: true },
+			subtext: { type: String },
+			type: { type: String, reflect: true },
+			_state: { type: String, reflect: true }
```
dbatiste, line 28: If we render the `_state` to the element with class `d2l-alert-toast-container`, then it _looks_ like we can avoid reflecting the `_state` to the host as an attribute.  I know we do that is a few places, but it's leaky and not really ideal.

components/alert/alert-toast.js
AllanKerr: I played around with this a ton and the only thing that I found that worked is waiting two frames or longer. Waiting for the next frame after re-render unfortunately isn't enough.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+
+	get open() {
+		return this._open;
+	}
+
+	set open(val) {
+		const oldVal = this._open;
+		if (oldVal !== val) {
+			this._open = val;
+			this.requestUpdate('open', oldVal);
```
dbatiste, line 111: Is this `requestUpdate()` call needed?  It looks like `render` does not depend on `open` value.

components/alert/alert-toast.js
dbatiste: I've been playing around with this, and it is indeed a bit confusing.  I am aware of at least one bug (which may or may not be related) when setting secondary properties from a property setter, but I was still unable to get around this.  I did however find that this also works:
```
requestAnimationFrame(() => {
    this._state = states.PREOPENING;
    requestAnimationFrame(() => {
        this._state = states.OPENING;
    });
});
```
But is it still unclear what the root problem is, because prior to that code, there are no pending state changes that need to be rendered, painted, etc.

I also found that the `await this.updateComplete;` (line 140) is not necessary in this method... because there isn't anything in this method relies on previous changes being rendered, painted, etc.

components/alert/alert-toast.js
dbatiste: ^ In experimenting, I found this is necessary in order for Lit to reflect the property update to the `open` attribute.  Otherwise, it would not be needed.

components/alert/alert-toast.js
AllanKerr: The `await this.updateComplete` is only unneeded if we move the `this._state = states.PREOPENING` into the `requestAnimationFrame`. I'm not entirely sure why but Lit seems to skip reflecting the `_state` if its updated in `_openChanged` without the `await this.updateComplete` causing the attribute go straight from `closed` to `opening`.

I want to avoid moving `states.PREOPENING` into the `requestAnimationFrame` because then there is a brief period where the component is in `CLOSED` state but is actually opening. This causes the component to end up open if `open` is quickly toggled from `false` to `true` then back to `false`.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+			type: { type: String, reflect: true },
+			_state: { type: String, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: block;
+			}
```
dlockhart, line 36: Normally I'd say we should include the `:host([hidden]) { display: none; }` styles here since we're overriding the default display... but it's hidden by default and the hidden-ness is controlled by the `open` attribute instead. So maybe it's fine to exclude it.

components/alert/alert-toast.js
dlockhart: Agreed, and it looks like that would clean up a lot of the CSS selectors too.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+			this._open = val;
+			this.requestUpdate('open', oldVal);
+			this._openChanged(val);
+		}
+	}
+
+	render() {
+		return html`
+			<div class="d2l-alert-toast-container" @transitionend=${this._onTransitionEnd}>
+				<d2l-alert class="d2l-alert-toast" type=${ifDefined(this.type)} @d2l-alert-closed=${this._onCloseClicked} button-text=${ifDefined(this.buttonText)} ?has-close-button=${!this.hideCloseButton} subtext=${ifDefined(this.subtext)}>
```
dlockhart, line 119: Nit: I think we typically try to wrap attribute values in quotes... I don't think it's necessary but seems to be the convention we've used thus far.

components/alert/alert-toast.js
AllanKerr: It looks like moving `state` to be on the `div` rather than reflected removes the need for `await this.updateComplete`. Almost feels like there is a Lit bug or gotcha I'm not aware of there related to reflected properties.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
...
+					<slot></slot>
+				</d2l-alert>
+			</div>
+		`;
+	}
+
+	_onCloseClicked(e) {
+		e.target.removeAttribute('hidden');
```
dlockhart, line 127: What's `e.target` in this context, the close button itself?

components/alert/alert-toast.js
```diff
@@ -0,0 +1,183 @@
+import '../button/button-icon.js';
+import '../button/button-subtle.js';
+import '../colors/colors.js';
```
dlockhart, line 3: I don't believe `button-icon`, `button-subtle` or `colors` are used directly by the toast alert?

components/alert/alert-toast.js
AllanKerr: It is the `d2l-alert`. It sets itself to `hidden` when the close button is clicked but this means the close transition isn't visible.

dlockhart: Oh wow, so this is like undoing what the alert did to itself? So what we'd really want is a way to tell the alert to have a close button but to not actually do anything when it's clicked. Then the toast would still get the event but could handle the "closing" itself.

AllanKerr: Its definitely a bit weird, it feels a bit off that the alert actually closes itself when the button is clicked. There didn't feel like a great way to prevent it from closing itself given the current design of `d2l-alert` unless we add a `no-close` property but that felt bad so I carried this through from the Polymer version.

dlockhart: I'm OK with keeping it as-is for now, but I think it's a good candidate for cleanup later.

dbatiste: What if `d2l-alert` provided a slot, and the default content would be the button that exists.  The toast would put it's own button in that slot, which it wire up to.  That way `d2l-alert` can have it's default behavior, and we can override it for toast without dirtying up the API.  Thoughts??

dlockhart: Another similar thought: what if when alert fired the `d2l-alert-close` event, it waited to see if someone called `preventDefault()` on it? If that was called, it wouldn't actually go through with the close. That's more what the spirit of `preventDefault` is for, ya?

AllanKerr: I like it, that feels like a very appropriate use of `preventDefault`!

dbatiste: Yeah that would be another clean option.  So alert would be updated to listen for it's own `d2l-alert-close` event... and then in a `setTimeout` or `rAF` it would check `e.defaultPrevented` and proceed with closing if depending.

dlockhart: I think it's even cleaner than that... events bubble up synchronously. So it could literally do this:
```
const e = new CustomEvent(
    'd2l-alert-closed', { bubbles: true, cancelable: true, composed: true }
);
this.dispatchEvent(e);
if (!e.defaultPrevented) {
    this.hidden = true;
}
```

AllanKerr:
Making a few random changes that I'll be backing out in my most recent commit to try and debug why tests are only failing in CI so if you see anything weird in the last commit it can be ignored for now.

components/alert/test/alert-toast.test.js
```diff
@@ -0,0 +1,147 @@
...
+			expect(toast._state).to.equal('preopening');
+			await aTimeout(75);
+			expect(toast._state).to.equal('opening');
+		});
+
+		it('should transition from opening to open on animation end', async() => {
+
+			toast.open = true;
+			await toast.updateComplete;
+			toast._state = 'opening';
```
dbatiste, line 61: I am a bit apprehensive about defining internals of the component states.  This _kind of_ tests the animation, but not really.  The internals are subject to change, but unfortunately we don't have a way (other than manual testing) to _really_ test the animation is working as expected.

AllanKerr: I had a lot of internal debate about whether to add them or not because it didn't feel great to be poking at internals but I ended up going for it because it felt worse to the state transitions at all. If we'd rather not poke at externals to this extent I can remove them.

components/alert/alert-toast.js
```diff
@@ -0,0 +1,177 @@
...
+		if (oldVal !== val) {
+			this._open = val;
+			this.requestUpdate('open', oldVal);
+			this._openChanged(val);
+		}
+	}
+
+	render() {
+		return html`
+			<div class="d2l-alert-toast-container" state="${this._state}" @transitionend=${this._onTransitionEnd}>
```
dlockhart, line 113: Technically this is invalid HTML, and I'd be worried that browsers may ignore it since `state` is not a valid attribute on `div`s. Probably it'll work, but might be safer to use `data-state`.

components/alert/alert.js
```diff
@@ -125,10 +125,10 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 	}
 
 	close() {
-		this.hidden = true;
-		this.dispatchEvent(new CustomEvent(
-			'd2l-alert-closed', { bubbles: true, composed: true }
-		));
+		const event = new CustomEvent('d2l-alert-closed', { bubbles: true, composed: true, cancelable: true });
+		if (this.dispatchEvent(event)) {
```
dlockhart, line 129: Oh nice, didn't know the dispatch returned false if default was prevented -- cool!

PR #530, opened 2020-05-06
Add a lint rule to catch console logging

Scy-D2L:
- Lint fails when code contains `console.log(...)`
- Still allows `console.warn()` and `console.error()`
- This rule ignores our demo pages where we're often console logging on purpose
- Has an exception for getLocalizeResources which seems to be logging deliberately only on debug


.eslintrc.json
```diff
@@ -4,6 +4,15 @@
     "Prism": false
   },
   "rules": {
-    "linebreak-style": ["error", "unix"]
-  }
+    "linebreak-style": ["error", "unix"],
+    "no-console": ["error", { "allow": ["warn", "error"] }]
```
dbatiste, line 8: For this part, I think we need to fix [the config that's being extended](https://github.com/Brightspace/eslint-config-brightspace/blob/master/index.js).

.eslintrc.json
```diff
@@ -4,6 +4,15 @@
...
+  },
+  "overrides": [{
+    "files": [
+      "**/demo/**.html",
```
dbatiste, line 12: This is pretty cool - I did not know we could do this.  :)

.eslintrc.json
dbatiste: That is, I think `eslint:recommended` ruleset changed within es-lint, and we didn't realize.

dlockhart: Agreed, and as long as we do a minor release because it's still `<0` it'll be treated like a major one.

PR #531, opened 2020-05-06
Add focus ordering and rules to generic list item layout

zinabat:
## **Context**
[US115380](https://rally1.rallydev.com/#/detail/userstory/382436537132?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#Clickable-Areas-and-Focus)

Forces the specified focus order for the action areas defined earlier by `d2l-list-item-generic-layout`. The order is always as follows:

1. `content-action`
2. `outside-control-action`
3. `control-action`
4. `actions`

The content area should have its focus removed and action should not occur.

**New helper:** `getNextAncestorSibling`
The new helper takes a node and an optional predicate and returns the next sibling at the closest DOM level it occurs at least one level up. This helper existed as a private method previously, and has been migrated.

This helper was added to allow removal of focus from any items that live within the content area.

**Default width** added for `outside-control` based on previous discussions.

**Result**
Screen Shot 2020-05-12 at 5 23 13 PM

## **Quality**
* New unit tests for `getNextAncestorSibling`
* Manually tested
  - [ ] Credge
  - [x] Legacy Edge (see known issues)
  - [x] Chrome
  - [x] Safari
  - [x] Firefox
* To test, pull the branch and run `npm start`, then visit http://localhost:8000/components/list/demo/list-item-sample.html

**Known Issues**
* Focus redirection does not work in Legacy Edge due to an inability to capture events. This is not a functionality issue, so we've decided not to support this for now. Developers are the only users affected.


components/list/demo/list-item-sample.js
```diff
@@ -19,9 +20,28 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
...
+				display:block;
+				height:100%;
+			}
+			:host([href]) {
+				--d2l-list-item-content-text-color: var(--d2l-color-celestine);
+			}
+			a[href]:focus + .d2l-list-item-content,
+			a[href]:hover + .d2l-list-item-content {
+				--d2l-list-item-content-text-decoration: underline;
+			}
```
zinabat, line 33: These will not trigger correctly on Legacy Edge and Firefox for some reason. Any input as to why or other ways to do this are welcome.

components/list/list-item-generic-layout.js
```diff
@@ -56,18 +57,45 @@ class ListItemGenericLayout extends LitElement {
...
+		this._preventFocusFromNonActions();
+	}
+
 	render() {
 		return html`
+		<slot name="content-action"></slot>
+		<slot name="outside-control-action"></slot>
+		<slot name="control-action"></slot>
+		<slot name="actions"></slot>
+
```
zinabat, line 72: Specifying slot order this way forces the focus order by relying on the DOM structure.

components/list/list-item-generic-layout.js
```diff
@@ -56,18 +57,45 @@ class ListItemGenericLayout extends LitElement {
...
+			'control',
+			'outside-control'
+		];
+		for (const slot of slots) {
+			const content = this.querySelector(`[slot="${slot}"]`);
+			const focusables = content ? getAllFocusableDescendants(content) : [];
+			for (const focusable of focusables) {
+				// remove focus and click events from focusable items.
+				// Items requiring focus MUST be placed within an action area
+				focusable.setAttribute('tabindex', '-1');
```
zinabat, line 91: This was the cleanest way I could figure to prevent focus reliably - alternatives if this is not ideal are welcome

helpers/focus.js
```diff
@@ -14,6 +14,31 @@ const focusableElements = {
 	textarea: true
 };
 
+export function getAllFocusableDescendants(node, includeHidden = true) {
```
m6jones, line 17: Could you add this file to the helper readme. I noticed it wasn't there and since you added to it you get more work as a prize!

zinabat: Oh! Good point

components/list/demo/list-item-sample.js
m6jones: One solution `:host(:hover) a + .d2l-list-item-content` not sure how I feel about it

zinabat: Won't that trigger even if something like the `outside-control-action` is being hovered?

helpers/README.md
```diff
@@ -11,6 +11,16 @@ import { announce } from '@brightspace-ui/core/helpers/announce.js';
 announce('...');
 ```
 
+## AsyncStateEvent
```
zinabat, line 14: Moved for alphabetical-ness

components/list/demo/list-item-sample.js
m6jones: Yeah :( I knew there was a reason this was bad.

zinabat: I have fixed this using javascript event listeners. Not exactly the cleanest way to add styles, but it works.

components/list/demo/list-item-sample.js
```diff
@@ -30,17 +50,46 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
...
 				</div>
 			</d2l-list-item-generic-layout>
 		`;
 	}
+
+	_anchorEventHandler(event) {
+		const content = this.shadowRoot.querySelector(`#${this._contentId}`);
+		switch (event.type) {
+			case 'mouseover':
+				content.setAttribute('hover', '');
```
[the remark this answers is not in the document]
zinabat: Will do - I was hesitant about doing this in general; this is great suggestion.

components/list/demo/list-item-sample.js
```diff
@@ -30,17 +50,46 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
...
-				<div slot="control-action">${ this._renderCheckboxAction() }</div>
+				<div slot="control-action" aria-labelledby="${this._contentId}">${ this._renderCheckboxAction() }</div>
+				` : nothing }
+				${ this.href ? html`
+				<a slot="content-action"
+					class="d2l-list-item-link"
+					href="${this.href}"
+					aria-labelledby="${this._contentId}"
+					@mouseover="${this._anchorEventHandler}"
+					@mouseout="${this._anchorEventHandler}"
```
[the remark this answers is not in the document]
zinabat: Arrow functions are not allowed in the eslint, which was my only real reasoning for making the shared handler. Is it better to have four separate handlers in this case?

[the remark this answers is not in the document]
m6jones: The linter was hating on `@mouseenter="${() => this._hovering = true;}"` is that because we don't need the `${}` part? That would be awesome for speed.

components/list/list-item-checkbox-mixin.js
```diff
@@ -76,7 +77,13 @@ export const ListItemCheckboxMixin = superclass => class extends superclass {
 
 	_renderCheckboxAction(inner) {
 		return this.selectable ? html`
-			<div @click="${this._handleCheckboxActionClick}" class="d2l-checkbox-action" ?disabled="${this.disabled}">${inner}</div>
+			<label @click="${this._handleCheckboxActionClick}"
+				class="d2l-checkbox-action"
+				?disabled="${this.disabled}"
+				for="${this._checkboxId}"
+				tabindex="0">
```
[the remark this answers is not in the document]
zinabat: I agree this is a bit weird.

The way that the designs have defined the checkbox, because it lives in the `control` area rather than `control-action`, the input itself is not meant to receive focus - but your points make me want to confirm this with @m6jones. What do you think?

components/list/list-item-generic-layout.js
```diff
@@ -56,18 +57,67 @@ class ListItemGenericLayout extends LitElement {
 		`;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+
+		this._preventFocusFromNonActions();
```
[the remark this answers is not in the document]
zinabat: Your assumption that this is to prevent misuse is correct. It's also for things like the above, where the action for the checkbox is independent from the checkbox itself, thus the checkbox theoretically should not have focusability.

I agree that the DOM walking is expensive and not ideal. I think I forgot about the `focusin` event and assumed that there didn't exist a bubbling event when one of the children of the non `action` slots receives focus/is about to. 🤷 

I will play with this; listening for the `focusin` event from the slot level should work and be infinitely more efficient if I can simply redirect focus to the next expected focusable item.

components/list/demo/list-item-sample.js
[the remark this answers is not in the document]
dbatiste: I see, the rationale can be found [here](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/no-template-arrow.md).  I was wondering if that were the case... indeed.   This is just a sample, but if it were me I would define 4 different handlers.  Once the `_hovering` and `_focusing` properties are defined, then handlers just become:
```
_handleMouseEnter() { this._hovering = true; }
_handleMouseLeave() { this._hovering = false; }
etc.
```

components/list/demo/list-item-sample.js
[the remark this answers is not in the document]
zinabat: Any pros/cons for classes vs. attributes mapped to these properties?

components/list/list-item-generic-layout.js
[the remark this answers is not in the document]
dbatiste: Cool, let me know how it goes.  :)

Note that `focus` event can be captured on the container even though it doesn't bubble.

With the standard capture (if supporting IE11) in `firstUpdated`:
```
.addEventListener('some-event', () => { ... }, true);
```
With standard options (if not supporting IE11)  in `firstUpdated`:
```
.addEventListener('some-event', () => { ... }, { capture: true });
```
With Lit directly in `render` ([ref to doc](https://lit-html.polymer-project.org/guide/writing-templates#add-event-listeners)):
```
constructor() {
    super();
    this._clickHandler = {
        // handleEvent method is required.
        handleEvent(e) { 
            console.log('clicked!');
        },
        // event listener objects can also define zero or more of the event 
        // listener options: capture, passive, and once.
        capture: true,
    }
}
render() {
    return html`<button @click=${this._clickHandler}>Click Me!</button>`;
}
```
^ While we typically always prefer to wire-up directly in the template with the `@`, I think this would be the first time using this approach of the handler **object**.

components/list/demo/list-item-sample.js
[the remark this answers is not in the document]
m6jones: They make it seem so dire. It isn't every time render happens. It is every time that part gets render. Even then, the defaultRenderProcessor might cache it. Anyways, easier to define the smaller method then going into the lower levels and check it out.

components/list/list-item-checkbox-mixin.js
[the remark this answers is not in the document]
m6jones: Yup this is an issue. So focus should be brought to checkbox. I got so wrapped up in my action naming. 

I really wanted to treat all the areas the same. But that seems overly messy. I also rather the checkbox mixin not to worry about placing tabindex. 

So my thoughts right now is to change things up a bit and set focus based on what is in the control areas. Treat the content area the same as you are doing now. 

I would still prefer all the areas get treated the same.

components/list/list-item-generic-layout.js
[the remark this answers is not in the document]
zinabat: Capturing is totally new to me - thanks for this. That's awesome.

components/list/list-item-checkbox-mixin.js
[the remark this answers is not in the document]
zinabat: I've changed things to only prevent focus on the `content` area, as the dragger will also have focusability.

components/list/list-item-generic-layout.js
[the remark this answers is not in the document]
zinabat: I've updated this to use a captured `click` handler and `focus` handler attached to the slot.

components/list/list-item-checkbox-mixin.js
```diff
@@ -76,7 +71,12 @@ export const ListItemCheckboxMixin = superclass => class extends superclass {
 
 	_renderCheckboxAction(inner) {
 		return this.selectable ? html`
-			<div @click="${this._handleCheckboxActionClick}" class="d2l-checkbox-action" ?disabled="${this.disabled}">${inner}</div>
+			<label
+				class="d2l-checkbox-action"
+				?disabled="${this.disabled}"
+				for="${this._checkboxId}">
+				${inner}
+			</label>
```
zinabat, line 79: Changing this to a `label` makes the click work natively in browsers, so need for a handler.

components/list/list-item-generic-layout.js
```diff
@@ -56,16 +60,51 @@ class ListItemGenericLayout extends LitElement {
 		`;
 	}
 
+	constructor() {
+		super();
+		this._preventFocus = {
+			handleEvent(event) {
+				event.preventDefault();
+				// target content slot only for now - can add others later
+				// TODO findComposedAncestor should be replaced with event.path || event.composedPath() when supported
```
dbatiste, line 69: I think `composedPath` should be usable.  We use it in a few places, usually to get at the real composed target.  It would have been polyfilled for IE11, and we won't be supporting IE11 very shortly.  That being said, I think what you have is ok as well.

components/list/list-item-generic-layout.js
```diff
@@ -56,16 +60,51 @@ class ListItemGenericLayout extends LitElement {
...
+					node.nodeName === 'SLOT' && ['content'].includes(node.name)
+				);
+				const ancestorSibling = getNextAncestorSibling(slot);
+				const next = getNextFocusable(ancestorSibling, true);
+				// related target is often on the parent
+				const related = getFirstFocusableDescendant(event.relatedTarget);
+				if (!event.relatedTarget) {
+					next.focus();
+				} else {
+					if (event.relatedTarget === next || related === next) {
```
dbatiste, line 80: Be careful here... given a scenario where focus is being moved from this `button`, say backward tabbing...
```
<my-elem>
    --shadowRoot
        <button>
```
The relatedTarget will be automatically retargeted by the browser to be the element in the context of the handler,  `my-elem`, whereas `next` (in this case) will be the composed DOM element that had the focus, `button`, and this if-statement won't catch it.  It's an edge case, but it could easily come up if people are wrapping focusables in their own components shadowRoots and putting them into the content slot.  If this doesn't get caught it's not a huge deal, but it could make the focus order confusing for screen reader users.

dbatiste: Or, could keyboard users get into a focus loop.  Brain 🌫️ ... need more ☕ .

zinabat: Yeah, this was my solution to a focus loop that happened because of what you said above - this _should_ be caught by the calculated `related`, since it refers to the composed item. This happens on the demo page, where `relatedTarget` refers to `d2l-button` but the focusable item is the `button` that lives in the shadow.

components/list/list-item-generic-layout.js
zinabat: Will leave it for now with the todo

PR #532, opened 2020-05-07
Modify page template to render RHP as BP

stevenao:
image


dlockhart:
Sorry I don't really understand what's trying to be done here -- what's "RHP" and "BP"?

stevenao:
@dlockhart To increase usability while we wait for the full mobile experience implementation, Joseph wants to render the right hand panel as a bottom panel in mobile for now.

dlockhart:
Thanks I understand now. The rough plan @wongvincent and I had for mobile was to just adjust the CSS grid layout to flip things from column-based to being stacked. I've pushed a change that accomplishes this, giving the main content 2/3 and the secondary 1/3. That can be tweaked if necessary.

PR #535, opened 2020-05-07
added d2l-list-item-drag-mixin

iFun:
[US115495
](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?qdp=%2Fdetail%2Fuserstory%2F383973313956)
[Designs](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#d2l-list-item-drag-mixin)

Screen Shot 2020-05-07 at 14 58 20

This is early draft as I am still not sure how some the methods works together, and which implementations need to be added


components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+
+	constructor() {
+		super();
+		this._itemDragId = getUniqueId();
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		if (this.key === undefined) {
+			this.draggable === undefined;
```
zinabat, line 22:
```suggestion
			this.draggable = undefined;
```

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+	// TODO
+	_copy() {
+
+	}
+
+	// TODO
+	_removeCopy() {
+	}
+
+	_dropHanlders() {
```
zinabat, line 99:
```suggestion
	_dropHandlers() {
```

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	_dragExit() {
+		this._dispatchDragEvent(null);
+		this._dragStopHanlders();
+	}
+	_dragEnter() {
+		this._dragStartHanlders();
+	}
+	_dragStartHanlders() {
```
zinabat, line 114:
```suggestion
	_dragStartHandlers() {
```

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+			}));
+		}
+	}
+
+	_renderDragHandle() {
+		return this.draggable ? html`
+			<div
+				id="${this._itemDragId}"
+				class="d2l-list-item-drag"
+				.draggable="${this.draggable}"
```
zinabat, line 47: Are we using this attribute anywhere on the drag handle itself?

m6jones: Nope, I don't think so.

This should be `?draggable="${this.draggable}"` I feel like Zina's point is this doesn't even need to be here though.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+
+	// TODO
+	_removeCopy() {
+	}
+
+	_dropHandlers() {
+		this._dispatchDragEvent(null);
+	}
+
+	_dragOverHanlders() {
```
zinabat, line 103: Couple more 🍬 
```suggestion
	_dragOverHandlers() {
```

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,120 @@
...
+		this._dispatchDragEvent(null);
+		this._dragStopHanlders();
+	}
+	_dragEnter() {
+		this._dragStartHanlders();
+	}
+	_dragStartHandlers() {
+		this.moveBefore();
+	}
+	_dragStopHanlders() {
```
zinabat, line 117:
```suggestion
	_dragStopHandlers() {
```

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	constructor() {
+		super();
+		this._itemDragId = getUniqueId();
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		if (this.key === undefined) {
```
zinabat, line 21:
```suggestion
		if (!this.key) {
```
This should handle all falsey cases to maintain consistency with the `checkbox-mixin`

m6jones: The question though is `0` an acceptable key?

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	moveBefore(itemKey) {
+		if (itemKey === undefined) {
+			return;
+		}
+		if (itemKey === null) {
+			this._dispatchDragEvent(null);
+		}
+		this._dispatchDragEvent(itemKey);
```
m6jones, line 69: If itemKey is null then `this._dispatchDragEvent(itemKey);` === `this._dispatchDragEvent(null);`.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+
+	_renderDragHandle() {
+		return this.draggable ? html`
+			<div
+				id="${this._itemDragId}"
+				class="d2l-list-item-drag"
+				@drop="${this._dropHandlers}"
+				@dragover="${this._dragOverHandlers}"
```
m6jones, line 49: These two events won't be connected here and can be removed for now. Future story will cover all the drop and drag over stuff. It will need to spawn something over the entire object the action slot.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,103 @@
...
+		` : nothing;
+	}
+
+	_renderDragAction(inner) {
+		return this.draggable ? html`
+			<div @click="${this._handleDragActionClick}" class="d2l-list-item-drag-action">${inner}</div>
+			` : nothing;
+	}
+
+	moveBefore(itemKey) {
```
AKobets, line 60: implementation of moveBefore and moveAfter almost identical. If there should be no difference, then duplication should be reduced.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		this._dispatchDragEvent(itemKey);
+	}
+
+	moveAfter(itemKey) {
+		if (itemKey === undefined) {
+			return;
+		}
+		if (itemKey === null) {
```
AKobets, line 72: null handling seems redundant because it still calls _dispatchDragEvent

components/list/list-item-drag-mixin.js
iFun: hmm the implementation is not done yet, so i think it will not be duplicate after

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,100 @@
...
+			this.draggable = false;
+		}
+	}
+
+	_dispatchDragEvent(detail) {
+		this.dispatchEvent(new CustomEvent('d2l-list-item-position', {
+			detail: detail,
+			bubbles: true
+		}));
+		return true;
```
dbatiste, line 31: Is this helper intended to return `true`? Usually we would not return anything, but every once in a couple of places we may return the result of `dispatchEvent` (bool) which indicates whether the default behavior is prevented (by one of the handlers calling `e.preventDefault()`.

m6jones:
The linting rules about position of methods is working again so that is the cause of the errors

PR #536, opened 2020-05-08
calendar: remove grid

margaree:
Context: iOS + Voiceover does not properly read table with `role="grid"`. It also does not seem to support anything `aria-*` in a table.

Notes:
- `selected` has been moved into a description which uses `d2l-offscreen` instead of `aria-label` because of iOS + voiceover not reading `aria-*` in table.
- I added comments to note why certain decisions were made
- If `role="gridcell"` is not on the `td`, NVDA will always says "Not selected" even if `aria-selected` is true. It does now read the selected state twice. This seems not great but not sure what to do otherwise.


components/calendar/calendar.js
```diff
@@ -419,9 +476,11 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 				const year = day.getFullYear();
 				const month = day.getMonth();
 				const date = day.getDate();
+				const description = `${formatDate(day, {format: 'medium'})}. ${selected ? this.localize('selected') : this.localize('notSelected')}`;
+				// role="gridcell" used for NVDA selected behavior to work properly
+				// d2l-offscreen used instead of aria-label for consistency in announcement across screen readers
 				return html`
 					<td
-						aria-label="${formatDate(day, {format: 'medium'})}"
 						aria-selected="${selected ? 'true' : 'false'}"
```
dlockhart, line 484: Does `aria-selected` still provide value? Curious why the combination of it and `role="gridcell"` is still needed given that the `grid` role isn't anymore. I see the comment about NVDA... what weird stuff is it doing?

margaree: Yeah so I've found with NVDA it needs both `aria-selected` and `role="gridcell"` because without both those things it will read "Not Selected" on every cell (even if both `aria-selected` and `role` are not present). I'm guessing this is because they have a tabindex and are "interactive" but I'm not sure.

dlockhart: That seems bizarre to me... because without those things isn't it just a normal table? Maybe it's the `role="columnheader"`?

margaree: Sadly doesn't seem to be. I made a minimal codepen with a simple table and no extra roles, just tabindex, and experimenting with role and aria-selected, and it's still behaving like this.

dlockhart: Crazy! Silly NVDA.

PR #540, opened 2020-05-11
Add d2l-expand-collapse component to create collapsible areas

AllanKerr:
This is a componentized version of our [d2l-dom-expand-collapse](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-expand-collapse.js) that I'm looking for some input on if we want this added as generic component in core.

# Goals
1. Short-term:
    1. Use this to add expand / collapse support to `d2l-alert`.
1. Long-term:
    1. Update `d2l-more-less` to use this internally.
    1. Replace the 2 LMS usages of the `d2l-dom-expand-collapse` with this and delete `d2l-dom-expand-collapse` entirely.


components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,132 @@
...
+	get expanded() {
+		return this._expanded;
+	}
+
+	set expanded(val) {
+		const oldVal = this._expanded;
+		if (oldVal !== val) {
+			this._expanded = val;
+			this.requestUpdate('expanded', oldVal);
+			this._expandedChanged(val);
```
dlockhart, line 80: Could we test out our next theory that if we're using reflected attributes that we can just rely on the `attributeChangedCallback` to be notified of changes (and avoid setters/getters)?

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,132 @@
...
+			:host {
+				display: block;
+			}
+
+			:host([hidden]) {
+				display: none;
+			}
+
+			.d2l-expand-collapse-container {
+				height: 0px;
```
dlockhart, line 36: Nit: no units on zero-value

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,132 @@
...
+			this._expanded = val;
+			this.requestUpdate('expanded', oldVal);
+			this._expandedChanged(val);
+		}
+	}
+
+	firstUpdated() {
+		super.firstUpdated();
+
+		const content = this.shadowRoot.querySelector('.d2l-expand-collapse-content');
```
dlockhart, line 87: Nit: There's a `_getContent` helper -- should either use it or remove it.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,132 @@
...
+		this._resizeObserver = new ResizeObserver(this._onContentResize);
+		this._resizeObserver.observe(content);
```
dlockhart, line 89: If the observer ends up being buggy, we could set the height to `auto` once it's fully expanded. i.e. only calculate the height when it initially goes from collapsed -> expanded to do the animation. Then when the animation finishes,  remove the fixed height.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	_expandedChanged(val) {
+		if (val) {
+			this._state = reduceMotion ? states.EXPANDED : states.EXPANDING;
+			this._updateHeight();
+		} else {
+			this._state = reduceMotion ? states.COLLAPSED : states.COLLAPSING;
+			this._height = null;
+		}
```
dlockhart, line 112: I feel like this component (or maybe it should be a mixin) would become really powerful if it could handle `aria-expanded` for us. The W3C [has an examplle](https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element) of how this should work. But that would imply that the component/mixin is aware of the "opener" (the thing that when clicked toggles the state).

Ideally in my mind, the reusable stuff that'd be valuable for other components:
- The animation of the content area (which you have)
- Wire-up to an opener that toggles things and sets `aria-expanded` on it
- Ability for the opener to be anything and styled in any way

Basically... if we look at more/less and at the design for collapsible alerts and the design for the FACE panels... if we could refactor whatever they share into a mixin/component, what would it look like?

AllanKerr: I did explore making it a mixin but the big drawback there is that we either need to either manipulate the DOM directly by adding styles outside of render or create methods people would need to call themselves as part of render, both of which felt bad if people are going to be using this outside of core.

How would you feel about creating something like a `d2l-expand-collapse-opener` that behaves in basically the same manner as the `d2l-dropdown-opener`? Then that could be used along with this to accomplish all of those goals.

AllanKerr: Or I guess if we are following the `dropdown` model for adding openers it would be something along the lines of renaming this to `d2l-expand-collapse-content` then adding `d2l-expand-collapse` for the opener logic.

dlockhart: Yeah it's complicated eh. Like the opener needs to be able to be literally anything... typically a button of some kind of with various stylings and sometimes the whole "bar" is the click target. But I don't believe we need to have `id` references... unless the `aria-controls` needs to be used.

AllanKerr: I think we'd need some sort of `id` reference just as some way to keep the opener and content in sync. At least it feels like it'd be the simplest option if we could just do something like `<d2l-expand-collapse opener="my-opener-id">` then it could find the opener element and add click handlers and set the `aria-expanded` attribute.

It wouldn't have the same shadow boundary ID issues that tooltip has since the actual linking would be done using Javascript.

components/expand-collapse/expand-collapse.js
AllanKerr: `attributeChangedCallback` does get called but it doesn't appear to be correctly triggering a re-render when `_state` and `_height` get set. 😕

components/expand-collapse/expand-collapse.js
dlockhart: Was just thinking we'd do what dropdown does where it uses a CSS class applied to an element somewhere inside a slot.

AllanKerr: The reason I don't think that works in this case is because that also has a dependency on the `d2l-dropdown-content` being inside the same slot. That works for dropdowns because they are absolutely positioned so where the `d2l-dropdown-content` is in the DOM tree doesn't really matter.

With an expand collapse area that becomes an issue because there is no guarantee that the opener and the collapsible section are next to each other in the DOM:
Capture

dlockhart: Yeah true! So in that case, some 3rd party is going to need to "marshall" the click event from the opener/closer (which may be 2 different elements like in this example) and notify the expand/collapse container. It _could_ pass the opener to the container as a DOM node and the container could apply some ARIA to it... but at that point we're getting diminishing returns.

Similarly, in theory the container could expand or collapse based on another party interacting with it, and that event would need to be marshalled back to the opener/closer.

So essentially I think I've come around to the container itself being the interesting bit of reusable UI here I think!

components/expand-collapse/expand-collapse.js
dbatiste: Perhaps [this bug](https://github.com/Polymer/lit-element/issues/965).

components/expand-collapse/expand-collapse.js
dbatiste: 👀 ... 🤔

components/expand-collapse/demo/expand-collapse.html
```diff
@@ -0,0 +1,57 @@
...
+
+<body unresolved>
+
+	<d2l-demo-page page-title="d2l-expand-collapse">
+
+		<h2>Default</h2>
+		<d2l-demo-snippet>
+			<template>
+				<d2l-button primary>Toggle</d2l-button>
+				<d2l-expand-collapse>
```
dbatiste, line 25: Another scenario worth checking is with `expanded` initially `true`.  Be careful that any delay caused by the snippet template being stamped and inserted isn't misleading. If in doubt, try it outside a snippet.  I'd expect not to see an animation in such case.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,140 @@
...
+	_onContentResize() {
+		if (!this.expanded) {
+			return;
+		}
+		this._updateHeight();
+	}
+
+	_updateHeight() {
+		const content = this._getContent();
+		this._height = content.scrollHeight;
```
dbatiste, line 136: Watch out here - I think `content` element will not exist initially and if initially expanded, this will error.

components/expand-collapse/expand-collapse.js
dbatiste: I think this is preferred (reverting to `auto`) - currently, if the browser is resized while open, the `height` property is updated.  I think it would be better to let the browser handle this natively.  

If we _do_ need the observer, we should try to make use of its [entries](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry) to avoid making measurements during such times where the user is resizing, etc. The observer provides this because it's such a common (costly) pattern.

AllanKerr: Reverting to auto is turning out to be surprisingly painful because transitions won't run to / from `auto. ` This makes the close transition flakey because the height needs to be computed then set to a specific value before collapsing which requires waiting for a timeout for the recalc to occur, then setting the height to 0. This works the majority of the time but if `expanded` is changed while the timeout is in progress then we run into race conditions. Still working to see if I can find a way around this.

components/expand-collapse/README.md
```diff
@@ -0,0 +1,37 @@
...
+</script>
+```
+
+**Properties:**
+
+- `expanded` (Boolean, default: `false`): Specifies the expanded/collapsed state of the content
+
+**Accessibility:**
+
+To make your usage of `d2l-expand-collapse` accessible, the `aria-expanded` attribute should be added to the element that controls expanding and collapsing section with `"true"` or `"false"` to indicate that the section is expanded or collapsed.
```
dlockhart, line 33: Could you provide a link to the [W3C docs](https://www.w3.org/TR/wai-aria/#aria-expanded) on the `aria-expanded` state here?

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
+
+const states = {
+	PRECOLLAPSING: 'precollapsing', // setting up the styles so the collapse transition will run
+	COLLAPSING: 'collapsing', // in the process of collapsing
+	COLLAPSED: 'collapsed', // fully collapsed
+	PREEXPANDING: 'preexpanding', // setting up the styles so the expand transition will run
+	EXPANDING: 'expanding', // in the process of expanding
+	EXPANDED: 'expanded', // fully expanded
+};
```
dlockhart, line 13: Question: what happens if someone does this in their initial markup?

```html
<d2l-expand-collapse expanded>Stuff in here</d2l-expand-collapse>
```

Does it animate right away? I think ideally we'd only animate user-initiated state changes. If they want the thing expanded initially, chances are they "saved" the state and are just keeping it in that same state -- like we do with widgets on the homepage.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+	updated(changedProperties) {
+		if (changedProperties.has('expanded')) {
+			this._expandedChanged(this.expanded);
+		}
+	}
+
+	render() {
+		const styles = { height: this._height };
+		return html`
+			<div class="d2l-expand-collapse-container" data-state=${this._state} @transitionend=${this._onTransitionEnd} style=${styleMap(styles)}>
```
dlockhart, line 80: Nit: quotes around attributes

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+	_getContent() {
+		return this.shadowRoot.querySelector('.d2l-expand-collapse-content');
+	}
+
+	async _getUpdateComplete() {
+		const fontsPromise = document.fonts ? document.fonts.ready : Promise.resolve();
+		await super._getUpdateComplete();
+		/* wait for the fonts to load because browsers have a font block period
+		where they will render an invisible fallback font face that may result in
+		improper height calculations if the expand-collapse starts expanded */
```
dlockhart, line 130: See my comment up higher, but if it starts expanded I'm hoping we can avoid the transition altogether, which would mean an initial height of `auto` so no need to wait for fonts.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+				await new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r)));
+				if (this._state === states.PRECOLLAPSING) {
+					this._state = states.COLLAPSING;
+					this._height = '0';
+				}
+			}
+		}
+	}
+
+	_getContent() {
```
dlockhart, line 121: Nit: since this is only ever used in `_expandedChanged`, there isn't a ton of value in pulling it out into a separate method and polluting the element with a "private" (but not really everyone will see it) method.

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+		if (this._state === states.EXPANDING) {
+			this._state = states.EXPANDED;
+			this._height = 'auto';
+		} else if (this._state === states.COLLAPSING) {
+			this._state = states.COLLAPSED;
+		}
+	}
+
+}
+customElements.define('d2l-expand-collapse', ExpandCollapse);
```
dlockhart, line 144: In terms of the component name... wondering what folks think. This component will end up being paired with an opener and likely wrapped in a visible "box" of some kind, so maybe its name should be a little more narrow in scope? Example: `d2l-expand-collapse-content`

components/expand-collapse/expand-collapse.js
```diff
@@ -0,0 +1,144 @@
...
+				const content = this._getContent();
+				this._height = `${content.scrollHeight}px`;
+				await this.updateComplete;
+				await new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r)));
+				if (this._state === states.PRECOLLAPSING) {
+					this._state = states.COLLAPSING;
+					this._height = '0';
+				}
+			}
+		}
```
dlockhart, line 118: What do folks think about firing a `d2l-expand-collapse-state-change` (or similarly named) event when it completes its transition from expanded -> collapsed or vice versa? I can imagine consumers will want to know so that they can swap icons on openers, text, etc. For example if more/less used this it would need to switch its text and icon.

components/expand-collapse/test/expand-collapse.visual-diff.html
```diff
@@ -0,0 +1,46 @@
...
+
+<body class="d2l-typography">
+	<div class="visual-diff">
+		<d2l-expand-collapse id="collapsed">
+			<ul>
+				<li>Coffee</li>
+				<li>Tea</li>
+				<li>Milk</li>
+			</ul>
+		</d2l-expand-collapse>
```
dlockhart, line 33: Interesting that the blue border still appears even when it's collapsed. It makes sense given it the internal container that gets a `display: none` on it, but as a consumer of this I wouldn't have expected it. I can't think of a super clean way to fix this without reflecting that `_state` property up to the host. Hmm...

AllanKerr: I think having the border still show up will work well. It aligns with FACE's use of expand collapse where they have a line where the expand collapse section is when collapsed. Implementing it like this means it can be done with an attribute selector: 
Capture

It will also make it a lot easier to use with `d2l-more-less` since that component doesn't fully collapse by default, it leaves 4 ems visibile.

components/expand-collapse/expand-collapse.js
AllanKerr: I was thinking that could be done in the same callback that toggles with `expanded` attribute but I certainly don't think it'd hurt to add an event, there are probably some cases it would come in handy.

components/expand-collapse/expand-collapse.js
AllanKerr: I like that, it aligns it a bit more closely with dropdown.

components/expand-collapse/expand-collapse.js
AllanKerr: That's a good call, I think I should be able to make that work so there is no animation. I'll give look into it. Right now it works but if the page loads quickly you do see a bit of a transition.

components/expand-collapse/expand-collapse.js
dlockhart: I can also see a future where some of the wrapper components -- like the one FACE is using -- also get pulled into core and could live in this directory. So the directory could still be `expand-collapse` but then we'd have a few bits and bobs in here for people to use.

components/expand-collapse/expand-collapse.js
dlockhart: I'm trying to think of a scenario where we'd want the animation initially and I'm coming up empty. It's always going to look weird and jarring and it'll force a bunch of repaints and cause other content below it to move and reflow.

components/expand-collapse/README.md
```diff
@@ -0,0 +1,37 @@
...
+<script type="module">
+  import '@brightspace-ui/core/components/expand-collapse/expand-collapse.js';
+</script>
+
+<button>Toggle</button>
+<d2l-expand-collapse>
+  <p>My expand collapse content.</p>
+</d2l-expand-collapse>
+
+<script type="module">
```
dbatiste, line 17: I don't think this script (or the button HTML) is really needed in the sample code.  The property is already listed below, so this doesn't _really_ add anything.  I would consider adding `expanded` attribute in the above HTML snippet though.

components/expand-collapse/demo/expand-collapse.html
dbatiste: I think it would be worthwhile adding a second snippet that has `expanded` initially set.

components/expand-collapse/expand-collapse.js
dbatiste: Yeah my current thinking is there should be no animation initially.

components/expand-collapse/README.md
AllanKerr: The only reason I included it was to try and give users an example of a correct usage with `aria-expanded` but I'm happy to trim it down since thats explained in the README.

components/expand-collapse/expand-collapse.js
dbatiste: Agreed.  I think an event could be useful.  An idea I've mulled over a little bit is to actually dispatch the event before the animation begins:
```
    'd2l-expand', {
    bubbles: true,
    detail: { expandComplete: /* promise that resolves when transition complete */  }
    }
));

    'd2l-collapse', {
    bubbles: true,
    detail: { collapseComplete: /* promise that resolves when transition complete */  }
    }
));
```
Or one event for both states could be fine too.  

It sorta solves the issue where a component dispatches and event, but not at the time when a consumer is interested, because it gives the consumer both.  I think there are a couple places were we dispatch an event at the beginning (ex. so that nest menu can translate at the same time as the dropdown height animates), but many times we are interested in when the animation is finished.

dbatiste: ^ the above mentioned problem is traditionally solved by dispatching different events for before/after.  The above approach is sorta cleaner, but not what consumers expect.

components/expand-collapse/README.md
dbatiste: Sorry, I skimmed over the aria part.  🤷‍♂️ Either way I guess.

components/expand-collapse/expand-collapse.js
dlockhart: That's a really good point, and I think consumers like more/less would absolutely want the event at the start. I like the pattern of including the promise in the detail.

PR #541, opened 2020-05-11
[increment minor] mobile view single scroll bar (#538)

stevenao:
I missed the chance to bump the version, so I am recreating the PR I previously merged.


dlockhart:
Merged & released for you, so should be published by the time you get in!

PR #543, opened 2020-05-13
input-date: Improve mobile experience

margaree:
I'll add comments on the bugs that each part fixes, but in general the fixes/improvements are:
- Scroll behavior - `input-text` should remain visible when dropdown is opened given that it has focus. Especially important on mobile
- Tapping icon would clear the `value`
- `input-text` width was often not wide enough on iOS

### DIFF
--- a/components/inputs/input-date.js
+++ b/components/inputs/input-date.js
@@ -192,7 +192,7 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	}
 
 	render() {
-		const formattedWideDate = formatISODateInUserCalDescriptor('2020-12-20');
+		const formattedWideDate = formatISODateInUserCalDescriptor('2323-12-23');
 		const inputTextWidth = `calc(${this._hiddenContentWidth} + 0.75rem + 3px)`; // text and icon width + paddingRight + border width + 1
 		const shortDateFormat = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
 
--- a/components/inputs/input-date.js
+++ b/components/inputs/input-date.js
@@ -192,12 +192,15 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	}
 
 	render() {
-		const shortDateFormat = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+		const formattedWideDate = formatISODateInUserCalDescriptor('2020-12-20');
 		const inputTextWidth = `calc(${this._hiddenContentWidth} + 0.75rem + 3px)`; // text and icon width + paddingRight + border width + 1
+		const shortDateFormat = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+
 		this.style.maxWidth = inputTextWidth;
+
 		return html`
 			<div aria-hidden="true" class="d2l-input-date-hidden-content">
-				<div><d2l-icon icon="tier1:calendar"></d2l-icon>${formatISODateInUserCalDescriptor('2020-12-20')}</div>
+				<div><d2l-icon icon="tier1:calendar"></d2l-icon>${formattedWideDate}</div>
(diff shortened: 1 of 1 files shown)


components/inputs/input-date.js
```diff
@@ -191,12 +192,15 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	}
 
 	render() {
-		const shortDateFormat = (this._dateTimeDescriptor.formats.dateFormats.short).toUpperCase();
+		const formattedWideDate = formatISODateInUserCalDescriptor('2020-12-20');
```
margaree, line 195: Fixes iOS width being too small

components/inputs/input-date.js
```diff
@@ -268,16 +271,16 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
...
 			this._setFormattedValue();
 
 			if (e.keyCode === 40) e.preventDefault();
 		}
 	}
 
-	async _handleChange(e) {
-		const value = e.target.value;
+	async _handleChange() {
+		const value = this._textInput.value;
```
margaree, line 283: Fixes the icon clearing the value (`mouseup` on `input-text` was triggered when `d2l-icon` was clicked but `icon` wouldn't have a value)

components/inputs/input-date.js
```diff
@@ -313,20 +316,25 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	_handleDropdownClose() {
 		this._calendar.reset();
 		this._dropdownOpened = false;
+		this._textInput.scrollIntoView({block: 'nearest', behavior: 'smooth', inline: 'nearest'});
 	}
 
 	_handleDropdownOpen() {
-		this.shadowRoot.querySelector('d2l-focus-trap').scrollIntoView({block: 'nearest', behavior: 'smooth', inline: 'nearest'});
+		if (!this._dropdown.openedAbove) this.shadowRoot.querySelector('d2l-focus-trap').scrollIntoView({block: 'nearest', behavior: 'smooth', inline: 'nearest'});
```
margaree, line 323: Scroll to the `input-text` since it has focus on initial dropdown opening. Don't scroll to show `focus-trap` if opened above since that results in jumpiness (scroll jumps to top of calendar dropdown content, then jumps down to input-text)

components/inputs/input-date.js
```diff
@@ -335,6 +343,7 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 		const date = getToday();
 		this._updateValueDispatchEvent(formatDateInISO(date));
 		this._dropdown.close();
+		this.focus();
```
margaree, line 346: I noticed that sometimes when clicking "Set to Today", the dropdown would close but `input-text` would not gain focus (which I think would be the behavior we'd want).

components/inputs/input-date.js
```diff
@@ -313,20 +316,25 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 	_handleDropdownClose() {
 		this._calendar.reset();
 		this._dropdownOpened = false;
+		this._textInput.scrollIntoView({block: 'nearest', behavior: 'smooth', inline: 'nearest'});
```
margaree, line 319: Sometimes (I think mostly noticed on iPhone) when the dropdown closes the input-text does not get scrolled to, and it seems like it should give that it has focus.

components/inputs/input-date.js
dlockhart: Ah so like the numbers can be wider than the characters in the format descriptor - crazy! Is there one number that's the widest? Like "3"? i.e. maybe `2323-12-23` might be wider.

margaree: Yeah so in the case of english the placeholder is M/D/YYYY so just having more characters I think causes the numbers to be wider. In Lato it looks like all numbers are the same width, but it could be in other languages with different fonts that that isn't true so I like the idea of using 3 - it seems wider.

PR #544, opened 2020-05-14
list item drag handle

m6jones:
STATEMENT:

The drag handle controls rendering the visual drag handle and the keyboard interactions that drag n drop will support.

BACKGROUND:

https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#d2l-drag-handle

ACCEPTANCE CRITERIA:

- Render the drag handle as shown above and implemented in the prototype version: d2l-activity-collection-editor-drag.js
- Sends events up to notify when it has entered keyboard mode.
- In keyboard mode, sends events based on key pressed
-- Right Arrow: commit change
-- Up Arrow: move previous
-- Down Arrow: move next
-- Enter/Space: commit move
-- ESC: cancel move
--  Home: first event
-- End: last event
-- Tab: Commit move and move to next element
-- Shift Tab: Commit move and move to previous element


components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
+
+import '../button/button-icon.js';
+import '../icons/icon.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { buttonStyles } from '../button/button-styles.js';
+import { getFirstFocusableDescendant } from '../../helpers/focus.js';
+
+const keyCodes = Object.freeze({
```
omsmith, line 8: I think at least ESC is used for focus-trap. Maybe it'd be good to have a key-codes module?

`export const ESC = 27;
`import { ESC } from '../../helpers/key-codes';`

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+				display: flex;
+				margin: 0.25rem;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.d2l-list-item-drag-handle-dragger-button,
+			.d2l-list-item-drag-handle-keyboard-button {
+				display: grid;
+				grid-auto-columns: auto;
```
omsmith, line 56: Isn't this the default?

zinabat:
Looks like the button's disabled state isn't set when its parent is disabled.
Screen Shot 2020-05-14 at 10 51 38 AM

components/list/list-item-drag-handle.js
m6jones: I am thinking the same thing. Because I had to code this up in 2 places just for this PR. The aria-grid story will also need keys.

dbatiste: I've considered this in the past.  The only reason I didn't was to avoid an extra import.  🤷‍♂️ I would not be opposed to it, and certainly there are a few places that define key codes for same purposes, as well as mocking key events in tests.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	constructor() {
+		super();
+		this.keyboardActive = false;
+	}
+
+	focus() {
+		const node = getFirstFocusableDescendant(this);
+		if (node) {
```
dbatiste, line 94: Micro-nit: in the olden days we always formatted statements like this, but in the past year or so we've relaxed that a bit in the interest of more succinct code.  Feel free to go `if (node) node.focus()` if you prefer.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+			node.focus();
+		}
+	}
+
+	render() {
+		return html`${this.keyboardActive ? this.renderKeyboardDragging() : this.renderDragger()}`;
+	}
+
+	renderDragger() {
```
dbatiste, line 103: Are these render helpers meant to be private?

components/list/list-item-drag-handle.js
m6jones: I did notice you start doing that - I'll change that as I prefer the shorter in some cases.

components/list/list-item-drag-handle.js
m6jones: They totally are suppose to be private.

components/list/list-item-drag-handle.js
m6jones: I'll double check and if it is i'll remove. It be surprising if it isn't default.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+		return html`
+			<button class="d2l-list-item-drag-handle-dragger-button" @click="${this._handleDeactiveKeyboard}" @keydown="${this._handleDeactiveKeyboard}" aria-label="${this.text}">
+				<d2l-icon icon="tier1:dragger"></d2l-icon>
+			</button>
+		`;
+	}
+
+	renderKeyboardDragging() {
+		return html`
+			<button class="d2l-list-item-drag-handle-keyboard-button" @blur="${this._onBlur}" @keydown="${this._handleActiveKeyboard}" aria-label="${this.text}">
```
dbatiste, line 113: @m6jones : Is there a reason we use `keydown` here instead of `keyup`?  Usually we prefer to use `keyup` for a11y.

@dlockhart : I know we went through and switched a bunch of our core components to use `keyup` instead since it's preferred.  Do you know the use-cases behind that preference? ... is it to allow cancelling the key somehow (and if so how do i, as a user, do it), is it something to do with modifier keys?

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+			case keyCodes.SPACE:
+			case keyCodes.RIGHT:
+				action = dragActions.save;
+				this.updateComplete.then(() => this.keyboardActive = false);
+				break;
+			default:
+				return;
+		}
+		this._dispatchAction(action);
+		e.preventDefault();
```
dbatiste, line 178: What is the behavior for a long list for a keyboard user?  Are they able to "drag" the item to a list item that is below the fold?

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+	cancel: 'keyboard-deactivate-cancel',
+	nextElement: 'next-element',
+	previousElement: 'previous-element'
+});
+
+class listItemDragHandle extends LitElement {
+
+	static get properties() {
+		return {
+			keyboardActive: { type: Boolean, attribute: 'keyboard-active', reflect: true },
```
dbatiste, line 38: Looks like this should be private...?

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	_onBlur() {
+		this.keyboardActive = false;
+		if (!this._tabbing) {
+			this._dispatchAction(dragActions.save);
+		}
+	}
+
+	_handleActiveKeyboard(e) {
```
dbatiste, line 143: This handler, for a native `button` expects keys to be pressed that users would not expect for that element.  I will am interested in how this will be communicated.  I think this is what you might have eluded to in our recent Slack conv...?

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -0,0 +1,44 @@
...
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+
+		it('simple', async function() {
+			const rect = await visualDiff.getRect(page, '#drag-handle');
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+		it('focus', async function() {
+			await visualDiff.resetFocus(page);
+			await focusMethod('d2l-list-item-drag-handle');
```
dbatiste, line 31: This could be... `await page.$eval('d2l-list-item-drag-handle', item => item.focus());`.

Not a big deal either way... 🤷‍♂️

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,196 @@
...
+			<button class="d2l-list-item-drag-handle-dragger-button" @click="${this._handleDeactiveKeyboard}" @keydown="${this._handleDeactiveKeyboard}" aria-label="${this.text}">
+				<d2l-icon icon="tier1:dragger"></d2l-icon>
+			</button>
+		`;
+	}
+
+	renderKeyboardDragging() {
+		return html`
+			<button class="d2l-list-item-drag-handle-keyboard-button" @blur="${this._onBlur}" @keydown="${this._handleActiveKeyboard}" aria-label="${this.text}">
+				<d2l-icon class="sub-icon" icon="tier1:arrow-toggle-up" @click="${this._dispatchActionUp}"></d2l-icon>
```
dbatiste, line 114: Looks like the `sub-icon` class is not being used.

components/list/list-item-drag-handle.js
dlockhart: It's not an accessibility thing AFAIK, but mouse down vs. up [definitely is](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-21/#252-pointer-cancellation-a).

I think it's mostly that `keydown` isn't actually a "key press". Like if I keydown on something and don't release it for a second or two, when should the action occur -- when I first press down, or when I release? I'd expect on release, which would mean listening for `keyup`.

m6jones: I'll change it to keyup.

dbatiste: You may or may not still need a `keydown` to `preventDefault`, depending on scrolling behavior (as per my other comment attached to that line).

components/list/list-item-drag-handle.js
m6jones: Yeah, so when people got to this part they were actually pretty good about using the arrow keys. Somethings like `tab` and `shift+tab` were never touched. Also, home and end weren't used either but I figured those would be useful for longer lists.

Yeah, in general Abbie is looking into the best way we can communicate the uses. Probably going to go with a tooltip but waiting on the designs.

components/list/list-item-drag-handle.js
m6jones: I need to test this.

m6jones: So I tested and it does scroll with the movement of the card. Although, I was testing in the prototype currently up on my site lord site. We should make a note to test this when we have a list that actually moves.

components/list/list-item-drag-handle.js
m6jones: I have a feeling this might need to be public. I have a method that will flip this into keyboard but honestly, I think there should be a better solution for this workflow and so I am going to make this private so when/if it comes up to be public we can see about another solution to the problem it is solving. 

The reason it is public right now is I copied my prototyped version of this component and there it required to be public. So similar the the `sub-icon` class this might be a relic of the prototype that doesn't need to be in.

components/list/test/list-item-drag-handle.visual-diff.js
m6jones: Kept it since refactoring the `keyboardActive` to private I used the `focusMethod` in another spot.

components/list/list-item-drag-handle.js
m6jones: I made the new file and updated the readme. I am instead of adding all the key codes at once I figured people can add them as the need them.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,197 @@
...
+				break;
+			case HOME:
+				action = dragActions.first;
+				break;
+			case END:
+				action = dragActions.last;
+				break;
+			case TAB:
+				action = e.shiftKey ? dragActions.previousElement : dragActions.nextElement;
+				break;
```
zinabat, line 138: I'm curious what the plan is for tabbing to the next item here. Currently, this creates a focus trap that prevents the checkbox or actions from being focused once the dragger is focused.

m6jones: This is only when keyboard mode is active. The designs have it so it goes to the next item in keyboard mode.

zinabat: But won't preventDefault always be triggered with the @keydown handler? Currently when I try to tab to each action, I get stuck on the dragger.

m6jones: Yes! That is a problem but not here in the code. Non-active keyboard should not have the `keydown` preventdefault.

m6jones: This has been fixed.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,204 @@
...
+	up: 'up',
+	down: 'down',
+	active: 'keyboard-active',
+	save: 'keyboard-deactivate-save',
+	cancel: 'keyboard-deactivate-cancel',
+	nextElement: 'next-element',
+	previousElement: 'previous-element'
+});
+
+class listItemDragHandle extends LitElement {
```
AKobets, line 21: Class name usually starts with capital letter

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,204 @@
...
+				this.updateComplete.then(() => this._keyboardActive = false);
+				break;
+			default:
+				return;
+		}
+		this._dispatchAction(action);
+		e.preventDefault();
+	}
+
+	_handleInactiveKeyboard(e) {
```
AKobets, line 156: would it be better name `_activateKeyboard` ?

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	_renderDragger() {
+		return html`
+			<button
+				class="d2l-list-item-drag-handle-dragger-button"
+				@click="${this._handleInactiveKeyboard}"
+				@keyup="${this._handleInactiveKeyboard}"
+				@keydown="${this._handleInactiveKeyDown}"
```
AKobets, line 180: why keydown handler is different? Does it have special meaning when user presses key DOWN?

m6jones: Yes, so `keydown` happens when a key is being pressed. The reason for this being different is because I want all keys but the ones I am defining on `keyup` to work as normal. However, I need to prevent the keys in the `keyup` to not doing anything since `enter` also activates from the `click` event. So I get a double activation which makes it flash back and forth.

components/list/list-item-drag-handle.js
m6jones: I like this name better for this method. But this current name follows the pattern for the ones above it which are nice. Is there a better name for the `_handleActiveKeyboard`?

components/list/demo/list-item-sample.js
```diff
@@ -94,6 +95,10 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
 	_handleMouseLeave() {
 		this._hovering = false;
 	}
+
+	_renderDragHandle() {
+		return html`<d2l-list-item-drag-handle ?disabled=${this.disabled}></d2l-list-item-drag-handle>`;
```
dbatiste, line 100: Nit: While Lit will handle with and without quotes around the attribute value, we've start to opt for trying to use quotes always.  Reason being is that it's closer to the HTML spec.  Minor thing, but will probably come up again in other PRs.

helpers/keyCodes.js
```diff
@@ -0,0 +1,10 @@
+export const DOWN = 'ArrowDown';
```
dbatiste, line 1: For the currently written exports, it should be possible to do:
```
export const
    DOWN = 'ArrowDown',
    END = 'End',
    ENTER = 'Enter',
    ESC = 'Escape',
    HOME = 'Home',
    LEFT = 'ArrowLeft',
    RIGHT = 'ArrowRight',
    SPACE = 'Space',
    TAB = 'Tab',
    UP = 'ArrowUp';
```
However, I am a bit on the fence in terms of exporting these unqualified constants.  The chance of collision is low, but certainly not zero.  Another approach, which I kind of prefer is:
```
export const keyCodes = {
    DOWN: 'ArrowDown',
    END: 'End',
    ENTER: 'Enter',
    ESC: 'Escape',
    HOME: 'Home',
    LEFT: 'ArrowLeft',
    RIGHT: 'ArrowRight',
    SPACE: 'Space',
    TAB: 'Tab',
    UP: 'ArrowUp'
};

import { keyCodes } from '../../helpers/keyCodes.js';
const { LEFT, RIGHT } = keyCodes;
```
The destructuring statement being optional for the consumer.  It gives the consumer the choice on how they want to reference the key codes.

omsmith: > The chance of collision is low, but certainly not zero

You can rename imports as well, no?

```js
import { LEFT as KeyLeft } from '../../helpers/keyCodes.js';
```

The separate exports is one of the big-benefits of ESM, and is better for dead code removal.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,204 @@
...
+		const node = getFirstFocusableDescendant(this);
+		if (node) node.focus();
+	}
+
+	render() {
+		return html`${this._keyboardActive && !this.disabled ? this._renderKeyboardDragging() : this._renderDragger()}`;
+	}
+
+	updated(changedProperties) {
+		if (changedProperties.has('_keyboardActive')) {
```
dbatiste, line 90: I think it is best to call `super.updated(changedProperties);`.

Nit: this could just be: `if (changedProperties.has('_keyboardActive')) this.focus();`

helpers/keyCodes.js
dbatiste: Yeah that's a fair point.  Perhaps we should just qualify the names like that by default.  That is, `KeyLeft` vs `LEFT`.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,204 @@
...
+			this._dispatchAction(dragActions.save);
+		}
+	}
+
+	_handleActiveKeyboard(e) {
+		if (!this._keyboardActive) {
+			return;
+		}
+		let action = null;
+		switch (e.key) {
```
dbatiste, line 123: I think `keyCode` has been deprecated for a while but we were unable to avoid it, possibly due to IE11.  Mozilla actually says to use `code`, but also mentions that it may not be supported by all browsers.  Is `key`?  Note that we have a whole bunch of other places that use numeric key codes, so in order for them to use the new import, we'll need to make some additional adjustments, assuming all works.

dbatiste: Apparently there are differences in key ~ `code` mappings.  Fairly long list of them on Mozilla.

components/list/test/list-item-drag-handle.test.js
```diff
@@ -0,0 +1,94 @@
...
+describe('ListItemDragHandle', () => {
+
+	describe('Events to activate keyboard mode', () => {
+		let element;
+		beforeEach(async() => {
+			element = await fixture(html`<d2l-list-item-drag-handle></d2l-list-item-drag-handle>`);
+			element.focus();
+		});
+
+		it(`Expect ${dragActions.active} event results when clicked`, async() => {
```
dbatiste, line 14: Nit - for BDD tests, names are usually written to read naturally starting with "it..." to describe the behavior.  The underlying expectations can change over time.

ex. "it _dispatches drag action event for [action] when clicked_"

While this is a pretty low level test, the idea behind it is that it leads us to take a more behavior driven mindset.

helpers/keyCodes.js
zinabat: If we're just using the straight up name for the key, what's the difference between just using `e.code === 'KeyLeft'`? The original point of the module was to hide the magic keyCode constants but exporting them as their actual code seems silly to me.

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -0,0 +1,45 @@
...
+	const visualDiff = new VisualDiff('list-item-drag-handle', __dirname);
+
+	let browser, page;
+
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await visualDiff.createPage(browser);
+		await page.goto(`${visualDiff.getBaseUrl()}/components/list/test/list-item-drag-handle.visual-diff.html`, {waitUntil: ['networkidle0', 'load']});
+		await page.bringToFront();
+		await visualDiff.resetFocus(page);
```
dbatiste, line 15: Normally the focus reset is done between tests if focus is applied during tests:
```
beforeEach(async() => {
	await visualDiff.resetFocus(page);
});
```
Currently, the `resetFocus` would be called before all these tests run, and only once.  It's not really doing anything for the tests.  For the current tests, focus is applied in the 2nd test, and would remain focused for the 3rd test, which makes its own focus call moot.

helpers/keyCodes.js
dbatiste: That's a good question - previously there was value when we were using the numeric codes with `keyCode`.  If we are going to use `key` then I don't see a lot of value here, but I think we need to figure which API we need to use (previously it was `keyCode`) - `keyCode` (deprecated), `code` (recommended by Mozilla but possibly has issues with mappings), or `key` (which was previously not used for some reason).  If necessary, we can then update other components that currently use `keyCode`.

My intent in the previous comment wasn't to just export the same thing - I actually didn't really pay attention to the actual values.  Rather, was just wanting to avoid collision and choose default names that clearly represent "key"s, avoiding rename in the import statement and addressing the somewhat minor concern of collision.

components/list/list-item-drag-handle.js
m6jones: So I didn't use code because it apparently maps to buttons not the character. They have a warning for code:

> Warning: This ignores the user's keyboard layout, so that if the user presses the key at the "Y" position in a QWERTY keyboard layout (near the middle of the row above the home row), this will always return "KeyY", even if the user has a QWERTZ keyboard (which would mean the user expects a "Z" and all the other properties would indicate a "Z") or a Dvorak keyboard layout (where the user would expect an "F").
>https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

This made me want to use key instead of code.

Why did I go against the grain with using keycode? The main reason is that the tests failed because old edge doesn't support building a `KeyboardEvent` with keycode anymore. So all my tests failed in old edge.

I missed this when I first scanned the page.
>     case "Left": // IE/Edge specific value

However, I don't have this and `case "arrowLeft";` seemed to have worked. I am going to go back a double check that this actually works in the browser. Worried I got excited to finally fixed the tests and didn't check.

If this doesn't work, I guess using `code` would be best? I really don't like that warning. 

The other option is to code up a helper method that does this for us and we never have to worry about it again. My issue with that right now is the only way I can see that working well is having a messy signature for the method.

components/list/test/list-item-drag-handle.test.js
m6jones: That sounds awesome. Will do that

components/list/list-item-drag-handle.js
m6jones: Thanks I did miss `super.updated(changedProperties);`

helpers/keyCodes.js
m6jones: I was more partial towards the object as well but basically what Owen mentioned is main reason I decided against it.

So I still prefer to have the constants since then I don't have to look up if it is `Esc` or `Escape` or maybe it is both ahh! 

So I don't like the preface of key on every key. We are importing from keyCodes we know they are keys. However, I hate all caps constants so having `keyLeft` is better for me.

components/list/demo/list-item-sample.js
m6jones: Yup, will change that. This could be something a linter code do?

components/list/list-item-drag-handle.js
m6jones: So I think this is growing bigger then this PR and should be thought more about before any patterns are made in core about this. I found that Margaree has fixed my edge problems with tests. So I am going to copy what Calendar did.

helpers/keyCodes.js
m6jones: So this discussion about keys is in two places:
>So I think this is growing bigger then this PR and should be thought more about before any patterns are made in core about this. I found that Margaree has fixed my edge problems with tests. So I am going to copy what Calendar did.

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	constructor() {
+		super();
+		this._keyboardActive = false;
+		this.disabled = false;
+	}
+
+	render() {
+		return html`${this._keyboardActive && !this.disabled ? this._renderKeyboardDragging() : this._renderDragger()}`;
```
dbatiste, line 97: We can simplify the template result and avoid a set of HTML comments from being rendered if we use:
```
if (this._keyboardActive && !this.disabled) return this._renderKeyboardDragging();
else return this._renderDragger();
```

components/list/list-item-drag-handle.js
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('_keyboardActive')) this.focus();
```
dbatiste, line 102: This seems to force-focus to the drag handle when the page is first loaded.  I don't think we expect that to happen... ?

components/list/test/list-item-drag-handle.test.js
```diff
@@ -0,0 +1,112 @@
...
+	DOWN: { key: 'down arrow', code: 40 },
+	SPACE: { key: 'space', code: 32 },
+	ENTER: { key: 'enter', code: 13 },
+	ESC: { key: 'escape', code: 27 },
+	TAB: { key: 'tab', code: 9 },
+	LEFT: { key: 'left arrow', code: 37 },
+	RIGHT: { key: 'right arrow', code: 39 }
+});
+
+describe('ListItemDragHandle', () => {
```
dbatiste, line 17: We should include a constructor test for the drag handle - take a peek at one of the other components but I think it would basically be...
```
it('should construct d2l-list-item-drag-handle', () => {
	runConstructor('d2l-list-item-drag-handle');
});
```
The reason for this is - there are certain things that are not allowed (such as sprouting attributes on the element) in the constructor.  We've seen this come up a couple times and it's easily missed since it's not often that we construct elements using `document.createElement` in Lit elements, fixtures, or our demos.

components/list/test/list-item-drag-handle.visual-diff.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="../../../test/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '../list-item-drag-handle.js';
+		</script>
+		<title>d2l-list-item-drag-handel</title>
```
dbatiste, line 9: Nit: typo on "handel" => "handle".

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -0,0 +1,48 @@
...
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await visualDiff.createPage(browser);
+		await page.goto(`${visualDiff.getBaseUrl()}/components/list/test/list-item-drag-handle.visual-diff.html`, {waitUntil: ['networkidle0', 'load']});
+		await page.bringToFront();
+	});
+
+	after(async() => await browser.close());
+
+	beforeEach(async() => {
```
dbatiste, line 19: Nit: in the interest of brevity and matching the line directly above it, this could be:
```
beforeEach(async() => await visualDiff.resetFocus(page));
```
No big deal though.

components/list/list-item-drag-handle.js
m6jones: No, we want it when it focus but not the first time. Nice catch.

components/list/test/list-item-drag-handle.test.js
m6jones: Cool, didn't know about this at all. Will add it.

PR #547, opened 2020-05-14
Focus the checkbox when label is clicked

zinabat:
Fixes a bug that prevented the checkbox from being focused when the action is clicked.

Caught in this story: [US115380](https://rally1.rallydev.com/#/detail/userstory/382436537132?fdp=true)


AKobets:
verified on MS Edge and chrome

PR #549, opened 2020-05-19
Adding generalized selectable mixin

zinabat:
## **Context**
[US116587](https://rally1.rallydev.com/#/detail/userstory/391283528060?fdp=true)

Generalizes the `list-item-checkbox-mixin` to `selectable-mixin` for easier use by non `list-item` components.

selectable

* Readme with documentation and screenshot added
* Event name decoupled from `list-item`
* Demo page for the mixin added

Screen Shot 2020-05-19 at 1 05 49 PM

Screen Shot 2020-05-19 at 1 05 38 PM

## **Quality**
- Unit tests from `list-item-checkbox-mixin` migrated and adjusted
- Manually tested with:
  - [x] Legacy Edge
  - [ ] CrEdge
  - [x] Firefox
  - [x] Chrome
  - [x] Safari

**Issue with aXe automated test**
This test fails:
```js
const element = await fixture(`<${tag} key="1234" selectable></${tag}>`);
await expect(element).to.be.accessible()
```
aXe will cite this rule: `Form elements must have labels`
However, the checkbox is correctly labelled. I've removed this test as a result. If someone knows why this happens, I'd love to know so I can fix it.


mixins/selectable/test/selectable-mixin.test.js
```diff
@@ -0,0 +1,66 @@
...
+		it('dispatches custom event when checkbox is checked', async() => {
+			// simulate a checkbox click
+			setTimeout(() => {
+				const checkbox = element.shadowRoot.querySelector('.d2l-input-checkbox');
+				checkbox.checked = true;
+				checkbox.dispatchEvent(new Event('change'));
+			});
+			const { detail } = await oneEvent(element, element._selectedEventName);
+			expect(detail.key).to.equal('1234');
+			expect(detail.selected).to.equal(true);
```
AKobets, line 36: also need test to show state could change from selected to unselected

AKobets: Tests are missing assertion on initial `element.selected` state and after the "click"

mixins/selectable/test/selectable-mixin.test.js
```diff
@@ -0,0 +1,66 @@
...
+				${this._renderSelectAction()}
+			`;
+		}
+	}
+);
+
+describe('SelectableMixin', () => {
+	it('Sets checked status to undefined when no key is given', async() => {
+		const element = await fixture(`<${tag} selected="true"></${tag}>`);
+		expect(element.selected).to.be.undefined;
```
AKobets, line 19: what if key is undefined, but selectable property is specified ?

mixins/selectable/test/selectable-mixin.test.js
```diff
@@ -0,0 +1,66 @@
+import { defineCE, expect, fixture, oneEvent } from '@open-wc/testing';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { SelectableMixin } from '../selectable-mixin.js';
+
```
AKobets, line 4: there could be 8 combinations of disabled, selectable, selected  states. 
All probably should be covered by testing expected corresponding dom state, and state change as result of change and click events

mixins/selectable/selectable-mixin.js
```diff
@@ -0,0 +1,89 @@
...
+			` : nothing;
+	}
+
+	_renderSelectAction(inner) {
+		return this.selectable ? html`
+			<label @click="${this._handleSelectActionClick}"
+				class="d2l-select-action"
+				?disabled="${this.disabled}"
+				for="${this._checkboxId}">
+				${inner}
```
AKobets, line 85: what is purpose of the inner ?

zinabat: It is an optional argument that is used to place text or html or other components within the `label`, but it should probably be using `nothing` instead to prevent an empty text node

m6jones: nothing is for slots to default. so null is good

mixins/selectable/test/selectable-mixin.test.js
zinabat: I've adjusted the tests so that all of these combinations are taken into account - definitely important!

mixins/selectable/selectable-mixin.js
dlockhart: Be a little careful here. `<label>` elements should only be used to wrap or link to inputs. And even then, we ran into a lot of issues with date/time where we had a label wrapping a text input and AT ends up reading a lot of things you don't want it to. Is it expected that they'll pass `_renderCheckbox` in as `inner`? Because then things get double-labelled, which is weird.

Also... if they don't pass anything for `inner`, then the label that's linked to the checkbox won't have any text associated with it, so the checkbox will be unlabelled, which we can't have.

zinabat: I've been contemplating adding an `aria-labelledby` to the `label` for cases where `inner` isn't passed, which I'm realizing now is a necessary thing.

If users decide to pass `_renderCheckbox` as `inner`, is that bad because the `label` already has the `for` attribute associated with the checkbox?

zinabat:
> This is interesting! I definitely see a need for multiple things to be "selectable", and so having that logic shared makes sense I think. However, I'm not sure we're at the point yet where we have all those different use cases really nailed down?[...]

These are fair points. The use case that was pointed out to me were things like cards, which currently aren't selectable but could be. You're probably right and I am jumping the gun here because others noticed that the `list-item-checkbox-mixin` does not need to be specific to list items and there _could_ be use cases for other 'selectable' components.

I'm fine with this being shelved until those use cases are better understood.

dlockhart:
I love the instinct towards reusability and factoring out useful bits for sure! So don't stop thinking that way. :) We just need to find that balance between factoring out before we know exactly how all the different usages will work vs. dealing with some duplicated code potentially.

Factored out code isn't perfect either as you're seeing... it can end up being complicated to use & understand when you come back to it months later (or someone else does), and hard to change/test later too once it's used in a bunch of ways you didn't initially imagine.

I'd suggest waiting until you have 3 or 4 solid "selectable" things that you can immediately use this thing in -- then I think if it works well for those and makes them each "better" (less code but still understandable), then go for it.

PR #551, opened 2020-05-19
separating axe tests out

dlockhart:
What do people think of this approach? Once CI finishes we'll see how much time it saves by not running all the aXe stuff in every browser.


dlockhart:
This seems to run no slower than before, and typically a little bit faster. There's a bit of added overhead from kicking off a new Karma session, which seems to currently mostly balance out from not needing to run aXe tests in a bunch of different browsers.

I was able to remove the timeouts though, so overall I think this is a win. Plus as we add more aXe tests, that overhead will become less and less relevant.

PR #552, opened 2020-05-20
input-date/calendar: fix incorrect focus date on initial open

margaree:
Problem: Initially when `selectedValue` goes from `undefined` to `''` when used in `input-date` the `_focusDate` becomes the first of the month instead of today. This change checks if today is in the current month and then that becomes focus date if there isn't a `selectedValue`.


components/calendar/calendar.js
```diff
@@ -439,6 +439,7 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 				this._focusDateAddFocus();
 			} else if (prop === 'selectedValue') {
 				if (this.selectedValue) this._focusDate = getDateFromISODate(this.selectedValue);
+				else if (this._today.getMonth() === this._shownMonth) this._focusDate = new Date(this._today);
```
dlockhart, line 442: Don't we also want to check the year? i.e. won't this select May 20 vs. May 1st if the shown year is 1999?

PR #553, opened 2020-05-20
d2l-list-item-placement-marker

AKobets:
Context: US115500

Status: ready for final review


components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,29 @@
...
+		return {
+			color: { type: String },
+			displayed: {type: Boolean}
+		};
+	}
+
+	constructor() {
+		super();
+		this.color = "blue";
+		this.strokeWidth = 4;
```
AKobets, line 14: all dimensions are calculated based on the stroke width

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,29 @@
...
+		this.radius = this.strokeWidth * 2;
+		this.height = this.radius * 2 + this.strokeWidth * 2;
+	}
+
+	render() {
+	  return this.displayed ? html`
+		<svg height="${this.height}"  width="100%">
+		  	<circle cx="${this.strokeWidth + this.radius}" cy="${this.strokeWidth + this.radius}" r="${this.radius}" stroke="${this.color}" stroke-width="${this.strokeWidth}" fill="white" />
+	  		<line x1="${this.radius*2 + this.strokeWidth}" y1="${this.height/2}" x2="98%" y2="${this.height/2}" stroke="${this.color}" stroke-width="${this.strokeWidth}" stroke-linecap="round"/>
```
AKobets, line 23: x2="98%" - not sure how to improve this, If I  set x2 = 100%, the linecap is cut off"

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,29 @@
...
+	static get properties() {
+		return {
+			color: { type: String },
+			displayed: {type: Boolean}
+		};
+	}
+
+	constructor() {
+		super();
+		this.color = "blue";
```
AKobets, line 13: what default color should be ?

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,29 @@
...
+		this.strokeWidth = 4;
+		this.radius = this.strokeWidth * 2;
+		this.height = this.radius * 2 + this.strokeWidth * 2;
+	}
+
+	render() {
+	  return this.displayed ? html`
+		<svg height="${this.height}"  width="100%">
+		  	<circle cx="${this.strokeWidth + this.radius}" cy="${this.strokeWidth + this.radius}" r="${this.radius}" stroke="${this.color}" stroke-width="${this.strokeWidth}" fill="white" />
```
zinabat, line 22: In general, we will want as much of these style-specific things to be in `static get styles()` as possible. They're easier to view and manipulate this way. See [Maya's prototype for an example](https://github.com/BrightspaceHypermediaComponents/activities/blob/majones/PROTOTYPE/dragList/components/d2l-activity-collection-editor/d2l-activity-collection-list-item.js#L358)

It's also possible to have expressions or calculated values in these styles: [Lit-element styles](https://lit-element.polymer-project.org/guide/styles#expressions)

components/list/list-item-placement-marker.js
zinabat: http://design.d2l/style/colour/
Celestine
You can do this the way Maya mentioned with `var(--d2l-color-celestine)` in css

iFun:
I think test failed on some lint errors. do you have the auto lint fix in place?

components/list/list-item-placement-marker.js
zinabat: Assigning via property makes sense to me, with the default being `3` as I think that's what the designs use

components/list/list-item-placement-marker.js
zinabat: For this, I would recommend using a third shape as in the Prototype [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/majones/PROTOTYPE/dragList/components/d2l-activity-collection-editor/d2l-activity-collection-list-item.js#L363)
Just place the line-cap piece overlapping the existing line so we get that nice rounded edge

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,49 @@
...
+				stroke: var(--d2l-color-celestine);
+				stroke-linecap: round;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.strokeWidth = 4;
+		this.radius = this.strokeWidth * 2;
```
zinabat, line 35: The designs have the radius set to 4 when the stroke width is 3 - I'm unsure if we should be calculating the radius this way or what formula we should use. @m6jones ?
**Current**
Screen Shot 2020-05-21 at 11 03 20 AM
**Design**
Screen Shot 2020-05-21 at 11 04 04 AM

m6jones: Yeah, we should be following the designs. Which Abbie just posted!

components/list/list-item-placement-marker.js
m6jones: Yes, what zina said about the third shape is the only way to do it.
```
<div class="${htmlClass}">
<svg viewBox="0 0 12 12" class="d2l-list-drag-indicator-circle">
	<circle cx="6" cy="6" r="4" fill="none"/>
</svg>
<svg class="d2l-list-drag-indicator">
	<line x1="0" y1="50%" x2="100%" y2="50%" />
</svg>
<svg viewBox="0 0 5 12" class="d2l-list-drag-indicator-linecap">
          <line x1="-5" y1="50%" x2="0" y2="50%" />
</svg>
```
I also found it helpful to be more dynamic to split them into separate svgs. This allows the user of this component to define the width without any stretching of the circle or the round end cap.

components/list/list-item-placement-marker.js
AKobets: image

components/list/list-item-placement-marker.js
m6jones: The strokeWidth is private so it should have a `_` to denote it.
https://blog.d2l.dev/2020/05/11/LitElement-Best-Practices-Gotchas/

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,49 @@
...
+		this.height = this.radius * 2 + this.strokeWidth * 2;
+	}
+
+	render() {
```
m6jones, line 39: I would define height, radius and strokewidth within the render method over the constructor. I don't see reason to have their scope extended to the class.

zinabat: The advantage of the constructor would be that users can alter these properties if they need to, but I'd argue in that case we should have a single property such as `size` that does all of that.

m6jones: Unless we have a use case for a user to modify the size I dont think we should support it. Or if we do have it modified by css height not a var

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,69 @@
...
+			 }
+
+		`;
+	}
+
+	render() {
+		return html`
+			<div class="d2l-list-drag-marker">
+				<svg class="d2l-list-drag-marker-circle">
+					<circle cx="50%" cy="50%" r="4px"/>
```
m6jones, line 55: `</circle>` not `/>`

components/list/list-item-placement-marker.js
dbatiste: I notice there are some off effects with the width - specifically the svg is overlapping the border in my mock-up, but without the border, it is overflowing. (zoomed screenshot)

image

Without the border, the edge of the circle appears to get cut off. (zoomed screenshot)

image

I also notice some odd things with the width of the line - like it doesn't always go 100% of it's container the way I expect it to.  (not zoomed screenshot, svg on top, div on bottom)

image


Was a `div` sol'n considered?  It seems simpler to me.

```

class ListItemPlacementMarker extends RtlMixin(LitElement) {

	static get styles() {
		return css`
			:host {
				display: block;
				position: relative;
				line-height: 0;
			}

			:host([hidden]) {
				display: none;
			}

			.d2l-list-drop-marker-circle {
				border: 3px solid var(--d2l-color-celestine);
				border-radius: 6px;
				box-sizing: border-box;
				display: inline-block;
				height: 11px;
				width: 11px;
			}

			.d2l-list-drop-marker-line {
				border-bottom: 2px solid var(--d2l-color-celestine);
				border-top: 1px solid var(--d2l-color-celestine);
				border-radius: 1px;
				position: absolute;
				top: 4px;
				left: 10px;
				width: calc(100% - 10px);
			}
		`;
	}

	render() {
		return html`
			<div class="d2l-list-drop-marker-circle"></div>
			<div class="d2l-list-drop-marker-line"></div>
		`;
	}

}

customElements.define('d2l-list-item-placement-marker-2', ListItemPlacementMarker);

```

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,69 @@
...
+			:host {
+				display: block;
+				position: relative;
+			}
+
+			:host([hidden]) {
+				display: none;
+			}
+
+			.d2l-list-drag-marker-line {
```
m6jones, line 17: This should have `width: 100%`

components/list/list-item-placement-marker.js
AKobets: @dbatiste thank you for catching the cut-off issue. I've tried to implement your exampla and found another problem - line some times is split when zooming page with ctrl+/-
I've adjusted margins in "my approach" and it looks better.
image

AKobets:
> I think test failed on some lint errors. do you have the auto lint fix in place?
> 
> /home/travis/build/BrightspaceUI/core/components/list/list-item-placement-marker.js
> 13:16 error Strings must use singlequote quotes
> 20:2 error Mixed spaces and tabs no-mixed-spaces-and-tabs
> 23:29 error Operator '*' must be spaced space-infix-ops
> 23:70 error Operator '/' must be spaced space-infix-ops
> 23:101 error Operator '/' must be spaced space-infix-ops
> 25:8 error Strings must use singlequote quotes
> 27:1 error Expected indentation of 0 tabs but found 2 spaces indent
> 29:1 error Expected indentation of 0 tabs but found 2 spaces indent
> ✖ 8 problems (8 errors, 0 warnings)
> 7 errors and 0 warnings potentially fixable with the `--fix` option.

fixed.

AKobets:
visual diff tests failed. Not related to my changes. Should I do anything about it?
```
d2l-input-date-time
       "before all" hook for "basic":
     Error: Failed to launch the browser process!
Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!
TROUBLESHOOTING: https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md
      at onClose (node_modules/puppeteer/lib/Launcher.js:750:14)
      at ChildProcess.<anonymous> (node_modules/puppeteer/lib/Launcher.js:740:61)
      at ChildProcess.EventEmitter.emit (domain.js:485:12)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:276:12)
  2) d2l-input-date-time
       "after all" hook for "basic-focus":
     TypeError: Cannot read property 'close' of undefined
      at Context.<anonymous> (components/inputs/test/input-date-time.visual-diff.js:18:33)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
  3) d2l-tabs
       "before all" hook in "d2l-tabs":
     Error: Protocol error (Target.setDiscoverTargets): Target closed.
      at /home/travis/build/BrightspaceUI/core/node_modules/puppeteer/lib/Connection.js:74:56
      at new Promise (<anonymous>)
      at Connection.send (node_modules/puppeteer/lib/Connection.js:73:12)
      at Function.create (node_modules/puppeteer/lib/Browser.js:34:22)
      at ChromeLauncher.launch (node_modules/puppeteer/lib/Launcher.js:246:37)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at async Context.<anonymous> (components/tabs/test/tabs.visual-diff.js:11:13)
  4) d2l-tabs
       "after all" hook in "d2l-tabs":
     TypeError: Cannot read property 'close' of undefined
      at Context.<anonymous> (components/tabs/test/tabs.visual-diff.js:19:33)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

dbatiste:
Hey @AKobets , you'll just need to rerun the tests unfortunately, and yep, it is unrelated to your changes.  

We've seen this error a few time before and it's kinda random.  From my brief research, I think it is due to a threading problem in the deep recesses of the OS.  

I have some idea to share a launched browser but I have not completed the change, and remains to be seen if it will actually resolve the issue.  We can also raise this in the Puppeteer repo, although I think it has already been raised, and they may only be able to work around the issue.

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,91 @@
...
+				margin-right: -1px;
+			}
+
+			.d2l-list-drag-marker-linecap {
+				height: 12px;
+				width: 4px;
+				stroke: none;
+				fill: var(--d2l-color-celestine);
+				margin-left: -1px;
+				margin-right: 0px;
```
dbatiste, line 32: Nit: `0px` should just be `0`.  It used to be a micro-optimization for performance, but nowadays I'd say it's just a consistency thing.

components/list/test/list-item-placement-marker.axe.js
```diff
@@ -0,0 +1,13 @@
+import '../list-item-placement-marker.js';
+import { expect, fixture, html } from '@open-wc/testing';
+
+describe('d2l-list-item-placement-marker', () => {
+	it('visible', async() => {
+		const element = await fixture(html`<d2l-list-item-placement-marker></d2l-list-item-placement-marker>`);
+		await expect(element).to.be.accessible();
+	});
+	it('visible-rtl', async() => {
```
dbatiste, line 9: I don't think we really need to check a11y for the RTL version.

components/list/test/list-item-placement-marker.visual-diff.js
```diff
@@ -0,0 +1,39 @@
...
+	describe('placement-marker', () => {
+		it('ltr', async function() {
+			const rect = await visualDiff.getRect(page, '#placement-marker');
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+		it('rtl', async function() {
+			const rect = await visualDiff.getRect(page, '#placement-marker-rtl');
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+		it('ltr-with-border', async function() {
```
dbatiste, line 30: Would it be sufficient to just capture images with the border (i.e. only 2 screenshots)?  While the screenshots capture so much that a unit test can't, they definitely have a cost to them.  If the previous two tests capture something that these don't then it's perfectly fine, but would opt being conservative with the test time if possible.

components/list/test/list-item-placement-marker.visual-diff.js
```diff
@@ -0,0 +1,39 @@
...
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await visualDiff.createPage(browser);
+		await page.goto(`${visualDiff.getBaseUrl()}/components/list/test/list-item-placement-marker.visual-diff.html`, {waitUntil: ['networkidle0', 'load']});
+		await page.bringToFront();
+	});
+
+	after(async() => await browser.close());
+
+	beforeEach(async() => await visualDiff.resetFocus(page));
```
dbatiste, line 19: I think we can remove this from these tests.  In some tests where we capture screenshots of components in their focused states, we will use this helper to reset the focus between tests if the page is not being reloaded.  In this case, focus is not being applied to the component, so this isn't needed.

components/list/test/list-item-placement-marker.axe.js
m6jones: I agree, I don't think this offers any differences for accessibility from testing ltr

components/list/test/list-item-placement-marker.visual-diff.js
dbatiste: Sure sounds fine.  :)

components/list/list-item-placement-marker.js
```diff
@@ -0,0 +1,90 @@
...
+			.d2l-list-drag-marker {
+				flex-wrap: nowrap;
+				display: flex;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.role = 'img';
```
dbatiste, line 69: This is interesting.  What's the rationale for marking this up to be an image, semantically?  What does the screen reader communicate?  I think this is the first time I've seen `img` being assigned for the role.

AKobets: @m6jones  could you please comment on dave's question ?

m6jones: So to me this is an image component. So I wanted to express it as an image. Secondary, I want people to be able to define this as decorative or have alt text if their use case needs it.

dbatiste: My thinking is that this would always be decorative.  In what case would we want a consumer of this component be able to define it as decorative or not?  How would this get surfaced to AT and would it be meaningful as an image?

dbatiste: i.e. what is the best semantic to convey this as a drop target, and is it even useful for AT?

dbatiste: Turns out there is a bit of Aria for this, although it is deprecated with no replacement. The `aria-grabbed` and `aria-dropeffect` are meant to indicate drag actions and drop targets, which can be any element in the base markup.  https://w3c.github.io/aria/#attrs_dragdrop

m6jones: Those would be out of scope for this component. I believe this is more of keeping this component clean to maybe I went a little far on making it perfect for it singular purpose. This component to me is just a image that is a circle with a line. :)

So I guess since this in the context of drag n drop which the name is very pointed at it will always be decorative so we don't really need the name. The accessibility for drag n drop will be handled not by this component at all.

dbatiste: Yeah that was my understanding of the a11y for the d&d as well.

Mozilla describes the use of [role="img"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img) differently here, but makes a point of suggesting it needs to be labelled, and if applied to an `svg` (that is not decorative), that it should be applied to the outer `svg`.

For this, I see the image as just an internal detail for the visual representation.  It could have been constructed with `div`s without affecting the meaning.  i.e. we're not surfacing this to AT with the expectation that it will get announced and described as an image.

I feel like I am making a big deal out of something that is small, but I don't agree with the semantics here.

AKobets: @dbatiste is there anything else that need to be done with this PR before owners approval ?

dbatiste: The `role="img"` should be removed, since it does not define this group of svgs in an accessible way.  If it is used (which I do not think it should be in this case), the element needs to be labelled.

PR #554, opened 2020-05-20
[20.20.6] - DE39211 - Delay tab panel selected event

svanherk:
Withe the migration of tabs to Lit, my-courses is missing the first `d2l-tab-panel-selected` event which can get dispatched before the panels are connected or event listeners are setup.  It looks like the old behavior had this delay: https://github.com/BrightspaceUI/tabs/blob/master/d2l-tab-panel-behavior.js#L75

### DIFF
--- a/components/tabs/tab-panel-mixin.js
+++ b/components/tabs/tab-panel-mixin.js
@@ -33,11 +33,23 @@ export const TabPanelMixin = superclass => class extends superclass {
 		this.role = 'tabpanel';
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldVal, prop) => {
+			if (prop === 'selected') {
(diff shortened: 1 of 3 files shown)


components/tabs/tab-panel-mixin.js
```diff
@@ -33,11 +33,23 @@ export const TabPanelMixin = superclass => class extends superclass {
...
+			}
+		});
+	}
+
 	async attributeChangedCallback(name, oldval, newval) {
 		super.attributeChangedCallback(name, oldval, newval);
-		if (name === 'selected') {
-			if (this.selected) this._dispatchSelected();
-		} else if (name === 'text') {
+		if (name === 'text') {
```
svanherk, line 52: In `master`, I'll move this to `updated` too, but keeping the cert fix more selective

PR #558, opened 2020-05-21
fixes #557: making key shortcuts to menu items case-insensitive

### LINKED ISSUE #557: Should menu shortcut keys be case-insensitive?
Watching @svanherk's GAAD demo today made me wonder if we shouldn't be treating the printable characters in menu items in a case sensitive way? It seemed really unintuitive to me to need to know whether a menu item is "Delete" vs. "delete" to know whether to hold down shift.

[The spec](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) is ambiguous:
> Any key that corresponds to a printable character (Optional): Move focus to the next menu item in the current menu whose label begins with that printable character.

Thoughts?

dlockhart:
If we want to make these checks case insensitive, here's a way to do it.

### DIFF
--- a/components/menu/menu.js
+++ b/components/menu/menu.js
@@ -256,7 +256,7 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 		e.stopPropagation();
 
 		const startsWith = function(item, value) {
-			if (item.text && item.text.length > 0 && item.text.substr(0, 1) === value) {
+			if (item.text && item.text.length > 0 && item.text.toLowerCase().substr(0, 1) === value) {
 				return true;
 			}
 			return false;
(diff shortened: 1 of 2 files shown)


components/menu/test/menu.test.js
```diff
@@ -117,6 +117,14 @@ describe('d2l-menu', () => {
 			expect(document.activeElement).to.equal(elem.querySelector('#c1'));
 		});
 
+		it('sets focus to next item that starts with uppercase character pressed', () => {
+			const eventObj = document.createEvent('Events');
+			eventObj.initEvent('keypress', true, true);
+			eventObj.charCode = 67;
```
dlockhart, line 123: Hopefully we can kill these legacy events and just use `new KeyboardEvent()` and `new Event()` soon now that IE11 is officially dead.

components/menu/menu.js
```diff
@@ -256,7 +256,7 @@ class Menu extends HierarchicalViewMixin(LitElement) {
 		e.stopPropagation();
 
 		const startsWith = function(item, value) {
-			if (item.text && item.text.length > 0 && item.text.substr(0, 1) === value) {
+			if (item.text && item.text.length > 0 && item.text.toLowerCase().substr(0, 1) === value) {
```
dbatiste, line 259: Nit: could lower case just one char instead of all the text.
```
item.text.substr(0, 1).toLowerCase()
```
That being said, this code only runs each time they hit a key... probably can't even measure the difference in such case.

PR #559, opened 2020-05-22
applying link styles using CSS classes instead of custom attributes

dlockhart:
I'm trying to get `lit-analyzer` strict mode working, so this is the first in a series of PRs to address some of the issues it's finding.

The issue here is that we're putting custom attributes like `main` and `small` on native `<a>` elements. Lit-analyzer checks to make sure that attributes you're putting on things are correct, so isn't a fan of this.

The verdict on whether this is valid HTML or not is murky. But my opinion is this: I could disable the lint rule, except that it also found a bunch of legit places where we were incorrectly putting attributes on things. And this is a pretty easy fix for us to just switch to using CSS classes...


components/link/link.js
```diff
@@ -1,7 +1,28 @@
 import '../colors/colors.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { classMap} from 'lit-html/directives/class-map.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
-import { linkStyles } from './link-styles.js';
```
dlockhart, line 4: Note: I collapsed the styles into here and `export`ed them while I was at it. This aligns with our best practice for not having unnecessary imports. I did a code search for `linkStyles` that didn't turn up anything, so think this is safe.

PR #561, opened 2020-05-22
Fix legacy edge bug

rgoessl:
Object spread not supported in legacy edge, stopped FRA from loading

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -248,11 +248,10 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			width: this._width ? `${this._width + 18}px` : '',
 		};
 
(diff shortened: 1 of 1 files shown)


dlockhart:
Shouldn't this have gotten transpiled by Babel/etc. in your FRA? Make sure that your configuration for browsers is set up properly... if it's set to "last 2 versions" then legacy-Edge no longer falls into that category.

rgoessl:
> 
> 
> Shouldn't this have gotten transpiled by Babel/etc. in your FRA? Make sure that your configuration for browsers is set up properly... if it's set to "last 2 versions" then legacy-Edge no longer falls into that category.

Updated my babel config, still doesn't seem to work:

```
{
	"presets": [
		[ "@babel/env", { 
			"targets": {
				"edge": "17"
			} 
		} ],
		"@babel/react"
	]
}
```

PR #562, opened 2020-05-22
Lit analyzer datetime fixes

dlockhart:
A couple things the analyzer caught with date/time.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -51,6 +51,16 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
+			noPaddingFooter: {
+				type: Boolean,
+				reflect: true,
+				attribute: 'no-padding-footer'
+			},
+			noPaddingHeader: {
+				type: Boolean,
+				reflect: true,
+				attribute: 'no-padding-header'
+			},
```
dlockhart, line 63: When time-input applied this attribute, the analyzer correctly complained that the property didn't actually exist... which is true -- there was only CSS for it.

components/inputs/input-date.js
```diff
@@ -219,7 +219,6 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 					title="${this.localize('openInstructions', {format: shortDateFormat})}"
 					.value="${this._formattedValue}">
 					<d2l-icon
-						?disabled="${this.disabled}"
```
dlockhart, line 222: Icons can't actually be disabled, so pretty sure this wasn't doing anything.

PR #566, opened 2020-05-22
DE39212 - Delay tab panel selected event

svanherk:
`master` version of https://github.com/BrightspaceUI/core/pull/554


svanherk:
> Looks good to me but I was wondering if it makes sense to have the ability to check what panel is selected? It feels a little weird to check an event that can be missed to get the initial state.

So we do reflect the `selected` attribute for consumers to query if a `tab-panel` is selected or not.  The problem is that the all-courses overlay isn't using this event just to keep track of the selected tab, it's relying on this event to kick off its whole initial data retrieval workflow, which is not really what the event is for.  I'll take a look at this as part of the next piece I pick up.

PR #571, opened 2020-05-25
removing unnecessary language attribute

dlockhart:
Couple minor fixes for lit-analyzer strict mode.

### DIFF
--- a/components/demo/code-view.js
+++ b/components/demo/code-view.js
@@ -27,7 +27,7 @@ class CodeView extends LitElement {
 	render() {
 		return html`
 			<div class="d2l-code-view-src"><slot @slotchange="${this._handleSlotChange}"></slot></div>
-			<div class="d2l-code-view-code">${this._codeTemplate}</div>
+			<div data-language="${this.language}" class="d2l-code-view-code">${this._codeTemplate}</div>
 		`;
 	}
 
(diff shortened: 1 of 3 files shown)


components/demo/code-view.js
```diff
@@ -27,7 +27,7 @@ class CodeView extends LitElement {
 	render() {
 		return html`
 			<div class="d2l-code-view-src"><slot @slotchange="${this._handleSlotChange}"></slot></div>
-			<div language="${this.language}" class="d2l-code-view-code">${this._codeTemplate}</div>
+			<div class="d2l-code-view-code">${this._codeTemplate}</div>
```
dlockhart, line 30: This wasn't doing anything.

components/demo/demo-snippet.js
```diff
@@ -24,8 +24,9 @@ class DemoSnippet extends LitElement {
 	}
 
 	render() {
+		const dirAttr = this._dir === 'rtl' ? 'rtl' : 'ltr';
 		return html`
-			<div class="d2l-demo-snippet-demo" dir="${this._dir}">
+			<div class="d2l-demo-snippet-demo" dir="${dirAttr}">
```
dlockhart, line 29: It was complaining that `dir` can't be any arbitrary string -- so forcing it to be one of these values fixes things.

components/demo/code-view.js
dbatiste: Currently I think this is used by:
```
:host .d2l-code-view-code::before {
    content: attr(language);
}
```

The language shows here:
image

Maybe we can just change it to a data attribute?

dlockhart: Ah cool, I was having trouble figuring out how it got applied. Will fix!

dlockhart: K fixed!

PR #572, opened 2020-05-25
using data-* attributes for custom attributes

dlockhart:


components/tabs/tabs.js
```diff
@@ -332,10 +332,9 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 
 		return html`
 			<div class="${classMap(tabsLayoutClasses)}">
-				<div ?allow-scroll-next="${this._allowScrollNext}"
-					?allow-scroll-previous="${this._allowScrollPrevious}"
+				<div ?data-allow-scroll-next="${this._allowScrollNext}"
+					?data-allow-scroll-previous="${this._allowScrollPrevious}"
 					class="d2l-tabs-container"
-					?scroll-collapsed="${this._scrollCollapsed}"
```
dlockhart, line 338: This doesn't appear to have been referenced by anything.

components/tabs/tabs.js
```diff
@@ -332,10 +332,9 @@ class Tabs extends LocalizeStaticMixin(ArrowKeysMixin(RtlMixin(LitElement))) {
 
 		return html`
 			<div class="${classMap(tabsLayoutClasses)}">
-				<div ?allow-scroll-next="${this._allowScrollNext}"
-					?allow-scroll-previous="${this._allowScrollPrevious}"
+				<div ?data-allow-scroll-next="${this._allowScrollNext}"
+					?data-allow-scroll-previous="${this._allowScrollPrevious}"
```
dlockhart, line 336: Could also go with CSS classes for these -- let me know if you'd prefer that.

components/tabs/tabs.js
dbatiste: Good find - I suspect impl for this behavior pivoted and this never got cleaned up.

PR #576, opened 2020-05-25
Calendar: min-value and max-value

margaree:
- adds `minValue` and `maxValue` to calendar properties
- changes the inner date to be a `button`. This is so `disabled` on dates outside of the valid range semantically makes sense.
- a bug fix was made related to the heading

Functionality expected:
- users cannot navigate by keyboard to the invalid dates
- users cannot click on invalid dates
- when users click page up and the date that they would page up to is disabled, their focus will instead go to the earliest focusable date (which would be closest to the date they would have gone to). If there is not one, their focus will go to the previous month icon button.
- when users click page down and the date that they would page down to is disabled, their focus will instead go to the latest focusable date (which would be closest to the date they would have gone to). If there is not one, their focus will go to the next month icon button.

TO DO:
- [x] update readme


components/calendar/calendar.js
```diff
@@ -430,6 +465,10 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			getCalendarData(true);
 			this.requestUpdate();
 		});
+
+		if (this.minValue && this.maxValue && (getDateFromISODate(this.minValue).getTime() > getDateFromISODate(this.maxValue).getTime())) {
+			console.warn('d2l-calendar component expects min-value to be before max-value');
```
margaree, line 470: I added a warning for if max is before min - not sure if that's the best way to indicate this?

components/calendar/calendar.js
```diff
@@ -467,31 +506,35 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 					'd2l-calendar-date-initial': this._isInitialFocusDate,
 					'd2l-calendar-date-selected': selected,
 					'd2l-calendar-date-today': checkIfDatesEqual(day, this._today)
 				};
 				const year = day.getFullYear();
 				const month = day.getMonth();
 				const date = day.getDate();
-				const description = `${formatDate(day, {format: 'medium'})}. ${selected ? this.localize('selected') : this.localize('notSelected')}`;
-				// role="gridcell" used for NVDA selected behavior to work properly
+				const description = `${date}. ${selected ? this.localize('selected') : this.localize('notSelected')} ${formatDate(day, {format: 'monthYear'})}`;
```
margaree, line 520: updated the `aria-label` with feedback from Carin

components/calendar/calendar.js
```diff
@@ -467,31 +506,35 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
+				// role="gridcell" used for screen reader (e.g., JAWS and VoiceOver) behavior to work properly
```
margaree, line 521: with the change to `button`, `gridcell` seems necessary for JAWS

components/calendar/calendar.js
```diff
@@ -509,7 +552,7 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 		};
 		const labelId = `${this._tableInfoId}-heading`;
 		const labelledBy = this._dialog ? labelId : undefined;
-		const heading = `${calendarData.descriptor.calendar.months.long[this._shownMonth]} ${this._shownYear}`;
+		const heading = formatDate(new Date(this._shownYear, this._shownMonth, 1), {format: 'monthYear'});
```
margaree, line 555: I realized that I had just been formatting my own Month Year heading rather than using `formatDate` for this. A few visual diff screenshots will be updated to reflect this.

components/calendar/calendar.js
```diff
@@ -663,87 +707,122 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 				this._keyboardTriggeredMonthChange = true;
 
 				// handle when current month has more weeks than previous month and page up pressed from last week
-				if (this._focusDate.getMonth() === this._shownMonth) this._focusDate.setDate(this._focusDate.getDate() - 7);
+				if (possibleFocusDate.getMonth() === this._shownMonth) possibleFocusDate.setDate(possibleFocusDate.getDate() - 7);
 
 				this._updateShownMonthDecrease(true, true);
+				await this.updateComplete;
+				this._updateFocusDateDependentOnDisabled(possibleFocusDate);
+				if (this._focusDate) this._focusDateAddFocus();
```
margaree, line 720: for `PAGEUP`: if there is not a focusable focusDate, focus on the previous month button icon (I'm thinking I should call that out more explicitly instead of using `this.focus()` in case for some reason `focus()` functionality changes.)

components/calendar/calendar.js
```diff
@@ -663,87 +707,122 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
-			this._isInitialFocusDate = false;
-			this._updateFocusDateOnKeyDown(numDaysChange);
-		}
-
 		if (preventDefault) {
 			e.preventDefault();
 			e.stopPropagation();
 		}
+
+		if (numDaysChange) {
```
margaree, line 759: I moved the functionality of `_updateFocusDateOnKeyDown` into here since it was not used anywhere else and then tweaked a little to deal with disabled focus dates.

components/calendar/calendar.js
```diff
@@ -663,87 +707,122 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
-		const date = await this._getDateElement(this._focusDate);
-		if (!date) {
-			this._keyboardTriggeredMonthChange = true;
-			const upDown = Math.abs(numDays) !== 1; // use left/right animation if difference is 1 day
-			if (oldFocusDate < this._focusDate) {
-				this._updateShownMonthIncrease(true, upDown);
-			} else {
-				this._updateShownMonthDecrease(true, upDown);
-			}
+	_updateFocusDateDependentOnDisabled(possibleFocusDate, latestPossibleFocusDate) {
```
margaree, line 813: 1. If desired focus date is between min and max, it becomes focus date
2. If there are dates in the month that are not disabled, the focus date is either the earliest or latest depending on if `latestPossibleFocusDate` is true
3. If there are no enabled dates, focus date is undefined.

components/inputs/input-date.js
```diff
@@ -273,8 +273,7 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 			await this._handleChange();
 			this._calendar.focus();
 			this._setFormattedValue();
-
-			if (e.keyCode === 40) e.preventDefault();
+			e.preventDefault();
```
margaree, line 276: there was a bug introduced where if enter was pressed to open the calendar on the input, the first date (e.g., today) would just be selected (dropdown would never be seen open, the text box would just fill)

components/calendar/calendar.js
margaree: `PAGEDOWN` works similarly, but focus would go to the next month button if no focusable dates

margaree:
I'm doing a bit of a re-factor to fix some things for NVDA so ignore for now

margaree:
This is ready for review when anyone gets a chance

components/calendar/calendar.js
```diff
@@ -133,6 +140,8 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 
 	static get properties() {
 		return {
+			maxValue: { type: String, attribute: 'max-value' },
+			minValue: { type: String, attribute: 'min-value' },
```
dlockhart, line 144: We should start reflecting new attributes I think, now that we've decided it's the best practice in out Lit guide.

components/calendar/calendar.js
dlockhart: Hmm, or you could even throw a [RangeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError)? They'll likely experience unexpected things otherwise, no?

PR #577, opened 2020-05-25
Alert Update to Design Spec

Scy-D2L:
Per [US114956](https://rally1.rallydev.com/#/detail/userstory/378675675652?fdp=true): Alerts - Update to align with design

- Removed warning type/color for `d2l-alert`
- Removed all type/colors for `d2l-alert-toast`
- Updated size/spacing


components/alert/alert.js
```diff
@@ -53,9 +53,6 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			:host([type="error"]) .d2l-alert-highlight {
 				background-color: var(--d2l-color-feedback-error);
 			}
-			:host([type="warning"]) .d2l-alert-highlight {
```
AllanKerr, line 56: I don't think we can safely remove this. We can mark it as deprecated in the docs but there seem to be quite a few places currently using it.

components/alert/alert-toast.js
```diff
@@ -71,6 +70,15 @@ class AlertToast extends LitElement {
 			d2l-alert {
 				animation: none;
 			}
+
+			@media (max-width: 615px) {
```
AllanKerr, line 74: This doesn't need to be in a media query since the max width of a toast is `600px`

components/alert/alert.js
Scy-D2L: Right now this would remove the colored band entirely, so I suspect something has to happen either way, but it's a question for @geurts  I think. I could assign `[type="warning"]` to the same style as the default if that's the goal.

components/alert/alert-toast.js
AllanKerr: @Scy-D2L and I talked a bit and she correctly pointed out that the `bottom: 12px` is still necessary in the media query.

components/alert/README.md
```diff
@@ -50,7 +50,6 @@ a pop-up at the bottom of the screen that automatically dismisses itself by defa
 - `no-auto-close` (Boolean, default: `false`) prevents the alert from automatically closing 2.5 seconds after opening.
 - `open` (Boolean, default: `false`): open or close the toast alert.
 - `subtext` (optional, String) The text that is displayed below the main alert message.
-- `type` (String, default: `'default'`): type of the alert being displayed. Can be one of  `default`, `critical`, `success` , `warning`
```
dlockhart, line 53: So despite the design saying this, there are definitely several places that are using toasts of different colours. I specifically noticed [one in quizzing](https://search.d2l.dev/xref/lms/le/quizzing/webpages/user/attempt/quiz_attempt_page_auto.js?r=38b26bb2#227) during my unbundling work where it's using a red and green toast for "Connect Lost" / "Connection Restored" toasts.

So we either:
1. Say "too bad" and force them all to be the default colour
2. Don't officially support the attribute in docs but behind the scenes we keep allowing it
3. Adjust the design somehow to allow for this in edge cases

@geurts @cheics-d2l any thoughts?

components/alert/alert.js
dlockhart: Yeah if there's a clear migration path for those alerts we should take it... I'm guessing to the default. There's also the [MVC version](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Alert/AlertTypes.cs?r=7c3878c6#7) of this control that'll need updating as well.

components/alert/alert.js
```diff
@@ -66,9 +63,12 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 
 			.d2l-alert-text {
 				flex: 1;
-				padding: 1rem 1.2rem 1rem 1.5rem;
+				padding: 0.9rem 1.2rem 0.9rem 1.2rem;
```
dlockhart, line 66: Nit: shorthand `0.9rem 1.2rem`

components/alert/alert.js
```diff
@@ -66,9 +63,12 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 				position: relative;
 			}
+			:host([has-close-button]) .d2l-alert-text {
+				padding-right: 0.9rem;
+			}
```
dlockhart, line 71: If you set any right/left-specific styles, you need to reverse them for RTL:

```css
:host([dir="rtl"][has-close-button]) .d2l-alert-text {
    padding-left: 0.9rem;
    padding-right: 1.2rem;
}
```

components/alert/alert.js
```diff
@@ -78,7 +78,7 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 			}
 
 			.d2l-alert-action {
-				margin: 0.3rem;
+				margin: 0.6rem 0.6rem 0.6rem 0;
```
dlockhart, line 81: Ditto here for RTL.

components/alert/test/alert.visual-diff.js
```diff
@@ -20,7 +20,6 @@ describe('d2l-alert', () => {
 		'type-default',
 		'type-success',
 		'type-critical',
-		'type-warning',
```
dlockhart, line 23: Might be worth adding a few RTL visual diff tests to make sure we don't forget about RTL in the future.

components/alert/README.md
cheics-d2l: We had a previous [conversation here in Slack](https://d2l.slack.com/archives/G0147DRG8CQ/p1589400472017500), which seemed to imply:

2. Don't officially support the attribute in docs but behind the scenes we keep allowing it

As far as colour - I think we could probably get away with changing all the Blue -> Green, but not so much with changing Red -> Blue, since Red is laden with different meaning.

Maybe an option 1/2 combo? 

- Say "too bad" and force all GREEN toasts to blue, don't officially support the attribute but allow legacy toasts to still be red

Again - the quizzing use case is more along "we need a pattern for when a connection is lost and restored", rather than usage of the specific toast pattern.

@geurts Thoughts?

geurts: I am okay with approach # 2. I am also okay with #1.5 if it's not a bunch of extra effort. 

(I guess I need to follow my notifcations in gmail more closely, sorry guys - Scy alerted me to this thread or I wouldn't have found it)

components/alert/test/alert.visual-diff.js
Scy-D2L: [There is one ](https://github.com/BrightspaceUI/core/blob/d33e4be8995142432edd303412f3d35fd606685d/components/alert/test/alert.visual-diff.js#L28) that covers the regular differences, I think I missed it because the padding changes meant all of the v-diff tests were failing either way (and for some reason when I run the tests locally I get the 'before' and 'after' pictures, but not the one that highlights what the difference is). I'm not sure if there's a second case that would be meaningful if added 🤔 open to ideas, though.

What I notice we don't have is a test for what happens on a small-width viewport. I can't recall seeing tests for that elsewhere - are the special spec'd adjustments for the toasts a fairly unique case or is the tiny screen test something we don't usually do?

dlockhart: We typically only do it if there are responsive rules to test. Here's [dialog's example](https://github.com/BrightspaceUI/core/blob/master/components/dialog/test/dialog.visual-diff.js#L43) of it.

dbatiste: For dialogs (and other things) I tried to focus the responsive tests on that would be different.  i.e. there are quite a few tests, and it wasn't really necessary or good use of build-time to repeat them all for both wide & narrow.

Scy-D2L: Given that the only changes are the text size and the distance from the edge of the screen in the case of toasts, I probably only need to have one extra test for this. The other tests here are mostly colors and internal elements that shouldn't need to be duped. :)

Scy-D2L:
Breaking up this PR since it seems like the color updates will need some special handling. The non-color updates are up in [another PR](https://github.com/BrightspaceUI/core/pull/589) already.

dlockhart:
I'm going to close this PR due to staleness for now. When/if we decide to come back to the rest of Scy's changes here, we can take another stab at it.

PR #585, opened 2020-05-29
input-date: min and max

margaree:
Does not deal with validation, just deals with supporting these attributes.

There was some `await`-ing added because of dates needing to be loaded before determining what the focus date should be. Not sure if there is a better way to organize this, but definitely open to suggestions there.


components/calendar/calendar.js
```diff
@@ -575,13 +575,12 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 	updated(changedProperties) {
 		super.updated(changedProperties);
 
-		changedProperties.forEach((oldVal, prop) => {
+		changedProperties.forEach(async(oldVal, prop) => {
```
margaree, line 578: I don't know if this is okay? The `this.updateComplete` was added due to needing the dates to be ready before checking what the focus date should be

components/inputs/test/input-date.visual-diff.html
```diff
@@ -15,26 +15,34 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta http-equiv="X-UA-Compatible" content="ie=edge">
 		<meta charset="UTF-8">
+		<style>
```
margaree, line 18: I did a re-order of these to be alphabetical and have simpler names.
I increased the padding to avoid the other `input-text`s showing up in the visual-diff screenshots and causing a ton of false failures. I can remove this in a subsequent PR where we aren't necessarily concerned with these false failures.

components/inputs/test/input-date.visual-diff.js
```diff
@@ -244,24 +244,55 @@ describe('d2l-input-date', () => {
 			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 		});
 
-		it('open with placeholder', async function() {
-			await helper.open(page, '#no-value');
-			const rect = await helper.getRect(page, '#no-value');
-			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-			await helper.reset(page, '#no-value');
+		describe('placeholder', () => {
```
margaree, line 247: I moved `placeholder` and `min-max` into their own `describe` so that the afterEach to reset (which is needed otherwise close isn't necessarily complete) is unique to each version of `input-date`. I want to do the same for `basic` but it would have caused too many failures of visual diff which could distract from any real failures.

components/calendar/calendar.js
dlockhart: I think it should be fine... it just means the callbacks won't run sequentially which probably doesn't matter.

PR #586, opened 2020-05-29
Serge: add support into core, move lang terms

margaree:
~~Very WIP - opening a PR since this works but I don't want to do too much more until I make sure this is on the path that we were discussing.~~

- Created `LocalizeCoreElement` which handles the dynamic import portion
- Added lang files for ~~en and fr~~ all languages containing all core lang terms
- Modified ~~`input-date` and `more-less`~~ all core components that use lang terms to use `LocalizeCoreElement` ~~for en and fr (will add everything else if this seems good so far)~~

Notes:
- there seems to only be 1 request to fetch the language file (only for the correct language) even when multiple of same component or different components on the page


components/inputs/input-date.js
```diff
@@ -216,7 +216,7 @@ class InputDate extends LocalizeStaticMixin(LitElement) {
 					@mouseup="${this._handleMouseup}"
 					placeholder="${shortDateFormat}"
 					style="${styleMap({maxWidth: inputTextWidth})}"
-					title="${this.localize('openInstructions', {format: shortDateFormat})}"
+					title="${this.localize('inputDate.openInstructions', {format: shortDateFormat})}"
```
dlockhart, line 219: Wondering if it should be `components.inputs.date.openInstructions` in terms of namespacing? We can have as many dots as we want right? Just thinking about when `templates` have translations, or helpers, mixins, etc.

core.serge.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "core",
+  "parser_plugin": {
+    "plugin": "parse_js"
+  },
+  "source_match": "en\\.js$",
+  "source_dir": "src/lang",
```
dlockhart, line 7: Won't it be in just `lang`?

package.json
```diff
@@ -29,6 +29,7 @@
     "/directives",
     "/generated",
     "/helpers",
+    "/lang",
```
dlockhart, line 32: There are a few other places we list out all the directories (like linting). Really wish we'd made a top-level `src` but too late for that!

components/calendar/test/calendar.visual-diff.js
```diff
@@ -145,15 +145,19 @@ describe('d2l-calendar', () => {
 
 	describe('interaction', () => {
 		afterEach(async() => {
-			await page.reload();
+			await page.$eval(firstCalendarOfPage, async(calendar) => {
```
margaree, line 148: The fetching of the files added some time to reloading and having the calendars ready for interaction. If I added `{waitUntil: ['networkidle0', 'load']}` into the args for `reload` then that fixes things, but that also slows down the tests so I rearranged a bit in order to remove the reloading.

components/calendar/test/calendar.visual-diff.js
```diff
@@ -258,48 +262,13 @@ describe('d2l-calendar', () => {
 					await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 				});
 
-				it('END max value', async function() {
```
margaree, line 261: These were rearranged into their own section for a more specified `afterEach`

components/calendar/test/calendar.visual-diff.js
```diff
@@ -389,7 +404,6 @@ describe('d2l-calendar', () => {
 	const tabToDates = async function() {
 		await page.keyboard.press('Tab');
 		await page.keyboard.press('Tab');
-		await page.keyboard.press('Tab');
```
margaree, line 392: The `calendar.focus()` added in the `afterEach` block causes only 2 tabs to be needed

lang/localize-core-element.js
```diff
@@ -0,0 +1,71 @@
+import { LocalizeMixin } from '../mixins/localize-mixin.js';
+
+export const LocalizeCoreElement = superclass => class extends LocalizeMixin(superclass) {
+
+	static async getLocalizeResources(langs) {
+		let translations;
+		for await (const lang of langs) {
+			switch (lang) {
+				case 'ar':
+					translations = await import('./ar.js');
```
svanherk, line 10: Does the delayed imports mean we need to do something special in BSI?

lang/localize-core-element.js
```diff
@@ -0,0 +1,71 @@
...
+					translations = await import('./zh-cn.js');
+					break;
+				case 'zh-tw':
+					translations = await import('./zh-tw.js');
+					break;
+				case 'zh':
+					translations = await import('./zh.js');
+					break;
+			}
+			if (translations && translations.default) {
```
dbatiste, line 58: The static mixin would automatically assign values for missing translations by iterating over the languages in reverse order.  With this new way, how are missing translations handled, say at dev-time, and prod?

lang/ar.js
```diff
@@ -0,0 +1,25 @@
+/* eslint quotes: 0 */
+
+export default {
+	"components.alert.alert.close": "إغلاق التنبيه",
```
dbatiste, line 4: The `components` part of this NS is kind of ambiguous.  `core` seems more meaningful, but I know this matches the folder structure, and also that not everything is a `component`.

core.serge.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "core",
+  "parser_plugin": {
+    "plugin": "parse_js"
+  },
+  "source_match": "en\\.js$",
+  "source_dir": "lang",
+  "output_file_path": "lang/%LANG%.js",
+  "output_lang_rewrite": [
+    "ar-sa ar",
```
dbatiste, line 10: I know this is not special for this PR, but how does this array work?  i.e. I see a culture-specific locale and a non-culture-specific locale separated by a space - how do they get used?

components/alert/alert.js
```diff
@@ -100,28 +100,12 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
-			'zh': { 'close': '关闭警报' }
-		};
-	}
-
 	constructor() {
 		super();
 		this.hasCloseButton = false;
 		this.type = 'default';
+
+		this._namespace = 'components.alert.alert';
```
dlockhart, line 108: Thoughts about flattening the namespace to just `component.<component-name>`? I think we'd be safe enough with `components.alert` and `components.alert-toast` and `components.input-date` for the namespace -- `components.alert.alert.close` seems like overkill.

Also, I think these namespaces could just be constant strings outside the class definition? No need to redefine it for each instance.

```javascript
const LANG_NAMESPACE = 'components.alert';
```

lang/localize-core-element.js
margaree: ah I see how that works - yeah with this current implementation the expectation would be that the translations are complete for each language and so it would just get the first language that has a translations file. I like that idea. It would increase the number of requests (I think up to 4, though that would generally be unlikely and usually 2-3)

components/dialog/dialog.js
```diff
@@ -71,30 +71,11 @@ class Dialog extends LocalizeStaticMixin(AsyncContainerMixin(DialogMixin(LitElem
...
-			'zh-cn': { close: '关闭此对话' },
-			'zh-tw': { close: '關閉此對話方塊' }
-		};
-	}
-
 	constructor() {
 		super();
 		this.width = 600;
+
+		this._namespace = 'components.dialog.dialog';
```
dlockhart, line 78: Nit: For cases like this where there's only 1-2 terms, I don't think I'd pull it out into a variable -- just reference it directly in the `render()`.

lang/ar.js
dlockhart: Haha I asked Margaree to put `components` in as part of the original draft review. Because we're going to have page templates, mixins and other stuff at the top level. `core` is redundant since this lang file is only used within core.

core.serge.json
margaree: So I think it's saying take these translations for this more specific locale code (e.g., ar-sa) and put them in a file called "ar" instead because that's what we're expecting. I do need to fix some of these based on the files we have here.

lang/localize-core-element.js
dlockhart: I'm really, really hoping not. There's a bunch of lang file copying in the BSI build that I don't want to be necessary. I _think_ the fact that we're explicitly listing these all out means the build system can find them.

lang/localize-core-element.js
dbatiste: Besides the extra requests, another downside to what was being done was - it was doing that object merge per instance of the component.  I think we had a store to cache them.

It would be nice to avoid that runtime processing, and also the extra requests?  What are people's thoughts around doing some type of validation to ensure we don't publish with missing translations?

lang/localize-core-element.js
svanherk: I also hope not, although I'm not sure the specific reasons the ones in BSI are needed.

lang/localize-core-element.js
dlockhart: Personally if the extra requests are only helpful during dev... I'd rather not do them. Seems silly for all our users to make 2-3 extra requests forever for no reason just to make dev a bit easier.

I'd be pro some kind of validation around catching missing translations -- there's already some Serge-based validation I believe that tries to catch replace strings that are messed up.

lang/localize-core-element.js
dlockhart: I specifically [added this note](https://github.com/Brightspace/brightspace-integration/blob/master/rollup/rollup-wc.config.js#L50) to the new stuff to try and stop the bleeding.

lang/en.js
```diff
@@ -0,0 +1,25 @@
+/* eslint quotes: 0 */
+
+export default {
+	"components.alert.alert.close": "Close Alert",
```
dbatiste, line 4: FYI - Serge has some capability to accept translation hints that will aid the translator.  With the `parse_js` parser, if the key-value pair line ends with a comment, it will be parsed and passed as the hint for the Serge translators.
```
"components.tabs.tabs.scroll.next": "Scroll Forward", // text for button that scrolls next set of tabs into view
```
Most of our English terms in core are hopefully fairly self-explanatory though.

lang/localize-core-element.js
margaree: For validation, what are thoughts there? My initial thought was adding a script that runs after build when travis is deploying (run it here https://github.com/BrightspaceUI/core/blob/master/.travis.yml#L54 - it could also run as part of the tests stage) which checks if all the lang files have the same keys. It could/probably should also check if the values are empty, though that would be problematic right now since we are missing some translations (should we add english values which would then get translated?).

lang/localize-core-element.js
margaree: I'll test this with my local BSI and LMS. If my LMS is able to load the components and they look fine/have proper lang terms in a couple of languages, is it fine or is there something else I should be watching for?

lang/en.js
margaree: Ah cool that's good to know - thanks!

lang/localize-core-element.js
dlockhart: That should do it. Even just building BSI and then looking in the `build` directory to make sure the files exist should be enough.

lang/localize-core-element.js
dbatiste: I think the deployment stage runs on all builds, ya?  It just doesn't actually deploy to NPM unless it's for a tag.

So... do we want this validation on every build?  and do we want to block PRs if those language files are missing translations?

I think if we go that route, we would manually populate missing values with English while we are waiting for Serge - hopefully that would not cause problems with Serge.  I imagine Serge would eventually just overwrite those values.

Another option is to hook into NPM's `prepublishOnly` step to validate and/or populate missing translation values just for the code being published.  This has the advantage of working in the event that the package needs to be published outside of CI.  It leaves the translation files in the source unmodified, which is good in the sense that we aren't doctoring the source files, but bad in the sense that there would be missing translation values while dev'ing on core - which is perhaps not a big deal.

margaree: So just to confirm I understand what using `prepublishOnly` would do, is that we would run a script that would replace missing translations with some value which would then go into the publish code, but would not be in the source code? If this is the case, would we still want to have some sort of alert to whoever published?

lang/localize-core-element.js
margaree: Cool, seems fine. Files are present in the build, more/less loads as expected in discussions for en and ar

components/dialog/dialog.js
```diff
@@ -71,27 +71,6 @@ class Dialog extends LocalizeStaticMixin(AsyncContainerMixin(DialogMixin(LitElem
...
-			'es': { close: 'Cerrar este cuadro de diálogo' },
-			'fr': { close: 'Fermer cette boîte de dialogue' },
-			'ja': { close: 'このダイアログを閉じる' },
-			'ko': { close: '이 대화 상자 닫기' },
-			'nl': { close: 'Dit dialoogvenster sluiten' },
-			'pt': { close: 'Fechar esta caixa de diálogo' },
-			'sv': { close: 'Stäng dialogrutan' },
-			'tr': { close: 'Bu iletişim kutusunu kapat' },
-			'tr-tr': { close: 'Bu diyalog kutusunu kapat' },
-			'zh': { close: '关闭此对话框' },
```
margaree, line 89: @dbatiste It seems that we generally assign `zh-cn` translations to `zh` - do you know why there is a separate `close` lang term for `zh` vs `zh-cn`? Similarly with Turkish we generally just have `tr`.

dbatiste: I think I copied from elsewhere, so no other particular reason.

lang/localize-core-element.js
dbatiste: Yeah that's what the idea would be if we want to hook into `prepublishOnly`.  

We could write a message into the log but I suspect it would almost always go unnoticed.  I can't speak for everyone, but I usually only look if CI failed, I suspect NPM publish didn't work, or I am checking which goldens were updated.

I tend to lean that way, but I don't feel super strongly about it.  It also makes it a bit easier to identify untranslated terms when looking at the code.

If we were to go the other route, and populate English values for missing translations and check them into GitHub while we're waiting for the Serge process, would those values cause problems for Serge, or will it still identify them as being untranslated and post an update for them?

margaree: Based on the my courses PR request here https://github.com/Brightspace/d2l-my-courses-ui/pull/821/files it looks like if we use English values they will be replaced with translations. @svanherk was there anything extra needed to have translations done that way or was that just automatically how it worked?
If that's the case then I think it would probably just be easiest to populate with English values pre-translation.

svanherk: The way I believe it works is:
1. Serge will look to see if you've added or changed any English terms.
2. It will send only those terms to the translators, regardless of whether you have those terms translated already or not.
3. Serge will come back with the translated terms and open a PR.

Because we're setting serge up for the first time, all the English terms we have will be sent.  We might get back some different terms for the ones we already translated, and the English terms in other languages should be replaced with the correct translated ones.

So sort answer: Yes, I think you're good to populate untranslated terms with the English versions and serge will properly replace them

dbatiste: My only reservation with populating in our source code is that it makes it harder for us to identify untranslated terms thereafter.  My suspicion is that the logic for validating is almost identical to populating, and the timing of when it gets called doesn't really introduce any difference - it just gets tacked onto existing script vs adding a line to package.json for `prepublishOnly`.

margaree: okay yeah that makes sense to me. I'll work on adding a script to run as part of `prepublishOnly`

cli/verify-translations.js
```diff
@@ -0,0 +1,51 @@
+const chalk = require('chalk'),
+	fs = require('fs'),
+	path = require('path');
+
+const langs = ['ar', 'da', 'de', 'es', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
+const langTermsPath = path.join(__dirname, '../lang');
+
+function _parseFile(fileName) {
+	const file = fs.readFileSync(`${langTermsPath}/${fileName}.js`).toString();
+	const stringContent = file.split('default ')[1].split(';')[0];
```
margaree, line 10: This relies on the only semi-colon being at the end of the file. Is this too risky and I should split on something like `};` and then add the `}`?

dlockhart: Seems OK for now. NPM is pretty close to supporting ES modules natively, although it may require that the extension be `.mjs` unfortunately.

PR #589, opened 2020-06-01
Update alert spacing and toast dismissal

Scy-D2L:
Stripped down version of [the alert design-update PR](https://github.com/BrightspaceUI/core/pull/577/files) with the non-color-related updates.

- Spacing/padding/margins of alerts and toasts adjusted to spec (inc new v-diff tests)
- Toast auto-close is disabled while hovered or focused and restarts when mouse or focus exits.


components/alert/alert-toast.js
```diff
@@ -97,8 +104,20 @@ class AlertToast extends LitElement {
...
-				<d2l-alert type="${ifDefined(this.type)}" @d2l-alert-closed=${this._onCloseClicked} button-text="${ifDefined(this.buttonText)}" ?has-close-button="${!this.hideCloseButton}" subtext="${ifDefined(this.subtext)}">
+			<div
+				class="d2l-alert-toast-container"
+				data-state="${this._state}"
+				@transitionend=${this._onTransitionEnd}>
+				<d2l-alert
+					@blur=${this._closeTimerStop}
+					button-text="${ifDefined(this.buttonText)}"
+					@d2l-alert-closed=${this._onCloseClicked}
+					@focus=${this._closeTimerStart}
```
dlockhart, line 115: Shouldn't focus and blur be reversed? Like blurring should start the timing and focussing should stop it?

Scy-D2L: Yep! 😕 Thanks, not sure how that slipped through.

components/alert/alert.js
```diff
@@ -127,12 +152,16 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 		return html`
 			<div class="d2l-alert-highlight"></div>
-			<div class="d2l-alert-text">
+			<div class="d2l-alert-text d2l-body-standard">
 				<slot></slot>
 				${this.subtext ? html`<p class="d2l-body-compact d2l-alert-subtext">${this.subtext}</p>` : null}
 			</div>
-			${this.buttonText && this.buttonText.length > 0 ? html`<d2l-button-subtle class="d2l-alert-action" text=${this.buttonText} @click=${this._onButtonClick}></d2l-button-subtle>` : null}
-			${this.hasCloseButton ? html`<d2l-button-icon class="d2l-alert-action" icon="d2l-tier1:close-default" text="${this.localize('close')}" @click=${this.close}></d2l-button-icon>` : null}
+			<div class="d2l-alert-action">
```
dlockhart, line 159: So by always wrapping these with `d2l-alert-action`, I noticed that when an alert doesn't have a close button or buttonText, there's an extra `40px` of space taken up by this container.

Also, the spec says that the text container should have 30px (1.5rem) of padding around it, but on the right it only has 24px currently.

Since the padding of the `d2l-alert-text` container is conditional on there being actions or not, I think I'd use conditional classes. So like `d2l-alert-text` gives it 1.5rem of padding but then it would add `d2l-alert-text-with-actions` if there are actions that would reduce the right padding.

To simplify things, I might also suggest positioning the coloured band absolutely -- that way it isn't taking up any actual space, it would just float on top.

components/alert/alert-toast.js
```diff
@@ -97,8 +104,20 @@ class AlertToast extends LitElement {
...
-				<d2l-alert type="${ifDefined(this.type)}" @d2l-alert-closed=${this._onCloseClicked} button-text="${ifDefined(this.buttonText)}" ?has-close-button="${!this.hideCloseButton}" subtext="${ifDefined(this.subtext)}">
+			<div
+				class="d2l-alert-toast-container"
+				data-state="${this._state}"
+				@transitionend=${this._onTransitionEnd}>
+				<d2l-alert
+					@blur=${this._closeTimerStart}
+					button-text="${ifDefined(this.buttonText)}"
+					@d2l-alert-closed=${this._onCloseClicked}
+					@focus=${this._closeTimerStop}
```
dlockhart, line 115: Another slight hick-up I noticed: if the toast has focus, we never want it to close. Currently you can open it, focus on it, then move your mouse away and it'll close.

A good solution to this might be to track whether focus is inside or not (using the focus and blur events) and then in `_closeTimerStop` just bail if focus is still inside.

components/alert/alert-toast.js
```diff
@@ -118,6 +137,19 @@ class AlertToast extends LitElement {
 		}
 	}
 
+	_closeTimerStart() {
+		if (!this.noAutoClose) {
+			const duration = this.buttonText ? 10000 : 4000;
```
dlockhart, line 142: I think there are or will be scenarios where multiple things can cause `_closeTimerStart()` to run. For example, you could mouse over, then focus, then mouse out, then blur, causing 2 calls.

The best approach here might be to clear any existing timeouts before creating a new one.

components/alert/alert-toast.js
Scy-D2L: How do we approach the situation where you move focus and still have the mouse overtop? I think I'd have to track `hasMouseover` and `hasFocus` and make sure it's got neither before it starts closing if these are equivalent prompts to keep it open.

dlockhart: Yep you're right, you'd need to track both. Fun times!

components/alert/alert.js
```diff
@@ -41,6 +42,7 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 				left: 0;
 				margin: -1px;
 				min-width: 0.3rem;
+				position:absolute;
```
dlockhart, line 45: Nitty nit: space

components/alert/alert.js
```diff
@@ -66,19 +68,33 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
+				padding-left: 0.9rem;
 			}
+
 			.d2l-alert-subtext {
 				margin: 0.5rem 0 0;
 			}
 
 			.d2l-alert-action {
-				margin: 0.3rem;
+				margin: 0.6rem;
```
dlockhart, line 87: I think the design calls for only the 18px between the actions and the text, so with this going all the way around it ends up with (0.9rem + 0.6rem) = 1.5rem = 30px instead of 18.

components/alert/alert.js
```diff
@@ -125,14 +141,20 @@ class Alert extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 	}
 
 	render() {
+		const hasActions = this.buttonText && this.buttonText.length > 0  || this.hasCloseButton;
 		return html`
 			<div class="d2l-alert-highlight"></div>
-			<div class="d2l-alert-text">
+			<div class="d2l-alert-text ${hasActions ? 'd2l-alert-text-with-actions' : null} d2l-body-standard">
```
dlockhart, line 147: The classMap stuff is handy to make this cleaner.

components/alert/alert.js
Scy-D2L: Somehow I'm surprised we don't lint this.

components/alert/alert.js
dlockhart: We probably could! 🤔

components/alert/alert-toast.js
```diff
@@ -22,6 +22,8 @@ class AlertToast extends LitElement {
 			open: { type: Boolean, reflect: true },
 			subtext: { type: String },
 			type: { type: String, reflect: true },
+			_hasFocus: { type: Boolean },
+			_hasMouse: { type: Boolean },
```
dlockhart, line 26: If you make these properties that Lit is tracking, then it'll cause `render()` to be called when they change. I don't _think_ you need that, so just exclude them from here and set them in the constructor like you already are and you should be good.

components/alert/alert-toast.js
```diff
@@ -118,11 +141,45 @@ class AlertToast extends LitElement {
 		}
 	}
 
+	_closeTimerStart() {
+		clearTimeout(this._setTimeoutId);
+		if (!this.noAutoClose && !this._hasFocus && !this._hasMouse) {
```
dlockhart, line 146: I think you want the focus and mouse checks to move inside the timeout. Otherwise I could mouse over, mouse out, then focus and it'd close.

dlockhart: Hmm actually no it wouldn't, since focusing would cancel the timeout. So probably OK!

components/alert/alert-toast.js
```diff
@@ -118,11 +141,45 @@ class AlertToast extends LitElement {
...
+			const duration = this.buttonText ? 10000 : 4000;
+			this._setTimeoutId = setTimeout(() => {
+				this.open = false;
```
dlockhart, line 149: We need a good place to reset `_hasFocus` and `_hasMouse` to `false`, otherwise if you open and then close and then open it again they'll be their old values. Maybe inside here where we set `open = false` is a good place?

Scy-D2L: If we reset it there it seems like we ought to reset it when we click the close button too, or there'd be a similar problem.

dlockhart: Yeah, so maybe on the `updated` when `open` changes eh?

dlockhart: Otherwise `elem.open = false` won't reset it.

PR #594, opened 2020-06-03
S akulick/prevent double lable

Scy-D2L:
Using a span for the label on the time input caused the screenreader on iOS to give it independent focus and read it twice.


components/inputs/input-time.js
```diff
@@ -170,17 +170,20 @@ class InputTime extends LitElement {
 	render() {
 		initIntervals(this.timeInterval);
 		const input = html`
-			<span class="${this.label && !this.labelHidden ? 'd2l-input-label' : 'd2l-offscreen'}" id="${this._dropdownId}-label">${this.label}</span>
+			<label
+				class="${this.label && !this.labelHidden ? 'd2l-input-label' : 'd2l-offscreen'}"
+				for="${this._dropdownId}-input"
+				id="${this._dropdownId}-label">${this.label}</label>
```
dlockhart, line 176: Does it still need an `id` now that nothing is `aria-labelledby` pointing at it?

Scy-D2L: It does, because the [dropdown menu](https://github.com/BrightspaceUI/core/blob/3667a91c1db69d6525eabef7a3566379da2ca745/components/inputs/input-time.js#L196) still needs to know what's labelling it - otherwise NVDA says '`undefined` menu' when identifying it.

dlockhart: Ah I see that down there now -- thanks.

PR #595, opened 2020-06-03
Hide the tabs (without animation) if there are less than two initially.

dbatiste:
Previous work to hide the tablist when there are less than two apparently only considered adding/removing tabs after the initial rendering.  This change hides the list of tabs if there are initially less than two.

It must opt out of animation however, similar to reduced motion so that a) the user doesn't see animation as soon as the page loads; and b) the component isn't in an animating state while hidden, which would prevent the `transitionend` event from firing, causing subsequent problems displaying the list later if tabs are added.


dbatiste:
Did not really expect those two other golden image updates.  They are the last two fixtures, but the page height should be enough to accommodate.  They look the same from what I can tell - will roll with it and see if this comes up again.

PR #597, opened 2020-06-04
Keyboard traversal added to list-item-generic-layout

zinabat:
This is a large PR because all of the functions for DOM traversal of the grid cells had to be written. Don't be scared! 😸 

## **Context**
[US115381](https://rally1.rallydev.com/#/detail/userstory/382446116064?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#Generic-List-Item)

Kapture 2020-06-04 at 12 39 29

This adds keyboard traversal to the `list-item-generic-layout` component and follows the `aria-grid` guidelines:
* **ArrowLeft** moves to the next left item in a row
* **ArrowRight** moves to the next right item in a row
* **ArrowUp** moves to the same item in the row above, if available
* **ArrowDown** moves to the same item in the row below, if available
* **PageUp** moves to the same item in the row **five** rows above, if available
* **PageDown** moves to the same item in the row **five** rows below, if available
* **Home** moves to the first item in the row
* **Ctrl+Home** moves to the first item of the first row
* **End** moves to the last item in the row
* **Ctrl+End** moves to the last item of the last row
* **Space** and **Enter** simulate a click on the focused item

Cells are defined with the `.d2l-cell` class and `data-cell-num` attribute to determine cell order, which is different from the DOM ordering of each cell.

### **Changes to `d2l-list`**
* The grid is turned on by adding the `grid` boolean attribute to the parent list
* The role of a `list-item` will be set to `rowgroup` if grid is enabled, `listitem` if not. Similarly, the `d2l-list` role is set to `list` if grid is disabled, nothing if enabled.

## **Quality**
* New tests for all key actions have been added
* Manual testing performed on browsers:
  - [x] Legacy Edge
  - [x] Chrome
  - [ ] CrEdge
  - [x] Firefox
  - [x] Safari

### **Known Limitations**
* `role="grid"` is not well-supported cross browser and reads out misinformation, thus was not added to `d2l-list`
* The `d2l-list` component should become a single tab stop in future work
* Usability instructions for use of the keyboard while in grid mode should be added in future work
* Attempting to PageUp to a row that does not have the same cell as the current focused item will not move focus. This is by design, but we may want to default to the first focusable cell instead.


components/list/list.js
```diff
@@ -34,13 +36,22 @@ class List extends LitElement {
...
 				<slot></slot>
 			</div>
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('grid')) {
+			const items = this._getItems();
+			items.forEach(item => item.role = this.grid ? 'rowgroup' : 'listitem');
```
dbatiste, line 51: It's not super amazing for the `d2l-list` to reach down to its children and mutate them.  We do do this sort of thing every once in a while though, but it always feels wrong.  

Currently this is tied into the update for the `grid` property of the list, which won't catch cases where additional items are added to the list (ex. load more), or if items are transplanted from one list to another.  A better place to capture those scenarios is in a slotchange` handler for the slot.

An alternative approach (that we've also utilized) is to have the item check it's closest `d2l-list` ancestor when it's connected. It is possible to have timing/race issues though, and unless the `grid` attribute of the list os observed, it won't be notified of changes... so also not amazing.

zinabat: Yeah, I agree this felt weird. Your point about an update to the list items not occurring when transplanted is important - I think using a `slotchange` handler is the best option for this, partly to avoid having to observe changes to the ancestor `d2l-list` from the list item.

@m6jones What do you think?

components/list/list.js
```diff
@@ -69,7 +80,7 @@ class List extends LitElement {
 
 	_getItems() {
 		return this.shadowRoot.querySelector('slot').assignedNodes().filter((node) => {
```
dbatiste, line 82: We don't need to add it right away, but something to be mindful/careful of is - slotted elements of the children's slots that are being distributed to this slot (nested slots).  To resolve that scenario, the flatten option can be passed.  `slot.assignedNodes({flatten: true})`.  Not something that is needed right away, but definitely something that could come up depending on how the consumer is composing the child items.

components/list/list.js
m6jones: Slot change would be the best. 

Another approach which might be better is I would expect this value not to change. Ie, you make an aria-grid navigation or list navigation. The application would be super confusing if we switched it up it locations. UX would be terrible.

So doing the alternative approach that Dave suggested is good since the problem shouldn't happen. We could enforce it by having d2l-list and d2l-list-aria-grid components (better names pending)

zinabat: I'll change this to be controlled from the `list-item-generic-layout` checking its ancestor on connection

components/list/demo/list-item-sample.js
```diff
@@ -46,11 +48,19 @@ class ListItemSample extends ListItemCheckboxMixin(LitElement) {
 		this._contentId = getUniqueId();
 	}
 
+	connectedCallback() {
```
dbatiste, line 51: I think this logic needs to be moved into a mixin.

zinabat: hmm, good point.

zinabat: So, I have moved this `connectedCallback` logic hunting for `d2l-list` as a parent to `ListChildMixin`; this seems sufficient. The grid can still be turned on for those implementing their own items by setting `grid-active` on the generic-layout component in whatever way they choose.

components/list/list-item-generic-layout.js
```diff
@@ -1,8 +1,35 @@
...
+	RIGHT: 39,
+	SPACE: 32,
+	UP: 38
+};
+
+class ListItemGenericLayout extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			role: { type: String, reflect: true },
```
dbatiste, line 29: Looks like this `role` property is not needed here.

dbatiste: Oh I see it is calculated in the connectedCallback.

components/list/list.js
```diff
@@ -1,5 +1,6 @@
 import '../colors/colors.js';
```
dbatiste, line 1: Looks like this colors import is not needed here.

components/list/list-child-mixin.js
```diff
@@ -0,0 +1,18 @@
+import { findComposedAncestor } from '../../helpers/dom.js';
+
+export const ListChildMixin = superclass => class extends superclass {
```
dbatiste, line 3: I think this should be called `ListItemMixin`.

PR #602, opened 2020-06-08
[a11y] Calendar: Use presentation & application roles to have more control over screen readers

margaree:
Problem:
Huge inconsistencies across screen reader/browser combinations in terms of what information is announced and in what order (e.g., a main functionality we want is for screen readers to read the column header, the weekday, first so users can quickly navigate across rows).

I'll comment on the changes made throughout, which give us much more control over how the calendar is presented (though they sacrifice built-in table semantics which was the hesitation for doing this).

Overall the desired experience (which has now been achieved across most screen reader/browser combinations):
- User enters into table, "[month] [year] table" is announced to them _(not working on Legacy Edge + NVDA and VoiceOver + iOS Safari)_
- as user arrow key navigates around, on each day they here (before anything else) "[weekday] [date] [selection state] [month] [year]" and maybe something about them being on a button.
- Page up/down will announce month/year when month changes _(does not work on Firefox + JAWS)_
- User does not have to swipe through the entire header row before getting into the dates _(was a problem on iOS)_


components/calendar/calendar.js
```diff
@@ -544,29 +537,30 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
-						text="${this._computeText(getPrevMonth(this._shownMonth))}"
-						icon="tier1:chevron-left">
-					</d2l-button-icon>
-					<h2 aria-atomic="true" aria-live="polite" class="d2l-heading-4" id="${labelId}">${heading}</h2>
-					<d2l-button-icon
-						@click="${this._onNextMonthButtonClick}"
-						text="${this._computeText(getNextMonth(this._shownMonth))}"
-						icon="tier1:chevron-right">
-					</d2l-button-icon>
+				<div role="application">
```
margaree, line 540: `role="application"` gives us control over the key events, which I believe was most problematic for NVDA + Firefox

components/calendar/calendar.js
```diff
@@ -544,29 +537,30 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
+							icon="tier1:chevron-left">
+						</d2l-button-icon>
+						<h2 aria-atomic="true" aria-live="polite" class="d2l-heading-4" id="${labelId}">${heading}</h2>
+						<d2l-button-icon
+							@click="${this._onNextMonthButtonClick}"
+							text="${this._computeText(getNextMonth(this._shownMonth))}"
+							icon="tier1:chevron-right">
+						</d2l-button-icon>
+					</div>
+					<table aria-labelledby="${labelId}" role="presentation">
```
margaree, line 554: `role="presentation"` lets us control what information is communicated to the user for the most part.

margaree:
Taking another look at the button vs. div change so this can be ignored for now

components/calendar/calendar.js
```diff
@@ -490,39 +486,36 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
...
 				const date = day.getDate();
-				const description = `${date}. ${selected ? this.localize('selected') : this.localize('notSelected')} ${formatDate(day, {format: 'monthYear'})}`;
+				const weekday = calendarData.descriptor.calendar.days.long[calendarData.daysOfWeekIndex[index]];
+				const description = `${weekday} ${date}. ${selected ? this.localize('selected') : this.localize('notSelected')} ${formatDate(day, {format: 'monthYear'})}`;
 				// role="gridcell" used for screen reader (e.g., JAWS and VoiceOver) behavior to work properly
 				return html`
 					<td
-						aria-selected="${selected ? 'true' : 'false'}"
+						aria-disabled="${disabled}"
+						aria-selected="${selected}"
```
dlockhart, line 509: I think this should still be a string and not a boolean being passed in... it'll probably `.toString()` it and still work but just for safety maybe?

components/calendar/calendar.js
```diff
@@ -794,6 +789,10 @@ class Calendar extends LocalizeStaticMixin(RtlMixin(LitElement)) {
 		}
 	}
 
+	_onMousedown(e) {
+		e.preventDefault();
```
dlockhart, line 793: What's this doing?

margaree: This was since I switched the button to a div, if the button was disabled it could still receive focus when clicked so this prevented it. I've figured out that I can change the div back to a button so I'm going to get rid of that portion of the change.

dlockhart: Cool sounds good!

components/calendar/calendar.js
dbatiste: I think I actually ran into a problem with this (using a boolean) not too long ago - I think with tabs.  I don't recall whether it was a linting/analyzer error, or an actual error though.

PR #604, opened 2020-06-08
Form Validation: `d2l-validation-custom`

AllanKerr:
# Changes
- Add a `ValidationCustomMixin` and `d2l-validation-custom` component to allow native and custom form elements to support `async` validation.
- Parent forms or form elements will receive the `d2l-validation-custom-connected` events to know that they should call `validate`.

## `ValidationCustomMixin`
- This mixin was created so frequently used custom validators can be created and reused by implementing `validate()` to do anything.

## `d2l-validation-custom`
- This component allows validation to be performed in the `d2l-validation-custom-validate` event handler. The event handler is responsible for calling `e.detail.resolve(true)` or `e.detail.resolve(false)` to indicate if validation succeeded or failed.

## Example
```
<d2l-form>
    <d2l-validation-custom for="mycheck" @d2l-validation-custom-validate=${this._validateCheckbox} failure-text="${this.localize('checkValidationFailure', { label: this.localize('checkTitle') })}" >
    </d2l-validation-custom>
    <input type="checkbox" id="mycheck" name="checkers"/>
</d2l-form>
```


components/validation/validation-custom-mixin.js
```diff
@@ -0,0 +1,51 @@
+
+export const ValidationCustomMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			failureText: { type: String, attribute: 'failure-text' },
```
AllanKerr, line 6: Does anyone have any thoughts on whether we want `failure-text` as a property or if we want to be more flexible by instead allowing a `failureText` string to be returned from `validate`?

components/validation/validation-custom-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	get target() {
+		return this._target;
+	}
+
+	async validate() {}
+
+	_updateTarget() {
+		const root = this.getRootNode();
```
dlockhart, line 47: Interesting, I hadn't seen this DOM API before. Is this roughly equivalent to `this.shadowRoot`?

components/validation/validation-custom-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+		this._target = root.getElementById(this.for);
```
dlockhart, line 48: Wondering if we should throw here if we don't find the target? Otherwise page developers may not realize why things later don't behave properly...

components/validation/validation-custom-mixin.js
dlockhart: That's a good question. I'm trying to think of a scenario where a single validator could result in different failure messages, and I'm not coming up with a good one. Very often there might be different error conditions (e.g. number must be greater than 4 AND less than 10), but in those cases we'd just craft an error message to reflect both conditions I think (e.g. "Number must be between 4 and 10"). And worst case for very complex scenarios they could just use multiple custom validators?

components/validation/validation-custom-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+				this._updateTarget();
+			}
+		});
+	}
+
+	get target() {
+		return this._target;
+	}
+
+	async validate() {}
```
dlockhart, line 44: Should this return `true` be default? Can't think of why someone wouldn't implement this though.

components/validation/validation-custom.js
```diff
@@ -0,0 +1,18 @@
...
+import { LitElement } from 'lit-element/lit-element.js';
+import { ValidationCustomMixin } from './validation-custom-mixin.js';
+
+class ValidationCustom extends ValidationCustomMixin(LitElement) {
+
+	async validate() {
+		const validation = new Promise(resolve => {
+			const details = { detail: { target: this.target, resolve } };
+			const event = new CustomEvent('d2l-validation-custom-validate', details);
+			return this.dispatchEvent(event);
```
dlockhart, line 11: What would it look like for someone to handle this event in practice? Something like...

```html
<d2l-validation-custom @d2l-validation-custom-validate="${this._validate}">
```

And then:
```javascript
_validate(e) {
    const val = e.detail.target.value;
    const isValid = val < 10 && val > 4;
    e.detail.resolve(isValid);
}
```

It'd be amazing if they could just return `true`, `false` or a promise from their event handler, but I can't think of a way to make that work. Plus I guess the way it's set up now, in theory there could be _many_ event handlers and only one of them needs to handle it (and could use `preventDefault` to stop the others), so... cool!

Also, events have targets already (`e.target` in this case, which will be the `d2l-validation-custom` element), so it's a little confusing for it to have `e.target` and `e.detail.target`. Trying to think of a better name... something like `forElement` maybe?

components/validation/test/validation-custom.test.js
```diff
@@ -0,0 +1,70 @@
...
+			const expectedTarget = root.querySelector('#target');
+			expect(custom.target).to.equal(expectedTarget);
+		});
+
+	});
+
+	describe('events', () => {
+
+		it('should fire connected event', async() => {
+			const custom = document.createElement('d2l-validation-custom');
```
dlockhart, line 43: Careful with scopes here... there's already a `custom` defined above so this initially confused me.

components/validation/test/validation-custom.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+			it('should fire validate event', async() => {
+				const validationHandler = async(e) => {
+					await aTimeout(0);
+					e.detail.resolve(expectedIsValid);
+				};
+				custom.addEventListener('d2l-validation-custom-validate', validationHandler);
+				const isValid = await custom.validate();
+				expect(isValid).to.equal(expectedIsValid);
+			});
```
dlockhart, line 64: Maybe worth testing this with a non-async handler?

components/validation/validation-custom-mixin.js
dbatiste: Yep, except that if it's in the document's context, it will return that instead.

components/validation/validation-custom-mixin.js
AllanKerr: I like throwing here (or at the very least warning) if `for` is defined and a target isn't found. I am planning on allowing `for` to omitted which will cause whatever handles `d2l-validation-custom-connected` to be the `target` but that shouldn't cause an issue here if we only throw if `for` is set.

components/validation/validation-custom-mixin.js
AllanKerr: I'm fine either way. I didn't add a return value because the sole purpose of the mixin is to have users override `validate`,

components/validation/validation-custom-mixin.js
AllanKerr: I was thinking there could be cases where server-side validation would be performed and the server would return an error message or enum but I can't think of any exact cases off the top of my head. Having multiple custom validators is always an option but it could get kind of cumbersome and would result in duplicate validation calls if its to a single API call with multiple messages.

components/validation/validation-custom.js
AllanKerr: Yep, thats exactly how it would be implemented. I did try to come up with options to just allow the value to be returned rather than having to pass in a `resolve` callback but as soon as the validate handler becomes `async` `e.preventDefault` doesn't work and return values can't be received anymore.

I'm happy renaming `target` to something else to avoid confusion.

PR #605, opened 2020-06-09
stickify footer in FACE mobile

spandanabaruah:
https://rally1.rallydev.com/#/29180338367d/detail/userstory/391269578948?fdp=true


templates/primary-secondary/primary-secondary.js
```diff
@@ -74,6 +74,15 @@ class TemplatePrimarySecondary extends LitElement {
...
 					grid-template-rows: auto 1px auto;
+					overflow-y: auto;
+				}
+				main {
+					grid-area: primary;
+					overflow-y: visible;
+				}
+				aside {
+					grid-area: secondary;
+					overflow-y: visible;
```
dlockhart, line 85: Do you need to specify the `grid-area` here? `main` should already be in `primary` and `aside` should already be in `secondary` from the CSS defined above.

spandanabaruah: You're right - will put up a new commit without the grid-area specified again in the media query.

PR #610, opened 2020-06-11
d2l-backdrop: Allow longer transition, and no transition on hide

emil-furniss:
In FACE, we would like to use `d2l-backdrop` as a "saving shim" to obscure the activity editor while we save the activity. Due to the save-cancel model being used there, saving can take some time due to several API calls - sometimes in the order of seconds. This results in bad UX if no immediate feedback is given on the "Save & Close" button.
However, design requested that the transition be longer in this case (1200ms) and that there be no transition when hiding the backdrop in the case of a mid-save or validation error.

To enable this functionality, this change adds 2 optional and independent attributes:
1. `slow-transition` which increases the transition time to 1200ms. It was suggested that having a second preset would be preferable to a freely configurable value, so different consumers can have a choice but still remain consistent.
2. `cut-out` which disables the fade-out transition while the backdrop is being hidden


dbatiste:
> design requested that the transition be longer in this case (1200ms) and that there be no transition when hiding the backdrop in the case of a mid-save or validation error.

What's the rationale behind this?  i.e. 1.2s seems like a long time... do they really want a longer animation, or is it to create some perceived delay (which should be accomplished with an actual delay)?

components/backdrop/backdrop.js
```diff
@@ -14,6 +14,7 @@ class Backdrop extends LitElement {
 		return {
 			shown: { type: Boolean },
 			forTarget: { type: String, attribute: 'for-target' },
+			cutOut: { type: Boolean, attribute: 'cut-out' },
```
dbatiste, line 17: I think something like `transitions: 'show-hide | show-only'` (string) or `no-animate-hide` (boolean) would be clearer.   In either case, if we ever need to expand on this to include _show_, we can add another valid string value `show-only`, or with the boolean approach another property, `no-animate-show`.  The `no-*` naming has some precedent in our other components.

emil-furniss: Sounds good, I'll change it to `no-animate-hide`. I know my name was terrible ^^
And yes, design explicitly wanted the animation to be a lot slower. I think it's about giving feedback that something is happening, while hacking user perception when the save process is being slow. 1.2s was arrived at by demoing different values on my local instance.

dbatiste: One thing to note (not sure if the designer is aware) is that if the user has reduced motion preference set, the backdrop will be shown immediately.

emil-furniss: Right, yes. I had mentioned this before in passing, but will make sure they are aware.

PR #611, opened 2020-06-11
Add localize-mixin deduping and collation

AllanKerr:
# Changes
- Add deduping and collation to the `localize-mixin` so that multiple mixins can contain their own lang terms without interfering with other `localize-mixin`s in the inheritance chain or terms on the component itself. Now all terms will be combined so that they can all be used.


mixins/localize-mixin.js
```diff
@@ -20,13 +20,21 @@ export const LocalizeMixin = superclass => class extends superclass {
...
+						if (results.length === 0) {
 							return;
 						}
-						this.__language = res.language;
-						this.__resources = res.resources;
+						const resources = {};
+						for (const res of results) {
+							const language = res.language;
+							for (const [key, value] of Object.entries(res.resources)) {
+								resources[key] = { language, value };
```
dlockhart, line 34: Interesting! So now in theory each group of resources could be in a different language.

mixins/localize-mixin.js
```diff
@@ -143,10 +152,26 @@ export const LocalizeMixin = superclass => class extends superclass {
...
+	static _getAllLocalizeResources(possibleLanguages) {
+		let resourcesLoadedPromises;
+		const superCtor = Object.getPrototypeOf(this);
+		if ('_getAllLocalizeResources' in superCtor) {
+			resourcesLoadedPromises = superCtor._getAllLocalizeResources(possibleLanguages);
+		} else {
+			resourcesLoadedPromises = [];
+		}
+		// eslint-disable-next-line no-prototype-builtins
+		if (this.hasOwnProperty('getLocalizeResources') || this.hasOwnProperty('resources')) {
```
dlockhart, line 164: Could this be an `&&` instead of an `||`? Wondering if the way it's written, a component could have a `resources` property but not be using localization at all but it would now get picked up by this code. A long shot, but it'd be nice to confirm that they have `resources` AND they're actually using the localize mixin.

mixins/localize-mixin.js
```diff
@@ -20,13 +20,21 @@ export const LocalizeMixin = superclass => class extends superclass {
...
+					.then((results) => {
+						if (results.length === 0) {
 							return;
 						}
-						this.__language = res.language;
-						this.__resources = res.resources;
+						const resources = {};
+						for (const res of results) {
+							const language = res.language;
+							for (const [key, value] of Object.entries(res.resources)) {
```
dlockhart, line 33: Wondering if we can intercept the case where two mixins are specifying the same key and maybe throw? Otherwise I think depending on which order the promises resolve they'll get random results as to which one gets used.

mixins/localize-mixin.js
AllanKerr: This is written as an `||` because the `_hasResources` checks that `getLocalizeResources` exists somewhere in the inheritance chain. If it doesn't exist then this code is skipped altogether.

The `||` is needed because when using `LocalizeStaticMixin`, `getLocalizeResources` is defined in the superclass while `resources` is defined in the subclass so `this.hasOwnProperty('getLocalizeResources')` is only `true` for the superclass. Since the `LocalizeStaticMixin` defines an empty `resources` block calling `getLocalizeResources` from the `superclass` causes the actual terms to be missed.

mixins/localize-mixin.js
AllanKerr: The order shouldn't be random. In `_getAllLocalizeResources` resources are loaded starting at the top of the inheritance chain working down so that the order of `localizeResources` has the same order. This allows mixins lower in the inheritance chain to override terms defined higher up in the inheritance chain.

Whether that's behavior we want or not, I'm open to convincing either way.

mixins/localize-mixin.js
AllanKerr: Yeah, it seemed like the best way to guarantee backwards compatibility while still allowing language fallbacks. I don't expect it to come up in production a lot but will likely happen a lot in dev.

mixins/localize-mixin.js
dlockhart: Yeah I guess the order of the promises will be consistent... they may resolve out of order but you're right the array itself will be consistent.

I don't see this cropping up too often (it wasn't even possible until this PR!) so maybe we can ignore it. Mostly if we wanted to throw, now's our only opportunity.

AllanKerr: I do think allowing overriding might be slightly more intuitive.

The main issue I see with throwing an exception is that adding a term with a colliding key in a mixin could cause a component using that mixin to blow up even if the mixin itself works and passes all tests. 

Given that most of our UI repos don't have scheduled CI runs it could take a long time for it to be noticed if its in a repo that isn't having a ton of work done in it.

dlockhart: Cool sounds good 👍

PR #612, opened 2020-06-11
verify-translations script: sort keys

margaree:
Otherwise new keys get added to the end instead of in alphabetical order


margaree:
Looks like the visual diff tests are failing because input-date has one diff off by 2 pixels. I can update the goldens in here to get that fixed.

PR #613, opened 2020-06-11
Added duplicateOffscreen directive

zinabat:
## Context
[US115498](https://rally1.rallydev.com/#/detail/userstory/383982961632?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#LongLoadDirective)

Originally named `LongLoadDirective`, `duplicateOffscreen` takes a content node and duplicates it offscreen, invisible to screen readers. The duplicated item sits just after the original content node in the DOM.

This directive will be used to assist with copying elements that take a long time to load, such as course images. It is one of the requirements for Drag N Drop lists.

Screen Shot 2020-06-11 at 2 48 23 PM

## Quality
- the `directives` folder has been added to the karma configuration file
- a unit test that tests some basics for the directive was added
- Manual browser testing with VoiceOver
  - [x] Chrome
  - [x] Legacy Edge 
  - [ ] CrEdge
  - [x] Safari
  - [x] Firefox


directives/duplicate-offscreen.js
```diff
@@ -0,0 +1,78 @@
...
+	style.type = 'text/css';
+	style.appendChild(document.createTextNode(offscreenStyles));
+
+	// workaround for polyfill issue. ShadowRoot isn't a real thing in polyfill land,
+	// so attach the style to the caller instead.
+	try {
+		shadow.prepend(style);
+	} catch (e) {
+		containerPart.options.eventContext.prepend(style);
+	}
```
zinabat, line 63: This is a gross "python-ic" way to do this workaround. I would love a suggestion to make this cleaner.

directives/duplicate-offscreen.js
```diff
@@ -0,0 +1,76 @@
...
+
+	// set the copy
+	copy.setValue(value);
+	copy.commit();
+
+	// nextSibling is how we access the actual node we need
+	// This is hardly intuitive and requires knowledge of Part's inner workings
+	// https://github.com/Polymer/lit-html/issues/388
+
+	const copyNode = copy.startNode.nextSibling;
```
m6jones, line 41: This is to skip over the comment the defines the part?

directives/demo/duplicate-demo.js
```diff
@@ -0,0 +1,14 @@
...
+
+class DuplicateDemo extends LitElement {
+	render() {
+		const item = html`<div>This message is present in the DOM twice</div>`;
+
+		return html`
+			${duplicateOffscreen(item)}
+		`;
+	}
+}
```
dbatiste, line 12: Being a duplicate, I would imagine that this would be valid...?
```
render() {
		const item = html`<div>This message is present in the DOM twice</div>`;

		return html`
			item
			${duplicateOffscreen(item)}
		`;
	}
```

dbatiste: ... or does it insert it into the DOM on-screen, and duplicate it for offscreen?

directives/demo/duplicate-demo.js
```diff
@@ -0,0 +1,14 @@
...
+import { duplicateOffscreen } from '../../../directives/duplicate-offscreen.js';
+
+class DuplicateDemo extends LitElement {
+	render() {
+		const item = html`<div>This message is present in the DOM twice</div>`;
+
+		return html`
+			${duplicateOffscreen(item)}
+		`;
+	}
```
dbatiste, line 11: I have a couple thoughts here.

My first impression is that I would use this directive like this, specifying the item, and using the directive to create an offscreen version of it:
```
render() {
    const item = html`<div>This message is present in the DOM twice</div>`;
    return html`
        ${item}
        ${duplicateOffscreen(item)}
    `;
}
```
But it looks like this directive would actually create two parts.  One that is onscreen and one that is offscreen.  Perhaps this is just a naming issue.

While the directives are super cool and quite powerful, I wonder if a directive is really needed to accomplish this.  That is, could the same result be accomplished without a directive:
```
render() {
    const item = html`<div>some content</div>`;
    return html`
        ${item}
        <div class="d2l-offscreen">${item}<div>
    `;
}
```

From what I understand - the goal is to render a copy of the list-item to be used for dragging, and if there is slow loading content in the item (like an image), we want that loaded as soon as the user begins to drag.  We don't really want to do a lot of upfront work duplicating item content, especially if the user isn't going to drag any items.  It would be very interesting to explore options for deferring this work, whilst still having the copy ready for dragging if/when needed.

Besides that, I'm also interested in seeing the bigger picture of D&D - previously we implemented D&D in a very generic way in MVC to allow it to be used by several different components, but that was before browsers natively supported it, so there was considerable value in doing so.  To what extent would we get value our of generalizing D&D now, and should this node copying stuff just get layered in with that.

directives/duplicate-offscreen.js
dbatiste: What's the error that gets thrown?  Could we avoid this by checking.
```
if (shadow) {
    ...
} else {
    ...
}
```
Also, do you know what `eventContext` is?  The API docs for `NodePart` aren't too informative.  i.e. could the above be accomplished by using `getRootNode()`?  An alternative approach is also to use inline styles and avoid insertion of a stylesheet.

directives/duplicate-offscreen.js
zinabat: Yes, it's a little bit weird and a known problem for readability in directives.

directives/demo/duplicate-demo.js
m6jones: The problem that this directive solves is that some elements that will be the contents of a list item can take forever to load. The example I ran into is course images on my prototype.

In the prototype I just copied the list and made a second one. This caused some weird scrolling issues? That is probably just how the prototype works. But for this I wanted something better. Something that could handle individual parts that need time to load. And say, if we don't have any then there no need for any duplication.

So basically, this directive is what people can use if they have those longer loading elements. They wrap it in this directive and our new in time copy method can sub in the elements that have already loaded.

So the point of this is to have just in time copy for everything unless the user of the list knows something will take longer to load. In that case they can use the directive to handle that case.

So basically, how someone would probably code this up is:
- Writes out there list and enables drag n drop.
- When they test it they notice the image on their card doesn't show up while dragging and becomes a grey box.
- They check out the read me and find that they just need to wrap their image in this directive and now it works.

We can also write in the readme that requiring to use this mixin probably indicates that cache isn't working since images have to be refetched each time they load.

I am using images since that is where I ran into the problem. However, this could happen with other components too and this is a way to standardize the copy of the element for us to use.

directives/duplicate-offscreen.js
m6jones: The comments define where parts begin and end.

directives/duplicate-offscreen.js
```diff
@@ -0,0 +1,76 @@
...
+	copyNode.setAttribute('data-duplicate', '');
```
m6jones, line 42: Nit: You can use `toggleAttribute`

Could you export `data-duplicate` as a const so when we write our copy method we can use the exported constant as the attribute.

directives/demo/duplicate-demo.js
dbatiste: If caching is working correctly, is this a non-issue?

Assuming it is an issue that needs to be dealt with... I would tend to prefer a sol'n that the list-item consumer doesn't need to opt into.  I'm not convinced that consumers will always know ahead of time which elements are slow loading (consider user-authored content), and this may also depend significantly on their network conditions.  The app developer might not notice anything slow, but under different conditions (ex. crappy mobile) they might, so I don't think it's safe to rely on the developers noticing, or realizing it's something they need to address.

Currently I don't think this is JIT... i.e. it renders as soon as the directive runs - is that true?   To me, just-in-time would load the image and render the copy just before it's needed, and only if it's needed.

Is it possible to elaborate more on how the offscreen duplicates would get used?  From what I gather, if only certain elements are duplicated offscreen (i.e. not the entire list-item) as indicated by use of the directive, then there must be some other code that will query for these elements and then move them into the copy for dragging.

What other use cases do we have besides images?  Possibly videos (which have thumbnails).  We would not be making duplicate HM requests, correct?  User authored content could contain lots of different things that are not known in advance.

Is the non-cached scenario an edge case that we can deal with simply by showing loading spinner?  If not, my current thinking is that we should aim for something more automatic - possibly creating the list item copies automatically and deferring the work using requestIdleCallback and/or IntersectionObserver.

m6jones: Can I setup a demo over zoom to show you the problem this solves that way you might be able to think of a better solution? I think that will be the easiest thing to do.

zinabat:
An alternative solution was found by @m6jones here: https://codepen.io/m6jones/pen/LYGRxzw?editors=0010

Issue #249, opened 2019-10-09
[Localization] Arabic translation for dialog


maboelsoudd2l:
the translation for https://github.com/BrightspaceUI/core/blob/master/components/dialog/dialog.js#L62
isn't the best. Can we change it?

dbatiste:
We certainly can. I copied this from elsewhere in our system, so I think there's likely more than one spot that we'd want to update - I'm not sure where the original translation came from.

dlockhart:
@maboelsoudd2l just circling back on this issue -- do you have a recommendation on what this translation should be instead? We'll need to update this code but also the corresponding translation in the LMS.

dlockhart:
I'm going to close this due to staleness. @maboelsoudd2l if you think this is still an issue, please re-open with a recommendation on what to use instead. Thanks!

PR #618, opened 2020-06-15
Provide a more palatable translation for "close this dialog"

maboelsoudd2l:
The translator seems to have understood the statement "Close this dialog" as "Close this conversation". Then the translated it to "Close this conversation right here". It sounded weird. So I just put the nearest translation I can think of and just put "Close", because there isn't really a one-to-one translation for "dialog box".

should fix the issue here: https://github.com/BrightspaceUI/core/issues/249


dbatiste:
Thanks Mohamed.

@margaree , now that Serge has been integrated, is this safe to change?  i.e. is Serge going to overwrite this fix?

svanherk:
I expect it will overwrite it, but it will most likely be another alternative.  Maybe we wait to merge this until we get the serge translations back next week, and @maboelsoudd2l you can advise which is better?

PR #622, opened 2020-06-15
Initial switch component with on/off semantics.

dbatiste:
This is a WIP - creating to get some initial feedback.  This introduces the `d2l-switch` which uses the Aria `switch` role ([spec](https://w3c.github.io/aria/#switch)) for "on/off" semantics.  It basically has the same functionality as checkbox and button toggle, but presents the widget more closely to its appearance.

Changes to follow once we get design details.  This is the current state based on a couple screenshots.

image


components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+
+	static get properties() {
+		return {
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
+			offIcon: { type: String, attribute: 'off-icon' },
+			on: { type: Boolean, reflect: true },
+			onIcon: { type: String, attribute: 'on-icon' },
+			role: { type: String, reflect: true },
+			tabindex: { type: String, reflect: true }
```
dlockhart, line 16: Should we go with `reflect: true` for everything seeing as it's our recommendation in the best practices?

components/switch/demo/switch.html
```diff
@@ -0,0 +1,54 @@
...
+				<template>
+					<d2l-switch label="Tool" label-hidden></d2l-switch>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>Visibility</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-switch label="Visibility" on-icon="tier1:visibility-show" off-icon="tier1:visibility-hide"></d2l-switch>
```
dbatiste, line 39: I think the "visibility" use-case may be fairly common.  In the current implementation it is supported by exposing `on-icon` and `off-icon` property.  However I am considering a couple of other options:

1. expose a `type="visibility"` property and bake the icons into the component for that variant; or
2. factor this logic into a mixin and provide two separate components - `d2l-switch` and `d2l-switch-visibility`

It may be necessary to change the text depending on the state... "Visible" and "Hidden", however we should be careful with this because screen reader (at least VoiceOver announces: `"{label}, {on/off}, switch"`, so given that, changing the label could make the announcement very confusing.

components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.labelHidden = false;
+		this.on = false;
+		this.onIcon = 'tier1:check';
+		this.role = 'switch';
```
dbatiste, line 103: Surface the `role` and the `tabindex` on the host so the switch can be associated with a tooltip.

components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+		});
+		this.addEventListener('keyup', (e) => {
+			if (e.keyCode === 32) this.on = !this.on;
+		});
+	}
+
+	render() {
+		const offIcon = (this.offIcon && this.offIcon.length > 0)
+			? html`<d2l-icon class="d2l-switch-icon-off" icon="${this.offIcon}"></d2l-icon>`
+			: html`<d2l-icon-custom size="tier1">
```
dbatiste, line 124: We don't have an icon for the off state in our iconset... so this defines that.

components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+
+		return html`
+			<div class="d2l-switch-container">
+				<div class="d2l-switch-inner">
+					<div class="d2l-switch-toggle"><div></div></div>
+					<d2l-icon class="d2l-switch-icon-on" icon="${this.onIcon}"></d2l-icon>
+					${offIcon}
+				</div>
+			</div>
+			${!this.labelHidden ? html`<span aria-hidden="true">${this.label}</span>` : ''}
```
dbatiste, line 138: This hides the visible text label from screen readers since we surface it on the host via `aria-label`.

components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('label')) {
+			this.setAttribute('aria-label', this.label);
+		}
+		if (changedProperties.has('on')) {
+			if (this.on) this.setAttribute('aria-checked', 'true');
+			else this.setAttribute('aria-checked', 'false');
```
dlockhart, line 149: Probably want to fire an event of some kind, maybe just a `change` event so folks can do `<d2l-switch @change="${this._handleSwitch}">`?

components/switch/switch.js
```diff
@@ -0,0 +1,155 @@
...
+			:host(:hover) .d2l-switch-container {
+				border-color: var(--d2l-color-celestine);
+				border-width: 2px;
+			}
+			d2l-icon, d2l-icon-custom {
+				height: 0.8rem;
+				width: 0.8rem;
+			}
+			d2l-icon-custom {
+				stroke: var(--d2l-color-tungsten);
```
dlockhart, line 82: Kind of sucks you had to do this... should we be making a change to `d2l-icon-custom`?

components/switch/demo/switch.html
dlockhart: Yeah I'd be in favour of exposing this as a separate component seeing as this is such a common use case.

components/switch/demo/switch.html
```diff
@@ -0,0 +1,54 @@
...
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>Visibility (with tooltip)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-switch label="Tool" label-hidden id="tool-switch"></d2l-switch>
+					<d2l-tooltip for="tool-switch">Toogle tool enabled/disabled.</d2l-tooltip>
```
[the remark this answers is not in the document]
dbatiste: Correct.  That would be an option though - move the `role` and `tabindex` into the shadowDOM, and bake a `d2l-tooltip` in for it.

components/switch/switch.js
dbatiste: Ah yep... I forgot to add that!

components/switch/demo/switch.html
dbatiste: Updated to define `d2l-switch` and `d2l-switch-visibility`.

components/switch/demo/switch.html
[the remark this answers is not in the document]
dbatiste: Moved focus target, role, etc. into local DOM.

components/switch/switch.js
dbatiste: Was able to remove this by using the variable in the `svg` for the custom icon.

components/switch/switch-visibility.js
```diff
@@ -0,0 +1,22 @@
+import '../icons/icon.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { SwitchMixin } from './switch-mixin.js';
+
+class VisibilitySwitch extends SwitchMixin(LitElement) {
+
+	constructor() {
+		super();
+		this.label = 'Visibility';
```
dbatiste, line 9: This needs to be languified - in some discussion with Design about the label.  Stay tuned.

components/switch/switch.js
dbatiste: This code is changed significantly, but `change` event has been added.

components/switch/switch-mixin.js
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	constructor() {
+		super();
+		this.disabled = false;
+		this.labelHidden = false;
+		this.on = false;
+		this._labelId = getUniqueId();
+	}
+
```
dlockhart, line 133: I'm assuming a label is required? If so, worth `console.warn`-ing in `firstUpdated` if they forget to provide one? I'd like to start using warnings more often for invalid usages like this. We could even get really aggressive and throw from `render()`...

components/switch/switch-mixin.js
```diff
@@ -0,0 +1,181 @@
...
+	_handleKeyUp(e) {
+		// space pressed... toggle state
+		if (e.keyCode === 32) this._toggleState();
+	}
+
+	_toggleState() {
+		if (this.disabled) return;
+		this.on = !this.on;
```
dlockhart, line 178: Mostly I'm curious: what should happen if folks manually change the "on" state by setting the property or attribute themselves when the component is disabled? I just tried this with checkboxes and it goes ahead with the state change, which I think is correct since if you're mucking around in JavaScript you probably know what you're doing. And that matches your current behavior, but might be worth ensuring there's a test for it.

components/switch/switch-mixin.js
```diff
@@ -0,0 +1,181 @@
...
+import { getUniqueId } from '../../helpers/uniqueId.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+export const SwitchMixin = superclass => class extends RtlMixin(superclass) {
+
+	static get properties() {
+		return {
+			disabled: { type: Boolean, reflect: true },
+			label: { type: String, reflect: true },
```
dlockhart, line 12: With the label (and labelHidden) being defined in the mixin, it means that it can be overridden even for the visibility switch. I'm fine with that, but just wanted to make sure that's what you intended.

components/switch/switch-mixin.js
```diff
@@ -0,0 +1,181 @@
...
+				cursor: pointer;
+				display: inline-block;
+				font-size: 0;
+				line-height: 0;
+				outline-style: none;
+				padding: 0.1rem;
+				vertical-align: middle;
+			}
+			.d2l-switch-container:focus,
+			.d2l-switch-container:hover {
```
dlockhart, line 41: You'll likely need a `focus()` method on the host that transfers focus inside to this element.

components/switch/switch-mixin.js
dbatiste: Thanks yep... that's the intent so far.

components/switch/switch-mixin.js
dbatiste: Thanks, indeed... added this.

components/switch/switch-mixin.js
dbatiste: Cool yep... add a warning.  A label is indeed required.

components/switch/switch-mixin.js
```diff
@@ -0,0 +1,199 @@
...
+			console.warn('Switch components require accessible text.');
+		}
+	}
+
+	render() {
+		const tabindex = (!this.disabled ? '0' : undefined);
+		const textPosition = (this.textPosition === 'start' || this.textPosition === 'hidden'
+			? this.textPosition : 'end');
+		return html`
+			${textPosition === 'start' ? html`<span id="${this._textId}" class="d2l-switch-text">${this.text}</span>` : ''}
```
dlockhart, line 148: Worth factoring this view out into a variable and just conditionally including it?

components/switch/switch-visibility.js
```diff
@@ -0,0 +1,22 @@
+import '../icons/icon.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { SwitchMixin } from './switch-mixin.js';
+
+class VisibilitySwitch extends SwitchMixin(LitElement) {
+
+	constructor() {
+		super();
+		this.text = 'Visibility';
```
dlockhart, line 9: If you want to languify this out of the gate, you could likely reuse the translations from [this term](https://search.d2l.dev/xref/lms/ipsct/D2L.IM.IPSCT/LangTerms/definitions/OnlineRooms.xml?r=d4b17a73#544).

dbatiste: Thanks.  I think there might still be some conversation with Design.  In several places they've been flipping the text along with the state of the switch, which ends up being very confusing, and is not recommended. So... I plan to localize this sometime in the next couple of days.

PR #625, opened 2020-06-16
Initial work getting custom-elements.json

margaree:
I added necessary comments to `more-less` to get the mostly correct documentation (would remove `analyze-more-less` from the package.json scripts).

A problem is that it does include `d2l-localize-behavior-language-changed` in events for any localized component. I so far have not found a way to ignore it.

Result:
```
{
  "version": "experimental",
  "tags": [
    {
      "name": "d2l-more-less",
      "path": "./components/more-less/more-less.js",
      "description": "The `d2l-more-less` element can be used to minimize the display of long content, while providing a way to reveal the full content.",
      "attributes": [
        {
          "name": "blur-color",
          "description": "The gradient color of the blurring effect.",
          "type": "\"hex color code\""
        },
        {
          "name": "h-align",
          "description": "The h-align property of the more-less button.",
          "type": "\"\"|\"text\""
        },
        {
          "name": "inactive",
          "description": "Whether the component is active or inactive.",
          "type": "boolean"
        },
        {
          "name": "expanded",
          "description": "Indicates whether element is in \"more\" state.",
          "type": "boolean",
          "default": "false"
        },
        {
          "name": "height",
          "description": "The maximum height of the content when in \"less\" state.",
          "type": "string",
          "default": "\"4em\""
        }
      ],
      "properties": [
        {
          "name": "blurColor",
          "attribute": "blur-color",
          "description": "The gradient color of the blurring effect.",
          "type": "\"hex color code\""
        },
        {
          "name": "hAlign",
          "attribute": "h-align",
          "description": "The h-align property of the more-less button.",
          "type": "\"\"|\"text\""
        },
        {
          "name": "inactive",
          "attribute": "inactive",
          "description": "Whether the component is active or inactive.",
          "type": "boolean"
        },
        {
          "name": "expanded",
          "attribute": "expanded",
          "description": "Indicates whether element is in \"more\" state.",
          "type": "boolean",
          "default": "false"
        },
        {
          "name": "height",
          "attribute": "height",
          "description": "The maximum height of the content when in \"less\" state.",
          "type": "string",
          "default": "\"4em\""
        }
      ],
      "events": [
        {
          "name": "d2l-more-less-render",
          "description": "Dispatched when the component finishes rendering"
        },
        {
          "name": "d2l-localize-behavior-language-changed"
        }
      ],
      "slots": [
        {
          "name": "",
          "description": "Default content placed inside of the component"
        }
      ]
    }
  ]
}
```


components/more-less/more-less.js
```diff
@@ -8,15 +8,41 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
...
 	static get properties() {
 		return {
-			blurColor: { type: String, attribute: 'blur-color' }, // The gradient color of the blurring effect. Must be hex color code.
-			expanded: { type: Boolean, reflect: true }, // Indicates whether element is in "more" state.
-			hAlign: { type: String, attribute: 'h-align' }, // The h-align property of the more-less button.
-			height: { type: String }, // The maximum height of the content when in "less" state.
-			inactive: { type: Boolean, reflect: true }, // Whether the component is active or inactive.
+			/**
+			 * The gradient color of the blurring effect.
+			 * @type {"hex color code"}
```
dlockhart, line 22: Probably want to use `string` here so that it's valid JSDoc, but maybe we can use the [typedef](https://jsdoc.app/tags-typedef.html) to define a custom HEX-string?

components/more-less/more-less.js
```diff
@@ -8,15 +8,41 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
...
+
+			/**
+			 * The maximum height of the content when in "less" state.
+			 */
+			height: { type: String },
+
+			/**
+			 * Whether the component is active or inactive.
+			 */
+			inactive: { type: Boolean, reflect: true },
```
dlockhart, line 45: It picked up the `false` default for `expanded` but not for `inactive` -- likely because we're not setting a default for this in the constructor like we should be.

components/more-less/more-less.js
```diff
@@ -8,15 +8,41 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
...
+			blurColor: { type: String, attribute: 'blur-color' },
+
+			/**
+			 * Indicates whether element is in "more" state.
+			 */
+			expanded: { type: Boolean, reflect: true },
+
+			/**
+			 * The h-align property of the more-less button.
+			 * @type {""|"text"}
```
dlockhart, line 33: Same here... should likely be a string but maybe we can define a complex type.

package.json
```diff
@@ -63,6 +65,7 @@
     "node-sass": "^4",
     "puppeteer": "^3",
     "sinon": "^9",
+    "web-component-analyzer": "^1.0.3",
```
dlockhart, line 68: Nit: can just have `^1`

components/more-less/more-less.js
margaree: So declaring it as a typedef makes sense (possibly more for the h-align case). It doesn't seem to affect the `custom-elements.json` file other than if I use `@type {HexString}` then that's noted as the type, so I don't know if that will potentially cause confusion. Where would be the best place to put the `@typedef`? Does right above `@type` make sense? I initially put them at the top of the file but that looks a bit cluttered.

dlockhart: I'm not sure... I doubt there'd be a way to put it somewhere centralized such that we could share it across all the things that need HEX strings. We could also just go with `{string}` for now and defer figuring out complex types until later. :)

margaree:
I've added comments as needed to the a and b components. I'll do the others in separate PR(s) so this one isn't too massive.

components/backdrop/backdrop.js
```diff
@@ -8,13 +8,27 @@ const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 const scrollKeys = [];
 let scrollOverflow = null;
 
+/**
+ * The `d2l-backdrop` element is a web component to display a semi-transparent backdrop behind a specified sibling element. It also hides elements other than the target from assistive technologies by applying `role="presentation"` and `aria-hidden="true"`.
+ */
 class Backdrop extends LitElement {
 
 	static get properties() {
```
margaree, line 16: This file uses an attribute `slow-transition` in the css (https://github.com/BrightspaceUI/core/blob/e5a10aaaeaeb37ca4c07038481e281b2a39e5c65/components/backdrop/backdrop.js#L49) that's called out in the readme but isn't listed in properties so it doesn't show up in the custom-elements.json. Should it get added to properties?

package.json
```diff
@@ -7,14 +7,15 @@
     "access": "public"
   },
   "scripts": {
+    "analyze": "wca analyze components/**/*.js --format json --outFile custom-elements.json",
```
dlockhart, line 10: You may want to include `templates` as well to pick up our page templates... there's just one right now and it's not really finished though.

PR #632, opened 2020-06-18
d2l-link: fix smalll

margaree:


svanherk:
> Wow. The first actual bug caused by my broken keyboard. :(

Time for a new laptop!

PR #635, opened 2020-06-18
ci: only building master and release branches in CI

dlockhart:
This should speed our CI builds up a bit. With this change, Travis CI won't run on branches other than release/maintenance branches named things like `1.x`, `1.53.x`, etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: node_js
 node_js: node
+branches:
+  only:
+  - master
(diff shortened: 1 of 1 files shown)


dbatiste:
Don't we need CI to run on other branches - for instance when updating goldens?

dlockhart:
CI will still run on pull requests, and the "regenerate goldens" button triggers a custom build with [its own `.travis.yml` config](https://github.com/BrightspaceUI/visual-difference-bot/blob/master/index.js#L316) anyway.

PR #636, opened 2020-06-19
ci: only including font addons in visual diff stage

dlockhart:
These are pretty slow to install, so doing it in all 3 stages shouldn't be necessary. I'm not sure if this is actually going to work since there isn't a lot of documentation on this. 🤞


dlockhart:
It did yeah! And seems to have shaved 1-1.5 minutes off the first and third stages. ⏩

PR #643, opened 2020-06-22
Added NewPositionEventDetails

zinabat:
## Context
[US115496](https://rally1.rallydev.com/#/detail/userstory/383979103344?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#newPositionEventDetails)

The new `NewPositionEventDetails` class is an event class with helper methods included. Currently, the event is not hooked up to anything - @m6jones 's work will be in tandem. The new class will replace the current `detail` in the 'd2l-list-item-position' event.

Comments will be reduced or eliminated when the drag n drop feature is merged into `master`.

### Event Properties
- `targetKey`: The key of the node that is moving
- `destinationKey`: The key of the item in the position the node is moving to
- `temporaryMovement`: Whether the movement is entering or exiting a temporary movement (for use with keyboard mode - currently not used)

### Event Methods
- `announceMove`: Announces the movement with a given array and a callback function
- `fetchPosition`: Fetches the position of the node given an array and a key
- `reorder`: Reorders a given array in place. Announces the movement if a callback function is given

## Quality
- Added unit tests to ensure reordering works

### DIFF
--- a/components/list/list-item-drag-mixin.js
+++ b/components/list/list-item-drag-mixin.js
@@ -107,23 +107,30 @@ export class NewPositionEventDetails {
 	 * @param { String } object.destinationKey The item key of the list-item in the position we are moving to
 	 * @param { String } object.temporaryMovement Information on whether the item is entering or exiting temporary movement
 	 */
-	constructor(object) {
-		this.targetKey = object.targetKey;
-		this.destinationKey = object.destinationKey;
-		this.temporaryMovement = object.temporaryMovement;
+	constructor({targetKey, destinationKey, temporaryMovement}) {
+		if (!targetKey || !destinationKey) {
+			throw new Error(`NewPositionEventDetails must have a targetKey and destinationKey\nGiven: ${targetKey} and ${destinationKey}`);
+		}
+		this.targetKey = targetKey;
+		this.destinationKey = destinationKey;
+		this.temporaryMovement = temporaryMovement;
 	}
 
 	/**
 	 * Announces a move to screen readers on an array
 	 * @param { Array<Node> } list The array to announce the move on.
-	 * @param { function(any, Number): String } announceCallback Callback function that returns the announcement text
+	 * @param { Object } obj Object containing callback functions
+	 * @param { function(any, Number): String } obj.announceFn Callback function that returns the announcement text
 	 *        Takes the item in the array and the index as arguments. Should return the text to announce.
+	 * @param { function(Node): String } obj.keyFn Callback function that returns a key given a listitem
 	 */
-	announceMove(list, announceCallback) {
-		const index = this.fetchPosition(list, this.targetKey);
-		if (index === null) throw new Error('Item to announce move not found in given array.');
-		const item = list[index];
-		const message = announceCallback(item, index);
+	announceMove(list, {announceFn, keyFn}) {
+		const targetIndex = this.fetchPosition(list, this.targetKey, keyFn);
+		const destinationIndex = this.fetchPosition(list, this.destinationKey, keyFn);
+		if (targetIndex === null) throw new Error(`Target "${this.targetKey}" not found in array`);
+		if (destinationIndex === null) throw new Error(`Destination "${this.destinationKey}" not found in array`);
+
+		const message = announceFn(list[targetIndex], destinationIndex);
 		if (message) announce(message);
 	}
 
@@ -131,9 +138,10 @@ export class NewPositionEventDetails {
 	 * Fetches an index position within a list given an item key
 	 * @param { Array<Node> } list Array of nodes
 	 * @param { String } key Item key to fetch position of
+	 * @param { function(Node): String } keyFn Function that returns the key from a list item
 	 */
-	fetchPosition(list, key) {
-		const index = list.findIndex(x => x.key === key);
+	fetchPosition(list, key, keyFn) {
+		const index = list.findIndex(x => keyFn(x) === key);
 		return index === -1 ? null : index;
 	}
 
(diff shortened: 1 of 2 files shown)


components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+
+export class NewPositionEventDetails {
+	/**
+	 * @param { Object } object An simple object with the position event properties
+	 * @param { String } object.targetKey The item key of the list-item that is moving
+	 * @param { String } object.destinationKey The item key of the list-item in the position we are moving to
+	 * @param { String } object.temporaryMovement Information on whether the item is entering or exiting temporary movement
+	 */
+	constructor(object) {
+		this.targetKey = object.targetKey;
```
AKobets, line 111: at which point validation of object properties should happen? How it is guaranteed that the constructor will get valid object at a runtime?

components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+	 *        This optional callback will be passed to announceMove if given.
+	 */
+	reorder(list, announceCallback = null) {
+		if (this.destinationKey === undefined || this.destinationKey === this.targetKey) return;
+
+		const origin = this.fetchPosition(list, this.targetKey);
+		const destination = this.fetchPosition(list, this.destinationKey);
+
+		if (origin === null || destination === null) {
+			throw new Error('Reorder failed. Position does not exist in list.');
```
AKobets, line 156: should error message be  more informative for debugging purpose ?

components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		}
+
+		// move the item in the list to a new position in place
+		const item = list[origin];
+		// now that we have a reference to the item, shove everything between the
+		// destination to the origin over one
+		if (origin > destination) {
+			// shove to the right ♪┗ ( ･o･) ┓♪
```
AKobets, line 164: what this `♪┗ ( ･o･) ┓♪` need to do in the comment ?

components/list/test/list-item-drag-mixin.test.js
```diff
@@ -19,3 +19,94 @@ describe('ListItemDragMixin', () => {
 		expect(element.draggable).to.be.false;
 	});
 });
+
+describe('NewPositionEventDetails', () => {
+	describe('reorder', () => {
+		const tests = [
+			{
+				description: 'does nothing when new and old key are the same',
```
AKobets, line 27: nit:  better `target and destination` key are the same

components/list/test/list-item-drag-mixin.test.js
```diff
@@ -19,3 +19,94 @@ describe('ListItemDragMixin', () => {
...
+				});
+				const objects = test.input.array.map(x => ({key : x }));
+				event.reorder(objects);
+				expect(objects.map(x => x.key)).to.deep.equal(test.expected);
+			});
+		}
+
+		it('throws an error when a position does not exist', () => {
+			const event = new NewPositionEventDetails({
+				targetKey: 'bar',
```
AKobets, line 105: should be 2 more tests to cover branching: when object with targetKey does not exists, when object with destinationKey  does not exist

components/list/test/list-item-drag-mixin.test.js
```diff
@@ -19,3 +19,94 @@ describe('ListItemDragMixin', () => {
 		expect(element.draggable).to.be.false;
 	});
 });
+
```
AKobets, line 22: missing tests for `announcement callbacks executed`

components/list/list-item-drag-mixin.js
zinabat: fiiiine

components/list/list-item-drag-mixin.js
zinabat: I can add validation to the constructor; this is the responsibility of the functions dispatching the event within the mixin.

components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
 };
+
+export class NewPositionEventDetails {
+	/**
+	 * @param { Object } object An simple object with the position event properties
+	 * @param { String } object.targetKey The item key of the list-item that is moving
+	 * @param { String } object.destinationKey The item key of the list-item in the position we are moving to
+	 * @param { String } object.temporaryMovement Information on whether the item is entering or exiting temporary movement
+	 */
+	constructor(object) {
```
m6jones, line 110: Can you do `constructor({targetKey, destinationKey, temporaryMovement})` doing the cool es6 stuff that we see in our services.

components/list/list-item-drag-mixin.js
m6jones: Not sure if this object will know if the key is valid or can know.

components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		const message = announceCallback(item, index);
+		if (message) announce(message);
+	}
+
+	/**
+	 * Fetches an index position within a list given an item key
+	 * @param { Array<Node> } list Array of nodes
+	 * @param { String } key Item key to fetch position of
+	 */
+	fetchPosition(list, key) {
```
m6jones, line 135: Let key be a callback so a user can define their own key. Not all arrays will have `.key` property. We might want to use the `href` or `self` as a key.

components/list/list-item-drag-mixin.js
```diff
@@ -98,3 +99,84 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+	/**
+	 * Reorders an array in place with the current event information
+	 * The item will be moved to the position of the destinationKey. Array elements shift
+	 * forward one to make room.
+	 * @param { Array<Node> } list The array to reorder
+	 * @param { function(any, Number): String } [announceCallback] Callback function that returns the announcement text
+	 *        Takes the item in the array and the index as arguments. Should return the text to announce.
+	 *        This optional callback will be passed to announceMove if given.
+	 */
+	reorder(list, announceCallback = null) {
```
m6jones, line 149: It be cool if this had the `key callback` as well. You could make the second parameter an object => `reorder(list, {announceCallBack, keyfn})` Define similar to https://lit-html.polymer-project.org/guide/template-reference#repeat

components/list/list-item-drag-mixin.js
m6jones: It is required or the code would break.

components/list/list-item-drag-mixin.js
m6jones, line 149: An actual real life interview question.

components/list/list-item-drag-mixin.js
AKobets: could key property name be passed via constructor ?

zinabat: Callback works better I think - it's consistent with what `lit-html` does and is more versatile than a property name

m6jones: The list is constructing and not the user. So it won't know about where the key is either.

PR #644, opened 2020-06-22
Add tooltip to SwitchMixin.

dbatiste:


dbatiste:
Yeah I was playing with that and I noticed a small pixel difference for the vertical placement of the tooltip in the test page.  Very strange.

PR #646, opened 2020-06-23
Jsdoc comments part 4

margaree:
list, loading-spinner, menu

FYI @m6jones @zinabat for the comments added in list components. Let me know if any of them need changing.


components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -11,9 +11,18 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 
 	static get properties() {
 		return {
-			shown: { type: Boolean, reflect: true },
+			/**
+			 * @ignore
+			 */
```
dlockhart, line 16: Oh neat!

dbatiste: Does jsdoc support single line comments?  One of my gripes with comments is basically how bloaty they can be.  Course being consistent in any given block of properties is probably preferred.  Anyway, I'll get used to whatever we decide.

margaree: It does seem to as long as they're formatted like `/** Whether to extend the separators beyond the content's edge */` and not `// Whether...` or `/* ... */`. 

I'm good either way. Most of these property description comments are one line so it could be pretty consistent and less bloaty, though if we do at some point get more into the typing then we would want to bring in the blocks in some cases.

margaree: I'll merge this PR and then do a follow up PR for consistency and tweaks and can try out the one line comment blocks

PR #648, opened 2020-06-23
jsdoc comments part 5

margaree:
- meter, offscreen, status-indicator, tabs, tooltip, primary secondary page template
- I cleaned up some properties from meter that were also included in the mixin

### DIFF
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -4,14 +4,10 @@ import { bodyStandardStyles } from '../typography/styles.js';
 import { MeterMixin } from './meter-mixin.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
+/**
+ * A circular progress indicator.
+ */
 class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
-	static get properties() {
-		return {
-			value: { type: Number },
-			max: { type: Number }
-		};
-	}
-
 	static get styles() {
 		return [ bodyStandardStyles, css`
 		:host {
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -76,22 +76,72 @@ const computeTooltipShift = (centerDelta, spaceLeft, spaceRight) => {
 	}
 };
 
+/**
+ * A component used to display additional information when users focus or hover on a point of interest.
+ * @slot - Default content placed inside of the tooltip
+ * @fires d2l-tooltip-show - Dispatched when the tooltip is opened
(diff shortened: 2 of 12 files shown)


components/tooltip/tooltip.js
```diff
@@ -76,22 +76,72 @@ const computeTooltipShift = (centerDelta, spaceLeft, spaceRight) => {
...
+ */
 class Tooltip extends RtlMixin(LitElement) {
 
 	static get properties() {
 		return {
-			align: { type: String }, /* Valid values are: 'start' and 'end' */
+			/**
+			 * Align the tooltip with either the start or end of its target.
+			 * If not set, the tooltip will attempt be centered.
+			 * Valid values: "start", "end"
```
margaree, line 92: I started putting valid values in a standard format on a separate line since we aren't using types yet, and this makes it easy to see and parse in the documentation site

components/meter/meter-circle.js
```diff
@@ -4,14 +4,10 @@ import { bodyStandardStyles } from '../typography/styles.js';
...
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
+/**
+ * A circular progress indicator.
+ */
 class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
-	static get properties() {
-		return {
-			value: { type: Number },
-			max: { type: Number }
```
dlockhart, line 11: Good catch.

components/tooltip/tooltip.js
dlockhart: Guessing there isn't a JSDoc way to do this eh. [TSX seems to support it](https://github.com/Shopify/polaris-react/blob/master/src/components/Button/Button.tsx#L50) as part of the property definition. 

Wonder if [this could work](https://stackoverflow.com/questions/19093935/how-to-document-a-string-type-in-jsdoc-with-limited-possible-values)?

margaree: So trying the enum kind of works but it doesn't actually print what the enum options are, just the enum name. It looks like using `@type {('start'|'end')}` is supported but the docs have not yet been updated so we could just use that.

dlockhart: Yeah was hoping the `@type` thing could work... as long as we can grab it somehow to render a `<select>` in the interactive demo thing / property list.

margaree: Cool yeah, it ends up as `"type": "('info'|'error')"` in custom-elements.json which seems parsable

PR #649, opened 2020-06-23
Feature/add breadcrumbs

miadabrin:
Port d2l-breadcrumbs to Core


components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+		`;
+	}
+
+	constructor() {
+		super();
+		this.href = '#';
+		this.compact = false;
+		this.text = '';
+		this.target = '';
+		this.ariaLabel = '';
```
AllanKerr, line 80: I would recommend not initializing `ariaLabel` and [using `ifDefined`](https://github.com/BrightspaceUI/core/blob/69dda0d1ca5d1a9d048f1fe7601cb4e376318a71/components/button/button-subtle.js#L159)

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.href = '#';
+		this.compact = false;
+		this.text = '';
+		this.target = '';
```
AllanKerr, line 79: I would also recommend not initializing and using `ifDefined` for `target` since its optional.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
+import '../icons/icon.js';
+import '../link/link.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class BreadCrumb extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			compact: {
```
AllanKerr, line 10: It might be worth making this private. Its technically accessed by `d2l-breadcrumbs` so its not truly private but we also don't want others setting it directly. Not sure if anyone else has any thoughts here.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				align-items: center;
+				display: inline-flex;
+			}
+
+			div.d2l-breadcrumb-wrapper {
```
AllanKerr, line 41: `.d2l-breadcrumb-wrapper` doesn't seem to be used anywhere else so `div` can be removed to simplify these rules.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+			}
+
+			d2l-icon {
+				height: 8px;
+				padding-left: 10px;
+				padding-right: 5px;
+				width: 8px;
+			}
+
+			d2l-icon:dir(rtl) {
```
AllanKerr, line 57: This should be `:host([dir="rtl"]) d2l-icon { ... }`

miadabrin: @AllanKerr I think the icon gets rtled as well. but what you suggested makes more sense

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+
+	static get styles() {
+		return css`
+			:host {
+				align-items: center;
+				display: inline-flex;
+			}
+
+			div.d2l-breadcrumb-wrapper {
+				align-items: center;
```
AllanKerr, line 42: I think you should be able to move `align-items` into the rule with `display: flex`. Without `dispaly: flex` `align-items` doesn't do anything.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,93 @@
...
+				padding-left: 5px;
+				padding-right: 10px;
+			}
+
+			:host([compact]) div.d2l-breadcrumb-wrapper d2l-icon {
+				padding-right: 10px;
+				padding-left: 0;
+			}
+
+			:host([compact]) div.d2l-breadcrumb-wrapper d2l-icon:dir(rtl) {
```
AllanKerr, line 67: This should also be using `:host([dir="rtl"])`, you can simply the rule to `:host([dir="rtl"][compact]) d2l-icon { ... }`

miadabrin: @AllanKerr this particular one doesn't see to work on edge doing it this way. for some reason `div.d2l-breadcrumb-wrapper` should be added as well or the scoping doesn't work correctly

components/breadcrumbs/breadcrumb.js
dlockhart: I was going to suggest the same -- `_compact`.

miadabrin: @AllanKerr @dlockhart considering it should be an attribute as well. Should I name the attribute as `compact` still?

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			:host(:dir(rtl))::after {
+				background: -webkit-gradient(linear, right top, left top, from(rgba(255, 255, 255, 0)), to(var(--d2l-breadcrumbs-background-color)));
+				background: -webkit-linear-gradient(right, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
+				background: -o-linear-gradient(right, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
+				background: linear-gradient(to left, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
+				left: 0;
+				right: auto;
+			}
```
AllanKerr, line 53: In the future you could write this like this so you don't end up duplicating the rule:
```
:host-context([dir="rtl"])::after,
:host(:dir(rtl))::after {
    ...
}
```

That being said, we should be okay with just single `:host([dir="rtl"])::after { ... }` rule.

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
dbatiste: Yep - `host-context` is not something we can rely on (lots of history behind it, and may never be supported as needed)

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
AllanKerr: If we're making it private I would recommend adding it to the `d2l-breadcrumb-wrapper` with `?data-compact=${this.compact}` and using `.d2l-breadcrumb-wrapper[data-compact]` instead. That way we don't need to reflect it to the host anymore.

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: The rule works without `d2l-breadcrumb-wrapper` . The class prefixing is only needed for `Slotted` rules it seems

miadabrin: ✔️ Done

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,116 @@
...
+		});
+		return breadCrumbs;
+	}
+
+	_handleSlotChange() {
+		this._setChildrenCompactProp();
+	}
+
+	_setChildrenCompactProp() {
```
dlockhart, line 98: @dbatiste @AllanKerr and I chatted a bit about what to recommend here. We think moving this to the child `d2l-breadcrumb` is a better approach for now -- you're correct that it won't get updates to the parent's attribute, but we think that's worth the trade-off to avoid walking the assigned nodes in the shadow root.

There are 2 ways you could accomplish this from the child:
1. Use `findComposedAncestor` (from our DOM helpers in core) in `connectedCallback` -- this is what the old Polymer version was doing
2. Fire an event from the child in `connectedCallback` and then have the parent attach a value for `composed` to the event detail. Since events are synchronous, the child can then access that value on the next line.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,89 @@
+import '../icons/icon.js';
+import '../link/link.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
```
dlockhart, line 6: @margaree has been going through and adding JSDoc comments to the classes and properties in core. We're going to use these to build up a documentation site for everything automatically. Would you be able to add this in based on some examples from the other components in core?

components/breadcrumbs/breadcrumb.js
dbatiste: Yeah I think this is an awkward limitation of shadyCSS/shadyDOM which comes into play for slotted selectors + legacy Edge.

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,116 @@
...
+				--d2l-breadcrumbs-background-color: rgb(251, 252, 252);
+				display: block;
+				position: relative;
+			}
+
+			:host::after {
+				background: -webkit-gradient(linear, left top, right top, from(rgba(255, 255, 255, 0)), to(var(--d2l-breadcrumbs-background-color)));
+				background: -webkit-linear-gradient(left, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
+				background: -o-linear-gradient(left, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
+				background: linear-gradient(to right, rgba(255, 255, 255, 0), var(--d2l-breadcrumbs-background-color));
```
dlockhart, line 27: You can remove the vendor prefixes for `linear-gradient`. All the browsers we care about support the standard one:

IE v10, released September 2012
Firefox v16, released August 2012
Chrome v26, released February 2013
Safari v6.1, released October 2013
Opera 12.1, released November 2012

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,116 @@
...
+				reflect: true
+			}
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				--d2l-breadcrumbs-background-color: rgb(251, 252, 252);
+				display: block;
```
dlockhart, line 19: If you set a `display` value for components, also provide CSS for the `hidden` attribute. [More information](https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-style-the-display-value-of-the-host)

components/breadcrumbs/breadcrumbs.js
miadabrin: @dlockhart which one do we like more ?

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,116 @@
...
+				type: Boolean,
+				reflect: true
+			}
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				--d2l-breadcrumbs-background-color: rgb(251, 252, 252);
```
dlockhart, line 18: I couldn't find a reference to this variable being overridden anywhere through code search. Can we just remove it and hardcode the colour wherever it's used? Just not something we need to support forever if it doesn't need to be extensible.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,89 @@
...
+			:host([dir="rtl"][compact]) d2l-icon {
+				padding-right: 0;
+				padding-left: 10px;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.href = '#';
```
dlockhart, line 74: I don't love setting the default value here to `#`... I know that's what it was before but it makes me really sad. I'm guessing this was being done so that the link would get the hand cursor and a tab stop even when it didn't have an `href` set. Curious if that's actually a valid use case...

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,116 @@
...
+				right: auto;
+			}
+
+			.d2l-breadcrumbs-wrapper {
+				overflow: hidden;
+				position: relative;
+				white-space: nowrap;
+			}
+
+			:host([compact]) .d2l-breadcrumbs-wrapper ::slotted(d2l-breadcrumb:not(:last-of-type)) {
```
AllanKerr, line 52: You can simplify this rule by removing `.d2l-breadcrumbs-wrapper`

index.html
```diff
@@ -92,6 +92,7 @@ <h2 class="d2l-heading-3">Components</h2>
 		<li><a href="components/switch/demo/switch.html">d2l-switch</a></li>
 		<li><a href="components/tabs/demo/tabs.html">d2l-tabs</a></li>
 		<li><a href="components/tooltip/demo/tooltip.html">d2l-tooltip</a></li>
+		<li><a href="components/breadcrumbs/demo/breadcrumbs.html">d2l-breadcrumbs</a></li>
```
dlockhart, line 95: Can you put this higher so that it's alphabetical?

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,89 @@
+import '../icons/icon.js';
+import '../link/link.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class BreadCrumb extends RtlMixin(LitElement) {
```
AllanKerr, line 7: For consistency I think the the `C` should be lowercase: `class Breadcrumb ...`

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,89 @@
...
+	constructor() {
+		super();
+		this.href = '#';
+		this.compact = false;
+		this.text = '';
+	}
+
+	render() {
+		return html`
+			<div class="d2l-breadcrumb-wrapper" role="navigation" aria-label="${this.ariaLabel ? this.ariaLabel : this.text}">
```
dlockhart: I know this was just copied from the Polymer element, but this might be a good time to align the accessibility with [ARIA best practices](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html).

They recommend that the `navigation` role actually be the parent -- we could turn `<div class="d2l-breadcrumbs-wrapper">` into a `<nav>`. That navigation element is the one that needs a label, but it seems like they recommend just using "Breadcrumb" as a generic label always, so that lang term could just live inside this component.

They further recommend using an unordered list and markup up the last non-link breadcrumb with `aria-current="page"`. I don't _think_ these breadcrumbs have that concept?

All of this actually aligns with how we do breadcrumbs in the monolith. Here's the markup they generate:
```html
<nav aria-labelledby="d2l_1_31_683">
  <h2 class="d2l-heading vui-offscreen" id="d2l_1_31_683">Breadcrumb:</h2>
  <ol aria-labelledby="d2l_1_31_683">
    <li><a class="d2l-link" href="/d2l/link1">Link 1</a></li>
    <li><a class="d2l-link" href="/d2l/link2">Link 2</a></li>
    <li><span class="d2l-breadcrumb-text">Page</span></li>
  </ol>
</nav>
```

We can also proceed as-is and Gaudi can look into making these ARIA changes afterwards.

components/breadcrumbs/demo/breadcrumbs.html
```diff
@@ -0,0 +1,63 @@
...
+			color: var(--d2l-color-tungsten);
+			font-size: 0.7rem;
+		}
+		.align-item.bullet::before {
+			color: var(--d2l-color-tungsten);
+			content: '\002022';
+			margin-left: 5px;
+			margin-right: 5px;
+		}
+	</style>
```
AllanKerr, line 25: It looks like the styles in this block are unused so you should be good to remove it.

components/breadcrumbs/breadcrumb.js
miadabrin: Sure. Sounds great

components/breadcrumbs/breadcrumb.js
AllanKerr: It looks like there are at least some places that use it with `click` event handlers without an `href` so I don't think we can remove it as default unless we also clean those places up.

dlockhart: Right. So we can either keep it, or add `tabindex="0"` and `cursor: pointer` when `href` is missing.

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Removed ✔️

components/breadcrumbs/test/breadcrumb.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+		it('should reflect "_compact" property to attribute', async() => {
+			elem._compact = true;
+			await elem.updateComplete;
+			expect(elem.hasAttribute('compact')).to.be.true;
+		});
+	});
+
+	describe('constructor', () => {
+		it('should construct', () => {
```
AllanKerr, line 47: Can you add a constructor test for `d2l-breadcrumbs` too?

index.html
miadabrin: Done ✔️

components/breadcrumbs/test/breadcrumb.test.js
miadabrin: Sure. will do

components/breadcrumbs/demo/breadcrumbs.html
miadabrin: Removed ✔️

components/breadcrumbs/test/breadcrumb.test.js
miadabrin: @AllanKerr it already exists under `describe('basic attribute tests'` . Should I moved it out?

AllanKerr: Sorry, I completely missed that! 😳 Normally I'd put it in a higher up `describe` block but thats super nit-picky so your call.

components/breadcrumbs/demo/breadcrumbs.html
```diff
@@ -0,0 +1,63 @@
...
+
+<head>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+	<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '../../demo/demo-page.js';
+		import '../breadcrumbs.js';
```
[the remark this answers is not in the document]
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: I will bring it back for now.

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: Actually I didn't remove it in the first place . So I guess: Done ✔️  for now

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,90 @@
...
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class Breadcrumb extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			_compact: {
+				attribute: 'compact',
+				type: Boolean,
+				reflect: true
```
AllanKerr, line 14: You might be doing this with the `compact` pass-through changes you're making but just to make sure it doesn't get missed. This doesn't need to be reflected anymore now that its added to the `d2l-breadcrumb-wrapper`.

components/breadcrumbs/test/breadcrumb.test.js
miadabrin: yeah let me move that out then.

components/breadcrumbs/test/breadcrumbs.test.js
```diff
@@ -0,0 +1,161 @@
...
+				<d2l-breadcrumbs>
+					<d2l-breadcrumb href="#" text="Basic Item 1"></d2l-breadcrumb>
+					<d2l-breadcrumb href="#" text="Basic Item 2"></d2l-breadcrumb>
+					<d2l-breadcrumb href="#" text="Basic Item 3" aria-label="Aria for Item 3"></d2l-breadcrumb>
+				</d2l-breadcrumbs>
+			`);
+			items = list.querySelectorAll('d2l-breadcrumb:not([hidden])');
+		});
+
+		it('should display three items', () => {
```
dbatiste, line 100: I find the visual-diff tests address this, and much more... making many such tests redundant.  I think even some of the property reflection is captured by the visual-diff test.  In some cases, the property reflection is simply a means to achieve some visual presentation, which is really what we're interested in.  Some of these tests were typical before we had visual-diff testing as a tool available to us.

components/breadcrumbs/test/breadcrumb.test.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: Yeah makes sense

components/breadcrumbs/test/breadcrumbs.test.js
miadabrin: @dbatiste do visual-diff tests also compare against multiple browsers? I think it already caught some of the issues with edge which I wouldn't have caught on my own

dbatiste: Currently we have it configured to cover Chromium browsers.  So, we can make a call on what's interesting to test cross-browser.  Many of the unit tests from previous Polymer components don't survive the migration.

miadabrin: So do you suggest removing the these tests for now? or should it be done at a later stage when we are more confident with cross-browser visual diffs?

dlockhart: Yeah I agree with Dave here... I'd probably remove everything except the constructor tests and maybe the default value ones. Basically if someone later mucked around with the code, we want to make sure that we catch things that the visual diff tests will not, but without going overboard.

dbatiste: Re cross-browser visual-diff: I think it's possible to launch experimental FF with Puppeteer.  It would be a big deal though, as it would significantly increase our build times, and undoubtedly introduce some build instability.  I think there is some mention of having Puppeteer drive headless Safari too, but to my knowledge this is merely an idea.

miadabrin: Removed ✔️

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: @AllanKerr @dlockhart @dbatiste So I went with the event processing way. with just minor different with what Dave suggested. Having the event being called in constructor doesn't seem to work properly as the constructors don't happen in any specific order so the event gets missed sometimes. I moved it to firstUpdated instead

dbatiste: Did you try `connectedCallback`, or did that cause some trouble too?

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	constructor() {
+		super();
+		this.href = '#';
+		this._compact = false;
+		this.text = '';
+	}
+
+	firstUpdated() {
```
dlockhart, line 79: So it won't work from the constructor as it's not attached to the DOM, but what about `connectedCallback`? `firstUpdated` is OK, but this means it'll render twice.

components/breadcrumbs/breadcrumbs.js
miadabrin: yeah both have the same issue

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('d2l-breadcrumb-connected', this._handleBreadCrumbConnected, true);
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.removeEventListener('d2l-breadcrumb-connected', this._handleBreadCrumbConnected, true);
```
dlockhart, line 67: You could avoid all this wire-up by putting the handler directly here:
```html
<div class="d2l-breadcrumbs-wrapper" @on-d2l-breadcrumb-connected="${this. _handleBreadCrumbConnected}">
```

It'll then handle the add/remove of the listener.

components/breadcrumbs/breadcrumbs.js
dlockhart: In that case, I think Dave's original solution of using `findComposedAncestor` would be preferable, since it avoids the double-render of using `firstUpdated`. Haha sorry...

miadabrin: Actually it might be the issue with console.log not being able to track the changes properly (at least the way I do it). It seems to be working fine in practice

miadabrin: I moved it to connectedCallback . (Also I don't mind doing it the other way as well)

components/breadcrumbs/breadcrumb.js
miadabrin: `connectedCallback` seems to be working fine. I was testing it bad I guess with console.log

components/breadcrumbs/breadcrumbs.js
miadabrin: Nice. Yeah I totally forgot that existed ✔️ done

components/breadcrumbs/breadcrumb.js
miadabrin: It does work. I was just using console.log to track changes which did not work properly. (Sorry)

components/breadcrumbs/breadcrumb.js
miadabrin: Added comments based on what I saw on other components. Please let me know if something should be added/removed

components/breadcrumbs/breadcrumb.js
miadabrin: I think I managed to make the parent to be `<nav>` and the slot to be rendered inside and `<ol>`
so this would be the parent:
```
<nav aria-label="Breadcrumb" class="d2l-breadcrumbs-wrapper" @d2l-breadcrumb-connected="${this._handleBreadCrumbConnected}">
				<ol>
				<slot></slot>
				</ol>
			</nav>
```
these extra styles added to parent as well:
```
.d2l-breadcrumbs-wrapper ol {
				margin: 0;
				padding-left: 0;
				list-style: none;
			}
```
and then the child would look like this:
```
<li class="d2l-breadcrumb-wrapper" ?data-compact=${this._compact} role="navigation">
				<d2l-link aria-label="${this.ariaLabel ? this.ariaLabel : this.text}" href="${this.href}" target="${ifDefined(this.target)}">${this.text}</d2l-link>
				${this._compact ? html`<d2l-icon icon="d2l-tier1:chevron-left"></d2l-icon>` : html`<d2l-icon icon="d2l-tier1:chevron-right"></d2l-icon>`}
			</li>
```
but it doesn't cover the fact that we should set `aria-current="page"` unless the user of the component sets it properly. Thoughts?

miadabrin: Also the separators are added with css in the original suggestion but we have them as icons. Do we want to change that?

dlockhart: Cool! Don't worry about the `aria-current="page"` for now. Based on what [design has specified](http://design.d2l/components/breadcrumbs/), they don't want it:
> Do not include the current page title in the breadcrumb as a non-clickable link – repetition is repetitive

For the list part, I'd want to test it out in a screen reader, but I think having `<ol><d2l-breadcrumb><li>` in the resolved DOM is going to trip them up -- list items are supposed to be direct children of the list itself, which doesn't play nice across the shadow boundaries. There are a couple ways I think we could solve that:
1. `<ol><d2l-breadcrumb role="presentation"><li>` will tell the screen reader to ignore that element in terms of semantics.
2. `<div role="list"><d2l-breadcrumb role="listitem">` applies the list/listitem roles to existing elements

In both cases, our `d2l-breadcrumb` element is "sprouting" an attribute, which isn't amazing but I don't really see an alternative. @dbatiste not sure if you have thoughts there.

Haha and we thought breadcrumbs would be an easy conversion... now you're learning all about accessibility and ARIA! 🎉

dbatiste: No other options to offer up here, other than dropping the list rendering altogether, which does not align with Aria example.  In either case above it would require sprouting, so I would be inclined to sprout `role="listitem"`, but perhaps still try to use the `ol`.  While not ideal, we've had to sprout role in a few other places.  The consumer could mess with it, but if they do (unlikely), they are doing bad things.

miadabrin: I went with the first option (setting a presentation role for breadcrumb in the middle per Dave's suggestion)

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * A component to help users understand where they are within the LMS. would be used as a child of d2l-breadcrumbs component (in slot)
+ * @fires d2l-breadcrumb-connected - Dispatched when the component is initialized to notify parent (so parent can set the compact attribute in event detail)
+ */
+class Breadcrumb extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			_compact: {
+				type: Boolean,
```
dlockhart, line 16: Can you also set `attribute: false` so that this won't sprout an attribute and so that setting an attribute won't set this value.

components/breadcrumbs/breadcrumb.js
```diff
@@ -0,0 +1,134 @@
...
+				detail: { }
+			}
+		);
+		this.dispatchEvent(event);
+		this._compact = !!event.detail.compact;
+	}
+
+	render() {
+		return html`
+			<li class="d2l-breadcrumb-wrapper" ?data-compact=${this._compact} role="navigation">
```
dlockhart, line 127: You can remove `role="navigation"` from here since the parent has the `<nav>` element now.

components/breadcrumbs/breadcrumb.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,88 @@
...
+				overflow: hidden;
+				position: relative;
+				white-space: nowrap;
+			}
+
+			.d2l-breadcrumbs-wrapper ol {
+				margin: 0;
+				padding-left: 0;
+				padding-right: 0;
+				list-style: none;
```
dlockhart, line 59:
```suggestion
				list-style: none;
				margin: 0;
				padding: 0;
```

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,88 @@
...
+				width: 10px;
+			}
+
+			:host([dir="rtl"])::after {
+				background: linear-gradient(to left, rgba(255, 255, 255, 0), rgb(251, 252, 252));
+				left: 0;
+				right: auto;
+			}
+
+			.d2l-breadcrumbs-wrapper {
```
dlockhart, line 49: The CSS class isn't really needed now, this could just be `nav {`

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,88 @@
...
+				overflow: hidden;
+				position: relative;
+				white-space: nowrap;
+			}
+
+			.d2l-breadcrumbs-wrapper ol {
```
dlockhart, line 55: Could just be `ol {`

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,88 @@
+import './breadcrumb.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+/**
+ * A component to help users understand where they are within the LMS. would be used the parent of d2l-breadcrumb and will render them in a slot
+ */
+class BreadCrumbs extends RtlMixin(LitElement) {
```
dlockhart, line 8: Class should just be `Breadcrumbs` to match casing.

components/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	constructor() {
+		super();
+		this.compact = false;
+	}
+
+	render() {
+		return html`
+			<nav aria-label="Breadcrumb" class="d2l-breadcrumbs-wrapper" @d2l-breadcrumb-connected="${this._handleBreadCrumbConnected}">
```
dlockhart, line 75: This label is going to need to be localized... also wondering whether it should be configurable. We could start with just localizing it and later if folks needs to set custom labels we could provide a custom label.

Have a look at some of the other components (e.g. more/less) for examples of using the `LocalizeMixin`. I think we can just copy [the existing translations](https://search.d2l.dev/xref/lms/lp/_lang/LangTerms/definitions/Framework.xml?r=174eecc7#272) from the MVC version of this component.

components/breadcrumbs/breadcrumb.js
miadabrin: Removed ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumbs.js
miadabrin: Done ✔️

components/breadcrumbs/breadcrumb.js
dbatiste: Is that preferred?  How does it announce screenreader-wise?

miadabrin: @dbatiste I tested with chromevox and it doesn't appear to read out the breadcrumbs part. I guess it doesn't work as expected?

components/breadcrumbs/breadcrumbs.js
miadabrin: @dlockhart I think the existing translations are for singular mode but the breadcrumbs is plural. Is there anyone I can contact to fix that for me? (or use google translate?).

miadabrin: also on another note: Do we prefer to use the `LocalizeCoreElement` instead of `LocalizeMixin` and centralize translations or do we like it to be inside the component itself? asking since it seems that we have used `LocalizeCoreElement` everywhere instead of `LocalizeMixin`

margaree: We've been centralizing them in the `lang/` files and using `LocalizeCoreElement`.

components/breadcrumbs/breadcrumb.js
miadabrin: I tried the other way as well. The screen reader doesn't seem to read it this way either:

```
			<div class="d2l-breadcrumbs-wrapper" role="list" aria-label="Breadcrumbs" @d2l-breadcrumb-connected="${this._handleBreadCrumbConnected}">
				<slot></slot>
			</div>
```

```
.d2l-breadcrumbs-wrapper {
				list-style: none;
				margin: 0;
				padding: 0;
				overflow: hidden;
				position: relative;
				white-space: nowrap;
			}
```

```
<div class="d2l-breadcrumb-wrapper" ?data-compact=${this._compact} role="navigation">
				<d2l-link aria-label="${this.ariaLabel ? this.ariaLabel : this.text}" href="${this.href}" target="${ifDefined(this.target)}">${this.text}</d2l-link>
				${this._compact ? html`<d2l-icon icon="d2l-tier1:chevron-left"></d2l-icon>` : html`<d2l-icon icon="d2l-tier1:chevron-right"></d2l-icon>`}
			</div>
```

components/breadcrumbs/breadcrumbs.js
dlockhart: The existing translations should work -- they look and behave essentially the same as this. Despite sometimes pluralizing it as "breadcrumbs" you could also can it a "breadcrumb" in reference to the collection, i.e. "breadcrumb trail".

dlockhart: I think you could remove the colon from the existing translations though if that's easy.

components/breadcrumbs/breadcrumb.js
dlockhart: The child `div` there needs `role="listitem"` not `navigation`.

components/breadcrumbs/breadcrumbs.js
miadabrin: yeah it should be easy. Also there is some minor differences between the langterms from what we have. e.g. there is a `de-de` langterm in LMS which I think would be `de` here. how should I organize that part?

margaree: Yup, you would just use the `de-de` term in `de`. The mappings are here https://github.com/BrightspaceUI/core/blob/master/core.serge.json for which language-locale combo ends up in each file. That should cover most cases.

miadabrin: Done ✔️

lang/nl.js
```diff
@@ -2,6 +2,7 @@
 
 export default {
 	"components.alert.close": "Waarschuwing sluiten",
+	"components.breadcrumbs.breadcrumb": "Breadcrumb",
```
dlockhart, line 5: Oh weird I guess we don't have a Dutch translation?

components/breadcrumbs/breadcrumb.js
m6jones: So this isn't great for accessibility. I don't think the default should be `#` when you have an @click. In discover, we use @click since it is an fra and we use single page navigation but our href still equals the actual url that will take you to that page. Prevent default is great for these situations!  

This is what we should make people do for accessibility! Making this default in core says, this is something that should be done when it isn't.

lang/nl.js
miadabrin: Yeah it seems like it. I can do it myself though . not sure if that's a good idea though

components/breadcrumbs/breadcrumb.js
miadabrin: @mthjones so just to confirm. removing the default is safe by itself? or should I add the properties Dave mentioned? `(tabindex ="0")` and `cursor: "pointer"`

dlockhart: It's not safe by itself, since Allan found some places that are _only_ setting `onclick`, which if we remove this those links will no longer get tab stops.

I think the proper solution here would be to first fix those places that are relying on this and then I agree with Maya that we should remove the default and include some documentation about how to properly provide a `href` AND `click` handler and avoid navigation using `preventDefault`.

I'm still worried that there'll be cases where there literally is no navigation to do. This is why Labs exists -- so that we could make breaking changes like this without impacting places that use the non-core thing... but that ship has sailed for this component.

dlockhart: But it's a lot to ask of @miadabrin to go through all those places and fix them up... so I'm not sure how to proceed here.

lang/nl.js
dlockhart: Nah it's OK -- once this merges the Serg.io automatic translation stuff should kick in and it'll get translated. 👍

components/breadcrumbs/breadcrumb.js
miadabrin: So Dave took care of it himself . So ✔️

components/breadcrumbs/breadcrumb.js
AllanKerr: I only found the one case and stopped looking so it might be worth seeing how many occurrences there actually are. To me the two path forwards seem to be:
1. Documenting it as required even though there are technically some places that don't set it
2. Cleaning up places that don't set it and actually making it required but that could prove to be difficult because we'd have to be able to find an appropriate URL to set which may not be obvious.

The former doesn't seem unreasonable to me.

miadabrin: So this should be the full list of places that don't set it:

https://search.d2l.dev/xref/GitHub/Brightspace/publish-fra/src/publish-activity-log.js?r=ecd4c66b#134
https://search.d2l.dev/xref/GitHub/Brightspace/publish-fra/src/publish-packages.js?r=a87e11c0#398
https://search.d2l.dev/xref/GitHub/Brightspace/smart-curriculum/src/react/ui/ContentPanel/CreateMaterials.jsx?r=394264f6#212
https://search.d2l.dev/xref/GitHub/Brightspace/smart-curriculum/src/react/ui/ContentPanel/AddMaterials.jsx?r=0ed0dffb#336

If we can figure out what to set for the href (it can just "#" for now I guess) for these places I think we can safely remove the default here (but documenting it like Allan suggested)

dlockhart: Cool yeah I like the idea of setting it to `#` but then in the PR asking for help to determine what the `href` should really be. Ensures we're not breaking anything, but let's us move forward.

miadabrin: PRs added :
https://github.com/Brightspace/publish-fra/pull/140
https://github.com/Brightspace/smart-curriculum/pull/1055

currently contacting code owners to review it

miadabrin: closed: https://github.com/Brightspace/smart-curriculum/pull/1055
(because fork doesn't run the CI)
added: https://github.com/Brightspace/smart-curriculum/pull/1056

m6jones: This is awesome! Thanks for figuring this out :)

miadabrin: The default is now removed ✔️

PR #650, opened 2020-06-24
jsdoc comments: tweaks for consistency

margaree:
- remove periods at ends of small sentences
- change required fields to have "REQUIRED:" at start
- use `@type` and `@default` where applicable
- simplify component descriptions where possible

For quotation marks, I used `"` when defining a `@default` since that's what the analyzer does with defaults it finds and I wanted to keep that consistent. For defining `@type` I used `'` because when displayed in the documentation table they look quite bulky with the double quote. However, this does take away some consistency.

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -11,7 +11,8 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 	static get properties() {
 		return {
 			/**
-			 * Optionally align dropdown to either `start` or `end`. If not set, the dropdown will attempt be centred.
+			 * Optionally align dropdown to either start or end. If not set, the dropdown will attempt be centred.
+			 * @type {('start'|'end')}
 			 */
 			align: {
 				type: String,
--- a/components/inputs/input-time.js
+++ b/components/inputs/input-time.js
@@ -89,6 +89,7 @@ class InputTime extends LitElement {
 		return {
 			/**
 			 * Set default value of input. Valid values are times in ISO 8601 time format ("hh:mm:ss"), "startOfDay", "endOfDay".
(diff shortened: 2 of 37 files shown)


components/dropdown/dropdown-content-mixin.js
```diff
@@ -11,7 +11,8 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 	static get properties() {
 		return {
 			/**
-			 * Optionally align dropdown to either `start` or `end`. If not set, the dropdown will attempt be centred.
+			 * Optionally align dropdown to either start or end. If not set, the dropdown will attempt be centred.
+			 * @type {('start'|'end')}
```
dlockhart, line 15: This is a weird attribute eh... the default is `undefined` which centres it? Hmm... 🤔

components/inputs/input-time.js
```diff
@@ -112,7 +113,8 @@ class InputTime extends LitElement {
 			 */
 			maxHeight: { type: Number, attribute: 'max-height' },
 			/**
-			 * Number of minutes between times shown in dropdown. Valid values are "five", "ten", "fifteen", "twenty", "thirty", "sixty".
+			 * Number of minutes between times shown in dropdown
+			 * @type {('five'|'ten'|'fifteen'|'twenty'|'thirty'|'sixty')}
```
dlockhart, line 117: We don't need to call out the `@default` here because lit-analyzer gets it from the constructor, ya?

margaree: Yup that's correct.

components/dropdown/dropdown-content-mixin.js
margaree: Yeah - we could add a 'centered' and do a bit of restructuring

PR #651, opened 2020-06-24
provider-mixin readme example: firstUpdate

johngwilkinson:
The provider-mixin wasn't working for me, until I realized calling reqestInstance in the constructor wasn't hitting the event handler, and moved it to firstUpdated().

This just adds that note and changes the sample code accordingly.

### DIFF
--- a/mixins/provider-mixin.md
+++ b/mixins/provider-mixin.md
@@ -23,7 +23,7 @@ class InterestingFactProvider extends ProviderMixin(LitElement) {
 Once this has been set up, child components can request your provider's data via the `RequesterMixin` mixin, using the same key as the provider. Since the event that is generated to request a provider is bubbled until it is fulfilled, this means there can be an arbitrary number of components in the hierarchy between the provider and the requester, none of which have any knowledge of the data being requested or provided.
 
 NB: before the first render (i.e. in the constructor) the event handler won't be attached to the DOM, so in practice `this.requestInstance()` needs
-to be called in `firstUpdated()`, which will cause a second render.
+to be called in `connectedCallback()` (or `firstUpdated()`).
 
 ```js
 import { RequesterMixin } from '@brightspace-ui/core/mixins/provider-mixin.js'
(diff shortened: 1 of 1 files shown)


mixins/provider-mixin.md
```diff
@@ -22,6 +22,9 @@ class InterestingFactProvider extends ProviderMixin(LitElement) {
 
 Once this has been set up, child components can request your provider's data via the `RequesterMixin` mixin, using the same key as the provider. Since the event that is generated to request a provider is bubbled until it is fulfilled, this means there can be an arbitrary number of components in the hierarchy between the provider and the requester, none of which have any knowledge of the data being requested or provided.
 
+NB: before the first render (i.e. in the constructor) the event handler won't be attached to the DOM, so in practice `this.requestInstance()` needs
```
dbatiste, line 25: It sounds like the requestor could call `requestInstance` in the `connectedCallback` if it needs the dependency before first render.  Care ought to be taken however, because if the element is removed from the DOM and reconnected (ex. drag & drop), `connectedCallback` would be invoke again.

mixins/provider-mixin.md
```diff
@@ -53,7 +53,7 @@ class InterestingFactUI extends RequesterMixin(LitElement) {
 		`;
 	}
 	
-	firstUpdated() {
+	connectedCallback() {
```
dbatiste, line 56: I think the first line in here should be `super.connectedCallback();`

mixins/provider-mixin.md
dlockhart: That should be OK though... `requestInstance` doesn't actually add any event listeners, it just fires an event. So the only downside here would be that it would be called again and get a "fresh" value for whatever it was requesting.

mixins/provider-mixin.md
```diff
@@ -22,6 +21,8 @@ class InterestingFactProvider extends ProviderMixin(LitElement) {
 
 Once this has been set up, child components can request your provider's data via the `RequesterMixin` mixin, using the same key as the provider. Since the event that is generated to request a provider is bubbled until it is fulfilled, this means there can be an arbitrary number of components in the hierarchy between the provider and the requester, none of which have any knowledge of the data being requested or provided.
 
+NB: due to its reliance on DOM events, `requestInstance()` needs to be called after the element has been attached to the DOM, such as in `connectedCallback()`.
```
svanherk, line 24: Side question... what does NB mean?

PR #652, opened 2020-06-25
Fix for tab not working on list item sample

zinabat:
@AKobets found a bug that prevents users from tabbing to and from items if grid is turned on. There's also some weirdness with NVDA, probably because the events are still propagating.


components/list/list-item-generic-layout.js
```diff
@@ -374,8 +375,13 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				// focus five rows down
 				this._focusNextRow(false, 5);
 				break;
+			default:
+				preventDefault = false;
```
AKobets, line 379: would it be possible to add test for TAB key event, to test the `preventDefault = false` flow ?

zinabat: good idea

zinabat: So, I've added a spy to make sure that preventDefault isn't called for TAB, but there's no good way to hook into the `focusin` event that's fired, as it's handled by the browser and doesn't fire in headless mode. Spying seems like enough for me.

components/list/test/list-item-generic-layout.test.js
```diff
@@ -326,12 +330,23 @@ describe('d2l-list-item-generic-layout', () => {
...
-				setTimeout(() => dispatchKeyEvent(layout, test.key.code, test.key.ctrl));
+				let event = null;
+				setTimeout(() => event = dispatchKeyEvent(layout, test.key));
 				await test.event();
 				expect(test.activeElement()).to.equal(test.expected());
+				expect(event.preventDefault).to.have.been.calledOnce;
 			});
 		}
 
+		it('does not preventDefault when Tab is pressed', async() => {
```
AKobets, line 341: when I was testing manually, bug symptoms where such: pressing tab does not go move focus outside the grid.
If you can change `longFixture` to have grid and then button, assertions could be made on document.activeElement
In case grid should capture event - the outside of the grid button is not active. For tab key it should be active.
Then you don't need a spy and can do a test in a black-box style.
Does it make sense ?

zinabat: Resolved via zoom

AKobets: as Zina explained earlier. focus will not change properly in headless mode and the suggested test changes will not work

PR #654, opened 2020-06-26
demo snippet: add public functionality for forcing the code to update

margaree:
The scenario is for live demos - have the code update when the component demoing changes.

`slotchange` doesn't seem to work for this and I wasn't sure how to force it to, but overall this seems to work.

### DIFF
--- a/components/demo/code-view.js
+++ b/components/demo/code-view.js
@@ -51,6 +51,10 @@ class CodeView extends LitElement {
 		`;
 	}
 
+	forceUpdate() {
+		this._updateCode(this.shadowRoot.querySelector('slot'));
+	}
+
 	get _codeTemplate() {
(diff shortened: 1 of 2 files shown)


dbatiste:
Do we know why the slot change event didn't work?  Just curious of the new use case.

I have some other idea on things I'd like to change with these components to make them editable.  Not sure if that overlaps with anything envisioned for the docs stuff.

margaree:
I think it's related to `the slotchange event doesn't fire if the children of a slotted node change` from https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event since I'm just modifying attributes and not deleting, though I didn't dig in too much. I don't have any other changes to make to these components so definitely go ahead.

PR #656, opened 2020-06-29
input-date-range: Initial work

margaree:
- Does not include validation yet
- I added `label-hidden` to `d2l-input-fieldset` as it looks like we wont always want to use fieldset to show the label
- lang changes were automatically done to move the `components.input-date-time.*` terms to be alphabetized according to how the verify-translations script alphabetized them and there should not be any actual changes to those terms.

To do:
- [x] readme


components/inputs/input-date-range.js
```diff
@@ -0,0 +1,133 @@
...
+		if (!this.label) {
+			console.warn('d2l-input-date-range component requires label text');
+		}
+	}
+
+	render() {
+		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-date-range.startDate');
+		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-date-range.endDate');
+		return html`
+			<d2l-input-fieldset label="${ifDefined(this.label)}" ?label-hidden=${this.labelHidden}>
```
dlockhart, line 93: Nit: wrap `label-hidden` value in quotes for consistency.

components/inputs/input-date-range.js
```diff
@@ -0,0 +1,133 @@
+
```
dlockhart, line 1: Nit: leading newline

PR #663, opened 2020-07-01
Fix positioning defect with body as offsetParent

AllanKerr:
# Changes
This is necessary because browsers include margin in `offsetTop` and `offsetLeft` when the `offsetParent` is the `body`. To avoid positioning incorrectly in this case we need to set `parentTop` and `parentLeft` to `0` so any `margin` will be included in the calculation. Only using the `offsetParent`'s bounding client rect excludes `margin`.


AllanKerr:
Didn't make any changes, just force-pushed to edit the commit message to be prefixed with `fix` for a patch release.

PR #653, opened 2020-06-25
Feature: Update list item to use list-item-generic-layout

zinabat:
## Context
[US115503](https://rally1.rallydev.com/#/detail/userstory/384004440540?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#d2l-list-item)
[UX Designs](http://design.d2l/components/lists/)

## Changes
* `d2l-list-item` now uses the `d2l-list-item-generic-layout` component to generate the dom
* New hover and focus states added to styles for selectable list items
* Properties for focus state added to capture all `focusin` events within the item
* New selected style

**Note: Colours are not well represented in gif**
Kapture 2020-06-25 at 16 06 11

**Actual Colours**
Screen Shot 2020-06-25 at 4 07 55 PM

### Coming Soon
- There is new spacing for this UX that will be in a future PR
- Mixin integration
- Checkbox action is broken; this is expected before mixin integration 
- The grid will not work yet, as it requires mixin integration

## Quality
- Visual diff tests should pass except for selected, hover, and focus states
- Layout and spacing is identical to master

PR #665, opened 2020-07-02
Draft Drag and Drop Functionality Mixin

m6jones:


components/list/list-item-checkbox-mixin.js
```diff
@@ -27,6 +28,9 @@ export const ListItemCheckboxMixin = superclass => class extends superclass {
 				cursor: default;
 			}
 		` ];
+
+		super.styles && styles.unshift(super.styles);
```
zinabat, line 32: smart

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		dropSpots.clear();
 	}
+	dropSpots = null;
+	return dropSpotTemp;
+}
 
-	_dragStopHandler() {
-		this._dispatchDragEvent(null);
-		this.moveafter(null);
+class keepTrackOfThoseDropSpots {
```
zinabat, line 297: Classes should always be capitalized - this can also probably be simplified to `DropSpotsState` since it is in effect a state object

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+
+class NewPositionEventDetails {
 	/**
 	 * @param { Object } object An simple object with the position event properties
 	 * @param { String } object.targetKey The item key of the list-item that is moving
 	 * @param { String } object.destinationKey The item key of the list-item in the position we are moving to
 	 * @param { String } object.temporaryMovement Information on whether the item is entering or exiting temporary movement
 	 */
-	constructor({targetKey, destinationKey, temporaryMovement}) {
+	constructor({targetKey, destinationKey, moveBeforeDestination, temporaryMovement}) {
```
zinabat, line 384: Can we add a comment to the function to specify this is a boolean parameter? It wasn't obvious to me at first

components/list/list-item-drag-mixin.js
m6jones: Yay, thank you I really hated the name :D

components/list/list-item-drag-mixin.js
m6jones: Yes, also I am not a big fan of this name either. Any suggestions on something better?

components/list/demo/list-drag-n-drop-sample.js
```diff
@@ -0,0 +1,78 @@
...
+					</d2l-list-item-sample>
+				`)}
+			</d2l-list>
+		`;
+	}
+
+	_moveItems(e) {
+		e.detail.reorder(this.list, { keyFn: (item) => item.key });
+		this.requestUpdate('list', []);
+	}
```
m6jones, line 75: This is how people will use the drag n drop list.

components/list/demo/list-item-sample.js
```diff
@@ -1,28 +1,30 @@
...
+import { ListItemDragMixin } from '../list-item-drag-mixin.js';
 import { ListItemMixin } from '../list-item-mixin.js';
-//import { ListItemDragMixin } from '../list-item-drag-mixin.js';
 import { nothing } from 'lit-html';
 
-class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
-	// TODO: Breakpoints to live elsewhere
+class ListItemSample extends ListItemMixin(ListItemDragMixin(ListItemCheckboxMixin(LitElement))) {
+	// TODO: Role and breakpoints to live elsewhere
 	// TODO: draggable should be part of the ListItemDragMixin and removed here
```
m6jones, line 13: These can be removed

components/list/demo/list-item-sample.js
```diff
@@ -52,34 +80,37 @@ class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
...
-					<slot></slot>
-				</div>
-
-				<div slot="actions">
-					<slot name="actions"></slot>
-				</div>
-			</d2l-list-item-generic-layout>
+			${this._renderTopPlacementMarker(html`<d2l-list-item-placement-marker></d2l-list-item-placement-marker>`)}
+			${this._renderDropArea()}
+			<div class="d2l-list-item-drag-image">
```
m6jones, line 85: This class is used in the mixin to determine what the drag image should be based off of.

components/list/demo/list-item-sample.js
```diff
@@ -52,34 +80,37 @@ class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
...
+				<d2l-list-item-generic-layout ?grid-active="${this.role === 'rowgroup'}">
+					${this._renderDragHandle((dragHandle) => html`<div slot="outside-control">${dragHandle}</div>`)}
+					${this._renderDraggableArea((dragArea) => html`<div slot="outside-control-action">${dragArea}</div>`)}
+					${this.selectable ? html`
+					<div slot="control">${ this._renderCheckbox() }</div>
+					<div slot="control-action">${ this._renderCheckboxAction(null, this._contentId) }</div>
+					` : nothing }
```
m6jones, line 92: There is a difference in pattern here. `${this._renderDraggableArea((dragArea) => html`<div slot="outside-control-action">${dragArea}</div>`)}` is more complicated but because it is more powerful and the drag n drop mixin takes use of this power. We probably will want to switch the checkbox mixin to do the same.

components/list/demo/list-item-sample.js
```diff
@@ -52,34 +80,37 @@ class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
...
+					<div slot="content"
+						id="${this._contentId}"
+						class="d2l-list-item-content ${ classMap(classes) }">
+						<slot name="illustration"></slot>
+						<slot></slot>
+					</div>
+					<div slot="actions">
+						<slot name="actions"></slot>
+					</div>
```
m6jones, line 110: This was added so I could add images to the demo. It was copied from Zina's work https://github.com/BrightspaceUI/core/pull/653

components/list/list-item-drag-handle.js
```diff
@@ -107,6 +107,11 @@ class ListItemDragHandle extends LitElement {
 		if (changedProperties.has('_keyboardActive') && typeof changedProperties.get('_keyboardActive') !== 'undefined') this.focus();
 	}
 
+	activateKeyboardMode() {
+		this._dispatchAction(dragActions.active);
+		this._keyboardActive = true;
+	}
```
m6jones, line 113: This allows the drag n drop mixin to trigger keyboard mode on a larger area.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
 import { getUniqueId } from '../../helpers/uniqueId.js';
-import { html } from 'lit-element/lit-element.js';
 import { nothing } from 'lit-html';
 
+const move = Object.freeze({
+	above: true,
+	below: false
+});
+
+const timeDelayForLeavingTheDropArea = 1000; //ms
```
m6jones, line 12: This is to give drag n drop more usability. Basically, if you go past the drop point ie top of the list it will still work for 1 second. I found this to be a better solution then trying to increase the drop target size.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
 import { announce } from '../../helpers/announce.js';
+import { dragActions } from './list-item-drag-handle.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
-import { html } from 'lit-element/lit-element.js';
 import { nothing } from 'lit-html';
 
+const move = Object.freeze({
+	above: true,
+	below: false
+});
```
m6jones, line 10: Makes some of the code more readible.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
+
+const timeDelayForLeavingTheDropArea = 1000; //ms
+
 export const ListItemDragMixin = superclass => class extends superclass {
 
 	static get properties() {
 		return {
 			draggable: { type: Boolean, reflect: true },
-			key: { type: String, reflect: true }
+			dragText: { type: String, attribute: 'drag-text' },
```
m6jones, line 19: If someone drags a course to the URL bar or a notepad. This is what will be dropped. Gives some neat functionality with no cost so why not.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
+			key: { type: String, reflect: true },
+			beingDragged: { type: Boolean, reflect: true, attribute: 'being-dragged' },
```
m6jones, line 21: Allows the component using this mixin to style the item left behind during the drag.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
+			_bottomPlacementMarker: { type: Boolean },
+			_dropTarget: { type: Boolean },
```
m6jones, line 23: This triggers whether or not to show the special drop area that gives better usability.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,77 @@
...
+			}
+		` ];
+
+		super.styles && styles.unshift(super.styles);
+		return styles;
+	}
+
 	constructor() {
 		super();
 		this._itemDragId = getUniqueId();
```
m6jones, line 73: I never use this and needs to be cleaned up.

components/list/list-item-drag-mixin.js
```diff
@@ -24,6 +81,11 @@ export const ListItemDragMixin = superclass => class extends superclass {
 		}
 	}
 
+	firstUpdated(changedProperties) {
+		this.addEventListener('dragenter', this._hostDragEnter);
```
m6jones, line 85: This makes the entire host a the ability to do things when someone first enters it. This is used to show the better drop targets that block other pointer events. So thats why they can only show up when something is being dragged

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
 		}
-		this._dispatchDragEvent(itemKey);
+		this._moveItem(itemKey, this.key, false);
 	}
 
-	moveBefore(itemKey) {
+	moveItemBeforeMe(itemKey) {
 		if (itemKey === undefined) {
 			return;
 		}
```
m6jones, line 108: Goal is to remove these methods.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+
+	_dragAreaClick(e) {
+		this.shadowRoot.querySelector(`#${this._itemDragId}`).activateKeyboardMode();
+		e.preventDefault();
 	}
 
-	// TODO
-	_copy() {
+	_dragBottomEnter(e) {
+		e.dataTransfer.dropEffect = 'move';
```
m6jones, line 119: You will see this a lot. This tells the browser that we are doing a move drop effect. On a mac this does nothing but it is suppose to change the cursor and user indication. Although it does nothing, it might in the future? It is part of the standards and I wanted to stick with it.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		const dropSpots = dropSpotsFactory();
+		dropSpots.setDestination(this, move.below);
```
m6jones, line 121: See comments I say about the dropSpots where I define the class.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		this._inBottomArea = true;
```
m6jones, line 122: When someone is moving from the bottom area to the main area this is what triggers if the item should be moved above the element. Giving a better user experience. Pictures and explaining is done here: https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#Drag-Areas-for-a-List-Item

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+	}
 
+	_dragEnd(e) {
+		const dropSpot = dropSpotsBlowUp();
+		this.beingDragged = false;
+		this.tilt = false;
+		if (dropSpot.shouldDrop(e.timeStamp)) {
+			this._moveItem(dropSpot.targetKey, dropSpot.destinationKey, dropSpot.moveBeforeDestination);
+		}
```
m6jones, line 131: Instead of moving the item during the drop event I move it here. This is because it gives that 1 second area around the drop areas for a better user experience.

I also, using the timestamps produced by the events for a consistent differences in time.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
 
-	_dispatchDragEvent(detail) {
+	_dragHandleActions(e) {
+		let destinationKey;
+		switch (e.detail.action) {
+			case dragActions.active:
+				this._keyboardMode = true;
+				break;
+			case dragActions.cancel:
+			case dragActions.save:
```
m6jones, line 141: This will be where the temp movement is used. Basically, in keyboard mode I want the user to be able to press escape and have things reset. So the idea is that it will send up the event saying save the temp movement or go back. It is expect that the user component handles the different states and decides if they want to use this functionality.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+				this._keyboardMode = false;
+				break;
+			case dragActions.up:
+				destinationKey = this.previousElementSibling && this.previousElementSibling.key;
```
m6jones, line 145: I will be modifying the object sent by the event in a way where we can remove the need to know about our siblings! This might come as a separate PR. Tight deadline and this works? Tests are more important and I will see if I get to this.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		const dropSpots = dropSpotsFactory();
+		dropSpots.updateTime(e.timeStamp);
+		e.preventDefault();
+	}
+
+	_dragStart(e) {
+		e.dataTransfer.setData('text/plain', `${this.dragText}`);
+		e.dataTransfer.effectAllowed = 'move';
+
+		const nodeImage = this.shadowRoot.querySelector('.d2l-list-item-drag-image') || this;
```
m6jones, line 190: This is where we define the image. If the list item defines a specific image use that. Otherwise just use the entire element.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		e.dataTransfer.setDragImage(nodeImage, 50, 50);
+
+		dropSpotsFactory(this);
+
+		setTimeout(() => {
+			this.beingDragged = true;
+		});
```
m6jones, line 197: This allows us to style the element that is left behind from dragging it.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+			const dropSpots = dropSpotsFactory();
+			dropSpots.setDestination(this, move.below);
+			this._inTopArea = false;
+		}
+	}
+
+	_drop() {
+		const dropSpots = dropSpotsFactory();
+		dropSpots.setDestination(this, dropSpots.moveBeforeDestination);
+	}
```
m6jones, line 219: Make sure that where we are dropping it is where we will move it to.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+
+	_hostDragEnter(e) {
+		const dropSpots = dropSpotsFactory();
+		if (this === dropSpots.target) {
+			return;
+		}
+		dropSpots.addVisitor(this);
```
m6jones, line 226: This will make sure we clean up and set `this._dropTarget=false` when we are done draggin

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		e.dataTransfer.dropEffect = 'move';
+	}
+
+	_moveItem(targetKey, destinationKey, moveBeforeDestination = false, temporaryMovement = false) {
 		this.dispatchEvent(new CustomEvent('d2l-list-item-position', {
-			detail: detail,
+			detail: new NewPositionEventDetails({targetKey, destinationKey, moveBeforeDestination, temporaryMovement}),
 			bubbles: true
 		}));
 	}
```
m6jones, line 236: The signature of this methods needs to be cleaned up. It bugs me and I want to add the feature of instead of a destination to have move up or down by a number.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
 
-	_dragEnter() {
-		this._dragStartHandler();
+	_renderBottomPlacementMarker(renderTemplate) {
+		return this._bottomPlacementMarker ? html`<div class="d2l-list-item-drag-bottom-marker">${renderTemplate}</div>` : null;
+	}
+
+	_renderDraggableArea(templateMethod) {
+		templateMethod = templateMethod || (dragArea => dragArea);
+		return this.draggable && !this._keyboardMode ? templateMethod(html`
```
m6jones, line 244: `this.draggable && !this._keyboardMode` see more complicated cases on when to show and hide the drag items.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
-		this._dispatchDragEvent(null);
+	_renderDropArea(templateMethod) {
+		templateMethod = templateMethod || (dropArea => dropArea);
+		return this.draggable && this._dropTarget ? templateMethod(html`
+			<div class="d2l-list-item-drag-drop-grid" @drop="${this._drop}" @dragover="${this._dragOver}">
+				<div @dragenter="${this._dragTopEnter}"></div>
+				<div @dragenter="${this._dragUpperEnter}"></div>
+				<div @dragenter="${this._dragLowerEnter}"></div>
+				<div @dragenter="${this._dragBottomEnter}"></div>
+			</div>
```
m6jones, line 271: Defines: https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#Drag-Areas-for-a-List-Item

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
-			detail: detail,
+			detail: new NewPositionEventDetails({targetKey, destinationKey, moveBeforeDestination, temporaryMovement}),
 			bubbles: true
 		}));
 	}
 
-	_dragEnter() {
-		this._dragStartHandler();
+	_renderBottomPlacementMarker(renderTemplate) {
+		return this._bottomPlacementMarker ? html`<div class="d2l-list-item-drag-bottom-marker">${renderTemplate}</div>` : null;
```
m6jones, line 239: Lets the list item style the placement markers. Maybe not all lists will have the same one?

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
 
-	_dragStartHandler() {
-		this.moveBefore(null);
-		this._dispatchDragEvent(null);
+	_renderTopPlacementMarker(renderTemplate) {
+		return this._topPlacementMarker ? html`<div class="d2l-list-item-drag-top-marker">${renderTemplate}</div>` : null;
+	}
+};
+
+let dropSpots = null;
```
m6jones, line 280: So, I needed information about what is happening in the drag during all events. The native behaviour of data transfer object doesn't always give information you store. So this was my way of getting it. Since only one drag event can happen at once this is safe.

components/list/list-item-drag-mixin.js
m6jones, line 297: So what is up with this object? 
Why do we need it? 
Can't we just clean up using the leave events that drag n drop offer? 

Well, my first pass was using the leave events. (I kept a commit with them if people are interested) This would lead to things not always being cleaned up. Which was bad. Not to mention the janky user experience it caused. 

So I instead decided to go with one point state system while dragging. This is the object that holds the state and keeps track of cleaning things up.

This was also the place to implement the 1 second area around drop targets.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,299 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+			this._destination = null;
+		}, timeDelayForLeavingTheDropArea);
+	}
+
+	_cleanUpOnLeave() {
+		if (!this._destination) return;
+		this._destination._topPlacementMarker = false;
+		this._destination._bottomPlacementMarker = false;
+		this._destination._inTopArea = false;
+		this._destination._inBottomArea = false;
```
m6jones, line 368: This class is friends with the component above.

components/list/list-item-drag-mixin.js
```diff
@@ -171,10 +445,12 @@ export class NewPositionEventDetails {
 		// now that we have a reference to the item, shove everything between the
 		// destination to the origin over one
 		if (origin > destination) {
+			destination = this.moveBeforeDestination ? destination : destination + 1;
```
m6jones, line 448: Placing before or after the destination was based off of the location compared with the origin. So this allows the component to decide if it goes above or below.

components/list/list-item-generic-layout.js
```diff
@@ -64,6 +64,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 
 			::slotted([slot="actions"]) {
 				grid-column: actions-start / actions-end;
+				justify-self: end;
```
m6jones, line 67: This push aligns actions to always be at the end of the list item. So they line up. This is also fixed in Zina's PR and will need to be removed to take in Zina's fix.

m6jones:
My next goal is to add some tests.

There are a few other things I want to add but it doesn't take away from the functionality and this is a blocking story for the rest of the work. So I am going to deferrer it so it becomes not blocking.

components/list/demo/list-drag-n-drop-sample.js
```diff
@@ -0,0 +1,78 @@
...
+				secondary: 'This course explores the geological process of the Earth\'s interior and surface. These include volcanism, earthquakes, mountains...',
+				img: 'https://s.brightspace.com/course-images/images/36c5813d-2ac3-4a73-8f39-3d6e1b381fe3/tile-high-density-max-size.jpg'
+			}
+		];
+	}
+
+	render() {
+		return html`
+			<d2l-list>
+				${repeat(this.list, (item) => item.key, (item) => html`
```
m6jones, line 59: If you use keyboard mode and move an item up. It works!! Move an item down and we lose focus and thus keyboard mode.

My solution to this is making a better directive. If anyone has something better let me know! Not sure how much time we have to create the directive :O

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,298 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+	if (dropSpots) {
+		dropSpots.clear();
 	}
+	dropSpots = null;
+}
 
-	_dragStopHandler() {
-		this._dispatchDragEvent(null);
-		this.moveafter(null);
+class DropSpotsState {
```
m6jones, line 296: So what is up with this object? 
Why do we need it? 
Can't we just clean up using the leave events that drag n drop offer? 

Well, my first pass was using the leave events. (I kept a commit with them if people are interested) This would lead to things not always being cleaned up. Which was bad. Not to mention the janky user experience it caused. 

So I instead decided to go with one point state system while dragging. This is the object that holds the state and keeps track of cleaning things up.

This was also the place to implement the 1 second area around drop targets.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,298 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		if (this === dropSpots.target) {
+			return;
+		}
+		dropSpots.addVisitor(this);
+		this._dropTarget = true;
+		e.dataTransfer.dropEffect = 'move';
+	}
+
+	_moveItem(targetKey, destinationKey, moveBeforeDestination = false, temporaryMovement = false) {
 		this.dispatchEvent(new CustomEvent('d2l-list-item-position', {
```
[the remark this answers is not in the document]
dbatiste: I think `d2l-list-item-position-change` would be a good name to use for this.

components/list/demo/list-drag-n-drop-sample.js
```diff
@@ -0,0 +1,78 @@
+import '../list-item-content.js';
+import './list-item-sample.js';
+import '../list.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { repeat } from 'lit-html/directives/repeat';
+
+class listDragNDropSample extends LitElement {
```
dbatiste, line 7: Nit: I think in the past we've done something like `CustomListItem` or `DemoListItem`.

components/list/list-item-drag-mixin.js
```diff
@@ -158,9 +432,8 @@ export class NewPositionEventDetails {
 	 */
 	reorder(list, {announceFn, keyFn}) {
```
dbatiste, line 433: The reorder logic doesn't seem like it should belong to the event details.  After all, these is just positioning event info.  While it is nice that it works on a generic list, I think this logic better resides in a list component/mixin.

components/list/demo/list-drag-n-drop-sample.js
dbatiste: Let's get to the bottom of that - sounds like focus is being moved elsewhere for some reason.  Maybe worst case scenario is we have to reapply focus after it's been moved.

components/list/demo/list-item-sample.js
dbatiste: This results in new functions being created for each `render` - in the context of event handlers, the linter will catch this for perf reasons.  The downside to moving it to functions on the class though, is that it's less clear to readers what's going to get rendered.

components/list/demo/list-drag-n-drop-sample.js
m6jones: That makes sense. I'll fix that.

components/list/demo/list-drag-n-drop-sample.js
m6jones: Yeah, this is a tricking one. So in the prototype the issue was that the object moving got re-render when moving down. I believe it has to do with how repeat handles its slices. 

I thought about doing the focus after it has been moved. What makes that hard and out of my reach is the blur happens at this level and so the list items don't really know when it is done. I would love a solution where the person using the drag n drop list doesn't need to know about it. I really like how simple it is to actually use drag n drop list.

components/list/demo/list-item-sample.js
```diff
@@ -40,6 +42,32 @@ class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
 			[slot="content-action"]:focus {
 				outline: none;
 			}
+			.d2l-list-item-drag-image {
```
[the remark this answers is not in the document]
m6jones: Yes, so if this is not given then it will set the image to :host. This is if you want a more special effect. I pushed the rotation of drag image to the list-item since it is visual aspect and maybe different list items would want different visualizations. 

There is also a tech reason - the mixin would need some knowledge about how the dom is setup in the list item to make the tilt work. So the above justifies this approach to fix that problem. 

I do believe in my justification since I do want the list-item components to define the style while using these mixins for their functionality and logic. I also realize, I more inclined to believe that reason has justification due to the fact there are tech reason another way won't work.

components/list/demo/list-item-sample.js
```diff
@@ -40,6 +42,32 @@ class ListItemSample extends ListItemMixin(ListItemCheckboxMixin(LitElement)) {
...
+				transform: rotate(-1deg);
+			}
+			d2l-list-item-generic-layout {
+				background: white;
+				transform: rotate(1deg);
+			}
+			[slot="content"] {
+				justify-content: stretch;
+				padding: 0.55rem 0px;
```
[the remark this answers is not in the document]
m6jones: I always do this! Thank you. I do actually look too.

components/list/demo/list-item-sample.js
m6jones: I'll split it off into a method and hopefully the name can hopefully make it still easier to read. The docs on how to make a new list item style should write them in the more readable way and make a note to split it off.

components/list/list-item-drag-mixin.js
m6jones: Never mind this is real!

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,298 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+	moveItemBeforeMe(itemKey) {
 		if (itemKey === undefined) {
 			return;
 		}
 
-		this._dispatchDragEvent(itemKey);
+		this._moveItem(itemKey, this.key, true);
+	}
+
+	_dragAreaClick(e) {
```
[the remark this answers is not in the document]
m6jones: I agree. It also helps organize them better. If you notice some of them are separated which made it really annoying to find and work on them. Will fix.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,298 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		this.shadowRoot.querySelector(`#${this._itemDragId}`).activateKeyboardMode();
```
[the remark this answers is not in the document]
m6jones: Thanks!

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,298 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		e.dataTransfer.dropEffect = 'move';
+		const dropSpots = dropSpotsFactory();
+		dropSpots.setDestination(this, move.below);
+		this._inBottomArea = true;
+	}
 
+	_dragEnd(e) {
+		const dropSpot = dropSpotsFactory();
+		this.beingDragged = false;
+		this.tilt = false;
```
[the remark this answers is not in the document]
m6jones: I missed it when deleting it. It an artifact from when I was trying to get the tilt to happen from within the mixin.

components/list/list-item-drag-mixin.js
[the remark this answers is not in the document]
m6jones: I agree. Will update it.

components/list/list-item-drag-mixin.js
m6jones: So all the rendering is going to be handled by the hypermedia components that are using this component. The functionality of the list is basically send events up to say this has move above/below another item in the list.

So any consumer will have a list of items. When we tell them, hey this item should move here. They would then have to grab their list and reorder it based on the information. Which they would be like, that is easy enough I'll just write a private method that does it. So we would then get a lot of different ways on reordering a list. It is fun to write real world interview questions but I figured why not just send up a method that does it for them. So then we only have one piece of code the reorders a list.

This also solve another issue where we need to announce what is moving and where it is going for screen readers. Our list component is dumb and has no knowledge of its contents. Each list item doesn't even know what position in the list it is in. So instead of relaying on every implementation of the list to do things right with accessibility. I figured let's make it as easy as possible by giving them ability to run the above method to announce the change where the information required is known.

Also, I plan on making another change where destination could be move up 1 or move down 1 (which would easily translate to move by n and move down by n) So all I would have to do is update this method and both types of destinations would work easily and the user wouldn't have to worry about which type of destination is sent up.

If it is a performance worry. We could always export the method instead of sending up the class and the new method would just take the event as a parameter. 

I would still prefer to keep as part of the detail class information since it more organized and easier to see that it exists for someone using it. 

As for the logic living in the mixin or list component sadly the information required to run this method is not something either has access too.

components/list/demo/list-drag-n-drop-sample.js
m6jones: Side note this is actually, a component that uses the list component. But I will name it with Demo in front.

components/list/demo/list-demo-item-sample.js
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	_renderOutsideControl(dragHandle) {
+		return html`<div slot="outside-control">${dragHandle}</div>`;
+	}
+
+	_renderOutsideControlAction(dragArea) {
+		return html`<div slot="outside-control-action">${dragArea}</div>`;
+	}
+}
```
m6jones, line 138: Here is what I changed in the file

components/list/demo/list-demo-item-sample.js
```diff
@@ -0,0 +1,140 @@
...
+			.d2l-list-item-drag-image {
+				transform: rotate(-1deg);
+			}
+			d2l-list-item-generic-layout {
+				background: white;
+				transform: rotate(1deg);
+			}
+			[slot="content"] {
+				justify-content: stretch;
+				padding: 0.55rem 0;
```
m6jones, line 52: Updated this to be 0

components/list/list-item-drag-mixin.js
dbatiste: Yeah I am all for writing this logic once and only once.  I most definitely don't want to see consumers having to write this sort of logic, or the complex layout logic we have.  It all needs to be very simple for consumers to use.

> As for the logic living in the mixin or list component sadly the information required to run this method is not something either has access too.

Hmmm, this is unfortunate.  A details/info class I usually expect to be just that.  I'll give it another look.

components/list/list-item-drag-mixin.js
zinabat: Maybe `insertBefore`?

components/list/list-item-generic-layout.js
zinabat: #678

components/list/list-item-drag-mixin.js
dbatiste: It would be nice to not need to keep track of drop targets in the state map (`_visitedDestination`).  Some other options to consider:

* only conditionally rendering based on `draggable`, or
* reset `_dropTarget` in `dragleave` or `dragexit` events, thereby avoiding holding references in state object

If we really need to conditionally render, perhaps consider renaming this `_dropTarget` property since it is currently named as an object/element rather than a boolean.  Perhaps `_draggingOver` would be better.

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,76 @@
...
+	below: false
+});
+
+const timeDelayForLeavingTheDropArea = 1000; //ms
+
 export const ListItemDragMixin = superclass => class extends superclass {
 
 	static get properties() {
 		return {
 			draggable: { type: Boolean, reflect: true },
```
dbatiste, line 18: While draggability is controlled by this property, any item would be a drop target, correct?

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +94,286 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+
+	addVisitor(destination) {
+		if (destination && !this._visitedDestination.has(destination)) {
+			this._visitedDestination.set(destination, null);
+		}
+	}
+
+	clear() {
+		this._cleanUpOnLeave();
+		this._visitedDestination.forEach((_, visitedDestination) => visitedDestination._dropTarget = false);
```
dbatiste, line 299: Would be nice to avoid having to do this (per previous comment).

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,76 @@
...
 		return {
 			draggable: { type: Boolean, reflect: true },
-			key: { type: String, reflect: true }
+			dragText: { type: String, attribute: 'drag-text' },
+			key: { type: String, reflect: true },
+			dragging: { type: Boolean, reflect: true, attribute: 'dragging' },
+			_bottomPlacementMarker: { type: Boolean },
+			_dropTarget: { type: Boolean },
+			_keyboardMode: { type: Boolean },
+			_topPlacementMarker: { type: Boolean }
```
dbatiste, line 25: I think this `_topPlacementMarker` and `_bottomPlacementMarker` are named like objects/elements but are actually booleans, which makes the code a little harder to understand.  These booleans are derived from the state object's `insertBefore` boolean, and is therein set by `setDestination(destination, insertBefore)`.  Ultimately the value comes from the `move` enum (which being a verb, is named like a method).  

I think some of this can be cleaned up by:

* renaming `move` to `moveLocations` or perhaps even `dropLocation` might be more appropriate
* updating `setDestination(destination, insertBefore)` signature to be `setDestination(destination, dropLocation)`
* updating `_insertBefore` in the state object to be `_dropLocation`
* replacing `_topPlacementMarker`and `_bottomPlacementMarker` with `_dropLocation`

Basically, I think if we name this stuff consistently then it makes it easier to follow.

components/list/list-item-drag-mixin.js
m6jones: I really tired to make dragexit or dragleave work. Sadly, I ran into problems with the mouse being moved to quickly. This is why I decided for a central holder of the state. It also allows us to extend the 1 second border. Your other comments below will make it better. 

I did keep the commit around for when I did use drag leave. It worked, but there were edge cases that were wonky.

This comment has similar information: https://github.com/BrightspaceUI/core/pull/665#discussion_r449159726

dbatiste: Yeah maybe complicated by the setTimeout...?  I'd expect the event handlers to be triggered in a consistent order, and synchronously.

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,283 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+function dropSpotsBlowUp() {
+	if (dropSpots) {
+		dropSpots.clear();
+	}
+	dropSpots = null;
+}
+
+class DropSpotsState {
+	constructor(target) {
+		this._target = target;
```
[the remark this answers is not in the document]
m6jones: So I have settled on - which I won't explain which is which to see if they are good:
dragTarget
dropTarget

components/list/list-item-drag-mixin.js
```diff
@@ -33,86 +95,283 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		this._destination = null;
+		this._visitedDestination = new Map();
+		this._insertBefore = false;
+		this._time = 0;
+	}
+
+	addVisitor(destination) {
```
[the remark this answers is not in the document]
m6jones: fixed

components/list/list-item-drag-mixin.js
m6jones: As explained not possible :(

components/list/list-item-drag-mixin.js
```diff
@@ -1,20 +1,82 @@
...
-export const ListItemDragMixin = superclass => class extends superclass {
+export const dropLocation = Object.freeze({
+	above: 1,
+	below: 2,
+	first: 3,
+	last: 4,
+	shiftDown: 5,
+	shiftUp: 6,
+	void: 0
+});
```
m6jones, line 15: I am not 100% on these names:
- **Above**: Move the element above the dropTarget.
- **Below**: Move the element below the dropTarget
- **First**: Move the element to the first position of the list.
- **Last**: Move the element to the last position of the list.
- **ShiftDown**: Shift the position of an element down by 1.
- **ShiftUp**: Shift the position of an element up by 1.
- **Void**: No drop location. Used to hide both markers.

zinabat: This is so much better!

components/list/demo/list-demo-drag-n-drop-usage.js
```diff
@@ -0,0 +1,79 @@
+import '../list-item-content.js';
+import './list-demo-item-sample.js';
+import '../list.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import {ifDefined} from 'lit-html/directives/if-defined';
```
dbatiste, line 5: Nit: inconsistent spacing.

components/list/list-item-drag-mixin.js
```diff
@@ -24,96 +84,285 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+		if (this._inBottomArea) {
+			const dropSpots = getDragState();
+			dropSpots.setActiveDropTarget(this, dropLocation.above);
+			this._inBottomArea = false;
+		}
 	}
 
-	_dragExit() {
-		this._dispatchDragEvent(null);
+	_onDropAreaUpperDragEnter(e) {
```
dbatiste, line 192: Sounds like we decided on "drop target" term... do these handlers and reference to "dropArea" in `render` also need to be updated?

components/list/list-item-drag-mixin.js
```diff
@@ -24,96 +84,285 @@ export const ListItemDragMixin = superclass => class extends superclass {
...
+function createDragState(target) {
+	clearDragState();
+	dragState = new DragState(target);
+	return dragState;
+}
+
+function getDragState() {
+	if (!dragState) createDragState();
+	return dragState;
+
```
dbatiste, line 264: Nit: whitespace.

PR #668, opened 2020-07-02
fix: introducing a delay of 800ms to the shim before the fade transition starts

spandanabaruah:
https://rally1.rallydev.com/#/29180338367d/detail/defect/398651793780?fdp=true

adding a 800ms delay to the shim after which the fade transition will start. 

Use case in FACE: When the shim is activated after the 'Save and Close' button is clicked the first time on the Edit Assignment page, it blocks all user interactions and so this solves the problem of clicking 'save and close' multiple times resulting in multiple copies of the file being attached to an assignment - while also maintaining the delay required by design between clicking 'Save and Close' and being directed to the 'Manage Assignments' page.

### DIFF
--- a/components/backdrop/README.md
+++ b/components/backdrop/README.md
@@ -30,6 +30,7 @@ button.addEventListener('click', () => {
 - `shown` (Boolean): used to control whether the backdrop is shown
 - `slow-transition` (Boolean): Increases the fade transition time to 1200ms (default is 200ms)
 - `no-animate-hide` (Boolean): disables the fade-out transition while the backdrop is being hidden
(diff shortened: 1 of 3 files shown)


dlockhart:
Hi @spandanabaruah -- I know the solution you folks are working on is kind of in flux, but does this new delay on the transition mean that the `no-animate-hide` attribute that was added previously is no longer required?

spandanabaruah:
> Hi @spandanabaruah -- I know the solution you folks are working on is kind of in flux, but does this new delay on the transition mean that the `no-animate-hide` attribute that was added previously is no longer required?

Hi @dlockhart, the `no-animate-hide` attribute will still be required in the case of an error where we will want to hide the backdrop instantly and return the UI to the user

PR #669, opened 2020-07-03
Add style linting

dbatiste:
This PR adds style linting for components/mixins in `core`.  Many of the corrections are stylistic, however there are a few legitimate syntax "errors" (that browsers tolerate) that are fixed as well.  There are a few things I left out...

* there are many options for enforcing whitespace - we can tweak more if desired.  I did not enforce them all because sometimes it's nice to have a little flexibility for whitespace when organizing selectors.

* there is a rule to enforce value functions (ex. `calc`) as lower case (which is the default), however I could not enable it due to variables being substituted in tooltip styles without more changes.  Not a big deal - since I can't remember the last time i needed to correct that.

* there is a default rule that enforces no descending specificity (selectors further down that have lower specificity), however I did not enable it because it will be very disruptive, and the value is iffy because it can result in selectors being organized differently. The rationale for this rule is that that scenario can be confusing since people generally expect rules further down the page to trump rules above. We can play with that in the future if we want.

Generally, this config catches almost all the things we generally nitpick in reviews - there are many we missed that I fixed in this PR which gives an idea of what it found for us.

List of rules can be found... [here](https://github.com/stylelint/stylelint/blob/a8042e7192cad993081cf982901948dbd5004048/docs/user-guide/rules/list.md#possible-errors).


.stylelintrc.json
```diff
@@ -0,0 +1,102 @@
+{
+  "extends": "stylelint-config-recommended",
+  "plugins": [
+    "stylelint-order"
+  ],
+  "ignoreFiles": [
```
dbatiste, line 6: We'll need to exclude the prismjs styles indefinitely.

I ignored the list components for the moment because @m6jones and others are in the midst of lots of changes under tight timeline and the linting fixes will be disruptive for their PR.

.stylelintrc.json
```diff
@@ -0,0 +1,102 @@
...
+    "at-rule-name-space-after": "always-single-line",
+    "at-rule-semicolon-newline-after": "always",
+    "block-closing-brace-newline-after": "always",
+    "block-closing-brace-newline-before": "always-multi-line",
+    "block-closing-brace-space-before": "always-single-line",
+    "block-opening-brace-newline-after": "always-multi-line",
+    "block-opening-brace-space-after": "always-single-line",
+    "block-opening-brace-space-before": "always",
+    "color-hex-case": "lower",
+    "color-hex-length": "long",
```
dbatiste, line 22: We have the choice of `short`, `long`, or omitting this rule.  We've typically always used the long hex format so that's what I set here.

.stylelintrc.json
```diff
@@ -0,0 +1,102 @@
...
+    "comment-whitespace-inside": "always",
+    "custom-property-pattern": [ "^d2l-", {
```
dbatiste, line 24: This rule (regex) will allow selectors for custom elements such as `d2l-icon { ... }`.  Otherwise it will flag them as unknown.

.stylelintrc.json
```diff
@@ -0,0 +1,102 @@
...
+      "declarations"
+    ],
+    "order/properties-alphabetical-order": true,
+    "property-case": "lower",
+    "selector-attribute-brackets-space-inside": "never",
+    "selector-attribute-operator-space-after": "never",
+    "selector-attribute-operator-space-before": "never",
+    "selector-combinator-space-after": "always",
+    "selector-combinator-space-before": "always",
+    "selector-class-pattern": "^d2l-",
```
dbatiste, line 84: This rule (regex) enforces that custom classes must start with `d2l-`.

components/backdrop/backdrop.js
```diff
@@ -49,15 +49,16 @@ class Backdrop extends LitElement {
 			:host([slow-transition]) {
 				transition: opacity 1200ms ease-in;
 			}
-			:host([_state=null][no-animate-hide]) {
```
dbatiste, line 52: This is an example where I "corrected" the descending specificity of the selectors (which I did not ultimately enforce).

components/dropdown/dropdown-content-styles.js
```diff
@@ -13,50 +13,47 @@ export const dropdownContentStyles = css`
...
 	.d2l-dropdown-content-pointer > div {
 		background-color: #ffffff;
 		border: 1px solid var(--d2l-color-mica);
 		border-radius: 0.1rem;
-		box-shadow: -4px -4px 12px -5px rgba(73, 76, 78, .2); /* ferrite */
+		box-shadow: -4px -4px 12px -5px rgba(73, 76, 78, 0.2); /* ferrite */
 		height: 16px;
-		width: 16px;
-		transform: rotate(45deg);
 		-webkit-transform: rotate(45deg);
```
dbatiste, line 44: Note for devs: ignore browser prefixes when ordering.

components/expand-collapse/expand-collapse-content.js
```diff
@@ -47,8 +47,8 @@ class ExpandCollapseContent extends LitElement {
 			}
 
 			/* prevent margin colapse on slotted children */
-			.d2l-expand-collapse-content-inner:before,
-			.d2l-expand-collapse-content-inner:after {
+			.d2l-expand-collapse-content-inner::before,
```
dbatiste, line 50: Found a bunch of these cases where we were incorrectly using pseudo-class syntax (single `:`) instead of pseudo-element (doubel `::`).

components/tooltip/tooltip.js
```diff
@@ -253,8 +253,8 @@ class Tooltip extends RtlMixin(LitElement) {
 
 			.d2l-tooltip-content {
 				background-color: var(--d2l-tooltip-background-color);
-				border-radius: ${contentBorderRadius}px;
 				border: ${contentBorderSize}px solid var(--d2l-tooltip-border-color);
+				border-radius: ${contentBorderRadius}px;
```
dbatiste, line 257: These variables prevent us from enforcing lower case for property value functions.  If we convert these to lower case then we can enable, but that's gross too.  Some other work-arounds are possible too.

.stylelintrc.json
```diff
@@ -0,0 +1,102 @@
...
+    "media-query-list-comma-newline-after": "always-multi-line",
+    "media-query-list-comma-space-after": "always-single-line",
+    "media-query-list-comma-space-before": "never",
+    "no-descending-specificity": null,
+    "no-eol-whitespace": [ true, {
+      "ignore": [ "empty-lines" ]
+    } ],
+    "number-leading-zero": "always",
+    "number-no-trailing-zeros": true,
+    "order/order": [
```
dbatiste, line 73: These two rules configure the order.  Optionally, we can specify an order other than alphabetic.

components/dropdown/dropdown-content-styles.js
dlockhart: Except they should precede the non-prefixed version so that browsers who understand both use the standard one.

package.json
```diff
@@ -66,6 +67,9 @@
     "puppeteer": "^5",
     "semantic-release": "^17",
     "sinon": "^9",
+    "stylelint": "^13.6.1",
+    "stylelint-config-standard": "^20.0.0",
+    "stylelint-order": "^4.1.0",
```
dlockhart, line 72: Nit: can exclude minor/patch versions here

components/dropdown/dropdown-content-styles.js
dbatiste: I'll see if I can find a way to enforce this.

dbatiste: As it turns out, the order plugin already enforces this.  :)

PR #676, opened 2020-07-03
Integrate mixins into d2l-list-item (Part 1)

zinabat:
## Context
[US115502](https://rally1.rallydev.com/#/detail/userstory/384002687644?fdp=true)
[Design Document](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/996083189/US114586+wc-list+Refactor+for+Drag+n+Drop#d2l-list-item)

- Adds the `ListItemMixin` and `ListItemCheckboxMixin` to the `ListItem`
- Removes the logic now contained in these mixins from the `ListItem`
- Grid can now be enabled on default lists using the `grid` property
  - New demo snippet showing grid enabled added

### Upcoming Changes
- `ListItemDragMixin` will be integrated when it is finished and tested
- Fixes for NVDA screen reading with grid (currently doesn't play nicely)

## UX Result
Kapture 2020-07-03 at 15 24 36

## Quality
- Unit tests for generic list item were changed to use `d2l-list-item`
- New demo snippet allows for manual testing

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -39,7 +39,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 					[start outside-control-start] minmax(0, min-content)
 					[control-start outside-control-end] minmax(0, min-content)
 					[control-end content-start] auto
-					[content-end actions-start] auto
+					[content-end actions-start] minmax(0, max-content)
 					[end actions-end];
 			}
 			::slotted([slot="outside-control"]),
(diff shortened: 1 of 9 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -49,12 +67,12 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				grid-row: 1 / 2;
 			}
 			::slotted([slot="outside-control"]) {
-				width: 42px;
+				width: 2.1rem;
```
m6jones, line 70: Fixed in https://github.com/BrightspaceUI/core/pull/678 (as you know since you wrote it.)

components/list/list-item-mixin.js
```diff
@@ -4,6 +4,9 @@ export const ListItemMixin = superclass => class extends superclass {
```
m6jones, line 4: Could you name this class `ListItemRoleMixin`?

PR #677, opened 2020-07-07
input-date: initial work on validation

margaree:
~~Still WIP. Wanting to confirm that I'm on the right track and using this correctly.~~

The behavior we want is that if a user enters an invalid date, the text stays as is (which is why there is the addition of `_shownValue` so as not to affect `value`) but the validation tooltip would appear. There is further work I'm currently doing on calendar to support this behavior.


components/calendar/calendar.js
```diff
@@ -93,13 +93,6 @@ export function getDatesInMonthArray(shownMonth, shownYear) {
 	return dates;
 }
 
-export function getDisabled(date, min, max) {
```
margaree, line 96: The change to calendar was just to move this functionality to `dateTime` helper since input-date now needs it.

components/inputs/input-date.js
```diff
@@ -20,7 +23,7 @@ export function formatISODateInUserCalDescriptor(val) {
  * A component that consists of a text input field for typing a date and an attached calendar (d2l-calendar) dropdown. It displays the "value" if one is specified, or a placeholder if not, and reflects the selected value when one is selected in the calendar or entered in the text input.
  * @fires change - Dispatched when a date is selected or typed. "value" reflects the selected value and is in ISO 8601 calendar date format ("YYYY-MM-DD").
  */
-class InputDate extends LocalizeCoreElement(LitElement) {
+class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
```
margaree, line 26: I used `FormElementMixin` in with input-date as well as validation-custom. Let me know if I used these incorrectly. I had added a console log within form element mixin locally to confirm the errors were happening when expected, so it seems to work.

components/inputs/input-date.js
```diff
@@ -51,11 +54,13 @@ class InputDate extends LocalizeCoreElement(LitElement) {
 			/**
 			 * Value of the input
 			 */
+			validationCustomValid: { type: Boolean },
```
margaree, line 57: Not sure if this needs to be public?

components/inputs/input-date.js
AllanKerr: `FormElementMixin` does support `d2l-validation-custom` so there's no issue with that. Although, I think in this case we can likely accomplish validation without it since its only needed for `async` validation.

components/inputs/input-date.js
```diff
@@ -297,25 +333,34 @@ class InputDate extends LocalizeCoreElement(LitElement) {
...
-		this._formattedValue = this.value ? formatISODateInUserCalDescriptor(this.value) : (this.emptyText ? this.emptyText : '');
+		this._formattedValue = this._shownValue ? formatISODateInUserCalDescriptor(this._shownValue) : (this.emptyText ? this.emptyText : '');
 	}
 
-	_updateValueDispatchEvent(dateInISO) {
-		if (dateInISO === this.value) return;
+	async _updateValueDispatchEvent(dateInISO) {
+		if (dateInISO === this._shownValue) return; // prevent validation from happening multiple times for same change
+		this._shownValue = dateInISO;
+		this.validationCustomValid = !dateInISO || isDateInRange(getDateFromISODate(dateInISO), getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue));
```
AllanKerr, line 350: Since this validation check is synchronous I would recommend using `setValidity` which is modeled after Form Associated Custom elements / the HTML 5 validation API.

It would look something along the lines of:
```
if (this.minValue && date.getTime() < getDateFromISODate(this.minValue).getTime()) {
    this.setValidity({ rangeUnderflow: true });
} else if (this.maxValue && date.getTime() > getDateFromISODate(this.maxValue).getTime()) {
    this.setValidity({ rangeOverflow: true });
} else {
    this.setValidity({});
}
```

Then to customize the messages you can override these:
```
get validationMessageRangeOverflow() {
    return 'It is too large, must be below max!';
}

get validationMessageRangeUnderflow() {
    return 'It is too low, must be above min!';
}
```

components/inputs/input-date.js
AllanKerr: Normally when using `d2l-validation-custom` the logic should be entirely contained within the event handler. You could do something along the lines of:
```
async _validate(e) {
    const isValid = !this._shownValue || isDateInRange(getDateFromISODate(this._shownValue), getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue));
	e.detail.resolve(isValid);
}
```
`isValidationCustomValid` is public in the test component just so that I can assert that `validate()` returns a value that matches with the result returned from the `d2l-validation-custom`.

components/inputs/input-date.js
```diff
@@ -297,25 +333,34 @@ class InputDate extends LocalizeCoreElement(LitElement) {
...
+		await this.validate();
```
AllanKerr, line 351: `validate` returns a list of errors that occurred so I'd recommend checking the result of this to determine if validation succeeded.

But we do need some discussion on whether we want `validate` to be called by the components themselves or if they need to be wrapped in a form / validation group to perform automatic validation.

The main reason for having the form / validation group do it is that we need to support both native and custom form elements and native elements have no way of calling their own `validate` function. We might be able to do some hybrid but definitely something that should be discussed.

components/inputs/input-date.js
margaree: Awesome, that seems to work well - thanks!

components/inputs/input-date.js
margaree: Yup for sure - when I had talked to Jeff about how input-date should validate, he wanted the validation to happen on enter or blur, so it did seem to make the most sense to put the validation directly in the component. However, if the component would be only within this form/validation group that would do the automatic validation, then that could work as well. I don't really have a strong preference, I just wouldn't want the validation to get missed if we end up in an in between state where input-date is being used in a place without whatever wrapper we use to do this automatic validation (not sure if that's a potential case).

components/form/form-element-mixin.js
```diff
@@ -96,7 +96,7 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 	}
 
 	get label() {
-		return null;
+		return this.label ? this.label : null;
```
margaree, line 99: So the lit analyzer was giving me the error
```
 This attribute binds the type 'null' which can be 'null'.
    173:  label="${ifDefined(this.label)}"
    no-nullable-attribute-binding
```

This change fixes it but I'm not sure if this is correct/causes problems with form element mixin. I did try different combinations of `label="${ifDefined(this.label === null ? undefined : this.label)"` in input-date and still had problems.

dlockhart: Looks like `ifDefined` only checked for `undefined` values, not `null` values. So @AllanKerr maybe we could change the `FormElementMixin` to return `undefined` values instead of `null` ones?

margaree: That also does not fix things:
```
This attribute binds the type 'undefined' which can be 'undefined'.
    172:  label="${ifDefined(this.label)}"
    no-nullable-attribute-binding
```

It actually seems like something weird is going on. I tried things from https://marketplace.visualstudio.com/items?itemName=runem.lit-plugin#no-nullable-attribute-binding and they don't fix it

dlockhart: Yeah weird eh... it just says to wrap it in an `ifDefined` call, which you're doing.

> Binding undefined or null in an attribute binding will result in binding the string "undefined" or "null". Here you should probably wrap your expression in the "ifDefined" directive.

AllanKerr: The act of defining `label` as a property should mean the getter in `form-element-mixin` never gets called because it automatically gets overrided. Lit creates getters for properties that look like:
```
get() {
    return this[key];
}
```

AllanKerr: Unfortunately we can't self-reference the property inside the getter like this:.
```
return this.label ? this.label : null;
```
If we do and someone chooses not to override `get label` then we'll get an infinite loop when trying to fetch the label for the element.

margaree: Interesting - so perhaps there is some sort of bug with the linter?

components/inputs/input-date.js
AllanKerr: @dlockhart and I discussed this and we think it makes sense to have components be responsible for calling `validate` themselves so this is the approach we want to take.

The change I would make is to change it to:
```
const errors = await this.validate();
if (errors.length > 0) return;
```
The reason is that `validate` internally calls `checkValidity()` and also handled any `d2l-validation-custom` elements so its always safer to go with `validate`. `checkValidity()` only checks the result of `setValidity(...)`

components/inputs/input-date.js
```diff
@@ -297,19 +338,33 @@ class InputDate extends LocalizeCoreElement(LitElement) {
...
+			if (this.minValue && getDateFromISODate(dateInISO).getTime() < getDateFromISODate(this.minValue).getTime()) {
+				this.setValidity({ rangeUnderflow: true });
+			} else if (this.maxValue && getDateFromISODate(dateInISO).getTime() > getDateFromISODate(this.maxValue).getTime()) {
+				this.setValidity({ rangeOverflow: true });
+			} else {
+				this.setValidity({});
+			}
+		} else {
+			this.setValidity({});
+		}
```
AllanKerr, line 365: Something that came to mind is since the validity flags are booleans you could rewrite this as:
```
this.setValidity({
    rangeUnderflow: dateInISO && this.minValue && getDateFromISODate(dateInISO).getTime() < getDateFromISODate(this.minValue).getTime(),
    rangeOverflow: dateInISO && this.maxValue && getDateFromISODate(dateInISO).getTime() > getDateFromISODate(this.maxValue).getTime()
});
```
Either should work, just depends what your preference is.

margaree: Cool yeah that looks much cleaner - good idea!

components/inputs/input-text.js
```diff
@@ -41,6 +41,10 @@ class InputText extends RtlMixin(LitElement) {
 			 * Disables the input
 			 */
 			disabled: { type: Boolean, reflect: true },
+			/**
```
margaree, line 44: I added `hideInvalidIcon` because in the validation designs, input-date has the calendar icon replaced with the alert icon when invalid input occurs.

components/inputs/input-date.js
```diff
@@ -161,12 +165,14 @@ class InputDate extends LocalizeCoreElement(LitElement) {
 			</div>
 			<d2l-dropdown ?disabled="${this.disabled}" no-auto-open>
 				<d2l-input-text
+					aria-invalid="${this._invalidInput ? 'true' : 'false'}"
```
AllanKerr, line 168: Calling `validate` automatically adds `aria-required` to your component so (with the current pattern) it'd be preferable to add an `aria-required` property and pass that down to `d2l-input-text`. That way you don't need to worry about tracking whether its currently valid or invalid.

components/inputs/input-date.js
```diff
@@ -20,10 +21,14 @@ export function formatISODateInUserCalDescriptor(val) {
...
  */
-class InputDate extends LocalizeCoreElement(LitElement) {
+class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 
 	static get properties() {
 		return {
+			/**
+			 * Indicates that the input value is invalid
+			 */
+			ariaInvalid: { type: String, attribute: 'aria-invalid', reflect: true },
```
dlockhart, line 31: I know this is something that the `FormElementMixin` is setting... but seeing how it ends up working here, I'm wondering if it could instead be a non-reflected property called something like `invalid`? That way each component can do with that what they like... in input-date's case it would transfer that as `aria-invalid='true'` onto its text input. Other components might do other things.

AllanKerr: Its definitely open to change. This was based off of `d2l-input-text` but if thats something we're not happy with now is the time to change it.

dlockhart: I think it's worth changing... otherwise we'll have custom elements sprouting `aria-invalid` attributes which don't actually mean anything. `aria-invalid` should really only apply to the underlying element(s) that are invalid -- in `d2l-text-input`'s case, that's the `<input type="text">`.

AllanKerr: Something that does concern me a bit is every time we make changes like these we diverge from how native and custom form elements are handled. The major cons I see with this are:
1. It increases the learning curve since.
1. It complicates `d2l-form` because there needs to be a bunch of branching logic for if an element is custom or native.
1. It will make it more difficult to ever roll out form associated custom elements if we reach a point when all browsers support them.

I think its reasonable to do in some cases but we should think really carefully when we do to make sure we're actually making the right long term decision.

dlockhart: Hmm ok I wasn't aware that `aria-invalid` was a form-associated-custom-elements thing. Typically ARIA is only ever used as a crutch until standards can catch up -- they rarely actually make use of it themselves.

AllanKerr: Form associated custom elements basically involve extending the HTML 5 validation API to custom components so attributes that work on native components will work on custom form elements. This also includes `aria-invalid`.

It doesn't need to be a form associated custom element thing but it would likely be what we use because `:invalid` has major short-comings that make it unusable:
1. `:invalid` being added on page load for required elements.
1. Not being able to add `:invalid` for `d2l-validator-custom`'s.

I do think we can remove it, I just want to make sure we thinking about the broad picture when we make the decision.

dlockhart: Right. So FACEs (haha) could essentially say... "Ignore the text input(s) inside my shadow root, they're not invalid but I am". And that might end up being accomplished using `aria-invalid`... although ideally the AOM will expose a way to say an element is invalid without using ARIA -- much the same way it's supposed to allow elements to take on particular roles without using the `role` attribute (part of ARIA).

I still feel like we're OK to rename this without deviating too much from FACE. Until FACE + the AOM exist, we'll likely need to be marshalling `aria-invalid` attributes onto nested native elements anyway.

margaree: So just so I'm clear before I make changes - is the thought that `FormElementMixin` should have a property `invalid` that is set instead of the `this.setAttribute('aria-invalid', 'true|false');` and then we have components like `input-date` listen for that and do what they need to?

AllanKerr: Alternatively if we're going the route of not using `aria-invalid` it might make the most sense to combine `showValidationTooltip` / `hideValidationTooltip` and just use a single `validationError` property that can be the current error message or `null`. Then the component can choose what to do with it.

dlockhart: Cool I like that 👍

components/form/form-element-mixin.js
```diff
@@ -217,10 +222,7 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this.localize('components.form-element-mixin.valueMissingMessage', { label: this.labelText });
 	}
 
-	validationTooltipHide() { }
-
-	// eslint-disable-next-line no-unused-vars
-	validationTooltipShow(message) {}
+	validationTooltipShowHide() {}
```
AllanKerr, line 225: I think it makes sense to get rid of this altogether. I was thinking the component could do conditional rendering based on if `validationError` is non-null:
```
{this.validationError !== null ? html`<d2l-tooltip>${this.validationError}</d2l-tooltip>`} : null
```
The only risk there is it might make it a bit easier for people to forget to add tooltips to their custom form elements.

margaree:
I'll create a follow up PR after this one for adding the tooltip into `input-text` (for accessibility) and then using that with `input-date`

components/form/form-element-mixin.js
```diff
@@ -76,12 +76,17 @@ export class FormElementValidityState {
 
 export const FormElementMixin = superclass => class extends LocalizeCoreElement(superclass) {
 
+	static get properties() {
+		return { validationError: { type: String } };
```
dlockhart, line 80: I don't know if we need to explicitly prevent properties from reflecting, but I've seen `attribute: false` used here to at least prevent this getting set from an attribute.

Issue #399, opened 2020-02-26
z-index of list-item actions obscures dropdown 


ktonon:
![Screen Shot 2020-02-26 at 10 38 43 AM](https://user-images.githubusercontent.com/22655/75366676-1be75580-588d-11ea-946e-1c8eeabd679a.png)

```html
<d2l-list-item>
    <div slot="actions"><!-- gets a z-index of 4-->
        <d2l-dropdown>
            <d2l-button-icon
                class="d2l-dropdown-opener"
                icon="tier1:more"
            ></d2l-button-icon>
            <d2l-dropdown-content>...</d2l-dropdown-content>
        </d2l-dropdown>
    </div>
</d2l-list-item>
```

This can be worked around by playing with `z-index` on the dropdown, but should work out of the box.

ktonon:
Workaround:

```html
<div slot="actions" style="position:relative;z-index:2">
<d2l-dropdown-more text="Open!" onclick="myClickFunction(event)" onblur="myBlurFunction(event)">
...
</d2l-dropdown-more>
</div>
```

```javascript
function myClickFunction(e) {
    if (e.target.parentNode) {
        e.target.parentNode.style.zIndex = '5';
    }
}
function myBlurFunction(e) {
    if (e.target.parentNode) {
        e.target.parentNode.style.zIndex = '2';
    }
}
```

dlockhart:
Thanks @ktonon. Sounds like @m6jones is planning some list optimizations soon that will resolve this!

dbatiste:
Yep... thanks guys.  Cool to see how these components are being used together.  😄

PR #678, opened 2020-07-07
Fix: generic list item styles and behaviour

zinabat:
## Changes
- Fixes the z-index dropdown/tooltip issue for list items using the generic layout #399 
- Fixes spacing problem for the action area
- Targets list-items for row changes based on role instead of tag name
- Style change to use `rem` 

**Before**
Screen Shot 2020-07-07 at 10 40 51 AM

**After**
Screen Shot 2020-07-07 at 10 57 32 AM

## Quality
- Two new tests added to ensure this works
- Manually tested
  - [x] Chrome
  - [x] Legacy Edge
  - [ ] CrEdge
  - [x] Firefox
  - [x] Safari


components/list/list-item-generic-layout.js
```diff
@@ -390,6 +409,13 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		this._cellNum = parseInt(slot.getAttribute('data-cell-num'));
 		this._cellFocusedItem = this._getFocusedItemPosition(event.target);
 	}
+
+	_unsetFocusInfo() {
+		setTimeout(() => {
+			this._cellNum = null;
+			this._cellFocusedItem = null;
+		}, 100);
```
zinabat, line 417: It needs a timeout to account transition animations, which look super weird without this.

components/list/list-item-generic-layout.js
```diff
@@ -133,18 +147,28 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
 		return html`
 		<slot name="content-action" class="d2l-cell" data-cell-num="5"></slot>
 		<slot name="outside-control-action" class="d2l-cell" data-cell-num="1"></slot>
 		<slot name="outside-control" class="d2l-cell" data-cell-num="2"></slot>
 		<slot name="control-action" class="d2l-cell" data-cell-num="3"></slot>
 		<slot name="control" class="d2l-cell" data-cell-num="4"></slot>
-		<slot name="actions" class="d2l-cell" data-cell-num="6"></slot>
+		<slot name="actions" class="${classMap(actionClasses)}" data-cell-num="6"
+			@click="${this._clickActionSlot}"
+			@focusout="${this._unclickActionSlot}"></slot>
```
zinabat, line 162: This is a `focusout` instead of a blur so we can catch bubbling events

components/list/list-item-generic-layout.js
```diff
@@ -390,6 +414,17 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
 	}
+
+	_unclickActionSlot(event) {
+		const related = event.relatedTarget;
+		const slot = (event.path || event.composedPath()).find(node =>
+			node.nodeName === 'SLOT' && node.classList.contains('d2l-cell'));
+		if (!related || !isComposedAncestor(slot, related)) {
+			setTimeout(() => {
+				this._actionFocused = false;
+			}, 200);
```
zinabat, line 425: Timeout ensures animations have time to process

m6jones: Can you use the animation end event to get a better timer?

components/list/list-item-generic-layout.js
```diff
@@ -49,12 +60,12 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				grid-row: 1 / 2;
 			}
 			::slotted([slot="outside-control"]) {
-				width: 40px;
+				width: 2.1rem;
```
dbatiste, line 63: I recently added style linting to core but I configured it to ignore list components for the moment, as to not create conflicts for you and Maya.  That being said, this will fail linting once enabled - the properties must be order alpha.  Feel free to correct if you like, or I will fix once enables for the list components.

components/list/list-item-generic-layout.js
```diff
@@ -66,6 +77,9 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				grid-column: actions-start / actions-end;
 				z-index: 4;
 			}
+			.focused::slotted([slot="actions"]) {
```
dbatiste, line 80: Same thing for this one - there is a rule that enforces class names to begin with `d2l-`.  It doesn't protect us 100%, but it's better than nothing.  The rationale is - for legacy Edge, which does not support shadowDOM, elements within components are possible targets for such selectors in the page.

components/list/list-item-generic-layout.js
dbatiste: This event handler is named as if they are ghost click methods.  Suggest `_onActionSlotClick` and `_onActionSlotFocusOut`.  The `_handle` prefix is fine too.

components/list/list-item-generic-layout.js
dbatiste: Which animation are you waiting for?  I agree with Maya here - it would be best to wait for an animation or transition event, but be mindful of components that support prefers reduced-motion, in such case there will be no-animation and therefore no event.  We have a few examples in core that check this in JS.

components/list/list-item-generic-layout.js
dbatiste: Oh, currently this is wiring up to `click` - does your current sol'n work for keyboard users?  Seems like we'd need to wire-up to `focusin` instead of `click`...?

Note: `d2l-card` had to deal with this as well.  [see here](https://github.com/BrightspaceUI/card/blob/master/d2l-card.js#L252).  It specifically wired up to dropdown open/close events.  Not sure I really have a preference on approach - it specifically targets the dropdown scenario with its fix.

components/list/list-item-generic-layout.js
zinabat: Yup, that is a better solution, I agree.

components/list/list-item-generic-layout.js
zinabat: I struggled with this a bit - technically I think it's only ever a problem for dropdown open and close events, so I'm fine with specifically targeting those instead of using `focusout` and `close`. Perhaps I should also be testing this with tooltips.

components/list/list-item-generic-layout.js
m6jones: I am really happy this is a linting rule. Team Media is really big on it and I try to be as well. I even have an extension that allows me to do it easily. 

I don't mind going through everything in lists after and making sure it is all good.

components/list/list-item-generic-layout.js
m6jones: I've been switching all the list to `_on` basically, I had `_handleDragHandleblah` and I was like there is too many handles! Also, `_on` is shorter.

components/list/list-item-generic-layout.js
zinabat: This is now a moot point because they are hooked into events instead.

components/list/list-item-generic-layout.js
zinabat: Fixed

components/list/list-item-generic-layout.js
zinabat: Changed to `_on`. I've also expanded the fix so that it includes tooltips and is available on every cell using a property called `front-row`. Although it's only these two things that need it right now, you never know!

components/list/list-item-generic-layout.js
zinabat: This is now using an attribute on the host to target

components/list/list-item-generic-layout.js
```diff
@@ -39,23 +49,41 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
+			:host([front-row]) ::slotted([slot="control-action"]) {
+				z-index: 4;
+			}
+			:host([front-row]) ::slotted([slot="content-action"]) {
+				z-index: 5;
+			}
+			:host([front-row]) ::slotted([slot="actions"]) {
+				z-index: 6;
+			}
+
```
zinabat, line 72: It's possible I'm missing something and there's an easier way to do this by targeting the host, but I couldn't find one.

components/list/list-item-generic-layout.js
```diff
@@ -384,6 +416,26 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		}
 	}
 
+	_onDropdownClose() {
+		if (!this._tooltipShowing) this.frontRow = false;
+		this._dropdownShowing = true;
```
dbatiste, line 421: I think this should be setting `false` instead.

components/list/list-item-generic-layout.js
```diff
@@ -384,6 +416,26 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
+	}
+
+	_onDropdownOpen() {
+		this.frontRow = true;
+		this._dropdownShowing = true;
+	}
+
+	_onTooltipHide() {
+		if (!this._dropdownOpen) this.frontRow = false;
```
dbatiste, line 430: Looks like this `_dropdownOpen` variable should actually be `_dropdownShowing` (based on naming immediately above above).  Whether we care that the variable name matches the event name... 🤷‍♂️ .

components/list/list-item-generic-layout.js
```diff
@@ -26,8 +26,18 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 
 	static get properties() {
 		return {
+			/**
+			 * @ignore
+			 */
+			frontRow: { type: Boolean, attribute: 'front-row', reflect: true },
```
dbatiste, line 32: I think this should be made private... with leading underscore `_frontRow`, so that consumers who may look at this code thing it's something they can/should try setting.

Alternatively, we could reflect `_dropdownShowing` and `_tooltipShowing` and set up CSS selectors that target both those, to avoid having to come up with a name for this property to describe its purpose.

components/list/list-item-generic-layout.js
dbatiste: Yeah it's not great to sprout attributes for internals on the host, given that the host attributes are sorta part of the public API... subject to be manipulated by consumers if used incorrectly.  That being said, we'd need to apply this attribute or a class to an internal element to avoid it in this case... and in this case, we only have slots, and I'm not 100% sure we can apply custom attributes to slots and effectively target them with CSS in all browsers.

components/list/list-item-generic-layout.js
zinabat: derp.

components/list/list-item-generic-layout.js
zinabat: I abandoned this approach because I had already made a typo in the handlers trying to keep the logic straight. Also, it's 100% cool to put classes on slots, and we can target them very easily. 

The alt approach would have been to put a wrapper `div` around the slots, but I didn't want to overcomplicate the DOM unnecessarily. This means that there's further reason to not put focusable/actionable things in the `content` slot as well.

components/list/list-item-generic-layout.js
```diff
@@ -130,16 +144,25 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 	firstUpdated() {
 		this.addEventListener('keydown', this._handleKeydown.bind(this));
 		this.addEventListener('focusin', this._setFocusInfo.bind(this));
+		this.addEventListener('d2l-dropdown-open', this._onDropdownOpen.bind(this));
```
dbatiste, line 147: It would also be fine to:
```
this.addEventListener('d2l-dropdown-open', () => this._dropdownOpen = true);
```
Etc.  Fine either way though.

components/list/list-item-generic-layout.js
zinabat: I'm reflecting these two properties with an underscore to denote that they are "private".

Should work in all browsers now.

components/list/list-item-generic-layout.js
zinabat: Good note - I'll probably do this in a different PR

components/list/list-item-generic-layout.js
zinabat: After looking into alternative options, the best thing to do was to reflect an obviously "private" attribute on the host.

PR #679, opened 2020-07-07
Misc Form Validation Fixes

AllanKerr:
When adding form validation I ran into a few minor issues in different places. This PR groups the fixes together.

# Changes
1. Update tooltip `align` so that it is properly reflected
1. Remove `:invalid` selector for `d2l-input` so elements don't appear invalid on first load
1. Add missing warning for default `rangeUnderflow` message
1. Update `tryGetLabelText` to only get child text nodes for implicit labels
1. Make the tooltip touch-start event passive so scroll doesn't block on touch


components/form/form-helper.js
```diff
@@ -6,9 +6,12 @@ export const isCustomFormElement = (node) => isCustomElement(node) && !!node.for
 
 export const tryGetLabelText = (ele) => {
 	if (ele.labels && ele.labels.length > 0) {
-		const labelText = ele.labels[0].textContent.trim();
-		if (labelText) {
-			return labelText;
+		const textNode = [...ele.labels[0].childNodes].find(node => node.nodeType === Node.TEXT_NODE);
+		if (textNode) {
+			const labelText = textNode.wholeText.trim();
```
dlockhart, line 11: `wholeText` doesn't appear to exist [until Safari 13](https://caniuse.com/#feat=mdn-api_text_wholetext)... we're still supporting 10 unfortunately.

AllanKerr: Ouch, I saw IE11 on MDN and thought I'd be good to go, I'll swap it out for doing it manually.

dlockhart: Cool or you could feature detect it and fall back to something else, 🤷

AllanKerr: I just went with `filter` + `reduce`, its not quite as fancy but should be more than sufficient given.

PR #682, opened 2020-07-09
Drag n Drop Mixin to master

m6jones:
I'm going to do a first pass. A lot of this has already been reviewed so hopefully it won't be that bad even though it is giant.


m6jones:
We are thinking to actually pull these changes into `polaris/updateListItem` then make one PR. So closing this up.

PR #683, opened 2020-07-09
Feature: Integrate mixins into list item (Part 2)

zinabat:
## Context
[US118153](https://rally1.rallydev.com/#/detail/userstory/401236562332?fdp=true)

This ties together work done on the `polaris/dragndrop` feature branch into the `polaris/updateListItem` branch.

- Integrates the `ListItemDragDropMixin` into the `ListItem`.
- Fixes a hover state bug that occurs when browsers incorrectly identify targets as hovering
- Style update for drag handle when in a list item
- Adds demo flags to the `d2l-list-demo-drag-n-drop-usage` component
- Adds the demo usage page

## UX Result
Kapture 2020-07-09 at 14 58 25

## Quality
- Existing tests
- Manually tested with Chrome on the `list-drag-n-drop` demo page

### Known Issues
- There is a minor bug when an item is dropped. Occasionally, the drag handle will still show.
- Reordering an item in keyboard mode when pressing down will work, but the drag handle then loses focus. This will be fixed with a directive.


components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,504 @@
+import { css, html } from 'lit-element/lit-element.js';
```
m6jones, line 1: This file was reviewed in detail already in https://github.com/BrightspaceUI/core/pull/665

A good read over would be good though. Nothing bad about double checking.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,504 @@
...
+			case dragActions.first:
+				this._annoucePositionChange(this.key, null, dropLocation.first);
+				break;
+			case dragActions.last:
+				this._annoucePositionChange(this.key, null, dropLocation.last);
+				break;
+			default:
+				break;
+		}
+
```
m6jones, line 152: Extra space

components/list/list-item.js
```diff
@@ -72,6 +70,17 @@ class ListItem extends ListItemCheckboxMixin(ListItemRoleMixin(RtlMixin(LitEleme
...
+			:host([dragging]) d2l-list-item-generic-layout {
+				filter: grayscale(75%);
+				opacity: 0.4;
+			}
+			.d2l-list-item-drag-image {
+				transform: rotate(-1deg);
+			}
+			d2l-list-item-generic-layout {
+				background: white;
+				transform: rotate(1deg);
```
m6jones, line 82: We needed tilt. Tilt that supported images. This, this was the best solution. 
`background: white` is required so the drag image isn't transparent.

components/list/list-item.js
```diff
@@ -219,53 +232,56 @@ class ListItem extends ListItemCheckboxMixin(ListItemRoleMixin(RtlMixin(LitEleme
...
+					class="${classMap(classes)}"
+					data-breakpoint="${this._breakpoint}"
+					data-separators="${ifDefined(this._separators)}"
+					?grid-active="${this.role === 'rowgroup'}">
+					${this._renderDragHandle(this._renderOutsideControl)}
+					${this._renderDraggableArea(this._renderOutsideControlAction)}
+					${this.selectable ? html`
+					<div slot="control">${ this._renderCheckbox() }</div>
+					<div slot="control-action">${ this._renderCheckboxAction('', this._contentId) }</div>
+					` : nothing }
```
m6jones, line 261: In another PR this would be great to clean up and make consistent with `listItemDragDropMixin`

components/list/test/list-item-drag-handle.test.js
```diff
@@ -28,7 +28,7 @@ describe('ListItemDragHandle', () => {
 			element.focus();
 		});
 
-		it(`It dispatch drag handle action event for ${dragActions.active} event when clicked.`, async() => {
+		it(`Dispatch drag handle action event for ${dragActions.active} event when clicked.`, async() => {
```
m6jones, line 31: Maya is special and so this is a fix from misunderstanding how to name tests :)

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,498 @@
...
+export const ListItemDragDropMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			/**
+			 * Whether the item is draggable
+			 */
+			draggable: { type: Boolean, reflect: true },
+			dragging: { type: Boolean, reflect: true,  },
```
dbatiste, line 28: Dangling comma here.

Also, is this meant to be private?  If so, we should prefix is with underscore so it's clear that consumers are not meant to touch it.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	_annoucePositionChange(dragTargetKey, dropTargetKey, dropLocation) {
+		this.dispatchEvent(new CustomEvent('d2l-list-item-position-change', {
+			detail: new NewPositionEventDetails({dragTargetKey, dropTargetKey, dropLocation}),
+			bubbles: true
+		}));
+	}
+
+	_onDragAreaClick(e) {
```
dbatiste, line 114: Should this be named `_onDragTargetClick`?

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,498 @@
...
+				break;
+			default:
+				break;
+		}
+	}
+
+	_onDragOver(e) {
+		if (!this.key) return;
+		const dropSpots = getDragState();
+		dropSpots.updateTime(e.timeStamp);
```
dbatiste, line 157: I think these two lines should be:
```
const dragState = getDragState();
dragState.updateTime(e.timeStamp);
```
... just missed during the rename I think.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,498 @@
...
+		createDragState(this);
+
+		setTimeout(() => {
+			this.dragging = true;
+		});
+	}
+
+	_onDrop() {
+		const dropSpots = getDragState();
+		dropSpots.setActiveDropTarget(this, dropSpots.dropLocation);
```
dbatiste, line 180: Same thing here... `dropSpots` -> `dragState`.  Several more below.

components/list/list-item.js
```diff
@@ -72,6 +70,17 @@ class ListItem extends ListItemCheckboxMixin(ListItemRoleMixin(RtlMixin(LitEleme
 			:host(:last-child) d2l-list-item-generic-layout[data-separators="between"] {
 				border-bottom: 1px solid transparent;
 			}
+			:host([dragging]) d2l-list-item-generic-layout {
```
dbatiste, line 73: Oh hmmm... I guess maybe this does need to remain public.

components/list/list-item-drag-mixin.js
m6jones: It is a read only style one. So making it private would be a good idea. Then set the `attribute: 'dragging'` for it to still work

components/list/list-item.js
m6jones: It is read only. Give the ability to style the left over image when it is being dragged.

PR #684, opened 2020-07-09
Calendar: allow disabled dates to be focusable

margaree:
Scenario:
In `input-date` when a user types a date that is outside of the valid range and then opens the calendar, the "selected" date should appear focused (but not selected). If they arrow key left, right, up, or down towards the direction of a valid date then the next valid date should receive focus.

I did some refactoring, as the logic around the keydown behaviour being jumbled together got confusing, and I thought I could simplify the updating of focus date into one function.


components/calendar/calendar.js
```diff
@@ -531,11 +541,12 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		}
 	}
 
-	async reset() {
+	async reset(allowDisabled) {
```
margaree, line 544: used by `input-date` on date being typed

components/calendar/calendar.js
```diff
@@ -696,100 +727,100 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 			const validDates = this.shadowRoot.querySelectorAll('.d2l-calendar-date:enabled');
 			const focusDate = validDates[latestPossibleFocusDate ? (validDates.length - 1) : 0].parentNode;
 			const year = focusDate.getAttribute('data-year');
 			const month = focusDate.getAttribute('data-month');
 			const date = focusDate.getAttribute('data-date');
 			this._focusDate = new Date(year, month, date);
+			return true;
 		} else {
-			this._focusDate = undefined;
+			return false;
```
margaree, line 816: `_focusDate` should not be set to undefined (other than with the page up/down case, which I moved in there) as this causes problems when all the dates in a month are disabled and the user presses multiple buttons

components/calendar/calendar.js
```diff
@@ -696,100 +727,100 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 		const selectedValueDate = this.selectedValue ? getDateFromISODate(this.selectedValue) : null;
 		const dateElem = selectedValueDate ? await this._getDateElement(selectedValueDate) : null;
-		if (dateElem && isDateInRange(selectedValueDate, getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue))) {
-			this._focusDate = selectedValueDate;
-		} else {
-			await this._updateFocusDateDependentOnDisabled(new Date(this._shownYear, this._shownMonth, 1));
-		}
-	}
-
-	_updateShownMonthDecrease(keyboardTriggered, transitionUpDown) {
```
margaree, line 771: I separated this and `_updateShownMonthIncrease` into separate functions (`_monthIncrease/Decrease` and `_triggerMonthChangeAnimations` so that the arrow key navigation to the next valid date would trigger the animations (even if more than 1 month before/after)

PR #686, opened 2020-07-09
Feature: Update list-item for drag 'n' drop, grid

### LINKED ISSUE #399: z-index of list-item actions obscures dropdown

zinabat:
## Changes
- Adds drag 'n' drop support to `d2l-list-item`
  - Includes new demo page for using drag 'n' drop
- Adds keyboard grid support to `d2l-list-item`
- Adds hover/focus styles to `selectable` list items
- Fixes  #399 

### Drag 'n' drop
Kapture 2020-07-09 at 14 58 25

### Hover/focus states for selectable items
Kapture 2020-06-25 at 16 06 11

### Keyboard grid
Kapture 2020-07-03 at 15 24 36

## Quality
- New unit tests
- New visual diff tests for using `tooltip` and `dropdown` with `list`
- Manually tested with demo pages
  - [x] Chrome
  - [x] Firefox
  - [x] Legacy Edge
  - [ ] CrEdge
  - [x] Safari
- Light manual testing with local LMS instance


components/list/demo/list.html
```diff
@@ -306,6 +311,91 @@ <h2>List (selectable with href)</h2>
...
+							<d2l-list-item-content>
+								<div>Geomorphology and GIS </div>
+								<div slot="secondary">This course explores the geological processes of the Earth's interior and surface. These include volcanism, earthquakes, mountain...</div>
+							</d2l-list-item-content>
+						</d2l-list-item>
+					</d2l-list>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>List (tooltips and dropdown menu)</h2>
```
zinabat, line 344: Demo for a common use case with `d2l-tooltip` and `d2l-dropdown-menu`

components/list/test/list.visual-diff.js
```diff
@@ -15,7 +17,7 @@ describe('d2l-list', () => {
 	});
 
 	beforeEach(async() => {
-		await visualDiff.resetFocus(page);
+		await page.reload();
```
zinabat, line 20: Unsure how much this slows tests down - was the easiest way to restore things.

index.html
```diff
@@ -79,7 +79,13 @@ <h2 class="d2l-heading-3">Components</h2>
 			</ul>
 		</li>
 		<li><a href="components/link/demo/link.html">d2l-link</a></li>
-		<li><a href="components/list/demo/list.html">d2l-list</a></li>
+		<li>
+			List
+			<ul>
+				<li><a href="components/list/demo/list.html">d2l-list</a></li>
+				<li><a href="components/list/demo/list-drag-n-drop.html">d2l-list (drag 'n' drop)</a></li>
```
zinabat, line 86: I think this should be left as a separate demo page, but could be persuaded otherwise

components/list/list-item-generic-layout.js
```diff
@@ -144,12 +156,8 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 	}
 
 	firstUpdated() {
-		this.addEventListener('keydown', this._handleKeydown.bind(this));
+		this.addEventListener('keydown', this._onKeydown.bind(this));
 		this.addEventListener('focusin', this._setFocusInfo.bind(this));
-		this.addEventListener('d2l-dropdown-open', this._onDropdownOpen.bind(this));
-		this.addEventListener('d2l-dropdown-close', this._onDropdownClose.bind(this));
-		this.addEventListener('d2l-tooltip-show', this._onTooltipShow.bind(this));
-		this.addEventListener('d2l-tooltip-hide', this._onTooltipHide.bind(this));
```
zinabat, line 152: These listeners have been moved to the `list-item`. The `postition: relative` on the `list-item` prevents this fix from working. I'm considering putting in the `ListItemRoleMixin` and renaming that mixin back to `ListItemMixin`

m6jones: So in the future if we find we need to make custom list items in HM. I want `ListItemMixin` to be extendable such that it the same as `list-item` but instead of slots there is a render method with `_renderListItem({differentSlots})` They will also need to extend the style too. So that is why I ask for the rename.

components/list/list-item-drag-mixin.js
```diff
@@ -0,0 +1,498 @@
...
+		this._focusingDragHandle = true;
+	}
+
+	_onFocusoutDragHandle() {
+		this._focusingDragHandle = false;
+	}
+
+	_onHostDragEnter(e) {
+		const dragState = getDragState();
+		if (this === dragState.target) {
```
m6jones: This is broken

components/list/list-item-generic-layout.js
m6jones: Yes, I'd be happy with it being named to something like `listItemRequiredMixin` Again, `listItemMixin` simple name to be style based in the future. So when we have other styles they can also be mixins.

components/list/test/list.visual-diff.js
zinabat: I used reload because the `tooltip-helper` didn't have a nifty `reset` method. https://github.com/BrightspaceUI/core/blob/master/components/tooltip/test/tooltip.visual-diff.js#L19

I could write one, but it seems weird to write it within the list tests.

dbatiste: Wow yeah, and quite a few tests as well.  That makes me wonder how much time is actually being spent on the page loads.  Tooltip does have public methods for hiding though... so seems possible to reset.

dbatiste: It'd be nice to find a way to measure the page load time so we can say definitely what it costs us, and whether we should continue to avoid reloading or not.  Don't need to figure it our here/now, but would certainly be good to know going forward.

zinabat: I'll just write one in a separate PR.

PR #688, opened 2020-07-10
List item mixin

m6jones:
Here is my thoughts on the list item mixin. That way if people want to extend the style of the mixin to have an href this is how it will work.


components/list/list-item-mixin.js
```diff
@@ -0,0 +1,357 @@
+import '../colors/colors.js';
```
m6jones, line 1: All the styling goes here.

components/list/list-item.js
```diff
@@ -31,327 +8,20 @@ const defaultBreakpoints = [842, 636, 580, 0];
...
-			href: { type: String },
-			_breakpoint: { type: Number },
-			_dropdownOpen: { type: Boolean, attribute: '_dropdown-open', reflect: true },
-			_hoveringLink: { type: Boolean },
-			_focusing: { type: Boolean },
-			_focusingLink: { type: Boolean },
-			_tooltipShowing: { type: Boolean, attribute: '_tooltip-showing', reflect: true }
-		};
-	}
+class ListItem extends ListItemMixin(LitElement) {
```
m6jones, line 11: This is how they would extend it to make like a `d2l-activity-list-item`

components/list/list-item.js
```diff
@@ -31,327 +8,20 @@ const defaultBreakpoints = [842, 636, 580, 0];
...
-		ro.unobserve(this);
-	}
-
-	firstUpdated(changedProperties) {
-		this.addEventListener('d2l-dropdown-open', () => this._dropdownOpen = true);
-		this.addEventListener('d2l-dropdown-close', () => this._dropdownOpen = false);
-		this.addEventListener('d2l-tooltip-show', () => this._tooltipShowing = true);
-		this.addEventListener('d2l-tooltip-hide', () => this._tooltipShowing = false);
-		super.firstUpdated(changedProperties);
+		return [super.styles];
```
zinabat, line 15: is this actually needed since we aren't adding any styles? The component should just naturally inherit them, right?

components/list/list-item.js
```diff
@@ -31,327 +8,20 @@ const defaultBreakpoints = [842, 636, 580, 0];
...
-		const lastBreakpointIndexToCheck = 3;
-		this.breakpoints.some((breakpoint, index) => {
-			if (width >= breakpoint || index > lastBreakpointIndexToCheck) {
-				this._breakpoint = lastBreakpointIndexToCheck - index - (lastBreakpointIndexToCheck - this.breakpoints.length + 1) * index;
-				return true;
-			}
+		return this._renderListItem({
+			illustration: html`<slot name="illustration"></slot>`,
+			content: html`<slot></slot>`,
+			actions: html`<slot name="actions"></slot>`
```
zinabat, line 22: Cool.

components/list/list-item.js
m6jones: You are totally right. Although this makes for a good example how someone could have styles.

m6jones:
This was a proof of concept. Will be re-doing it since I want to make sure I get all the changes.

PR #689, opened 2020-07-10
Add slot for supporting information

ciss1995:
Context:
US115504 - https://rally1.rallydev.com/#/357252966636d/detail/userstory/384006109564
This PR adds a new slot called supporting information to list-item-content.

However, this PR doesn't strictly follow the slot naming from design. 
Indeed, we should have `1:Primary Text`, `2:Secondary Text`, and `3:Supporting Information`. 
Screen Shot 2020-07-10 at 11 32 04 AM

If we were to apply the above changes, the current implementations of list-item-content that are using `secondary slot` would be affected. To avoid causing confusion, we will keep `secondary` as `supporting information` and add `supporting information` as `secondary`. In a following PR, we will switch the slot names. 
Here is a list of some current usage of `slot="secondary"`:
https://search.d2l.dev/xref/GitHub/Brightspace/d2l.dev/sites/www/dist/index.html?r=1d5644b7
https://search.d2l.dev/xref/GitHub/BrightspaceUI/core/components/list/test/list.visual-diff.html?r=22557f41
https://search.d2l.dev/xref/GitHub/BrightspaceUI/core/components/list/README.md?r=0d53388b
https://search.d2l.dev/xref/GitHub/Brightspace/d2l.dev/sites/github/dist/index.html?r=7ce90665
https://search.d2l.dev/xref/GitHub/BrightspaceUI/core/components/dialog/demo/dialog-async-content.js?r=9ce4b8f7
https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/consistent-evaluation/components/left-panel/consistent-evaluation-submission-item.js?r=48ffa290
https://search.d2l.dev/xref/GitHub/Brightspace/cdo-blog/blog/index.html?r=143b17be
https://search.d2l.dev/xref/GitHub/Brightspace/d2l.dev/sites/github/dist/users/index.html?r=7addb9b0
https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-activity-collection-editor/d2l-activity-collection-editor.js?r=ebb0c324
https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/organizations/components/d2l-organization-admin-list/d2l-organization-admin-list.js?r=7ec38741
https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/enrollments/components/d2l-enrollment-collection-view/d2l-enrollment-collection-view.js?r=99274473#397
https://search.d2l.dev/xref/GitHub/Brightspace/d2l.dev/sites/github/dist/link/index.html?r=e46ab362

This PR:
Screen Shot 2020-07-10 at 11 05 31 AM

Quality:
Tested it on http://localhost:8000/components/list/demo/list.html.


components/list/list-item-content.js
```diff
@@ -26,13 +26,21 @@ class ListItemContent extends LitElement {
...
+			}
+
 		`];
 	}
 
 	render() {
 		return html`
 			<div class="d2l-list-item-content-text d2l-body-compact"><slot></slot></div>
-			<div class="d2l-list-item-content-text-secondary d2l-body-small"><slot name="secondary"></slot></div>
+			<div class="d2l-list-item-content-text-secondary d2l-body-small"><slot name="supportingInformation"></slot></div>
```
zinabat, line 42: I think this slot should be named `supporting-information`  for consistency with other multiple-word slots

components/list/list-item-content.js
```diff
@@ -26,13 +26,21 @@ class ListItemContent extends LitElement {
 				overflow: hidden;
 			}
 
+			.d2l-list-item-content-text-supportingInformation {
```
zinabat, line 29: Similarly, class names are generally kebab case, i.e., `.d2l-list-item-content-text-supporting-information`.

components/list/list-item-content.js
AKobets: what is a trick for structuring the html this way: div with class secondar has slot 'supporting', and div with class supporting has slot secondary ?

AKobets: > what is a trick for structuring the html this way: div with class secondary has slot supporting, and div with class supporting has slot secondary ?

actualy should be reading PR description. it has an explanation

dbatiste: `supporting-info` would be fine too.

components/list/list-item-content.js
```diff
@@ -26,13 +26,21 @@ class ListItemContent extends LitElement {
...
+
 		`];
 	}
 
 	render() {
 		return html`
 			<div class="d2l-list-item-content-text d2l-body-compact"><slot></slot></div>
-			<div class="d2l-list-item-content-text-secondary d2l-body-small"><slot name="secondary"></slot></div>
+			<div class="d2l-list-item-content-text-secondary d2l-body-small"><slot name="supportingInformation"></slot></div>
+			<div class="d2l-list-item-content-text-supportingInformation d2l-body-small"><slot name="secondary"></slot></div>
```
dbatiste, line 43: I am a little confused here though - the class that is being applied here for the secondary slot is `d2l-list-item-content-text-supportingInformation`.  Similarly above.

ciss1995: Yeah so where you see `class="d2l-list-item-content-text-supportingInformation d2l-body-small">` we should in fact have `<slot name="supporting-info">` and not `<slot name="secondary"` like here. We are doing it this way because some projects are using `secondary` instead of `supporting-info` e.g: https://search.d2l.dev/xref/GitHub/Brightspace/d2l.dev/sites/www/dist/index.html?r=1d5644b7#35, which should be `supporting-info`. Basically, design added a new slot between the `Primary` and `Secondary` and changed the slot that was named `secondary` to `supporting info`, and that new slot is now named `secondary`. Now because we want to get our stuff moving and don't want to make this a breaking change, I thought we could do this, and once everyone is ready to switch, we can just switch the slot names here. Lol, I hope that makes sense.

ciss1995: But for consistency, I can have the class name match the slot name and when we switch, we'll just switch the class name as well.

dbatiste: Oh yikes.  We _really_ try to avoid changing slot names because they are part of the public API.  If we need to change the slot name we should really be adding the new slot, going to all the apps that are using the old slot name and update them to the new slot name.  I definitely would not want to see this class/slot naming for very long, since it's going to confuse almost everyone who comes across it, and will likely be subject to being "corrected".

ciss1995: Oh yeah, I completely agree, that was the initial plan but then we didn't want to disturbed other teams and have them test their apps again. But I'll bring up your concern to the team and see if we can reconsider going and updating the apps directly. FYI @zinabat @m6jones

zinabat: We discussed various approaches to this, but I didn't think about the possibility of it being "corrected" by others if it stays for too long with this approach. We should consider adding the new slot and do as Dave suggests

dbatiste: Yeah just spoke with Jenny as well.  Let's try to get this & and apps set up the way we want so we don't have any debt to clean up.  If you need help updating the various places that depend on this, with approvals or what-not, I can probably give a hand.  It should just be a matter of updating to latest core version, updating to the new slot name, and releasing for new bsi.

ciss1995: Cool thanks! Once this is in, I'll start opening PR in the repos that I listed above.

PR #691, opened 2020-07-13
feat: Fix lost of focus when moving down in safari

m6jones:
Fix lost of keyboard mode in chrome
Fix lost of focus when moving up in legacy edge
generic grid uses keyup so we don't get double input

Kapture 2020-07-13 at 15 01 53


components/list/list-item-drag-handle.js
```diff
@@ -169,13 +175,23 @@ class ListItemDragHandle extends LitElement {
...
 		this._dispatchAction(action);
 		e.preventDefault();
-	}
-
-	_onBlur() {
-		this._keyboardActive = false;
-		if (!this._tabbing) {
-			this._dispatchAction(dragActions.save);
+		e.stopPropagation();
+		const cell = findComposedAncestor(this, (parent) =>  parent.hasAttribute && parent.hasAttribute('draggable'));
```
zinabat, line 179: Is this the demo component that holds the `d2l-list`? This is the ancestor of the list that only has this attribute for demo purposes; is it possible to wait for the list component's update instead?

m6jones: It is looking for `d2l-list-item`

PR #693, opened 2020-07-14
input-date-range: initial work adding validation

margaree:
Checking to see if this is on the right track then I'll work on tests and such.


components/inputs/input-date-range.js
```diff
@@ -129,16 +141,44 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-		// TODO: validation that start is before end
 		const elem = e.target;
-		if (elem.classList.contains('d2l-input-date-range-start')) this.startValue = elem.value;
-		else this.endValue = elem.value;
+		if (elem.classList.contains('d2l-input-date-range-start')) {
+			this.startValue = elem.value;
+		} else {
+			this.endValue = elem.value;
+		}
+		await this.validate();
```
AllanKerr, line 168: This should be `requestValidate()` instead of `validate()`

components/inputs/input-date-range.js
```diff
@@ -129,16 +141,44 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+		const errors = await super.validate();
+		if (errors.length !== 0) {
+			return errors;
+		}
+		return Promise.all([
+			await this.shadowRoot.querySelector('.d2l-input-date-range-start').validate(),
+			await this.shadowRoot.querySelector('.d2l-input-date-range-end').validate()]
+		).then((res) => {
+			return res.reduce((acc, errors) => [...acc, ...errors], []);
+		});
```
AllanKerr, line 154: This will likely be okay but we'll have to see how the tooltips end up working. We may only want to validate one or the other to prevent tooltip collisions.

components/inputs/input-date-range.js
```diff
@@ -95,18 +103,21 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-date-range.endDate');
 		return html`
+			<d2l-validation-custom @d2l-validation-custom-validate=${this._validateRange}></d2l-validation-custom>
 			<d2l-input-fieldset label="${ifDefined(this.label)}" ?label-hidden="${this.labelHidden}">
 				<d2l-input-date
 					@change="${this._handleChange}"
 					class="d2l-input-date-range-start"
 					?disabled="${this.disabled}"
-					label="${startLabel}"
+					?invalid="${this.validationError ? true : false}"
```
AllanKerr, line 119: This will likely need to change so that its only passed in if the `validationError` is present. If it isn't present then we don't want to provide the error at all so that the component itself can set `invalid`.

components/inputs/input-date-range.js
```diff
@@ -50,7 +53,9 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			/**
 			 * Value of the start date input
 			 */
-			startValue: { attribute: 'start-value', reflect: true, type: String }
+			startValue: { attribute: 'start-value', reflect: true, type: String },
+			_endLabel: { type: String },
+			_startLabel: { type: String }
```
dlockhart, line 58: Probably want `attribute: false` here so that setting a `_endlabel` attribute on the element doesn't set this.

components/inputs/input-date-range.js
```diff
@@ -95,18 +103,21 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		if (!this.label) {
 			console.warn('d2l-input-date-range component requires label text');
 		}
+
+		this._endLabel = this.endLabel ? this.endLabel : this.localize('components.input-date-range.endDate');
+		this._startLabel = this.startLabel ? this.startLabel : this.localize('components.input-date-range.startDate');
```
dlockhart, line 108: Technically I think these may need to be re-computed in `updated` since any time `end-label` or `start-label` are set you'd need to re-compute. Instead of properties, I think it might be easier to just create private getters for these and then call them from `render()` and `validationMessageBadInput()`.

components/inputs/input-date.js
```diff
@@ -33,6 +33,10 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 			 * Text to reassure users that they can choose not to provide a value in this field (usually not necessary)
 			 */
 			emptyText: { type: String, attribute: 'empty-text'},
+			/**
+			 * Style the component as invalid
+			 */
+			invalid: { type: Boolean, reflect: true },
```
dlockhart, line 39: I'm late to the party here, so this may have been discussed already -- I think Allan's already adding this to attribute to `FormElementMixin`?

components/inputs/input-date-range.js
AllanKerr: Once my PR gets in to add `invalid` and `forceInvalid` you'll be able to change these to:
```
.forceInvalid=${this.invalid}
```

margaree:
I'm going to close and re-open so I can get a better look at the diff and do modifications if needed

PR #695, opened 2020-07-14
Add invalid and force-invalid to FormElementMixin

AllanKerr:
# Changes
- The `invalid` attribute is used to style custom form-elements error state. All error styling should be based on this attribute. Previously this was done using `aria-invalid`.
- The `forceInvalid` property was added so that custom form elements can be forced into an invalid state even if they didn't fail validation. An example use case is in the `input-date-range` component when the start date is later than the end date. Even though neither of the inner `date-range` components failed validation we want to be able to show them in their error state.

### DIFF
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -77,7 +77,11 @@ export class FormElementValidityState {
 export const FormElementMixin = superclass => class extends LocalizeCoreElement(superclass) {
 
 	static get properties() {
-		return { validationError: { attribute: false, type: String } };
+		return {
+			forceInvalid: { type: Boolean, attribute: false },
+			invalid: { type: Boolean, reflect: true },
+			validationError: { type: String, attribute: false },
+		};
 	}
 
 	constructor() {
@@ -85,11 +89,20 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		this._validationCustoms = new Set();
 		this._validationMessage = '';
 		this._validity = new FormElementValidityState({});
+		this.forceInvalid = false;
 		this.formValue = null;
 		this.validationError = null;
 
 		this.addEventListener('d2l-validation-custom-connected', this._validationCustomConnected);
 		this.addEventListener('d2l-validation-custom-disconnected', this._validationCustomDisconnected);
+	}
+
+	updated(changedProperties) {
+		changedProperties.forEach((_, propName) => {
+			if (propName === 'forceInvalid' || propName === 'validationError') {
+				this.invalid = this.forceInvalid || this.validationError !== null;
+			}
+		});
 	}
 
 	checkValidity() {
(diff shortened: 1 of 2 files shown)


components/form/form-element-mixin.js
```diff
@@ -77,21 +77,34 @@ export class FormElementValidityState {
 export const FormElementMixin = superclass => class extends LocalizeCoreElement(superclass) {
 
 	static get properties() {
-		return { validationError: { attribute: false, type: String } };
+		return {
+			forceInvalid: { type: Boolean, attribute: false },
+			invalid: { type: Boolean, reflect: true },
```
dlockhart, line 82: Does `invalid` need to reflect or be bound to an attribute at all? Seems like it's almost more of a computed property -- its value is based on whether there's a validation error or `forceInvalid` is set. So we don't want consumers seeing it being sprouted as an attribute on elements or thinking they should be setting it themselves.

components/form/form-element-mixin.js
```diff
@@ -77,21 +77,34 @@ export class FormElementValidityState {
...
+			validationError: { type: String, attribute: false },
+		};
 	}
 
 	constructor() {
 		super();
 		this._validationCustoms = new Set();
 		this._validationMessage = '';
 		this._validity = new FormElementValidityState({});
+		this.forceInvalid = false;
```
dlockhart, line 92: Does `invalid` need to be defaulted to `false`, or does setting `forceInvalid` here cause the `updated()` call to set it?

components/form/form-element-mixin.js
AllanKerr: The idea was to have some attribute that could be used specifically for styling components when they're in an invalid state. Similar to `:invalid` except with an attribute selector because we can't create our own pseudo-class selector. Forcing people to reflect it themselves becomes a bit awkward.

components/form/form-element-mixin.js
AllanKerr: Yeah, I didn't initialize it because it should get set to its value when the component does its first `updated` call. Since its effectively a computed property I only wanted its value to be set in `updated()`.

components/form/form-element-mixin.js
dlockhart: Ah ok gotcha, that makes sense! `attribute: false` _might_ still make sense then... reflect will cause it to sprout an attribute for CSS selectors, but then `attribute: false` I think will mean that if they manually change that attribute it won't have any effect on the property, which I think is what you want.

components/form/form-element-mixin.js
dlockhart: Not sure if not initializing it in the constructor will mess up @margaree's JSDoc stuff. Alternately, we might need to add `@default false` to the JSDoc comments on the property.

AllanKerr: If it makes other things more difficult we might as well initialize it, it shouldn't matter if its `undefined` or `false` before the first `updated()` call.

components/form/form-element-mixin.js
AllanKerr: Unfortunately `attribute: false` seems to have issues with `reflect`, this is from the [docs](https://lit-element.polymer-project.org/guide/properties#conversion):
>  If attribute is false, the converter, reflect and type options are ignored.

dlockhart: Boo. Ok thanks for investigating!

PR #696, opened 2020-07-15
fix: accessibility for list item dragging

zinabat:
- Fixes the `NewPositionEventDetails.announceMove` function
- Adds a default label to the drag button
- Adds an attribute for users to change this label if they want to

### DIFF
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -27,7 +27,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			 */
 			draggable: { type: Boolean, reflect: true },
 			dragging: { type: Boolean, reflect: true },
-			dragHandleAriaLabel: { type: String, attribute: 'drag-handle-aria-label' },
+			dragHandleText: { type: String, attribute: 'drag-handle-text' },
 			dropText: { type: String, attribute: 'drop-text' },
 			key: { type: String, reflect: true },
 			_draggingOver: { type: Boolean },
@@ -254,7 +254,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			<d2l-list-item-drag-handle
 				id="${this._itemDragId}"
 				class="${classMap(classes)}"
-				text="${ifDefined(this.dragHandleAriaLabel)}"
+				text="${ifDefined(this.dragHandleText)}"
 				@focusin="${this._onFocusinDragHandle}"
 				@focusout="${this._onFocusoutDragHandle}"
 				@d2l-list-item-drag-handle-action="${this._onDragHandleActions}">
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -3,6 +3,7 @@ import { announce } from '../../helpers/announce.js';
 import { classMap } from 'lit-html/directives/class-map.js';
 import { dragActions } from './list-item-drag-handle.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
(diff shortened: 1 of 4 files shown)


components/list/list-item-drag-drop-mixin.js
```diff
@@ -252,6 +254,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			<d2l-list-item-drag-handle
 				id="${this._itemDragId}"
 				class="${classMap(classes)}"
+				text="${this.dragHandleAriaLabel || this.localize('components.list.list-item-drag-handle')}"
```
m6jones, line 257: I would expect that the person using the list item would localize this. Does this make it easier?

zinabat: I figured having a default label was good because in most cases the label should be the same?

zinabat: I decided you were right

m6jones: Am I? My original thought was to have a default too. Didn't read this right. Blaming head ache.

dbatiste: The static text, "Drag Handle" makes sense to me, and I'm not sure it needs to be customizable.  I wouldn't add to the API until we have a really compelling use-case (do we already?).  I think the "button" verbiage can be omitted, because it should already be announced as the element getting focus is an actual `button`.

If we did need to expose this, I would omit "Aria" from the property name and just expose it similar the the handle component's property - "DragHandleText", since Aria is an internal detail of `d2l-list-item-drag-handle`.

m6jones: We have a use case. We will want to give the position which is something unknown. Unless there is a better way to do it.

m6jones: This is what we have in the design: 
image

m6jones: Another change, we will have to make is that the label for the drag handle. 
While it is in non-keyboard mode be the static default value.
When it switches to keyboard mode then we can add the custom so we have the position value.

Does that make sense?

m6jones: I am also happy if they are both static but if they are going to be static maybe just do it in the drag handle?

dbatiste: I see.  Yeah so that has Activity specific verbiage.  Is "Activity current position 1/4" meaningful?  I guess that's a design question.  I'm not sure what "1/4" means... i.e. does it mean position 1 of 4?

Aside from that, I think we should be careful about baking the up/down verbiage into the app-specific terms.  I think that verbiage is best owed by the core component because that's where the interaction is implemented/defined.  I'm all for conveying that info, but if ever change the keyboard interaction we don't want to be having to fix lang-terms in apps.

m6jones: I agree with you

components/list/list-item-drag-handle.js
```diff
@@ -219,7 +225,7 @@ class ListItemDragHandle extends LitElement {
 				@click="${this._onInactiveKeyboard}"
 				@keyup="${this._onInactiveKeyboard}"
 				@keydown="${this._onInactiveKeyDown}"
-				aria-label="${this.text}"
+				aria-label="${this.text || this._defaultLabel}"
```
m6jones, line 228: Remove `text` from this one?

components/list/list-item-drag-drop-mixin.js
zinabat: I've renamed the attribute, and also changed the default text to `Reorder action` and `Reorder item`

PR #697, opened 2020-07-15
fix sauce flakiness 

margaree:


margaree:
This seems to fix things. I think the `browserDisconnectTolerance : 2` portion was what fixed things the most, as legacy Edge seems to now consistently have 2 disconnects before getting through all the tests.

dlockhart:
Nice, good work @margaree!

PR #698, opened 2020-07-15
Add showErrors to FormElementMixin validate

AllanKerr:
# Changes
- This allows validate to be called during live validation to remove errors if the component becomes valid but not show new errors if the component goes from a valid to an invalid state.
- This is required because our designs for live validation only have new errors shown on focus out.

## Example
- For a `d2l-input-text` the usage would look something like this with `requestValidate(false)` when the text value changes and `requestValidate(true)` on blur.
```
updated(changedProperties) {

		if (prop === 'value') {
			this._prevValue = (oldVal === undefined) ? '' : oldVal;
			this.requestValidate(false); // don't show new errors but clear existing errors if we go from invalid to valid
		}
	});
}

_handleBlur(e) {
	this._focused = false;

	const browserType = window.navigator.userAgent;
	if (this._prevValue !== e.target.value && (browserType.indexOf('Trident') > -1 || browserType.indexOf('Edge') > -1)) {
		this._handleChange();
	}
	this.requestValidate(true); // show errors and enter an invalid state if validation fails 
}

```

### DIFF
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -123,7 +123,7 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this.localize('components.form-element-mixin.defaultFieldLabel');
 	}
 
-	async requestValidate(showErrors) {
+	async requestValidate(showErrors = true) {
 		if (this.dispatchEvent(new CustomEvent('d2l-form-element-should-validate', { cancelable: true }))) {
 			await this.validate(showErrors);
 		}
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -123,9 +123,9 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this.localize('components.form-element-mixin.defaultFieldLabel');
 	}
 
-	async requestValidate() {
+	async requestValidate(showErrors) {
 		if (this.dispatchEvent(new CustomEvent('d2l-form-element-should-validate', { cancelable: true }))) {
-			await this.validate();
+			await this.validate(showErrors);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


components/form/form-element-mixin.js
```diff
@@ -123,9 +123,9 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this.localize('components.form-element-mixin.defaultFieldLabel');
 	}
 
-	async requestValidate() {
+	async requestValidate(showErrors) {
```
AllanKerr, line 126: I'm not a huge fan of `showErrors` so if anyone has any better names let me know or if we want this to default to `true` rather than `false`.

margaree: Not sure on the name but I'm thinking we may want to default to `true` instead of `false` to avoid confusion around why errors aren't working (which may have just happened to me when I was trying this out and it took me a bit to figure out where I was missing a true)

dlockhart: Yeah I'm not thinking of anything better. You could reverse it to be something like `suppressErrors` or `hideErrors`... not sure that's an improvement. `displayErrors`, `visualizeErrors`, `errorsAreShown`... haha.

AllanKerr: Sounds good to me, in that case I'll leave the name as is but make it default to `true`.

AllanKerr: Changed `showErrors` to default to `true` but only for `requestValidate()` since `validate()` should never be called manually. Only thing to watch out for is that you don't lose the `showErrors` value when you override `validate()`.

components/form/form-element-mixin.js
```diff
@@ -123,9 +123,9 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this.localize('components.form-element-mixin.defaultFieldLabel');
 	}
 
-	async requestValidate() {
+	async requestValidate(showErrors = true) {
```
svanherk, line 126: Woohoo for no IE11! 🎉

PR #699, opened 2020-07-16
fix: Click Focus, Click Keyboard

m6jones:
Abbie: Re: Clicking the drag handle -- I’m thinking we should only make it change to keyboard on click if it’s already focused. I don’t want mouse users to get disoriented if they click right on the drag handle the first time.

Kapture 2020-07-16 at 2 55 37

### DIFF
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -34,6 +34,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			_dropLocation: { type: Number },
 			_focusingDragHandle: { type: Boolean },
 			_hovering: { type: Boolean },
+			_isActivitingKeyboardMode: {type: Boolean },
 			_keyboardActive: { type: Boolean }
 		};
 	}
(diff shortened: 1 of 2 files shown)


components/list/list-item-drag-drop-mixin.js
```diff
@@ -171,8 +171,19 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 	}
 
 	_onDragTargetClick(e) {
-		this.shadowRoot.querySelector(`#${this._itemDragId}`).activateKeyboardMode();
+		if (this._keyboardActiveOnNextClick) {
+			this.shadowRoot.querySelector(`#${this._itemDragId}`).activateKeyboardMode();
+		} else {
+			this.shadowRoot.querySelector(`#${this._itemDragId}`).focus();
```
dbatiste, line 177: What is the user is trying to click on another focusable element in the list-item?  Seems like this would style focus away unexpectedly.

AKobets:
Is this PR affects something that is used outside of learning path scope ?

AKobets:
Does this PR affect something that is used outside of learning path scope ?

components/list/list-item-drag-drop-mixin.js
zinabat: Other focusable items will always appear in front because of the layered grid

PR #702, opened 2020-07-16
Fix validation-custom connection events

AllanKerr:
## Issue
In browsers that use ShadyDOM events don't bubble correctly for slotted descendants: https://github.com/Polymer/lit-element/issues/658

## Fix
This resolves the issue by firing the events from a proxy node that bypasses ShadyDOM and adding the
element that dispatched the event to the detail.

A one off fix was used instead of creating a helper because there is some hope this might be fixed in the near future: https://github.com/webcomponents/polyfills/pull/332


AllanKerr:
Looks like a few more issues need to be resolved, closing to rethink this slightly.

PR #705, opened 2020-07-16
fix: Make drag handle appear always for mobile

zinabat:
- Adds a media query that makes the drag handle hover effect only occur on desktop


components/list/list-item-drag-drop-mixin.js
```diff
@@ -73,12 +73,14 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 				width: 100%;
 				z-index: 100;
 			}
-			d2l-list-item-drag-handle {
-				opacity: 0;
-			}
-			d2l-list-item-drag-handle.d2l-hovering,
-			d2l-list-item-drag-handle.d2l-focusing {
-				opacity: 1;
+			@media screen and (min-device-width: 615px) {
```
dbatiste, line 76: I assume that what we are _really_ interested in is touch device, or rather, devices that do not have the concept of "hover".  Screen width is not a great query to use in such case.  The way we handled this for `VisibleOnAncestorMixin`

```
@media only screen and (hover: none), only screen and (-moz-touch-enabled: 1) {
   /* show it */
}
```

zinabat: oooh, neat, I did not know you could query for hover capability.

zinabat: I have used the same approach, except this media query targets only hover-capable and touch-disabled devices, since the styles are only applicable to desktop/hoverable screens.

dbatiste: Did you happen to see... https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-moz-touch-enabled?

zinabat: I'm just now seeing `pointer: coarse` as a "replacement" for `-moz-touch-enabled: 1`. Perhaps this query should really be targetting `pointer: fine` along with `hover: hover`.

dbatiste: Looks like FF supports [hover: none](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover) as of v64.  I _think_ the `-moz-touch-enabled` would only really be needed for versions before that, and as of v73, it is no longer supported.

zinabat: We'll be testing this more thoroughly with real devices shortly, but for now I have left in `pointer: fine`, which is probably overkill but 🤷

dbatiste: Yeah that's fine.  Take care though - the original code (for visible=on-ancestor) only relied on the media-query to adapt the behavior for the case where hovering was not possible.  The current code here flips the logic around, and relies on the media query for hover-enabled devices, which is probably almost all of our usage.  That is why the fact that IE didn't support that query didn't matter - no one would really be using IE on a mobile device.

PR #706, opened 2020-07-16
allowing primary panel's overflow to be hidden

dlockhart:
This change is to support a use case in consistent-evaluation where various legacy viewers are embedded in the primary panel via a `<iframe>`. When the iframe becomes scrollable, scrollbars appear for both the iframe and the primary panel.

In this scenario, the ideal solution is for the iframe's height to correspond to the height of the primary panel, and for the panel's scrolling to be disabled -- since the iframe itself will handle scrolling.

@jeroach, I tested this out with an iframe like this (note the new `primary-overflow="hidden"` attribute):

```html
<d2l-template-primary-secondary primary-overflow="hidden" width-type="fullscreen">
    <div slot="header">Header</div>
    <iframe slot="primary" src="some-page.html" style="border: none; height: 100%; width: 100%;"></iframe>
    <p slot="secondary">Secondary</p>
    <div slot="footer">Footer</div>
</d2l-template-primary-secondary>
```

It's cool that you can drop the iframe right into the slot like that! Make sure you kill the border on the iframe though otherwise you get a bit of a weird gap.


templates/primary-secondary/primary-secondary.js
```diff
@@ -12,9 +12,14 @@ class TemplatePrimarySecondary extends LitElement {
...
 	static get properties() {
 		return {
+			/**
+			 * Controls how the primary panel's contents overflow
+			 * @type {'default'|'hidden'}
+			 */
+			primaryOverflow: { attribute: 'primary-overflow', reflect: true, type: String },
 			/**
 			 * Whether content fills the screen or not
-			 * @type {('fullscreen'|'normal')}
```
margaree, line 17: Thanks for catching this!

PR #707, opened 2020-07-17
fix: Menu + dropdown menu: keypress to keyup and mouseup to click

m6jones:
I ran into problems with a drop down menu in the list component. Basically, the list component does things on `keyup` which is still registered for double action on the key.

I also noticed dropdown was using `onMouseUp` and changed it to the newer `click`.

I replaced `keypress` with `keyup` since keypress is depreciated.

I don't know the risks of this change. As there could be good reasons for the choices I changed away from. If that is the case, I am happy to find another solution. 

It would be amazing if this could be a cert fix.

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -1,3 +1,16 @@
+const keyCodes = Object.freeze({
+	DOWN: 40,
+	END: 35,
+	ENTER: 13,
+	HOME: 36,
+	LEFT: 37,
+	PAGEUP: 33,
+	PAGEDOWN: 34,
+	RIGHT: 39,
+	SPACE: 32,
+	UP: 38
+});
+
 export const DropdownOpenerMixin = superclass => class extends superclass {
 
 	static get properties() {
@@ -34,8 +47,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this.noAutoOpen = false;
 		this.disabled = false;
 
-		this.__onKeyPress = this.__onKeyPress.bind(this);
-		this.__onMouseUp = this.__onMouseUp.bind(this);
+		this.__onKeyUp = this.__onKeyUp.bind(this);
+		this.__onClick = this.__onClick.bind(this);
 	}
 
 	connectedCallback() {
@@ -48,8 +61,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 				return;
 			}
 			opener.setAttribute('aria-haspopup', 'true');
-			opener.addEventListener('keypress', this.__onKeyPress);
-			opener.addEventListener('mouseup', this.__onMouseUp);
+			opener.addEventListener('keyup', this.__onKeyUp);
+			opener.addEventListener('click', this.__onClick);
 			opener.setAttribute('aria-expanded', (content && content.opened || false).toString());
 		});
 	}
@@ -60,8 +73,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		if (!opener) {
 			return;
 		}
-		opener.removeEventListener('keypress', this.__onKeyPress);
-		opener.removeEventListener('mouseup', this.__onMouseUp);
+		opener.addEventListener('keyup', this.__onKeyUp);
+		opener.addEventListener('click', this.__onClick);
 	}
 
 	firstUpdated(changedProperties) {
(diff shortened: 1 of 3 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -112,15 +130,18 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
...
 		if (this.noAutoOpen) return;
-		this.toggleOpen(true);
-	}
-
-	__onMouseUp() {
-		if (this.noAutoOpen) return;
-		this.toggleOpen(false);
+		switch (e.keyCode) {
+			case keyCodes.ENTER:
+			case keyCodes.SPACE:
```
m6jones, line 137: I also fixed an accessibility issue!

components/menu/menu.js
```diff
@@ -259,17 +259,18 @@ class Menu extends HierarchicalViewMixin(LitElement) {
...
-		if (e.keyCode === keyCodes.DOWN || e.keyCode === keyCodes.UP
-			|| e.keyCode === keyCodes.SPACE || e.keyCode === keyCodes.ENTER
+		if (e.keyCode === keyCodes.SPACE || e.keyCode === keyCodes.ENTER
 			|| e.keyCode === keyCodes.ESCAPE) {
 			return;
 		}
 
 		e.stopPropagation();
 
+		if (e.keyCode === keyCodes.DOWN || e.keyCode === keyCodes.UP) return;
```
m6jones, line 272: We don't want these escaping the menu do we?

components/dropdown/dropdown-opener-mixin.js
dbatiste: Thanks Maya.  Instead of a switch, I'd be tempted to just use:
```
if (e.keyCode !== keyCodes.ENTER && e.keyCode !== keyCodes.SPACE) return;
this.toggleOpen(true);
...
```
I assume preventing default is for the Space key, which would normally scroll the page?  What's the reason doe stopping the propagation?

components/menu/menu.js
dbatiste: What do you mean by "escaping the menu"?  What behavior does this change/fix?

dbatiste:
Yeah, there could certainly be "bugs" in the original, but I'm also fairly certain almost all of it was written intentionally, and we have sort of a "special" keyboard/mouse behavior as far as screen readers are concerned... which certainly makes me a little cautious here.

components/menu/menu.js
m6jones: Right now, when you press up and down in a drop down while in an aria-grid. Such as lists the keys leak out to the list and move focus away from the drop down and into the list item.

dbatiste: I see.  That's a little concerning.  We can,'t always expect components to stop propagation for that scenario.  Like we can do it here for this component, but what else might we have to deal with?  These events naturally bubble and usually it would be the responsibility of the handlers higher up to check target, etc.

m6jones: That is a solution. How would the list know what target uses what keys though?

m6jones:
The other thing we need to standardize is how we use `keydown` and `keyup`. Hopefully, upgrade to no longer use `keypress` (It is deprecated and this comment is totally based on that fact so I could totally be missing the boat on why we should keep it around) 

From comments in updating the list I think we are agreeing to use `keyup` for functionality. Making sure everything follows that is important so we don't get an action on a down that moves you into a new state that will trigger the up action in that state.

components/menu/menu.js
dbatiste: The only alternative that I can think of is that the list-grid would have to be aware of which types of elements (including checking aria role) to ignore.  Also not amazing.

I think in this case we can probably safely stop the propagation, since the menu is providing some very specific behavior and it's unlikely that parent components would be interested in those events.  I wouldn't say that about all events though, or all key codes, which is what makes me pause here.

dbatiste:
Re `keydown` vs `keyup` - yeah agreed that we should be consistent.  I think we've basically standardized on `keyup`.  This is especially true for things like `setDismissible`.  The `keydown` event is still needed to cancel certain default behavior though.

I think I wrote this code 3 or 4 years years ago so I don't recall all the specifics, but I think `keypress` was used to handle the case where the user can press a key to move focus.  We only very recently changed this logic to be case-insensitive.  This was also very recently migrated to Lit, and these events were not changed, possibly intentionally.

I wouldn't mind taking a comprehensive look at the use of key events, `keyCode`, `charCode`, and `key` and aligning where possible.

m6jones:
More thought and discussion to follow.

PR #708, opened 2020-07-17
fix: Wire touch events for drag and drop on mobile devices

zinabat:
## Context
So, fun fact: `DragEvent` [is not a thing](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent#Browser_compatibility) for most **mobile browsers** with the exception of Chrome and Firefox on newer phones. This pull request hooks into the `TouchEvent` API to synthetically reproduce `drag` events on phones.

- `touchstart`: Waits for the user to hold for a moment before `dragstart` happens
- `touchmove`: Fires on the `drag-target` being moved (aka _touched_). We dig into the DOM of the `list` in order to detect what we're dragging over. Depending on where we are, it will fire:
  - `dragover`: on the `drop-grid` we are dragging over
  - `dragenter`: on list items and `drop-grid` areas
- `touchend`: Will fire a `drop` if we're over a `drop-grid`, followed by a `dragend` on the `drag-target`
- `touchcancel`: Happens in various ways on devices when a touch is cancelled, such as by a flick

A custom solution was used because currently, this is the only known place where native drag events exist.

**Usability Notes**
- Users on mobile touch devices can **scroll normally** as long as they don't hold down on a draggable item.
- Because of this new feature, holding on an item will **not allow users to open their context menu** (i.e., copy, inspect, etc.). This is the same expectation for phones with native `DragEvent` support
- Users can also **scroll if they drag** the item to the bottom or top of the viewport.
- The **drag image does not show** for mobile phones with this PR - that is work that will likely come later. Because of the placement line, this isn't needed for usability.

## UX Result
Kapture 2020-07-16 at 21 03 25

## Quality
- Manually tested on Chrome with the Device toolbar
- Manually tested on iOS with an LMS build that includes this change

**Important Testing Note**
Because this assumes that the `ondragenter` property is not a function for browsers that don't support it, the context menu will appear in the Chrome device window if you hold on an item for too long without dragging. This _shouldn't_ occur on phones, but just something to keep in mind for testing.


components/list/list-item-drag-drop-mixin.js
```diff
@@ -111,6 +130,17 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+	_findListItemFromCoordinates(x, y) {
+		const listNode = findComposedAncestor(this.parentNode, (node) => node && node.tagName === 'D2L-LIST');
+		return listNode.shadowRoot.elementFromPoint(x, y);
+	}
+
+	_onContextMenu(e) {
+		if (isDragSupported()) return;
+		e.preventDefault();
+		e.stopPropagation();
+	}
```
zinabat, line 142: Separate fun fact. Did you know `contextmenu` is an event??? 🍪

PR #709, opened 2020-07-17
Fix lit analyzer errors

AllanKerr:
# Changes
The two culprits were `no-unknown-attribute` and `no-incompatible-type-binding`.
- `no-unknown-attribute` errored because we marked attributes as `@ignore` so they wouldn't be documented but apparently Lit Analyzer doesn't like that because it also then thinks they don't exist.
- `no-incompatible-type-binding` errored because we're passing values Lit Analyzer thinks are strings into what is effectively an enum type. Having to add comment types in these cases feels pretty bad so not sure if anyone has any thoughts on disabling this rule.

### DIFF
--- a/components/inputs/input-time-range.js
+++ b/components/inputs/input-time-range.js
@@ -103,6 +103,11 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 	render() {
 		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-time-range.startTime');
 		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-time-range.endTime');
+
+		/**
+		 * @type {'five'|'ten'|'fifteen'|'twenty'|'thirty'|'sixty'}
+		 */
+		const timeInterval = this.timeInterval;
 		return html`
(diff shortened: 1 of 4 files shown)


components/inputs/input-time-range.js
```diff
@@ -103,6 +103,11 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 	render() {
 		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-time-range.startTime');
 		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-time-range.endTime');
+
+		/**
+		 * @type {'five'|'ten'|'fifteen'|'twenty'|'thirty'|'sixty'}
+		 */
```
dlockhart, line 109: Yeah in other places I've created something like a `_getTimeInterval()` that has a `switch` statement and that seems to work as well. Lots of code for no reason though.

zinabat:
I would be pro turning off `no-incompatible-type-binding` because it's also the reason we are now forced to use `boolean-attribute="true"` instead of the more succinct `boolean-attribute`

PR #712, opened 2020-07-18
US116586 Sticky footer for primary-secondary-template on mobile

emil-furniss:
On mobile when the secondary panel is repositioned below the primary panel, it can be quite a long scroll to the footer. This change makes the footer sticky on mobile devices.

Tested on Safari & Chrome on iOS, and Chrome on Android.

Demo using the FACE assignment editor:
prim-sec-template-sticky-footer

https://rally1.rallydev.com/#/29180338367d/iterationstatus?detail=%2Fuserstory%2F391269578948


dlockhart:
I pushed a change to the white background colour to apply it always -- also `--d2l-color-white` isn't a variable anymore so that wasn't working.

Is the header supposed to always be visible too? It looks like it is in your screenshot animation but I'm not seeing that behaviour in the demo.

emil-furniss:
@dlockhart Thanks for the colour fix - sorry I should have checked the demo more thoroughly.
And regarding the header, according to design the ultimate goal for the template header on mobile is a "snap back approach": It goes away when you scroll down, but as soon as the user scrolls up again it shows up. The current sticky header behavior is just a side-effect of using the immersive nav.

dlockhart:
Ah OK gotcha. So I'm good with this change, although ultimately I think when we implement more of the mobile panel stuff this may need to be reworked. But we'll cross that bridge when we come to it.

PR #713, opened 2020-07-20
fix: Drag and drop Safari and Firefox redirects to "undefined"

zinabat:
closes https://rally1.rallydev.com/#/detail/defect/405152852372

[DE39813](https://rally1.rallydev.com/#/detail/defect/405152852372?fdp=true)

- Fixes a bug in some versions of Safari and Firefox that redirects users to "undefined"


components/list/list-item-drag-drop-mixin.js
```diff
@@ -158,8 +158,10 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 	}
 
 	_onDragStart(e) {
-		e.dataTransfer.setData('text/plain', `${this.dropText}`);
 		e.dataTransfer.effectAllowed = 'move';
+		if (e.dataTransfer.setData && this.dropText) {
+			e.dataTransfer.setData('text/plain', `${this.dropText}`);
```
AKobets, line 163: what if condition is evaluated to false, how it will affect user experience and drag/drop functionality?

m6jones: It's an extra neat thing I put in. At the end of the day it can be removed.

PR #714, opened 2020-07-20
input-date-range: validation

margaree:
- `input-date` contains a tooltip which shows the out of range (i.e., min and max) errors
- `input-date-range` contains tooltips which show the bad input errors (i.e., end date before start date)


components/inputs/input-date-range.js
```diff
@@ -129,16 +149,71 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		if (input) input.focus();
 	}
 
+	async validate(showErrors) {
+		const errors = await super.validate(showErrors);
+		if (errors.length !== 0) {
+			return errors;
+		}
+		setTimeout(async() => {
```
margaree, line 157: I added this `setTimeout` because when going from an invalid range to just an invalid input-date, I was getting this error:
```
run-async.js:50 Uncaught (in promise) Error
    at run-async.js:50
    at NodePart.commit (parts.ts:210)
    at TemplateInstance.update (template-instance.ts:53)
    at NodePart.__commitTemplateResult (parts.ts:271)
    at NodePart.commit (parts.ts:221)
    at render (render.ts:55)
    at Function.render (shady-render.ts:277)
    at HTMLElement.update (lit-element.ts:216)
    at HTMLElement.performUpdate (updating-element.ts:757)
    at HTMLElement._enqueueUpdate (updating-element.ts:713)
```

It still functioned properly, but something about the timing seemed right. `setTimeout` was the only thing I had found that "fixed" it.

components/inputs/input-date-range.js
```diff
@@ -98,24 +105,37 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 		return html`
+			${tooltipStart}
+			${tooltipEnd}
+			<d2l-validation-custom @d2l-validation-custom-validate=${this._validateRange}></d2l-validation-custom>
 			<d2l-input-fieldset label="${ifDefined(this.label)}" ?label-hidden="${this.labelHidden}">
 				<d2l-input-date
 					@change="${this._handleChange}"
 					class="d2l-input-date-range-start"
+					@d2l-form-element-should-validate="${this._handleNestedFormElementValidation}"
+					@d2l-input-date-dropdown-open="${this._handleDropdownOpen}"
```
margaree, line 119: The events coming from `d2l-input-date` were necessary in order to have the same behaviour where the tooltip hides when the dropdown opens, and the tooltip appears when the dropdown closes.

components/inputs/input-date-range.js
```diff
@@ -98,24 +105,37 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 
 	render() {
-		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-date-range.startDate');
-		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-date-range.endDate');
+		const tooltipStart = this.validationError ? html`<d2l-tooltip align="start" for="${this._startInputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
+		const tooltipEnd = this.validationError ? html`<d2l-tooltip align="start" for="${this._endInputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
 		return html`
+			${tooltipStart}
+			${tooltipEnd}
+			<d2l-validation-custom @d2l-validation-custom-validate=${this._validateRange}></d2l-validation-custom>
```
AllanKerr, line 113: We should be able to get rid of this `d2l-validation-custom` and in `_handleChange` we'd then call:
```
this.setValidity(...);
await this.requestValidate();
```

components/inputs/input-date-range.js
```diff
@@ -129,16 +149,71 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 			'change',
 			{ bubbles: true, composed: false }
 		));
 	}
 
+	_handleDropdownOpen(e) {
+		const id = e.target.id;
+		const tooltip = this.shadowRoot.querySelector(`d2l-tooltip[for="${id}"]`);
+		if (tooltip && tooltip.showing) tooltip.hide();
+	}
```
AllanKerr, line 193: We might be able to replace this with tooltip `close-on-click`. It was introduced so that when a button gets interacted with to open a dropdown the tooltip will close.

components/inputs/input-date-range.js
```diff
@@ -129,16 +149,71 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+		if (tooltip && !tooltip.showing) tooltip.show();
+	}
+
+	_handleNestedFormElementValidation(e) {
+		e.preventDefault();
+	}
+
+	get _startLabel() {
+		return this.startLabel ? this.startLabel : this.localize('components.input-date-range.startDate');
+	}
```
AllanKerr, line 208: I assume this isn't inlined in render for performance reasons? If the localize mixin doesn't automatically cache terms is this something we should consider adding? It feels a bit weird to have components be responsible for caching their own lang terms.

components/inputs/input-date-range.js
```diff
@@ -129,16 +149,71 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+		const tooltip = this.shadowRoot.querySelector(`d2l-tooltip[for="${id}"]`);
+		if (tooltip && tooltip.showing) tooltip.hide();
+	}
+
+	_handleInnerFocus(e) {
+		// in order for calendar close to cause tooltip to show
+		const id = e.target.id;
+		const tooltip = this.shadowRoot.querySelector(`d2l-tooltip[for="${id}"]`);
+		if (tooltip && !tooltip.showing) tooltip.show();
+	}
```
AllanKerr, line 200: I assume in this case the tooltip's target already has focus but the tooltip doesn't open because it got focus before the tooltip was added? If thats the case this feels like something that should be fixed in `d2l-tooltip` so it starts opened when added to a target that already has focus or hover.

components/inputs/input-date-range.js
AllanKerr, line 157: Why is `setTimeout` needed here?

components/inputs/input-date.js
```diff
@@ -309,10 +315,16 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 	_handleDropdownOpen() {
 		if (!this._dropdown.openedAbove) this.shadowRoot.querySelector('d2l-focus-trap').scrollIntoView({block: 'nearest', behavior: 'smooth', inline: 'nearest'});
 		// use setTimeout to wait for keyboard to open on mobile devices
+		const tooltip = this.shadowRoot.querySelector('d2l-tooltip');
+		if (tooltip && tooltip.showing) tooltip.hide();
```
AllanKerr, line 319: You might also be able to replace this with tooltip `close-on-click`.

components/inputs/input-date-range.js
AllanKerr: I'm not entirely sure what the root cause is but it seems to be triggered by the icon change in `d2l-input-date`. I believe its an issue with the `runAsync` directive.

You can remove the need for `setTimeout` by changing it from this:
```
<d2l-icon icon="${this.invalid ? 'tier1:alert' : 'tier1:calendar'}" slot="left" style="${styleMap({color: this.invalid ? 'var(--d2l-color-cinnabar)' : ''})}"></d2l-icon>
```
to this:
```
${this.invalid
	: html`<d2l-icon icon="tier1:calendar" slot="left"></d2l-icon>`
}
```

margaree: Thanks for figuring that out!

components/inputs/input-date-range.js
margaree: So the tooltip's target is the `input-date` but when the calendar closes, the focus moves to the `d2l-input-text` within the `input-date` which does not trigger focus up to the `input-date`

AllanKerr: I do think we can likely make this work with tooltip automatically since the `activeElement` will be reported as the `input-date` even if its the `input-text` with focus. Is this code specifically to hand the case where they click a date triggering an invalid range and the calendar closes or are there other cases it handles?

margaree: Okay yeah if we can handle that without this then that would be good.

Yup, basically, it's for the case where the end date is before start date and then the calendar is open and then closes (could be that the selected dates were bad before opening and they did not fix it while the calendar was open, and so it's still a bad input when the calendar closes).

AllanKerr: This _should_ now automatically be handled by tooltip, let me know if you run into any issues.

components/inputs/input-date-range.js
margaree: So this kind of works, but it does only close on click and not on enter or down arrow.

components/inputs/input-date-range.js
margaree: It was suggested here https://github.com/BrightspaceUI/core/pull/693#discussion_r455058275, more for simplicity than performance. I think localize mixin does some sort of caching of the lang terms but I haven't been in that code for a bit so I don't exactly recall - will take a look.

components/inputs/input-date-range.js
AllanKerr: Thats unfortunate, it would be really nice to be able to handle it without an event. We could add something like `close-on-keypress=[key-code-1, ..., key-code-n]` but not sure whether that's something we'd want in tooltip.

components/inputs/input-date-range.js
AllanKerr: I'm fine leaving it as is if it benefits simplicity. But it would be nice to remove the `startLabel` property if `localize` is caching it for us so we can just `return this.localize('components.input-date-range.startDate');`

margaree: So `startLabel` is for if the consumer of `input-date-range` wants to specify what the start label is (e.g., instead of "Start Date" it's "Begin on.." or something like that) whereas `_startLabel` is the internal usage of that, so I'm thinking I should rename `_startLabel`

AllanKerr: Sorry about that, missed that there was a corresponding public `startLabel` property for users of the component. That makes sense to me!

margaree:
The rebasing of this has become a mess so I think I'm going to just close this again and open a fresh PR off master.

PR #715, opened 2020-07-20
Dbatiste/fullscreen dialog

dbatiste:
`d2l-dialog-fullscreen` variant of the generic `d2l-dialog`.  I made this a separate component as to not overload the API since `width` property does not really apply for fullscreen.  This leaves open the opportunity to change layout more significantly if needed, without affecting `d2l-dialog`.

dialog-fullscreen

Also reorganized the demo pages a little, and renamed a couple visual-diff tests for consistency.


components/dialog/README.md
```diff
@@ -45,24 +45,20 @@ document.querySelector('#open').addEventListener('click', () => {
   });
 });
 ```
-Make sure that in the 'd2l-dialog-close' event handler that you sync any bound properties if you decided to use them:
+
+*Important:* The user may close the dialog in a few different ways: clicking the dialog workflow buttons (marked up with `data-dialog-action`); clicking the `[x]` button in the top-right corner; or pressing the `escape` key. Therefore, if your component tracks the `opened` state of the dialog with its own property, it's important to keep it in sync by listening for the `d2l-dialog-close` event.
```
dbatiste, line 49: Simplified this by combining with the note that followed.  Hopefully it's clearer.

components/dialog/demo/dialog-confirm.html
```diff
@@ -0,0 +1,64 @@
+<!DOCTYPE html>
```
dbatiste, line 1: This is same demo as before, I just moved it into its own file since I was creating a new page for fullscreen dialog anyway.

components/dialog/test/dialog.visual-diff.js
```diff
@@ -48,19 +48,6 @@ describe('d2l-dialog', () => {
 						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 					});
 
-					it('scroll bottom shadow', async function() {
```
dbatiste, line 51: These tests did not need to be repeated for both wide & narrow cases.  Trying to optimize.

components/dialog/dialog-mixin.js
```diff
@@ -18,6 +18,8 @@ if (window.D2L.DialogMixin.preferNative === undefined) {
 	window.D2L.DialogMixin.preferNative = true;
 }
 
+window.D2L.DialogMixin.preferNative = false;
```
dlockhart, line 21: Accident?

components/dialog/dialog-mixin.js
```diff
@@ -35,6 +37,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			 * The optional title for the dialog
 			 */
 			titleText: { type: String, attribute: 'title-text' },
+			_autoSize: { type: Boolean },
```
dlockhart, line 40: Probably want `attribute: false` so that nothing gets reflected back & forth?

components/dialog/test/dialog-fullscreen.test.js
```diff
@@ -0,0 +1,14 @@
+import '../dialog.js';
```
dlockhart, line 1: Should this be importing `dialog-fullscreen.js` and `runConstructor('d2l-dialog-fullscreen')`?

components/dialog/dialog-mixin.js
dbatiste: Thanks... yep.  😵

components/dialog/dialog-mixin.js
dbatiste: It should not get reflected, but yeah, we're not expecting consumer to set this.  Will add.

components/dialog/test/dialog-fullscreen.test.js
dbatiste: Yikes, yes.  Thanks.  Too much copy pasty.

PR #716, opened 2020-07-21
Tooltip will now start open if added to a target that already has focus

AllanKerr:
# Changes
- Now if a tooltip is added to a target that already has focus the tooltip will start opened. [This change was made to address this use case](https://github.com/BrightspaceUI/core/pull/714#discussion_r457619606).


components/tooltip/tooltip.js
```diff
@@ -812,6 +812,8 @@ class Tooltip extends RtlMixin(LitElement) {
 
 		if (this.showing) {
 			this.updatePosition();
+		} else if (this.getRootNode().activeElement === target) {
```
dlockhart, line 815: There's also `getComposedActiveElement` is the focus helpers.

AllanKerr: `getComposedActiveElement` is slightly different since it continually traverses down the DOM until it finds the actual element that has focus. In this case I only want the active element within the context of the tooltip and target's root node.

dlockhart: Ah gotcha makes sense to use a quicker short-circuit. 👍

PR #717, opened 2020-07-21
Add check-manual icon

margaree:


dbatiste:
The `catalogue.md` needs to be updated, ya?

PR #722, opened 2020-07-22
input-date-range: validation

margaree:
Redo of https://github.com/BrightspaceUI/core/pull/714 (I think I've address everything mentioned so far in that PR within this one).


components/inputs/input-date-range.js
```diff
@@ -129,16 +149,62 @@ class InputDateRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-		else this.endValue = elem.value;
+		let startDate, endDate;
+		if (elem.classList.contains('d2l-input-date-range-start')) {
+			this.startValue = elem.value;
+			startDate = elem.value;
+			endDate = this.shadowRoot.querySelector('d2l-input-date.d2l-input-date-range-end').value;
+		} else {
+			this.endValue = elem.value;
+			endDate = elem.value;
+			startDate = this.shadowRoot.querySelector('d2l-input-date.d2l-input-date-range-start').value;
```
AllanKerr, line 187: Is there a reason we need to get this from the input rather than `this.startValue`?

AllanKerr:
At some point we'll also want to make these elements set their form values using `setFormValue` so their data gets submitted when a form gets submitted but I'm fine with putting that in a different PR.

components/inputs/input-date-range.js
margaree: I think the way I'd originally implemented it needed it but no it doesn't need to be like this anymore - will fix.

PR #723, opened 2020-07-23
Replace form-element showErrors with validationType

AllanKerr:
# Changes
## Why?
- `showErrors = true` was equivalent to `ValidationType.SHOW_NEW_ERRORS` and `showErrors = false` was equivalent to `ValidationType.UPDATE_EXISTING_ERRORS ` so the only new addition is `ValidationType.SUPPRESS_ERRORS`.
- We need `ValidationType.SUPPRESS_ERRORS` for complex nested custom form elements so that a parent form element can suppress its error message if its children are invalid or vice versa. This may be necessary if both the parent and children show tooltips that overlap. We still want to validate all of them but one may need to suppress its error so that they can render nicely.

## Validation Type
- `ValidationType` provides more control over how errors are rendered in a custom form element's sub-tree.
    - `SHOW_NEW_ERRORS` causes validation to show any errors that occurred immediately. This can be used during blur or change to notify users that their input was invalid.
    - `UPDATE_EXISTING_ERRORS` causes validation to only update an error message if the error message changed to a different one or remove it if the input become valid. This should be used during live validation when a user is in the process of input so we don't show new errors if they are half-way through inputting something like the start of an email address.
    - `SUPPRESS_ERRORS` will validate but suppress any errors that occur. This can be used in the case of nested custom form elements so that the parent won't show any validation errors if the children that the parent depends on are invalid. This might needed if the parent's error message overlaps with a child's error message.


components/inputs/input-date-range.js
```diff
@@ -149,17 +149,13 @@ class InputDateRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(LitEl
...
-		if (errors.length !== 0) {
-			return errors;
-		}
-		return Promise.all([
-			await this.shadowRoot.querySelector('.d2l-input-date-range-start').validate(showErrors),
-			await this.shadowRoot.querySelector('.d2l-input-date-range-end').validate(showErrors)]
-		).then((res) => {
-			return res.reduce((acc, errors) => [...acc, ...errors], []);
-		});
+	async validate(validationType) {
```
margaree, line 152: Thanks for updating this!

PR #728, opened 2020-07-24
Add d2l-form-error-summary Component

AllanKerr:
# Changes
1. Add the `d2l-form-error-summary` component to summarize a group of errors.
error-summary

1. Integrate `d2l-form-error-summary` into `d2l-form`.
form-error-summary


components/form/form-errory-summary.js
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	constructor() {
+		super();
+		this.errors = [];
+		this._expanded = true;
+	}
+
+	render() {
+		const errorSummaryTerm = this.errors.length === 1 ? 'components.form-error-summary.singleError' : 'components.form-error-summary.multipleErrors';
```
dlockhart, line 67: Ooooh, this looks like a fun opportunity to use [the "plural" keyword](https://formatjs.io/docs/icu-syntax#plural-format) in the Intl Message syntax!

components/form/form-errory-summary.js
```diff
@@ -0,0 +1,109 @@
...
+						aria-expanded=${this._expanded ? 'true' : 'false'}
+						icon=${this._expanded ? 'tier1:arrow-collapse-small' : 'tier1:arrow-expand-small' }
+						@click=${this._toggleExpandCollapse}>
+					</d2l-button-icon>
+				</div>
+				<d2l-expand-collapse-content ?expanded=${this._expanded}>
+					<ul class="d2l-form-error-summary-error-list">
+						${this.errors.map(error => html`
+							<li>
+								<d2l-link href=${error.href} @click=${error.onClick}>${error.message}</d2l-link>
```
dlockhart, line 82: Can we just apply `linkStyles` to a native `<a>` here? I think our general best-practice should be to always do that when we can, instead of using extra web components unnecessarily.

components/form/form-errory-summary.js
```diff
@@ -0,0 +1,109 @@
...
+		return this.errors.length > 0 ? errorSummary : html``;
+	}
+
+	async focus() {
+		if (this.errors.length === 0) {
+			super.focus();
+			return;
+		}
+		await this.updateComplete;
+		await new Promise(resolve => requestAnimationFrame(() => requestAnimationFrame(resolve)));
```
dlockhart, line 98: Eeek, seems pretty fragile eh. Were you able to see why the list items weren't available without delaying? I haven't seen us need to do this in other `focus()` methods.

AllanKerr: Yeah, I'm not super happy with it but it seems to be the same reason animations don't run when going from `display: none` to `display: block` without waiting for 2xRAF.

The reason its needed is because `focus` is called immediately after setting `errors` which can trigger's the conditional rendering. If it goes from being hidden due to having no errors to being visible then we need to:
1. Wait for `updateComplete` so that the `errors` has been updated.
1. Wait for the first `requestAnimationFrame` which runs just before the browser actually renders the new content.
1. Wait for the second `requestAnimationFrame` at which point we can guarantee the errors have rendered so the browser will properly focus on them.

components/form/form-errory-summary.js
AllanKerr: I did consider that approach but I'm a bit apprehensive to use it because the more complex syntax often gets mangled during translation. But I can definitely change it if we're not concerned about that.

dlockhart: I was actually interested in running that experiment to see what we got back haha...

components/form/form-errory-summary.js
dlockhart: Hmm... could `d2l-expand-collapse-content-expand`'s promise be incorporated somehow?

dlockhart: Also, what happens if it's collapsed?

AllanKerr: In this case the wait is needed because of the conditional render: `return this.errors.length > 0 ? errorSummary : html``;`. We can't use the expand event because it starts pre-expanded so there is no event fired.

AllanKerr: > Also, what happens if it's collapsed?

If its collapsed then there is no focus change. We could move it to the `expand-collapse` button if we want.

dlockhart: I see -- that's complicated. I'd be curious what Jeff thinks RE: focus management if things are collapsed. Picturing situations where there are long forms and the summary isn't visible, so moving focus to it would bring it into view.

AllanKerr: Chatted with Jeff a bit and we're happy with moving focus to the expand button if the error list is collapsed.

PR #730, opened 2020-07-24
Split List Item into a Mixin

m6jones:
This PR adds the option to use a mixin variant of the list item. For those cases where we want to modularize the HM components a little more.

Allows a consumer to use `d2l-list-item` as normal or extend `ListItemMixin` on a custom list item where the custom part is the data. This allows people to use the mixin for HM components.


components/list/list-item-mixin.js
```diff
@@ -0,0 +1,351 @@
...
+	static get properties() {
+		return {
+			/**
+			 * Breakpoints for responsiveness in pixels. There are four different breakpoints and only the four largest breakpoints will be used.
+			 */
+			breakpoints: { type: Array },
+			/**
+			 * Address of item link if navigable
+			 */
+			href: { type: String },
```
m6jones, line 38: This will conflict with HM `href`. We will need a new name for this attribute. 

When we do we need to backwards compatible it with `href` on `d2l-list-item` .

The new attribute should also handle javascript methods for click events.

zinabat: For backwards compatibility, we could check if the `href` value is set to a hypermedia entity URL. If it isn't, we can safely assume it is meant as a `d2l-list-item` navigation URL, and fill the new property with that value.

m6jones: I'm thinking, we just update `d2l-list-item` component and handle the backwards compatibility there. Since it will always be a list.

components/list/README.md
```diff
@@ -78,6 +78,41 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
 - `selectable` (Boolean): Indicates a checkbox should be rendered for selecting the item
 - `selected` (Boolean): Whether the item is selected
 
+## ListItemMixin
+
+What if you want to use a list item need more modularity? This mixin is for you! This mixin allows you to make a component a list item without the worry of styling. All the properties from `d2l-list-item` will be added to your new component.
```
zinabat, line 83:
```suggestion
Want to maintain consistency with `d2l-list-item` but need more modularity? This mixin is for you! This mixin allows you to make a component into a list item without requiring custom styling. All of the properties and functionality from `d2l-list-item` will be added to your new component.
```

components/list/README.md
```diff
@@ -78,6 +78,41 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
...
+import { ListItemMixin } from './list-item-mixin.js';
+
+class ListItem extends ListItemMixin(LitElement) {
+...
+```
+
+How add the styles:
+```javascript
+static get styles() {
+	return [ super.styles ];
```
zinabat, line 98: I think this is only necessary if people need to extend the css on the component.
```suggestion
	return [ super.styles, css`
	    .some-custom-class {
	        padding: 1rem;
	     }
	` ];
```

m6jones: It's true. However, I sort of want to show how people can make their own. So that is why I added it.

components/list/README.md
```diff
@@ -78,6 +78,41 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
...
+render() {
+	return this._renderListItem({
+		illustration: html`[Image HTML here]`,
+		content: html`[Content here such as d2l-list-item-content]`,
+		actions: html`actions here`
+	});
+}
+```
+Where the parameters correspond to the slots of `d2l-list-item`:
+- illustration (TemplateResult):  Provide a illustration for your list item.
```
dbatiste, line 113: Nit: should be `an`.

components/list/README.md
```diff
@@ -78,6 +78,41 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
...
+- content (TemplateResult): Core content of the list item. You can use content components such as `d2l-list-item-content` here. (This corresponds to the default slot in `d2l-list-item`.
```
dbatiste, line 114: Suggest making this a little more succinct.  "Core content of the list item, such as a `d2l-list-item-content` element."

components/list/list-item-mixin.js
```diff
@@ -0,0 +1,351 @@
...
+			.d2l-list-item-content.d2l-hovering,
+			.d2l-list-item-content.d2l-focusing {
+				--d2l-list-item-content-text-decoration: underline;
+			}
+			[slot="content-action"]:focus {
+				outline: none;
+			}
+			[slot="content"] {
+				justify-content: stretch;
+				padding: 0.55rem 0px;
```
dbatiste, line 112: Nit: no units on `0` values.  Did the style linter not catch this?

components/list/list-item-mixin.js
```diff
@@ -0,0 +1,351 @@
...
+				border-bottom: 1px solid var(--d2l-color-mica);
+				border-top: 1px solid var(--d2l-color-mica);
+			}
+			[slot="content"].d2l-list-item-content-extend-separators {
+				padding-left: 0.9rem;
+				padding-right: 0.9rem;
+			}
+			a[href].d2l-list-item-link {
+				width: 100%;
+				height: 100%;
```
dbatiste, line 98: Nit: a few styles out of order here and below.

components/list/list-item-mixin.js
```diff
@@ -0,0 +1,351 @@
...
+				margin-left: 0.9rem;
+				margin-right: 0;
+			}
+			:host([selectable]:not([disabled]):not([draggable]):hover) d2l-list-item-generic-layout,
+			:host([selectable]:not([disabled])) d2l-list-item-generic-layout.d2l-focusing,
+			:host([selectable][draggable]:not([disabled])) d2l-list-item-generic-layout.d2l-hovering {
+				background-color: var(--d2l-color-regolith);
+			}
+			:host([selected]:not([disabled])) d2l-list-item-generic-layout {
+				background-color: #F3FBFF;
```
dbatiste, line 183: Normally we apply 0.5 opacity for disabled.  Does that not work in this case?  This appears to be a color that's not in our palette.

components/list/list-item-mixin.js
dbatiste: Oh, the list components are still being ignored for style linting.  We should get that enabled as soon as convenient.

m6jones: I'll enable that and every time. You think I would catch this by now lol.

components/list/list-item-mixin.js
m6jones: Will dig into this more.

m6jones: This is actually when it isn't disabled but selected.

m6jones:
So there are a couple of problems that came to my attention with styling the slots. The mixin no longer has the slots to style. So going to close the PR and fix that before putting it up again.

components/list/list-item.js
```diff
@@ -31,327 +8,33 @@ const defaultBreakpoints = [842, 636, 580, 0];
...
 			 * Address of item link if navigable
 			 */
-			href: { type: String },
-			_breakpoint: { type: Number },
-			_dropdownOpen: { type: Boolean, attribute: '_dropdown-open', reflect: true },
-			_hoveringLink: { type: Boolean },
-			_focusing: { type: Boolean },
-			_focusingLink: { type: Boolean },
-			_tooltipShowing: { type: Boolean, attribute: '_tooltip-showing', reflect: true }
+			action: { type: String, reflect: true },
```
dbatiste, line 17: It looks like this new `action` property is already on the `ListItemMixin` - just need to update it to reflect.

From what I gather, the `href` on the item (for navigation) potentially conflicts with the `href` for a hypermedia-drive component - yes?  One might argue that these `href`s ought to be the same.  Food for thought.  🤔 

Before going much further with this though, I think it is worthwhile settling on our solution for the JS action scenario... I recall we discussed this a bit a few weeks ago.  To preface, we will need the ability to define the primary action to be either JS (button semantics) or navigation (anchor semantics), while we currently only support the navigation scenario.  It would be ideal if the consumer could provide the element, leaving the list-item un-opinionated and non-limiting.  For instance, currently we're only exposing `href`, but there are other link specific properties we are omitting, such as `target`.  Otherwise, the consumer needs to tell the list item what type of action it is, so that it can render an `a` or `button`.  I think we talked about the button dispatching a custom event that the consumer could listen for.  In short, let's make sure any API changes we make jive with that use-case.

.stylelintrc.json
```diff
@@ -2,7 +2,6 @@
   "extends": "@brightspace-ui/stylelint-config",
 	"ignoreFiles": [
     "components/demo/code-dark-plus-styles.js",
-    "components/demo/code-tomorrow-styles.js",
-    "components/list/**/*.js"
```
dbatiste, line 6: Thanks for updating this!

components/list/list-item.js
m6jones: >From what I gather, the href on the item (for navigation) potentially conflicts with the href for a hypermedia-drive component - yes? One might argue that these hrefs ought to be the same. Food for thought.

I would really like them to be the same but are used differently. One links to a webpage where the other links to an api. If you have thoughts on how we can treat them as the same I would love to hear that when you get back.

>It looks like this new action property is already on the ListItemMixin - just need to update it to reflect.

I am reflecting in this case so `href` is backwards compatible. Since all other cases don't need it reflected I decided to just add the special case on the `list-item` component.

Yeah, I decided in another PR to handle the event. I believe we agreed that people can either set `action` as a url or define a click handler using a custom event.

I would also prefer not to use name `action` as the url attribute. If anyone can think of a better name I'm all ears.

dbatiste: What about `actionHref` / `action-href`?  That's the approach that was taken with the MVC DataList... it uses `ActionLocation` (`ILocation`) and `ActionOnClick` (`IJavaScriptFunction`).

m6jones: Awesome!

dbatiste (approved):

PR #733, opened 2020-07-27
Add getFileIconType helpers

zhangium:
Helper functions that return a file icon type based on the given file extension or filename string. Returns string of form `file-` + type, corresponding to an icon name for the `<d2l-icon>` component.

Corresponding names:
- file-archive
- file-audio
- file-document (_default_)
- file-image
- file-presentation
- file-video

```js
import { getFileIconTypeFromExtension, getFileIconTypeFromFilename } from '@brightspace-ui/core/helpers/getFileIconType.js';
getFileIconTypeFromExtension('docx'); // returns 'file-document'
getFileIconTypeFromFilename('MY_SONG.MP3'); // returns 'file-audio'
```


helpers/getFileIconType.js
```diff
@@ -0,0 +1,103 @@
+export function getFileIconTypeFromExtension(extensionString) {
+	var iconType;
```
dlockhart, line 2: Instead of using a `var` here (which the linter is complaining about), could you just return from the `case` statements?

helpers/getFileIconType.js
```diff
@@ -0,0 +1,103 @@
...
+	switch(extensionString.toLowerCase()) {
```
dlockhart, line 3:
```suggestion
	switch (extensionString.toLowerCase()) {
```

helpers/getFileIconType.js
```diff
@@ -0,0 +1,103 @@
...
+		case '3gp':
+		case 'asf':
+		case 'asx':
+		case 'avi':
+		case 'divx':
+		case 'flv':
+		case 'm4v':
+		case 'mkv':
+		case 'mov':
+		case 'mov':
```
dlockhart, line 75:
```suggestion
		case 'mov':
```

helpers/README.md
```diff
@@ -170,3 +170,15 @@ import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
 // gets a unique indexed id (for lifetime of page)
 getUniqueId();
 ```
+
```
dlockhart, line 173: I'm torn between putting this in "helpers" like you have it, and just putting it in with "components/icons". Everything here is icon-related... so having it there and the docs there with the rest of the icon docs makes sense I think. My impression of "helpers" was they were always meant for non-component-specific stuff... but this is somewhat component related.

Anyone else have opinions here?

zhangium: I originally thought to put it in helpers because I didn't think this standalone file fit within "components/icons". I agree this is only used within icons, and I can also add to the icon readme to point to this helper. Awaiting more comments/opinions on this.

AllanKerr: I think it'd make sense to include these in an `icon-helper`  file within `components/icons` just because the output of both of these functions is tightly coupled to `d2l-icon`. We've included component specific helpers in a few different places like with the [`form-helper`](https://github.com/BrightspaceUI/core/blob/master/components/form/form-helper.js) and [`input-time` helpers](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-time.js#L38)

dlockhart: I'm leaning the same way, except I'd suggest naming it `getFileIconType.js` as you have here though as I don't think we'd add more things to it since these utility functions will be rarely used and if we merge them then users would need to always download them all together. The docs could just go into the icon README though!

PR #735, opened 2020-07-29
input field long text scroll to the begin when off focus

stevenao:


dlockhart:
From my testing, it looks like Safari and new-Edge also scroll to the beginning on-blur. So it's just Firefox that doesn't. It feels wrong applying this to all browsers when only Firefox is behaving slightly differently -- is it that big a deal? Our Firefox usage is about 5.3% of users.

stevenao:
@dlockhart I spoke to Joseph about it. He is ok to ignore FF. Thanks for checking. I will close this PR now.

PR #736, opened 2020-07-29
input-time-range: validation

margaree:
Overall this is very similar to `input-date-range`, with the exception of `input-time` not having individual validation since it can't really be invalid on its own.

The accessibility piece with the tooltip/error message will be done separate as it will be more general to all `FormElementMixin` usage.


components/inputs/input-time-range.js
```diff
@@ -98,6 +109,30 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		if (!this.label) {
 			console.warn('d2l-input-time-range component requires label text');
 		}
+
+		if (!this.startValue && !this.endValue) {
```
margaree, line 113: It was necessary to give `startValue` and `endValue` default values depending on if the consumer sets them.

`startValue` should be midnight be default (used `getDefaultTime` from `input-time` in case that changes) and `endValue`, if it does not have a value, should be one time interval after (as seen in current time range usage in Brightspace).

components/inputs/input-time.js
```diff
@@ -17,7 +18,7 @@ const TODAY = getToday();
 const END_OF_DAY = new Date(TODAY.year, TODAY.month, TODAY.date, 23, 59, 59);
 const INTERVALS = new Map();
 
-function getIntervalNumber(size) {
+export function getIntervalNumber(size) {
```
margaree, line 21: This and `getDefaultTime` were needed in `input-time-range` so I chose to export them rather than move them to `dateTime` helper since they are very fairly `input-time` specific. I can move them to the helper instead if that makes more sense.

components/inputs/input-time.js
```diff
@@ -74,16 +75,11 @@ function formatValue(time) {
 	return value;
 }
 
-function parseValue(val) {
```
margaree, line 77: This was moved to be `getDateFromISOTime` in the `dateTime` helper

components/inputs/test/input-time-range.test.js
```diff
@@ -47,10 +47,84 @@ describe('d2l-input-time-range', () => {
...
+				inputElem.value = '00:15:00';
+				setTimeout(() => dispatchEvent(inputElem, 'change'));
+				await oneEvent(elem, 'change');
+				expect(elem.startValue).to.equal('0:15:00');
+				expect(elem.endValue).to.equal('00:30:00');
+				expect(elem.invalid).to.be.false;
+				expect(elem.validationError).to.be.null;
+			});
+
+			it('should be valid if end time and no start time', async() => {
```
dlockhart, line 69: Just to confirm... it's not that there's no start value in the range, it's that one wasn't provided and we default it to midnight? Like visually to the user there's no way for a range to be missing a start and/or end value -- it'll always have both?

margaree: Yup that's correct - I copy/pasted from the date range tests and forgot to update the context here.

components/inputs/input-time-range.js
```diff
@@ -98,6 +109,30 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+			} catch (e) {
+				// case where startValue is invalid
+				endValue = new Date(getDefaultTime());
+			}
+			endValue.setMinutes(endValue.getMinutes() + interval);
+			this.endValue = formatTimeInISO({hours: endValue.getHours(), minutes: endValue.getMinutes(), seconds: endValue.getSeconds()});
+		} else if (!this.startValue && this.endValue) {
+			const startValue = getDefaultTime();
+			this.startValue = formatTimeInISO({hours: startValue.getHours(), minutes: startValue.getMinutes(), seconds: startValue.getSeconds()});
+		}
```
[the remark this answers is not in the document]
margaree: Yup I think that should work and simplify things - thanks!

components/inputs/input-time-range.js
```diff
@@ -90,6 +96,62 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 		this.timeInterval = 'thirty';
+
+		this._endDropdownOpened = false;
+		this._endInputId = getUniqueId();
+		this._startDropdownOpened = false;
+		this._startInputId = getUniqueId();
+	}
+
+	get endValue() { return this._endValue; }
+	set endValue(val) {
```
AllanKerr, line 107: I think we'll run into some problems with these checks if the start or end value are valid times but don't match the `timeInterval`. The inner `d2l-input-time` will have the right value and the UI will show the right value but `startValue` or `endValue` will be incorrect.

components/inputs/input-time-range.js
```diff
@@ -98,6 +160,17 @@ class InputTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+
+		if (!this.startValue) {
+			const startValue = getDefaultTime();
+			this.startValue = formatTimeInISO({hours: startValue.getHours(), minutes: startValue.getMinutes(), seconds: startValue.getSeconds()});
+		}
+		if (!this.endValue) {
+			const endValue = getDateFromISOTime(this.startValue);
+			endValue.setMinutes(endValue.getMinutes() + getIntervalNumber(this.timeInterval));
+			this.endValue = formatTimeInISO({hours: endValue.getHours(), minutes: endValue.getMinutes(), seconds: endValue.getSeconds()});
+		}
```
AllanKerr, line 172: Is there a reason we can't put these in the constructor?

margaree: The assignment of `startValue` could but `endValue` depends on if `startValue` has a value so I think we'd want to keep that here.

components/inputs/input-time-range.js
margaree: Ah yup that's a good point.

PR #742, opened 2020-07-30
feat: d2l-form nested validation and submission

AllanKerr:
# Changes

## Add support for nested form submission and validation:
- When a parent `d2l-form` is validated or submitted it will validate or submit all descendant `d2l-form` components; both directly nested and composed within the shadow DOM of descendant custom elements.
- All errors for nested forms are bubbled up and displayed within a single error summary in the root form.
- If all `d2l-form` components pass validation then they are submitted individually.
## Add support for `no-nesting`:
- This allows descendant forms to opt-out of being validated and submitted when an ancestor `d2l-form` is submitted or validated.
## Add support for `async-submit`:
- When the `async-submit` attribute is present the form will be submitted using XHR rather than native form submit.
- When forms are nested they can only be submitted using XHR because each `d2l-form` is submitted individually.

### DIFF
--- a/components/form/demo/form-dialog-demo.js
+++ b/components/form/demo/form-dialog-demo.js
@@ -1,0 +1,85 @@
+
+import '../../button/button.js';
+import '../../button/floating-buttons.js';
+import '../../dialog/dialog.js';
+import '../../inputs/input-text.js';
+import '../../validation/validation-custom.js';
+import '../form.js';
+import './form-panel-demo.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { inputStyles } from '../../inputs/input-styles.js';
+import { selectStyles } from '../../inputs/input-select-styles.js';
+
+class FormDialogDemo extends LitElement {
+
+	static get styles() {
+		return [inputStyles, selectStyles, css`
+
+			.d2l-form-dialog-demo-container {
+				margin-bottom: 10px;
+			}
+		`];
+	}
+
+	render() {
+		return html`
+			<d2l-form id="dialog-main-form">
+				<div class="d2l-form-dialog-demo-container">
+					<label>Email<input class="d2l-input" name="email" type="email"></label>
+				</div>
+				<div class="d2l-form-dialog-demo-container">
+					<d2l-validation-custom for="password-dialog" @d2l-validation-custom-validate=${this._validatePassword} failure-text="Expected hunter2" ></d2l-validation-custom>
+					<label>Password<input class="d2l-input" id="password-dialog" name="password" required type="password"></label>
+				</div>
+				<d2l-dialog id="dialog" title-text="My Favorites">
+					<d2l-form id="dialog-secondary-form" async-submit no-nesting @d2l-form-submit=${this._onDialogSubmit}>
+						<fieldset class="d2l-form-dialog-demo-container">
+							<legend>Choose your favorite monster</legend>
+							<label><input type="radio" name="monster" value="kraken">&nbsp;Kraken</label>
+							<label><input type="radio" name="monster" value="sasquatch">&nbsp;Sasquatch</label>
+						</fieldset>
+						<div class="d2l-form-dialog-demo-container">
+							<label>Favorite Pet<br/>
+								<select class="d2l-input-select" name="pets" required>
+									<option value="">--Please choose an option--</option>
+									<option value="porpoise">Porpoise</option>
+									<option value="house hippo">House Hippo</option>
+									<option value="spiker monkey">Spider Monkey</option>
+									<option value="capybara">Capybara</option>
+								</select>
+							</label>
+						</div>
+					</d2l-form>
+					<d2l-button slot="footer" primary @click=${this._onDialogSubmitClicked}>Save</d2l-button>
+					<d2l-button slot="footer" data-dialog-action>Cancel</d2l-button>
+				</d2l-dialog>
+				<div class="d2l-form-dialog-demo-container">
+					<d2l-button @click=${this._openDialog}>Show Favorites Dialog</d2l-button>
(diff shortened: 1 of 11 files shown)


components/form/demo/form-panel-demo.js
```diff
@@ -0,0 +1,108 @@
...
+							</label>
+						</div>
+					</d2l-form>
+				</d2l-expand-collapse-content>
+			</div>
+		`;
+	}
+
+	_onInvalid() {
+		this._expanded = true;
```
dlockhart, line 99: Doesn't have to be part of this PR, but I'd love to be able to handle this case automatically. Like fire a bubbling event originating at invalid form elements and then have `d2l-expand-collapse-content` (and similar components) listen for it and expand automatically. The MVC & legacy frameworks currently handle something like this.

components/form/demo/form-dialog-demo.js
```diff
@@ -0,0 +1,85 @@
...
+			<d2l-form id="dialog-main-form">
+				<div class="d2l-form-dialog-demo-container">
+					<label>Email<input class="d2l-input" name="email" type="email"></label>
+				</div>
+				<div class="d2l-form-dialog-demo-container">
+					<d2l-validation-custom for="password-dialog" @d2l-validation-custom-validate=${this._validatePassword} failure-text="Expected hunter2" ></d2l-validation-custom>
+					<label>Password<input class="d2l-input" id="password-dialog" name="password" required type="password"></label>
+				</div>
+				<d2l-dialog id="dialog" title-text="My Favorites">
+					<d2l-form id="dialog-secondary-form" async-submit no-nesting @d2l-form-submit=${this._onDialogSubmit}>
```
dlockhart, line 35: What are your thoughts about a use case similar to this, where you have a bunch of content that's hidden (like in a closed dialog) that when hidden you don't want to participate in the form but when displayed you do? 

Could be as simple as them manually adding/removing the `no-nesting` attribute based on the visibility for now. But it would be nice for the nested `d2l-form` to not actually be necessary at all -- just have some hidden content that only participates when it's visible.

components/form/form.js
```diff
@@ -1,22 +1,53 @@
...
 
 	static get properties() {
 		return {
+			/**
+			 * The URL that processes the form submission
+			 */
 			action: { type: String },
+			/**
+			 * Causes the form to be submitted as an asynchronous XHR request rather than a native form submission.
+			 * If the form has nested forms then
```
dlockhart, line 28: Looks like the comment got cut off. I'm curious about when you would want async XHR vs. not.

components/form/form.js
```diff
@@ -1,22 +1,53 @@
...
+			enctype: { type: String },
+			/**
+			 * The HTTP request method to use to submit the form
+			 * @type {'POST'|'PUT'}
+			 */
 			method: { type: String },
+			/**
+			 * Prevents the form from being validated and submitted when an ancestor form is validated or submitted
+			 */
+			noNesting: { type: Boolean, attribute: 'no-nesting', reflect: true },
```
dlockhart, line 44: I don't _love_ the name, but can't really think of an alternative.

components/form/demo/form-dialog-demo.js
AllanKerr: Something like this would need to be approached carefully since I don't want to deviate too far from how native forms work where hidden inputs still get submitted.

The best solution that comes to mind is to use a `MutationObserver` and then conditional rendering can be used to include or exclude sections. This might not play well if we want to add Polymer support due to how `dom-if` and stamping works but I'm not sure how concerned we'd be about that.

components/form/form.js
AllanKerr: The main use case for non-XHR would be if you want to display the response in some target (`_self`, `_blank`, `_parent`) like  with native form submission. Supporting this also means we can swap a native `form` out for a `d2l-form` and submission should still work the same.

XHR would be used whenever we want to be able to do some custom logic after submit rather.

That being said, I fully expect most modern use cases to have an event listener for `@d2l-form-submit`, call `preventDefault()` and submit via a Hypermedia action.

components/form/demo/form-dialog-demo.js
dlockhart: Good point about native forms, maybe I'm overthinking things. 👍

components/form/form.js
dlockhart: Gotcha. Wouldn't this need a `target` attribute for the form then? Just wondering if we could get away without the attribute... like if they set the target then we use a native form, otherwise we XHR (assuming action is also set)?

AllanKerr: Yeah, it'll likely need a `target` attribute. I've been trying to think of a good way to allow options like that to be configured without having a bunch of attributes that only apply if other attributes are set but have yet to come up with a good solution.

dlockhart: What about having `target`, `action`, `method` and `enctype` only apply to native form submission. If you want to do an XHR, add a listener for the `d2l-form-submit` event and do it yourself via fetch?

Because yeah otherwise... `target` doesn't make sense with XHRs and `action` is more of a form-y thing.

AllanKerr: The main reasons `d2l-form` knows how to do XHR submissions are:
1. When using nesting submission has to be `async` so if we'd then default to only firing a `d2l-form-submit` event and not actually submitting anything unless they have their own custom submit behavior. Its possible we're okay with that.
1. Building up a `FormData` object from a JSON object can be a pain if it includes things like files. If the form does the XHR submission then I can build it internally from the private native form with `new FormData(this._form)` rather than trying to piece it together manually.

If we aren't too concerned about those then we can get rid of it and just have native submission and custom submission if the have an event listener.

PR #743, opened 2020-07-30
Update components/list/README.md

zinabat:
Adds documentation for drag n drop


components/list/README.md
```diff
@@ -34,6 +34,22 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
...
+* **ArrowUp** moves to the same item in the row above, if available
+* **ArrowDown** moves to the same item in the row below, if available
+* **PageUp** moves to the same item in the row **five** rows above, if available
+* **PageDown** moves to the same item in the row **five** rows below, if available
+* **Home** moves to the first item in the row
+* **Ctrl+Home** moves to the first item of the first row
+* **End** moves to the last item in the row
+* **Ctrl+End** moves to the last item of the last row
+* **Space** and **Enter** simulate a click on the focused item
+
```
dbatiste, line 52: As an aside, what are the valid slots for devs to place focusable elements?  Is it just the secondary actions slot?  Do you think there is value is calling attention to this?  I suspect we're going to see lots of devs using the slots incorrectly.

zinabat: Good point, I meant to put this in there.

PR #744, opened 2020-07-30
fix: Fix tryGetLabelText in Safari when used on elements in the Shadow DOM

AllanKerr:
# Changes
## Issue
The [labels](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) property is fully supported in Safari but is broken when called on an
element in the shadow DOM. In this case it always returns an empty list even
if the element has an implicit or explicit label.

## Fix
To fix it, we need to manually check the parent for an implicit label or look
for labels with a for attribute pointing to the label.


AllanKerr:
> Wow crazy! Wonder if there's a Safari bug already tracking this, or if not we could add one...

Surprisingly I couldn't find any existing bug reports that match

PR #746, opened 2020-07-31
DE39995 - Fix event firing in consumer of the tab-panel-mixin

svanherk:
In 20.20.7 I tried to cleanup this private helper method (https://github.com/BrightspaceUI/core/pull/566), but didn't realize it was being overwritten by a consumer (https://github.com/BrightspaceUI/facet-filter-sort/blob/master/components/d2l-filter-dropdown/d2l-filter-dropdown-category.js#L188).  This means the `d2l-filter-dropdown-category-selected` event hasn't been firing since then.  From what I can tell, Portfolio is the only place currently using this event and since it's an FRA, I'll be releasing a patch fix on the version it needs (rather than making new BSI releases).

Putting everything back to get back to green ASAP, but maybe the consumer should be doing something different and less fragile.

Tests in `filter-facet-sort` would have caught this months ago, but it uses npm caches.  😒 So I'm going to turn that off.

### DIFF
--- a/components/tabs/tab-panel-mixin.js
+++ b/components/tabs/tab-panel-mixin.js
@@ -69,13 +69,18 @@ export const TabPanelMixin = superclass => class extends superclass {
 			if (prop === 'selected') {
 				if (this.selected) {
 					requestAnimationFrame(() => {
-						this.dispatchEvent(new CustomEvent(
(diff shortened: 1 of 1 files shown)


svanherk:
Jumping ship on this plan - talking with Allan a fix in `facet-filter-sort` will be easier, especially for the backport.  And Edge disconnects in CI are making me angry.

PR #747, opened 2020-07-31
1.49.1 - DE39995 - Fix event firing in consumer of the tab-panel-mixin

svanherk:
Same as https://github.com/BrightspaceUI/core/pull/746, but applied against the `1.49.0` version of `core`.  This is the `core` version Portfolio has in their 5.14 version.  Since this is before we setup semantic-release, going to have to release it the old fashioned way.


svanherk:
> I think you'll also need to update the npm keys for the release to work since they were regenerated when we switched to Semantic Release.

Thanks, makes sense!  But closing anyways, going to fix in `facet-filter-sort`

PR #748, opened 2020-07-31
fix: update lang term values with translations

margaree:
Related to https://github.com/BrightspaceUI/core/pull/739 going into master


svanherk:
Should this have been updated? https://github.com/BrightspaceUI/core/blob/master/lang/localize-core-element.js

Going through this with my-courses right now

margaree:
Yup it does - good catch! Thanks!

PR #749, opened 2020-07-31
provide a way to give the inner input field an id

stevenao:
https://rally1.rallydev.com/#/29180338367ud/iterationstatus?detail=%2Fuserstory%2F376732857096
https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-container-mixin.js#L60

The story is making use of the `isFocusable` api, finding all the invalid inputs, and then put a focus on them. I need a way to ignore certain input fields. Using their id's provides just what I need and it is less intrusive.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -196,10 +196,6 @@ class InputText extends RtlMixin(LitElement) {
 		this._lastSlotWidth = 0;
 	}
 
-	_getInputId() {
-		return this.id || this._uniqueId;
-	}
-
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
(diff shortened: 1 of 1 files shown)


dlockhart:
I'm not sure this is what you want... the `id` on the internal input should definitely not be something we expose, since it should be completely internal. We may choose one day to render something completely different. And regardless, the `id` on the `d2l-text-input` needs to be for it, not for something else inside of it, otherwise you wouldn't be able to `document.getElementById()` for `d2l-text-input`... or if you did in a browser that doesn't support shadow DOM (like Edge), you'd get back both elements.

dlockhart:
@AllanKerr FYI as it appears as the new form validation stuff may be what they actually want here...

stevenao:
I had a look at the Validation stuff from the demo. It doesn't look like it is what I am tackling here. It is more of a problem - after the `getFocusableDescendant` is returning a element that I should have no knowledge of (an element inside a shadow dom).  `d2l-input-text` actually provides a `focus` function to abstract away the need to set focus on the inner element. However, the foucs helper function is aware of the fact that the component can set focus. 

I am going to close this PR and send a new one.

PR #750, opened 2020-07-31
fix: add new languages to localize-core-element

margaree:


svanherk:
Serge config?  Did we determine if that does anything?  Probably only needed in master anyways.

margaree:
Ah yeah I think serge config is fine to just be updated in master (or not at all in these cases - unless we want `cy-gb` to go to `cy`?). I'll do that as part of my `master` PR.

svanherk:
Yeah I ended up going with https://github.com/Brightspace/d2l-my-courses-ui/pull/838/files#diff-b1ebc4241aea524d9577292b391c966b for completeness, but I think its the same as doing nothing

lang/localize-core-element.js
```diff
@@ -9,6 +9,9 @@ export const LocalizeCoreElement = superclass => class extends LocalizeMixin(sup
 				case 'ar':
 					translations = await import('./ar.js');
 					break;
+				case 'cy':
```
svanherk, line 12: If the language set is `cy-gb`, will the `langs` object passed in have both `cy` and `cy-gb`?

margaree: yup it would get passed [`cy-gb`, `cy`, `en`] (and I think `en-us` before `en` but I don't recall for sure)

PR #751, opened 2020-07-31
fix: Hide the error icons when the element has focus

AllanKerr:
# Changes 
- Hide the `input-select` error icon on focus
- Hide the `textarea` error icon on focus
- Hide the `input-text` error icon on focus

error-icon-focus


AllanKerr:
Found some unwanted changes in the `textarea` visual diff so added a fix for those.

PR #752, opened 2020-07-31
Fix overflow on list containers.

m6jones:
There was a problem with consistent evaluation with a list item having a name that is way too long and single character. Things started to over extend to different parts grid cells. Here is the fix:
image

Picture of the problem: 
image

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -65,8 +65,6 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			::slotted([slot="content"]),
 			::slotted([slot="actions"]) {
 				grid-row: 1 / 2;
-				overflow: hidden;
-				position: relative;
 			}
 			::slotted([slot="outside-control"]) {
 				grid-column: outside-control-start / outside-control-end;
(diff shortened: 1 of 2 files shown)


zhangium:
I found that this change breaks any expanded menus visibility (working in master).

image

components/list/list-item-generic-layout.js
```diff
@@ -65,6 +65,8 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			::slotted([slot="content"]),
 			::slotted([slot="actions"]) {
```
zhangium, line 66: I think if you pull the actions slot out on its own without the overflow and position styling it prevents the tooltips and expanded menu from getting covered up.

dbatiste: Yeah we definitely need to be careful of overflow with the dropdown components.  The dropdown has some magic positioning, and if you place it in a container that does not allow overflow, there may be some issues with it being cut off or not displaying correctly.

m6jones: Yup! I totally forgot about this PR :( I'll fix the drop down issue and only hide on the overflow-x

PR #754, opened 2020-08-04
fixing apostrophes in French

dlockhart:
I think we just want these to be the standard ones, not a special unicode one.


svanherk:
In the #localization channel, there's a slack workflow to report errors the translators made.  I sent one off for my-courses when they unnecessarily escaped all the apostrophes to try to avoid having this come back next time, by just sending a link to the github commit diff.  Seems like it might make sense to send this one too, and maybe the changes @margaree made in the other PR?

dlockhart:
Thanks Stacey -- I submitted the workflow for the apostrophe issue.

PR #756, opened 2020-08-04
Build changes: Experiment with BrowserStack

margaree:

PR #757, opened 2020-08-04
Lock mocha version

margaree:
With the latest mocha (8.1) I'm seeing the error `TypeError: Cannot read property 'Runnable' of undefined` when running the karma tests with the problem tracing to `var Runnable = window.Mocha.Runnable;` and `window.Mocha` being undefined. Locally locking the version seems to fix it.


margaree:
Done in https://github.com/BrightspaceUI/core/pull/756

PR #760, opened 2020-08-05
feat: Make d2l-input-text a form element

AllanKerr:
# Changes
## Original Changes
- `d2l-input-text` is now a custom form element
- It will now validate itself, enter error states, and show error tooltips for the following attributes:
     - `pattern`
     - `type="email"`
     - `type="url"`
     - `minlength`
     - `maxlength`
     - `min`
     - `max`
     - `required`
- It can also be included in `d2l-form` to be validated and submitted when the form is submitted.
- `d2l-validation-custom` elements can be attached to it to add custom validation logic.

## Additional Changes
- Add `noValidate` to `FormElementMixin` to allow automatic validation to be disabled for custom form elements.
- Add `invalid-change` event so that users can be notified when the `invalid` state of a custom form element changes.


AllanKerr:
> Looks good! Since @krshnpatel-d2l is going to be working on building a `d2l-input-number`, I'm wondering if we should be pulling the number-related stuff out of here? We really don't want folks using `<d2l-input-text type="number">`, since it won't get any of our number formatting or parsing logic. We'd need to search to make sure nobody was using it for this. 🤞

Not supporting overflow and underflow validation isn't great because we support `min` and `max` which trigger those error types. So its less of a "do we want to support it" question and more of a "do we want an improved error message if it gets triggered"?

As long as we allow those constraints I think it'd be best to include corresponding error messages but I'm open to discussion there.

AllanKerr:
I ended up making one more change to move `requestValidate` back to blur instead of having it in change. This is so that the input is validated if the user typed something then removed it all. This should trigger a `required` error but since the value started empty and ended empty it doesn't trigger a `change` event, only `blur`.

PR #766, opened 2020-08-06
Add alert-toast-close event so consumers can keep their private properties in sync.

### LINKED ISSUE #759: Alerts should dispatch an event when closed.
Consumers commonly declare a private property for the alert state in their component, however they need to keep it in sync with the alert open/closed state so that subsequent changes to reopen the alert cause it to be displayed again.  To allow this, we should dispatch a "close" event when the alert is closed, that consumers can listen for and keep their private state property in sync.

dbatiste:
This fixes #759.

I've also proceeded to update the event names for the alert to be present tense.  I suspect the past-tense names were an artifact of the previous Polymer alerts.  Unfortunately this will require a little cleanup in apps.  I would have left it as it, except that I did not want `d2l-alert` and `d2l-alert-toast` following different convention, and I also didn't want to continue with the past-tense naming since there is strong precedent for present tense in dropdown, dialog, browser events, and other components.

### DIFF
--- a/components/alert/README.md
+++ b/components/alert/README.md
@@ -25,8 +25,8 @@ The `d2l-alert` component can be used to communicate important information relat
 - `type` (String, default: `'default'`): type of the alert being displayed. Can be one of  `default`, `critical`, `success` , `warning`
 
 **Events:**
-* `d2l-alert-closed`: dispatched when the alert's close button is clicked
-* `d2l-alert-button-pressed`: dispatched when the alert's action button is clicked
+* `d2l-alert-close`: dispatched when the alert's close button is clicked
+* `d2l-alert-button-press`: dispatched when the alert's action button is clicked
 
 ## d2l-alert-toast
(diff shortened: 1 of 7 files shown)


svanherk:
I assume we want to avoid a major version bump of `core`, so is the plan to update all the places quickly?  Looks to be about 2 places in the LMS, 2 components in BSI, and 10 FRA repos?

dbatiste:
Currently both events will be dispatched, but yeah, updating the apps to use the new event name is preferred vs keeping it around indefinitely.  Yeah this will just be a minor bump, and non-breaking.

svanherk:
> Currently both events will be dispatched, but yeah, updating the apps to use the new event name is preferred vs keeping it around indefinitely. Yeah this will just be a minor bump, and non-breaking.

Ah gotcha, looked too quickly and didn't notice you were still sending the old ones as well

PR #768, opened 2020-08-07
input-date-time: validation

margaree:
This PR adds validation for date/times being selected outside of min and max value. It prevents the `input-date` validation. The tooltip is attached to the date input but shows when time input receives focus/hover. Tooltip is not rendered when either dropdown is opened.


components/inputs/input-date-time.js
```diff
@@ -138,30 +157,98 @@ class InputDateTime extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+
+	_handleInputTimeBlur() {
+		const tooltip = this.shadowRoot.querySelector('d2l-tooltip');
+		if (tooltip) tooltip.hide();
+	}
+
+	_handleInputTimeFocus() {
+		const tooltip = this.shadowRoot.querySelector('d2l-tooltip');
+		if (tooltip) tooltip.show();
+	}
```
AllanKerr, line 233: This feels like something that could be made a lot easier. 🤔

Not sure if anyone else has any thoughts but could it be worth expanding tooltip to support multiple targets? We could then do `for="${dateId} ${timeId}"` and the tooltip would position itself relative to the smallest rectangle that contains both along with handling focus events from both?

It would be a bit of work to add so probably not immediately but maybe something that'd be of value?

dlockhart: I like that idea!

PR #772, opened 2020-08-10
input-date-time-range: Validation

margaree:
This functions very similarly to `input-date-range` validation. I added max value and min value in as well.


components/inputs/input-date-time-range.js
```diff
@@ -85,25 +104,39 @@ class InputDateTimeRange extends RtlMixin(LocalizeCoreElement(LitElement)) {
 	}
 
 	render() {
-		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-date-time-range.startDate');
-		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-date-time-range.endDate');
+		const tooltipStart = (this.validationError && !this._startDropdownOpened) ? html`<d2l-tooltip align="start" for="${this._startInputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
```
AllanKerr, line 107: One issue with having the tooltips attached to the `d2l-input-date-time` components is that if the tooltip opens above it'll be placed above the label rather than above the input box. A fix would be to set `position="bottom"` so  it always opens below but that would cause the tooltip to position itself slightly less optimally.
Capture

components/inputs/input-date-time.js
```diff
@@ -220,6 +220,10 @@ class InputDateTime extends FormElementMixin(LocalizeCoreElement(RtlMixin(LitEle
 			await this.updateComplete;
 			this._handleInputTimeFocus();
 		}
+		this.dispatchEvent(new CustomEvent(
+			'd2l-input-date-time-dropdown-toggle',
+			{ bubbles: true, composed: false, detail: { opened: e.detail.opened } }
```
AllanKerr, line 225: Does this need to `bubble`? It might be nice to force people to handle it on the `d2l-input-date-time` itself

PR #774, opened 2020-08-10
add requestResize() method to dropdown-content

johngwilkinson:
To cover the case discussed here: https://d2l.slack.com/archives/C0PHG3QB0/p1596220392383400

The new method is just like the page resize handler, but public. It has one behaviour which is arguably a bug, but which I'm treating as out-of-scope because it already happens on page resize: if there is a scroll bar on the dropdown content, it jumps to the top on resize instead of staying where it was.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -264,6 +264,13 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		this.opened = true;
 	}
 
+	async requestResize() {
```
AllanKerr, line 267: Can we rename this to `resize` so its consistent with the public `resize` method in dialog? https://github.com/BrightspaceUI/core/tree/master/components/dialog

AllanKerr, line 267: Its probably worth updating `__onResize` to call this method.

johngwilkinson: Done

johngwilkinson: Done

components/dropdown/README.md
```diff
@@ -209,6 +209,11 @@ To make your usage of `d2l-dropdown-more` accessible, use the following property
 * `footer`: Sticky container at the bottom of the dropdown
 * `default`: Anything inside of `d2l-dropdown-content` that isn't in the `header` or `footer` slots appears as regular content
 
+**Methods:**
+* `async resize()`: Call if the size of the content changes due to a change in
+a nested component. The nested component may choose to fire a custom event, which
+the component containing the `d2l-dropdown-content` can catch and call this method.
```
AllanKerr, line 215: Nit: I'm not sure if mentioning custom events brings any value since calling `resize` from an event listener is more about one specific use case than how it could be used in general. Its really just a public method that triggers the `dropdown-content` to resize to fit its content. It could also get called due to things like a `MutationObserver`, `ResizeObserver` or even something else altogether.

johngwilkinson:
There are two tests that have failed at least twice in a row now. They seem unrelated to this change.

AllanKerr:
> There are two tests that have failed at least twice in a row now. They seem unrelated to this change.

Looks like it might have just been a bit of flake, its passed now.

PR #776, opened 2020-08-11
dismissible tests: fix flakiness

margaree:
I tried some other timeout related changes but rearranging the tests seemed to be the only change that helped. I reran several times and all passed (we'll see if this works out in the long run). Haven't dug much into why this works. This flakiness only seems to happen on chrome and when the full test suite is run.


margaree:
Nevermind - seems like this doesn't consistently fix it. I'll dig more.

PR #779, opened 2020-08-11
FormElementMixin: announce errors on blur and when updated

margaree:
Notes:
- `blur` seems to work fine when tested (mostly tested with `input-text` in order to test both change and blur, did some testing with `input-date`)
- There is some repetition with `input-text` since the tooltip is attached to the `input` (depends on the browser combination)

Results:
Does not work (announce just doesn't seem to trigger anything):
- JAWS + firefox
- NVDA + edge

Repetition
- NVDA + firefox (minor repetition - seems to state tooltip message on first blur + also states announce)
- NVDA + chrome (states both tooltip message and announcement on blur and update)
- JAWS + chrome (states both tooltip message and announcement on blur and update)
- VoiceOver + Chrome (minor repetition - tooltip announces once when focus is received and once when change happens, so not really related to this)
- VoiceOver + Safari (minor repetition - tooltip announces once when focus is received and once when change happens, so not really related to this) - Also worth noting that when typing caused an updated error, VoiceOver would interrupt the announce to state what letter had been typed

### DIFF
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -1,4 +1,5 @@
 import { isCustomFormElement, tryGetLabelText } from './form-helper.js';
+import { announce } from '../../helpers/announce.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 
 export const ValidationType = {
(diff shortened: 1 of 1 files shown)


margaree:
Discussed this further and decided to close this for now, since most cases will be covered by the addition of `announce` to tooltip, along with accessibility in form error summary.

PR #780, opened 2020-08-11
Create input-number component

krshnpatel-d2l:
## [Demo Page](http://localhost:8000/components/inputs/demo/input-number.html)

To-Do:
- [x] autocomplete
- [x] autofocus
- [x] disabled
- [x] label
- [x] labelHidden
- [x] max
- [x] maxFractionDigits
- [x] min
- [x] minFractionDigits
- [x] placeholder
- [x] required
- [x] value
- [x] format input based on locale
- [x] _handleChange
- [x] validation and forms
- [x] demo page
- [x] unit tests
- [x] accessibility tests
- [x] visual diff tests
- [x] documentation


components/inputs/input-number.js
```diff
@@ -0,0 +1,83 @@
...
+	static get properties() {
+		return {
+			autocomplete: { type: String },
+			autofocus: { type: Boolean },
+			disabled: { type: Boolean },
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
+			max: { type: Number },
+			min: { type: Number },
+			name: { type: String },
```
dlockhart, line 19: Let's leave `name` off for now since I'm not sure what setting it on the inner input will even do (it's for form submission).

components/inputs/input-number.js
```diff
@@ -0,0 +1,83 @@
...
+			placeholder: { type: String },
+			required: { type: Boolean },
+			step: { type: Number },
+			value: { type: Number },
+			width: { type: String }
```
dlockhart, line 24: I don't think we need `width`, since they can set it using inline styles `<d2l-input-text style="width:400px">`.

components/inputs/input-number.js
```diff
@@ -0,0 +1,83 @@
...
+			this.style.width = this.width;
+		}
+
+		return html`
+			<label
+				class="${this.label && !this.labelHidden ? 'd2l-input-label' : 'd2l-offscreen'}"
+				for="${this._inputId}"
+				id="${this._inputId}-label">${this.label}</label>
+			<input
+				aria-labelledby="${this._inputId}-label"
```
dlockhart, line 65: The `for` attribute on the label links these two together -- no need for `aria-labelledby` on the input (or the `id` on the label).

components/inputs/input-number.js
dlockhart: Actually... @AllanKerr what do you think here? With `d2l-form-native` I guess it'll do something?

krshnpatel-d2l: I added `name` because the `autocomplete` attribute requires it to map which fields can be autofilled. Not sure if a numeric input would be used for auto-filling things though.

AllanKerr: `name` is important for form submission. Without it `validate` will still work but the `formValue` won't get included in the submission.

dlockhart: Cool OK so sounds like it's worth keeping it around.

components/inputs/input-number.js
```diff
@@ -0,0 +1,83 @@
...
+			autofocus: { type: Boolean },
+			disabled: { type: Boolean },
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
+			max: { type: Number },
+			min: { type: Number },
+			name: { type: String },
+			placeholder: { type: String },
+			required: { type: Boolean },
+			step: { type: Number },
```
AllanKerr, line 22: `step` should be a `String` because it has the special value `"any"`

components/inputs/input-number.js
```diff
@@ -0,0 +1,83 @@
...
+
+	static get properties() {
+		return {
+			autocomplete: { type: String },
+			autofocus: { type: Boolean },
+			disabled: { type: Boolean },
+			label: { type: String },
+			labelHidden: { type: Boolean, attribute: 'label-hidden' },
+			max: { type: Number },
+			min: { type: Number },
```
AllanKerr, line 18: I think we also want to make these `String`s since `min` and `max` are string properties on `input`

dlockhart: Hmm. `max` and `min` on the generic input are weird though because they need to handle dates and times. In our case we know this is a number, so couldn't we make it a number and then convert it to a string when we set it on the input?
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/max

AllanKerr: I couldn't conclusively find if there were any special values allowed for `type="number"` that couldn't be represented with `Number` so figured it'd be safer than later finding out there is one that someone needs and then having to change it from `Number` to `String` once out in the wild.

But if we're not worried about that I'm fine with keeping it as a `Number`.

dlockhart: I'm not worried, especially since the table in the MDN docs lists `<Number>` as the syntax for numeric inputs. The HTML spec basically says that if the string parses to a number, then that's the minimum/maximum.

dbatiste: Scientific notation may be worth checking, both for the attributes, and also for the user specified value.

components/inputs/test/input-number.visual-diff.js
```diff
@@ -0,0 +1,95 @@
+const puppeteer = require('puppeteer');
+const VisualDiff = require('@brightspace-ui/visual-diff');
```
dlockhart, line 2: For the visual diff tests, I think we want to be careful that we're not creating a lot of tests that just do the same thing as the unit tests.

Unit tests are super fast and really easy to write, and are great for things like checking that the validation code is resulting in the errors we expect, or that setting a particular value is resulting in the formatted value we expect, events are firing when they should, etc.

Visual tests are slow, but are good at making sure a particular state is rendered properly. So I'd suggest only doing visual tests for things like: default, hidden label, required, disabled, placeholder, invalid (no focus), invalid (with focus), invalid (focus then blur -- if that was a bug you had trouble with). No need to have a visual test for every single type of validation error, or every single type of number output, since those are easy to write unit tests for.

components/inputs/input-number.js
```diff
@@ -0,0 +1,158 @@
...
+	_handleBlur() {
+		this.requestValidate(true);
+	}
+
+	async _handleChange(e) {
+		const value = e.target.value;
+		this._formattedValue = value;
+		await this.updateComplete;
+		const oldValue = this.value;
+		this.value = parseNumber(value);
```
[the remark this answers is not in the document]
krshnpatel-d2l: This expression would call the setter for `value` so if a `NaN` is passed in, `formatValue` will throw an error and set `_formattedValue = ''` and `value = undefined`.

krshnpatel-d2l:
Ready for review!

dlockhart (approved):

PR #781, opened 2020-08-11
refactor: Split d2l-form into d2l-form-native and FormMixin

AllanKerr:
# Context
Based on discussion in [this PR](https://github.com/BrightspaceUI/core/pull/742) we decided to split `d2l-form` into a `d2l-form-native` that handles native form submission and `d2l-form` that allows form nesting and requires the user to handle submission. This handles the refactor to split out the shared logic into the `FormMixin` and rename the `d2l-form` to `d2l-form-native`.

The new version of `d2l-form` will be coming in the next PR.

## Changes
1. Rename `d2l-form` to `d2l-form-native`.
1. Add `action`, `enctype`, `method`, and `target` to align with native forms.
1. Add `requestSubmit` to align with native forms.
1. Rename the `d2l-form-submit` event to `submit` to align with native forms.
1. Add a `formdata` event to allow users to get the raw `FormData` like in native forms.
1. Split any shared logic between `d2l-form-native` and the upcoming `d2l-form` into the new `FormMixin.`


components/form/form-errory-summary.js
```diff
@@ -53,6 +53,10 @@ class FormErrorSummary extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				margin-left: 0;
 				margin-right: 1.2rem;
 			}
+
+			d2l-alert {
+				max-width: none;
+			}
```
AllanKerr, line 59: This change is unrelated but got missed in previous PRs.

components/form/form-mixin.js
```diff
@@ -30,7 +27,9 @@ class Form extends LocalizeCoreElement(LitElement) {
 		this._validationCustoms = new Set();
 		this._errors = new Map();
 
-		this.addEventListener('d2l-form-errors-change', this._formErrorsChange);
+		this._form = document.createElement('form');
```
AllanKerr, line 30: This is created in the constructor so the setters in `d2l-form-native` can set properties on it. `firstUpdated` and `connectedCallback` are too late.

components/form/form-native.js
```diff
@@ -0,0 +1,148 @@
...
+		for (const entry of Object.entries(customFormData)) {
+			const input = document.createElement('input');
+			input.type = 'hidden';
+			input.name = entry[0];
+			input.value = entry[1];
+			this._form.appendChild(input);
+			tempInputs.push(input);
+		}
+		const submit = this.dispatchEvent(new CustomEvent('submit', { bubbles: true, cancelable: true }));
+		this.dispatchEvent(new CustomEvent('formdata', { detail: { formData: new FormData(this._form) } }));
```
AllanKerr, line 111: Normally calling `e.preventDefault()` on submit would prevent the `formdata` event from firing. That works with native forms because users can call `new FormData(e.target)`. Unfortunately that only works with a native form so unless we want to expose the internal native form in the `detail` we need to fire `formdata` even if default submission is prevented.

components/form/form-native.js
```diff
@@ -0,0 +1,148 @@
...
+			target: { type: String },
+		};
+	}
+
+	get action() {
+		return this._form.action;
+	}
+
+	set action(val) {
+		this._form.action = val;
```
dlockhart, line 44: What's the benefit to referencing the form's values vs. rendering the form as part of `render()` and then binding these as normal attributes on the form? Does the mixin really need to "own" the form?

AllanKerr: The form is special because its actually added as a child inside the slot, not rendered within `d2l-form-native`'s shadow DOM. Its done this way because native form elements need to be part of the form which is done via ID linking.

Capture

AllanKerr: I've changed this so the `form` is created just-in-time during submission rather than in the `constructor`. If it causes any issues with native browser validation I can always swap it back.

PR #782, opened 2020-08-12
tooltip: announce inner text when used with custom elements

margaree:
Added the `announced` attribute so that initially this is opt-in in order to avoid repetition when used with components that have already added their own `announce`.

`innerText` seems to work but I'm wondering if there is something else I should be using?

### DIFF
--- a/components/tooltip/README.md
+++ b/components/tooltip/README.md
@@ -77,6 +77,7 @@ In the following example to constrain the tooltip to the dashed boundary we can 
 ```
 
 **Advanced Properties:**
+* `announced` (Boolean) - announce the tooltip inner text to screen reader users when the tooltip is shown. Use with custom elements.
 * `boundary` (Object) - optionally provide boundaries to constrain where the tooltip will appear. Valid properties are `"top"`, `"bottom"`, `"left"` and `"right"`. The boundary is relative to the tooltip's [offset parent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent).
 * `close-on-click` (Boolean, default: `false`) - causes the tooltip to close when its target is clicked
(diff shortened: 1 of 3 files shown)


components/tooltip/tooltip.js
```diff
@@ -772,6 +778,7 @@ class Tooltip extends RtlMixin(LitElement) {
 			this.dispatchEvent(new CustomEvent(
 				'd2l-tooltip-show', { bubbles: true, composed: true }
 			));
+			if (this.announced && !this._isInteractive(this._findTarget())) announce(this.innerText);
```
AllanKerr, line 781: We should be able to use `_target` instead of `_findTarget()`.

AllanKerr, line 781: One concern here is that if `!this._isInteractive(this.__target)` returns `false` it doesn't guarantee the tooltip won't be announced, all it guarantees is that it doesn't announce properly on all browser / screen reader combinations which could cause some duplicate announcements.

Thoughts on removing `aria-describedby` if `this.announced && !this._isInteractive(this.__target)` is `true`?

components/tooltip/README.md
```diff
@@ -77,6 +77,7 @@ In the following example to constrain the tooltip to the dashed boundary we can
 ```
 
 **Advanced Properties:**
+* `announced` (Boolean) - announce the tooltip inner text to screen reader users when the tooltip is shown. Use with custom elements.
```
AllanKerr, line 80: There is an `accessibility` section in the README that should also be updated to basically say use this if only if none of the other options mentioned work.

components/tooltip/tooltip.js
margaree: Yup that seems like a good idea.

Issue #526, opened 2020-05-05
Announce accumulating messages


astadelbauer:
I'm trying to use announce to notify screenreaders when I delete rubrics from FACE one-by-one. 

This seems to work fine in NVDA+FF, but in Chrome, it accumulates the messages so if I delete one rubric, I get "Rubric removed", delete another "rubric removed rubric removed", and so on (also seems to do this with JAWS).  

I guess it happens because of announce re-using the same live region. If I wait the 10s between deleting each rubric for the live region to reset, it will only show the 1 message.

![image](https://user-images.githubusercontent.com/10656471/81083223-4c1d2700-8ec2-11ea-9a25-b267ac419b70.png)

PR #783, opened 2020-08-12
date and time inputs: add announced to tooltip

margaree:


AllanKerr:
I did notice its really easy to get duplicate announcements due to this issue so I think we're going to have to put a fix in: https://github.com/BrightspaceUI/core/issues/526

Issue #718, opened 2020-07-21
Disabled menu items not setting aria-disabled


dlockhart:
When disabling a `<d2l-menu-item disabled>`, while visually and functionally the item acts disabled, `aria-disabled="true"` isn't getting applied to the element with the `menuitem` role. So to screen readers, it does not announce that the item is disabled.

Sample resolved markup:

```html
<d2l-menu-item disabled class="d2l-contextmenu-item" text="Paste" role="menuitem" tabindex="-1">
    <!-- shadow root --->
</d2l-menu-item>
```

PR #786, opened 2020-08-12
fix: menu-item-mixin: add aria-disabled if disabled

margaree:
Fixes https://github.com/BrightspaceUI/core/issues/718


components/menu/menu-item-mixin.js
```diff
@@ -91,6 +91,9 @@ export const MenuItemMixin = superclass => class extends superclass {
 	}
 
 	__initializeItem() {
+		if (this.disabled) {
+			this.setAttribute('aria-disabled', 'true');
```
dlockhart, line 95: Should we do this in `updated()` instead so that changes to the property are reflected?

dlockhart: Could also have a private `_ariaDisabled` that binds to `aria-disabled` that we set from `updated()` -- that would avoid modifying DOM outside of render.

margaree: Agreed on the changing in `updated`. I'm not totally clear on what the suggestion is about binding to `aria-disabled` since `menu-item-mixin` doesn't have a `render` and it seems a bit complex to modify in each component that uses this mixin.

dlockhart: I just meant it could have a property:
```
_ariaDisabled: { type: String, attribute: 'aria-disabled' }
```

And then in the `updated()` for `disabled` it could go `this._ariaDisabled = this.disabled ? 'true' : 'false'` which would result in the same as going `this.setAttribute('aria-disabled', 'true')` except that you'd be letting Lit handle it for you.

PR #788, opened 2020-08-13
Handle input change, min/max values, min/max fraction digits, localization

krshnpatel-d2l:
This PR includes the following:
~ _handleChange method implementation
~ handle min/max values attributes (any input values less than min will be set to min and any input values more than max will be set to max)
~ handle min/max fraction digits attributes (input values will be updated based on min/max fraction digits on change)
~ handle number localization (input values will be formatted based on locale)

### DIFF
--- a/components/inputs/demo/input-number.html
+++ b/components/inputs/demo/input-number.html
@@ -59,7 +59,7 @@
 			<h2>Min and Max Fraction Digits</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-input-number label="Points" value="12345.67" min-fraction-digits="1" max-fraction-digits="2" style="width: 6rem;"></d2l-input-number>
+					<d2l-input-number label="Points" value="1.23" min-fraction-digits="1" max-fraction-digits="2"></d2l-input-number>
 				</template>
 			</d2l-demo-snippet>
 
--- a/components/inputs/demo/input-number.html
+++ b/components/inputs/demo/input-number.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="fr">
+<html lang="en">
 	<head>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta http-equiv="X-UA-Compatible" content="ie=edge">
--- a/components/inputs/demo/input-number.html
+++ b/components/inputs/demo/input-number.html
@@ -56,13 +56,6 @@
 				</template>
 			</d2l-demo-snippet>
 
-			<h2>Step</h2>
-			<d2l-demo-snippet>
-				<template>
-					<d2l-input-number label="Grade" value="85.76" step="0.01" style="width: 5rem;" ></d2l-input-number>
-				</template>
-			</d2l-demo-snippet>
-
 			<h2>Min and Max Fraction Digits</h2>
 			<d2l-demo-snippet>
 				<template>
--- a/components/inputs/demo/input-number.html
+++ b/components/inputs/demo/input-number.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="fr">
 	<head>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta http-equiv="X-UA-Compatible" content="ie=edge">
@@ -59,7 +59,7 @@
 			<h2>Min and Max Fraction Digits</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-input-number label="Points" value="8504.76" min-fraction-digits="1" max-fraction-digits="2" style="width: 6rem;"></d2l-input-number>
+					<d2l-input-number label="Points" value="12345.67" min-fraction-digits="1" max-fraction-digits="2" style="width: 6rem;"></d2l-input-number>
 				</template>
 			</d2l-demo-snippet>
 
--- a/components/inputs/demo/input-number.html
+++ b/components/inputs/demo/input-number.html
@@ -59,7 +59,14 @@
 			<h2>Step</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-input-number label="Grade" style="width: 5rem;" value="85.76" step="0.01"></d2l-input-number>
+					<d2l-input-number label="Grade" value="85.76" step="0.01" style="width: 5rem;" ></d2l-input-number>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>Min and Max Fraction Digits</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-number label="Points" value="8504.76" min-fraction-digits="1" max-fraction-digits="2" style="width: 6rem;"></d2l-input-number>
 				</template>
 			</d2l-demo-snippet>
 
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -106,7 +106,7 @@ class InputNumber extends LitElement {
 		const value = e.target.value;
 		this._formattedValue = value;
 		await this.updateComplete;
-		this.value = value;
+		this.value = parseNumber(value);
 	}
 }
 customElements.define('d2l-input-number', InputNumber);
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -1,9 +1,18 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { formatNumber, parseNumber } from '@brightspace-ui/intl/lib/number.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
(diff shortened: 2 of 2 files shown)


components/inputs/input-number.js
```diff
@@ -45,9 +62,18 @@ class InputNumber extends LitElement {
 		this.labelHidden = false;
 		this.required = false;
 
+		this._formattedValue = '';
 		this._inputId = getUniqueId();
 	}
 
+	firstUpdated(changedProperties) {
```
dlockhart, line 69: I actually think you want this code to run in `updated()`, since anytime the `value` is updated you want to re-check it and re-format it. For example (and good tests to add when you have them), running these should both result in the value being re-formatted which today it won't:

```
// setting the HTML attribute
elem.setAttribute('value', '39.2');

// setting the underlying property
elem.value = 39.2;
```

Another approach (which the time input uses) is to run this in a setter for `value`.

components/inputs/input-number.js
```diff
@@ -63,14 +89,24 @@ class InputNumber extends LitElement {
...
 	}
+
+	_handleChange(e) {
+		const inputValue = e.target.value.trim();
+		this.value = inputValue !== '' ? parseNumber(inputValue) : undefined;
+
+		if (!isNaN(this.value) && this.value !== undefined) {
+			this.value = checkMinMaxValue(this.value, this.min, this.max);
+			this._formattedValue = formatValue(this.value, this.minFractionDigits, this.maxFractionDigits);
+			e.target.value = this._formattedValue;
```
dlockhart, line 106: You never want to be setting `e.target.value`, since it's already bound to `this._formattedValue`. Simply setting `this._formattedValue` should be enough to trigger a re-render, which should set the value of the input.

Similarly, with the changes called out above, I don't think you even need to set `_formattedValue` since it should be a function of setting value.

components/inputs/input-number.js
```diff
@@ -63,14 +89,24 @@ class InputNumber extends LitElement {
...
+		} else {
+			e.target.value = '';
+		}
```
dlockhart, line 109: One piece you could add here (in this PR or a future one) is to fire `change` event if that actual value changes as a result of the change from the input.

components/inputs/input-number.js
```diff
@@ -63,14 +95,18 @@ class InputNumber extends LitElement {
 				class="d2l-input"
 				?disabled="${this.disabled}"
 				id="${this._inputId}"
-				max="${ifDefined(this.max)}"
-				min="${ifDefined(this.min)}"
 				name="${ifDefined(this.name)}"
 				placeholder="${ifDefined(this.placeholder)}"
-				step="${ifDefined(this.step)}"
```
AllanKerr, line 70: I'm assuming you removed `max`, `min`, and `step` because they don't work well with the formatted value?

components/inputs/input-number.js
```diff
@@ -45,9 +62,24 @@ class InputNumber extends LitElement {
...
+	get value() { return this._value; }
+	set value(val) {
+		const oldValue = this.value;
+		try {
+			const newValue = checkMinMaxValue(val, this.min, this.max);
+			this._formattedValue = formatValue(newValue, this.minFractionDigits, this.maxFractionDigits);
+			this._value = parseNumber(this._formattedValue);
+		} catch (err) {
+			this._formattedValue = '';
+			this._value = undefined;
```
AllanKerr, line 78: If theres an error do we want to wipe out the value or revert back to `oldVal`? Or maybe even allow it and display a type mismatch error?

components/inputs/demo/input-number.html
```diff
@@ -56,10 +56,10 @@ <h2>Min and Max Values</h2>
 				</template>
 			</d2l-demo-snippet>
 
-			<h2>Step</h2>
+			<h2>Min and Max Fraction Digits</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-input-number label="Grade" style="width: 5rem;" value="85.76" step="0.01"></d2l-input-number>
+					<d2l-input-number label="Points" value="12345.67" min-fraction-digits="1" max-fraction-digits="2" style="width: 6rem;"></d2l-input-number>
```
AllanKerr, line 62: nit: It'd be nice to remove the `style="width: 6rem"` from the demo. The reason being that people often copy and paste from demo examples.

components/inputs/input-number.js
```diff
@@ -1,10 +1,25 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { formatNumber, parseNumber } from '@brightspace-ui/intl/lib/number.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
 import { inputStyles } from './input-styles.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 
+function checkMinMaxValue(value, min, max) {
```
AllanKerr, line 9: nit: `check` makes me think it returns a `boolean` depending on if its within the range. Something like `clampValue` might be a bit more clear.

components/inputs/input-number.js
```diff
@@ -15,12 +30,14 @@ class InputNumber extends LitElement {
 			label: { type: String },
 			labelHidden: { type: Boolean, attribute: 'label-hidden' },
 			max: { type: Number },
+			maxFractionDigits: { type: Number, attribute: 'max-fraction-digits' },
 			min: { type: Number },
+			minFractionDigits: { type: Number, attribute: 'min-fraction-digits' },
```
AllanKerr, line 35: Do we need defaults for these?

components/inputs/input-number.js
```diff
@@ -63,14 +95,18 @@ class InputNumber extends LitElement {
...
+				type="text"
+				.value="${this._formattedValue}">
 		`;
 	}
+
+	async _handleChange(e) {
+		const value = e.target.value;
+		this._formattedValue = value;
+		await this.updateComplete;
+		this.value = parseNumber(value);
```
AllanKerr, line 109: I don't think we should be setting `_formattedValue` here. We should be able to wait for `set value()` to update `_formattedValue` so it doesn't get updated twice.

components/inputs/input-number.js
krshnpatel-d2l: So I removed `max`, `min`, and `step` values because they have no function with a 'text input' element. However, I still handle `min` and `max` values with the `checkMinMaxValue` method.

components/inputs/input-number.js
krshnpatel-d2l: Is type mismatch error something we can do in form validation? If yes, then I think that's the better route to go and I'll include that in my next PR (form validation stuff).

AllanKerr: If we want to display an error then it'd be done as part of integrating form validation. If you take a native number input and enter something like `+--33.....` then call `input.validity` you'll see the `badInput` flag is set to `true`.

But from playing around with native inputs it looks like they don't even allow you to type non-numeric characters so there isn't an exact parallel to what we're doing here.

components/inputs/input-number.js
krshnpatel-d2l: Nope

components/inputs/input-number.js
krshnpatel-d2l: So this is the only way to update the user's input when the input loses focus in certain cases.
For example: Assuming that the input has `max-fraction-digits=2`, the following flow would not work as expected if we don't set `_formattedValue` first then re-set it:
1. Initial value is `1.23`
2. User updates the value to `1.234`
3. The input value will remain `1.234` because no actual change in value was detected because the formatted value for `1.234` would result in `1.23` (same as initial value) due to `max-fraction-digits`

components/inputs/input-number.js
krshnpatel-d2l: So I don't think this numeric input would be exactly like the native number input because we would allow formatted numbers to be typed in as well based on locale (i.e. 3,400.23 in `en` or 30,40 in `fr`). However, throwing a validation error in this case would be better since the user may not notice that their input was changed to the "old value". I'm assuming we want an error like "Invalid value" or something when the user focuses out of the input field.

components/inputs/input-number.js
AllanKerr: Have you looked into why the `input` isn't rerendering the `_formattedValue` correctly? I would have expected the re-render from setting `value` to be enough. Even if you step through it shows the `render` is called with the correct `_formattedValue` but the value inside the `input` isn't correctly getting re-rendered.

krshnpatel-d2l: Yeep, I looked into this quite a bit. The problem is that the `render()` function will be called but the `input` element still detects no change and therefore will not be updated. I expected that setting `value` would be enough as well but I think only the elements with changed properties will be re-rendered.

components/inputs/input-number.js
AllanKerr: See the discussion [here](https://github.com/BrightspaceUI/core/pull/788#discussion_r475856822) about `_formattedValue` being a function of `set value()`

PR #792, opened 2020-08-14
refactor: Additional FormElementMixin Simplifcations

AllanKerr:
# Changes
- Remove `FormElementMixin.setCustomValidity`
    - This is redundant with the ability to override `validationMessage` and setting `setValidity({ customError: true })` so removing it to simplify the interface.
- Add `FormElementMixin` `name` property
    - Name is used by all form elements during submission so it should be a property of all of them rather than having to add it explicitly to every single form element
- Remove `FormElementMixin.checkValidity`
    - This is redundant with `validity.valid` so removing it to simplify the interface

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -78,10 +78,6 @@ class InputText extends FormElementMixin(RtlMixin(LitElement)) {
 			 */
 			minlength: { type: Number },
 			/**
-			 * Name of the input
-			 */
-			name: { type: String },
-			/**
 			 * Regular expression pattern to validate the value
 			 */
 			pattern: { type: String },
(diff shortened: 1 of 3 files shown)


components/inputs/input-text.js
```diff
@@ -77,10 +77,6 @@ class InputText extends FormElementMixin(RtlMixin(LitElement)) {
 			 * Imposes a lower character limit
 			 */
 			minlength: { type: Number },
-			/**
-			 * Name of the input
-			 */
-			name: { type: String },
```
dlockhart, line 83: @krshnpatel-d2l FYI you'll want to remove `name` from numeric input as you'll get it for free now from the FormElementMixin.

PR #796, opened 2020-08-17
Another "Skeleton" loading approach for discussion

mdgbayly:
Just following on from some of the discussions we've had about coming up with a re-usable pattern/solution for implementing the Skeleton Page Loading effects documented on http://design.d2l/patterns/skeleton-loading/

As mentioned previously, we need a solution for FACE that will allow us to implement the design shown in the mockup linked here:
https://xd.adobe.com/view/399d5110-bf02-401d-9041-c60fcc11b1c2-6cce/

I looked at the solutions being proposed by both Maya and Chris. They are both good solutions and are not mutually exclusive to the solution shown in this PR.  However, the approach shown in this PR is one that fits quite nicely into how FACE works as it is quite lightweight and un-opinionated, and just uses a couple of CSS classes to style potentially any (OK, probably reaching there! :-) ) component.

The pattern shown in this PR boils down to styling a component with a css `::before` pseudo selector that will render a skeleton block that is absolutely positioned to cover the full dimensions of the component being skeletized. In this way the skeleton automatically adapts to the size of the component being skeletized based on that component's normal styling and so we don't have to add any other hints to `skeleton` components indicating what size they should be rendered at or conditionally render skeleton placeholder components instead of the normal components, both of which can lead to duplicated code. (see `skeleton-styles.js`)

This approach also makes it easy to skeletize more native leaf components like `label` and `d2l-text-input`

The `skeletize` class can be coupled with a `skeleton` attribute host selector such that the `skeletize` styles are only applied when the host component is rendering in `skeleton` mode.

Within FACE, we have a lot of component nesting and each sub component may be a self-contained hypermedia resource enabled component. During loading, we need each sub component to be rendered so that it can kick off it's own hypermedia API fetches. So in most cases, we can't replace a component with a dummy skeleton object as that will delay the overall fetch. Also each sub-component is in the best position to know what might be an appropriate skeleton rendering for that component. 

So in many cases it's sufficient to just apply the `skeletize` class to a sub-component (e.g. for labels, text inputs, buttons). But where the sub-component might want to take more control, it can support the `skeleton` property too, and then it can use that property to apply styling internally and potentially conditionally render certain aspects of it's own children. 

e.g. we have a `Score` component in FACE that renders both score and grade information. During loading we don't know yet whether there is just a score or score and grade, so the `Score` component only renders a skeleton view of the `score` input field, but it can do so just by adding a `skeletize` class to that input.

Anyway, let me know your thoughts. No intention for requesting this PR be merged or become part of core at this point. I do think that different scenarios might require different approaches and that for some scenarios the approaches shown by Maya and Chris might be more appropriate. But this approach seems to be a better fit for FACE and web forms, so unless you guys can see explicit red flags e.g. due to a11y issues or cross browser issues, we will most likely proceed with this approach in FACE at least for an initial implementation. I've tried the demo page included here across a number of desktop and mobile browsers and so far have not hit any major blockers.

### DIFF
--- a/components/skeleton/demo/index.html
+++ b/components/skeleton/demo/index.html
@@ -1,0 +1,108 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+	<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '../../demo/demo-page.js';
+		import '../../button/button.js';
+		import '../../inputs/input-text.js';
+		import '../../inputs/input-date.js';
+		import './skeleton-host.js';
+	</script>
+	<style>
+		@keyframes loadingPulse {
+			0% { background-color: var(--d2l-color-sylvite); }
+			50% { background: var(--d2l-color-regolith); }
+			75% { background: var(--d2l-color-sylvite); }
+			100% { background: var(--d2l-color-sylvite); }
+		}
+
+		.skeletize::before {
+			content: "";
+			position: absolute;
+			top: 0;
+			bottom: 0;
+			right: 0;
+			left: 0;
+			border-radius: 0.2rem;
+			animation: loadingPulse 1.8s linear infinite;
+			background-color: var(--d2l-color-sylvite);
+			z-index: 2000;
+		}
+
+		.skeletize {
+			position: relative;
+			color: transparent;
+			border: none;
+			box-shadow: none;
+		}
+
+		.flex-container {
+			display: flex;
+			flex-wrap: wrap;
+		}
+
+		.item-container {
+			margin-right: 40px;
+		}
+
+		.item-container[dir="rtl"] {
+			margin-right: 0;
+			margin-left: 40px;
+		}
+
+	</style>
+</head>
+
+<body unresolved>
+	<d2l-demo-page page-title="d2l-skeleton">
+
+		<h2>Custom Element handling skeleton behaviour internally</h2>
+		<d2l-demo-snippet>
+			<template>
+				<d2l-skeleton-host skeleton>
+				</d2l-skeleton-host>
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Label & Date</h2>
+		<d2l-demo-snippet>
+			<template>
+				<label class="skeletize d2l-label-text">Birth date</label>
+				<d2l-input-date class="skeletize" style="display:block" label="Date" label-hidden></d2l-input-date>
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Label & Text Input</h2>
+		<d2l-demo-snippet>
+			<template>
+				<label class="skeletize d2l-label-text">Hello you old skeleton you</label>
+				<d2l-input-text class="skeletize"></d2l-input-text>
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Flex Items</h2>
+		<d2l-demo-snippet>
+			<template>
+			<div class="flex-container">
+				<div class="item-container">
+					<label class="skeletize d2l-label-text">Hello</label>
+					<d2l-input-text size="6" class="skeletize"></d2l-input-text>
+				</div>
+				<div class="item-container">
+					<label class="skeletize d2l-label-text">You lovely skeletal world</label>
+					<d2l-input-text size="6" class="skeletize"></d2l-input-text>
+				</div>
+			</div>
+			</template>
+		</d2l-demo-snippet>
+
+	</d2l-demo-page>
+</body>
+
+</html>
--- a/components/skeleton/demo/index.html
+++ b/components/skeleton/demo/index.html
@@ -56,6 +56,10 @@
 			margin-left: 40px;
 		}
 
+		d2l-input-text {
+			display: block;
+		}
+
 	</style>
 </head>
 
@@ -92,11 +96,11 @@
 			<div class="flex-container">
 				<div class="item-container">
 					<label class="skeletize d2l-label-text">Hello</label>
-					<d2l-input-text size="6" class="skeletize"></d2l-input-text>
+					<d2l-input-text class="skeletize"></d2l-input-text>
 				</div>
 				<div class="item-container">
 					<label class="skeletize d2l-label-text">You lovely skeletal world</label>
-					<d2l-input-text size="6" class="skeletize"></d2l-input-text>
+					<d2l-input-text class="skeletize"></d2l-input-text>
 				</div>
 			</div>
 			</template>
--- a/components/skeleton/demo/skeleton-host.js
+++ b/components/skeleton/demo/skeleton-host.js
@@ -1,0 +1,38 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { skeletonStyles } from '../skeleton-styles.js';
+
+class SkeletonHost extends LitElement {
+
+	static get properties() {
+		return {
+			skeleton: { type: Boolean, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [
+			skeletonStyles,
+			css`
+				:host {
+					display: block;
+				}
+
+				:host([skeleton]) ul {
+					height: 1rem;
+				}
+			`
+		];
+	}
+
+	render() {
+		return html`
+			<label class="skeletize">I am alive</label>
+			<ul class="skeletize">
+				<li>One</li>
+				<li>Two</li>
+				<li>Three</li>
+			</ul>
+		`;
+	}
+}
+customElements.define('d2l-skeleton-host', SkeletonHost);
--- a/components/skeleton/skeleton-styles.js
+++ b/components/skeleton/skeleton-styles.js
@@ -1,0 +1,30 @@
+import { css } from 'lit-element/lit-element.js';
+
+export const skeletonStyles = css`
+	@keyframes loadingPulse {
+		0% { background-color: var(--d2l-color-sylvite); }
+		50% { background-color: var(--d2l-color-regolith); }
+		75% { background-color: var(--d2l-color-sylvite); }
+		100% { background-color: var(--d2l-color-sylvite); }
+	}
+
+	:host([skeleton]) .skeletize::before {
+		content: "";
+		position: absolute;
+		top: 0;
+		bottom: 0;
+		right: 0;
+		left: 0;
+		border-radius: 0.2rem;
+		animation: loadingPulse 1.8s linear infinite;
+		background-color: var(--d2l-color-sylvite);
+		z-index: 2000;
+	}
+
+	:host([skeleton]) .skeletize {
+		position: relative;
+		color: transparent;
+		border: none;
+		box-shadow: none;
+	}
+`;
--- a/index.html
+++ b/index.html
@@ -131,6 +131,11 @@
 		<li><a href="mixins/demo/localize-mixin.html">localize-mixin</a></li>
 	</ul>
 
+	<h2 class="d2l-heading-3">Skeletons</h2>
+	<ul>
+		<li><a href="components/skeleton/demo/index.html">Skeletons</a></li>
+	</ul>
+
 	<h2 class="d2l-heading-3">Templates</h2>
 	<ul>
 		<li><a href="templates/primary-secondary/demo/index.html">Primary Secondary Template (fullscreen width)</a></li>
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lint:lit": "lit-analyzer \"{components,directives,helpers,mixins,templates,test,tools}/**/*.js\" --strict",
     "lint:style": "stylelint \"**/*.js\"",
     "prepublishOnly": "node ./cli/verify-translations.js && npm run analyze",
-    "start": "es-dev-server --node-resolve --dedupe --watch --open",
+    "start": "es-dev-server --node-resolve --dedupe --watch --open --hostname 192.168.1.91",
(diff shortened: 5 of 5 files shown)


components/skeleton/demo/index.html
```diff
@@ -0,0 +1,112 @@
...
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '../../demo/demo-page.js';
+		import '../../button/button.js';
+		import '../../inputs/input-text.js';
+		import '../../inputs/input-date.js';
+		import './skeleton-host.js';
+	</script>
+	<style>
+		@keyframes loadingPulse {
```
mdgbayly, line 18: For the demo, I just copied the skeleton styles to the demo page as didn't want to spend time making them work both in a component host and here. But ordinarily these would only be needed within a host using the the `skeleton-styles.js` style module.

components/skeleton/skeleton-styles.js
```diff
@@ -0,0 +1,30 @@
+import { css } from 'lit-element/lit-element.js';
+
+export const skeletonStyles = css`
+	@keyframes loadingPulse {
+		0% { background-color: var(--d2l-color-sylvite); }
+		50% { background-color: var(--d2l-color-regolith); }
+		75% { background-color: var(--d2l-color-sylvite); }
+		100% { background-color: var(--d2l-color-sylvite); }
+	}
+
```
mdgbayly, line 10: This is the core of the solution using an absolutely position before style to cover the component being styled.

components/skeleton/skeleton-styles.js
```diff
@@ -0,0 +1,30 @@
...
+		top: 0;
+		bottom: 0;
+		right: 0;
+		left: 0;
+		border-radius: 0.2rem;
+		animation: loadingPulse 1.8s linear infinite;
+		background-color: var(--d2l-color-sylvite);
+		z-index: 2000;
+	}
+
```
mdgbayly, line 23: Possibly? a little hacky but this style sets the component being styled to `position: relative` to anchor the absolutely positioned `::before` selector. Not sure if that could cause issues in other places where position is being used for other things, but it works for us in FACE for now. Also this style includes other overrides that might 'bleed' out of the skeleton like borders, box-shadows etc.

components/skeleton/demo/skeleton-host.js
```diff
@@ -0,0 +1,38 @@
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { skeletonStyles } from '../skeleton-styles.js';
+
+class SkeletonHost extends LitElement {
```
mdgbayly, line 4: This is just a demo component showing how other custom components could support the `skeleton` property and then internally use combinations of the `skeleton` attribute on the host and the `skeletize` class to style their own internal content.

e.g. here the component has a list, but as it might not know how long this list will be until loaded it just renders the list at a fixed height. (This example is taken directly from the right hand panel summarizers that we have in FACE, where we only want to render a single block to represent the potential content - other scenarios might want to represent a 'pseudo list' in which case the approaches shown by Maya/Chris might be more appropriate).

components/skeleton/skeleton-styles.js
dbatiste: This seems reasonable to me.

Yeah I am a little concerned about clobbering styles on the component as you mentioned.  Perhaps those can be addressed on a case by case basis - for example, in such case, the `skeletize` class would perhaps be rendered on another nested element.

What's the rationale for the `z-index: 2000;` value?  I think this could cause the skeleton to be displayed on-top of some other elements that we would not normally expect it to.  For this skeleton approach, I would expect a `z-index` value of something like `1-10` range.

I think it would be appropriate to also define a media query to respect user preference for reduced motion.
```
@media (prefers-reduced-motion: reduce) {
	:host {
		animation: none;
	}
}
```

dbatiste: After thinking about this more, I suspect a high z-index is actually what we want to ensure everything gets covered in the skeletized element.  Still worthwhile knowing why `2000` was picked though and if this were merged also applying a brief comment to avoid regressions etc.

mdgbayly: Yeh- the high z-index was kind of plucked randomly out of thin air because when I was first experimenting with this idea, the skeleton overlay wasn't working for masking things like regular input fields. I meant to go back and dig more into it at some point. I think even just setting a z-index of 1 on the `::before` works around that issue.

But it's things like the z-index that make me feel a little bit uncomfortable about this whole approach - and the use of the ::before pseudo selector in general makes me nervous - mostly because I don't have much experience with using it and coupling it with absolute positioning just seems like it's pushing the boundaries...not sure how consistent browsers are at implementing such things and whether the approach I'm using is a "designed css feature" or just happens to be working in a way that I hoped it would.

Initially I started with the idea of directly rendering an actual `d2l-skeleton` component in each custom element when it was in `shadow` mode and trying to get that `d2l-skeleton` to `fill` the component in such a way that you didn't have to explicitly set dimensions on it. But then it just occurred to me that using `::before` with `content: ""` might be a simple and elegant? way to achieve a similar effect. But I was mighty surprised when it just seemed to work.

components/skeleton/demo/skeleton-host.js
dlockhart: Could probably also be exposed as a simple `SkeletonMixin` that components could extend -- with it they'd get the `skeleton` property (ensuring naming consistency across components) and it would load the `skeletonStyles` automatically enabling them to use the CSS class.

components/skeleton/skeleton-styles.js
dlockhart: I think it's a really elegant solution, and one I'm not sure I would have thought of! I'm less worried about the style clobbering and agree we could just tackle any issues that come up with it case-by-case.

The `z-index` is worrying, but I think as long as we tested it out with a bunch of places we also leverage `z-index` (dropdowns, tooltips & dialogs come to mind) hopefully we can sort that out. We've been talking about having some kind of "z-index manager" that helps us keep track of where various things layer in... if things get hairy we could always look into that.

components/skeleton/demo/index.html
```diff
@@ -0,0 +1,112 @@
...
+				<d2l-skeleton-host skeleton>
+				</d2l-skeleton-host>
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Label & Date</h2>
+		<d2l-demo-snippet>
+			<template>
+				<label class="skeletize d2l-label-text">Birth date</label>
+				<d2l-input-date class="skeletize" style="display:block" label="Date" label-hidden></d2l-input-date>
```
dlockhart, line 81: The duplicate labels here isn't great, and probably won't pass an accessibility audit. We can't have a `<label>` just floating around not associated with a form element, but due to shadow DOM problems we also can't actually link that label with the underlying text input. That's why all the `d2l-input-*` components come with their own label.

But that's OK -- we can just make these components "skeleton aware", and then you could do something like:

```html
<d2l-input-date skeleton style="display:block" label="Birth date"></d2l-input-date>
```

components/skeleton/demo/index.html
```diff
@@ -0,0 +1,112 @@
...
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Label & Text Input</h2>
+		<d2l-demo-snippet>
+			<template>
+				<label class="skeletize d2l-label-text">Hello you old skeleton you</label>
+				<d2l-input-text class="skeletize"></d2l-input-text>
```
dlockhart, line 89: Similarly to above, this will fail in a different way -- `d2l-input-text` must have a label specified on it.

mdgbayly:
@dbatiste @dlockhart We are going to move ahead with using this solution in FACE assignments (until we hit a roadblock that makes me realize how naive I was). I like the idea of using a mixin to get the skeleton property and styles. Never tried mixing in styles before but I guess it should work just like mixing in properties?

I was wondering whether we should put our mixin/styles in BrightspaceUILabs but it seems kind of onerous to setup a whole new repo just for this.... unless we were also maybe to use it for adding some of the other 'skeleton' approaches being tried by other teams. The alternative is we just add it to our BrightspaceHypermediaComponent/activities repo for now.
Thoughts?

dbatiste:
> Never tried mixing in styles before but I guess it should work just like mixing in properties?

Styles are not automatically mixed the way properties are.  If the consumer is using the `SkeletonMixin` in their component, they'd need to do something like this:

```
class SomeComponent extends SkeletonMixin(LitElement) {

    static get styles() {
        return [ super.styles, css`
            :host { ... }
        `];
    }

}
```

My 2 cents re where this should reside: I can certainly imagine the skeleton pattern being used more, and we'd like to control that centrally.  However, we definitely want to avoid breaking changes in `core`, so if we're say less than 90% confident in the approach, then I'd hold off on core for the time being.  Although creating it in `labs` would be onerous, it has the advantage that one or two other apps might use it and expose the wrinkles, so that we can iron them out and feel more confident about moving it into `core`.  If it's just added to `activities`, it addresses the immediate need, but might not expose the limitations because it's tailored to your use case alone, with the added risk that someone else copies it with some modifications, or goes in a different direction altogether, making it that much harder to align down the road.  So, I'd be in favor of `core` (assuming high degree of confidence in approach), or labs.  I do like this solution,  but I'd like to better understand the use-case that Maya had for the solution she used - I think that is worth flushing out before committing to `core`.

Re the inputs and labels: Carin might not notice an issue if she's keyboarding through the page because focus would land on the internal input, which is labelled by the `aria-label`.  But, there would be an actual native `label` on the page that's pointing to an element that is not a form field.  What happens when a user clicks on the native label?  Normally we'd expect focus to move to the native `input` it labels.  Some a11y issues could result if this behavior does not work - for instance, if a sighted person tells the AT to move focus referring to the label.

components/skeleton/skeleton-styles.js
zinabat: I think using a `::before` pseudo-element is elegant and will fit for many use cases. It does remind me of early web dev days, though 😂

components/skeleton/demo/skeleton-host.js
zinabat: I like the idea of a mixin because it can help to standardize when we want to create more complex skeleton components.

components/skeleton/skeleton-styles.js
m6jones: Oh, this is interesting idea and I like it.

dlockhart:
I think the immediate need for this in text and date inputs for FACE is enough justification for us to take a bit of extra time now to get it into core. I can't see a good way for you to experiment with it outside of core if core components need to use it.

Dave's right though that if this ends up not working well, we definitely don't want to add it to core only for it to be abandoned or dramatically changed after it's been adopted in a bunch of places. Having said that, I feel fairly confident that at the very least we'll need some shared styles and a mixin that for components to skeleton-ify themselves... which the solution discussed here covers. If later we need to supplement with other styles and/or components, that wouldn't be a breaking change. This solution also elegantly solves simply applying the skeleton styles to a piece of text, which I think was a lot of what Maya's use cases where.

@m6jones and @ChrisKraljevicD2L, I know there's a lot to follow in this PR, but I'm curious about your thoughts and whether you feel the proposed solution would work for the situations you've used skeletons in the past?

components/skeleton/skeleton-styles.js
m6jones: Also worried about z-index

m6jones: It is nice to have skeletons over things. So when they hide there is not flicker.

components/skeleton/demo/skeleton-host.js
ChrisKraljevicD2L: I agree mixins are super handy for standardizing implementations of features for components like this. My concern is consumers of the mixin overriding property/function behaviour from it or utilizing properties and functions in odd ways.

The new members of Team Phoenix who haven't worked with polymer or lit before have all expressed frustration with the "opaqueness" of mixins. Mainly not being able to figure out where certain properties or functions come from when they're referenced or overridden in mixin consumers. That's the source of my concerns, otherwise a skeleton mixin would be a huge boon for this feature.

ChrisKraljevicD2L (approved):
Overall this looks super neat and I'm down for it :+1:

components/skeleton/demo/skeleton-host.js
m6jones: I sort of prefer this not to be a mixin. These don't need access to lifecycle lit so having a mixin might be over kill.

dlockhart: It doesn't need access to lifecycle methods... _yet_. :)

Without the mixin, components that want to be skeleton-able would need to import the styles themselves and add their own `skeleton` property. Not a big deal... but having the mixin also doesn't prevent them from doing those two things manually either.

m6jones (commented):
Overall, I think this is great. I don't think it will compete with how I thought of things but is the first step of making that happen.

I would still like to see the typography styles define the their skeleton height since it 100% based on how big the text is. 

Another thing if we are skeletize text won't match the exact height that it should.

mdgbayly:
> I would still like to see the typography styles define the their skeleton height since it 100% based on how big the text is.
> 
> Another thing if we are skeletize text won't match the exact height that it should.

Yeh, this approach is definitely only a starting point and I have not attempted to take things to the level of detail and precision as some of the ideas presented on http://design.d2l/patterns/skeleton-loading/. But I got the impression from our initial group discussions that the specs on design.d2l are not necessarily fully signed off yet? But yes, as things stand in this initial PR, if you apply a `skeletize` class to a text content like label or heading, it won't render at the `cap-height` as specified on design.d2l. But i think those are things that could potentially be added maybe in conjunction with typography styles.

But doing things like rendering a multi-line text content placeholder, for example, for a paragraph element where you want to show 3 lines and a ragged last line, are not going to "just work" with this approach, but it could possibly be used as the basis for achieving something like that.

But this approach does work (for now!) for things like FACE assignments where our requirements for the full skeleton rendering are probably a bit simpler.

dlockhart:
I'm going to close this PR as the majority of @mdgbayly's awesome work here has made it into our initial skeleton stuff in core. Thanks Martin (and others!) for the contribution and great discussion! 🎉

PR #797, opened 2020-08-18
Set up GitHub dependabot config.

dbatiste:


dlockhart:
We have a few ignore rules set up right now with the web version that we likely need to maintain:

```
    versions:
    - ">= 2.a"
    - "< 3"
    versions:
    - ">= 3.a"
    - "< 4"
  - dependency-name: intl-messageformat
    versions:
    - ">= 8.a"
    - "< 9"
  - dependency-name: intl-messageformat
    versions:
    - ">= 9.a"
    - "< 10"
  - dependency-name: karma-browserstack-launcher
    versions:
    - ">= 1.5.a"
    - "< 1.6"
  - dependency-name: karma-browserstack-launcher
    versions:
    - ">= 1.6.a"
    - "< 1.7"
  - dependency-name: karma-sauce-launcher
    versions:
    - ">= 4.a"
    - "< 5"
  - dependency-name: mocha
    versions:
    - ">= 8.1.a"
    - "< 8.2"
```

dbatiste:
A couple questions re those ignore statements:

* do you know why a couple of those are repeated (different versions)?  do we just want to ignore these altogether?
* do you know what the `a` refers to in the version?  is it the same as `x`?

If we want to ignore them altogether, we can just exclude the `versions`, which simplifies things.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
```
dbatiste, line 7: I've added these ignore rules (ignoring version).  I'm not sure if the explicit version numbers are needed.  It sorta seems like they are just ignoring the dependency altogether, except when a new major is released.  I can add them if we need them - if so I think this...
```
    versions:
    - ">= 2.a"
    - "< 3"
    versions:
    - ">= 3.a"
    - "< 4"
```
could become...
```
    versions: [">= 2.a", "< 3", ">= 3.a", "< 4" ]
```
or...
```
    versions:
    - ">= 2.a"
    - "< 3"
    - ">= 3.a"
    - "< 4"
```

dlockhart: Yeah I'm not really sure what it all means. These got created through the GitHub comments we made on dependabot PRs, so that could explain the duplicates. 

Basically `intl-pluralrules` we just want it to stick at `^1` for now. I'm not sure about the others or if they're still needed (@margaree added them somewhat recently so probably), but it looks like Mocha just needs to be `< 8.2`, `karma-sauce-launcher < 5` and `karma-browserstack-launcher < 1.5`.

margaree: I think Mocha needs to be `< 8.1` (I believe a fix has been merged for the problem I was seeing but there hasn't been a release) but otherwise those look right.

dlockhart: Does the YML support comments? Wondering if we could at least put in comments for each one explaining/linking to the issue.

dbatiste: Yep, we can certainly include comments.  

Re whether we include explicit versions, I think if we exclude them, we won't get PRs when new versions are released, whereas if we do specify versions, we will still get PRs for versions that fall outside of the ranges being ignored, so perhaps nice to be notified, but perhaps also a nuisance to maintain when new version continue to have the issue.  This is their example from docs...
```
    ignore:
      - dependency-name: "express"
        # For Express, ignore all updates for version 4 and 5
        versions: ["4.x", "5.x"]
        # For Loadash, ignore all updates
      - dependency-name: "loadash"
```
So I _think_ if we want to keep versions, I think it could look like (plus comments):
```
    ignore:
        versions: ["2.x", "3.x"]
      - dependency-name: "intl-messageformat"
        versions: ["8.x", "9.x"]
      - dependency-name: "karma-browserstack-launcher"
        versions: ["1.5.x", "1.6.x"]
      - dependency-name: "karma-sauce-launcher"
        versions: ["4.x"]
      - dependency-name: "mocha"
        versions: ["8.1.x"]
```

PR #799, opened 2020-08-21
Dbatiste/migrate card

dbatiste:
Migration of the [card components](https://github.com/BrightspaceUI/card).  This is pretty much a direct port of those components with adjustments made for Polymer vs. Lit implementation.  It should allow removal of all components in card repo.


components/card/card-footer-link.js
```diff
@@ -0,0 +1,178 @@
...
+
+	render() {
+		return html`
+			<a ?download="${this.download}"
+				href="${ifDefined(this.href)}"
+				hreflang="${ifDefined(this.hreflang)}"
+				rel="${ifDefined(this.rel)}"
+				target="${ifDefined(this.target)}"
+				type="${ifDefined(this.type)}">
+				<span class="d2l-card-footer-link-text d2l-offscreen">${this.text}</span>
```
svanherk, line 153: Is this class still needed?  Seems like it only handling the offscreen before

components/card/card.js
```diff
@@ -0,0 +1,337 @@
...
+			:host([subtle]:not([href])) {
+				box-shadow: none;
+				transform: none;
+			}
+			@media (prefers-reduced-motion: reduce) {
+				:host {
+					transition: none;
+				}
+			}
+		`];
```
svanherk, line 208: Nice!

components/card/card.js
```diff
@@ -0,0 +1,337 @@
...
+		this._hover = false;
+		this._tooltipShowing = false;
+		this._onBadgeResize = this._onBadgeResize.bind(this);
+		this._onFooterResize = this._onFooterResize.bind(this);
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		const badgeObserver = new ResizeObserver(this._onBadgeResize);
+		badgeObserver.observe(this.shadowRoot.querySelector('.d2l-card-badge'));
```
svanherk, line 228: Do we need to worry about unobserving these?

dbatiste: Good question.  I don't think so, since the observer is scoped to this component and will get cleaned up along with attached handlers when the component is garbage collected.

components/card/card.js
```diff
@@ -0,0 +1,337 @@
...
+	_onDropdownOpen() {
+		this._dropdownActionOpen = true;
+	}
+
+	_onFooterResize(entries) {
+		if (!entries || entries.length === 0) return;
+		const entry = entries[0];
+		// firefox has a rounding error when calculating the height of the contentRect
+		// with `box-sizing: border-box;` so check for numbers which are close to 0 as well
+		this._footerHidden = (entry.contentRect.height < 1);
```
svanherk, line 308: I think the logic here has _technically_ changed, but is definitely worth the simplification.  I don't think we'll run into a situation where the footer was previous hidden, and then gets something 0.5 pixels high that we want to show.

components/card/card-footer-link.js
dbatiste: Thank!  Good catch!  Will update.

components/card/card.js
dbatiste: Cool yep... thanks for noticing that!  I should have called that out as something that changed.

PR #805, opened 2020-08-25
fix: Fixing inability to click actions on Edge

zinabat:
## Context
[DE40163](https://rally1.rallydev.com/#/detail/defect/414252830536?fdp=true)
A regression was introduced in #752 that is fixed in this adjustment. The regression prevented Legacy Edge users from clicking on navigation links in list components as well as drag n drop. The fix ensures the original bugfix is maintained.

### Example of maintained fix that prevents content from overstretching its bounds
Screen Shot 2020-08-25 at 2 23 34 PM

### Screenshot of manual testing with Legacy Edge
Screen Shot 2020-08-25 at 2 31 16 PM

## Quality
* Manually tested with demo pages in Chrome and Legacy Edge
* Manually tested with local BSI and LE


components/list/list-item-generic-layout.js
```diff
@@ -78,8 +78,6 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 
 			::slotted([slot="content"]) {
 				grid-column: content-start / content-end;
-				overflow-x: hidden;
-				position: relative;
```
zinabat, line 82: This line was the primary offender.

PR #806, opened 2020-08-25
fix: defect DE40213 - Rubric buttons hidden behind face footer

adkulas:
# Overview
This PR is to fix a defect outlined here: 
- [DE40213](https://rally1.rallydev.com/#/detail/defect/416279793680?fdp=true): FACE > Rubrics > Create new rubric overlay buttons are hidden behind the FACE footer

Based on this [discussion](https://d2l.slack.com/archives/CS52A4T3M/p1597696256006000) it is likely the issue was created from #770 
## Setup
- Enable the FACE Feature flag and FACE config to on/opt in
- Run local BSI with the following repos
  - [@brightspace-ui/core](https://github.com/BrightspaceUI/core)

## Recreate Defect
- Run above local repos on the master branch.
- Navigate to `Course > Assignments > New Assignment > Evaluation & Feedback > Add Rubric > Create New`
- Scroll to bottom of page and observe header cover dialog closing "X" and footer covers bottom of overlay
defect

## Proposed Fix
- Changing the `z-index` of the `.d2l-template-primary-secondary-content` back to default value of auto.
Buttons Visible

## Methodology
Some properties for z-index
- The `z-index` for every html element is `auto` by default (evaluates to 0).
- `z-index` accepts negative numbers
- An element inherits its parents `z-index` if there is one.
- A child cannot override its parent's `z-index`
https://stackoverflow.com/questions/16057361/how-to-make-child-element-higher-z-index-than-parent 

I believe the original intent of #770 was to set the stacking order of the header and footer to be higher than the content. However, for things like `d2l-simple-overlay` (which is used in the defect above) the component sets its `z-index` to 1000 to place its content over all other elements. This is a problem because the component has a parent which has set an `z-index` of 1 and therefore inherits it's `z-index`. 

Having only the header and footer set a `z-index` will result in the same stacking priority but also give components in main the ability to set `z-index` when needed.

### Recreating the problem documented in #770 
- Execute: 
```shell
git checkout c80e1d8fccf5c371f879ec351ecedc46a8eab526 # commit prior to #770
npm start
```
- Navigate to: [Primary Secondary Template (fullscreen width)](http://localhost:8001/templates/primary-secondary/demo/index.html)
  - `box-shadow` renders properly with no `background-color`
image
- In the dev console Add a background color to the primary section
  - Example `<div slot="primary" style="background-color: #fefefe;">`
- Observe the `box-shadow` disappears on the primary section and not the secondary section:
image

### Verify PR #770 Corrects issue using same steps
- Execute: 
```shell
git checkout 55638ec54907879bcfc6242d78319e701536b546  # commit for PR#770
npm start
```
- Navigate to: [Primary Secondary Template (fullscreen width)](http://localhost:8001/templates/primary-secondary/demo/index.html)
- In the dev console Add a background color to the primary section
  - Example `<div slot="primary" style="background-color: #fefefe;">`
- Observe the `box-shadow` is visible primary section:
image
- This commit has the z-index changes:
```css
			header {
				grid-area: header;
				z-index: 2;
			}
```
```css
			.d2l-template-primary-secondary-content {
				display: grid;
				grid-area: content;
				grid-template-areas: "primary divider secondary";
				grid-template-columns: minmax(320px, 2fr) 1px minmax(320px, 1fr);
				grid-template-rows: auto;
				overflow: hidden;
				z-index: 1;
			}
```
```css
			footer {
				background-color: #ffffff;
				box-shadow: 0 -2px 4px rgba(73, 76, 78, 0.2); /* ferrite */
				grid-area: footer;
				padding: 0.75rem 1rem;
				z-index: 2; /* ensures the footer box-shadow is over main areas with background colours set */
			}
```

### Checking this change maintains the `box-shadow` fix
- Execute: 
```shell
git checkout usa/adkulas/fix-covered-rubric-buttons
npm start
```
- Navigate to: [Primary Secondary Template (fullscreen width)](http://localhost:8001/templates/primary-secondary/demo/index.html)
- In the dev console Add a background color to the primary section
  - Example `<div slot="primary" style="background-color: #fefefe;">`
- Observe the `box-shadow` is still visible on the primary section:
image
- confirm `z-index` of main content is set back to auto
```css
			header {
				grid-area: header;
				z-index: 2;
			}
```
```css
			.d2l-template-primary-secondary-content {
				display: grid;
				grid-area: content;
				grid-template-areas: "primary divider secondary";
				grid-template-columns: minmax(320px, 2fr) 1px minmax(320px, 1fr);
				grid-template-rows: auto;
				overflow: hidden;
				z-index: auto;
			}
```
```css
			footer {
				background-color: #ffffff;
				box-shadow: 0 -2px 4px rgba(73, 76, 78, 0.2); /* ferrite */
				grid-area: footer;
				padding: 0.75rem 1rem;
				z-index: 2; /* ensures the footer box-shadow is over main areas with background colours set */
			}
```


adkulas:
Currently confirming the following:

- [ ] Fix works on Chrome (latest)
- [ ] Fix works on Firefox (latest)
- [ ] Fix works on Edge (latest)
- [ ] Check that d2l-simple-overlay and `d2l-template-primary-secondary`  works on other places that use it
- [ ] Recreate the scenario Dave was originally trying to fix and verfiy it is still fixed (#770)

dbatiste:
This seems reasonable.  I think it would be worthwhile adding a visual-diff test that mocks up this scenario where something in the context sets a `z-index` with the expectation of overlapping the header and footer, like fullscreen dialog would.

To be clear about `z-index` however, it is not inherited like properties such as `color`.  What actually happens is - if the element has a `position` other than `static` and a `z-index` that's an integer (not `auto`) then the browser will create a stacking context, effectively a hierarchy of z-index values.  So given the following:

```
A1 (z-index 2)
    B1 (z-index 1)   
A2 (z-index 1)
    B2 (z-index 20)  
        C1 (z-index 5)
    B3 (z-index 10)
    B4
        C2 (z-index 30)
```

Stacking will occur within a given stacking context according to the `z-index` values in that context.  So A1 will be on top of A2, B2 will be on top of B3, and C2 will be on top of B2.  B1 will be visually on top of all of A2's children because A1 has a larger `z-index` than A2.

PR #808, opened 2020-08-26
fix for legacy edge RTL horizontal scroll

spandanabaruah:
This overflow-y here (https://github.com/BrightspaceUI/core/blob/master/templates/primary-secondary/primary-secondary.js#L68) was causing an endless horizontal scroll in RTL languages in Legacy Edge (see in screenshot). This code change is to fix this problem which is outlined in this trello card https://trello.com/c/TgSSaZ1d/245-faceedge-primary-template-panel-has-large-horizontal-scroll-bar-in-legacy-edge-rtl
face_edge_scrollbar


templates/primary-secondary/primary-secondary.js
```diff
@@ -65,7 +65,7 @@ class TemplatePrimarySecondary extends LitElement {
 			}
 			main {
 				grid-area: primary;
-				overflow-y: auto;
+				overflow-x: hidden;
```
spandanabaruah, line 68: This change doesn't seem to be breaking anything in the other browsers (for LTR and RTL)

dbatiste: To be clear, is the scrolling "endless", or is it just really far... to the point where our offscreen text would render?

I see why we would add the `overflow-x`, but I'm not sure why we would remove the `overflow-y`.  What's the reasoning for that?

@wongvincent - looks like this `overflow-y` was here (class was changed to name somewhere in the history) since you added this component.  Do you recall why this was included, and do you have concerns about this change?

wongvincent: I do not recall. In the initial commit, which was really just testing stuff out, the `overflow-y` CSS property was added along with a `height` CSS property. But then the `height` property was removed and the `overflow-y` remained. The `overflow-y` property may not be required for anything now.

spandanabaruah: @dbatiste yup you're right it is not really an endless scroll - but a really long horizontal scroll. With this change, that horizontal scroll in legacy edge for RTL languages is removed. 

I tried keeping both `overflow-y: auto` as well as `overflow-x: hidden` and it seemed to behave exactly the same as having only then `overflow-x: hidden ` rule

PR #817, opened 2020-09-01
input-date/time/date-time: required

margaree:
- If required and no default value is given, then for date and date-time the default should be today or closest valid date if min/max
- Time should default to midnight, which it already does so no changes were needed.
- "Clear" button should not be displayed when required
- If user tries to erase the text, should revert back to previous good value


components/inputs/input-date-time.js
```diff
@@ -35,6 +35,10 @@ class InputDateTime extends FormElementMixin(LocalizeCoreElement(RtlMixin(LitEle
 			 * Minimum valid date/time that could be selected by a user.
 			 */
 			minValue: { attribute: 'min-value', reflect: true, type: String },
+			/**
+			 * Indicates that a value is required
+			 */
+			required: { type: Boolean, reflect: true },
```
dlockhart, line 41: Should probably set default to `false` in the constructor?

helpers/dateTime.js
```diff
@@ -55,6 +55,24 @@ export function formatDateInISOTime(val) {
...
+		return dateTime ? formatDateTimeInISO(convertLocalToUTCDateTime(today)) : formatDateInISO(today);
+	} else {
+		if (minValue && maxValue) {
+			const diffToMin = Math.abs(todayDate.getTime() - minDate.getTime());
+			const diffToMax = Math.abs(todayDate.getTime() - maxDate.getTime());
+			if (diffToMin < diffToMax) return minValue;
+			else return maxValue;
+		} else if (minValue) return minValue;
+		else if (maxValue) return maxValue;
+	}
```
dlockhart, line 73: I'm guessing this never returns `undefined` because implicitly if there's no min/max specified, `isDateInRange` will always return `true`. Not sure if there's a way to make that more clear looking at the code (I'm surprised the linter isn't complaining).

margaree: Yup exactly - Good idea, I'll try to make it more clear within this function.

PR #819, opened 2020-09-01
Use d2l-input-text internally and add form validation to d2l-input-number

krshnpatel-d2l:
Changes made in this PR:
~ used d2l-input-text internally instead of using the native input in d2l-input-number
~ added form validation for d2l-input-number (i.e. rangeOverflow, rangeUnderflow)
~ localized validation messages for d2l-input-number


components/inputs/input-number.js
```diff
@@ -81,32 +70,69 @@ class InputNumber extends LitElement {
...
+		const errors = await super.validate(showErrors);
+		if (errors.length !== 0) {
+			return errors;
+		}
+		return await this.shadowRoot.querySelector('d2l-input-text').validate(showErrors);
+	}
+
+	get validationMessage() {
+		if (this.validity.rangeOverflow || this.validity.rangeUnderflow) {
+			const label = this.label && !this.labelHidden ? this.label : 'Number';
```
dlockhart, line 107: What's the thinking here behind using "Number" if they've chosen to hide the label? I think we always just want to use the label they've given OR just embed "Number" inside the lang term like Allan had for the existing range under/overflow terms -- i.e. `"Number must be lower than {max}"`.

components/inputs/input-number.js
```diff
@@ -81,32 +70,69 @@ class InputNumber extends LitElement {
...
+
 	async _handleChange(e) {
 		const value = e.target.value;
 		this._formattedValue = value;
 		await this.updateComplete;
 		this.value = parseNumber(value);
+		this.setFormValue(this.value);
+		this.setValidity({
+			rangeUnderflow: (this.min || this.min === 0) && this.value < this.min,
+			rangeOverflow: (this.max || this.max === 0) && this.value > this.max
```
dlockhart, line 129: Could maybe do `typeof(this.min) === 'number' && this.value < this.min` to remove some ambiguity around casting `this.min` to a boolean resulting in `false` if it's `0`.

components/inputs/input-number.js
krshnpatel-d2l: Hmm in that case I think it'd make more sense to just use the given label because it'd be clearer to the user.

On the other hand, I wanted to use "Number" if there was no label because the user may not know what the validation error was referring to if a field had a hidden label.

lang/en.js
```diff
@@ -33,6 +33,9 @@ export default {
 	"components.input-date.errorOutsideRange": "Date must be between {minDate} and {maxDate}",
 	"components.input-date.openInstructions": "Use date format {format}. Arrow down or press enter to access mini-calendar.",
 	"components.input-date.setToToday": "Set to Today",
+	"components.input-number.errorMaxNumberOnly": "{label} must be lower than {maxNumber}.",
+	"components.input-number.errorMinNumberOnly": "{label} must be higher than {minNumber}.",
```
AllanKerr, line 37: It'd recommend reusing the lang terms from `components.form-element.input.number.rangeOverflow` and `components.form-element.input.number.rangeUnderflow`

components/inputs/input-number.js
AllanKerr: When @geurts and I discussed it he mentioned only wanting `Number` because we want to avoid using the field label unless its absolutely necessary. It might be worth double checking just in case something has changed since then.

components/inputs/input-number.js
```diff
@@ -81,32 +70,69 @@ class InputNumber extends LitElement {
...
+		this.shadowRoot.querySelector('d2l-input-text').focus();
+	}
+
+	async validate(showErrors) {
+		const errors = await super.validate(showErrors);
+		if (errors.length !== 0) {
+			return errors;
+		}
+		return await this.shadowRoot.querySelector('d2l-input-text').validate(showErrors);
+	}
```
AllanKerr, line 103: I lead you slightly astray here so apologies for that. It looks like the PR was slightly out of date so there are a couple changes that should be made here. [This is what the most up to date one looks like](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-date-range.js#L211).
1. We should call `this.shadowRoot.querySelector('d2l-input-text').validate(showErrors)` first since we want to ensure the `input-text` passes validation first before we perform number validation on it.
1. We always want to call `validate` on both since validation both shows new errors AND hides errors that are no longer relevant. But if validation fails for the child then we just want to suppress number validation errors with `ValidationType.SUPPRESS_ERRORS`
1. `showErrors` should be `validationType`.

I believe this will do what you need:
```
	async validate(validationType) {
		const childErrors = await this.shadowRoot.querySelector('d2l-input-text').validate(validationType);
		const errors = await super.validate(childErrors.length > 0 ? ValidationType.SUPPRESS_ERRORS : validationType);
		return [...childErrors, ...errors];
	}
```

components/inputs/input-number.js
dlockhart: Yeah I think I agree with Jeff here... "Number must be greater than 4" when shown as a tooltip next to any input seems pretty clear to me. When we transplant arbitrary labels into lang terms we run the risk of running into all kinds of weird issues in other languages.

lang/en.js
```diff
@@ -33,6 +33,7 @@ export default {
 	"components.input-date.errorOutsideRange": "Date must be between {minDate} and {maxDate}",
 	"components.input-date.openInstructions": "Use date format {format}. Arrow down or press enter to access mini-calendar.",
 	"components.input-date.setToToday": "Set to Today",
+	"components.input-number.errorOutsideRange": "Number must be between {minNumber} and {maxNumber}.",
```
dlockhart, line 36: This could probably just go up with `components.form-element.input.number` as `.rangeError`?

dlockhart (approved):
This seems good to me!

lang/en.js
```diff
@@ -10,8 +10,9 @@ export default {
 	"components.form-element.defaultError": "{label} is invalid.",
 	"components.form-element.defaultFieldLabel": "Field",
 	"components.form-element.input.email.typeMismatch": "Email is not valid",
-	"components.form-element.input.number.rangeOverflow": "Number must be lower than {max}",
-	"components.form-element.input.number.rangeUnderflow": "Number must be higher than {min}",
+	"components.form-element.input.number.rangeError": "Number must be between {min} and {max}.",
```
AllanKerr, line 13: I'm not sure if its a strict requirement but should these be added with an empty string in the other languages?

krshnpatel-d2l: Would it hurt to add them? If not, then I can add them in this PR.

AllanKerr: It definitely won't hurt, [it seems to be the approach we take most often](https://github.com/BrightspaceUI/core/commit/2fbdc454b0a6fd022947cbedfb5bdd5999ea2eb1#diff-d91f9f7171040f8d75fd69c48184da63). I'm not sure if its a strict requirement but probably worth doing just incase.

AllanKerr (approved):

PR #821, opened 2020-09-01
input-date/time components: setFormValue

margaree:
Added `setFormValue` to the input-date/time components and rearranged a little bit to be more modelled off of the FormElementMixin documentation.


components/inputs/input-date-range.js
```diff
@@ -145,6 +145,21 @@ class InputDateRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(LitEl
...
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldVal, prop) => {
+			if (prop === 'startValue' || prop === 'endValue') {
+				this.setFormValue({
+					[`${this.name}-startValue`]: this.startValue,
+					[`${this.name}-endValue`]: this.endValue,
```
dlockhart, line 155: I guess if they don't set a name, then they likely don't care about the form value anyway?

components/inputs/input-date-range.js
```diff
@@ -145,6 +145,21 @@ class InputDateRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(LitEl
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		changedProperties.forEach((oldVal, prop) => {
```
dlockhart, line 151: This seems like a better place for this given `_handleChange` wouldn't have been called if values were set via JS. 👍

PR #822, opened 2020-09-01
Update Semantic Release Versioning Info

AllanKerr:
We don't want people accidentally trigger major version bumps so clarifying that it is in the message body rather than the head like the others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Releases use the [semantic-release](https://semantic-release.gitbook.io/) toolin
 
 Upon release, the version in `package.json` is updated, a tag and GitHub release is created and a new package will be deployed to NPM.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,7 +127,7 @@ Releases use the [semantic-release](https://semantic-release.gitbook.io/) toolin
 
 Upon release, the version in `package.json` is updated, a tag and GitHub release is created and a new package will be deployed to NPM.
 
-Commits prefixed with `feat` will trigger a minor release, while `fix` or `perf` will trigger a patch release. A commit containing `BREAKING CHANGE` will cause a major release to occur.
+Commits prefixed with `feat` will trigger a minor release, while `fix` or `perf` will trigger a patch release. A commit containing `BREAKING CHANGE` in the _**message body**_ will cause a major release to occur.
```
dlockhart, line 130: I think it needs a colon after "CHANGE"

PR #823, opened 2020-09-02
input-date/time ranges: required

margaree:
This PR is to implement `required` in the range input components. 
The most complex part of the change is the logic around start and end default values for the components that contain dates (`input-time-range` already handled this).
- If `startValue` then `endValue` = `startValue` + 1 day
- If `endValue` then `startValue` = `endValue` - 1 day
- If neither start or end then the default is retrieved from the start component.
     - If the day after is in range then `endValue` = value of start input + 1 day
     - If the day after is out of range (after maxValue) then value of end input = initial value of start input and value of start input = initial value of start input - 1 day

Hopefully that makes sense and covers all cases.


components/inputs/input-date-range.js
```diff
@@ -106,6 +111,42 @@ class InputDateRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(LitEl
 		if (!this.label) {
 			console.warn('d2l-input-date-range component requires label text');
 		}
+
+		if (this.required && (!this.startValue || !this.endValue)) {
+			if (this.startValue) {
+				const endInputDate = new Date(this.startValue);
+				endInputDate.setDate(endInputDate.getDate() + 1);
+				const parsedObject = parseISODate(endInputDate.toISOString());
+				this.endValue = formatDateInISO(parsedObject);
```
dlockhart, line 120: I guess we're kind of assuming that if they set a start date, that both it and the day after are "in range"? Things get really complicated if the max & min only span 1 day. 😕

margaree: Yup exactly - it seemed unnecessarily complicated to consider/handle other cases especially since they're fairly unlikely.

PR #824, opened 2020-09-02
Add row number for screen reader

ciss1995:
Context:
DE39906 - https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F407153740380
Add information about the current row for screen readers.

Change:
Reorder items -> Reorder items, current position {currentPosition} out of {size}.
Screen Shot 2020-09-02 at 12 51 32 PM

Quality:
Tested locally with the demo page.


components/list/list-item-drag-handle.js
```diff
@@ -38,6 +38,8 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
 		return {
 			disabled: { type: Boolean, reflect: true },
 			text: { type: String },
+			size: { type: Number },
```
zinabat, line 41: It's a bit strange to have a `size` associated with the handle of each item

components/list/list-item-drag-drop-mixin.js
```diff
@@ -47,8 +47,10 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			draggable: { type: Boolean, reflect: true },
 			dragging: { type: Boolean, reflect: true },
 			dragHandleText: { type: String, attribute: 'drag-handle-text' },
+			dropDestination: { type: Number, reflect: true },
```
zinabat, line 50: multi-word properties should always be translated to kebab-case.
```suggestion
			dropDestination: { type: Number, reflect: true, attribute: 'drop-destination' },
```

components/list/list-item-drag-handle.js
```diff
@@ -38,6 +38,8 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
...
+			currentPosition: { type: Number },
```
zinabat, line 42: This should be translated to `current-position`

components/list/list-item-drag-drop-mixin.js
```diff
@@ -47,8 +47,10 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
 			dropText: { type: String, attribute: 'drop-text' },
 			key: { type: String, reflect: true },
+			size: { type: Number, reflect: true },
```
dbatiste, line 53: I'm not really a fan of adding the size to the item component since it results in two sources of truth.  This is really a property of the list, which ought to be obtained by querying the list item count when needed.

components/list/list-item-drag-handle.js
dbatiste: Agreed, but for other reasons mentioned in previous comment.

zinabat:
Just spitballing - we can probably ask the siblings/parent for a count and position on `focusin` if in keyboard mode. This should ensure it's kept up to date when needed, I think. I'm pretty sure screenreaders read out after this event but I'm not positive.

dbatiste:
Yeah.  Cheick and I were chatting a bunch in Slack.  I think it would be worthwhile adding a helper to the list component to get the item count.  I think it should looks something like:
```
this.shadowRoot.querySelector('slot')
    .assignedNodes({ flatten: true })
    .filter((node) => node.nodeType === Node.ELEMENT_NODE);
```
It would be good to figure out position just-in-time as well as you indicated.

components/list/demo/list-demo-drag-n-drop-usage.js
```diff
@@ -70,6 +71,7 @@ class ListDemoDragNDropUsage extends LitElement {
 				${repeat(this.list, (item) => item.key, (item) => html`
 					<d2l-list-item
 						key="${ifDefined(item.key)}"
+						drop-destination="${ifDefined(this._dropLocation || item.key)}"
```
zinabat, line 74: Passing this information in the demo doesn't appear to work properly when I test with VoiceOver on Chrome.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -365,17 +366,25 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
 			'd2l-hovering': this._hovering
 		};
+
 		return this.draggable ? templateMethod(html`
 			<d2l-list-item-drag-handle
 				id="${this._itemDragId}"
 				class="${classMap(classes)}"
 				text="${ifDefined(this.dragHandleText)}"
+				current-position="${ifDefined(this.dropDestination)}"
+				list-size="${ifDefined(this._getListSize())}"
```
zinabat, line 387: I think based on @dbatiste and our discussions, we should move instead to get the position and list size when the drag handle is focused. This will avoid having to explicitly pass the information down the tree in this way.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -606,6 +615,11 @@ export class NewPositionEventDetails {
 			}
 		}
 		list[destination] = item;
+
+		return {
+			origin: origin,
+			destination: destination
+		}
```
zinabat, line 622: We won't need this anymore if we are asking the parent for information on `focusin`.

components/list/list-item-drag-handle.js
```diff
@@ -37,6 +37,8 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
 	static get properties() {
 		return {
 			disabled: { type: Boolean, reflect: true },
+			currentPosition: { type: Number, attribute: 'current-position' },
+			listSize: { type: Number, attribute: 'list-size' },
```
zinabat, line 41: Let's remove these properties and query for the information only when needed.

components/list/list.js
```diff
@@ -64,6 +64,12 @@ class List extends LitElement {
 		`;
 	}
 
+	getListSize() {
```
zinabat, line 67: Add this method to the readme so people know about it 🍰 A test or two for it wouldn't be a terrible idea

dbatiste: I also think this should be called `getListItemCount`.

components/list/list-item-drag-handle.js
```diff
@@ -240,6 +242,7 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
 				@focusout="${this._onFocusOut}"
 				@keyup="${this._onActiveKeyboard}"
 				@keydown="${this._onPreventDefault}"
+				aria-live="assertive"
```
zinabat, line 245: This line does the magic. It tells the screen reader to read out the label if it changes, and IMMEDIATELY. 🎉

components/list/list-item-drag-handle.js
```diff
@@ -150,6 +151,7 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
 			case keyCodes.UP:
 				this._movingElement = true;
 				action = dragActions.up;
+				this.updateComplete.then(() => this.blur());
```
zinabat: This line is also necessary to tell the screenreader to refocus the item.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -37,7 +38,7 @@ const isDragSupported = () => {
 	return typeof el.ondragenter === 'function';
 };
 
-export const ListItemDragDropMixin = superclass => class extends superclass {
+export const ListItemDragDropMixin = superclass => class extends LocalizeCoreElement(superclass) {
```
zinabat, line 41: This line breaks `list-item-generic-layout` tests >_> WHY.

zinabat: This is definitely a tomorrow problem, but I _believe_ what's happening in the tests is that the `shadowRoot` of the `list-item`s are not loaded until the locale resources load, which results in a bunch of `null` shadow roots when we try to query the individual list items.

We need a way to wait for the locale resources before attempting to query the shadow root.
```js
it('enables grid on list-item-generic-layout elements', async() => {
	const el = await fixture(normalFixture); // this waits for the list but not necessarily the list-items in it
	await elementUpdated(el._getItems()[0]); // this works sometimes, but not all the time
	el._getItems().forEach(item => {
		expect(item.shadowRoot.querySelector('d2l-list-item-generic-layout').gridActive).to.be.true; // these shadowRoots are null if the items haven't loaded completely
	});
});
```

ciss1995: We could also keep the LocalizeCoreElement in the `list-item-drag-handle` and we pass down an object with the position and the count instead of the string?

components/list/list.js
```diff
@@ -64,6 +64,14 @@ class List extends LitElement {
 		`;
 	}
 
+	getListItemCount() {
+		return this._getItems().length;
+	}
+
+	getListItemPosition(item) {
```
dbatiste, line 71: Does this return 0 for the first item?  Just wondering if this is index or position, and what's going to get announced.

dbatiste: Oh I see above that +1 is added to it.  I think I would call this `getListItemIndex`, but not a huge deal.  Usually I think of "position" starting at 1, but maybe that's just me.

ciss1995: Oh yeah this does return the index that's why we increment the value in `parent.getListItemPosition(this) + 1`

dbatiste (approved):
Just the comment about position vs. index to consider.

Would be good to have @zinabat approval on this too.  :)

components/list/list.js
ciss1995: Oh yeah good point! I can fix up the language there.

zinabat: Index right now, which means we should either change the name to `Index` or return the position.

components/list/README.md
```diff
@@ -27,6 +27,8 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
 
 **Methods:**
 
+- `getListItemCount`: gets the length of the items within the list
+- `getListItemPosition` (Object): gets the position of the given element within the list
```
zinabat, line 31: Readme update because you changed the name!
```suggestion
- `getListItemIndex` (Object): gets the index of the given element within the list
```

zinabat (approved):
Just the readme update :)

PR #826, opened 2020-09-02
input-date*: remove default date value

margaree:


components/inputs/input-date.js
```diff
@@ -190,6 +186,7 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 					@keydown="${this._handleKeydown}"
 					hide-invalid-icon
 					id="${this._inputId}"
+					@invalid-change="${this._handleInvalid}"
```
margaree, line 189: @AllanKerr is this the right way to do this? I want to listen for when `input-text` goes to `invalid` because of required not being met

dlockhart (approved):

components/inputs/input-date.js
AllanKerr: Ideally it should work the other way around where you override `validate`, call `validate()` on the `text-input` and disable the `text-input`'s self-validation with `@d2l-form-element-should-validate`

margaree: Right - like with the `input-date-range` component. Makes sense, thanks!

margaree: Ok so I've updated that in a way that works where I'm also listening for blur since I don't want an invalid state to happen until after the first blur. I don't know if it's completely what you meant so let me know if it's completely off.

AllanKerr: That looks right to me. The only part that it does make me wonder about is if `changed` should always use `ValidationType.UPDATE_EXISTING_ERRORS`. Displaying some new errors immediately when the value changes and others only on blur feels inconsistent.

margaree: So if it's only `ValidationType.UPDATE_EXISTING_ERRORS` then it doesn't work for something like min/max since there aren't necessarily already errors to update, whereas if it's only `ValidationType.SHOW_NEW_ERRORS` then when `required` the `input-date` starts immediately in an invalid state. I can do some experimenting to see if there's a better way to have this more consistent. I could track if there's been an initial blur and the always just use `SHOW_NEW_ERRORS` or something like that.

AllanKerr: But the idea there is that as long as the user hasn't left focus then they aren't done editing so even `min` and `max` errors shouldn't be showing.

margaree: Yup but I mean once they have happened. It looks like updated runs before blur so the errors don't exist yet to update.

AllanKerr: I'll see if I can take look

PR #827, opened 2020-09-03
using switches for demo snippet controls

dlockhart:
While working on integrating the skeleton stuff, I thought it'd be cool for the demo snippet to support a control to enable/disable the skeleton on the component(s) inside. While implementing that, using the new `<d2l-switch>` seemed to make a lot of sense! That involved moving things to more of a side panel, which I think lines up with our future plans to add responsive resize handles that you could drag to shrink/grow the size of the demo area.

I also cleaned up the background/margins to avoid having the demo overlap with the controls (which happened occasionally) and tidied up the radius/border on the code view to make it fit a bit nicer.

Before:
Screen Shot 2020-09-03 at 5 17 21 PM

After (with future skeleton switch, not included in this PR):

Screen Shot 2020-09-04 at 9 38 58 AM


components/demo/code-view-styles.js
```diff
@@ -33,7 +34,6 @@ export const styles = css`
 		display: none;
 	}
 	pre[class*="language-"] {
-		border-radius: 5px;
```
dlockhart, line 36: Having the parent `overflow: hidden` "clips" children when there's a border-radius, which is what this child border-radius was trying to emulate.

components/demo/demo-snippet.js
```diff
@@ -87,29 +118,9 @@ class DemoSnippet extends LitElement {
 			.replace(/=""/g, '');           // replace empty strings for boolean attributes (="")
 	}
 
-	_handleDirChange() {
```
dlockhart, line 90: Moved this into a more generic helper as I'm going to need the same code to apply `skeleton` to the stuff in the slot.

dlockhart:
I updated things to use a side panel instead of dedicated space at the top, which is more in line with how we'll want to implement responsive resizing in the future.

components/demo/demo-snippet.js
```diff
@@ -30,12 +66,14 @@ class DemoSnippet extends LitElement {
...
 		return html`
-			<div class="d2l-demo-snippet-demo" dir="${dirAttr}">
-				<div class="d2l-demo-snippet-actions">
-					<button id="d2l-demo-snippet-toggle-dir" @click="${this._handleDirChange}" title="toggle dir">${this._dirButton}</button>
+			<div class="d2l-demo-snippet-demo-wrapper">
+				<div class="d2l-demo-snippet-demo" dir="${dirAttr}">
+					<slot name="_demo"></slot>
+					<slot></slot>
+				</div>
+				<div class="d2l-demo-snippet-settings">
```
dbatiste, line 74: Not that it need to be handled in this PR, but another use-case that comes up in some of our demo pages are ad hoc actions.  Many times we end up adding extra buttons outside the snippet for testing purposes.  ex.

* a "focus" button that applies focus to the component (html editor)
* buttons for adding/removing content (floating buttons, tabs demos)
* buttons for manipulating hierarchical views

Many times these buttons, and their associated JS could be placed in the snippet, but the choice was made not to in order to keep the snippet code focused.

So, I am wonder if we would find it useful to have a slot to help group these things.  Focus is one that ought to be more common than it actually is, and perhaps we should build it in.  Sometimes though, the snippet contains several elements, like a menu + menu-items, and the first element isn't necessarily the one we're interested in applying settings to.

dbatiste (approved):
Looks good.  Just that one comment to consider.

components/demo/demo-snippet.js
dlockhart: Interesting, yeah this could easily have a slot and/or come with some built-in actions. I'll think about this more and it'll come up again when we look to improve this further for the component catalog.

PR #831, opened 2020-09-04
feat: US119264 Oslo: Use non-versioned cache key

pureooze:
This PR works with BrightspaceHypermediaComponents/activities#1071 and a change in the LMS (https://github.com/Brightspace/lms/pull/917) where we are changing how OSLO requests overrides. Instead of requesting based on the resource path we want to use the collection name. This helps us avoid the issue where the Cache Storage was being filled with caches from older versions of BSI because the resource path includes the BSI version.


helpers/getLocalizeResources.js
```diff
@@ -284,26 +284,23 @@ function shouldFetchOverrides() {
...
 
 	if (shouldUseBatchFetch()) {
 
 		// If batching is available, pool requests together.
 
-		url = formatFunc(OverrideLanguage);
-		url = new URL(url).pathname;
-
-		res = fetchWithPooling(url);
+		resource = formatFunc(OverrideLanguage);
```
JordanCorner, line 293: The formatFunc doesn't take any properties, should we remove this? https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/mixins/d2l-activity-editor-lang-mixin.js#L9

helpers/getLocalizeResources.js
```diff
@@ -284,26 +284,23 @@ function shouldFetchOverrides() {
...
 
 	} else /* shouldUseCollectionFetch() == true */ {
 
 		// Otherwise, fetch it directly and let the LMS manage the cache.
 
-		url = formatFunc(OverrideLanguage);
-		url = new URL(url).pathname;
-		url = documentLocaleSettings.oslo.collection + url;
+		resource = formatFunc(OverrideLanguage);
+		resource = formatOsloRequest(documentLocaleSettings.oslo.collection, resource);
```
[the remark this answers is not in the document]
pureooze: Yeah maybe, something like `requestURL` 🤔

helpers/getLocalizeResources.js
pureooze: Good catch, yeah I think this is good to remove. I'm actually not sure why it ever took args, maybe the plan was at some point to request per lang instead of one override file. Either way, I think this arg is good to remove.

JordanCorner (approved):

pureooze:
This PR is ready for review 😃

dlockhart (approved):

PR #833, opened 2020-09-08
Select / Textarea Error Icon Focus Tweaks

AllanKerr:
# Changes
- Since `textarea` and `select` add additional padding for the error icons we've decided they don't need to be hidden on focus. Instead, they should always show as long as the input is in an error state.
- `input-text` needs to be handled independently because its error icon is an actual overlay so it doesn't require additional padding.


dbatiste (approved):
Changes look ok.  The icon in the select does look pretty squished in there.  May want to double check with @geurts.

PR #834, opened 2020-09-09
Fix Tooltip webkit clipping defect

AllanKerr:
When transform rotate is combined with clip the div will occasionally clip incorrectly causing part of the pointer to be cut off. Adding a transparent border forces webkit to change how it renders the pointer which prevents this from happening.

Capture


dbatiste (approved):
Crazy!  Is this an issue for the d2l-dropdown content pointer as well?

AllanKerr:
> Crazy! Is this an issue for the d2l-dropdown content pointer as well?

I haven't been able to reproduce it with `d2l-dropdown` but I can't say for sure whether it does or doesn't impact it. I've noticed that tooltip is only impacted by this in certain places and even in those places it'll often fix itself as you use the page without anything actually changing.

dlockhart (approved):

PR #835, opened 2020-09-09
FormElementMixin Nesting Refactor

AllanKerr:
## Motivation
- The previous method of disabling validation for child form elements then overriding validate wasn't working.
- The two big issues were that:
    1. It was difficult to understand
    1. Form elements know when to call `validate` for themselves. Other components may call validate at different times in their lifecycle or in response to different events. A parent component doesn't have enough information to know when to call `validate` for its children.

## Changes
- Components are now solely responsible for validating themselves
- If a component is invalid then it will show display itself invalid and show a tooltip
    - If a parent component wants to hide its tooltip when a child component has an error it may use the `childErrors` property to conditionally render the `tooltip`
    - If a parent component wants to force a child component to look like its in an error state it may use `force-invalid`
- Components are responsible for passing `noValidate` up to child form elements
- Components are responsible for override `validate` and calling `validate` on any child form elements.


components/inputs/input-date-time-range.js
```diff
@@ -109,17 +109,20 @@ class InputDateTimeRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(L
...
 		return html`
 			${tooltipStart}
 			${tooltipEnd}
 			<d2l-input-fieldset label="${ifDefined(this.label)}" ?label-hidden="${this.labelHidden}" ?required="${this.required}">
 				<div class="d2l-input-date-time-range-start-container">
 					<d2l-input-date-time
+						?novalidate="${this.noValidate}"
 						@change="${this._handleChange}"
 						class="d2l-input-date-time-range-start"
-						@d2l-form-element-should-validate="${this._handleNestedFormElementValidation}"
```
dlockhart, line 122: Can `_handleNestedFormElementValidation` be removed?

components/form/form-element-mixin.js
```diff
@@ -92,12 +87,15 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 			name: { type: String },
 			noValidate: { type: Boolean, attribute: 'novalidate' },
 			validationError: { type: String, attribute: false },
+			childErrors: { type: Object, attribute: false },
+			_errors: { type: Array, attribute: false }
```
dlockhart, line 91: Just wanted to confirm that a change to either of these needs to force a re-render of the component into which `FormElementMixin` is mixed in? `_errors` seems like a definite yes, but doesn't `childErrors` typically just impact parents, which would be communicated via the `d2l-form-element-errors-change` event? If so, it could just be a private variable vs. a property.

dlockhart (commented):
This looks good to me, and seems clearer in terms of who's responsible for what.

components/form/form-element-mixin.js
AllanKerr: `_errors` is actually the one we could get away with without it being a property since `validationError` is a property that references the first error in `_errors` so a change to `_errors` is already guaranteed to cause a rerender.

`childErrors` maintains a mapping of errors that exist in the component's nested from elements. If `childErrors` changes we need to rerender because the parent may want to hide or show its tooltip in response to the child hiding or showing its tooltip.

components/inputs/input-date.js
```diff
@@ -248,16 +254,10 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 		if (this._textInput) this._textInput.focus();
 	}
 
-	async validate(validationType) {
-		let childErrors = [];
-		const inputTextElem = this.shadowRoot.querySelector('d2l-input-text');
-		if (inputTextElem) {
-			await inputTextElem.updateComplete;
-			childErrors = await Promise.resolve(inputTextElem.validate(validationType));
-			this._inputTextInvalid = inputTextElem.invalid;
```
margaree, line 257: So I think this still needs to be set, otherwise the icon doesn't change if `required` and no entry

components/inputs/input-time-range.js
```diff
@@ -160,17 +160,21 @@ class InputTimeRange extends FormElementMixin(RtlMixin(LocalizeCoreElement(LitEl
 		const startLabel = this.startLabel ? this.startLabel : this.localize('components.input-time-range.startTime');
 		const endLabel = this.endLabel ? this.endLabel : this.localize('components.input-time-range.endTime');
 
+		const startTimeInput = this.shadowRoot.querySelector('.d2l-input-time-range-start');
```
margaree, line 163: Start and end wont really have their own errors - they can never be empty so they wont have an error state if empty and required. The only error possibility for times is if start time is after end time in the parent (though maybe this change is necessary still)

margaree (commented):
Looks good! Seems much more clear.

components/inputs/input-date.js
AllanKerr: Good catch, I'll add `this.invalid || this.childErrors.size > 0` so the error icon shows if the `text-input` is invalid.

components/inputs/input-time-range.js
AllanKerr: I don't think this is technically needed. I wrote it this way so that if `input-time` gets modified to support some sort of error, `input-time-range` will automatically handle it correctly.

margaree (approved):

dlockhart (approved):

PR #838, opened 2020-09-10
Support time-default-value override attribute in input-date-time

emil-furniss:
Supports overriding the default midnight time when a date is first set in the date-time input. This is done by adding an attribute `time-default-value` which takes a time string in the `hh:mm:ss` format and passes it to the `default-value` attribute on the time input.

Use case is in assignments, where the default for due dates and availability dates should default to 11:59 PM.


components/inputs/docs/input-date-time.md
```diff
@@ -194,15 +194,14 @@ The `<d2l-input-date-time>` component consists of a `<d2l-input-date>` and a `<d
 
 **Properties:**
 
-Note: All `*value` properties should be in ISO 8601 combined date and time format (`YYYY-MM-DDTHH:mm:ss.sssZ`) and in UTC time (i.e., do NOT localize to the user's timezone).
```
dlockhart, line 197: Instead of removing this, can you change it to say something like "... `max-value`, `min-value` and `value` should be in ISO 8601..."

components/inputs/input-date-time.js
```diff
@@ -43,6 +43,10 @@ class InputDateTime extends FormElementMixin(LocalizeCoreElement(RtlMixin(LitEle
 			 * Indicates that a value is required
 			 */
 			required: { type: Boolean, reflect: true },
+			/**
+			 * Override for the time input default value. If not provided, this will default to midnight. Should be in format "hh:mm:ss".
+			 */
```
dlockhart, line 48: Let's borrow the description from `d2l-input-time`:

```suggestion
			/**
			 * Set default value of time portion of the input. Valid values are times in ISO 8601 time format ("hh:mm:ss"), "startOfDay", "endOfDay".
			 * @type {'startOfDay'|'endOfDay'|string}
			 */
```

dlockhart (approved):
Looks good! Just a couple suggestions about the docs.

dlockhart (approved):

components/inputs/docs/input-date-time.md
```diff
@@ -194,14 +194,15 @@ The `<d2l-input-date-time>` component consists of a `<d2l-input-date>` and a `<d
...
+Note: `max-value`, `min-value` and `value` should be in ISO 8601 combined date and time format (`YYYY-MM-DDTHH:mm:ss.sssZ`) and in UTC time (i.e., do NOT localize to the user's timezone).
 
 | Property | Type | Description |
 |--|--|--|
 | `label` | String, **required** | Accessible label for the input |
 | `disabled` | Boolean | Disables the input |
 | `max-value` | String | Maximum valid date/time that could be selected by a user |
 | `min-value` | String | Minimum valid date/time that could be selected by a user |
+| `time-default-value`| String, default midnight | Override for default time in format `hh:mm:ss`. |
```
AllanKerr, line 205: It'd be worth adding something here about `startOfDay` and `endOfDay`. Its described in `input-time` as:
> Set default value of input. Valid values are times in ISO 8601 time format ("hh:mm:ss"), "startOfDay", "endOfDay".

PR #840, opened 2020-09-11
fixing position of floating buttons inside the demo snippet

dlockhart:
Fixes the form demo page where floating buttons existed inside the demo snippet.


dbatiste (approved):

krshnpatel-d2l (approved):
Looks good on my local env

PR #841, opened 2020-09-11
Add advanced FormElementMixin docs

AllanKerr:
# [README](https://github.com/BrightspaceUI/core/blob/akerr/form-element-mixin/native-docs/components/form/docs/form-element-wrapping.md)
## Changes
These docs detail how to create a custom form element that wraps native form elements


components/form/docs/form-element-wrapping.md
```diff
@@ -0,0 +1,110 @@
...
+		this.value = e.target.value;
+	}
+
+}
+customElements.define('my-wrapping-form-element', MyWrappingFormElement);
+```
+
+**2. Use the input's build-in client-side form validation:**
+
+We must add attributes to the native input so that it will validate its own input. In this case we will use modify `render` to add `required` and `pattern`.
```
dlockhart, line 63:
```suggestion
We must add attributes to the native input so that it will validate its own input. In this case we will modify `render` to add `required` and `pattern`.
```

dlockhart (approved):
Very clear!

PR #842, opened 2020-09-14
Visual diff tests and minor bug fixes

krshnpatel-d2l:
Changes made in this PR:
~ Added visual diff tests
~ Added a couple unit tests (invalid values and re-formatting values)
~ Validate on blur event (for the case when user focuses on a required field then blurs)
~ Fix tooltip position (make it relative host element)
~ Fix validity null error (make sure inner element exists before trying to validate)

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -39,6 +39,7 @@ class InputNumber extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 			css`
 				:host {
 					display: inline-block;
+					position: relative;
 					width: 4rem;
 				}
 			`
@@ -76,6 +77,7 @@ class InputNumber extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 	}
 
 	render() {
+		const tooltip = this.validationError ? html`<d2l-tooltip for="${this._inputId}" state="error" align="start">${this.validationError}</d2l-tooltip>` : null;
 		return html`
 			<d2l-input-text
 				autocomplete="${ifDefined(this.autocomplete)}"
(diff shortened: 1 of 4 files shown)


components/inputs/input-number.js
```diff
@@ -69,7 +70,14 @@ class InputNumber extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 		this.requestUpdate('value', oldValue);
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this.addEventListener('blur', this._handleBlur);
```
dlockhart, line 76: Could this be wired up via `<d2l-input-text @blur="{this._handleBlur}">`? That way Lit will automatically remove the listener for you. I see the date input is doing it the same way you are, so maybe it needs to be this way?

components/inputs/test/input-number.test.js
```diff
@@ -82,6 +82,43 @@ describe('d2l-input-number', () => {
...
+
+			setTimeout(() => dispatchEvent(elem, 'change'));
+			await oneEvent(elem, 'change');
+
+			expect(elem.value).to.equal(undefined);
+			expect(elem._formattedValue).to.equal('');
+		});
+	});
+
+	describe('re-formatting values', () => {
```
dlockhart, line 98: These 2 tests are kind of just testing that the number formatting/parsing in the `d2l-intl` library are working correctly. I'm OK verifying that, but could we just call the tests something like `it should format value using Intl` and `it should parse value using Intl`?

dlockhart (approved):

PR #843, opened 2020-09-14
Documentation and dispatching change event

krshnpatel-d2l:
Changes made in this PR:
~ Added documentation for `input-number` component and links to it
~ Minor typo fix in `input-text.md`
~ Dispatching change event when value is changed
~ Added unit tests for dispatching change event


components/inputs/README.md
```diff
@@ -94,7 +95,6 @@ Alternately, the `<d2l-input-fieldset>` component can accomplish this for you. T
 ## Future Enhancements
 
 - Date & time range with validation
-- Numeric input
```
dlockhart, line 97: Good catch!

components/inputs/docs/input-number.md
```diff
@@ -0,0 +1,52 @@
+# Number Inputs
+
+The `<d2l-input-number>` element is a wrapper around the `<d2l-input-text>` component. It's intended for inputting numbers only.
```
dlockhart, line 3: We may choose one day to not simply wrap `d2l-input-text`, so I don't think we need to leak this implementation details. You could say something like:
> The `<d2l-input-number>` element is similar to `<d2l-input-text>`, except it's intended for inputting numbers only.

components/inputs/docs/input-number.md
```diff
@@ -0,0 +1,52 @@
...
+  label="Label"
+  value="0">
+</d2l-input-number>
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `label` | String, required | Label for the input. |
```
dlockhart, line 21: These seem to be alphabetical except for `label`

components/inputs/docs/input-text.md
```diff
@@ -24,7 +24,7 @@ The `<d2l-input-text>` element is a simple wrapper around the native `<input typ
 | `autocomplete` | String | Specifies which types of values [can be autofilled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) by the browser |
 | `autofocus` | Boolean | When set, will automatically place focus on the input |
 | `disabled` | Boolean | Disables the input |
-| `label-hidden` | Boolean | Hides the labe visually (moves it to the input's `aria-label` attribute) |
+| `label-hidden` | Boolean | Hides the label visually (moves it to the input's `aria-label` attribute) |
```
dlockhart, line 27: 👍

components/inputs/input-number.js
```diff
@@ -147,6 +147,11 @@ class InputNumber extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 		this._formattedValue = value;
 		await this.updateComplete;
 		this.value = parseNumber(value);
+
+		this.dispatchEvent(new CustomEvent(
+			'change',
+			{ bubbles: true, composed: false }
+		));
```
dlockhart, line 154: Could you add some tests for this as well? We want to make sure that if the underlying value didn't actually change (even though the value the user typed in might have), that we don't fire the event.

components/inputs/docs/input-number.md
krshnpatel-d2l: This is because it's a `required` attribute. I see that `label` is always first for other docs as well.

dlockhart: Ah ok cool 👍

dlockhart (approved):
Looks good!

PR #845, opened 2020-09-15
input-date: accessibility changes

margaree:
Opening as draft in case there are bigger changes to how we want to do this.

To do:
- [x] add `offscreen-description` to `input-text` readme

Problem:
Support for having both `aria-describedby` and `title` and reading both was mixed, so there wasn't a good way to have the `empty-text` read.

Solution:
[see user story for more info] Use `d2l-tooltip` for what is currently in `title` and put the `empty-text` as an `aria-describedby` on the `input` in `input-text`.

Notes:
- The tooltip has a 1 second delay on hover
- If user opens then closes the dropdown, the tooltip should not reappear until they remove focus and return


components/inputs/input-date.js
```diff
@@ -169,14 +171,16 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 		const icon = (this.invalid || this._inputTextInvalid)
 			? html`<d2l-icon icon="tier1:alert" slot="left" style="${styleMap({ color: 'var(--d2l-color-cinnabar)' })}"></d2l-icon>`
 			: html`<d2l-icon icon="tier1:calendar" slot="left"></d2l-icon>`;
-		const tooltip = (this.validationError && !this._dropdownOpened) ? html`<d2l-tooltip align="start" announced for="${this._inputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
+		const errrorTooltip = (this.validationError && !this._dropdownOpened) ? html`<d2l-tooltip align="start" announced for="${this._inputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
+		const infoTooltip = (this._showInfoTooltip && !this.validationError && !this._inputTextInvalid && !this._dropdownOpened) ? html`<d2l-tooltip align="start" announced delay="1000" for="${this._inputId}">${this.localize(`${this._namespace}.openInstructions`, { format: shortDateFormat })}</d2l-tooltip>` : null;
```
dlockhart, line 175: Could be slightly simplified: `(this._showInfoTooltip && !errorTooltip && !this._inputTextInvalid)`

components/inputs/input-date.js
```diff
@@ -193,10 +197,10 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 					?label-hidden="${this.labelHidden}"
 					live="assertive"
 					@mouseup="${this._handleMouseup}"
+					offscreen-description="${ifDefined(this.emptyText ? this.emptyText : undefined)}"
```
dlockhart, line 200: Is this just handling the case where `emptyText` is `''`? I would have thought `ifDefined(this.emptyText)` would have sufficed.

components/inputs/input-text.js
```diff
@@ -77,6 +78,10 @@ class InputText extends FormElementMixin(RtlMixin(LitElement)) {
 			 * Imposes a lower character limit
 			 */
 			minlength: { type: Number },
+			/**
+			 * Additional information communicated in the aria-describedby on the input
+			 */
+			offscreenDescription: { type: String, reflect: true, attribute: 'offscreen-description' },
```
dlockhart, line 84: What do you think about just calling this `description`? We use that in buttons as "for accessibility when text on button does not provide enough context" which is kind of what this is. How input chooses to render that (offscreen linked with `aria-describedby` or otherwise) seems like an internal implementation detail.

components/inputs/input-date.js
margaree: It is to handle that case. I would have thought so too but sadly not.

components/inputs/input-text.js
margaree: Yup that makes sense to me.

AllanKerr:
I'm a little concerned that this having tooltips in a bunch of different levels within the nested form element tree is going to become impossible to manage safely. An example of this is that in `d2l-input-range` the info tooltip will open overtop of the validation tooltip because the info tooltip only gets suppressed by errors in validating itself or its children, not parents. Thoughts?

margaree:
Ah right, so it should probably be checking for `this.invalid` instead of the validation error but I do see your point about it being potentially unsustainable. Should `input-text` be managing its own tooltips instead?

AllanKerr:
> Ah right, so it should probably be checking for `this.invalid` instead of the validation error but I do see your point about it being potentially unsustainable. Should `input-text` be managing its own tooltips instead?

I _think_ `this.invalid` will fix this case but if we were to ever add info tooltips to other date input components suddenly those would start overlapping as well. I don't know if there really is a good way to handle it since we're trying to build independent components that need to peer into each others internals to prevent tooltip overlap.

dlockhart (approved):

PR #846, opened 2020-09-15
feat: Empty PR to generate semantic release

ciss1995:


AllanKerr (approved):
In the future I'd recommend using `git commit --allow-empty` which lets you create an empty commit without changing any files.

PR #847, opened 2020-09-15
Refactor input-number to use the new FormElementMixin nesting

AllanKerr:
# Changes
- `FormElementMixin` was refactored to not need `ValiditionType` and `d2l-form-element-should-validate`. This is the corresponding `input-number` change.


krshnpatel-d2l (approved):
Looks nicer and cleaner 👍

PR #849, opened 2020-09-16
fix: acitonHref !== actionHref

m6jones:
There was a typo when I was doing actionHref.


dbatiste (approved):

dlockhart (approved):

components/list/list-item-mixin.js
```diff
@@ -321,7 +321,7 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 					` : nothing }
 					${ this.actionHref ? html`
 					<a slot="content-action"
-						href="${this.acitonHref}"
+						href="${this.actionHref}"
```
AKobets, line 324: verified on demo page that this fixes navigation

PR #851, opened 2020-09-16
Add more information for list item drag handle keyboard mode

ciss1995:
Context:
Add more information to screen readers on drag and drop handle.

Quality: 
Screen Shot 2020-09-16 at 2 27 08 PM

### DIFF
--- a/lang/en.js
+++ b/lang/en.js
@@ -39,7 +39,7 @@ export default {
 	"components.input-time-range.errorBadInput": "{startLabel} must be before {endLabel}",
 	"components.input-time-range.startTime": "Start Time",
 	"components.list-item-drag-handle.default": "Reorder item action for {name}",
-	"components.list-item-drag-handle.keyboard": "Reorder items, current position {currentPosition} out of {size}",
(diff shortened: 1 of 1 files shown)


AKobets (approved):

dbatiste (approved):

lang/en.js
```diff
@@ -39,7 +39,7 @@ export default {
 	"components.input-time-range.errorBadInput": "{startLabel} must be before {endLabel}",
 	"components.input-time-range.startTime": "Start Time",
 	"components.list-item-drag-handle.default": "Reorder item action for {name}",
-	"components.list-item-drag-handle.keyboard": "Reorder items, current position {currentPosition} out of {size}",
+	"components.list-item-drag-handle.keyboard": "Reorder item, current position {currentPosition} out of {size}. To move this activity, press up or down arrows.",
```
dbatiste, line 42: Oh, one sec... Are you sure you want "activity" baked into this message?  Seems like this should be "item".

ciss1995: Yeah good point, will double check with Abbie

PR #853, opened 2020-09-16
Skeleton Mixin

dlockhart:
### Overview

This builds on the great work of @mdgbayly's #796, @ChrisKraljevicD2L's #502 and @m6jones's work with SVG skeletons in activities.

The main concept here is that elements which are skeleton-aware will extend the `SkeletonMixin`, which automatically gives them a `skeleton` attribute which can be set and brings in the skeleton styles that they can apply to their components. I've implemented this in `d2l-input-text` as an example. If we like this approach, we can apply it to all our inputs and any other place it makes sense.

The main skeleton style is the `d2l-skeletize` CSS class. When present on an element, and when the host element has its `skeleton` attribute set, a pulsing rounded bar will be applied that fills the entire element, like this (unfortunately GIFs don't have enough colours to do this justice):

Sep-16-2020 17-37-36

### Motivation & Principles

1. Several existing and in-progress projects need a skeleton loader implementation. We have design buy-in, multiple use cases, so now is a good time to incorporate something into core.
2. We'd prefer making our components "skeleton aware" so that they can optimize themselves, instead of having outside parties (i.e. the app) apply their interpretation of what a "skeleton date picker" looks like -- which will lead to inconsistencies.
3. We'd prefer apps be able to have a single render path that simply toggles skeletons on or off, vs. needing to have a "skeleton render path" and a "not skeleton render path". Note: this doesn't mean we won't eventually eventually create actual skeleton components (e.g. `<d2l-skeleton-paragraph>`, `<d2l-skeleton-chart>`) that apps can use.

### Specific Problems This PR Solves

Beyond Martin's initial work, the main problems that still remained were:

**Cap Height**

For typography, we want the skeleton to only fill the cap height of the text -- slightly less tall than the full element height. I solved this by overriding the `top` and `bottom` for each of our typography styles (3 body styles, 4 headings, 3 link variations & input labels) with a custom value. This was pretty finicky and involved lots of hand tweaking -- I used some opacity & lots of zooming. Here's heading 2 in "debug mode" as an example:

Screen Shot 2020-09-16 at 5 43 20 PM

**Multi-Line Headings**

Even if a heading contains multiple lines of text, design would like the skeleton to only be a single line. I solved this by setting the `height` of the skeleton equal to the `line-height` of the individual headings.

heading

**Block Element Width**

For block elements (like headings), we don't always want the skeleton width to be 100%. Even for other types of elements, sometimes you just don't want the skeleton to fill the entire space, and setting a width on the element when it's not-a-skeleton isn't feasible.

In these cases, there are a series of CSS classes that can be applied to get different percentage widths from `5%` to `95%` in steps of `5`. So you could do something like:

```html
<h2 class="d2l-heading d2l-skeletize d2l-skeletize-width-65">${this.willBeSetLater}</h2>
```

That heading would be 65% width while the skeleton is visible, then be a normal block element later.

### What Isn't Included Here

**Paragraphs**

Notably present in the design and missing from there is some way to style a paragraph to look like a multi-line skeleton where the last line is 90% width. From the design:

Screen Shot 2020-09-15 at 3 53 31 PM

I started implementing this, but since it's not required in FACE right now I wanted to get what I had reviewed while the implementation for that paragraph stuff continues.

**Skeleton-Specific Components**

I've also avoided introducing actual components like `<d2l-skeleton>` or `<d2l-skeleton-chart>`. I still think a few components will be needed (maybe a `<d2l-skeleton-paragraph lines="5">` if I can't solve it using SVGs!) eventually, but for now I'm hoping Martin's innovative styles can simply be applied to existing elements -- allowing for a single render path.


components/demo/demo-snippet.js
```diff
@@ -66,14 +70,16 @@ class DemoSnippet extends LitElement {
...
 		return html`
 			<div class="d2l-demo-snippet-demo-wrapper">
 				<div class="d2l-demo-snippet-demo" dir="${dirAttr}">
 					<slot name="_demo"></slot>
 					<slot></slot>
 				</div>
 				<div class="d2l-demo-snippet-settings">
-					<d2l-switch text="RTL" ?on="${dirAttr === 'rtl'}" @change="${this._handleDirChange}"></d2l-switch>
+					<d2l-switch text="RTL" ?on="${dirAttr === 'rtl'}" @change="${this._handleDirChange}"></d2l-switch><br>
+					${skeleton}
```
dlockhart, line 82: This switch ends up looking like this in demo snippets that opt into skeletons:

Screen Shot 2020-09-16 at 6 13 20 PM

components/inputs/input-label-styles.js
```diff
@@ -3,14 +3,13 @@ import { css } from 'lit-element/lit-element.js';
...
+		display: inline-block;
 		font-size: 0.7rem;
 		font-weight: 700;
 		letter-spacing: 0.2px;
 		line-height: 1rem;
-		margin: 0;
-		padding: 0 0 7px 0;
-		width: 100%;
+		margin: 0 0 7px 0;
+		padding: 0;
```
dlockhart, line 12: In order for label skeletons to not be full-width, I switched them to `inline-block`. I moved the padding to margin so that the `7px` at the bottom would be outside the skeleton.

components/inputs/input-label-styles.js
```diff
@@ -27,6 +26,10 @@ export const inputLabelStyles = css`
 		left: auto;
 		right: 0.15rem;
 	}
+	:host([skeleton]) .d2l-input-label.d2l-skeletize::before {
+		top: 0.15rem;
+		bottom: 0.25rem;
+	}
```
dlockhart, line 32: This is for the cap height.

components/inputs/input-text.js
```diff
@@ -267,7 +265,7 @@ class InputText extends FormElementMixin(RtlMixin(LitElement)) {
 		`;
 		if (this.label && !this.labelHidden) {
 			return html`
-				<label class="d2l-input-label" for="${this._inputId}">${this.label}</label>
+				<div><label class="d2l-input-label d2l-skeletize" for="${this._inputId}">${this.label}</label></div>
 				${input}`;
```
dlockhart, line 269: Pretty simple process for a component to become "skeleton-aware"!

components/skeleton/demo/skeleton-test-typography.js
```diff
@@ -0,0 +1,67 @@
...
+			<a href="https://d2l.com" class="d2l-link d2l-skeletize">Link (Normal)</a><br>
+			<a href="https://d2l.com" class="d2l-link d2l-link-main d2l-skeletize">Link (Main)</a><br>
+			<a href="https://d2l.com" class="d2l-link d2l-link-small d2l-skeletize">Link (Small)</a>
+			<hr>
+			<h1 class="d2l-heading-1 d2l-skeletize d2l-skeletize-25">Heading 1</h1>
+			<h2 class="d2l-heading-2 d2l-skeletize d2l-skeletize-20">Heading 2</h2>
+			<h3 class="d2l-heading-3 d2l-skeletize d2l-skeletize-15">Heading 3</h3>
+			<h4 class="d2l-heading-4 d2l-skeletize d2l-skeletize-10">Heading 4</h4>
+			<hr>
+			<h3 class="d2l-heading-3 d2l-skeletize d2l-skeletize-15">Heading 3<br>spanning multiple lines</h3>
```
dlockhart, line 62: This component looks like this:

Screen Shot 2020-09-16 at 6 16 22 PM

And in skeleton-mode:

Screen Shot 2020-09-16 at 6 16 51 PM

components/skeleton/skeleton-mixin.js
```diff
@@ -0,0 +1,101 @@
...
+		bottom: 0;
+		content: '';
+		left: 0;
+		position: absolute;
+		right: 0;
+		top: 0;
+		z-index: 2000;
+	}
+	:host([skeleton]) .d2l-skeletize {
+		border-color: transparent;
```
dlockhart, line 25: @mdgbayly The one change I made to your original styles was to use a transparent border instead of `border: none`. Otherwise elements that had a border would change heights as they went from skeleton to not.

components/typography/styles.js
```diff
@@ -6,11 +6,19 @@ export const bodyStandardStyles = css`
 		font-weight: 400;
 		line-height: 1.4rem;
 	}
+	:host([skeleton]) .d2l-body-standard.d2l-skeletize::before {
+		top: 0.3rem;
+		bottom: 0.35rem;
+	}
```
dlockhart, line 12: These are all for the cap heights... good times this was.

components/typography/styles.js
```diff
@@ -45,11 +65,26 @@ export const heading1Styles = css`
...
+		top: 0.45rem;
+		bottom: 0.45rem;
+	}
 	@media (max-width: 615px) {
 		.d2l-heading-1 {
 			font-size: 1.5rem;
 			line-height: 1.8rem;
 		}
+		:host([skeleton]) .d2l-heading-1.d2l-skeletize {
+			height: 1.8rem;
```
dlockhart, line 82: The fixed heights on headings in skeleton mode is to handle that case where design would like multi-line headings to show up as a single line. So we set the height equal to the line height.

components/skeleton/skeleton-mixin.js
```diff
@@ -0,0 +1,101 @@
...
+		animation: loadingPulse 1.8s linear infinite;
+		background-color: var(--d2l-color-sylvite);
+		border-radius: 0.2rem;
+		bottom: 0;
+		content: '';
+		left: 0;
+		position: absolute;
+		right: 0;
+		top: 0;
+		z-index: 2000;
```
dbatiste, line 22: Do we need this z-index to be so high?  I think from @mdgbayly 's previous reply, it sounded like there wasn't a specific reason to choose this value, and that it quite possibly could have been something much smaller, perhaps even as low as 1.

components/skeleton/skeleton-mixin.js
```diff
@@ -0,0 +1,101 @@
...
+	}
+	:host([skeleton]) .d2l-skeletize-10::before {
+		width: 10%;
+	}
+	:host([skeleton]) .d2l-skeletize-5::before {
+		width: 5%;
+	}
+`;
+
+export const SkeletonMixin = dedupeMixin(superclass => class extends superclass {
```
dbatiste, line 89: Is the `dedupeMixin` for sure necessary?

components/skeleton/skeleton-mixin.js
dlockhart: Nope, not in my testing. I've set it to `1` as you suggested and we can adjust if it becomes necessary.

components/skeleton/skeleton-mixin.js
dlockhart: I don't think so, no -- I'll remove.

components/skeleton/skeleton-mixin.js
```diff
@@ -0,0 +1,100 @@
...
+		right: 0;
+		top: 0;
+		z-index: 1;
+	}
+	:host([skeleton]) .d2l-skeletize {
+		border-color: transparent;
+		box-shadow: none;
+		color: transparent;
+		position: relative;
+	}
```
dlockhart, line 28: One drawback to this approach is that you actually need a host element using the `SkeletonMixin` in order for the styles to get applied. This is fine for our core components, but if some app component just wanted to apply skeleton styles to a piece of typography (heading, link, etc), it's an extra step.

One solution to that could be to introduce another CSS class to force the skeleton ON, which would just add another selector to the styles, like this:

```css
:host([skeleton]) .d2l-skeletize,
.d2l-skeleton.d2l-skeletize {
   /* skeleton styles */
}
```

Then all a consumer would need to do would be `<h2 class="d2l-skeleton d2l-skeletize">` and then remove the `d2l-skeleton` class once things have loaded.

Thoughts? Can always add this later if it comes up.

dbatiste: Seems reasonable/useful.  Would they remove `d2l-skeleton` or `d2l-skeletize`?  Given the selector, it looks like ti wouldn't matter which they remove... just wondering what the "most correct" would be.

dlockhart: Yeah in this case adding/removing `d2l-skeleton` would be the most correct.

mdgbayly (commented):
Looking really good!

components/demo/demo-snippet.js
```diff
@@ -203,6 +217,25 @@ class DemoSnippet extends LitElement {
 		this._code = textNode.textContent;
 	}
 
+	_updateHasSkeleton() {
+
```
dlockhart, line 221: This code is auto-detecting whether the component in the demo snippet is skeleton-aware and if so, it shows the toggle.

components/inputs/input-label-styles.js
```diff
@@ -27,6 +26,13 @@ export const inputLabelStyles = css`
 		left: auto;
 		right: 0.15rem;
 	}
+	:host([skeleton]) .d2l-input-label.d2l-skeletize::before {
+		bottom: 0.25rem;
+		top: 0.15rem;
+	}
+	:host([skeleton][required]) .d2l-input-label.d2l-skeletize::after {
+		display: none;
+	}
```
dlockhart, line 35: This hides the "required" asterix when in skeleton-mode, which is what design would like.

components/inputs/input-text.js
```diff
@@ -140,8 +141,9 @@ class InputText extends FormElementMixin(RtlMixin(LitElement)) {
 				:host([hidden]) {
 					display: none;
 				}
-				label {
-					display: block;
+				.d2l-input-label {
+					display: inline-block;
+					vertical-align: bottom;
```
dlockhart, line 146: Block elements aren't great for skeletizing, since the skeleton fills the whole width instead of just the width of the text. Instead of making all `d2l-input-label`s inline -- which would break all the things -- I'm going to be a bit more surgical.

components/skeleton/skeleton-mixin.js
dlockhart: I ended up not doing this for now -- it adds a lot of complexity and bulk to the CSS. We can look into adding it later if necessary.

components/inputs/input-text.js
dlockhart: One side effect of this block -> inline change is that text inputs with labels have shifted down by `1px`, making the visual diffs `1px` taller. This mostly impacted the date input visual diffs.

dlockhart:
I think this PR is ready for review! I still plan to add documentation and "skeletize" the date/time inputs, but I'll do that in follow-up PRs. I added a few comments about things that have changed since the last round of changes.

dbatiste (approved):
LGTM!

mdgbayly (approved):
Looks great - keen for us to refactor our FACE implementation to use this once it's merged!
Thanks for putting in all this effort. These things always end up being bigger and more 'involved' than I initially expect!

PR #854, opened 2020-09-17
Add background-shading attribute

AllanKerr:
# Changes
- Add `background-shading` to the primary secondary template to allow either the primary panel, secondary panel, or neither to have their backgrounds shaded.

### `background-shading="primary"`
primary
### `background-shading="secondary"`
secondary


templates/primary-secondary/primary-secondary.js
```diff
@@ -70,6 +75,10 @@ class TemplatePrimarySecondary extends LitElement {
 			:host([primary-overflow="hidden"]) main {
 				overflow: hidden;
 			}
+			:host([background-shading="primary"]) main,
```
dbatiste, line 78: If an internal class is applied to the `div` with `class="d2l-template-primary-secondary-content"`, then it's possible to avoid reflecting/sprouting `none` on the public API.

dlockhart (approved):
Looks good! The `primary-secondary-desktop-fullscreen-width.visual-diff.html` is setting a background colour on its primary panel to simulate how folks have been using this -- it can likely be removed now.

templates/primary-secondary/primary-secondary.js
dlockhart: We could also just not set a default value, so nothing should sprout.

AllanKerr: I left the default value but moved the attribute to an inner `div` and disabled `reflect` so the interface should be more consistent with native elements.

dbatiste (approved):

PR #856, opened 2020-09-18
Update dropdown content position to respect ancestors with hidden or scrolling overflow.

dbatiste:
* updates logic to respect ancestors with hidden or scrolling overflow
* adds new getBoundingAncestor helper

Dialog showing calendar within - usable, but our stuff is so big, it makes it difficult to see the entire widget in a dialog. So, while large dropdown content may still be undesirable, this change addresses the current issue of overflow/clipping that makes the content inaccessible.
image


components/dropdown/test/dropdown-content-contained.visual-diff.js
```diff
@@ -0,0 +1,32 @@
...
+		'contained-bottom'
+	].forEach((testName) => {
+		it(testName, async function() {
+			const rect = await visualDiff.getRect(page, `#${testName}`);
+			const selector = `#${testName} d2l-dropdown`;
+			await helper.open(page, selector);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+	});
+
```
dlockhart, line 31: Where you expecting a diff to `d2l-dropdown-openers-context-menu.png `? It seems to be 4px less wide.

dlockhart (approved):
Looks great.

components/dropdown/test/dropdown-content-contained.visual-diff.js
dbatiste: Nope... I wasn't expecting that.  Really strange.

dbatiste: Looks like that was an actual regression related to my moving the width calc.  Applied fix and the diff image is back to original.  Just need to update golden again for my branch.

PR #860, opened 2020-09-22
Move to the new version of eslint config

svanherk:
All of this is rearranging so that variables and classes are not used before they are defined.

### DIFF
--- a/components/button/test/floating-buttons.visual-diff.js
+++ b/components/button/test/floating-buttons.visual-diff.js
@@ -7,6 +7,12 @@ describe('d2l-floating-buttons', () => {
 
 	let browser, page;
 
+	const scroll = async(page, selector, alignToTop) => {
+		await page.evaluate((selector, alignToTop) => {
+			document.querySelector(selector).scrollIntoView(alignToTop);
+		}, selector, alignToTop);
+	};
+
 	before(async() => {
 		browser = await puppeteer.launch();
 		page = await visualDiff.createPage(browser);
@@ -68,11 +74,5 @@ describe('d2l-floating-buttons', () => {
 		});
 
 	});
-
-	const scroll = async(page, selector, alignToTop) => {
-		await page.evaluate((selector, alignToTop) => {
-			document.querySelector(selector).scrollIntoView(alignToTop);
-		}, selector, alignToTop);
-	};
 
 });
--- a/components/form/form-element-localize-helper.js
+++ b/components/form/form-element-localize-helper.js
@@ -1,36 +1,12 @@
 import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
 import { tryGetLabelText } from './form-helper.js';
 
-export const localizeFormElement = (localize, ele) => {
-	if (ele.validity.valid) {
-		return null;
-	}
-	const tagName = ele.tagName.toLowerCase();
-	const labelText = tryGetLabelText(ele) || localize('components.form-element.defaultFieldLabel');
-	switch (tagName) {
-		case 'input':
-			return _localizeInputElement(localize, ele, labelText);
-		case 'textarea':
-			return _localizeInputTextElement(localize, ele, labelText);
+const _localizeGenericElement = (localize, ele, labelText) => {
+	switch (true) {
+		case ele.validity.valueMissing:
+			return localize('components.form-element.valueMissing', { label: labelText });
 		default:
-			return _localizeGenericElement(localize, ele, labelText);
-	}
-};
-
-const _localizeInputElement = (localize, ele, labelText) => {
-
-	const type = ele.type;
-	switch (type) {
-		case 'number':
-			return _localizeInputNumberElement(localize, ele, labelText);
-		case 'url':
-			return _localizeInputUrlElement(localize, ele, labelText);
-		case 'email':
-			return _localizeInputEmailElement(localize, ele, labelText);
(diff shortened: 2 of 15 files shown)


components/button/test/floating-buttons.visual-diff.js
```diff
@@ -7,6 +7,12 @@ describe('d2l-floating-buttons', () => {
 
 	let browser, page;
 
+	const scroll = async(page, selector, alignToTop) => {
```
svanherk, line 10: For the visual diff helpers, I just moved them to the top

components/form/form-element-localize-helper.js
```diff
@@ -1,36 +1,12 @@
 import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
```
svanherk, line 1: The diff for this is horrible, but all the helpers were moved above the exported const that uses them

helpers/gestures.js
```diff
@@ -19,9 +19,9 @@ function handleTouchStart(e) {
 
 	const reset = () => {
 		tracking = null;
-		node.removeEventListener('touchend', handleTouchEnd);
-		node.removeEventListener('touchermove', handleTouchMove);
-		node.removeEventListener('touchcancel', handleTouchCancel);
+		node.removeEventListener('touchend', handleTouchEnd); /* eslint-disable-line no-use-before-define */
+		node.removeEventListener('touchermove', handleTouchMove); /* eslint-disable-line no-use-before-define */
+		node.removeEventListener('touchcancel', handleTouchCancel); /* eslint-disable-line no-use-before-define */
```
svanherk, line 24: There's a circular dependency here where these functions call `reset`.  We can disable the linting in this case, or we could move `reset` to be a `function` instead of a `const`?

dlockhart: Good with either option.

dlockhart (approved):

components/list/list-item-drag-drop-mixin.js
```diff
@@ -37,6 +37,214 @@ const isDragSupported = () => {
 	return typeof el.ondragenter === 'function';
 };
 
+class DragState {
```
dbatiste, line 40: It's unfortunate to have to bring this up since the main focus is usually the component.  Not a big deal.

components/menu/test/menu-checkbox.visual-diff.js
```diff
@@ -7,6 +7,22 @@ describe('d2l-menu checkbox', () => {
 
 	let browser, page;
 
+	const contentResize = (page, selector) => {
```
dbatiste, line 10: Beyond the scope of this PR, but if you wanted to, you _could_ simply this code a little by:
``` javascript
const { oneEvent } = require('@brightspace-ui/visual-diff/helpers');

const click = async(page, selector) => {
    const resize = oneEvent(page, selector, 'd2l-menu-item-change');
    await page.$eval(selector, (item) => item.click());
    return resize;
};
```
I think the `await` in there isn't really doing anything, in which case that could be removed, and the function itself would not need to be `async`.

The `oneEvent` helper is new - I recently did a little refactor to the visual-diff library and added that as a helper, since waiting for a single event is a common pattern.

components/more-less/test/more-less.visual-diff.js
```diff
@@ -7,6 +7,25 @@ describe('d2l-more-less', () => {
 
 	let browser, page;
 
+	const contentResize = (page, selector) => {
+		return page.evaluate((selector) => {
+			return new Promise((resolve) => {
+				const elem = document.querySelector(selector);
+				elem.shadowRoot.querySelector('.d2l-more-less-content').addEventListener('transitionend', (e) => {
+					if (e.propertyName === 'height') resolve();
```
dbatiste, line 15: This one could almost be simplified like the other one, except we'd have to enhance the `oneEvent` helper to accept some type of option/predicate to check the property name.  Typically this check is made when there could be multiple events being triggered.  If we did enhance that function, we'd have to be careful though - it can't be a function predicate that is passed... whatever is passed into `evaluate`/`$eval`/`$$eval` mist be serializable. Again, technically beyond your current goal here.  :)

mixins/async-container/test/async-container-mixin.visual-diff.js
```diff
@@ -7,6 +7,16 @@ describe('d2l-async-container-mixin', function() {
 
 	let browser, page;
 
+	const getAsyncStateEvent = (page, selector, state) => {
+		return page.$eval(selector, (elem, state) => {
+			return new Promise((resolve) => {
+				elem.addEventListener('d2l-async-demo-container-changed', (e) => {
```
dbatiste, line 13: A slightly different flavor of oneEvent could handle this... except instead of checking an animation event property, it's checking some custom data.  🤔

dbatiste (approved):

components/list/list-item-drag-drop-mixin.js
svanherk: Yeah, we could turn off the rule for classes if we want them to always go at the end, but I also seems good to leave on

components/menu/test/menu-checkbox.visual-diff.js
svanherk: Updated this one and the `menu-radio` tests!  Will leave the other two for a more focused effort

PR #861, opened 2020-09-23
fix: add check for opener position to know if content is absolute wrt document

dbatiste:
Dropdown content clipping would normally be an issue in the LMS also, except the LMS has some "special" styles that override the opener's default position from `relative` to `static`, effectively making the dropdown content positioned `absolute` with respect to the document (instead of the opener within a bounding container).  In such a case, we do not need adjust the dropdown content position and dimensions according to what would normally be the bounding container (the wrapper element within `d2l-scroll-wrapper`).

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -469,8 +469,12 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		this._width = null;
 		await this.updateComplete;
 
+		const openerPosition = window.getComputedStyle(opener, null).getPropertyValue('position');
 		const boundingContainer = getBoundingAncestor(target);
 		const boundingContainerRect = boundingContainer.getBoundingClientRect();
+
+		// position check in case consuming app (LMS) has overriden position to make content absolute wrt document
+		const bounded = (openerPosition === 'relative' && boundingContainer !== document.documentElement);
 
 		const adjustPosition = async() => {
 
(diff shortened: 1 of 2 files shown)


components/dropdown/dropdown-content-mixin.js
```diff
@@ -469,37 +469,42 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		this._width = null;
 		await this.updateComplete;
 
+		const openerPosition = window.getComputedStyle(opener, null).getPropertyValue('position');
 		const boundingContainer = getBoundingAncestor(target);
 		const boundingContainerRect = boundingContainer.getBoundingClientRect();
 
+		// position check in case consuming app (LMS) has overriden position to make content absolute wrt document
+		const bounded = (openerPosition === 'relative' && boundingContainer !== document.documentElement);
```
dbatiste, line 477: As mentioned in the description, the LMS has special styles to make the opener positioned static, which results in the content absolutely positioned with respect to the document, meaning we can/should opt out of positioning the dropdown content relative to the bounding container (the `d2l-scroll-wrapper`).

components/dropdown/dropdown-content-mixin.js
```diff
@@ -469,37 +469,42 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
 		const adjustPosition = async() => {
 
 			const targetRect = target.getBoundingClientRect();
 			contentRect = contentRect ? contentRect : content.getBoundingClientRect();
 			const headerFooterHeight = header.getBoundingClientRect().height + footer.getBoundingClientRect().height;
 
 			let spaceAround;
-			if (boundingContainer === document.documentElement) {
+
+			if (bounded) {
```
dbatiste, line 487: Note: I just reversed the if/else here so avoid the negation.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -537,7 +542,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		};
 
 		const scrollWidth = Math.max(header.scrollWidth, content.scrollWidth, footer.scrollWidth);
-		const availableWidth = (boundingContainer === document.documentElement ? window.innerWidth - 40 : boundingContainerRect.width - 60);
+		const availableWidth = (bounded ? boundingContainerRect.width - 60 : window.innerWidth - 40);
```
dbatiste, line 545: Same here... I reversed things to avoid negation of the `bounded` value.

PR #862, opened 2020-09-24
Skeleton documentation

dlockhart:


components/skeleton/README.md
```diff
@@ -0,0 +1,104 @@
...
+Once a component is skeleton-aware, it can apply skeleton styles to native elements within its render root using the `d2l-skeletize` CSS class. These native elements can be anything, including our own headings, links and standard/compact/small body styles.
+
+```javascript
+render() {
+  return html`
+    <h2 class="d2l-heading-2 d2l-skeletize">Heading</h2>
+    <span class="d2l-body-compact d2l-skeletize">Description</span>
+    <a class="d2l-link d2l-skeletize" href="somewhere">Link</a>
+    <div class="widget d2l-skeletize">Widget</div>
+    <d2l-input-text label="text input" ?skeleton="{this.skeleton}"></d2l-input-text>
```
dlockhart, line 61: This example where the native elements use the `d2l-skeletize` CSS class but custom elements use the `skeleton` attribute makes me want to change them to be named consistently. What do you think? Options are obviously either "skeleton" or "skeletize".

components/inputs/demo/input-select-test.js
```diff
@@ -28,15 +29,17 @@ class TestInputSelect extends RtlMixin(LitElement) {
...
-			<select
-				aria-label="Choose a dinosaur:"
-				aria-invalid="${invalid}"
-				class="d2l-input-select"
-				?disabled="${this.disabled}">
-				<option>Tyrannosaurus</option>
-				<option>Velociraptor</option>
-				<option>Deinonychus</option>
-			</select>
+			<div class="d2l-skeletize">
```
dlockhart, line 32: Native `<select>`s don't support `::before` so I needed to do a special workaround that merited a call-out in the docs below.

components/skeleton/README.md
mdgbayly: yeh - i've wondered about that several times. Initially, I think I opted for different names because I hadn't included the recommended `d2l-` prefix on the class name so it seemed clearer in the css to name them differently. But with the usage of the `d2l-` prefix that argument is less compelling.  In my head, I still sort of like having them named differently because it feels like it makes it clearer that they have different roles, but that could be just because I'm so used to it by now. So I'm open to going with whatever you guys prefer.  If we go with one value, then `skeleton` or `skeletize`. I liked `skeletize` for the class name as it sounds more "verby" whereas the attribute was more "nouny" but again that might be just me being invested in those terms. Either works.

dlockhart: Now I'm also a bit torn... I agree they have subtly different roles -- the "skeletize" CSS class is more about what you'd like it to do when the host component is in skeleton-mode, vs. the "skeleton" attribute is you telling it to go into that mode. But I don't know that consumers of this will notice.

If we're going along with other native boolean attributes, maybe we should use "skeletized" to match its current state instead of a verb -- i.e. along the lines of native boolean attributes like "disabled", "required" and "hidden"?

dbatiste: I think this is why I previously was curious about which class the consumer was going to remove (skeleton sounding more like it represents a component that has a skeleton or is a skeleton, vs skeletize sounding like it's in its skeleton state. Haha.  👻

dlockhart: At this point I think I'm going to leave it... at the very least I'll get this PR in and convert a few more components over. FACE will be the only place using this in the short-term, so there's still a bit of time to change the wording around if we feel strongly about it.

PR #865, opened 2020-09-24
Zero max fraction digits bug

krshnpatel-d2l:
I noticed for max-fraction-digits=0 (integers only), the max-fraction-digits was resolving to "undefined".

So this PR includes the following:
~ remove conditions inside options for formatting number (not needed anymore because there is no loss of type since the function is not exported anywhere; i.e. undefined changes to type String when passed in from another component)
~ do not export formatValue function (not used anywhere else)
~ added unit tests for max-fraction-digits=0
~ included unit test for checking default value on "title" property


dlockhart (approved):
Nice, simpler IS better!

PR #867, opened 2020-09-24
skeletize number inputs

dlockhart:
Adds skeleton support to numeric inputs. This is going to fail visual diff until I get the goldens updated.


krshnpatel-d2l:
Nice! Looks good on my local env 👍

krshnpatel-d2l (approved):

PR #869, opened 2020-09-25
skeletize checkboxes

dlockhart:
This makes our `d2l-input-checkbox` component skeleton-aware.

I could almost get away with simply applying `d2l-skeletize` to the actual checkbox input, but that doesn't work for legacy-Edge so I had to resort to wrapping it and skeletizing that wrapper. Text inputs and selects are going to have the same issue, so I'll be adding a section to the skeleton documentation to let folks know how to work around this limitation.


components/inputs/input-checkbox.js
```diff
@@ -1,16 +1,58 @@
 import '../colors/colors.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
-import { checkboxStyles } from './input-checkbox-styles.js';
```
dlockhart, line 3: I pulled these styles into here as part of this change as they were only referenced by the list component. This will cut down on an extra web request in most scenarios.

components/inputs/input-checkbox.js
```diff
@@ -85,9 +130,16 @@ class InputCheckbox extends RtlMixin(LitElement) {
 					margin-left: 0;
 					margin-right: 0;
 				}
+				:host([skeleton]) .d2l-input-checkbox-text.d2l-skeletize::before {
+					bottom: 0.3rem;
+					top: 0.3rem;
```
dlockhart, line 135: Since checkboxes don't use one of our standard typography styles, I needed to explicitly adjust the top/bottom to get the right cap height on the skeleton.

components/inputs/input-checkbox.js
```diff
@@ -85,9 +130,16 @@ class InputCheckbox extends RtlMixin(LitElement) {
...
+				}
 				.d2l-input-checkbox-text-disabled {
 					opacity: 0.5;
 				}
+				:host([skeleton]) .d2l-input-checkbox-text-disabled {
+					opacity: 1;
```
dlockhart, line 141: Otherwise disabled checkboxes' skeleton over the text is 50% opacity.

components/skeleton/skeleton-mixin.js
```diff
@@ -27,7 +27,7 @@ export const skeletonStyles = css`
 	}
 	:host([skeleton]) .d2l-skeletize {
 		background-color: transparent;
-		border-color: transparent;
+		border-color: var(--d2l-color-sylvite);
```
dlockhart, line 30: @mdgbayly One more adjustment to this, so now it's gone from `none` -> `transparent` -> `sylvite`. Without this, components that do have borders (like checkboxes) end up looking like their skeletons have shrunk by the border width. By only setting the colour, elements that don't have borders won't be impacted.

dbatiste (approved):

PR #874, opened 2020-09-25
Skeletize date/time components

dlockhart:
This skeletizes all of our date/time components, plus fieldset since it was used internally by them.


components/demo/demo-snippet.js
```diff
@@ -122,7 +122,7 @@ class DemoSnippet extends LitElement {
 							nodes[i].setAttribute(name, value);
 						}
 					}
-					if (nodes[i].shadowRoot) {
+					if (applyToShadowRoot && nodes[i].shadowRoot) {
```
dlockhart, line 125: Slight change here as we don't want it to travel into shadow roots like the RTL one does and turn skeletons on -- a web component is responsible for passing down `skeleton` to any components inside itself.

components/inputs/input-date.js
```diff
@@ -176,7 +177,8 @@ class InputDate extends FormElementMixin(LocalizeCoreElement(LitElement)) {
 			? html`<d2l-icon icon="tier1:alert" slot="left" style="${styleMap({ color: 'var(--d2l-color-cinnabar)' })}"></d2l-icon>`
 			: html`<d2l-icon icon="tier1:calendar" slot="left"></d2l-icon>`;
 		const errorTooltip = (this.validationError && !this._dropdownOpened && this.childErrors.size === 0) ? html`<d2l-tooltip align="start" announced for="${this._inputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
-		const infoTooltip = (this._showInfoTooltip && !errorTooltip && !this.invalid && this.childErrors.size === 0) ? html`<d2l-tooltip align="start" announced delay="1000" for="${this._inputId}">${this.localize(`${this._namespace}.openInstructions`, { format: shortDateFormat })}</d2l-tooltip>` : null;
+		const infoTooltip = (this._showInfoTooltip && !errorTooltip && !this.invalid && this.childErrors.size === 0 && !this.skeleton) ? html`<d2l-tooltip align="start" announced delay="1000" for="${this._inputId}">${this.localize(`${this._namespace}.openInstructions`, { format: shortDateFormat })}</d2l-tooltip>` : null;
+
```
dlockhart, line 181: This had a bit of special handling so that the info tooltip wouldn't show up while the skeleton was on and clicking on it wouldn't trigger the dropdown.

components/inputs/input-text.js
```diff
@@ -249,7 +250,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 					?autofocus="${this.autofocus}"
 					@change="${this._handleChange}"
 					class="${classMap(inputClasses)}"
-					?disabled="${this.disabled}"
+					?disabled="${disabled}"
```
dlockhart, line 253: Disabling text inputs when skeletized so that you can't tab into nothingness.

margaree (approved):

PR #877, opened 2020-09-30
adding an entrypoint for all the form documentation

dlockhart:
I think having a single place we can send people (and link to in other places) to learn about forms would be nice. Also this makes browsing the source on GitHub nicer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,7 @@ npm install @brightspace-ui/core
   * [Dropdowns](components/dropdown/): dropdown openers and content containers
   * [Expand Collapse](components/expand-collapse): component to create expandable and collapsible content
   * [Focus Trap](components/focus-trap/): generic container that traps focus
-  * Forms
-	* [Forms](components/form/docs/form.md): Forms that aggregate data so the user can control how it is submitted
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -59,9 +56,11 @@ npm install @brightspace-ui/core
 * Mixins
   * [ArrowKeysMixin](mixins/arrow-keys-mixin.md): manage focus with arrow keys
   * [AsyncContainerMixin](mixins/async-container/): manage collective async state
+  * [FormElementMixin](components/form/docs/form-element-mixin.md): allow components to participate in forms and validation
   * [LocalizeMixin](mixins/localize-mixin.md): localize text in your components
   * [ProviderMixin](mixins/provider-mixin.md): provide and consume data across elements in a DI-like fashion
   * [RtlMixin](mixins/rtl-mixin.md): enable components to define RTL styles
+  * [SkeletonMixin](components/skeleton/): make components skeleton-aware
```
dlockhart, line 63: I added some special call-outs here for the mixin (and the skeleton one) despite these things living in `components`. I like the idea of grouping all the mixins together like this, but am fine to remove this if others disagree.

AllanKerr (approved):

PR #885, opened 2020-10-01
Fix tooltip console warning when invalid

### LINKED ISSUE #883: Numeric Input: tooltip console warning when invalid
When you put the numeric input into an invalid state, a console warning appears:

> d2l-tooltip may be being used in a non-accessible manner; it should be attached to interactive elements like 'a', 'button','input' 'select', 'textarea' or static / custom elements if a role has been set and the element is focusable.

This is caused by the `for` of the tooltip pointing at the `d2l-input-text` element and not the underlying native input. For/id references can't cross shadow boundaries though anyway.

I believe the date/time inputs get around this by using the `announced` attribute on the tooltip, which will cause it to be announced and not rely on the for/id linkage.

krshnpatel-d2l:
## Fixes #883

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -88,7 +88,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 	}
 
 	render() {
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Thanks that was fast!

PR #888, opened 2020-10-05
allowing input width to be set individually from the label width

dlockhart:
Currently, we're setting a width of `4rem` on the entire numeric input component. While that's a good default width for the input box itself, when there's a longer label it's also constraining the label:

Screen Shot 2020-10-05 at 1 19 45 PM

While less visible, this is also an issue with our standard `d2l-input-text`.

This change treats the width of the `d2l-input-number` component like our other inputs -- it'll take up as much space as is provided, while restricting the input box itself to `4rem`.

For folks who want a larger input box, they can no longer go `<d2l-input-number style="width: 10rem">` like they could before, so this also introduces a new `input-width` attribute to let them customize that. `d2l-input-text` has the same attribute.


components/inputs/test/input-number.test.js
```diff
@@ -132,17 +132,23 @@ describe('d2l-input-number', () => {
...
 			setTimeout(() => documentLocaleSettings.language = 'fr');
-			await oneEvent(elem, 'd2l-localize-behavior-language-changed');
-			await elem.updateComplete;
-			expect(getInnerInputValue(elem)).to.equal('');
+			return Promise.all([
+				oneEvent(elem, 'd2l-localize-behavior-language-changed'),
+				elem.updateComplete
+			]).then(() => {
+				expect(getInnerInputValue(elem)).to.equal('');
+			});
```
dlockhart, line 151: These tests (and the similar ones in time input) have been flaky since I added them last week. I suspect what's happening is the event and the `updateComplete` promises could be resolving out-of-order, so switching to a `Promise.all` instead which won't care about the sequence.

krshnpatel-d2l (approved):
🎉 🎉

dbatiste (approved):

PR #889, opened 2020-10-06
Add request instance helper.

dbatiste:
This PR factors the `requestInstance` logic out into a helper that can be used outside the context of a web component.  I've run into a case where I need the ability to request an instance from a custom tinyMCE plugin.

### DIFF
--- a/mixins/provider-mixin.js
+++ b/mixins/provider-mixin.js
@@ -24,7 +24,7 @@ export function requestInstance(node, key) {
 	});
 	node.dispatchEvent(event);
 	return event.detail.instance;
-};
+}
 
 export const RequesterMixin = superclass => class extends superclass {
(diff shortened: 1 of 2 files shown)


dlockhart (approved):
Looks good. I'm guessing you'll need to install this on a global from BSI -- which fine, that's the only way to do it today... but man it'd be great to be able to use JS modules from monolith code.

dbatiste:
Indeed it would.  I was considering factoring the provider out too.  After all, the mixin is basically attaching the `Map` to the node.  It would enable using it in that LMS provider code I just wrote.  Just not sure we want to advocate folks doing that.  Thoughts?

dlockhart:
If it removes duplicates of the event name (which we might change here without realizing it's used elsewhere) I'm all for it.

PR #895, opened 2020-10-16
Primary Secondary Template: Resizable + Mobile Drawer

AllanKerr:
# Primary Secondary Template
- Small tweaks will likely be made to this PR because I'm still doing some cross-device testing but this should contain the vast majority of the logic in a mostly final state.
- The PR is large but has been broken up into a few logical commits that should make it easier to review. 
### Included
- Add mobile drawer mode that allows mobile users to expand / collapse the secondary panel regardless of whether the page template is in resizable or fixed mode.
- Add resizable desktop mode that allows desktop users to drag the divider with their mouse or resize using their right and left arrow keys.
### Not Included
- Collapsing the header in mobile view
- Divider bounce when hitting the resizing bounds when resizing using arrow keys.
- Storing the user's divider position.

### Desktop Fixed
desktop-fixed
### Desktop Resizable
desktop-resizable
### Mobile Resizable
mobile-resizable


templates/primary-secondary/primary-secondary.js
```diff
@@ -1,5 +1,421 @@
...
+	}
+
+	_trackTouch(touch) {
+		if (this._touches.length === 5) {
+			this._touches.shift();
+		}
+		this._touches.push(touch.screenY);
+	}
+
+}
```
dlockhart, line 418: Wow that's a lot of different resizers!

templates/primary-secondary/primary-secondary.js
```diff
@@ -46,88 +477,390 @@ class TemplatePrimarySecondary extends LitElement {
...
+			:host([dir="rtl"]) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-right {
+				display: none;
+			}
+			d2l-icon {
+				display: none;
+			}
+
 			footer {
-				background-color: #ffffff;
+				background-color: white;
```
dlockhart, line 645: I think we'd talked about applying the white background to the header as well? Doesn't need to be part of this change though.

templates/primary-secondary/primary-secondary.js
```diff
@@ -46,88 +477,390 @@ class TemplatePrimarySecondary extends LitElement {
...
+									</svg>
+								</d2l-icon-custom>
+							</div>
+							<div class="d2l-template-primary-secondary-divider-handle-mobile">
+								${this._size === 0 ? html`<d2l-icon icon="tier1:chevron-up"></d2l-icon>` : html`<d2l-icon icon="tier1:chevron-down"></d2l-icon>`}
+							</div>
+						</div>
+					</div>
+					<div style=${styleMap(secondaryPanelStyles)} class="d2l-template-primary-secondary-secondary-container" @transitionend=${this._onTransitionEnd}>
+						<aside>
```
dlockhart, line 860: Nit: could the `aside` and parent `div` be combined so it's just the `aside`?

templates/primary-secondary/test/primary-secondary-mobile.visual-diff.html
```diff
@@ -0,0 +1,234 @@
...
+			<div slot="secondary">
+				Students of all ages get more engaged when they share their progress, reflections, and learning
+				experiences
+				with their classmates, teacher, and parents. Our Portfolio tool, as well as built-in badges and
+				certificates
+				provide instant recognition and keep students motivated. Parents love them too.
+			</div>
+			<div slot="footer">Footer</div>
+		</d2l-template-primary-secondary>
+	</div>
```
dlockhart, line 54: Very elegant!

dlockhart (commented):
First pass at this looks good to me... like there's a lot going on so it's hard to really dig too deeply into things, but that would be true for any review of like a month+ of work all at once. :)

templates/primary-secondary/primary-secondary.js
AllanKerr: The `aside` is contained within a `div` so that the `aside` can maintain its width while the parent div collapses. This allows the secondary panel to be collapsed and expanded without losing scroll position. It also means that when the secondary panel collapses it looks like it slides off the page rather than resizing to a width of `0`.

templates/primary-secondary/primary-secondary.js
AllanKerr: There is a little bit of duplication but it makes debugging and refactoring infinitively easier since any of them can be disabled with a single line change to isolate resizing behavior.

templates/primary-secondary/primary-secondary.js
dlockhart: Ah cool makes sense!

dlockhart (approved):

PR #899, opened 2020-10-21
fix null exception in disconnectedCallback

stevenao:
https://app.circleci.com/pipelines/github/Brightspace/assignments-ui/6/workflows/61410bab-e85d-4017-b971-358c3f68be53/jobs/1835

### DIFF
--- a/components/more-less/more-less.js
+++ b/components/more-less/more-less.js
@@ -118,8 +118,8 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 			this.__contentSlot.removeEventListener('slotchange', this.__reactToChanges.bind(this));
 			this.__contentSlot.removeEventListener('slotchange', this.__startObserving.bind(this));
 		}
-		this.__content.removeEventListener('focusin', this.__focusIn.bind(this));
-		this.__content.removeEventListener('focusout', this.__focusOut.bind(this));
+		this.__content && this.__content.removeEventListener('focusin', this.__focusIn.bind(this));
+		this.__content && this.__content.removeEventListener('focusout', this.__focusOut.bind(this));
 		this.shadowRoot.removeEventListener('transitionstart', this.__bound_transitionEvents);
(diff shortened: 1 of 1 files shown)


components/more-less/more-less.js
```diff
@@ -118,8 +118,8 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 			this.__contentSlot.removeEventListener('slotchange', this.__reactToChanges.bind(this));
 			this.__contentSlot.removeEventListener('slotchange', this.__startObserving.bind(this));
 		}
-		this.__content.removeEventListener('focusin', this.__focusIn.bind(this));
-		this.__content.removeEventListener('focusout', this.__focusOut.bind(this));
+		this.__content && this.__content.removeEventListener('focusin', this.__focusIn.bind(this));
```
dbatiste, line 121: What the scenario where `this.__content` is null/undefined?

Aside from that, I don't think this is doing what was intended because `bind` produces a new function, which isn't the function that was added, but I know that's what the original was doing.  I really wish this was wiring up handlers in `render` instead, which simplifies a lot of this, but I feel like that's a bigger change to clean up.

stevenao: It is failing in the clean up stage at one of the unit test in `assignment-ui`. the `more-less` component is used in the `d2l-assignment-viewer`. I don't see it doing anything special usage of it there. It is weird. So, I am not sure exactly what happen. My intuition is that it doesn't hurt to just put some null check there.

dbatiste: Yeah doesn't hurt to add... just really strange because that element gets created on the first render, and this variable is assigned in `firstUpdated`.  Is it possible the test isn't waiting for that before detaching the element?

dbatiste: Nit: if it were me, I would make it consistent with the conditional check that is almost exactly the same immediately above.  ;)

dbatiste (approved):

AllanKerr:
Just a heads up @stevenao , you'll need to merge in `master` for CI to pass.

PR #901, opened 2020-10-22
input-date-time-range: add inclusiveDateRange and autoShiftDates

margaree:
This is to mirror the control. Note that in the control, `autoShiftDates` is true by default but I have left it as false by default here for now.

I also added options to hide and start and end labels. It might make more sense to combine these into one property (e.g., `childLabelsHidden`) as I doubt there would be a case where one would be hidden and the other visible.

To do:
- [x] update README once properties seem good to go


dlockhart (approved):
Yeah agreed that `child-labels-hidden` is likely better.

margaree:
Replaced by https://github.com/BrightspaceUI/core/pull/902

PR #905, opened 2020-10-23
Primary Secondary: persist divider size to local storage

AllanKerr:
# Changes
- Save divider size to local storage so its remembered.
- The key is based on the window location so each page will have its own size stored.


templates/primary-secondary/primary-secondary.js
```diff
@@ -885,6 +893,10 @@ class TemplatePrimarySecondary extends RtlMixin(LitElement) {
 		};
 	}
 
+	_computeSizeKey() {
+		return `d2l-primary-secondary-${window.location.href}`;
```
dlockhart, line 897: I'm not sure how well this will work in practice... most URLs contain the orgUnitId in the query-string or even worse as part of the URL itself. Wonder if we want to introduce something like a `memory-key` attribute or something that could be used for this kind of thing, i.e. `<d2l-template-primary-secondary memory-key="quick-eval">`?

If this came up a few times, we could even throw it in a mixin.

templates/primary-secondary/primary-secondary.js
```diff
@@ -902,15 +914,22 @@ class TemplatePrimarySecondary extends RtlMixin(LitElement) {
 
 		if (this._size === undefined) {
 			// initialize size on first resize
-			if (this._isMobile) {
-				this._size = this._contentBounds.minHeight;
+			const key = this._computeSizeKey();
+			const size = parseFloat(localStorage.getItem(key));
```
dlockhart, line 918: You may want to test this in incognito mode in a few browsers... this used to throw in those scenarios, or if the locale storage was "full".

templates/primary-secondary/primary-secondary.js
AllanKerr: Good call, having to add a property for it is unfortunate but probably for the best.

dlockhart (approved):
Looks awesome!

PR #906, opened 2020-10-23
feat: adding a "supporting" slot to menu items

dlockhart:
This new slot enables supporting information to be displayed to the right of all menu item types. I chose "supporting" as a name to align with lists but also to leave the door open for a "secondary" slot down the road. Secondary = a second line of text vs. "supporting" is an "aside" bit of info.

It can be text, like in this example in our media player:
Screen Shot 2020-10-23 at 11 57 22 AM

Or other supporting information like in filters:
Screen Shot 2020-10-23 at 3 08 02 PM

In order to get this to work consistently across the different menu types, I've broken the styles out a bit so they can be better shared. I've introduced `d2l-menu-item-text` as a CSS class and in doing so created a breaking change in the `menuItemSelectableStyles` -- I'm going to follow up by fixing the face-filter-sort's usage of this.


components/menu/menu-item-mixin.js
```diff
@@ -94,7 +94,7 @@ export const MenuItemMixin = superclass => class extends superclass {
 	}
 
 	__initializeItem() {
-		const slot = this.shadowRoot.querySelector('slot');
+		const slot = this.shadowRoot.querySelector('slot:not([name])');
```
dlockhart, line 97: Now that there are 2 slots this had to be adjusted to look for a nameless slot.

components/menu/menu-item-selectable-styles.js
```diff
@@ -8,17 +8,10 @@ export const menuItemSelectableStyles = [ menuItemStyles,
 			align-items: center;
 			display: flex;
 			padding: 0.75rem 1rem;
-
 		}
 
-		span {
```
dlockhart, line 14: This is the breaking change -- from a `span` to just using the `d2l-menu-item-text` styles defined in the base shared styles.

components/menu/menu-item-styles.js
```diff
@@ -44,4 +44,23 @@ export const menuItemStyles = css`
...
 	}
+
+	.d2l-menu-item-text {
+		flex: auto;
+		line-height: 1rem;
+		overflow-x: hidden;
+		overflow-y: hidden;
+		text-overflow: ellipsis;
+		white-space: nowrap;
+	}
```
dlockhart, line 55: This replaces the generic `span` CSS duplicated in `menu-item.js` and `menu-item-selectable-styles.js`.

components/menu/menu-item.js
```diff
@@ -15,22 +15,18 @@ class MenuItem extends MenuItemMixin(LitElement) {
...
-					overflow-y: hidden;
-					text-overflow: ellipsis;
-					white-space: nowrap;
-				}
-
-				:host > d2l-icon {
+				d2l-icon {
 					flex: none;
-					margin-top: 0.1rem;
+					margin-left: 6px;
```
dlockhart, line 25: I added `6px` of left margin to the ">" icon to make sure the supporting text (and the menu item text too) doesn't butt up against it. I verified this value with Jeff G.

dbatiste (approved):
Looks great!

Jevan5 (approved):
Great features 👍

PR #909, opened 2020-10-26
Dark Mode for Menus & Dropdown Menus

dlockhart:
This adds limited support for a "dark mode" theme for our menus and dropdown menus, for limited use from the media player. 
I did not add any demos so that we can keep this relatively under wraps.

Screen Shot 2020-10-23 at 5 43 11 PM

I introduced a very basic `ThemeMixin` that we could one day extend to automatically populate the `theme` attribute based on a user's OS preference. For now, media player will need to set the theme manually.


components/menu/menu-item-selectable-styles.js
```diff
@@ -14,13 +14,6 @@ export const menuItemSelectableStyles = [ menuItemStyles,
 			text-decoration: none;
 		}
 
-		:host(:hover),
-		:host(:focus),
-		:host(:hover) > d2l-icon,
-		:host(:focus) > d2l-icon {
-			color: var(--d2l-color-celestine-minus-1);
-		}
```
dlockhart, line 22: This is a slight change in behaviour so that the checkbox behaviour on hover/focus matches that of the chevron -- it also simplifies how theming would work.

components/menu/demo/menu.html
```diff
@@ -20,6 +20,35 @@
 
 		<d2l-demo-page page-title="d2l-menu">
 
+			<h2>Dark Theme</h2>
```
margaree, line 23: Did you mean to commit this demo? Just wondering based on the PR description

components/menu/menu-item-styles.js
```diff
@@ -3,27 +3,26 @@ import { css } from 'lit-element/lit-element.js';
...
 export const menuItemStyles = css`
 	:host {
-		background-color: #ffffff;
+		background-color: var(--d2l-menu-background-color);
 		border: 1px solid transparent;
-		border-top-color: var(--d2l-color-gypsum);
+		border-top-color: var(--d2l-menu-border-color);
 		box-sizing: border-box;
-		color: var(--d2l-color-ferrite);
+		color: var(--d2l-menu-foreground-color);
```
dbatiste, line 10: Do you want to make a similar change to line #6 in `dropdown-content-styles.js`?  It think it's the color that would be used for generic dropdown content (not a menu).  It looks like some of the other style changes apply to that scenario, so it seems _partly_ supported in the current state.

components/menu/demo/menu.html
dlockhart: I did not! Will remove thx.

components/menu/menu-item-styles.js
dlockhart: So currently only `<d2l-dropdown-menu>` supports `theme="dark"` (it's the only one with the mixin)... so technically I think we can get away without changing `dropdown-content-styles.js`?

dlockhart: Actually no you're right... since `dropdown-content-styles.js` is already using some of these variables, they NEED to be defined with it otherwise dropdowns that aren't menus won't work -- which is why a bunch of visual diffs failed.

dlockhart: Ok fixed!

dbatiste (approved):
Looks good once goldens are fixed up.  😄

PR #910, opened 2020-10-27
fix: do not render href if null

dlockhart:
The enrollment card Polymer element is binding the card's `href` to a `null` value. Back when this was Polymer, that resulted in no `href` being set, but that's not the case in Lit so we need to handle it specially.


margaree (approved):

svanherk:
Thanks @dlockhart!

PR #913, opened 2020-10-28
Making dark dropdown menus 90% opacity

dlockhart:
Jeff would like the new "dark mode" menus to have a 90% opacity.

They'll look like this:

Screen Shot 2020-10-28 at 10 32 22 AM


test/styles.css
```diff
@@ -7,3 +7,12 @@ body {
...
 	margin-bottom: 30px;
 }
+.visual-diff-dark {
+	background: repeating-linear-gradient(
+		45deg,
+		#606dbc,
+		#606dbc 10px,
+		#465298 10px,
+		#465298 20px
+	);
```
dlockhart, line 17: I'd like to use this striped background for other dark mode visual diffs that have some opacity to them. It'll help us catch regressions that accidentally break the opacity.

dbatiste: Cool.  This might be better for the background that's used for the translucent button visual-diff too.

components/dropdown/dropdown-content-styles.js
```diff
@@ -19,21 +21,24 @@ export const dropdownContentStyles = css`
 	}
 
 	:host([theme="dark"]) {
+		--d2l-dropdown-above-animation-name: d2l-dropdown-above-animation-dark;
+		--d2l-dropdown-animation-name: d2l-dropdown-animation-dark;
 		--d2l-dropdown-background-color: #292a2b; /* ferrite @ 50% */
 		--d2l-dropdown-border-color: var(--d2l-color-ferrite);
 		--d2l-dropdown-foreground-color: var(--d2l-color-mica);
 		--d2l-dropdown-shadow-color: rgba(0, 0, 0, 1);
+		opacity: 0.9;
```
dbatiste, line 30: Are we sure we want to use `opacity` for this?  While it is convenient that is also makes the menu semi-transparent (which currently has a solid background color applied), it also makes the text, and everything else in the dropdown semi-transparent.

dlockhart: Yeah I know. I tried really hard to just apply an `rgba()` with alpha to the existing menu background colours -- the issue is that all 3 of the dropdown container, the `<d2l-menu>` container AND the individual `<d2l-menu-item>`s have a background colour set (to white in non-dark mode, different on-hover for the items). I can't have all 3 be 90% opaque... otherwise you end up with much less opacity than intended -- and different opacity depending on whether a menu is contained in a dropdown or not.

To get around that, we'd need to remove the background colour from dropdowns when a menu is contained inside (which it doesn't really know) and the menu container as well, which has implications for custom menu items and the header/footer areas. This solution was easier. :)

test/styles.css
dlockhart: Good call. I'll have follow-up PRs for dark mode button-icons so will include it with that.

components/dropdown/dropdown-content-styles.js
dbatiste: I see - all makes sense.  It seems like even if the nested component tries to set `opacity: 1`, it doesn't really seem to override the ancestor's opacity... it just means `opacity: 1` in the context of the `opacity: 0.9`.  The transparency we are using is pretty subtle, so maybe it's ok?... but if a higher transparency were used we might not be able to use this approach.

dbatiste (approved):

components/dropdown/dropdown-content-styles.js
dlockhart: Yeah it's very subtle. The other drawback to this approach is that it's only being applied to dropdown menus -- menus on their own (which are rare) don't get it. Fine for media player, but if we were doing a larger "dark mode" thing it may not be.

dbatiste: Good point.  The only other spot I know menu is used outside of dropdowns is the mobile fly-out.  Not sure if we'd want transparency applied there in such a case or not.  🤷‍♂️

PR #914, opened 2020-10-28
fix: translucent disabled button-icon hover styles

dlockhart:
While working on dark mode button-icons, I noticed that the translucent ones were incorrectly showing the hover style while disabled (and hovered). Coincidentally, the visual diff tests for this scenario weren't set up correctly and it was capturing the wrong element.

This fixes both of those things, and introduces the diagonal "dark mode" background to make the translucent visual diff better.

Screen Shot 2020-10-28 at 3 07 32 PM


components/button/test/button-icon.visual-diff.html
```diff
@@ -13,21 +13,11 @@
...
-			position: relative;
 			width: 200px;
 			height: 80px;
 		}
-		.translucent-container > img {
-			border-radius: 6px;
-			height: 100%;
-			position: absolute;
-			width: 100%;
-		}
```
dlockhart, line 26: Couldn't figure out what this was trying to do -- removed.

dbatiste: 🎉 .  I think that was left-over from when the visual-diff fixtures was set up like the demo, to include an actual image in the background... to look all fancy and such.  It was removed because the async nature of the image load make the test flakey, slower, and basically added little value for the test.

dbatiste (approved):

dlockhart:
Re-generating the goldens based on an expected visual-diff run:
https://s3.ca-central-1.amazonaws.com/visualdiff.gaudi.d2l/screenshots/BrightspaceUI/core/button-icon/2020-10-28-20.10.58.799-44d8f1dfa6b2eba9c123d6a5bfb7b2f50857d4aa-report.html

PR #915, opened 2020-10-28
HtmlBlock component for rendering user-authored HTML

dbatiste:
This component renders user-authored HTML, and lazy loads MathJax if it finds math in the content. Eventually we will probably move the MathJax loader outside of this module in order to use it within the LMS (Content topics and Equation Editing). Styles are copied for:
* for headings, lists, paragraphs, pre-formatted text : [bsi](https://github.com/Brightspace/brightspace-integration/blob/master/sass/html-block.scss)
* for links : [core:links](https://github.com/BrightspaceUI/core/blob/master/components/link/link.scss)

The API is basically:
``` html
<d2l-html-block>
    <template>
        <!-- some html -->
    </template>
</d2l-html-block>
```

There is some added complexity to get MathJax typesetting math within shadowDOM. The adaptor required to make this work is taken from this [MathJax thread](https://github.com/mathjax/MathJax/issues/2195). There is some odd error related to `<!---->` nodes at the root, which causes some issue that I've inquired about, but for the moment it is working because this sol'n is blowing away LitElement's comment nodes (not ideal, but currently not an issue). This sol'n uses MathJax v3 in order to avoid rendering as SVG, which would be the only option for the adapter with v2. The configuration has been generated from our v2 config, however some additional review needs to be done with respect to a11y and the changes that come with v3.

**Typography:**

image

**Math (block):**

image

**Math (inline):**

image


components/html-block/README.md
```diff
@@ -0,0 +1,23 @@
...
+## d2l-html-block
+
+The `d2l-html-block` element is a web component for displaying user-authored HTML. It includes styles for headings, lists, anchors and other elements.  In addition, it provides integration with MathJax for rendering MathML.
+
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/html-block/html-block.js';
+</script>
+
+<d2l-html-block html="..."></d2l-html-block>
```
dlockhart, line 12: Thoughts about an attribute vs. using a nested `<template>`? I think we're already doing that in the LMS to be able to defer rendering... although the attribute also accomplishes that.

The ergonomics of a template seem nicer if you're just crafting a manual block for testing/demo (no HTML encoding needed), but in practice folks will likely just be binding to the attribute via Lit so it won't matter.

components/html-block/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `html` | String | HTML to be rendered |
+
+## Future Enhancements
+
+Looking for an enhancement not listed here? Create a GitHub issue!
```
dlockhart, line 23: You could list replace strings here since it's a potential future enhancement we're already aware of?

components/html-block/html-block.js
```diff
@@ -0,0 +1,271 @@
...
+
+	return new Promise(resolve => {
+		if (!document.head.querySelector('#d2l-mathjax')) {
+			const script = document.createElement('script');
+			script.id = 'd2l-mathjax';
+			script.async = 'async';
+			script.onload = resolve;
+			script.src = 'https://s.brightspace.com/lib/mathjax/3.1.2/mml-chtml.js';
+			document.head.appendChild(script);
+		}
```
dlockhart, line 119: If Mathjax is already loaded this Promise won't resolve... should you call `resolve()` in an `else` block?

components/html-block/html-block.js
```diff
@@ -0,0 +1,271 @@
...
+			}
+			:host([dir="rtl"]) ul,
+			:host([dir="rtl"]) ol {
+				padding-left: 0;
+				padding-right: 3em;
+			}
+		`;
+	}
+
+	render() {
```
dlockhart, line 238: Can be excluded?

components/html-block/html-block.js
```diff
@@ -0,0 +1,271 @@
...
+	async updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (!changedProperties.has('html')) return;
+
+		if (this._hasMath()) {
+			if (!mathJaxLoaded) {
+				mathJaxLoaded = loadMathJax();
+			}
+			await mathJaxLoaded;
```
dlockhart, line 250: If `loadMathjax()` just always resolved the Promise you could turn these 4 lines into: `await loadMathJax();`.

components/html-block/README.md
dbatiste: I was thinking about this too and I was sorta on the fence about it, which is the main reason I started this PR as a draft.  😏   It would indeed make demo snippets a little nicer.  

Are you thinking `d2l-html-block` would require the content be provide in a `template`?  If it's in a `template`, the component can clone it and stamp it into local shadowDOM within the scope of the styles (`h1, h2, etc`), whereas if the HTML is rendered to the lightDOM, it would be outside the scope of those selectors, making applying those styles more difficult to apply, since the consumer context may be any arbitrary web component's shadowDOM.  

Rendering the HTML in the shadowDOM has the opposite consequence though - the consumer can't target elements in the user-authored HTML - something that I think is done in one or two places in the LMS to override default paragraph margins. Hoping we can find another sol'n to that, and such a use case should not exist.

components/html-block/README.md
dbatiste: I've had some discussion with Eric about replace-strings and I think we're in agreement that all the replace strings should be handled server-side.  The HM API's can return both the replaced (for display) and unreplaced (for editing) versions of the HTML. So, consumers using this component would be feeding it HTML that has replacements already done.

dlockhart: Gotcha -- makes sense, similar to HTML filtering. 👍

components/html-block/README.md
dlockhart: Yeah I was thinking it would need to be in a `<template>`, which would allow it to be lazy loaded. Otherwise I think the browser is going to start downloading and executing scripts/images/videos right away before our web component even boots up.

components/html-block/html-block.js
dbatiste: That's a good point.  I think I might adjust this to be more like:
``` javascript
let mathJaxLoaded;
const loadMathJax = () => {
    ...
    mathJaxLoaded = new Promise(resolve => {
        // add the script and resolve
    });
    return mathJaxLoaded;
};
```
Then the consumer (currently just `d2l-html-block`) doesn't need to worry about making a conditional check, it can just:
```
await loadMathJax();
```
But then maybe `requireMathJax` would be a better name.

components/html-block/html-block.js
dbatiste: Yeah probably.  I will try.

components/html-block/html-block.js
dbatiste: Haha... yep.  Read this after I replied to your previous comment.  😄

components/html-block/html-block.js
```diff
@@ -0,0 +1,271 @@
...
+
+const loadMathJax = () => {
+
+	window.MathJax = {
+		options: {
+			menuOptions: {
+				settings: { zoom: 'None' }
+			}
+		},
+		chtml: { scale: 1 },
```
brissons, line 14: I think the default configuration for `chtml: { scale }` is 1. Does this need to be set explicitly?

dbatiste: I think I can remove this.  Thanks.  :)

components/html-block/README.md
```diff
@@ -0,0 +1,21 @@
+# HTML Block
+
```
dlockhart, line 2: Nit: I think typically for READMEs that only have one thing in them we don't duplicate headings like this.

components/html-block/README.md
```diff
@@ -0,0 +1,21 @@
...
+## d2l-html-block
+
+The `d2l-html-block` element is a web component for displaying user-authored HTML. It includes styles for headings, lists, anchors and other elements.  In addition, it provides integration with MathJax for rendering MathML.
+
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/html-block/html-block.js';
+</script>
+
+<d2l-html-block>
```
dlockhart, line 12: Not sure if it's worth providing a sentence about why the `<template>` is necessary and how things won't work without it? I could see people not really looking at this, going `<d2l-html-block>${myStuff}</d2l-html-block>` and being confused as to why it doesn't work.

Actually... what would happen there? Can we detect that and make sure it doesn't "accidentally" kind-of work?

components/html-block/html-block.js
```diff
@@ -0,0 +1,270 @@
...
+				//  Now do the usual startup now that the extensions are in place
+				//
+				window.MathJax.startup.defaultReady();
+			}
+		}
+	};
+
+	mathJaxLoaded = new Promise(resolve => {
+		const script = document.createElement('script');
+		script.async = 'async';
```
dlockhart, line 114: I think `async` is correct... it's going to not render-block while downloading, but then it will. `defer` on the other hand would still download asynchronously but wait until HTML parsing is complete before executing. I don't actually think it matters since most web components are going to upgrade _after_ HTML parsing anyway I think.

components/html-block/html-block.js
```diff
@@ -0,0 +1,270 @@
...
+				padding-left: 0;
+				padding-right: 3em;
+			}
+		`;
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		if (!this._renderContainer) {
```
dlockhart, line 232: Nit: I'd inverse this to reduce nesting

dlockhart (approved):

components/html-block/README.md
dbatiste: Thanks.  I will fix that up next time I am in here making changes.

components/html-block/README.md
dbatiste: Yeah that's a good point.  I will update and do some tests around that scenario.

components/html-block/html-block.js
dbatiste: Yeah, I was basically following the config guidance from MathJax here - they were recommending `async` unless the config and the library were both being requested, in which case they recommend `defer` since `async` doesn't guarantee order, and the config is required to come before the library in order for initialization to be correct.  That all being said, I think we can switch it to `defer` if it matters because our config is within this module.  In this particular case it's being requested pretty late, only after we've detected `math` in the DOM.

PR #919, opened 2020-10-28
Focus visible

dlockhart:
This pulls together the `:focus-visible` polyfill into a mixin so that we can easily use it in our components. 🎉 

It also uses the mixin in our button-* components, which design would like before we implement dark mode buttons. Some other candidates we can look at will be tabs, cards and switches.


components/button/button-styles.js
```diff
@@ -21,7 +21,10 @@ export const buttonStyles = css`
 		white-space: nowrap;
 		width: auto;
 	}
-	button:focus {
+	/* Outline on focus for elements not using FocusVisiblePolyfillMixin
+	 * or when focus-visible for those who are */
+	:host(:not([data-js-focus-visible])) button:focus,
+	button.focus-visible {
```
dlockhart, line 27: Normally you'd just need `button.focus-visible { /* styles */ }`, but in this case a few things outside of core are using `buttonStyles` and aren't currently using the `FocusVisiblePolyfillMixin` -- with the first line here we're still applying the outline on normal focus if the host doesn't have a data attribute that the polyfill adds.

components/button/test/button-icon.visual-diff.js
```diff
@@ -21,22 +21,25 @@ describe('d2l-button-icon', () => {
...
 	].forEach((entry) => {
 		describe(entry.category, () => {
 			entry.tests.forEach((name) => {
 				it(name, async function() {
 					const selector = `#${entry.category}`;
 
 					if (name === 'hover') await page.hover(selector);
-					else if (name === 'focus') await page.$eval(selector, (elem) => elem.focus());
+					else if (name === 'focus') await page.$eval(selector, (elem) => { elem.focus(); elem.shadowRoot.querySelector('button').classList.add('focus-visible'); });
+					else if (name === 'click') await page.$eval(selector, (elem) => elem.focus());
```
dlockhart, line 36: A programatic `focus()` doesn't get the `:focus-visible` styles unless it was triggered by a keyboard event (like our menus will), so we need to manually add (and remove) the `focus-visible` class to get the "true" focus styles.

components/button/button-icon.js
```diff
@@ -84,7 +85,7 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 				:host([active]) button {
 					background-color: var(--d2l-color-gypsum);
 				}
-				button:focus {
+				button.focus-visible {
```
dlockhart, line 88: This is failing our CSS linting with:
> Expected class selector ".focus-visible" to match specified pattern   selector-class-pattern

So I guess I'll need to add an exception -- probably best to do it in the global stylelint-config eh @dbatiste?

dlockhart: Created a PR to add it:
https://github.com/BrightspaceUI/stylelint-config/pull/1

dbatiste: Yeah, the alternative would be that we disable the rule with inline comments, or remove that rule entirely... but adding it seems good!

dbatiste (approved):

helpers/focus.js
```diff
@@ -13,6 +13,18 @@ const focusableElements = {
 	textarea: true
 };
 
+export function forceFocusVisible(elem) {
+	if (!isFocusable(elem)) {
```
dlockhart, line 17: This handles the case where the element is something like a `d2l-button` where on its own it isn't focusable (but has a `focus()` method), so we need to find the real `button` element inside.

dlockhart (commented):
I made some changes to add a `forceFocusVisible()` helper and used it from our dialog code. This helper also simplifies the visual diff tests which is nice.

Issue #923, opened 2020-10-30
New dependency 'focus-visible' does not meet license requirements


devpow112:
Noticed when running the D2L license checker it fails on a dependency of this package.
```
"focus-visible@5.2.0" doesn't meet license requirements (W3C)
```

dbatiste:
Ref: https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document

The "...without fee or royalty..." is perhaps the concerning part - not sure if that's the reason this license was excluded from our list though.

devpow112:
The license checker is also not part of the CI for this repo. Feels like it would be a good add to catch stuff like this during development. I can work on adding it.

devpow112:
Adding the license checker there are a number of other licenses from dev dependencies that do not meet requirements either 😞
```
>>> Package "axe-core@4.0.2" doesn't meet license requirements (MPL-2.0).
    Repository: https://github.com/dequelabs/axe-core
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\axe-core

>>> Package "bin-links@1.1.8" doesn't meet license requirements (Artistic-2.0).
    Repository: https://github.com/npm/bin-links
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\npm\node_modules\bin-links

>>> Package "focus-visible@5.2.0" doesn't meet license requirements (W3C).
    Repository: https://github.com/WICG/focus-visible
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\focus-visible

>>> Package "gentle-fs@2.3.1" doesn't meet license requirements (Artistic-2.0).
    Repository: https://github.com/npm/gentle-fs
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\npm\node_modules\gentle-fs

>>> Package "npm-lifecycle@3.1.5" doesn't meet license requirements (Artistic-2.0).
    Repository: https://github.com/npm/lifecycle
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\npm\node_modules\npm-lifecycle

>>> Package "npm@6.14.8" doesn't meet license requirements (Artistic-2.0).
    Repository: https://github.com/npm/cli
    Path: C:\D2L\Development\github\BrightspaceUI\core\node_modules\npm

>>> Package "qrcode-terminal@0.12.0" doesn't meet license requirements (Apache 2.0).
    Repository: https://github.com/gtanner/qrcode-terminal
[...]

PR #924, opened 2020-10-30
Add license checker

devpow112:
This adds the d2l license checker to CI. Adds a number of manual overrides for exemptions/corrections. See #923.


dlockhart (approved):
Thanks for adding this in!

PR #927, opened 2020-10-30
Dark button icon

dlockhart:
This introduces a "dark" theme for `d2l-button-icon` and tweaks the menu dark styles to align with some design changes.


components/button/button-icon.js
```diff
@@ -51,9 +54,21 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
 				:host([hidden]) {
 					display: none;
 				}
+				:host([translucent]) {
```
dlockhart, line 57: I refactored "translucent" button-icon styles while I was in here, as it could take advantage of the new CSS variables.

components/button/button-icon.js
```diff
@@ -95,31 +110,9 @@ class ButtonIcon extends ButtonMixin(VisibleOnAncestorMixin(RtlMixin(LitElement)
...
 					transition-property: background-color, box-shadow;
 				}
-				:host([translucent]) button,
-				:host([translucent]) button[disabled]:hover,
-				:host([translucent]) button[disabled]:focus,
-				:host([active][translucent]) button[disabled] {
-					background-color: rgba(0, 0, 0, 0.5);
-				}
-				:host([translucent][visible-on-ancestor]) button {
-					transition-duration: 0.4s, 0.4s;
```
dlockhart, line 109: Hmm... this rule got lost. I'm not sure why the duration is longer in this one case (since we removed almost all of our transitions anyway), but I'll restore it.

dlockhart: Fixed.

AllanKerr (approved):
Approving so you can merge once the new visual diff test cases get added to `button-icon.visual-diff.js`

PR #928, opened 2020-10-31
html-block improvements

dbatiste:
This PR turned out to be slightly more than I'd originally intended, but it seems reasonable to roll these improvements together.

* updated readme to include better explanation on how to use (based on feedback)
* updated readme to include example of using within Lit context (use of `unsafeHTML` directive)
* updated to use mutation observer for detecting template content changes (a use-case i initially avoided but i think is typical/expected)
* updated to hide non-templated content (based on feedback)
* updated visual-diff tests to include no-template case
* updated visual-diff tests to include template content mutation case
* updated demo to include example wrapping and updating the template content

I thought it was worthwhile including this because it's useful for both manual unit testing, and also showing how this is expected to be used within other Lit components.
image


components/html-block/html-block.js
```diff
@@ -226,42 +229,54 @@ class HtmlBlock extends LitElement {
...
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		if (!this._renderContainer) {
+		if (this._renderContainer) return;
 
-			this.shadowRoot.innerHTML = '<div class="d2l-html-block-rendered"></div><slot></slot>';
+		this.shadowRoot.innerHTML = '<div class="d2l-html-block-rendered"></div><slot></slot>';
 
-			this.shadowRoot.querySelector('slot').addEventListener('slotchange', async e => {
```
dbatiste, line 236: This is likely the most interesting part of these changes and probably easier viewed split rather than unified.  The stamping logic has just been factored out so that it can be called for the two cases : 1) when the slotted template element is initially added or replaced; and 2) when the content of the template changes.

Case (2) is expected to be fairly common because when Lit consumers render this component, they will not be replacing their initially rendered `template` element, but rather the contents of it when they want to update.  This is because Lit will not recreate that `template` element.  To accommodate this scenario, a MutationObserver is wired up for the `template.content` to re-stamp the content.

components/html-block/demo/html-block.html
```diff
@@ -261,6 +261,168 @@ <h2>HTML Block (inline math)</h2>
...
 
+			<h2>HTML Block (wrapped)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-some-component></d2l-some-component>
+				</template>
+			</d2l-demo-snippet>
+
+			<d2l-code-view language="javascript">
```
dbatiste, line 272: I don't love that this is basically a copy of the actual demo code, but this is due to some limitations of the demo-snippet.  On the plus side, it does allow me to construct the snippet in a summarized fashion, excluding the less interesting bit of the code.

components/html-block/test/html-block.visual-diff.js
```diff
@@ -21,10 +21,12 @@ describe('d2l-html-block', () => {
...
 		{ name: 'empty', selector: '#empty' },
 		{ name: 'empty-template', selector: '#empty-template' },
+		{ name: 'no-template', selector: '#no-template' },
 		{ name: 'typography', selector: '#typography' },
-		{ name: 'update', selector: '#update', action: selector => page.$eval(selector, elem => elem.innerHTML = '<template>after update</template>') },
-		{ name: 'math (block)', selector: '#math-block' },
-		{ name: 'math (inline)', selector: '#math-inline' }
+		{ name: 'update-template', selector: '#update-template', action: selector => page.$eval(selector, elem => elem.innerHTML = '<template>after update</template>') },
+		{ name: 'update-content', selector: '#update-content', action: selector => page.$eval(selector, elem => elem.querySelector('template').content.textContent = 'after update') },
+		//{ name: 'math (block)', selector: '#math-block' },
```
dbatiste, line 28: Excluding these tests until I can figure out why they are flakey.  The image height is varying sometimes from 2-10px for these images.  Might be a timing issue.

components/html-block/README.md
```diff
@@ -1,9 +1,11 @@
 # HTML Block
 
-## d2l-html-block
-
 The `d2l-html-block` element is a web component for displaying user-authored HTML. It includes styles for headings, lists, anchors and other elements.  In addition, it provides integration with MathJax for rendering MathML.
 
+Place the user-authored HTML within a `template` and the `d2l-html-block` will stamp the content into its local DOM where styles are be applied, and math typeset.
```
dlockhart, line 5:
```suggestion
Place the user-authored HTML within a `template` and the `d2l-html-block` will stamp the content into its local DOM where styles will be applied, and math typeset.
```

dlockhart (approved):

PR #930, opened 2020-11-02
fix: extra check to make sure polyfill is really there

dlockhart:
I noticed the following JS error in Kibana from Spark since this was introduced:

> TypeError: window.applyFocusVisiblePolyfill is not a function
    at shadowRoot.window.applyFocusVisiblePolyfill.window.addEventListener.once (https://s.brightspace.com/lib/bsi/20.20.12-83/unbundled/button-styles.js:1:310)

So somehow, despite the event [not being fired until after the global is introduced](https://github.com/WICG/focus-visible/blob/master/src/focus-visible.js#L297), it's still not there when this code runs. Very confusing, but 🤷

### DIFF
--- a/mixins/focus-visible-polyfill-mixin.js
+++ b/mixins/focus-visible-polyfill-mixin.js
@@ -14,7 +14,10 @@ export const FocusVisiblePolyfillMixin = superclass => class extends superclass 
 		}
 
 		window.addEventListener('focus-visible-polyfill-ready', () => {
-			window.applyFocusVisiblePolyfill(this.shadowRoot);
+			// somehow it's occasionally still not available
+			if (this.shadowRoot && window.applyFocusVisiblePolyfill) {
+				window.applyFocusVisiblePolyfill(this.shadowRoot);
+			}
 		}, { once: true });
 
 		if (!polyfillLoading) {


mixins/focus-visible-polyfill-mixin.js
```diff
@@ -14,7 +14,10 @@ export const FocusVisiblePolyfillMixin = superclass => class extends superclass
 		}
 
 		window.addEventListener('focus-visible-polyfill-ready', () => {
-			window.applyFocusVisiblePolyfill(this.shadowRoot);
```
dbatiste, line 17: No harm in doing this check, but it certainly seems strange to need this, since the function is attached to the window before the event is dispatched. (https://github.com/WICG/focus-visible/blob/master/src/focus-visible.js#L283)

dbatiste (approved):

mixins/focus-visible-polyfill-mixin.js
dlockhart: Agreed. After this goes in, I'm going to monitor the JS errors closely and make sure it actually does solve the issue. All I can think of is that it's some kind of timing issue... maybe as the page is unloading the event is firing. I have no idea.

dlockhart: Actually... looking at when the errors occurred, I think this was some kind of automation that triggered it. It happened several hundred times all at the exact same second the morning of Nov 2 and then the same thing Oct 30.

PR #932, opened 2020-11-04
Add icon to d2l-alert

Jevan5:
While developing the `@brightspace-ui-labs/media-player` component,
we wanted to display an alert to the user. Some documentation
(http://design.d2l/components/alerts/) shows that `d2l-alert`'s
can support images. Upon slotting an ``, I realized that it
was moving things around within the `d2l-alert` component.

Instead of adding support for images, which may come in all shapes
and sizes, I decided to add support for the `d2l-icon`'s that we all know and love.

- With icon
Screen Shot 2020-11-04 at 7 13 02 AM

- With icon and action
Screen Shot 2020-11-04 at 7 13 31 AM

- With icon, action, and no padding
Screen Shot 2020-11-04 at 7 13 50 AM


components/alert/alert.js
```diff
@@ -102,19 +108,24 @@ class Alert extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+				padding-left: 0.5rem;
+			}
 
 			:host([dir="rtl"]) .d2l-alert-text-with-actions {
 				padding-left: 0.9rem;
 				padding-right: 1.5rem;
 			}
-			:host([no-padding]) .d2l-alert-text,
-			:host([no-padding]) .d2l-alert-text-with-actions {
+			:host([no-padding]) .d2l-alert-text {
```
Jevan5, line 119: With `no-padding`, it is targeting `.d2l-alert-text` _and_ `.d2l-alert-text-with-actions`. If an element belongs to the class `.d2l-alert-text-with-actions`, it also belongs to the class `.d2l-alert-text`:

```
const alertTextClasses = {
    'd2l-alert-text: true,
    'd2l-alert-text-with-actions': <conditional>
};
```

No need for the second half of the query selector.

components/alert/alert.js
```diff
@@ -102,19 +108,24 @@ class Alert extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			.d2l-alert-text-with-actions {
 				padding-right: 0.9rem;
 			}
+			.d2l-alert-text-with-icon {
+				padding-left: 0.5rem;
```
Jevan5, line 112: Reduced padding since the icon has some padding between the highlight bar already.

Jevan5:
Not a very common pattern, and people like custom layouts for images in their alerts.

PR #933, opened 2020-11-04
Close drop-dropdown menu on link select

Jevan5:
Selecting a menu item did not previously close menus,
because they were expected to navigate to another page.
If the target is a new tab, the original page would remain
intact, with the drop-down menu still open.

This change will close drop-down menus whenever a menu
item link is selected. Also, passes `download` attribute through to anchor.

example-video

### DIFF
--- a/components/dropdown/test/dropdown-menu.test.js
+++ b/components/dropdown/test/dropdown-menu.test.js
@@ -2,6 +2,7 @@ import '../dropdown.js';
 import '../dropdown-menu.js';
 import '../../menu/menu.js';
 import '../../menu/menu-item.js';
+import '../../menu/menu-item-link.js';
 import '../../menu/menu-item-radio';
 import { expect, fixture, html, oneEvent } from '@open-wc/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
@@ -15,6 +16,19 @@ const itemFixture = html`
 					<d2l-menu-item text="Introduction" id="first-item"></d2l-menu-item>
 					<d2l-menu-item text="Searching for the Heavens "></d2l-menu-item>
 					<d2l-menu-item text="The Universe"></d2l-menu-item>
+				</d2l-menu>
+			</d2l-dropdown-menu>
+		</d2l-dropdown>
+	</div>
+`;
+
+const linkFixture = html`
+	<div>
+		<d2l-dropdown>
+			<button class="d2l-dropdown-opener">Open it!</button>
+			<d2l-dropdown-menu id="dropdown">
+				<d2l-menu label="Helpful Links">
+					<d2l-menu-item-link text="menu.js (View)" href="../../menu/menu.js" id="first-item"></d2l-menu-item-link>
 				</d2l-menu>
 			</d2l-dropdown-menu>
 		</d2l-dropdown>
@@ -42,6 +56,7 @@ describe('d2l-dropdown-menu', () => {
 
 		[
 			{ name: 'should close when menu item is selected', fixture: itemFixture },
+			{ name: 'should close when menu link item is selected', fixture: linkFixture },
 			{ name: 'should close when menu radio item is changed', fixture: radioFixture },
 		].forEach((testCase) => {
 			it(testCase.name, async() => {
--- a/components/dropdown/test/dropdown-menu.test.js
+++ b/components/dropdown/test/dropdown-menu.test.js
@@ -28,7 +28,7 @@ const linkFixture = html`
 			<button class="d2l-dropdown-opener">Open it!</button>
 			<d2l-dropdown-menu id="dropdown">
 				<d2l-menu label="Helpful Links">
-					<d2l-menu-item-link text="menu.js (View)" href="../../menu/menu.js" id="first-item"></d2l-menu-item-link>
+					<d2l-menu-item-link text="menu.js (View)" href="#" id="first-item"></d2l-menu-item-link>
 				</d2l-menu>
 			</d2l-dropdown-menu>
 		</d2l-dropdown>
--- a/components/menu/demo/menu.html
+++ b/components/menu/demo/menu.html
@@ -158,6 +158,9 @@
 						<d2l-menu-item-link text="Dwarf Planets (link)" href="https://en.wikipedia.org/wiki/Dwarf_planet">
 							<span slot="supporting">39+ AU</span>
 						</d2l-menu-item-link>
+						<d2l-menu-item-link text="Dwarf Planets (link to download)" href="../menu.js" target="_blank" download>
+							<span slot="supporting">Something</span>
+						</d2l-menu-item-link>
 					</d2l-menu>
 				</template>
 			</d2l-demo-snippet>
--- a/components/menu/menu-item-link.js
+++ b/components/menu/menu-item-link.js
@@ -14,6 +14,10 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 	static get properties() {
 		return {
 			/**
+			 * Name to use for download
+			 */
+			download: { type: String },
+			/**
 			 * The url the menu item link navigates to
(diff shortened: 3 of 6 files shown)


components/dropdown/test/dropdown-menu.test.js
```diff
@@ -21,6 +22,19 @@ const itemFixture = html`
 	</div>
 `;
 
+const linkFixture = html`
+	<div>
+		<d2l-dropdown>
+			<button class="d2l-dropdown-opener">Open it!</button>
+			<d2l-dropdown-menu id="dropdown">
+				<d2l-menu label="Helpful Links">
+					<d2l-menu-item-link text="menu.js (View)" href="../../menu/menu.js" id="first-item"></d2l-menu-item-link>
```
Jevan5, line 31: Maybe the following test failure is related to this line opening content on new page:
```
Edge 18.18363 (Windows 10) ERROR
  Some of your tests did a full page reload!
04 11 2020 13:11:35.082:ERROR [reporter.browserlabs]: ✖ Test Errored
```

Thoughts?

components/menu/menu-item-link.js
```diff
@@ -13,6 +13,10 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 
 	static get properties() {
 		return {
+			/**
+			 * Name to use for download
+			 */
+			download: { type: String },
```
dlockhart, line 19: Reading the spec... haha this is a weird attribute eh... it's a string and becomes the suggested filename if set to a string, but otherwise it's a boolean attribute and uses other mechanisms to determine the filename. Weird!

For the description in the comment, I'd suggest something like:
```suggestion
			 * Prompts the user to save the linked URL instead of navigating to it.
			 * Must be to a resource on the same origin.
			 * Can be used with or without a value, when set the value becomes the filename.
			 */
			download: { type: String },
```

components/dropdown/test/dropdown-menu.test.js
dbatiste: Yeah I think the `click()` performed in the test will cause a navigation to happen for this test. For the purpose of this test, I think we should be able to use either `href="#"` (which scrolls to the top of the page) or `href="javascript:void(0)"` (which does nothing).

components/menu/menu-item-link.js
```diff
@@ -57,7 +61,7 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 	render() {
 		const rel = this.target ? 'noreferrer noopener' : undefined;
 		return html`
-			<a href="${ifDefined(this.href)}" rel="${ifDefined(rel)}" target="${ifDefined(this.target)}" tabindex="-1">
+			<a download="${ifDefined(this.download)}" href="${ifDefined(this.href)}" rel="${ifDefined(rel)}" target="${ifDefined(this.target)}" tabindex="-1">
```
dbatiste, line 64: Would our `download` attribute be required?  According to the specs:
> Prompts the user to save the linked URL instead of navigating to it. Can be used with or without a value:
>
> Without a value, the browser will suggest a filename/extension, generated from various sources:
The Content-Disposition HTTP header
The final segment in the URL path
The media type (from the (Content-Type header, the start of a data: URL, or Blob.type for a blob: URL)
Defining a value suggests it as the filename. / and \ characters are converted to underscores (_). Filesystems may forbid other characters in filenames, so browsers will adjust the suggested name if necessary.

This implies to me that both of these scenarios are valid:
``` javascript
<a download href="...">

<a download="some name" href="...">
```
But with the current rendering, if the consumer does not provide a value for `download`, the user will not be prompted to save instead of navigating.  Is that true?

dlockhart: That's a good question... worth testing, but I think just setting a boolean attribute will still bind to the string (not sure what value it'll end up with, maybe empty?).

components/menu/demo/menu.html
```diff
@@ -158,6 +158,9 @@ <h2>Supporting Slot</h2>
 						<d2l-menu-item-link text="Dwarf Planets (link)" href="https://en.wikipedia.org/wiki/Dwarf_planet">
 							<span slot="supporting">39+ AU</span>
 						</d2l-menu-item-link>
+						<d2l-menu-item-link text="Dwarf Planets (link to download)" href="../menu.js" target="_blank" download>
```
dbatiste, line 161: This seems a little strange - the `download` attribute is being used here as if the property is defined as a `Boolean`, but the component defines the property as a `String`.  This relates to my other common about the type and whether the value is required or not.  Curious what gets passed down in this case.

components/menu/menu-item-link.js
Jevan5: The `download` attribute would not be required.

The following 3 scenarios are valid:
```
// Navigates. Rendered when no 'download' attribute is passed
<a href="...">

// Downloads, and generates file name in the browser. Rendered when 'download' attribute is passed with no value
<a download href="...">

// Downloads, with custom file name. Rendered when 'download' attribute is passed with value
<a download="custom-name" href="...">
```

`ifDefined()` will render the attribute even if it does not have a value, so the `download` attribute is set without a value in that case.

components/menu/demo/menu.html
dlockhart: I think that's OK based [on the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) of how it's supposed to work.

components/dropdown/test/dropdown-menu.test.js
dlockhart: Agree with Dave here -- those options should work.

components/menu/menu-item-link.js
Jevan5: Should I add demo examples and tests with `download` and `download="value"`?

Jevan5: > That's a good question... worth testing, but I think just setting a boolean attribute will still bind to the string (not sure what value it'll end up with, maybe empty?).

`<d2l-menu-item-link download>` will gives the MenuItemLink object a `download` value of `''`, empty-string.

components/menu/demo/menu.html
Jevan5: The attribute is a string. Lit will initialize strings as empty-string if no value is passed

components/menu/menu-item-link.js
dbatiste: Thanks, yeah that's what I was wondering.  That might be ok.  I'm not sure if the empty string matters or not.  If it does, it could be avoided using something roughly along these lines:

```javascript
const hasDownloadName = this.download && this.download.length > 0;

return hasDownloadName 
    ? html`<a download="${this.download}" ...></a>` 
    : html`<a ?download="${!!this.download}" ...></a>`; 
```

components/menu/menu-item-link.js
dbatiste: Along these lines - will also want to update the [readme for menu](https://github.com/BrightspaceUI/core/tree/master/components/menu#d2l-menu-item-link).

components/menu/menu-item-link.js
Jevan5: If you write `<a download=""></a>` to the DOM, it will render as `<a download></a>`. Attributes with empty-string values have their values removed.

Jevan5:
This test is failing in CI:
```
  LocalizeMixin
    ✔ should only render once (static)
    ✖ should only render once (async)
```

But passing when I run `npm test` locally:
```
  LocalizeMixin
    ✔ should only render once (static)
    ✔ should only render once (async)
```

dlockhart:
That's a flakey test that we're looking into -- just re-run for now.

dbatiste (approved):
Look' good!

dlockhart (approved):
This looks great to me -- thanks Joshua for adding!

PR #934, opened 2020-11-05
fix: making form elements block

dlockhart:
The native `<form>` element is block by default, so ours should be too.

Fixed some file locations in the READMEs while I was in here.

### DIFF
--- a/mixins/test/localize-mixin.test.js
+++ b/mixins/test/localize-mixin.test.js
@@ -212,7 +212,7 @@ describe('LocalizeMixin', () => {
 				setTimeout(() => {
 					expect(renderCount).to.equal(1);
 					done();
-				}, 300);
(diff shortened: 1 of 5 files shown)


dbatiste (approved):

AllanKerr (approved):

mixins/test/localize-mixin.test.js
```diff
@@ -212,7 +212,7 @@ describe('LocalizeMixin', () => {
 				setTimeout(() => {
 					expect(renderCount).to.equal(1);
 					done();
-				}, 300);
+				}, 500); // larger timeout for legacy-Edge
```
dlockhart, line 215: @AllanKerr trying this as maybe legacy-Edge has just gotten suuuuuper slow

PR #935, opened 2020-11-05
Primary Seconday Style Tweaks

AllanKerr:
# Changes
- Add more subtle desktop focus and hover styles
- Change mobile breakpoint to 767px since it was 1px too large
- Add secondary panel shadow for mobile


templates/primary-secondary/primary-secondary.js
```diff
@@ -912,6 +933,8 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(LitEle
 						</div>
 					</div>
 					<div style=${styleMap(secondaryPanelStyles)} class="d2l-template-primary-secondary-secondary-container" @transitionend=${this._onTransitionEnd}>
+						<div class="d2l-template-primary-secondary-divider-shadow">
```
AllanKerr, line 936: This was a bit of a pain. I had to use a new div because:
1. If the shadow is on the divider its displayed on both sides.
2. If the shadow is on the container, it disappears when collapsed because the container has a height of `0px`.

dlockhart (approved):

PR #937, opened 2020-11-09
GitHub Actions

dlockhart:
Putting this up as a draft for visibility while Stacey and I work on some of the missing pieces.

TODO:
- [x] Convert the translation verification pre-deploy step into an action
- [x] Deploy via semantic-release
- [x] Add `actions/cache@v2` to `Brightspace/third-party-actions` and use that
- [x] Add license-checker step (part of CI, or maybe a separate workflow?)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+    branches:
+      - master
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - master
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
```
dlockhart, line 14: I'm not sure how much of the above `on` stuff we actually need... maybe it's fine to just run this on all branches and just limit the jobs down below to only run under more specific circumstances -- for example only deploy from `master` or `a.b.c` release branches.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+
+jobs:
+  lint:
+    name: Lint
+    timeout-minutes: 2
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
+      - uses: actions/cache@v2
```
dlockhart, line 24: This will need to be added to `Brightspace/third-party-actions` if caching is something we want. @j3parker what do you think?

.gitignore
```diff
@@ -1,6 +1,5 @@
 generated/
 node_modules/
-package-lock.json
```
dlockhart, line 3: A caveat to being able to cache `node_modules` and have fast builds is checking in our lock file. I think it would be easy to set up an automated workflow to keep this up to date for us automatically.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+          key: npm-${{ hashFiles('**/package-lock.json') }}
+      - name: Install dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: npm ci
+      - name: Build
+        run: npm run build
+      - name: BrowserStack
+        env:
+          BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
+          BROWSERSTACK_ACCESS_KEY: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
```
dlockhart, line 80: These are secrets local to this repo for now -- once we're all-in on BS (or not) we could move them out to the org.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: npm run test:bs
+  sauce:
+    name: Unit Tests (SauceLabs)
+    if: ${{ github.event_name == 'pull_request' }}
+    timeout-minutes: 10
+    continue-on-error: true
```
dlockhart, line 86: This means this step is allowed to fail, which is what we want while we're just comparing Sauce to BS.

components/inputs/test/input-time-range.test.js
```diff
@@ -103,7 +103,8 @@ describe('d2l-input-time-range', () => {
 			expect(elem.endValue).to.equal('00:30:00');
 		});
 
-		it('should update startValue as expected when set through property', async() => {
+		// timing out in legacy-Edge via GitHub Actions
+		it.skip('should update startValue as expected when set through property', async() => {
```
dlockhart, line 107: These tests are timing out in legacy-Edge on BrowserStack but only from GitHub Actions... super weird. :(

.github/workflows/ci.yml
svanherk: Yeah I was wondering if it'd be nice to divide into two workflows - one that runs when a PR is opened and added to, and then one were `master` or a version branch is committed to (i.e. merge).  Since we usually run a different set of tests for PR CI versus merge.

dlockhart: I like that... no need to lint on `master` since in theory nothing should have been able to merge into it without passing the stricter CI. Want me to split this out into two now or do you want to just add to this PR as you make progress with your stuff?

svanherk: Yeah no rush, I can split it out once I get deploy working... 🤞

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - '[0-9]+.x'
```
j3parker, line 7: `.x`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+      - '[0-9]+.[0-9]+.x'
+  pull_request:
+    types: [opened, reopened, synchronize]
```
j3parker, line 10: This is usually what you want... [so its the default :)](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request)

image

It's _not_ a big deal, but this can be simplified slightly and in the future you won't need to lookup the docs if you remember that the default is good:

```suggestion
```

.github/workflows/ci.yml
j3parker: You can add `if:` to steps and check the branch name. [Here's an example of that](https://github.com/Brightspace/d2l.dev/blob/master/.github/workflows/paste.d2l.dev.yml#L31).

Sometimes its nice to have one big workflow but disable some steps based on the scenario... but it can get overboard and one workflow might be right.

j3parker: P.S. for pull requests this is filtering on the source branch, not the target branch.... so you probably don't want `branches:`

.github/workflows/ci.yml
j3parker: Sounds good

.github/workflows/ci.yml
dlockhart: Our release branches are called `1.x` or `1.39.x`.

j3parker (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
+name: CI
+
+on: [push, pull_request]
```
j3parker, line 3: This is fine, but it will (afaik) run the job twice when you push to a PR: once for the push, once for the PR synchronized event.

A lot of people will do:

```yaml
on:
  push:
    branches:
      - master
      - etc.
  pull_request:
jobs:
  ...
```

It's not a big deal; up to you.

dlockhart: Ah gotcha thanks, will fix.

.travis.yml
```diff
@@ -24,39 +24,6 @@ addons:
...
-        if [ $TRAVIS_PULL_REQUEST != false ] && [ $TRAVIS_SECURE_ENV_VARS == true ]; then
-            echo "Pull request with secure environment variables, running Sauce tests...";
-            npm run build;
-            npm run test:sauce || true;
-        fi
-      name: "SauceLabs"
-      env:
-        - SAUCE_USERNAME: Gaudi1
-        # SAUCE_ACCESS_KEY
-        - secure: bwPUP8HCJ4x/VQHYT83fw28MF6fZqcCFYYLQIuYl5d8QXPuHNiON+36YSEkIix/2AjuQR8OtNqy7MPem013QUzEl9ayJ+P/Kto3hN84D6fNj4DXKUtDywDeAs6B0hv3OZZIUJ2ixqClyGDRoK9M51NzS3d9Nts8rEDTJY5CqnD7s+0mOUXfmr7ehStQfs4JdsGApdp5ByA7btC6t3vNAUSoGOlnfKGb4VP/cCXAS64Suy/8Q09fCCKr2BMswQwR6hwb52EJ50W9JbpXu/S0DwrwwM5ZGCG4+SSXqrMoaWXeoOjjQ+tdTx+Rj/cvYO0IE9vJvEMubFCzjs57d2r9IwoKmJg27Fh7YPlZk4YEhcge2vDO9AmlGKGDKeEYco4fXelAmmCcFjmrWXjH8Uc31n2wZj4yYg6AKU2M+GvMYNmh+aI1qGJ+urpQigYdf8IWpqryW19wiOzRm+N6JhdFGlfgnJMEv/OmdTwGOXKA1Qlmc0Bv8ZjWHa6bVMrwnueygoUwc7rWYQg0SjIaW8zhTWwy2JkBoCD1/TpKYTNOdL4VSaxpRSCQn4uAlJ2rNBIyv7FM0S7ycKMTVQGXIyaRo28vKUEGUDYGYWbu++5fBxRfaHWeE79ZFE823LyFQetTUmr7ms/i4pqkaM/m+q840s34lLwmupSoUoVBAGezVsk8=
```
dlockhart, line 59: For some reason this Travis config ends up running on NPM `7.0.8` which results in a bunch of visual diffs whereas [a fresh PR](https://github.com/BrightspaceUI/core/pull/939) runs on `6.14.8`. So confused...

.github/workflows/ci.yml
dbatiste: What does it cause to be cached?

dlockhart: This is configured to cache everything in `node_modules`, which saves roughly 2 minutes from each job in here (2 minutes times 5 jobs = 10 minutes), plus from the release workflow as well.

It will use a hash of `package-lock.json` to bust the cache, which means that we'd need to start using a lock file. But we can automate the lock file updates similar to how we do that in BSI, minimizing its impact on us. Bonus here would be that we'd have a lock file in release branches, which is nice to manage dependencies.

dlockhart: More information:
[Caching dependencies to speed up workflows](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows).

dlockhart:
This is now ready for review. We've decided due to the number of minor visual diff changes that @svanherk is going to add visual diff as a separate pull request after this merges -- so we'll temporarily be without visual diff.

package.json
```diff
@@ -52,25 +47,21 @@
...
     "eslint-plugin-html": "^6",
     "eslint-plugin-lit": "^1",
     "eslint-plugin-sort-class-members": "^1",
-    "karma-browserstack-launcher": "~1.4.0",
-    "karma-sauce-launcher": "^2",
     "lit-analyzer": "^1",
     "mocha": "^8",
     "node-sass": "^5",
     "puppeteer": "^5",
-    "semantic-release": "^17",
```
svanherk, line 73: We can remove the semantic-release config stuff below too

components/inputs/test/input-time-range.test.js
margaree: We had the `karma-browserstack-launcher` version locked at 1.4.0 before whereas it looks like it's not anymore - maybe that's somehow related?

dlockhart: Thanks, I'll give that a try. Although even if it resolves the issue I'm not sure if it's worth holding back on an old version just for legacy-Edge. We'll see.

margaree: Agreed

package.json
dlockhart: Yup thanks -- removed.

components/inputs/test/input-time-range.test.js
dlockhart: Still failing on 1.4.0. I think they [fixed the issues](https://github.com/karma-runner/karma-browserstack-launcher/pull/158/files) with post-1.4.0 timeouts, so this must be a different thing. What's weird is the error message:

> Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

But according to our config, our mocha timeout should be 10 seconds. 🤷

svanherk (approved):
Let's try it!

PR #938, opened 2020-11-09
Expose verify-translations CLI script as an export for other modules to use

grant-cleary:
For the latest work on the `htmleditor` component, we would ideally like to be able to rely on the same lang management as `core` does. Ideally this would also include the pre-publish scripts, so I would like to consider publishing this CLI script so other modules can consume this.


dlockhart:
Hmm... so we're right in the middle of migrating to GitHub Actions, and I think this kind of script would be a perfect thing to live in an action -- i.e. I want to delete/replace it soon, so publishing it as part of the NPM package isn't ideal.

grant-cleary:
> Hmm... so we're right in the middle of migrating to GitHub Actions, and I think this kind of script would be a perfect thing to live in an action -- i.e. I want to delete/replace it soon, so publishing it as part of the NPM package isn't ideal.

That sounds like a thing I could do. :)

dlockhart:
Awesome! I think @svanherk is working on a set of shared actions that our component repos could all pick and choose from, so she may have thoughts too.

grant-cleary:
> Awesome! I think @svanherk is working on a set of shared actions that our component repos could all pick and choose from, so she may have thoughts too.

Cool! I'll wait for Stacey to chime in before I close this out then.

svanherk:
Yeah this definitely seems like something we can add as an option to the deploy action.  I'm _just_ getting into it, so I don't have a timeframe... does "soon" work? 😛

grant-cleary:
> Yeah this definitely seems like something we can add as an option to the deploy action. I'm _just_ getting into it, so I don't have a timeframe... does "soon" work? 😛

Haha sure, "soon" is fine. If there's any way I can help out, do let me know, but I also don't want to step on your toes if you're not in a place where collaboration makes sense. 😄

grant-cleary:
Going to close this out now because either way it's not the direction we want to go. Thanks for your input, @dlockhart and @svanherk!

PR #939, opened 2020-11-09
re-run visual diff

dlockhart:
It looks like something in the config has changed such that a significant number of visual diffs need to be re-generated. Isolating this in an empty PR so it's obvious.


dlockhart:
Going to close this since it appears it's the NPM version that's causing the diff.

PR #940, opened 2020-11-10
Toggle dropdown opener on spacebar

Jevan5:
While testing `@brightspace-ui-labs/media-player`, we found that the `<d2l-button-icon>`'s were clickable with mouse, enter, or spacebar. The settings menu has a `<d2l-button-icon class="dropdown-opener">` to toggle it, but the toggling is only handled by mouse or enter.

This change would allow the dropdown menu to be toggled with spacebar as well.

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -113,7 +113,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__onKeyPress(e) {
-		if (![13, 32].includes(e.keyCode)) return;
+		if (e.keyCode !== 13 && e.keyCode !== 32) return;
 		if (this.noAutoOpen) return;
 		this.toggleOpen(true);
 	}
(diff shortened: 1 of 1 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -113,7 +113,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__onKeyPress(e) {
-		if (e.keyCode !== 13) return;
+		if (![13, 32].includes(e.keyCode)) return;
```
dbatiste, line 116: Is there a reason to write the check this way as opposed to the more conventional `if (e.keyCode !== 13 && e.keyCode !== 32)`?  I find the non-array based style easier to read, and it avoids allocating memory.  I haven't done a perf test - it probably requires 1000s of iterations to measure, but taken as a whole, the later is my preference.

Jevan5: Usually how I write a check to see if a value exists in some set of possible values. Makes more sense when the set is larger since the original value and `!==` does not have to be re-written for each item in the set.

That being said, seems like fewer characters is less important than convention and performance.

dbatiste: Cool.  Thanks. 😄   I do agree that I'd make that compromise at some point with a larger set.  In such a case, I'd consider factoring it out of the component so the allocation is only done once for all instances, and all key presses.

dbatiste (approved):

PR #942, opened 2020-11-10
US121245 Add bongo tier 1 image

cbelan-d2l:
The bongo images are used when creating a Bongo Tool Provider.

**Question:**
I have tier 2 and 3 from design. But we don't think these will be used. Should I add them anyways?
My thought was, no, but wanted to check if anyone has a different opinion

Related PR https://github.com/Brightspace/lms/pull/3278


components/icons/images/tier1/video-assignment.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg">
+  <g fill="#494C4E" fill-rule="nonzero">
```
dlockhart, line 2: Can you remove the fill rules? `<d2l-icon>` will add this colour back in, but if you specify it inside the SVG it can't be changed from the outside (e.g. on hover to celestine).

cbelan-d2l: Hmm... I see `<g fill="#494c4e"` in a whole bunch.

The difference is mine is using `fill-rule="nonzero"`
While the others have this `<g fill="#494c4e" fill-rule="evenodd">`

Is that what you're referring to?

dlockhart: Ah sorry nevermind, forgot we automatically remove `#494c4e` at runtime -- you're good!

dlockhart (approved):

PR #951, opened 2020-11-13
Add d2l-list-item-button.

dbatiste:
This PR refactors the rendering of the primary action to support alternate primary action rendering (i.e. button instead of anchor).  Currently, consumers are trying to wire-up for `click` events on standard list-items, however this is not keyboard accessible, nor does it provide any action styling.

* refactors ListItemMixin to support alternate primary action rendering
* new ListItemLinkMixin that contains the link rendering for primary action
* new ListItemButtonMixin that contains the button rendering for primary action
* new `d2l-list-item-button` component

The rendered button dispatches a `d2l-list-item-button-click` event allowing consumers to perform their action.

This change is made to be backwards compatible because there are two usages of `ListItemMixin` that depend on the `actionHref` property.  They will need to be modified to also extend `ListItemLinkMixin` before `actionHref` can be removed from `ListItemMixin` without breakage.  This removal will be done in a follow-up PR.

https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-basic.js?r=7d7905cd

https://search.d2l.dev/xref/GitHub/BrightspaceHypermediaComponents/activities/components/d2l-work-to-do/d2l-work-to-do-activity-list-item-detailed.js?r=7d7905cd

Once all looks good, I will update the readme and tests.


dlockhart (approved):

dlockhart:
I wonder if it makes sense to re-target the actual `click` event so that people could still wire up to the `@click` event? Would be less verbose that `d2l-list-item-blah-blah-click`.

m6jones (commented):
This looks awesome!!

Thanks Dave!

dbatiste:
> I wonder if it makes sense to re-target the actual `click` event so that people could still wire up to the `@click` event? Would be less verbose that `d2l-list-item-blah-blah-click`.

I thought about dispatching `click`, however they can click other things inside the list-item as well, which should not run their action code.  We'd need to stop propagation of those click events I think, which could be problematic if they are depending on them.

PR #953, opened 2020-11-13
Adding absolute positioning to hidden MathML

brissons:
MathJax creates (or passes on) some hidden `MathML` for screen readers. However, if the containing `MathML` is large enough (ex: contains `semantics` tags), it can push the width of its container (which might make some unnecessary horizontal scrollbars).

Adding a `position: absolute` to take it out of the visual flow of the document (since you can't see it anyway) so it doesn't add any width. Shouldn't affect screen reading either (tested with JAWS and NVDA w/MathPlayer just in case).

[Before](https://user-images.githubusercontent.com/52549862/99118385-0ea9fc80-25bd-11eb-80b2-906b85ad6a3b.png) and [after](https://user-images.githubusercontent.com/52549862/99118278-db676d80-25bc-11eb-99a7-0556bdff6b1a.png).

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -218,6 +218,9 @@ class HtmlBlock extends LitElement {
 				outline-width: 0;
 				text-decoration: underline;
 			}
+			mjx-assistive-mml math {
+				position: absolute;
+			}
 			:host([dir="rtl"]) {
 				text-align: right;
 			}


dbatiste:
Seems ok.  Any chance you have a snippet of MathML I can try?  The `mjx-assistive-mml` element is already being set the absolute (by MathJax), so it seems odd that the math element within would cause scrolling outside.  Also wondering if this is a bug with MathJax (that we should raise), or something to do with how we are using it.

brissons:
Here you go!
```
<math xmlns="http://www.w3.org/1998/Math/MathML" title="" display="inline">
   <semantics>
      <mstyle>
         <mi>x</mi>
         <mo>=</mo>
         <mfrac>
            <mrow>
               <mo>-</mo>
               <mi>b</mi>
               <mo>±</mo>
               <msqrt>
                  <msup>
                     <mi>b</mi>
                     <mn>2</mn>
                  </msup>
                  <mo>-</mo>
                  <mn>4</mn>
                  <mi>a</mi>
                  <mi>c</mi>
               </msqrt>
            </mrow>
            <mrow>
               <mn>2</mn>
               <mi>a</mi>
            </mrow>
         </mfrac>
      </mstyle>
      <annotation encoding="wiris">{"version":"1.1","math":"&lt;math xmlns="http://www.w3.org/1998/Math/MathML"&gt;&lt;mi&gt;x&lt;/mi&gt;&lt;mo&gt;=&lt;/mo&gt;&lt;mfrac&gt;&lt;mrow&gt;&lt;mo&gt;-&lt;/mo&gt;&lt;mi&gt;b&lt;/mi&gt;&lt;mo&gt;&amp;#xB1;&lt;/mo&gt;&lt;msqrt&gt;&lt;msup&gt;&lt;mi&gt;b&lt;/mi&gt;&lt;mn&gt;2&lt;/mn&gt;&lt;/msup&gt;&lt;mo&gt;-&lt;/mo&gt;&lt;mn&gt;4&lt;/mn&gt;&lt;mi&gt;a&lt;/mi&gt;&lt;mi&gt;c&lt;/mi&gt;&lt;/msqrt&gt;&lt;/mrow&gt;&lt;mrow&gt;&lt;mn&gt;2&lt;/mn&gt;&lt;mi&gt;a&lt;/mi&gt;&lt;/mrow&gt;&lt;/mfrac&gt;&lt;/math&gt;"}</annotation>
   </semantics>
</math>
```

I was playing around with the demo a bit, and it seems to occur when `display: inline`.

dbatiste:
Thanks.  Looks good.  I think the other way to handle this it to put `overflow: hidden` on the `div` that we render, just inside `mjx-body`, but that might have some other undesirable effects and this is more targeted.

dbatiste (approved):

PR #960, opened 2020-11-19
Only do the list-item tilt if the item is actually draggable.

dbatiste:
This change opts out of the tilt CSS if they item isn't draggable.

The reason for opting out (besides it being an unnecessary transform) is that the tilt increases the `scrollHeight` but not the `offsetHeight`, which is what dialogs base their height calcs off of, resulting in an unnecessary scrollbar in dialogs.  We _could_ also change dialogs to use `scrollHeight` instead but this would create extra undesirable whitespace whist avoiding the unnecessary scrollbar, and I would consider it a change to fragile code. In the end, I think opting out of this CSS is good either way.

Background: the tilt is applied this way so that images inside of list-items will be rendered in the tilted drag-image when a list-item is dragged.  This CSS is effectively setting up the drag image in advance. (hopefully that summarizes it correctly @m6jones).  From a pure technical perspective, I'd prefer not to use the tilt at all, and there is an open discussion with design for this... so for the time being it remains.

For ref... drag images are currently tilted like so:
image


dbatiste:
Haha... VD caught this legit thing I missed.
image

dlockhart (approved):

PR #961, opened 2020-11-20
Revert "Fix dropdown content overflowing page when aligned"

AllanKerr:
This was supposed to go to in a PR but I git pushed without realizing I was still on `master`. Reverting so it can go through a proper review.

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -485,12 +485,8 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			contentRect = contentRect ? contentRect : content.getBoundingClientRect();
 			const headerFooterHeight = header.getBoundingClientRect().height + footer.getBoundingClientRect().height;
 
-			const spaceRequired = {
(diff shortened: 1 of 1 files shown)


margaree (approved):

dlockhart (approved):

components/dropdown/dropdown-content-mixin.js
```diff
@@ -578,9 +579,9 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		}
 		const isRTL = this.getAttribute('dir') === 'rtl';
 		if ((this.align === 'start' && !isRTL) || (this.align === 'end' && isRTL)) {
-			constrained.left = Math.max(0, spaceRequired.width - (targetRect.width + spaceAround.right));
```
dbatiste, line 581: Do you know why this was being passed in and used in this calc?  i.e. what use-case it was meant for?

AllanKerr: This was supposed to be part of a new PR to fix [DE41401](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fdefect%2F457806818636) where start or end align doesn't honor page boundaries. I'm just reverting this then immediately putting up a PR to add it back since it got pushed to master accidentally bypassing review.

PR #962, opened 2020-11-20
Fix dropdown content overflowing page when aligned

AllanKerr:
## Issue
- `align` works by setting the space on either the left or right to `0`. This works fine except in situations where setting it to 0 means there isn't enough space for the dropdown. In these cases, the dropdown will end up off-screen because it doesn't know which side had its space reduced to be aligned or is an actual page boundary.

## Fix
When constraining the space to align the dropdown, it will always leave enough width for the dropdown even if it means it can't be perfectly edge aligned.

### Before
Before

### After
After


dbatiste (approved):
Looks good.

dbatiste:
Would be kinda nice to have a visual-diff case to catch any regression for this.

PR #964, opened 2020-11-24
Improve dropdown positioning logic

AllanKerr:
# Changes
If there isn't enough room to position the dropdown above or below, it will now choose the side with the most scrollable space rather than always picking the same arbitrary side.

This fixes an issue where dropdowns would open in a direction that prevented them from being interacted with.
Capture


dlockhart (approved):
Nice!

components/dropdown/dropdown-content-mixin.js
```diff
@@ -585,11 +593,14 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
-			(spaceAround.above > spaceAround.below)
-		);
+	_getOpenedAbove(spaceAround, spaceAroundScroll, spaceRequired) {
+		if (spaceAround.below >= spaceRequired.height) {
+			return false;
+		}
+		if (spaceAround.above >= spaceRequired.height) {
+			return true;
+		}
+		return spaceAroundScroll.above > spaceAroundScroll.below;
```
wongvincent, line 603: If there isn't enough space above or below, but there is enough scrollable space above and below, it will open to the side that has the most scrollable space?

Could that accidentally cause an element to open upwards even if there is enough scrollable space below, but not more than the amount of scrollable space above?

AllanKerr: Yeah, it could cause dropdowns to open up even if there is enough scrollable space below. Originally that was how I had implemented but having it choose the side with the most space ended up working a bit better when I tested it with dropdowns on a bunch of different pages. If there isn't enough visible space to open it up or down, I don't think the direction matters too much, as long as it opens somewhere that it can be scrolled to.

dlockhart (approved):

dlockhart (approved):

PR #966, opened 2020-11-30
Use viewport height/width for constraining dialog on small displays i…

dbatiste:
…nstead of body dimensions.


dlockhart (approved):
Is this testable with visual diff by setting the viewport super small and having a wide element on the page to force scroll?

dbatiste:
> Is this testable with visual diff by setting the viewport super small and having a wide element on the page to force scroll?

We can add a test for that scenario, but it wouldn't really catch a regression for the specific use-case since it seems to be Android specific.  That's probably a worthwhile scenario nonetheless - will add.

dlockhart (approved):

PR #967, opened 2020-12-01
input-date-time-range: style improvements

margaree:
- Add "to" between date and time when `child-labels-hidden`
- `input-date-time`s should appear next to each other when space and should wrap when not space
- "to" (if visible) should wrap to its own line between the `input-date-time` components
- if slotted content then no wrapping and `input-date-time` components should only ever be stacked

Notes:
- I used a `resizeObserver` on the parent to check for size change. I don't know if this is the best strategy/if there could be performance impact here so open to recommendations there


components/inputs/input-date-time.js
```diff
@@ -88,13 +89,6 @@ class InputDateTime extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(R
 			:host([hidden]) {
 				display: none;
 			}
-			d2l-input-date {
```
margaree, line 91: this was adding padding even when no time so I moved this into a styleMap

components/inputs/input-date-time.js
```diff
@@ -146,9 +152,9 @@ class InputDateTime extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(R
 					min-value="${ifDefined(this._minValueLocalized)}"
 					?required="${this.required}"
 					?skeleton="${this.skeleton}"
+					style="${styleMap(dateStyle)}"
 					.value="${this._parsedDateTime}">
-				</d2l-input-date>
-				<d2l-input-time
+				</d2l-input-date><d2l-input-time
```
margaree, line 157: There was some spacing being added that is no longer present if these are on the same line🤷‍♀️

components/inputs/input-date-time-range.js
```diff
@@ -189,44 +276,51 @@ class InputDateTimeRange extends SkeletonMixin(FormElementMixin(RtlMixin(Localiz
...
-						max-value="${ifDefined(this.maxValue)}"
-						min-value="${ifDefined(this.minValue)}"
-						?required="${this.required}"
-						?skeleton="${this.skeleton}"
-						value="${ifDefined(this.startValue)}">
-					</d2l-input-date-time>
-					<slot name="start"></slot>
+				<div class="${classMap(containerClassMap)}">
+					<div class="d2l-input-date-time-range-start-container">
+						<d2l-input-date-time
```
margaree, line 281: Changes here are mostly indentation

margaree:
Replaced by https://github.com/BrightspaceUI/core/pull/974

PR #970, opened 2020-12-02
fix: Allow skeleton Mixin to be apart of a Mixin chain

m6jones:
Basically, using `skeletonMixin(ListItemMixin(...))` will remove the ListItemMixin styles without this fix.


dbatiste (approved):
Thanks Maya.  :)

dlockhart (approved):

PR #972, opened 2020-12-03
input-time-range: style improvements (wrapping and "to")

margaree:
This PR is the same idea as https://github.com/BrightspaceUI/core/pull/967 with a very similar implementation other than the absence of the slots.


components/inputs/input-time-range.js
```diff
@@ -210,38 +282,47 @@ class InputTimeRange extends SkeletonMixin(FormElementMixin(RtlMixin(LocalizeCor
...
+							?label-hidden="${this.childLabelsHidden}"
+							?required="${this.required}"
+							?skeleton="${this.skeleton}"
+							time-interval="${ifDefined(timeInterval)}"
+							value="${ifDefined(this.startValue)}">
+						</d2l-input-time>
+					</div>
+					<div class="d2l-input-time-range-end-container">
+						<div class="d2l-input-time-range-to d2l-body-small d2l-skeletize">
+							${this.localize('components.input-time-range.to')}
```
dlockhart, line 306: Not sure if this is feasible, but now that this "to" thing with the wrapping is being used in 2 places within date/time would it be possible to centralize it in a component? Like a `<d2l-date-time-range-to>` or something with a slot for each side of the range.

If that makes things even more complex then ignore me, but just seems like if some of the styles and observers could be centralized that may save some code.

margaree: That sounds like a great idea

components/inputs/input-date-time-range-to.js
```diff
@@ -0,0 +1,198 @@
...
+import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
+
+class InputDateTimeRangeTo extends SkeletonMixin(LocalizeCoreElement(LitElement)) {
+
+	static get properties() {
+		return {
+			/**
+			 * Force block (stacked) range display if true
+			 */
+			blockDisplay: { attribute: 'block-display', type: Boolean },
```
dlockhart, line 16: I don't actually see this being set -- is it meant to be used in a use case not yet captured?

dlockhart: Ah I see it's used in the date-time-range PR 👍

dlockhart (approved):
Looks like a lot of whitespace tightening up in the visual diffs, which seems expected and originally unnecessary?

margaree:
@dlockhart yup exactly

PR #973, opened 2020-12-04
ignoring aXe rule

dlockhart:
This is failing the GHA-based builds with:

> Rule: presentation-role-conflict
    Impact: minor
    Elements of role none or presentation should be flagged (https://dequeuniversity.com/rules/axe/4.1/presentation-role-conflict?application=axeAPI)
    Issue target: d2l-calendar,table

We think it's safe to ignore this warning since without the label things don't get announced but we also can't ditch the presentation role.


dlockhart:
I can't get this to actually fail in Travis CI, only GitHub Actions so I'll just fix it there.

PR #979, opened 2020-12-08
Style scrollbars in primary-secondary template

emil-furniss:
Design expressed concern that the scrollbars in FACE were causing obtrusive resizing of the elements in the secondary panel of the primary-secondary-template as users open and close accordions and the scrollbar pops in and out of view. This is especially true on Windows where the scrollbar is quite chunky.

This is the current state with default scrollbars:

face-old-scrollbars

This PR styles the scrollbars on both the primary and secondary panels to mimic the styles of scrollbars on Mac, with the intent of making them less obtrusive. To avoid the elements moving around when the scrollbar pops in and out of view, it also sets `overflow: scroll` on the secondary panel to always show the scrollbar. Since it can be styled to blend into the background and fits into the default 10px gutter around the elements in the secondary panel, this results in a pretty smooth integration:

face-new-scrollbars

The primary panel also receives the styled scrollbars, but we retain the old behavior of only showing it when required. This is to avoid having a permanent white margin where the scrollbar is, since some face editors have designs that would clash with that. It is also thought that the primary panel does less resizing that the secondary panel, so the impact of the scrollbar popping in and out is less:

face-new-scrollbars-primary

The styling works on all Blink and Webkit browsers which covers the vast majority of our use cases, but Firefox and legacy Edge are the two outliers.

Firefox does not allow for the same degree of styling, so the scrollbars are not rounded at the top and bottom:

face-new-scrollbars-ff

And legacy Edge does not allow for any styling so it mostly looks unchanged, except for the permanently visible scrollbar on the secondary panel which has been deemed an acceptable evil for now:

edge-scrollbar-face


dbatiste (approved):

dbatiste:
Hey Emil, please hold off on merging this.  We are having a bit of discussion as to whether we should be sharing these style with a couple other things.

emil-furniss:
Actually not quite ready to merge this on my end either yet, still waiting to hear if this should go into 21.01. or wait for 21.02.

dlockhart:
I've updated this so that it only applies the scrollbars on Windows, since the macOS/Android ones are more fully-featured and we don't want to override those.

For now we're good just leaving this in the primary/secondary template, but if we see this pattern elsewhere we may break it out somewhere more shareable.

dlockhart (approved):

emil-furniss:
@dlockhart Am I right in thinking that the `overflow: scroll` to always show the scrollbar (on Windows) doesn't have a visual effect on Mac/Android since they always overlay the scrollbars over the content anyway?

dlockhart:
Yeah correct -- I had the same thought and tested it out... it doesn't seem to have any effect that I can see.

templates/primary-secondary/primary-secondary.js
```diff
@@ -27,6 +28,11 @@ function isMobile() {
 	return matchMedia('only screen and (max-width: 767px)').matches;
 }
 
+function isWindows() {
+	const val = window.navigator.userAgent.indexOf('Windows') > -1;
```
dbatiste, line 32: Nit: this could just be:
```javascript
const isWindows = window.navigator.userAgent.indexOf('Windows') > -1;
```
... so that it doesn't need to be evaluated for every `render` call.

dlockhart: Good call -- updated.

emil-furniss:
Thanks guys for all you help. PDM has confirmed that this should only go out in 21.02, so I'll hold off merging this until after branching next Fri, Dec 18.

emil-furniss:
Update is a result of some back-and-forth on the colors used by the track and thumb in order to meet the required a11y contrast ratios. We've darkened the thumb and added a transparent white overlay over the track to lighten any backgrounds which might be too dark.

dlockhart (approved):

PR #983, opened 2020-12-08
input-date-range: style tweaks: add "to" and wrapping

margaree:
Similar to the PRs for `input-date-time-range` and `input-time-range`. For this one, the axe tests and sometimes the unit tests were failing with the error `ResizeObserver loop limit exceeded`. There did seem to be an additional run of the ResizeObserver happening for the parent node. I have not yet been able to figure out why (looks like it's somewhat related to additional resizing happening because `input-date` is using `input-text` where as `input-time` does not but don't know much more). I have not seen this error in testing on the demo. Some research suggests it's safe to suppress this error, but open to ideas on why this would be happening.


dlockhart (approved):
Not sure if you want to re-enable the date/time visual diff tests as part of this or @svanherk will that be problematic for you?

svanherk:
Been in meetings all afternoon but I can put up these visual diff tests next if you want them checked before merging

PR #987, opened 2020-12-09
Add new icons for html-editor icon-set.

dbatiste:


dlockhart:
Are the existing icons in use somewhere and if so, we're OK with them changing right away?

dbatiste:
> Are the existing icons in use somewhere and if so, we're OK with them changing right away?

The monolith editor uses a completely different set of icons, all png based.  We could update them, but I think it's best to just leave them alone, and remove them when the new component replaced it. https://search.d2l.dev/xref/lms/lp/_lang/ImageTerms/images/Framework/HtmlEditor/

The existing html-editor component used by FRAs uses yet another set of icons that are also different (intentionally), so we can't update them easily either, and they will also become obsolete.  https://github.com/Brightspace/d2l-html-editor/tree/master/icons

I don't think there's much actually referring to these icons (here in core).  We _could_ move them all to the new editor repo.

dlockhart:
> We could move them all to the new editor repo.

Yeah that's kind of where I was going too... we'd need to be 100% sure that nothing is using them, but yeah they do seem kind of unnecessary in here.

dbatiste:
Closed this PR since moving editor icons to the HtmlEditor repo.  These icons will be removed in another PR once we're certain that we've copied all the required icons.

PR #1010, opened 2020-12-15
docs: fix visual-diff test path for core

dlockhart:
Because core is slightly different.


dbatiste (approved):

krshnpatel-d2l (approved):
Got 3 timeouts during golden generation and sometimes flakey when running the tests but seems to work! ✔️

PR #1012, opened 2020-12-16
fix: Remove input-width from input-number and input-text

krshnpatel-d2l:
## Summary
~ Removed `input-width` from `input-number` and `input-text` components

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -22,7 +22,6 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 			autocomplete: { type: String },
 			autofocus: { type: Boolean },
 			disabled: { type: Boolean },
-			inputWidth: { attribute: 'input-width', type: String },
 			label: { type: String },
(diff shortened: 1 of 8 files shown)


components/inputs/input-number.js
```diff
@@ -58,7 +57,6 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 		super();
 		this.autofocus = false;
 		this.disabled = false;
-		this.inputWidth = '4rem';
```
dlockhart, line 61: We're going to want to apply this `4rem` width to the `<input type="text">` in order to keep the same size.

dlockhart: Which... maybe means `<d2l-input-text>` needs to keep `input-width`?

krshnpatel-d2l: So if we keep this then adding an inline style of `width: 10rem` would set `input-number`'s container width to `10rem` and always default the `input-text`'s container `max-width` to `4rem`.

So that would mean that we would have to keep `input-width` for `input-number` as well because we would have to set both `style="width: 10rem;"` and `input-width="10rem"` inside the MVC control.

PR #1016, opened 2020-12-23
Fix RTL label causing an `overflow-x` which creates horizontal scrollbar

mostlyfabulous:
### FACE Assignments

Fixes: https://trello.com/c/DwgFvfRz/262-rtl-new-picker-causes-horizontal-scroll-in-rhp

**Cause:**
By setting the `time` field with the new date picker a horizontal scrollbar appears in the secondary panel.

**Initial investigation:**
`label` for the `time` fields triggers an overflow in the x-axis and so a horizontal scrollbar appears as the `label` is set to render at `left: -10000px`. Setting it to `right: -10000px` maintains functionality and has no overflow/horizontal scrollbar, not sure if the RTL mixin needs a fix to get this to happen.

**Before:**
image

**After:**
image


dbatiste (approved):

mostlyfabulous:
🙁 I didn't put `fix: ` in the right place for a release...

PR #1017, opened 2020-12-23
Ensure the switch toggle is stacked on-top of the switch icons.

dbatiste:
This is required specifically for RTL.

Before fix (RTL):
localhost_8000_components_switch_demo_switch html (1)

After fix (RTL):
localhost_8000_components_switch_demo_switch html


dbatiste:
Seems we have a couple really flakey tests.  I will probably merge this as-is, unless I can sort out the reason for the flake.  Ran CI 3 times and got different set of errors with this in both BrowserStack and Sauce.

svanherk (approved):
Should we have a visual-diff test for this?

dbatiste:
> Should we have a visual-diff test for this?

Haha good question - we actually have RTL tests for this, but for some strange reason, our visual-diff RTL setup doesn't capture this bug.  I will look at an alternate way to apply dir in the test that might capture it.

svanherk (approved):
Looks good!

Issue #882, opened 2020-10-01
Date input: info tooltip shows briefly on click


dlockhart:
The info tooltip is super handy, especially when using the keyboard as it tells you the date format and how to open the mini calendar.

However, when you click into the input the mini calendar opens automatically. But for a brief moment before it does, the info tooltip is shown, which creates some jitter.

![calendar](https://user-images.githubusercontent.com/5491151/94828493-86d6ad00-03d7-11eb-9e03-11b05e6cfc20.gif)

Ideally we would only show the tooltip when focused using the keyboard. If a click event was the source, we'd just open the calendar and not show the tooltip.

PR #1024, opened 2021-01-06
input-date: fix tooltip jitter on click

margaree:
Fixes https://github.com/BrightspaceUI/core/issues/882

When `input-text` is clicked, it triggers the tooltip to be shown because of it being focused on. In order for the click not to show the tooltip, I added a timeout in the focus event listener for `input-text` where it waits to see if the `mouseup` event happens which would then set `_inputTextFocusMouseup` to true and in that case the tooltip should stay hidden. There doesn't seem to be a way to easily detect if the mouse vs keyboard caused the focus just within the focus listener, which would be more ideal.


components/inputs/test/input-date-time.visual-diff.js
```diff
@@ -46,8 +46,10 @@ describe('d2l-input-date-time', () => {
 
 	it('basic-focus', async function() {
 		await page.$eval('#basic', (elem) => elem.focus());
-		const rect = await visualDiff.getRect(page, '#basic');
-		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		setTimeout(async() => {
+			const rect = await visualDiff.getRect(page, '#basic');
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		}, 160);
```
dlockhart, line 52: Not sure if using the sinon fake timers here to quickly advance the time could make these run faster, or if it'll just make them not work at all. 🤷

components/inputs/input-date.js
```diff
@@ -349,8 +353,19 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
+		this._inputTextFocusShowTooltip = true;
+	}
+
 	_handleInputTextFocus() {
 		this._formattedValue = this._shownValue ? formatISODateInUserCalDescriptor(this._shownValue) : '';
+
+		// hide tooltip when focus, wait to see if click happened, then show
+		this._inputTextFocusShowTooltip = false;
+		setTimeout(() => {
+			if (!this._inputTextFocusMouseup) this._inputTextFocusShowTooltip = true;
```
dlockhart, line 367: Crazy that there isn't a better way to detect the source (mouse vs. keyboard) of a focus event but you're right... I couldn't find another way either.

dlockhart (approved):
Change looks good, but the visual diffs look a little off. I think for some of them, we're now expecting the tooltip balloon triangle thing to go away, but there are more changes than just that.

components/inputs/test/input-date-time.visual-diff.js
margaree: That's a good idea - I'll give it a go

dlockhart (approved):

PR #1028, opened 2021-01-07
Fix whitespace in dialog footer for fullscreen mode.

dbatiste:
This PR makes a small fix to the whitespace in the dialog footer.  We set an unequal padding in the footer because the balance of the padding comes from bottom margin on the slotted buttons (which is used to space buttons vertically if/when they wrap).  Previously, the responsive CSS was not taking this into account, resulting in too much whitespace below the buttons.

**before fix:**
image

**after fix:**
image


components/dialog/dialog-styles.js
```diff
@@ -126,7 +126,7 @@ export const dialogStyles = css`
 			padding: 0 20px;
 		}
 		.d2l-dialog-footer {
-			padding: 20px;
+			padding: 20px 20px 2px 20px;
```
dlockhart, line 129: What happens if for some reason the footer doesn't have any buttons, or are we just not expecting that case?

dbatiste: Yeah, so currently we assume there will be at least one button in the footer.  We could add a check for slotted elements and hide the footer completely.  That might be a valid use-case that we haven't seen, or aren't aware of.

dbatiste: So... I correct myself... this is actually a use-case we handle for fullscreen dialogs.  We _could_ do a similar check for standard dialogs as well if we think it's worthwhile.

dlockhart: This definitely seems better than how it was regardless.

dlockhart (approved):

components/dialog/dialog-styles.js
dbatiste: Yeah... I think we want to add this check.  We need to decide how we want it to look when it overflows and there's scrolling... i.e. padding-wise and the overflow shadow.  I've put together a couple options and discussing with Jeff to decide what exactly we want.

dbatiste: This is the result we decided on... 5px padding for the footer when empty so user can still distinguish the overflow shadow.
image

Issue #884, opened 2020-10-01
Date/time: initial performance


dlockhart:
When date/time inputs and ranges are used inside of a larger app, they tend to be the last component on the page to render. This causes some jank as things shift around, but when skeleton loaders are used it also causes the skeleton "pulse" to be out of sync with the rest of the skeletons on the page.

There could be several things causing this:
- The components could be rendering multiple times initially, possibly unnecessarily
- They could be rendering expensive things that may be able to be deferred until later -- for example the date input calendar dropdown or the time input intervals.
- Their default non-rendered sizes may be off slightly, which could be causing some jank. This may not be possible to solve.

I'd recommend doing some instrumentation (`console.log` FTW!) inside the various `render()`s to see when, why and how often things are being rendered. Additionally, `performance.now()` could be used in various lifecycle positions to determine how long things are taking to render and attempt to make improvements.

PR #1033, opened 2021-01-11
input-date/time* components: improve initial load

margaree:
Fixes https://github.com/BrightspaceUI/core/issues/884

This works to minimize the number of renders that occur on initial component load. The following were improved:
- `input-date`: calendar is not rendered until the dropdown is first opened
- `input-time`: time intervals are not rendered until the dropdown is first opened
- `input-date-time`: remove extra render being required when `value` or `min/maxValue` set
- `input-date-time`: don't render time when date is not set (not sure if post-first-render it should just be hidden if date is unset?)

Other things that were investigated that were not possible to improve:
- `input-time` renders twice when no `value` due to needing to set a default value (likely needs to be done in `firstUpdated` since it relies on checking for both `value` and `defaultValue`)
- `input-date` renders twice in order to get the correct width since it's based on the maximum width of `empty-text`, formatted example, or widest possible value

I want to do another pass to see if it's possible to reduce the number of renders that happen at other points like when opening the calendar dropdown, but I'll do that separately.


components/calendar/calendar.js
```diff
@@ -538,8 +538,9 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		});
 	}
 
-	focus() {
+	async focus() {
```
margaree, line 541: This was needed for the focusing on today/the selected date to work when opening the calendar dropdown initially

components/inputs/input-date-time.js
```diff
@@ -101,10 +100,58 @@ class InputDateTime extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(R
 		this._dropdownOpened = false;
 		this._inputId = getUniqueId();
 		this._namespace = 'components.input-date-time';
-		this._parsedDateTime = '';
```
margaree, line 104: I moved this into render to prevent an additional render

components/inputs/input-date-time.js
```diff
@@ -130,7 +177,27 @@ class InputDateTime extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(R
 			dateStyle.paddingRight = '0.3rem';
 		}
 
+		const parsedValue = this.value ? (this.localized ? this.value : getLocalDateTimeFromUTCDateTime(this.value)) : '';
 		const tooltip = (this.validationError && !this._dropdownOpened && this.childErrors.size === 0) ? html`<d2l-tooltip align="start" announced for="${this._inputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
+		const inputTime = !timeHidden ? html`<d2l-input-time
```
margaree, line 182: I switched to only rendering time when there is a value/date selected. Should this only happen on the first render and then after that it's just hidden if the user clears the date?

svanherk:
> Could not generate new goldens - your code changes will update golden files that you do not have the latest version of. Please rebase or merge master into your branch.

Hmm... something went weird here. Will investigate.

components/inputs/input-date-time.js
dlockhart: I think this is fine the way you have it now... it adds a small rendering expense but only when a date is added/cleared.

dlockhart (approved):
Looks like great improvements!

PR #1035, opened 2021-01-12
announce: remove duplication of messages

margaree:
Fixes https://github.com/BrightspaceUI/core/issues/526

This fix checks if a message has already been added to the aria-live container. If not, it performs as before. If it has, it removes then re-adds it (otherwise it might not actually get read). 

This was tested with the `input-time-range` component validation which can end up adding the text of multiple "Start time must be before end time" tooltips to the announcement if the user is quickly tabbing around.

This assumes that there isn't a scenario where there should be duplicate messages that are repeated in succession. A different fix will need to be made if that is the case.

### DIFF
--- a/helpers/announce.js
+++ b/helpers/announce.js
@@ -1,5 +1,6 @@
 let timeoutId = null;
 let container = null;
+let messages = {};
 
 export function announce(message) {
 	if (!message) return;
@@ -25,8 +26,16 @@ export function announce(message) {
 	treat the change as an update. Firefox sometimes ignores changes if the region
 	and update are made too quickly in succession. RequestAnimationFrame is not
 	sufficient here. */
+	let child;
+	if (Object.keys(message).length === 0 || !Object.keys(messages).includes(message)) {
+		child = document.createTextNode(message);
+		messages[message] = child;
+	} else {
+		child = messages[message];
+		child.parentNode.removeChild(child);
+	}
 	setTimeout(() => {
-		container.appendChild(document.createTextNode(message));
+		container.appendChild(child);
 	}, 100);
 
 	/* Need to purge old messages so that they are not discovered by screen readers
@@ -36,6 +45,7 @@ export function announce(message) {
 		container.parentNode.removeChild(container);
 		container = null;
 		timeoutId = null;
+		messages = {};
 	}, 10000);
 
 }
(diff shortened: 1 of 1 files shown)


helpers/announce.js
```diff
@@ -25,8 +26,16 @@ export function announce(message) {
...
 	and update are made too quickly in succession. RequestAnimationFrame is not
 	sufficient here. */
+	let child;
+	if (Object.keys(message).length === 0 || !Object.keys(messages).includes(message)) {
+		child = document.createTextNode(message);
+		messages[message] = child;
+	} else {
+		child = messages[message];
+		child.parentNode.removeChild(child);
+	}
```
dlockhart, line 36: Another option here instead of keeping a Map/reference to all the DOM nodes (which can become stale and create circular references between the DOM and JavaScript) would be to do a runtime search for it. Something like this:

```javascript
const elem = [...container.children].find((c) => c.textContent === message);
if (elem) elem.parentNode.removeChild(elem);
container.appendChild(document.createTextNode(message);
```

margaree: Good idea - that looks a lot cleaner

dbatiste (approved):
If you want to try some other scenarios besides the input you're working on, there is a demo pages for `announce`.  Ignoring the legit duplicate message could come up as an issue, but maybe it's still better.

dbatiste:
I also highly recommend testing in as many screen reader / browser combos as possible.  The aria-live regions are extremely fragile with inconsistent behaviour across browsers and AT.  It's also extremely frustrating at times.

dlockhart (approved):

margaree:
I tested with the input-time-range page as well as the announce demo page with the following (announce doesn't actually work for VoiceOver + Firefox and JAWs + Firefox):

- NVDA + ...
   - Firefox
   - Edgium
   - Chrome
- JAWS + ...
   - Edgium
   - Chrome
- VoiceOver + ...
   - Chrome
   - Safari

Most worked as expected.

Notable findings:
- Does not work for rapidly repeating actions with VoiceOver + Chrome (e.g., user keeps pushing the "announce" button, only reads the first time) - overall not too worried about this. Seems like a power user issue as well as I don't think VoiceOver + Chrome is a common usage
- Timeout increase fixed ^ that problem for VoiceOver + Safari

dlockhart (approved):

PR #1036, opened 2021-01-13
fix: Verify cache storage is available before trying to use batch route

pureooze:
[DE41794](https://rally1.rallydev.com/#/110294864140d/dashboard?detail=%2Fdefect%2F466890048628)

This fixes an issue where Firefox private browsers could not use OSLO (and so not get overrides) because they would try to use the batch call but then fail when trying to open CacheStorage because it is not available in private mode.

**More info:**
https://bugzilla.mozilla.org/show_bug.cgi?id=1320796
https://bugzilla.mozilla.org/show_bug.cgi?id=1112134

### DIFF
--- a/helpers/test/getLocalizeResources.test.js
+++ b/helpers/test/getLocalizeResources.test.js
@@ -208,7 +208,8 @@ describe('getLocalizeResources', () => {
 
 		expect(formatFuncSpy).to.have.been.callCount(1); // 1 lms
 		expect(formatFuncSpy).to.have.been.calledWithExactly();
-		expect(openSpy).to.have.been.calledOnceWithExactly('d2l-oslo');
+		expect(openSpy).to.always.have.been.calledWithExactly('d2l-oslo');
+		expect(openSpy).to.have.callCount(3);
 		expect(matchSpy).to.have.been.callCount(1);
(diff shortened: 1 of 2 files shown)


helpers/test/getLocalizeResources.test.js
```diff
@@ -208,7 +208,8 @@ describe('getLocalizeResources', () => {
 
 		expect(formatFuncSpy).to.have.been.callCount(1); // 1 lms
 		expect(formatFuncSpy).to.have.been.calledWithExactly();
-		expect(openSpy).to.have.been.calledOnceWithExactly('d2l-oslo');
+		expect(openSpy).to.always.have.been.calledWithExactly('d2l-oslo');
+		expect(openSpy).to.have.callCount(3);
```
pureooze, line 212: So seems like this increases how many times the cache gets opened. This seems not ideal but not sure how else we can verify if CacheStorage is available to the browser.

dlockhart: Yeah in my experience I haven't seen a way to check if local storage is available without trying to write/read to it.

dlockhart (approved):

kieranderson (approved):

PR #1038, opened 2021-01-13
fix: increase skeleton z-index

mdulat:
The new html editor has a z-index = 1 on the toolbar, so the skeleton doesn't overlap it.

image

### DIFF
--- a/components/skeleton/skeleton-mixin.js
+++ b/components/skeleton/skeleton-mixin.js
@@ -18,7 +18,7 @@ export const skeletonStyles = css`
 		position: absolute;
 		right: 0;
 		top: 0;
-		z-index: 1;
+		z-index: 2;
 	}
 	@media (prefers-reduced-motion: reduce) {
 		:host([skeleton]) .d2l-skeletize::before {


dlockhart:
@dbatiste Do we want to make HTML editor use the `SkeletonMixin` so teams don't have to do this manually? At that point, it may very well run into this same issue, but at least we could solve it in that context.

dbatiste:
> @dbatiste Do we want to make HTML editor use the `SkeletonMixin` so teams don't have to do this manually? At that point, it may very well run into this same issue, but at least we could solve it in that context.

I think so yeah. @grant-cleary : not sure what all you have on your plate. Is this something you can help with as well?

grant-cleary:
> I think so yeah. @grant-cleary : not sure what all you have on your plate. Is this something you can help with as well?

Yeah, I can take some time to tackle this.

grant-cleary:
FYI that https://github.com/BrightspaceUI/htmleditor/pull/74 should remove the need for this. I think it makes sense for the editor component to override the z-index value instead of asking the skeleton component to do so.

dbatiste (approved):

dbatiste:
Can we close this then?

grant-cleary:
> Can we close this then?

I think so, yeah, assuming nobody has very strong objections to handling this in the editor component.

dlockhart:
Closing in favour of @grant-cleary's change in HTML editor.

PR #1039, opened 2021-01-14
Remove code owners for the visual-diff goldens

svanherk:
This won't do a lot if you're watching the whole `core` repo anyways, but I think it would result in one less email.  If you aren't watching it, then you'd no longer get notified about the visual-diff PRs unless you caused the commit that opened it.  For whatever reason, I couldn't do `screenshots/ci/golden/`, I had to pick just one level.

Thoughts?


dlockhart (approved):
If that works and results in 1 less email, 🙌

dbatiste (approved):

PR #1041, opened 2021-01-14
Revert "fix: Verify cache storage is available before trying to use batch route"

pureooze:
Reverts BrightspaceUI/core#1036

Seems like OSLO is having issues on quad with one of the activity editor collections being a 404 (check cache storage for status codes), and this seems like the only recent change to OSLO. Not sure why it would cause the issue but gonna try a rollback just incase because I am off tomorrow.

image


pureooze:
Seems like a different change in a different repo might be the actual cause of this, closing this PR.

PR #1042, opened 2021-01-15
calendar: reduce renders

margaree:
The biggest culprits were:
- `_focusDate` which sometimes was being set to the same value multiple times, so work was done to minimize that.
- Unnecessary `await this.updateComplete`

### DIFF
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -415,6 +415,8 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		);
 		if (dropdownContent) this._dialog = true;
 
+		this.addEventListener('blur', () => this._isInitialFocusDate = true);
+
 		this.addEventListener('d2l-localize-behavior-language-changed', () => {
 			calendarData = null;
 			getCalendarData(true);
@@ -795,8 +797,6 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		} else {
 			this._focusDateAddFocus();
 		}
-		await this.updateComplete;
-		this._isInitialFocusDate = true;
 	}
 
 	_triggerMonthChangeAnimations(increase, keyboardTriggered, transitionUpDown) {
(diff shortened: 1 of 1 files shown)


components/calendar/calendar.js
```diff
@@ -594,15 +606,14 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 		this.selectedValue = formatDateInISO({ year: year, month: (parseInt(month) + 1), date: date });
+		this._dateSelected = true;
 
 		const eventDetails = {
 			bubbles: true,
 			composed: true,
 			detail: { date: this.selectedValue }
 		};
 		this.dispatchEvent(new CustomEvent('d2l-calendar-selected', eventDetails));
-		await this._updateFocusDateOnChange();
```
margaree, line 604: This was moved to be in `_updateFocusDateOnChange` since it was duplicated effort with `updated`

components/calendar/calendar.js
```diff
@@ -811,11 +820,15 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	}
 
 	async _updateFocusDate(possibleFocusDate, latestPossibleFocusDate, allowDisabled) {
-		await this.updateComplete;
-		if (isDateInRange(possibleFocusDate, getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue)) || allowDisabled) {
+		if ((!this.minValue && !this.maxValue)
```
margaree, line 823: These parts don't need `updateComplete` to have happened

components/calendar/calendar.js
```diff
@@ -533,7 +537,11 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			if (prop === '_shownMonth' && this._keyboardTriggeredMonthChange) {
 				this._focusDateAddFocus();
 			} else if (prop === 'selectedValue' && this.selectedValue) {
-				await this._updateFocusDateOnChange();
+				if (!this._focusDate || !checkIfDatesEqual(this._focusDate, getDateFromISODate(this.selectedValue))) {
```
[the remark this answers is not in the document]
GeorgePlukov: Minor change but it may be nice to split this line into two to make the comparison easier to read. Something like
```suggestion
				const selectedDate = getDateFromISODate(this.selectedValue);
				if (!this._focusDate || !checkIfDatesEqual(this._focusDate, selectedDate)) {
```

components/calendar/calendar.js
```diff
@@ -811,11 +820,15 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	}
 
 	async _updateFocusDate(possibleFocusDate, latestPossibleFocusDate, allowDisabled) {
-		await this.updateComplete;
-		if (isDateInRange(possibleFocusDate, getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue)) || allowDisabled) {
+		if ((!this.minValue && !this.maxValue)
+			|| (isDateInRange(possibleFocusDate, getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue)) || allowDisabled)
```
GeorgePlukov, line 824: Similar to above these embedded function calls could be extracted onto their own lines

margaree:
Looks like that visual-diff test failure was flake. I'll keep an eye on it and if it continues I'll skip it and tweak. Maybe some timeouts/waiting is needed or something like that.

dlockhart (approved):

PR #1045, opened 2021-01-19
New d2l-input-textarea component.

dbatiste:
Don't be frightened by the number of files in this PR... it's almost all tests and visual-diffs.  The real interesting bit is the component, which isn't even that big.  😄 

This PR introduces a new `d2l-input-textarea` component that wraps the native `textarea`, and further provides validation (same as `d2l-input-text` along with `d2l-tooltip`), skeleton mode, and some nifty auto-grow behaviour (originally used in foundations component by @m6jones).

Auto-grow: the strategy used for auto-grow is to basically mirror the content in a hidden `div` with a common parent as the `textarea`, and utilize height: 100% on the `textarea`.  This has an a11y advantage of other solutions that utilize `contenteditable` or require measurement on every keystroke. This is essentially the same strategy that's employed by `iron-autogrow-textarea`, except for a Lit element.

Some screenshots...

**Default**
image

**Min-Height**
image

**Placeholder**
image

**Invalid/Tooltip**
image


components/inputs/docs/input-text.md
```diff
@@ -98,24 +98,3 @@ class MyElem extends LitElement {
 
 }
 ```
-
```
dbatiste, line 101: This just got moved to the dedicated readme for textareas.

components/inputs/input-textarea.js
```diff
@@ -0,0 +1,261 @@
...
+		return true;
+	}
+
+	_handleInvalid(e) {
+		e.preventDefault();
+	}
+
+}
+
+customElements.define('d2l-input-textarea', InputTextArea);
```
dbatiste, line 261: I chose this name to be consistent with the format of the rest of the components in the inputs folder.  `d2l-inout-*`.  If anyone has objection I'd be open to changing this to `d2l-textarea`, which would align more closely to the underlying element but be inconsistent with our other components. (ex. `d2l-input-fieldset` is one such similar case).

components/inputs/test/input-textarea.visual-diff.js
```diff
@@ -9,36 +9,64 @@ describe('d2l-input-textarea', () => {
...
 		browser = await puppeteer.launch();
-		page = await visualDiff.createPage(browser, { viewport: { width: 800, height: 1300 } });
+		page = await visualDiff.createPage(browser, { viewport: { width: 800, height: 2000 } });
 		await page.goto(`${visualDiff.getBaseUrl()}/components/inputs/test/input-textarea.visual-diff.html`, { waitUntil: ['networkidle0', 'load'] });
 		await page.bringToFront();
 	});
 
 	after(async() => await browser.close());
 
-	['wc', 'sass'].forEach((type) => {
```
dbatiste, line 19: I added a few more visual-diff tests for things specific to `d2l-input-textarea`.  There's a little duplication between this and the Sass, but there are enough differences that combining them starts to get messy, and I'd like to keep them both separate in code, and separate in reporting.

components/inputs/docs/input-textarea.md
```diff
@@ -0,0 +1,81 @@
...
+| `value` | String, default: `''` | Value of the `textarea` |
+
+**Accessibility:**
+
+To make your usage of `d2l-input-textarea` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `aria-invalid` | [Indicate that the `textarea` value is invalid](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-invalid) |
+| `aria-label` | Use when `label` does not provide enough context. Only applies if no `label-hidden`. |
```
dlockhart, line 41: This shouldn't be here, ya? Like `label` + `label-hidden` = `aria-label` I thought.

components/inputs/input-textarea.js
```diff
@@ -0,0 +1,261 @@
...
+				display: none;
+			}
+			.d2l-input-label {
+				display: inline-block;
+				vertical-align: bottom;
+			}
+			:host .d2l-input-textarea-container {
+				max-width: 100%;
+				position: relative;
+			}
```
dlockhart, line 84: Is the ':host' prefix needed for all of these? I thought it was always implied.

components/inputs/input-textarea.js
```diff
@@ -0,0 +1,261 @@
...
+			return elem.validity;
+		}
+		return super.validity;
+	}
+
+	_getAriaLabel() {
+		if (this.label && this.labelHidden) {
+			return this.label;
+		}
+		if (this.hasAttribute('aria-label')) {
```
dlockhart, line 223: Supporting `aria-label` was a backwards-compatibility thing that we had to do for `<d2l-input-text>` because it didn't initially support `label-hidden`. No need to support it here.

components/inputs/input-textarea.js
dlockhart: I'm OK either way here. I do like that all the inputs are grouped together like this, but this is the first one I believe that isn't natively `<input type="something">`.

components/inputs/docs/input-textarea.md
```diff
@@ -0,0 +1,81 @@
...
+
+| Property | Type | Description |
+|--|--|--|
+| `aria-invalid` | String | Indicates that the `textarea` value is invalid |
+| `description` | String | A description to be added to the `textarea` for accessibility |
+| `disabled` | Boolean | Disables the `textarea` |
+| `label` | String, required | Label for the `textarea` |
+| `label-hidden` | Boolean | Hides the label visually (moves it to the `textarea`'s `aria-label` attribute) |
+| `max-height` | String, default: 12rem | Maximum height of the input before scrolling. `none` allows `textarea` to grow infinitely. |
+| `min-height` | String, default: none | Minimum height of the input. If `min-height` and `max-height` are equal then auto-grow will be disabled. |
```
dlockhart, line 27: Thoughts about having the default min-height be multi-line, similar to how a default `<textarea>` would look? I'm worried that if it's single line it won't look like a textarea and folks will always be changing the default.

components/inputs/docs/input-textarea.md
```diff
@@ -0,0 +1,81 @@
...
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `aria-invalid` | String | Indicates that the `textarea` value is invalid |
+| `description` | String | A description to be added to the `textarea` for accessibility |
+| `disabled` | Boolean | Disables the `textarea` |
+| `label` | String, required | Label for the `textarea` |
+| `label-hidden` | Boolean | Hides the label visually (moves it to the `textarea`'s `aria-label` attribute) |
+| `max-height` | String, default: 12rem | Maximum height of the input before scrolling. `none` allows `textarea` to grow infinitely. |
```
dlockhart, line 26: Curious whether it would be possible to just rely on the CSS `style="max-height: 5rem; min-height: 2rem;"` values instead of needing to define our own? Or is it because changing them needs to trigger a re-render?

components/inputs/input-textarea.js
```diff
@@ -0,0 +1,261 @@
...
+				max-width: 100%;
+				position: relative;
+			}
+			:host textarea {
+				height: 100%;
+				left: 0;
+				position: absolute;
+				resize: none;
+				top: 0;
+				z-index: 2;
```
dlockhart, line 91: What does the default width end up being? Like does it fill any space that it's placed inside? I'm not sure what the correct thing would be here... but maybe something similar to the default `<textarea>` or maybe design has some thoughts. Same thoughts about the default height that I mentioned up above.

components/inputs/docs/input-textarea.md
dbatiste: This is true, but it falls back to `aria-label` on the component if neither are specified.  This is same as `d2l-input-text`.  We do mention that `label` is required, so I feel like this is overkill... it's shouldn't actually get used this way.

components/inputs/input-textarea.js
dbatiste: Thanks, these can mostly be removed.  There was one case where the `:host` was providing some necessary selectivity, but I will change it to `textarea.d2l-input` to make that clearer.

components/inputs/docs/input-textarea.md
dbatiste: Yeah that's a good question.  I'd be fine with it.  What do you think @geurts ?

components/inputs/docs/input-textarea.md
dbatiste: I can experiment with that - it might work.

components/inputs/input-textarea.js
dbatiste: The default width ends up being 100%, however the consumer can limit the width if they choose:
```html
<d2l-input-textarea style="width: 250px;">
```
However, it does not auto-grow horizontally.  So this doesn't quite work...
```html
<d2l-input-textarea style="display: inline-block;">
```
...because the mirror is a block and must match the `textarea` dimensions (it almost works).  We _might_ be able to enable horizontal auto-grow if that's required.

dlockhart: I don't think horizontal autogrow would be useful. 100% width of its container seems correct, unless design thinks having a default max-width would make sense?

geurts: I think defaulting to 100% width makes sense, that's the way it gets used most of the time these days.

components/inputs/docs/input-textarea.md
geurts: I'm fine with that too. Maybe 6rem would work?

components/inputs/docs/input-textarea.md
dbatiste: I can get the auto-grow to work properly (with the `style="min/max-height"`) by making the host `display: flex; flex-direction: column;`, the label `flex: none` and the remainder `flex: auto;` with `overflow: auto;`, and also move the `d2l-tooltip` out of the container our so it doesn't get clipped by the overflow.

However it creates some other problems: 

* the label ends up using some of the height, and the amount depends on whether the label wraps or not, so it's sensitive to browser resizing too.  Maybe we could be ok with this, but it seems weird, and if design wants a particular usage to have a certain input height, it wouldn't be guaranteed. 
* it surfaces the `display` requirement on the `host`, and if the consumer overrides it then the behaviour breaks.  If we want to allow the consumer to use `display: inline-block;` with horizontal auto-grow, then that further complicates things because they'd need to use `display: inline flex`, which so far I've been unable to get working properly with the auto-grow, but even if it worked, I don't think the consumer should need to know the internal flex requirement.

Maybe I am missing some other options, although I've played around with it longer than I'd like to admit.  😆  I'm open to exploring this further, but if the label issue is a deal breaker then I opt for not spending more time ironing out the `display` issue.

components/inputs/input-textarea.js
dbatiste: Awesome!  Thanks.  That makes the one issue in my other comment moot.  :smile:

components/inputs/docs/input-textarea.md
dlockhart: Yeah doesn't seem worth it if it's causing this much grief -- mostly I like avoiding creating properties/attributes for things that exist already in CSS. What if instead we called it `max-cols` and `min-cols` and then internally translated those integers to `rem` units? That way it's less confusing with the CSS.

dbatiste: I'm guessing you mean rows, or lines?  Maybe `min-lines` / `max-lines`?

dlockhart: Right yes, rows or lines!

dlockhart (approved):

PR #1051, opened 2021-01-21
Switch component focus styling and visual improvements

GeorgePlukov:
Changes:
    - Adds FocusVisibility polyfill 
    - Toggle size from 1.3rem -> 1.1rem
    - Switch width -> 3rem
    - Adds hover effect for mouseover
    - Only show outer border for keyboard events
    - Adds visual diff tests for hovering

https://user-images.githubusercontent.com/1743751/105231386-cd8e1580-5b34-11eb-888c-d413c3d151c1.mov


components/switch/switch-mixin.js
```diff
@@ -4,8 +4,9 @@ import { css, html } from 'lit-element/lit-element.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { FocusVisiblePolyfillMixin } from '../../mixins/focus-visible-polyfill-mixin.js';
 
-export const SwitchMixin = superclass => class extends RtlMixin(superclass) {
+export const SwitchMixin = superclass => class extends RtlMixin(FocusVisiblePolyfillMixin(superclass)) {
```
dlockhart, line 9: I'd recommend the `eslint` VS Code (or other IDE) plugin as it'll catch these as you go!

components/switch/test/switch.visual-diff.js
```diff
@@ -27,8 +37,20 @@ describe('d2l-switch', () => {
 			{ name: 'off', selector: '#off' },
 			{ name: 'off-focus', selector: '#off', action: (selector) => page.$eval(selector, (elem) => elem.focus()) },
```
dlockhart, line 38: I think you'll need to use our `forceFocusVisible` helper -- have a look at what `button-icon` is doing. Programatically calling `.focus()` doesn't trigger its heuristic, which requires a combination of keyboard events for it to believe that the focus merits the visible treatment.

components/switch/switch-mixin.js
```diff
@@ -40,20 +41,23 @@ export const SwitchMixin = superclass => class extends RtlMixin(superclass) {
...
+				box-shadow: 0 0 0 1px var(--d2l-color-celestine) inset;
+			}
 			:host([disabled]) .d2l-switch-container {
 				cursor: default;
 				opacity: 0.5;
 			}
 			:host([disabled]) .d2l-switch-container:focus,
-			:host([disabled]) .d2l-switch-container:hover {
-				border-color: transparent;
+			:host([disabled]) .d2l-switch-container:hover,
```
dlockhart, line 55: I think the first line (`:host([disabled]) .d2l-switch-container:hover,`) can be excluded since it doesn't have a `box-shadow`?

components/switch/switch-mixin.js
```diff
@@ -76,52 +83,69 @@ export const SwitchMixin = superclass => class extends RtlMixin(superclass) {
...
 			}
-
+			:host([on]) .d2l-switch-icon-off,
+			:host([on]) .d2l-switch-icon-off {
+				transform: scale(0.35);
+			}
+			:host(:not([on])) .d2l-switch-icon-on,
+			:host(:not([on])) .d2l-switch-icon-on {
+				color: var(--d2l-color-celestine);
+				transform: scale(0.35);
```
dlockhart, line 139: Cool way to accomplish this effect!

components/switch/test/switch.visual-diff.js
GeorgePlukov: Nice catch, I didn't double check that the focus tests were still working (it looks like they aren't). I'll take a look at the forceFocusVisible helper

components/switch/switch-mixin.js
GeorgePlukov: So I ended up noticing that the hover effect should be applied to the inner div when the container div is being hovered (that is the div listening for the click event) so the class changed slightly but is still necessary. I was able to remove the :focus line as the switch is not selectable via the keyboard anymore so we don't have to worry about the classes being applied

components/switch/switch-mixin.js
GeorgePlukov: I also ended up getting the `stylelint` extension, very helpful thanks

dbatiste (approved):

PR #1056, opened 2021-01-22
Updates for API compatibility.

dbatiste:
This PR makes some slight changes and additions to the API of the `d2l-input-textarea` in order to enable seamless transition of the [existing Polymer component](https://github.com/BrightspaceUI/inputs/blob/master/d2l-input-textarea.js) that is used in a variety of places.  This will enable us to completely replace that code with an import for this component.  To summarize:

* change `min-lines` and `max-lines` to `rows` and `max-rows`
* add `no-border` and `no-padding` options 
* add `select` method to select underlying `textarea` text
* add back `aria-label` fallback since all existing usages expect it


components/inputs/docs/input-textarea.md
```diff
@@ -23,12 +23,14 @@ The `<d2l-input-textarea>` is a wrapper around the native `<textarea>` element t
 | `disabled` | Boolean | Disables the `textarea` |
 | `label` | String, required | Label for the `textarea` |
 | `label-hidden` | Boolean | Hides the label visually (moves it to the `textarea`'s `aria-label` attribute) |
-| `max-lines` | Number, default: 11 | Maximum number of lines before scrolling. Less than 1 allows `textarea` to grow infinitely. |
-| `min-lines` | Number, default: 5 | Minimum number of lines. If `min-lines` and `max-lines` are equal then auto-grow will be disabled. |
+| `max-rows` | Number, default: 11 | Maximum number of lines before scrolling. Less than 1 allows `textarea` to grow infinitely. |
```
margaree, line 26: nit: should the description be changed to refer to rows instead of lines?

components/inputs/input-textarea.js
```diff
@@ -224,6 +250,9 @@ class InputTextArea extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))
 		if (this.label && this.labelHidden) {
 			return this.label;
 		}
+		if (this.hasAttribute('aria-label')) {
```
margaree, line 253: Should this be added to the readme?

margaree (approved):

components/inputs/docs/input-textarea.md
dbatiste: sure yep... that would avoid some confusion.

components/inputs/input-textarea.js
dbatiste: We could... although I figured I'd leave it out though since we actually want people to use `label` and `label-hidden`.  This code is really just so existing usages don't break a11y.

dbatiste: I'll add a comment here though so it's clear why this is here.

margaree: That makes sense to me. Sounds good

PR #1070, opened 2021-01-26
input-date: use resizeobserver for getting hiddenContentWidth

margaree:
This was noticed in activities where the input-date was unnecessarily wide. The input is initially hidden and so the hidden divs here that are used to calculate the maxWidth don't actually have widths themselves when `firstUpdated` is run. I moved this calculation into a ResizeObserver which seems to solve this issue.

In fixing this I also noticed that the info tooltip was showing up when the input-date was disabled so I hid that.

### DIFF
--- a/components/inputs/input-date.js
+++ b/components/inputs/input-date.js
@@ -134,7 +134,6 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		this._dropdownOpened = false;
 		this._dropdownFirstOpened = false;
 		this._formattedValue = '';
-		this._hiddenContentResizeObserver = null;
 		this._hiddenContentWidth = '8rem';
 		this._inputId = getUniqueId();
 		this._inputTextFocusMouseup = false;
@@ -144,15 +143,6 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		this._shownValue = '';
 
 		this._dateTimeDescriptor = getDateTimeDescriptorShared();
-	}
-
-	disconnectedCallback() {
-		super.disconnectedCallback();
-
-		if (this._hiddenContentResizeObserver) {
-			this._hiddenContentResizeObserver.disconnect();
-			this._hiddenContentResizeObserver = null;
-		}
 	}
 
 	async firstUpdated(changedProperties) {
--- a/components/inputs/input-date.js
+++ b/components/inputs/input-date.js
@@ -13,6 +13,7 @@ import { FormElementMixin } from '../form/form-element-mixin.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
+import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 import { styleMap } from 'lit-html/directives/style-map.js';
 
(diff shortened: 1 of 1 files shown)


components/inputs/input-date.js
```diff
@@ -165,8 +176,14 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
 		await (document.fonts ? document.fonts.ready : Promise.resolve());
-		const width = Math.ceil(parseFloat(getComputedStyle(this.shadowRoot.querySelector('.d2l-input-date-hidden-content')).getPropertyValue('width')));
-		this._hiddenContentWidth = `${width}px`;
+
+		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-date-hidden-content');
+		this._hiddenContentResizeObserver = this._hiddenContentResizeObserver || new ResizeObserver(() => {
+			const width = Math.ceil(parseFloat(getComputedStyle(hiddenContent).getPropertyValue('width')));
+			this._hiddenContentWidth = `${width}px`;
+		});
+		this._hiddenContentResizeObserver.disconnect();
```
dlockhart, line 185: Did you intend to disconnect right away?

margaree: That's what I had seen done in other places but looking closer they were handling the case where the observer could exist already so it does seem unnecessary here.

components/inputs/input-date.js
```diff
@@ -144,6 +146,15 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		this._dateTimeDescriptor = getDateTimeDescriptorShared();
 	}
 
+	disconnectedCallback() {
```
margaree, line 149: @dbatiste I noticed that tabs also uses a ResizeObserver in firstUpdated and then doesn't use disconnectedCallback. Is this piece here necessary or can I remove it?

dbatiste: So I don't _think_ so.  From a garbage collection perspective, the element and the handler being observed are all internal to the component, so when the component is GC'd I think the observer will be too.  

Aside from GC, functionally, `disconnectedCallback` doesn't necessarily mean that the component is going to be thrown away... i.e. the consumer may reattach it to the DOM.  In such case, there would need to be code in the `connectedCallback` to setup the observer again, assuming it was disconnected in the `disconnectedCallback`, except when `connectedCallback` is first run, the shadowDOM won't exist. The best an cleanest place to hook up listeners and observers is indeed `firstUpdated`, leaving them intact regardless of whether how many times the element is disconnected and reconnected.

margaree: Thanks for the detailed explanation. It sounds like it's reasonably safe to remove the `disconnectedCallback` then.

margaree:
Does this seem good as is or should I take a different approach?

dlockhart (approved):
I think this is good -- like we don't want `ResizeObserver`s all over the place if we can avoid them, but this seems like a good use of one.

dbatiste:
> Does this seem good as is or should I take a different approach?

So the idea is that the resize observer is trigger when it becomes visible?  If so, might want to check Safari if you haven't already.  I can't remember whether Safari calls ResizeObserver when visibility changes.  I vaguely recall some browser differences in how ResizeObserver is called when visibility changes, or maybe that was IntersectionObserver... 🤔

margaree:
So what's happening in the activities case is that when `firstUpdated` is initially called the component is hidden and so the div that is used to calculate the max-width is returning a width of `auto`. Then when the component becomes visible that div ends up getting a width which triggers the `ResizeObserver`. Thanks for the heads up about Safari - I tested it there and it's working.

dbatiste (approved):

PR #1072, opened 2021-01-27
fix: support for RTL in skeleton paragraphs (fixes #1069)

### LINKED ISSUE #1069: d2l-skeletize-paragraph does not support RTL
None of the paragraph skeletons currently supports RTL.

image

dlockhart:
I added RTL tests for all the skeleton stuff as well, so expect a bunch of new diffs. I renamed the existing LTR visual diff files, so they should be unchanged.


components/skeleton/skeleton-mixin.js
```diff
@@ -72,6 +72,12 @@ export const skeletonStyles = css`
 		content: '\\A \\A \\A \\A';
 		white-space: pre;
 	}
+	:host([skeleton][dir="rtl"]) .d2l-skeletize-paragraph-2,
+	:host([skeleton][dir="rtl"]) .d2l-skeletize-paragraph-3,
+	:host([skeleton][dir="rtl"]) .d2l-skeletize-paragraph-5 {
```
dlockhart, line 77: I ran this through a CSS autoprefixer with our current support settings and the `-webkit-*` prefix isn't needed.

components/skeleton/skeleton-mixin.js
```diff
@@ -131,7 +138,7 @@ export const skeletonStyles = css`
 	}
 `;
 
-export const SkeletonMixin = superclass => class extends superclass {
+export const SkeletonMixin = superclass => class extends RtlMixin(superclass) {
```
dlockhart, line 141: Alright this should fix the initial issues. I was testing it in the demo container which puts `dir="rtl"` right on the element. Without the `RtlMixin`, it wasn't picking up a `dir="rtl"` on the document.

Since having someone use `SkeletonMixin` AND `RtlMixin` -- even within our own components -- is quite common, I've added the de-duper to `RtlMixin`.

dbatiste (approved):

components/skeleton/test/skeleton.visual-diff.html
```diff
@@ -13,6 +13,10 @@
 			import '../demo/skeleton-test-paragraph.js';
 			import '../demo/skeleton-test-width.js';
 		</script>
+		<script>
```
dbatiste, line 16: Maybe try moving this before the imports module.

svanherk (approved):
Looks good to me

PR #1074, opened 2021-01-27
change top position for d2l-dialog on mobile

AKobets:
On mobile devices the d2l-dialog has default top offset of 42px. 
In some cases the bottom of the dialog is cutoff by device borders.
Suggested: use custom css property to pass top value. ( or use by default `top:0px`)
 See  screenshot and  rally DE42182   
Cutoff+Browse+AcivityLibrary+on+Android

Here is screenshot for suggested change: 
image

### DIFF
--- a/components/dialog/dialog.js
+++ b/components/dialog/dialog.js
@@ -68,9 +68,9 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 			@media (max-width: 615px), (max-height: 420px) and (max-width: 900px) {
 
 				.d2l-dialog-outer {
-					height: calc(100vh - 42px) !important;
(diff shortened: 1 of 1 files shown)


dlockhart:
I'm hoping this is something we can just fix globally without the need for an offset... just seems like a bug with our implementation?

AKobets:
> I'm hoping this is something we can just fix globally without the need for an offset... just seems like a bug with our implementation?
@dlockhart do you mean that it could be just changed to 
```
  @media (max-width: 615px), (max-height: 420px) and (max-width: 900px) {

				.d2l-dialog-outer {
					height: 100vh !important;
					margin: 0 !important;
					top: 0x;
					width: 100vw !important;
				}
...
```
I was afraid of some regression.

dlockhart:
Closing in favour of a more global solution in #1079 -- thanks for the bug report!

PR #1075, opened 2021-01-27
Move loadMathJax helper into its own module.

dbatiste:
This PR moves the `loadMathJax` helper into its own module so that it can be imported and called separately from `d2l-html-block`.  There will be additional changes to `loadMathJax` to support configuration options required for the Equation Editors.


helpers/mathJax.js
```diff
@@ -0,0 +1,124 @@
...
+		script.onload = resolve;
+		script.src = 'https://s.brightspace.com/lib/mathjax/3.1.2/mml-chtml.js';
+		document.head.appendChild(script);
+	});
+
+	return mathJaxLoaded;
+
+}
+
+window.D2L = window.D2L || {};
```
dbatiste, line 121: This is the only bit that's added here.  This is needed to expose the loader, for example in the context of Content Topics.

Another option I am contemplating is having the consumer use a dynamic import instead.  For example:
```javascript
import('..the-location-of/mathJax.js').then(module => {
    module.loadMathJax();
});
```
But then it requires the consumer to use a dynamic import, which I'm not sure if we'd encounter some FOUC.

dbatiste: Aside, this would be a technique that _could_ be used to use some of our other helpers in the monolith without having to expose them explicitly on the window in BSI.

dlockhart: Right, so typically we'd do this bit inside BSI, [like this](https://github.com/Brightspace/brightspace-integration/blob/master/web-components/bsi-unbundled.js#L16). I think I'd prefer we stick with one technique or the other.

Doesn't this approach still require that BSI import it (to make sure it's on the page).?

dbatiste: So in this case the BSI option won't be possible since it's not included in Content Topics.

dbatiste: But yeah, there will be code in the monolith that wants to call this too.

dlockhart: Right... so you'll need to expose this as a module similar to our other components and load that module on the page(s) that need it.

dbatiste: Ok, I've done a bunch of testing with this including different browsers, both Dev and Prod configurations of BSI and it seems to be working as expected for Content Topics.  I can move the `window.D2L.MathJax` bit to a module in BSI.

dlockhart (approved):

PR #1085, opened 2021-01-29
fix: let width of confirmation dialog be specified.

mbosiljevac:
I want to make the confirm dialog wider. Let me know if there was a reason why we had the max width/some reason why I shouldn't make this change!

### DIFF
--- a/components/dialog/README.md
+++ b/components/dialog/README.md
@@ -115,6 +115,7 @@ document.querySelector('#open').addEventListener('click', () => {
 | `text` | String, required | The required text content for the confirmation dialog |
 | `opened` | Boolean | Whether or not the dialog is open |
 | `title-text` | String | The optional title for the confirmation dialog |
+| `width` | Number, default: `420` | The preferred width (unit-less) for the dialog |
 
 **Events:**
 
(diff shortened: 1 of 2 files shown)


dbatiste:
Yeah the current `max-width` (and not specifying an explicit `width`) is intentional. If we allow explicit widths for `d2l-dialog-confirm` then we risk breaking the responsive design.  Instead, `d2l-dialog-confirm` imposes this max-width so it will work on smaller devices, and more generally, for smaller viewports. I think `max-width: 420px` was chosen both to work on small devices as well as to provide a reasonable aesthetic for anticipated usage.

Aside, `d2l-dialog` internally calculates an explicit width respecting the space available.

We can potentially look at adding a media-query for a different max-width when the viewport is larger, but I think we'd need this to be defined by Design.

mbosiljevac:
> Yeah the current `max-width` (and not specifying an explicit `width`) is intentional. If we allow explicit widths for `d2l-dialog-confirm` then we risk breaking the responsive design. Instead, `d2l-dialog-confirm` imposes this max-width so it will work on smaller devices, and more generally, for smaller viewports. I think `max-width: 420px` was chosen both to work on small devices as well as to provide a reasonable aesthetic for anticipated usage.
> 
> Aside, `d2l-dialog` internally calculates an explicit width respecting the space available.
> 
> We can potentially look at adding a media-query for a different max-width when the viewport is larger, but I think we'd need this to be defined by Design.

hmm ok thanks for the context! I'll close this PR and have a bigger chat with design about how to deal with my specific situation!

dbatiste:
> hmm ok thanks for the context! I'll close this PR and have a bigger chat with design about how to deal with my specific situation!

Sounds great. If necessary, feel free to look myself or one of the other from Gaudi in if there is a desire to make some changes here.

PR #1086, opened 2021-01-29
list-item skeleton updates

dbatiste:
This PR adds a bit more skeleton behavior to the list-items.

* hides the button primary action layer when skeleton is showing
* skeletizes the selectable list-item checkbox


components/list/list-item-link-mixin.js
```diff
@@ -29,10 +29,10 @@ export const ListItemLinkMixin = superclass => class extends ListItemMixin(super
 				outline: none;
 				width: 100%;
 			}
-			:host([skeleton]) a[href] {
```
dbatiste, line 32: This selector was unnecessarily qualified with the `href` attribute.

dlockhart (approved):

PR #1091, opened 2021-02-01
fix: list-item-link gap caused by border

bearfriend:
The list-item border causes links not to cover the entire element, leaving 1px at the top where the element can be hovered/clicked without engaging the link.

list-item-link-coverage-2

### DIFF
--- a/components/list/list-item-link-mixin.js
+++ b/components/list/list-item-link-mixin.js
@@ -19,16 +19,14 @@ export const ListItemLinkMixin = superclass => class extends ListItemMixin(super
 			:host([action-href]:not([action-href=""])) {
 				--d2l-list-item-content-text-color: var(--d2l-color-celestine);
 			}
-			:host([action-href]:not([action-href=""]):not([skeleton])) d2l-list-item-generic-layout.d2l-focusing,
-			:host([action-href]:not([action-href=""]):not([skeleton])) d2l-list-item-generic-layout.d2l-hovering {
+			:host([action-href]:not([action-href=""]):not([skeleton])) [slot="content"].d2l-focusing,
+			:host([action-href]:not([action-href=""]):not([skeleton])) [slot="content"].d2l-hovering {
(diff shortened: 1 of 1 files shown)


dbatiste:
Hey Danny, you'll want to rebase to pick up a fix that will avoid the Korean language issue with visual-diff workflow dependency (that cropped up on Friday for some unknown reason).

Aside from that, re my comment (https://github.com/BrightspaceHypermediaComponents/activities/pull/1429#discussion_r567942981), would that event avoid the need for this fix?  It seems like the need for this fix is arising due to a slight difference between the overall list-item dimensions and the dimensions of the anchor.  If we do still need this, I think it's fine, but if not it would be nice to avoid the negative margin shenanigans.

dbatiste:
Oh - the BrowserStack issue is a known flakey test.  This can just be rerun.

bearfriend:
Ahh, I keep forgetting I have this forked. Rebasing from `origin` is not doing me any good 🤦 

Still necessary for the hover discrepency. Looks like a `box-sizing` fix is broken by the shadow boundary. We could use an inset `box-shadow` instead of border but I think I prefer this. I'm fine with whatever though.

bearfriend:
Links now fire [`d2l-list-item-link-click`](https://github.com/BrightspaceUI/core/pull/1094) and this PR moves the hover state to the content `slot`, which was already set up to support it but was unused. The 1px border/gap is now consistently _outside_ both instead of _inside_ as I originally intended for this PR.

dbatiste:
Is it possible to see this issue in the demo page?  I'm trying to reproduce it in order to tell if this is also an issue for `d2l-list-item-button` (mixin), since it happens to also target `d2l-list-item-generic-layout`.  Unable to see this so I'm not sure.

bearfriend:
@dbatiste Yes, it's reproducible on all action, button, and selectable list-items. I'm closing this as the context and solution has changed significantly and will open a new PR that fixes all of them.

PR #1093, opened 2021-02-01
dropdown: use opener's parent for running getBoundingAncestor

margaree:
This is to fix the case with html editor where the dropdowns are opening in the incorrect direction. I found that the values used in the left position calculation here https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-content-mixin.js#L591 seemed really off (e.g., target width was 44, space around was -20) and that if I set bounded to false then things were fine. I also noticed that `getBoundingAncestor` for these cases was just returning the opener (`d2l-button-subtle`) since it was going into this case https://github.com/BrightspaceUI/core/blob/master/helpers/dom.js#L17. This is one of those fixes where it fixes the case at hand and the visual diffs locally were fine but I don't know if it was like this for a reason.

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -472,7 +472,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		this._width = null;
 
 		const openerPosition = window.getComputedStyle(opener, null).getPropertyValue('position');
-		const boundingContainer = getBoundingAncestor(target);
+		const boundingContainer = getBoundingAncestor(target.parentNode);
 		const boundingContainerRect = boundingContainer.getBoundingClientRect();
 		const scrollHeight = boundingContainer.scrollHeight;
 


dbatiste:
I think this looks good.

I'm not sure I follow why Allan's change seemed to affect this, but the fix relates to the Sept 20th change that I made for dropdowns constrained in dialogs and other small spaces (which would have made it into an earlier release that I think doesn't have this issue.

margaree:
I think it's because Allan's change adjusted this calculation https://github.com/BrightspaceUI/core/pull/962/files#diff-ae18b665ccdd284bba434d088bd1a535058b44d952092836cc7ac9c8dc6fd767R581 so that it would now use that Math.max calculation, which used the values from the `d2l-button-subtle` and resulted in an undesirable left.

dbatiste:
> I think it's because Allan's change adjusted this calculation https://github.com/BrightspaceUI/core/pull/962/files#diff-ae18b665ccdd284bba434d088bd1a535058b44d952092836cc7ac9c8dc6fd767R581 so that it would now use that Math.max calculation, which used the values from the `d2l-button-subtle` (since that was the bounding container) and resulted in an undesirable left.

Cool.  Yep, makes sense now.

dbatiste (approved):

PR #1094, opened 2021-02-01
Add d2l-list-item-link-click event so consumers can use (ex. telemetry).

dbatiste:
This PR surfaces a custom event for the primary navigation action on list items, so that consumers can wire-up for things like telemetry.  Without this, consumers are forced to wire-up click and key handlers on the list-item, which doesn't always translate to navigation when there are other things the user can interact with, such as selection, drag & drop, secondary actions, etc.


bearfriend (approved):
:shipit:

dlockhart (approved):

PR #1095, opened 2021-02-01
20.21.2: dropdown: use opener's parent for running getBoundingAncestor

margaree:


dbatiste (approved):
Looks like you are going to need that apt-get fix for the visual-diff workflow.

PR #1098, opened 2021-02-02
Make textarea compatible with Activity Feed 

mostlyfabulous:
### Currently in AF:

- Clicking the bin icon 🗑️ does not clear the text area
- Posting a comment does not clear the comment's text area either and the text content remains

### Context

I'm not entirely sure why the value of the text area is not being set when we fire off our events in AF. I've traced it down to potentially:
https://github.com/BrightspaceUI/core/blob/f127a2e2f84c423714d88df202c91fa8f2458e65/components/inputs/input-textarea.js#L209

I might be wrong in expecting `formValue` to update the `value` of the `textarea` but from debugging in dev tools, it seems all the right handlers are being called and the `value` property in `d2l-input-textarea` is correctly set, except it does not update the underlying `textarea`.

Trello card: https://trello.com/c/Uea0RZ8H/304-af-comment-d2l-text-area-display-is-wrong-not-functioning-normally

[US124446](https://rally1.rallydev.com/#/29180338367d/dashboard?detail=%2Fuserstory%2F489806220980&fdp=true?fdp=true): Regression test AF, FACE after `d2l-input-textarea` refactoring


dlockhart:
This might be a similar issue to how `<d2l-input-checkbox>` needs to bind to the native input's property -- the attribute is only used to initially populate the value, after that it's the property.

Could you try to fix with `<textarea value="${this.value}">` in the render code?

dbatiste:
> This might be a similar issue to how `<d2l-input-checkbox>` needs to bind to the native input's property -- the attribute is only used to initially populate the value, after that it's the property.
> 
> Could you try to fix with `<textarea value="${this.value}">` in the render code?

The `value` attribute isn't valid for element `textarea`.  The `value` property is though `<textarea .value="${this.value}">`.  I was just doing some testing though, and updating the `value` property on `d2l-input-textarea` seems to be working as expected?  @mostlyfabulous , is that not what you are trying to update?

Aside, the `textarea` property in this component here needs to be removed - it is merely here temporarily so that rubrics would not break while we were cutting over from the Polymer version.

dlockhart:
Sorry, value property via `.value="${this.value}"` is what I meant to write. 🤦

mostlyfabulous:
@dbatiste Yes, that's right.
@dlockhart `.value="${this.value}"`  works in AF!

dbatiste:
I'm still uncertain as to what's happening here.  When I set `value` property of `d2l-input-textarea` (after initial render) I see both the value change in the input, and also when I check the `value` property on the underlying `textarea` it is also updated as expected.

mostlyfabulous:
@dlockhart I noticed that behaviour as well. Although the `value` on `textarea` was also correct, it didn't update the contents inside the `textarea`. A wild guess is redux-polymer in AF isn't quite working with this refactored component.

dbatiste:
Looks like cursor moved to the end as a result of setting the property.  Ideally we match the behaviour of the unwrapped native textarea in this regard.

dlockhart:
I was able to reproduce what I think is happening in a codepen:
https://codepen.io/dlockhart/pen/vYyOrGa

If you immediately click the button to change the value, the rendered value is updated properly.

But if you first edit the value, then click the button, it does nothing.

dbatiste:
Awesome.  That's interesting behaviour. 

Looks like cursor moved to the end as a result of setting the property.  Ideally we match the behaviour of the unwrapped native textarea in this regard.  This is what I find with native `textarea`:

* when initialized with a value `<textarea>stuff</textarea>`, the caret is placed at the beginning when focusing
* when initialized with a value `<textarea>stuff</textarea>` and setting the `value` property, and then focusing, the cret is indeed placed at the end instead.

dbatiste:
I tested quickly with VoiceOver and it announces the value regardless, and it also announces "insertion at beginning of text" / "insertion at end of text" so the caret position might not matter too much.

dbatiste:
A consequence to this is that if the value is initially long (such that there is scrolling), focusing causing the textarea to scroll when it normally would not have.  If we want to match the behaviour we can do something along these lines:

FirstUpdated
```javascript
    this._valueInitialized = true;
}
```

In render
```javascript
const textarea = html`
    ...
    <textarea 
        ...
        .value="${ifDefined(this._valueInitialized ? this.value : undefined)}">${this.value}</textarea>
    ...
`;
```

mostlyfabulous:
@dlockhart 
That seems reasonable 👍

dbatiste (approved):

PR #1099, opened 2021-02-02
fix: list-item hover gap caused by border

bearfriend:
All hover effects should _not_ be triggered by hovering `d2l-list-item-generic-layout` itself, but by hovering its child elements (inside its 1px borders).

This prevents the background color from changing before any actionable child elements are hovered.

Fix is for all action, button, and selectable list-items.

(green background color is just for GIF contrast)
hover-gap

### DIFF
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -596,7 +596,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 
 	_renderDragTarget(templateMethod) {
 		templateMethod = templateMethod || (dragTarget => dragTarget);
-		return this.draggable && !this._keyboardActive ? templateMethod(html`
+		return this.draggable && !this._keyboardActive ? templateMethod.call(this, html`
 			<div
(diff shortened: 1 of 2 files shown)


components/list/list-item-drag-drop-mixin.js
```diff
@@ -606,7 +596,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 
 	_renderDragTarget(templateMethod) {
 		templateMethod = templateMethod || (dragTarget => dragTarget);
-		return this.draggable && !this._keyboardActive ? templateMethod(html`
+		return this.draggable && !this._keyboardActive ? templateMethod.call(this, html`
```
dbatiste, line 599: Was this causing an issue?

bearfriend: Just necessary to add the handlers to the outside control action.

dbatiste: I thought `this` would have been the component in either context.  🤷‍♂️

dbatiste (approved):

PR #1101, opened 2021-02-02
Issue 1083: fix flaky input-date-time test

margaree:
To fix the `d2l-input-date-time min and max value ✖ should set correct min and max on d2l-input-date (expected '2018-08-27' to equal '2018-08-26', Edge, Safari)` test (the other two haven't seemed to fail lately so I'll just keep an eye on them).

I've rerun CI 3 times and so far so good.

My theory is that because this is the first test that's being flaky, the `documentLocaleSettings.timezone.identifier = 'America/Toronto';` hasn't always "kicked in" yet. I tried adding an event listener for the localize event but that wasn't quite working (wasn't getting called) so I'm trying out a timeout.


margaree:
I've now rerun ~5-6 times. There was one was sauce failed but it looked to be a connection failure ( `Disconnectedreconnect failed before timeout of 50000ms (ping timeout)`) and I didn't see evidence of a test failures in the logs.

dlockhart (approved):
Cool -- always sucks when a timeout fixes things, but what you gonna do. Might be worth putting a comment there to remind us later that this was a flaky test.

margaree:
Okay this is good to go for a re-review. Thoughts on the script for the ResizeObserver error handling and if it seems good, should it be added to the initializer?

dlockhart (approved):
This is great! Yeah seeing as other people's tests could include our own components that may now be firing these errors, adding it to the initializer makes sense to me.

PR #1106, opened 2021-02-04
Change switch off background color to Regolith

GeorgePlukov:
We missed this during the last switch component style updates. Just changing the background color to a slight grey when the switch is off


components/switch/switch-mixin.js
```diff
@@ -68,6 +68,9 @@ export const SwitchMixin = superclass => class extends RtlMixin(FocusVisiblePoly
 			:host([on]) .d2l-switch-inner {
 				background-color: var(--d2l-color-celestine-plus-2);
 			}
+			:host(:not([on])) .d2l-switch-inner {
```
dbatiste, line 71: Could this be moved up to the `.d2l-switch-inner` selector (i.e. to avoid the `[on]`, `:not()`, and `:host` selector checks)?

dbatiste (approved):
New goldens look good.  The BrowserStack failure looks to be a date/time input flake.

PR #1108, opened 2021-02-05
input-time: Default change

margaree:
Sorry for the large PR - all the components are intertwined with their defaults and I think there could be some weird/incorrect behaviour if I don't do them all.

Changes:
- `input-time`: 
    - default when `default-value` is undefined is the current time rounded up to the next time interval
    - `startOfDay` is now 12:01 AM unless `enforce-time-intervals` is true in which case it is 12 AM (this also modifies the first element of the dropdown)
- `input-time-range`: start default is now rounded up to the next time interval, end is the following interval (same as before)
- `input-date-time`: no date default (same as before), time defaults to `startOfDay`
- `input-date-time-range`: no date default (same as before), start time defaults to `startOfDay`, end time defaults to `endOfDay`

Notes:
- This brings timezones into `input-time`. I kept the actual value as the localized since that's what it was always implied to be and then anyone using it will continue on as before.
- `startOfDay` - there was concern around changing to 12:01 AM. I think this change is safe. It isn't ever stored as a value, it is just used as a default value that a user will change. It's also only used by activities and not in existing LMS code. Let me know if there could be issues there that I'm missing.

To Do:
- [x] clarify with Jeff on if current time is exactly on an interval (e.g., 12 pm) then should default be now or +1 time interval (e.g., 12:30 pm) -> +1 interval


components/inputs/input-time.js
```diff
@@ -37,23 +39,23 @@ export function getIntervalNumber(size) {
...
+			const today = getToday();
+			const date = new Date(today.year, today.month - 1, today.date, today.hours, today.minutes, 0);
+			return getTimeAtInterval(timeInterval, date);
+		}
 		default:
 			return getDateFromISOTime(time);
 	}
 }
 
-export function getFormattedDefaultTime(defaultValue) {
```
margaree, line 52: Moved to `input-date-time`

components/inputs/input-time.js
```diff
@@ -63,12 +65,19 @@ export function getTimeAtInterval(timeInterval, time) {
 	return time;
 }
 
-function initIntervals(size) {
-	if (!INTERVALS.has(size)) {
+function initIntervals(size, enforceTimeIntervals) {
```
margaree, line 68: The render code was getting too messy with the end and start dropdown elements depending on `enforceTimeIntervals` so I moved it into here

components/inputs/test/input-date-time-range.visual-diff.js
```diff
@@ -223,8 +223,8 @@ describe('d2l-input-date-time-range', () => {
 				describe(testCase.name, () => {
 					before(async() => {
 						await page.$eval('#min-max', (elem) => elem.blur());
-						await changeInnerInputTextDate(page, '#min-max', startDateSelector, testCase.startDate);
-						await changeInnerInputTextDate(page, '#min-max', endDateSelector, testCase.endDate);
+						await changeInnerInputDateTime(page, '#min-max', startDateSelector, testCase.startDate);
```
margaree, line 226: This test was to trigger error handling but since the times were no longer the same (changed to startOfDay and endOfDay) there was no longer an error so the time has to be tweaked to be the same as well

dlockhart (approved):
LGTM!

dlockhart (approved):

PR #1116, opened 2021-02-05
fix: validate numeric inputs whenever the value changes

dlockhart:
The parameter here that's changing is whether or not to "show new errors". I think the reasoning for passing in `false` here was to avoid validating partial input while the user was still typing, instead waiting for the `blur` event to fire. But the numeric input only updates `value` on-blur anyway, not as the user types. @krshnpatel not sure if you remember more details here?

The problem with passing `false` is that validation is skipped if `value` is set programatically using JavaScript. This happens a lot in the MVC/legacy frameworks where we have wrappers around the component with their own APIs. I think we _do_ want to validate whenever `value` changes regardless.

The outcome of this change are:
* For `required` values, if a value isn't provided validation doesn't occur since `value` doesn't ever change -- this is good, as we don't want required inputs to be red initially until the user has interacted with them.
* For all other types of validation (min/max/etc), if the initial value is invalid, validation will occur and the input will be red. I think this is OK since the user will see that and know to fix it, instead of waiting for form submission time.
* When interacting with the input and causing it to be valid or invalid, the check occurs on-blur as before.
* When setting the value programatically, the validation occurs immediately (expected change)

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -140,7 +140,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 					rangeUnderflow: rangeUnderflowCondition,
 					rangeOverflow: rangeOverflowCondition
 				});
-				this.requestValidate(false);
+				this.requestValidate(true);
 			}
 		});
 	}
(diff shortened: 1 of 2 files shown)


krshnpatel (approved):
Yeep I remember this 😅

margaree (approved):

Issue #1083, opened 2021-01-28
Investigate flaky input-date* unit tests


margaree:
Tests to investigate:
- d2l-input-date-time-range > utility > getShiftedEndDateTime in America/Toronto with localized true > date change > ✖ should return correctly backward shifted end date (Actual 2020-10-26T03:00:00.000 expected 2020-10-26T04:00:00.000, Firefox, Safari, Chrome)
- d2l-input-date-time-range > utility > getShiftedEndDateTime in Australia/Eucla with localized true > date change > ✖ should return correctly backward shifted end date (Actual 2020-10-26T03:00:00.000 expected 2020-10-26T04:00:00.000, Firefox, Safari, Chrome)
- d2l-input-date-time min and max value ✖ should set correct min and max on d2l-input-date (expected '2018-08-27' to equal '2018-08-26', Edge, Safari)

margaree:
PR https://github.com/BrightspaceUI/core/pull/1101 should resolve `d2l-input-date-time min and max value ✖ should set correct min and max on d2l-input-date (expected '2018-08-27' to equal '2018-08-26', Edge, Safari)`. The other 2 will be resolved separately.

PR #1120, opened 2021-02-08
input-date-time-range: attempt to fix flaky tests

margaree:
Opening as draft in order to rerun the tests several times.

Should fix https://github.com/BrightspaceUI/core/issues/1083.

This is likely cause by the same problem as https://github.com/BrightspaceUI/core/pull/1101, which is the timezone not yet being set, and there isn't an event to listen to that I'm aware of so a timeout should help with this.


margaree:
I have rerun the tests ~6 times and these tests have not flaked

dlockhart (approved):
Nice!

PR #1121, opened 2021-02-08
lint: Bump eslint-config-brightspace version

svanherk:
Use latest `eslint-config-brightspace` version.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,9 +3,6 @@
   "globals": {
     "Prism": false
   },
-  "rules": {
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -3,9 +3,6 @@
   "globals": {
     "Prism": false
   },
-  "rules": {
```
svanherk, line 6: This was the rule that was added in `0.12.0`, so we can remove it here

dlockhart (approved):
Nice, thanks!

PR #1128, opened 2021-02-11
[ignore] visual-diff: fix flaky dialog and dropdown-content tests

margaree:


dbatiste (approved):

components/dialog/test/dialog.visual-diff.js
```diff
@@ -46,23 +46,30 @@ describe('d2l-dialog', () => {
 						await page.setViewport({ width: info.viewport.width, height: info.viewport.height, deviceScaleFactor: 2 });
 					});
 
-					it('opened', async function() {
-						if (info.category === 'short-narrow') return; // TODO: remove this; skipping for now due to flaking
-						await helper.open(page, '#dialog');
-						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+					it('opened', function() {
+						return Promise.all([
```
dbatiste, line 50: This parallelizes these tasks, but the screenshot task needs to happen after.  This might luckily work but it doesn't look like it would always work.

margaree: Ah I was under the impression that the tasks are happening in serial, though upon further searching it seems a bit vague and unknown. It has been working consistently for these cases for several runs but I'll see about reorganizing to be more sure.

components/dialog/test/dialog.visual-diff.js
```diff
@@ -46,23 +46,38 @@ describe('d2l-dialog', () => {
 						await page.setViewport({ width: info.viewport.width, height: info.viewport.height, deviceScaleFactor: 2 });
 					});
 
-					it('opened', async function() {
-						if (info.category === 'short-narrow') return; // TODO: remove this; skipping for now due to flaking
-						await helper.open(page, '#dialog');
-						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+					it('opened', function() {
+						return new Promise((resolve) => {
```
margaree, line 50: Any ideas on a better way to write these? It seems like it could be simplified. I tried with `reduce` but couldn't quite get it.

dlockhart: What about something like...

```javascript
    helper.open(page, '#dialog').then(() => {
        return page.waitForSelector('#bottom', { visible: true });
    }).then(() => {
        return visualDiff.screenshotAndCompare(page, this.test.fullTitle()
    }).then(() => resolve);
});
```

dbatiste: I think think it could be written a couple different ways:

```javascript
    await helper.open(page, '#dialog');
    await page.waitForSelector('#bottom', { visible: true });
});
```

However, specific to what's currently here, it is worth mentioning that all `async` methods automatically return a `Promise`, and `await` is synonyms with `then`.  So what's currently written could also be rewritten as:

```javascript
it('opened', function() {
    return helper.open(page, '#dialog').then(() => {
        return page.waitForSelector('#bottom', { visible: true });
    }).then(() => {
        return visualDiff.screenshotAndCompare(page, this.test.fullTitle());
    });
});
```

There are various permutations that combine `Promise`, `then`, `async`, `await` that could be written, but normally I would opt for keeping consistent with `async`/`await` or `Promise`/`then`/`catch`/`finally`.  There are some differences however in the way exceptions are handled, and also, if one wants to wait for multiple promises, the Promise API provides other very helpful methods such as `all`, `race`, `allSettled`.

On this topic, something I see people do sometimes which is pointless is:

```javascript
const someMethod = async() => {
    return await someOtherAsyncMethod();
};
// or
const someMethod = async() => {
    await someOtherAsyncMethod();
};

await someMethod();
```

Which could just be:

```javascript
const someMethod = () => {
    return someOtherAsyncMethod();
};

await someMethod();
```

components/dialog/test/dialog.visual-diff.js
```diff
@@ -47,22 +47,32 @@ describe('d2l-dialog', () => {
...
+							return visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+						});
 					});
 
 					it('opened-wide', async function() {
-						await page.$eval('#wideContainer', wideContainer => wideContainer.style.width = '1500px');
-						await helper.open(page, '#dialog');
-						await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
-						await page.$eval('#wideContainer', wideContainer => wideContainer.style.width = 'auto');
+						return Promise.all([
```
dbatiste, line 59: Is the `Promise.all` needed?  Does this work?

```javascript
await page.$eval('#wideContainer', wideContainer => wideContainer.style.width = '1500px');
await helper.open(page, '#dialog');
await page.waitForSelector('#bottom', { visible: true });

await page.$eval('#wideContainer', wideContainer => wideContainer.style.width = 'auto');
```
If we _do_ need the `Promise.all` for some reason, I think it would also be fine to write it like this:
```javascript
await Promise.all([
	page.$eval('#wideContainer', wideContainer => wideContainer.style.width = '1500px'),
	helper.open(page, '#dialog'),
	page.waitForSelector('#bottom', { visible: true })
]);
await page.$eval('#wideContainer', wideContainer => wideContainer.style.width = 'auto');
```
Not a big deal.  The `then` statements just aren't as tidy.

In the end, the key requirement here is that we need to use `page.waitForSelector('#bottom', { visible: true })` for some reason eh?

dlockhart: Worth noting that `Promise.all` does not run in sequence, but in parallel. So there's no guarantee that it'll wait for the first thing to execute before doing the second... which I think is what you need?

margaree: So I think for `opened-wide` it wasn't flaky before and we can go back to how it was with the `await`s though I'll need to test that out (it's hard to know because something will be totally fine running locally but really flaky in the PR). The `Promise.all` seems to be necessary for the `opened` and `rtl` tests (I believe when I had that was when there was consistently no flake) but I'll try to restructure if possible.

margaree:
Closing since this has become a bit convoluted and the dialog piece looks like it could be a separate problem. I'm going to open a new PR for the cursor flake and the `input-date-time-range` timeout that was added.

PR #1139, opened 2021-02-18
dialog: fix top border not appearing when small and resize occurs

margaree:
Ignore for now. I'll see if the visual diffs are no longer flaky and then open as a real PR.

Bug:
When dialog is opened at small width/height where it fills the page (height of page ~<443 px) and dev tools is opened then closed, the footer border is present when it shouldn't be. Hopefully this fix (thanks Dave B) also allows the visual diffs to not be flaky.


components/dialog/dialog-mixin.js
```diff
@@ -407,8 +407,10 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			this._height = this._getHeight();
 			await this.updateComplete;
 		}
-		this._updateOverflow();
-		await this.updateComplete;
+		requestAnimationFrame(async() => {
```
dbatiste, line 410: It looks like it won't make any difference to the current callers, however if we want the resulting `Promise` (returned by `_updateSize`) to still resolve when the update is complete, I think we would need to wrap the rAF in a `Promise` and either `return` it, or `await` it.  None of the callers actually rely on the `Promise` that's returned though... the method is only really marked as `async` in order to use `await` internally.

```javascript
...
...
await new Promise(resolve => {
        this._updateOverflow();
        await this.updateComplete;
        resolve();
    });
});
```

dbatiste (approved):

PR #1146, opened 2021-02-21
Add support for dynamically imported translation resources

bearfriend:
Adds a proper `localize-dynamic` mixin to simplify individual component complexity. Docs have been updated with example usage.

Dependent upon:
~https://github.com/Brightspace/brightspace-integration/pull/3350~
~https://github.com/Brightspace/eslint-config-brightspace/pull/39~

### DIFF
--- a/mixins/localize-dynamic-mixin.js
+++ b/mixins/localize-dynamic-mixin.js
@@ -1,7 +1,9 @@
 import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js';
 import { LocalizeMixin } from './localize-mixin.js';
 
+/*** WARNING! ¡ATENCIÓN! ACHTUNG! Adding to defaultLangs MUST be a BREAKING change with a MAJOR version bump! ***/
 const defaultLangs = ['ar', 'cy-gb', 'cy', 'da', 'de', 'en', 'es-es', 'es', 'fr-ca', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
+/****************************************************************************************************************/
 const fallbackLang = 'en';
 const defaultExportName = 'default';
 
@@ -17,8 +19,8 @@ export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(su
 		for (const lang of [...langs, fallbackLang]) {
 
 			if (supportedLangs.includes(lang)) {
-				const mod = await importFunc(lang);
-				const resources = mod[exportName];
+				const mod = await importFunc(lang).catch(() => {});
+				const resources = mod && mod[exportName];
 
 				if (resources) {
 
--- a/mixins/localize-dynamic-mixin.js
+++ b/mixins/localize-dynamic-mixin.js
@@ -1,0 +1,41 @@
+import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js';
+import { LocalizeMixin } from './localize-mixin.js';
+
+const defaultLangs = ['ar', 'cy-gb', 'cy', 'da', 'de', 'en', 'es-es', 'es', 'fr-ca', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
+const fallbackLang = 'en';
+const defaultExportName = 'default';
+
+export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(superclass) {
+
+	static get config() {
+		return {};
+	}
+
+	static async getLocalizeResources(langs, config) {
+		const { importFunc, osloCollection, supportedLangs = defaultLangs, exportName = defaultExportName } = config;
+
+		for (const lang of [...langs, fallbackLang]) {
+
+			if (supportedLangs.includes(lang)) {
+				const mod = await importFunc(lang);
+				const resources = mod[exportName];
+
+				if (resources) {
+
+					if (osloCollection) {
+						return await getLocalizeOverrideResources(
+							lang,
+							resources,
+							() => osloCollection
+						);
+					}
+
+					return {
+						language: lang,
+						resources
+					};
+				}
+			}
+		}
+	}
+};
--- a/mixins/localize-dynamic-mixin.js
+++ b/mixins/localize-dynamic-mixin.js
@@ -1,11 +1,17 @@
 import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js';
 import { LocalizeMixin } from './localize-mixin.js';
 
-/*** WARNING! ¡ATENCIÓN! ACHTUNG! Adding to defaultLangs MUST be a BREAKING change with a MAJOR version bump! ***/
-const defaultLangs = ['ar', 'cy-gb', 'cy', 'da', 'de', 'en', 'es-es', 'es', 'fr-ca', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
-/****************************************************************************************************************/
 const fallbackLang = 'en';
 const defaultExportName = 'default';
+
+function warnMissingFiles(failures, resolvedLang) {
+	const { path } = failures[0].err.message.match(/(?:https?:\/\/.*?\/)(?<path>.*)\/.*\./).groups;
+	const langs = failures.map(f => f.lang).join('","');
+	console.warn(
+		'WARNING: Unacceptable in production. Consider Manually Retrieved Resources method.\n\n' +
+		`Failed to load translation resources for languages "${langs}" from ${path}.`,
+		`Falling back to "${resolvedLang}". `);
+}
 
 export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(superclass) {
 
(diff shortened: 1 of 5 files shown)


mixins/localize-dynamic-mixin.js
```diff
@@ -0,0 +1,41 @@
+import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js';
+import { LocalizeMixin } from './localize-mixin.js';
+
+const defaultLangs = ['ar', 'cy-gb', 'cy', 'da', 'de', 'en', 'es-es', 'es', 'fr-ca', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
```
dlockhart, line 4: I love that all the default languages are just handled for folks, but I'm worried about what happens during the period of time from when we add a new language here (e.g. `cy` was added recently) and when downstream components actually add translations for that language.

I think what would happen is it would call `await importFunc('cy')` which would 404 and may blow up on the next line? The two alternatives I can think of:

1. We don't have a default list like this -- each implementation **must** provide the list of languages they support. We could maybe have `en` as the default since everyone should at least support that.
2. We somehow handle the 404 by detecting it and moving on to the next language in the list. I don't love having a 404 be a non-error though as users will see it in their consoles and our devs may not notice and never fix the underlying issue.

mixins/localize-mixin.md
```diff
@@ -56,27 +58,57 @@ class MyComponent extends LocalizeStaticMixin(LitElement) {
...
-
 #### Example 2: Dynamically Imported Resources
 
 This approach is better for components with many language resources. By importing them dynamically, only the resources for the requested language are actually fetched and downloaded.
 
 Store your resources in individual files, one for each language:
 ```javascript
 // en.js
-export const val = {
+export default {
```
dlockhart, line 68: I think this is going to be a breaking change for those using the Serge integration. Was this change just ergonomic? Or I guess is this what the `exportName` config value is for?

dlockhart (commented):
This is looking really promising!

mixins/localize-dynamic-mixin.js
bearfriend: Yeah, I was hoping we could just be conservative on adding new languages here. Maybe we wait for serge to send out PRs for the new language and then give people `x` amount of time before merging it in here. Perhaps it's a breaking change each time 😬, though it shouldn't be super frequent. Unless we can be smart and add error handling as a backup as well. With telemetry even?

Otherwise, forcing each component to declare supported languages isn't the end of the world.

mixins/localize-mixin.md
bearfriend: I don't think serge cares. I _believe_ it is just skipping over it with its regex anyway looking for the key/value pairs. I'll confirm with Sean. Just ergonomic and searching this repo I found 143 `export default` and only 30 `export const` in `lang` directories so I figured it was a good default.

dlockhart: Cool yeah agreed, and even better if it doesn't impact Serge.

bearfriend: Confirmed it doesn't care. But they need to be consistent between languages, obviously, which this will enforce as well.

dlockhart: Awesome, thanks for checking that.

mixins/localize-dynamic-mixin.js
```diff
@@ -0,0 +1,43 @@
...
+		return {};
+	}
+
+	static async getLocalizeResources(langs, config) {
+		const { importFunc, osloCollection, supportedLangs = defaultLangs, exportName = defaultExportName } = config;
+
+		for (const lang of [...langs, fallbackLang]) {
+
+			if (supportedLangs.includes(lang)) {
+				const mod = await importFunc(lang).catch(() => {});
```
bearfriend, line 22: Some people find the `await/catch()` mixture objectionable, but it's cleaner than a `try/catch`. This will only supress the javascript errors, not the initial network error.

mixins/localize-dynamic-mixin.js
```diff
@@ -0,0 +1,43 @@
+import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js';
+import { LocalizeMixin } from './localize-mixin.js';
+
+/*** WARNING! ¡ATENCIÓN! ACHTUNG! Adding to defaultLangs MUST be a BREAKING change with a MAJOR version bump! ***/
```
bearfriend, line 4: After sleeping on it, I guess this doesn't really help us given the BSI dependency situation. Maybe it's just better we require it to be explicit.

dlockhart: Yeah. Like we've really tried hard to avoid major version bumps of core, because coordinating that switch across all the repos that get pulled into BSI is a nightmare -- they all need to upgrade together anyway.

bearfriend: Unfortunate, as I was really patting myself on the back for this warning message.

mixins/localize-dynamic-mixin.js
dlockhart: At least this will suppress the JS error like you said, and it'll allow the loop through the supported languages to continue... which likely just means it'll fall back to English. But we also added `es-es` recently so it would have nicely fallen back to `es`.

Would a `console.warn` inside the `catch` bit make sense? Not sure if we can get the message to be meaningful enough for devs to actually be able to debug it though... like it would need to have the element name I think and I'm not sure we can get that.

mixins/localize-dynamic-mixin.js
dlockhart: Haha it's a fantastic message.

mixins/localize-dynamic-mixin.js
bearfriend: Since the actual error occurs on the arrow function that the components write, it should be in the stack.

mixins/localize-dynamic-mixin.js
dbatiste: Yeah, we would make heroic efforts to avoid a breaking change.  It would be **extremely** disruptive. We should ensure we have a strategy such that we never have to make a major version bump.

bearfriend:
I realized we don't actually need `supportedLanguages` at all, because rollup replaces the import with a switch. If the language isn't _truely_ supported (a file exists!) there is no `case` and it falls through and loops to the next language. If someone is using this outside BSI and for whatever reason can't use a build system that does the same, they're no worse off. They can either be ok with 404s, handle them themselves in their `importFunc`, or simply use the Manual method instead.

mixins/localize-dynamic-mixin.js
```diff
@@ -0,0 +1,57 @@
...
+const defaultExportName = 'default';
+
+function warnMissingFiles(failures, resolvedLang) {
+	const { path } = failures[0].err.message.match(/(?:https?:\/\/.*?\/)(?<path>.*)\/.*\./).groups;
+	const langs = failures.map(f => f.lang).join('","');
+	console.warn(
+		'WARNING: Unacceptable in production. Consider Manually Retrieved Resources method.\n\n' +
+		`Failed to load translation resources for languages "${langs}" from ${path}.`,
+		`Falling back to "${resolvedLang}". `);
+}
```
bearfriend, line 14: Definitely open to dealing with this differently.

mixins/localize-mixin.md
```diff
@@ -4,9 +4,9 @@ The `LocalizeMixin` and `LocalizeStaticMixin` allow you to localize text in your
 
 ## Providing Resources
 
-Your component must provide resources by either implementing a static `resources` getter for local resources, or a `getLocalizeResources` method to fetch resources asynchronously. The `getLocalizeResources` method will be passed an array of lowercase languages in preferential order.
+Your component must provide resources by either implementing a static `resources` getter for local resources, or a `config` getter to fetch resources asynchronously. The `importFunc` method of `config` will be called with lowercase languages in preferential order.
 
-Your implementation should find the resources that best match the languages passed in. It should then return an object containing two values:
+Static implementations should return an object containing two values:
```
dlockhart, line 9: Since it only applies to static, can you move this bit down to the static example?

mixins/localize-dynamic-mixin.js
```diff
@@ -0,0 +1,57 @@
...
+
+export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(superclass) {
+
+	static get config() {
```
dlockhart, line 18: Wondering if we should call this something like `localizeConfig` to avoid possible collisions with other mixins or properties/methods on the element?

mixins/localize-mixin.md
```diff
@@ -56,27 +58,56 @@ class MyComponent extends LocalizeStaticMixin(LitElement) {
...
+
+Then create your `config` getter:
+```javascript
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+
+class MyComponent extends LocalizeDynamicMixin(LitElement) {
+
+  static get config() {
+    return {
+      importFunc: lang => import(`../lang/${lang}.js`) // Path must be relative!
```
dlockhart, line 83: Could this not `import('../lang/${lang}.js').val` or `import('../lang/${lang}.js').default` and remove the need for the `exportName` config?

mixins/localize-mixin.md
```diff
@@ -56,27 +58,56 @@ class MyComponent extends LocalizeStaticMixin(LitElement) {
...
+
+```javascript
+return {
+  ...,
+  osloCollection: 'my-project\\myComponent', // To enable OSLO
+  exportName: 'resources' // If your resource files use named exports
+}
+```
+
+#### Example 3: Manually Retrieved Resources
```
dlockhart, line 99: We could even ditch this section from the docs altogether if you want. 🤷

mixins/localize-dynamic-mixin.js
dlockhart: I'd suggest removing the "Unacceptable in production" bit since chances are these types of problems _will_ sneak into production and it doesn't look great if clients see this.

Maybe just drop the first line entirely?

mixins/localize-mixin.md
bearfriend: Yeeeeesssss.

mixins/localize-mixin.md
bearfriend: I thought about it, but if people use this outside BSI it could be helpful.

mixins/localize-mixin.md
bearfriend: This appears to not even be accurate for static implementations?

dlockhart: Oh weird yeah they don't need to return anything eh.

dlockhart (approved):
I really like where this has arrived at! I'm good with merging this in. When you do, can you squash your commits and put `feat: ` in the squash commit to cause a minor release?

Thanks SO much for contributing this, it's going to make localizing components so much easier -- I'll feature it in tomorrow State of the UI as well.

PR #1148, opened 2021-02-22
input-time: visual-diff: flaky localization tests

margaree:


margaree:
The `input-time` localization tests have been really stable. I think I've done 10+ runs of visual-diff and they've been fine. Within this PR I also did some tweaks to how `d2l-input-time-hidden-content-width-change` listening is working and that seems stable as well.

If they `input-time` tests continue to be stable and this merges then I'll keep an eye on them.

dlockhart (approved):
Awesome stuff Margaree!

PR #1150, opened 2021-02-22
input-date: visual-diff: fix flaky test

margaree:


margaree:
This seems stable at this point. I'll try moving the input-time tests back to the way they were and see if it's still good.

components/inputs/input-time.js
```diff
@@ -241,23 +241,16 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
 			this._formattedValue = formatTime(time);
 		}
 
+		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-time-hidden-content');
 		this.addEventListener('d2l-localize-behavior-language-changed', () => {
 			this._formattedValue = formatTime(getDateFromISOTime(this.value));
 			INTERVALS.clear();
+			this.requestUpdate().then(() => this._onResize(hiddenContent));
```
margaree, line 248: This is how `input-date` is working and would just ensure that the size change happens on language change as well as trigger the event for the tests.

components/inputs/test/input-date-time-range.visual-diff.js
```diff
@@ -95,7 +95,7 @@ describe('d2l-input-date-time-range', () => {
 					return new Promise((resolve) => {
 						elem.updateComplete.then(() => {
 							const timeElem = dateElem.shadowRoot.querySelector('d2l-input-time');
-							timeElem.addEventListener('d2l-input-time-hidden-content-width-change', () => timeElem.updateComplete.then(resolve));
+							timeElem.addEventListener('d2l-input-time-hidden-content-width-change', () => timeElem.updateComplete.then(setTimeout(resolve, 100)));
```
margaree, line 98: Some tweaks to make flake less frequent/hopefully prevent it

components/inputs/test/input-date.visual-diff.js
```diff
@@ -198,6 +198,7 @@ describe('d2l-input-date', () => {
 							);
 							input.dispatchEvent(e);
 						});
+						await page.waitForTimeout(100);
```
margaree, line 201: I tried modifying this to wait for calendar updateComplete but that didn't work. I haven't seen this test flake since adding this.

components/inputs/test/input-time-range.visual-diff.js
```diff
@@ -235,14 +235,14 @@ describe('d2l-input-time-range', () => {
...
 		before(async() => {
-			await page.reload();
-			await page.$eval('#labelled', (elem) => {
-				const timeElem = elem.shadowRoot.querySelector('d2l-input-time');
-				return new Promise((resolve) => {
-					elem.updateComplete.then(() => {
-						timeElem.addEventListener('d2l-input-time-hidden-content-width-change', resolve);
-					});
-				});
+			await page.$eval('#hidden-labels', async(elem) => {
```
margaree, line 238: These skeleton tests were sometimes not run which seemed to be because `resolve` was never called. This seems to work without the `page.reload` and with just resizing these cases back to their original sizes instead.

margaree:
The tests have been stable for many runs now.

dlockhart (approved):

PR #1167, opened 2021-02-24
input-date-time-range: change test to no longer flake

margaree:
Problem:
Between 5 - 6 pm some of the tests end up being run on a machine that has its timezone set to Central Europe Time. In 2020, CET had Daylight Savings time end on Oct 25. Since our difference calculation for shifting dates is done using `Date`, because of the time change it becomes off by 1 hour (https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-date-time-range.js#L34).

"Solution":
I shifted all the tests to be further from that time shift.

What does this mean:
This is how our old date picker behaved as well and is not new. How it normally exhibits itself is that when a user picks a date that moves start to one side of daylight savings and end to the other side, the end time shifts by an hour. This usually should seem fine to a user since they would in theory have their system set to the same timezone as their LMS. This does bring up the interesting case where if the system timezone is different than the LMS timezone then they would see that weird shifting on a different date than potentially expected. In practice I don't think this would cause any user error as the user is likely to change the time anyways.

### DIFF
--- a/components/inputs/test/input-date-time-range.test.js
+++ b/components/inputs/test/input-date-time-range.test.js
@@ -49,10 +49,10 @@ describe('d2l-input-date-time-range', () => {
 				});
 
 				it('should return correctly forward shifted end date when localized without Z', () => {
-					const prevStartValue = '2020-10-25T04:00:00.000';
-					const prevEnd = '2020-10-27T04:00:00.000';
-					const start = '2020-10-26T04:00:00.000';
-					const newEndValue = '2020-10-28T04:00:00.000';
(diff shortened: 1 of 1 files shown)


dlockhart:
Wow crazy, awesome digging to figure that out! So they actually run the tests on different machines at different times of the day?!

I agree it doesn't sound like this is going to result in user errors, and just shifting the times around to fix the "flake" (I guess not really since it's consistent) makes sense to me.

dlockhart (approved):

dbatiste:
😮

PR #1168, opened 2021-02-24
Add text-align variable for input to allow overriding.

dbatiste:


components/inputs/input-styles.js
```diff
@@ -20,6 +20,7 @@ export const inputStyles = css`
 		min-height: calc(2rem + 2px);
 		min-width: calc(2rem + 1em);
 		position: relative;
+		text-align: var(--d2l-input-text-align, start);
```
dlockhart, line 23: This is for HTML editor?

dbatiste: Color picker.
image

dbatiste: For a bit more context:
image

PR #1170, opened 2021-02-25
Add d2l-input-percent.

dbatiste:
This PR adds a new input variant, `d2l-input-percent` with all the regular input fixins.  It wraps `d2l-input-number` adding the "%" in the margin beside the input.

image


components/inputs/input-number.js
```diff
@@ -22,6 +22,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 			autocomplete: { type: String },
 			autofocus: { type: Boolean },
 			disabled: { type: Boolean },
+			hideInvalidIcon: { attribute: 'hide-invalid-icon', type: Boolean, reflect: true },
```
dbatiste, line 25: Needed to expose these things on `d2l-input-number` so that `d2l-input-percent` could use them.

components/inputs/input-percent.js
```diff
@@ -0,0 +1,180 @@
...
+	render() {
+		return html`
+			<d2l-input-number
+				autocomplete="${ifDefined(this.autocomplete)}"
+				?autofocus="${this.autofocus}"
+				@blur="${this._handleBlur}"
+				@change="${this._handleChange}"
+				?disabled="${this.disabled}"
+				.forceInvalid="${this.invalid}"
+				hide-invalid-icon
```
dbatiste, line 122: Hide the invalid icon because it obscures too much of the value.

components/inputs/input-percent.js
```diff
@@ -0,0 +1,180 @@
...
+		const childErrors = await inputNumberElem.validate();
+		const errors = await super.validate();
+		return [...childErrors, ...errors];
+	}
+
+	async _handleChange(e) {
+		const oldValue = this.value;
+		this.value = e.target.value;
+
+		if (e.target.value === undefined) {
```
dbatiste, line 160: The `target.value` will from `d2l-input-number` will be `undefined` if the user entered an invalid value, so we reset to the old value.

components/inputs/docs/input-percent.md
```diff
@@ -0,0 +1,50 @@
...
+</script>
+<d2l-input-percent label="Label" value="100"></d2l-input-percent>
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `label` | String, required | Label for the input. |
+| `autocomplete` | String | Specifies which types of values [can be autofilled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) by the browser. |
```
dlockhart, line 19: `autocomplete` can likely be excluded since none of the value types that are numbers (like birthday-month) would apply to percents.

components/inputs/input-number.js
```diff
@@ -113,8 +115,11 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 				?required="${this.required}"
 				?skeleton="${this.skeleton}"
 				title="${ifDefined(this.title)}"
-				.value="${this._formattedValue}"
-			><slot slot="after" name="after"></slot></d2l-input-text>
+				.value="${this._formattedValue}">
+					<slot slot="left" name="left"></slot>
+					<slot slot="right" name="right"></slot>
+					<slot slot="after" name="after"></slot>
```
dlockhart, line 121: Be a little careful of spacing here... I seem to recall having to remove all the whitespace because when slots were missing there would be an extra `&nbsp;` in between them.

components/inputs/input-percent.js
```diff
@@ -0,0 +1,180 @@
+import './input-number.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { FormElementMixin } from '../form/form-element-mixin.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
+
+/**
+ * This component wraps the "<d2l-input-number>" tag and is intended inputting percent values.
```
dlockhart, line 10:
```suggestion
 * This component wraps the "<d2l-input-number>" tag and is intended for inputting percent values.
```

components/inputs/input-percent.js
dlockhart: Another option would be to just never let this go into an invalid state? Like you could always round up to 0 and down to 100 and if they enter something that isn't a number.

components/inputs/input-percent.js
```diff
@@ -0,0 +1,180 @@
...
+				min="0"
+				min-fraction-digits="${ifDefined(this.minFractionDigits)}"
+				name="${ifDefined(this.name)}"
+				?noValidate="${this.noValidate}"
+				placeholder="${ifDefined(this.placeholder)}"
+				?required="${this.required}"
+				?skeleton="${this.skeleton}"
+				title="${ifDefined(this.title)}"
+				value="${ifDefined(this.value)}">
+					<span slot="right" @click="${this._handleSymbolClick}">%</span>
```
dlockhart, line 137: Not to complicate things, but I just remembered that different locales have different characters and placements of the percent symbol. In reality, I think all the ones we care about are `%`, but the placement does vary slightly. If you grab the `getNumberDescriptor` [out of Intl](https://github.com/BrightspaceUI/intl/blob/master/lib/number.js#L104), you can access the `percentPositivePattern`. Then you can see if the pattern is `{num} %` or `% {num}`. The spaces are also optional, but I don't think you care about them.

We could also add this as a future enhancement.

dlockhart (approved):

components/inputs/input-percent.js
dbatiste: Yeah I think that could work.  Then I can enable the invalid icon for the required/empty state, which is a nice feature.

dbatiste: Got this working, so not for required fields, it will show the icon.
image

PR #1178, opened 2021-03-01
Add subtle mode for dropdown-more and button-icon

GeorgePlukov:
Button groups have the option to use `d2l-button-subtle`, when page is made small we automatically shift it to use a `d2l-button-more` but the button styling had no option for setting it to celestine. This allows us to pass the subtle flag to change the icon and hover colors. 

I thought this was a reasonable method to implement this but if there is a more intuitive api (wrapping in a parent component for subtle) let me know whats preferred. 

Screen Shot 2021-03-01 at 2 57 18 PM


dlockhart:
@geurts wondering if you have any thoughts here? My main worry is that our current subtle buttons don't support an icon-only mode. So this change is kind of merging the subtle button styling into button-icon... which doesn't seem quite right.

There's also the complexity where button-icon could be `translucent` AND `subtle`... which is weird.

Some possible outcomes:
- We change subtle buttons to support icon-only
- We don't use subtle buttons in this particular case -- where a group of subtle buttons collapses into a "..." more button
- ??

GeorgePlukov:
I discussed this with @geurts in terms of having the `...` icon match color to the rest of the button groups and we agreed that it would make sense (in that context) to have the icon the same color as the subtle buttons. I did think that potentially enabling them both translucent and subtle would look weird. I'm definitely open to implementing this in another way as you have valid concerns @dlockhart but due to the elements being children of the parent I couldn't manage to get the styling to match without exposing some sort of flag.

components/button/button-icon.js
```diff
@@ -100,6 +101,12 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				:host([active]) button {
 					background-color: var(--d2l-button-icon-background-color-hover);
 				}
+				:host d2l-icon {
+					color: var(--d2l-button-icon-text-color);
+				}
+				:host button:hover > d2l-icon {
```
dbatiste, line 107: Should this be applied in the focus state as well?

dbatiste (approved):

components/button/button-icon.js
GeorgePlukov: Nice catch! It does look like we are applying hover styling to the focus state for button-icons.

components/button/button-icon.js
```diff
@@ -49,6 +49,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 					--d2l-button-icon-min-height: calc(2rem + 2px);
 					--d2l-button-icon-min-width: calc(2rem + 2px);
 					--d2l-button-icon-h-align: calc(((2rem + 2px - 0.9rem) / 2) * -1);
+					--d2l-button-icon-text-color: var(--d2l-color-ferrite);
+					--d2l-button-icon-text-color-hover: var(--d2l-color-ferrite);
```
dlockhart, line 53: Hmm... rethinking this slightly since button-icon doesn't have "text" per se (my bad for suggesting that name). Looking through this, there's also that `--d2l-icon-fill-color` variable that defaults to ferrite that we're already overriding for dark mode and the translucent style. Could we leverage it maybe?

Something like:

```css
:host {
  --d2l-icon-fill-color: var(--d2l-button-icon-color, var(--d2l-color-ferrite));
}
/* same as existing rules -- i.e. you can't override the icon colour if it's a translucent or dark mode button-icon */
:host([translucent]) {
  --d2l-icon-fill-color: white;
}
:host([theme="dark"]) {
  --d2l-icon-fill-color: var(--d2l-color-sylvite);
}
/* I'm not sure if this part will work... changing a variable value in a pseudo-class */
button:hover,
button:focus,
:host([active]) button {
  background-color: var(--d2l-button-icon-background-color-hover);
  --d2l-icon-fill-color: var(--d2l-button-icon-color-hover, var(--d2l-color-ferrite));
}
```

components/button/test/button-icon.visual-diff.html
```diff
@@ -26,6 +26,8 @@
 			--d2l-button-icon-min-width: 1.5rem;
 			--d2l-button-icon-border-radius: 4px;
 			--d2l-button-icon-focus-box-shadow: 0 0 0 1px #ffffff, 0 0 0 3px #006fbf;
+			--d2l-button-icon-text-color: var(--d2l-color-celestine);
+			--d2l-button-icon-text-color-hover: var(--d2l-color-celestine-minus-1);
```
dlockhart, line 30:
```suggestion
			--d2l-button-icon-fill-color: var(--d2l-color-celestine);
			--d2l-button-icon-fill-color-hover: var(--d2l-color-celestine-minus-1);
```

dlockhart (approved):

PR #1180, opened 2021-03-03
US125413 - Allow support for LaTeX rendering in our MathJax implementation

grant-cleary:
This PR allows consumers to specify an optional MathJax config object when using an HTML block. Right now this is used to enable (or disable) LaTeX rendering within MathJax.

### DIFF
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -10,6 +10,17 @@
 			import '../../demo/demo-page.js';
 			import '../html-block.js';
 		</script>
+		<script>
+			if (window.location.search.indexOf('latex=true') !== -1) {
+				D2L = {};
+				D2L.LP = {};
+				D2L.LP.Web = {};
+				D2L.LP.Web.UI = {};
+				D2L.LP.Web.UI.Flags = {
+					Flag: (flagName, defaultValue) => { return true; }
+				}
+			}
+		</script>
 	</head>
 	<body unresolved>
 
@@ -41,6 +52,7 @@
 								<li>ordered item 2</li>
 							</ol>
 							<div><a href="https://d2l.com">anchor</a></div>
+							<div style="display: none">\( )\</div>
 						</template>
 					</d2l-html-block>
 				</template>
@@ -255,6 +267,33 @@
 										<mo>)</mo>
 									</mrow>
 								</math> and other symbols.
+							</div>
+						</template>
+					</d2l-html-block>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>HTML Block (LaTeX math)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-html-block>
+						<template>
+							<div>$$ f(x) = \int \mathrm{e}^{-x}\,\mathrm{d}x $$ $$ x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a} $$</div>
+						</template>
+					</d2l-html-block>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>HTML Block (LaTeX inline math)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-html-block>
+						<template>
+							<div>An equation rendered using LaTeX...
+								\( f(x) = \int \mathrm{e}^{-x}\,\mathrm{d}x \)
+								... and some text!
 							</div>
 						</template>
 					</d2l-html-block>
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -11,6 +11,7 @@
 			import '../html-block.js';
 		</script>
 		<script>
+			// mock flag for us125413-mathjax-render-latex
 			if (window.location.search.indexOf('latex=true') !== -1) {
 				window.D2L = {};
 				D2L.LP = {};
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -11,14 +11,16 @@
 			import '../html-block.js';
 		</script>
 		<script>
-			// mock flag for us125413-mathjax-render-latex
 			if (window.location.search.indexOf('latex=true') !== -1) {
 				window.D2L = {};
 				D2L.LP = {};
 				D2L.LP.Web = {};
 				D2L.LP.Web.UI = {};
 				D2L.LP.Web.UI.Flags = {
-					Flag: () => { return true; }
+					Flag: (feature, defaultValue) => {
+						if (feature === 'us125413-mathjax-render-latex') return true;
+						else return defaultValue;
+					}
 				};
 			}
 		</script>
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -19,7 +19,7 @@
 
 			<d2l-demo-snippet>
 				<template>
-					<d2l-html-block math-jax-config='{ "renderLatex": true }'>
+					<d2l-html-block>
 						<template>
 							<h1>heading 1</h1>
 							<h2>heading 2</h2>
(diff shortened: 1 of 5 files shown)


components/html-block/html-block.js
```diff
@@ -129,14 +140,14 @@ class HtmlBlock extends LitElement {
...
-				const hasMath = !!fragment.querySelector('math');
+				const hasMathML = !!fragment.querySelector('math');
 
 				const temp = document.createElement('div');
 				temp.appendChild(fragment);
 				const fragmentHTML = temp.innerHTML;
 
-				if (hasMath) {
-					await loadMathJax();
+				if (hasMathML || /\$\$|\\\(/.test(fragmentHTML)) {
```
grant-cleary, line 149: Right now we're only supporting the `$$` and `\(` delimiters because that's what Carlton currently needs. If others feel strongly about it, I can expand this to include the `\[` delimiter as well.

Trandrew17 (approved):

components/html-block/html-block.js
```diff
@@ -7,6 +7,12 @@ import { loadMathJax } from '../../helpers/mathjax.js';
  */
 class HtmlBlock extends LitElement {
 
+	static get properties() {
+		return {
+			mathJaxConfig: { type: Object, attribute: 'math-jax-config' }
```
dbatiste, line 12: I don't think we should surface "MathJax" specifically on the `d2l-html-block` API - the fact that we are using MathJax should be kept internal from our app components.  That aside, exposing the LaTeX aspect on `d2l-html-block`'s API allows for inconsistent UX that would be frustrating for users, and also introduces a bunch of complexity for those scenarios (ex. more than one `d2l-html-block` on the same page with different math settings).

Instead, I was thinking that we would internally check the client-side feature flag for the instance/org.  In effect, this is passed "under the covers" without exposing it on any APIs.

helpers/mathjax.js
```diff
@@ -110,7 +110,12 @@ export function loadMathJax() {
 		const script = document.createElement('script');
 		script.async = 'async';
 		script.onload = resolve;
-		script.src = 'https://s.brightspace.com/lib/mathjax/3.1.2/mml-chtml.js';
+
+		const component = mathJaxConfig && mathJaxConfig.renderLatex
```
dbatiste, line 114: 👍

components/html-block/html-block.js
grant-cleary: Oh, I see what you mean. Sure, yeah. Something like `if (D2L.LP && {FlagCheck}) ...` should be easy enough to do here. I suppose that won't work in FRAs though without designing a pass-through for flags.

components/html-block/html-block.js
```diff
@@ -129,14 +129,18 @@ class HtmlBlock extends LitElement {
...
-				const hasMath = !!fragment.querySelector('math');
+				const hasMathML = !!fragment.querySelector('math');
 
 				const temp = document.createElement('div');
 				temp.appendChild(fragment);
 				const fragmentHTML = temp.innerHTML;
 
-				if (hasMath) {
-					await loadMathJax();
+				const hasLatex = D2L.LP && D2L.LP.Web.UI.Flags.Flag('us125413-mathjax-render-latex', true) && /\$\$|\\\(/.test(fragmentHTML);
```
grant-cleary, line 138: Eventually this will depend on a config variable, at which point I imagine we'll need to implement a provider/requestor pattern here to fetch the right LMS information if it's available.

grant-cleary:
I'm not sure what's up with the visual diff tests here...

components/html-block/demo/html-block.html
```diff
@@ -41,6 +52,7 @@ <h6>heading 6</h6>
 								<li>ordered item 2</li>
 							</ol>
 							<div><a href="https://d2l.com">anchor</a></div>
+							<div style="display: none">\( )\</div>
```
dbatiste, line 55: Is this needed?

grant-cleary: So the first HTML block that renders always wins, which means if this first HTML block doesn't contain LaTeX, we'll load the wrong MathJax renderer. This was mostly just a quick and dirty way to achieve that without placing the math examples first. Feels like it would also be fine to just only display the LaTeX examples in the `latex=true` case and omit them in the base case.

components/html-block/html-block.js
dbatiste: I think we should move this down so that can rely on the `||` short-circuit and avoid the regex test in the case where mathML was found.

components/html-block/demo/html-block.html
```diff
@@ -10,6 +10,17 @@
...
+			if (window.location.search.indexOf('latex=true') !== -1) {
+				D2L = {};
+				D2L.LP = {};
+				D2L.LP.Web = {};
+				D2L.LP.Web.UI = {};
+				D2L.LP.Web.UI.Flags = {
+					Flag: (flagName, defaultValue) => { return true; }
+				}
+			}
+		</script>
```
grant-cleary, line 23: Ah good, linting doesn't like my trickery (though I also did legitimately miss a semi-colon here).

components/html-block/html-block.js
grant-cleary: So I don't think we can. Someone can use the equation editor to create an equation _and_ use inline LaTeX in the same block of HTML. In that case, we still need to load the LaTeX renderer.

components/html-block/demo/html-block.html
```diff
@@ -10,6 +10,18 @@
 			import '../../demo/demo-page.js';
 			import '../html-block.js';
 		</script>
+		<script>
+			if (window.location.search.indexOf('latex=true') !== -1) {
+				/* eslint-disable-next-line no-global-assign */
```
grant-cleary, line 15: Linting thinks that `D2L` is a global readonly variable... Which is generally true, but not on demo pages.

components/html-block/html-block.js
dbatiste: Oh, seems like we are mixing two things here: has math (mathML or LaTeX), and supports LaTeX.  Can we do something like this:
```javascript
const isLaTeXSupported = D2L.LP && D2L.LP.Web.UI.Flags.Flag('us125413-mathjax-render-latex', true);
if (hasMathML || (isLaTeXSupported && /\$\$|\\\(/.test(fragmentHTML)))  {
    const mathJaxConfig = { renderLatex: isLaTeXSupported };
    await loadMathJax(mathJaxConfig);
    ...
}
```
That way mathJax gets loaded with the correct configuration we support regardless of whether it contains mathML or LaTeX.  It's entirely valid scenario of an HtmlBlock to initially contain mathML, and then the contents get updated to contain LaTeX after the first render.

components/html-block/demo/html-block.html
dbatiste: I think you can also avoid the linting error by starting with `window.`... i.e. `window.D2L = {};`

grant-cleary: Ah, I'll try that then. Better than adding linting overrides.

components/html-block/html-block.js
grant-cleary: Ah yeah, I wasn't thinking of re-rendering. This makes sense, yeah.

dbatiste: (i think this will address the other thread about the latex that is display: none above)

grant-cleary: It did and I was able to remove that. I got it stuck in my head that we only wanted the LaTeX component loaded if LaTeX was on the page.

dbatiste (approved):

components/html-block/html-block.js
dbatiste: Yeah so just to close the look on this - when/if we need to support this in a FRA, we'll need to pass the proper config through ifrau to it can be used in the FRAs context.

grant-cleary: Thinking about ifrau makes me sad, Dave.

PR #1181, opened 2021-03-03
Awang/us124817 visibility mixin

AndyWangD2L:
[Rally Link - US124817](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F496667342648)

Creates the visibility mixin used for animating a component when it is shown or hidden.


mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+				{ bubbles: true, composed: false }
+			));
+		};
+		const dummyOnTransitionEnd = () => {
+			this.dummy.replaceWith(this);
+			this.dispatchEvent(new CustomEvent(
+				'd2l-visibility-mixin-remove-end',
+				{ bubbles: true, composed: false }
+			));
+			this.remove();
```
AndyWangD2L, line 96: I feel like this should come before the event dispatch. However, when I do so, the test checking for 'd2l-visibility-mixin-remove-end' event firing fails. It would say `event.target` is null (and I think `event` itself was null as well). But adding an event listener to it and using `console.log` shows that both `event` and `event.target` are not null.

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+					if (animateStyle.dummyOnTransitionEnd) {
+						animateStyle.dummyOnTransitionEnd();
+					}
+				}
+			};
+
+			Object.assign(this.dummy.style, animateStyle.finalDummy);
+			Object.assign(this.style, animateStyle.final);
+		}
+
```
dlockhart, line 198: Can these move up to the top and then you can return after them and remove the nested `if`-statement?

```js
if (reduceMotion) {
  if (animateStyle.thisOnTransitionEnd ) animateStyle.thisOnTransitionEnd();
  if (animateStyle. dummyOnTransitionEnd ) animateStyle. dummyOnTransitionEnd();
  return;
}
// non-reduce motion code here
```

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+			if (document.body.contains(this.dummy)) {
+				// preserve the current opacity & transform for when we switch directions of transition
+				this.style.opacity = window.getComputedStyle(this).opacity;
+				this.style.transform = window.getComputedStyle(this).transform;
+
+				// preserve the current dummy height for when we switch directions of transition
+				this.dummy.style.height = window.getComputedStyle(this.dummy).height;
+				this.dummy.replaceWith(this);
+			}
+			this.replaceWith(this.dummy);
```
dlockhart, line 157: Be careful with `replaceWith`... [caniuse says](https://caniuse.com/mdn-api_childnode_replacewith) it was only added in Safari 14. You can probably accomplish the same thing by `this.parentNode.insertBefore(dummy, this); dummy.appendChild(this);` I think.

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+
+	_animateHide() {
+		const thisOnTransitionStart = () => {
+			this.dispatchEvent(new CustomEvent(
+				'd2l-visibility-mixin-hide-start',
+				{ bubbles: true, composed: false }
+			));
+		};
+		const dummyOnTransitionEnd = () => {
+			this.dummy.replaceWith(this);
```
dlockhart, line 49: See comment below about using `replaceWith`.

mixins/visibility-mixin.js
dlockhart: Interesting, although not entirely surprising. It's unfortunate that we need `this.dummy.replaceWith(this)` (which will cause a full re-render) just to dispatch the event and then right after we remove it (causing another re-render). Wondering how important those events are beyond tests? Or can we just not do the `replaceWith` call and have `this` just be a child of `dummy` when the event fires?

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+				'd2l-visibility-mixin-remove-end',
+				{ bubbles: true, composed: false }
+			));
+			this.remove();
+		};
+		this._animateHideRemove(thisOnTransitionStart, dummyOnTransitionEnd);
+	}
+
+	_animateShow() {
+		this.style.display = this.displayOriginal;
```
dlockhart, line 102: I'm a little worried about all the code that's manipulating the DOM/style outside of the `render()` method. Technically in Lit we're not supposed to do this -- `render()` should be a function of the component's properties and all re-renders should occur by changing them. I'm not sure how else to accomplish this though...

dlockhart (commented):
This is starting to come together for sure! Highlighted a few concerns in the comments that we can either discuss here or in the Slack channel.

mixins/visibility-mixin.js
dlockhart: One option we could discuss around this would be for this to go back to being a container instead of a mixin -- I know we had briefly discussed that in Slack before this was written. That way it would have its own `render()` method so all these direct manipulations would go away.

The downside would be it would be wrapping `d2l-list-item`s, which might mess up the ARIA and other cases where list was expecting particular children. But maybe `role="presentation"` on itself and a few other tweaks could get around that though. Discuss!

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	constructor() {
+		super();
+		this.dummy = document.createElement('div');
+		this.dummy.style.overflow = 'hidden';
+		this.dummy.style.display = 'grid';
+	}
+
+	firstUpdated() {
```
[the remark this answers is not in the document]
bearfriend: `getComputedStyle` will actually cause a whole reflow, not just a paint. Not sure if `requestAnimationFrame` will help it.

mixins/visibility-mixin.js
AndyWangD2L: We could have `this` be a child of `dummy` when the event fires, and then do a `this.dummy.remove()` to get rid of both `this` and `dummy`. Seems to work without any issues, aside from it being strange to have the event dispatch before the actual remove happen.

mixins/visibility-mixin.js
AndyWangD2L: Looks like this works the same without issues. Fixed.

mixins/visibility-mixin.md
```diff
@@ -0,0 +1,92 @@
...
+			<p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+		`;
+	}
+}
+customElements.define('my-component', MyComponent);
+```
+
+To show the component with the transition, append it to the DOM with the attribute `animate='show'`. The opacity & transform are animated to the component's values at the time in which `animate` is first updated.
+To hide the component, apply the attribute `animate='hide'`. Alternatively, to remove the component, apply the attribute `animate='remove'`.
+Once the component is hidden (but not removed), show the component again by re-applying the attribute `animate='show'`.
```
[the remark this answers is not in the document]
AndyWangD2L: Ah, I see what you mean from the demos. Would simply `To show the component with the transition, render it with the attribute animate='show'` make more sense?

mixins/visibility-mixin.js
zinabat: For list items, I think that it may be worth making a special type of `d2l-list-item` that combines the visibility container with the `d2l-list-item-mixin`. It might need a bit of adjustment to do this effectively, but this seems cleaner than wrapping list-item components themselves.

zinabat: **Alternatively**, if we stick with a mixin, I wonder if we could change to use render methods that take templates as arguments like what we did with our list item mixins: https://github.com/BrightspaceUI/core/blob/master/components/list/list-item-checkbox-mixin.js

They require the consumer to call the render methods, but this seems like a reasonable thing that we already do elsewhere. This gives us indirect access to the render method so that we can rely on properties instead of manipulating the DOM and styles manually.

mixins/visibility-mixin.md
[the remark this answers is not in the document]
dlockhart: Yeah exactly

mixins/visibility-mixin.js
dlockhart: I like that, and Dave B suggested something similar. Let's continue this discussion the Slack thread in `#wc-visibility-animations`.

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
...
+				this.dummy.style.height = window.getComputedStyle(this.dummy).height;
+				this.dummy.replaceWith(this);
+			}
+			this.replaceWith(this.dummy);
+			this.dummy.appendChild(this);
+
+			// allow enough time for reflow to occur to ensure that the transition properly runs
+			await new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r)));
+
+			this.ontransitionstart = (event) => {
```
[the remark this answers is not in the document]
AndyWangD2L: You're right, changing it to `this.addEventListener('transitionstart')` is much better incase others were trying to listen for it as well. Fixed! Also cleans up listeners now.

mixins/demo/visibility-test-item.js
```diff
@@ -0,0 +1,46 @@
...
+			}
+			d2l-button-icon {
+				flex: 0 0 auto;
+			}
+		`;
+	}
+
+	render() {
+		return html`<span>Item ${this.number}</span>
+			<d2l-button-icon icon="tier1:delete" text="Remove" @click="${this._remove}"></d2l-button-icon>`;
```
m6jones, line 33: nit: 
```
return html`
            <span>Item ${this.number}</span>
			<d2l-button-icon icon="tier1:delete" text="Remove" @click="${this._remove}"></d2l-button-icon>
`;
```

mixins/visibility-mixin.js
AndyWangD2L: Yeah this looks way better. I also ended up moving all `reduceMotion` code out of `_animateVisibility` and directly into the top of each of show/hide/remove instead.

mixins/visibility-mixin.js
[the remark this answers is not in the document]
AndyWangD2L: Is there a way for me to tell if `requestAnimationFrame` helps?

mixins/visibility-mixin.js
```diff
@@ -0,0 +1,207 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+
+const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
+const transitionDuration = 300;
+const moveYValue = 10;
+
+export const VisibilityMixin = dedupeMixin(superclass => class extends superclass {
+	static get properties() {
+		return {
+			animate: { type: String }
```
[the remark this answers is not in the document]
AndyWangD2L: Yeah I agree, I like that better. I think it's clearer to the consumer what it's related to and used for. Waiting on more input about this before changing.

AndyWangD2L:
Closing this mixin PR in favour of using directives instead, as found in [this PR](https://github.com/BrightspaceUI/core/pull/1203).

PR #1183, opened 2021-03-04
Overflow group component

GeorgePlukov:


svanherk:
> Could not generate new goldens - your code changes will update golden files that you do not have the latest version of. Please rebase or merge master into your branch.

Sometimes this message is a red herring, which bugs me about it...  going to add a to do list item to try to make it smarter. 
 It seems like the actual issue is: `ReferenceError: changeButtonsTests is not defined`

GeorgePlukov:
Thanks! I forgot I ended up moving those tests to unit tests but guess I didn't remove all the references

components/button-group/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `auto-show` | Boolean | Automatically determine the min and maximum number of items to show based on which elements have classes `d2l-button-group-show` and `d2l-button-group-no-show`. Please consult the design team when using this attribute. |
+| `min-to-show` | Number | The minimum number of elements to always show. Please consult the design team when using this attribute. |
+| `max-to-show` | Number | The maximum number of elements to always show |
+| `subtle` | Boolean | Styles the dropdown menu in the button-group as a subtle button for action button groups |
```
dlockhart, line 43: Wondering if instead of introducing a new attribute, we could automatically figure this out based on what types of buttons were rendered inside?

components/button-group/README.md
```diff
@@ -0,0 +1,51 @@
...
+| `opener-type` | String | Currently the only option for this is `icon`, which will permanently render the dropdown menu symbol as `...` |
+
+
+## d2l-action-button-group
```
dlockhart, line 47: You can exclude this section -- we'll communicate this change in the old repo, or possibly we can just convert all the places using it over and pretend it never existed!

components/button-group/button-group.js
```diff
@@ -0,0 +1,162 @@
...
+import '../menu/menu.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { DynamicResizeMixin } from '../../mixins/dynamic-resize-mixin.js';
+import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
+import { offscreenStyles } from '../offscreen/offscreen.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+/**
+ *
+ * A button group component that can be used to display a set of buttons
```
dlockhart, line 19:
```suggestion
 * Wraps a set of buttons and responsively places overflowing buttons into a menu
```

components/button-group/button-group.js
```diff
@@ -0,0 +1,162 @@
...
+				display: block;
+
+			}
+			
+			:host([subtle]) {
+				--d2l-button-icon-fill-color: var(--d2l-color-celestine);
+				--d2l-button-icon-fill-color-hover: var(--d2l-color-celestine-minus-1);
+			}
+
+			:host .d2l-button-group-container {
```
dlockhart, line 51: No need to prefix any of these with `:host` -- the way shadow DOM works all these styles will be scoped to the host anyway.

components/button-group/button-group.js
```diff
@@ -0,0 +1,162 @@
...
+				display: flex;
+				flex: 0 1 auto;
```
dlockhart, line 53: I'm surprised this is doing anything? `display: flex` makes `.d2l-button-group-container` a flexbox container, but `flex: 0 1 auto` assumes that itself is contained in a flexbox, which it isn't.

components/button-group/button-group.js
```diff
@@ -0,0 +1,162 @@
...
+				margin-right: 0;
+			}
+
+			:host([subtle]) .d2l-dropdown-subtle-opener-text,
+			.d2l-dropdown-opener-text {
+				margin-right: 0.3rem;
+				vertical-align: middle;
+			}
+			:host(:dir(rtl)) .d2l-dropdown-opener-text,
+			:host([subtle]:dir(rtl)) .d2l-dropdown-subtle-opener-text {
```
dlockhart, line 95: You'll want `:host([dir="rtl"])` not the `:dir(rtl)` as the `RtlMixin` will sprout the `dir` attribute on the host for you.

components/button-group/button-group.js
```diff
@@ -0,0 +1,162 @@
...
+			:host([dir="rtl"]) .d2l-button-group-container ::slotted(d2l-dropdown:not(.d2l-overflow-dropdown)),
+			:host([dir="rtl"]) .d2l-button-group-container ::slotted(d2l-dropdown-button),
+			:host([dir="rtl"]) .d2l-button-group-container ::slotted(d2l-dropdown-button-subtle),
+			:host([dir="rtl"]) .d2l-button-group-container ::slotted(.d2l-button-group-custom-item) {
+				margin-left: 0.6rem;
+				margin-right: 0;
+			}
+
+			:host([subtle]) .d2l-button-group-container ::slotted(d2l-button-subtle) {
+				margin-right: 0.2rem;
```
dlockhart, line 81: Interesting... subtle buttons have a different margin?

mixins/dynamic-resize-mixin.js
```diff
@@ -0,0 +1,350 @@
...
+
+		if (minToShow !== undefined) {
+			this.minToShow = minToShow;
+		}
+		if (maxToShow !== undefined) {
+			this.maxToShow = maxToShow;
+		}
+	}
+	_chomp(items = this._layoutItems) {
+
```
dlockhart, line 148: The more I look at this mixin, the less sure I am that making it so generic is that useful. For it to work properly, it needs the things inside it to be pretty specific... buttons and stuff. I'm just not sure it'll be that useable beyond butotn-groups.

The fact that it's a mixin also means all this chomping needs to happen outside of the component's `render()` method, which isn't ideal. If all this code was in `button-group` itself, properties could just be updated, which would cause a re-render and the "chomp" would happen there -- so much simpler. 

@dbatiste curious about your thoughts here? We were originally thinking about the navbar and HTML editor "chompers" when hoping to make this generic.

dlockhart (commented):
Got part-way through and realized we should probably discuss the mixin bit first, so don't worry about the nit/docs comments for now.

components/button-group/README.md
```diff
@@ -0,0 +1,51 @@
+# Button Groups
+
```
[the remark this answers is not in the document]
GeorgePlukov: I was considering this but since we handle elements like `d2l-link` I wasn't sure if it made sense. It does seem like link is the exception and in general its buttons/dropdown buttons that it will be handling.

components/button-group/button-group.js
GeorgePlukov: Yeah, since there is no background color/borders around the subtle buttons they look really spread out with the `0.6rem` margins. This was suggested by Jeff so that the space between the button text looks similar to the non subtle buttons

components/button-group/README.md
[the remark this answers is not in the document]
dlockhart: Yeah let's shelve this until after our meeting... I think we may land on something more generic than button-group anyway.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+	}
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._removeEventListeners();
+	}
+	async firstUpdated() {
+		super.firstUpdated();
+
+		// selected elements
+		this._buttonSlot = this.shadowRoot.querySelector('.item-container');
```
dlockhart, line 187: Best practice is to scope CSS class names with the name of the component, so something like `.d2l-overflow-group-slot`. Alternately, since there's only one slot anyway you could just go `this.shadowRoot.querySelector('slot')`. 🤷

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+			this._autoDetectBoundaries(this._slotItems);
+		}
+
+		this._availableWidth = this._container.clientWidth;
+	}
+	render() {
+		let overflowMenu;
+		if (!this._overflowMenuHidden) {
+			overflowMenu = this._getOverflowMenu();
+		}
```
dlockhart, line 212:
```suggestion
		const overflowMenu = !this._overflowMenuHidden ? this._getOverflowMenu() : null;
```

Or alternatively, `_getOverflowMenu()` could just return `null` if `_overflowMenuHidden` is `true`.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+
+		if (changedProperties.get('minToShow')
+			|| changedProperties.get('maxToShow')) {
+			this._chomp();
+		}
+
+		// slight hack, overflowMenu isnt being rendered initially so wait until update
+		// to calculate its width
+		if (!this._overflowMenuWidth) {
```
dlockhart, line 242: Would `firstUpdated` be a more appropriate place to do this calculation?

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+			// this action needs to be deferred until first render of our overflow button
+			requestAnimationFrame(() => {
+				this._chomp();
+			});
+		}
+	}
+	connectedCallack() {
+		this._addEventListeners();
```
dlockhart, line 250: Since there's just the one, I think `_addEventListeners` could just be pulled into here. (same for remove)

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+		this._mini = this.openerType === OPENER_TYPE.ICON || swapToMiniButton;
+		this._chompIndex = this._overflowMenuHidden ? null : showing.count;
+
+		this.dispatchEvent(new CustomEvent('d2l-overflow-group-updated', { bubbles: true, composed: true }));
+	}
+	_convertToDropdownItem(node) {
+		const tagName = node.tagName.toLowerCase();
+		let menuItem;
+		if (tagName === 'd2l-button' || tagName === 'd2l-button-subtle' || tagName === 'button' || tagName === 'd2l-button-icon') {
+			menuItem = this._createMenuItem(node);
```
dlockhart, line 355: Should be able to just `return` from each of these blocks, then remove the `} else { return; } return menuItem;` at the end. A `switch` on the `tagName` may also look a little cleaner, and then handle the "separator" and "custom-item" case separately (which may go away anyway).

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+				menuItem = this._createMenuItem(node);
+			}
+		} else if (node.classList.contains('d2l-overflow-group-custom-item')) {
+			menuItem = this._createMenuItem(node);
+		} else {
+			return;
+		}
+		return menuItem;
+	}
+	_createMenuItem(node) {
```
dlockhart, line 376: It looks like most of these `_create*` methods are just helpers and don't actually reference `this`. So I think they can just be pulled outside of the class to the top of the file as functions.

It doesn't make a huge difference, but that way they'll be treated as static and shared amongst all the instances of this class, vs. having copies of them for each one.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+	}
+	_getLayoutItems(filteredNodes) {
+
+		const items = filteredNodes.map((node) => {
+			const computedStyles = window.getComputedStyle(node);
+			return {
+				type: node.tagName.toLowerCase(),
+				isChomped: false,
+				width: node.offsetWidth
+					+ parseInt(computedStyles.marginLeft.replace('px', ''))
```
dlockhart, line 426: Nit: I've seen this just written as `parseInt(computedStyles.marginLeft) || 0` which works too.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+		});
+
+		return items;
+	}
+	_getOverflowMenu() {
+
+		const moreActionsText = this.localize('components.overflow-group.moreActions');
+		const overflowItems = this._dropdownItems ? this._dropdownItems.slice(this._chompIndex) : [];
+		const menu = html`<d2l-dropdown-menu>
+			<d2l-menu id="overflowMenu" label="${moreActionsText}">
```
dlockhart, line 439: I don't _think_ this `id` is used?

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._removeEventListeners();
+	}
+	async firstUpdated() {
+		super.firstUpdated();
+
+		// selected elements
+		this._buttonSlot = this.shadowRoot.querySelector('.item-container');
+		this._buttonSlot.addEventListener('slotchange', this._handleSlotChange);
```
dlockhart, line 188: This should be able to be added inline in the `render()`:

```html
<slot @slotchange="${this._handleSlotChange}">
```

That should remove the need to remove the event listener later too since Lit will handle it automatically.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+			||  node.querySelector('d2l-dropdown-button-subtle');
+
+		const openerText = node.text || menuOpener.text;
+		const subMenu = node.querySelector('d2l-menu');
+		const subItems = [];
+
+		// iterate through any sub menus and turn them into menu items
+		for (const node of subMenu.children) {
+			subItems.push(this._convertToDropdownItem(node));
+		}
```
dlockhart, line 409:
```suggestion
		Array.from(subMenu.children).map((node) => this._convertToDropdownItem(node));
```

components/overflow-group/test/overflow-group.visual-diff.html
```diff
@@ -0,0 +1,159 @@
...
+					<d2l-button>Button 1</d2l-button>
+					<d2l-button>Button 2</d2l-button>
+					<d2l-button id="last">Button 3</d2l-button>
+				</d2l-overflow-group>
+			</div>
+		</div>
+
+		<!-- add a node to see if it shows the overflow -->
+		<div id="width-not-overflowing" class="visual-diff">
+			<template>
```
dlockhart, line 104: A bunch of these don't seem to be used in the `.js` file?

package.json
```diff
@@ -69,6 +69,7 @@
     "focus-visible": "^5",
     "intl-messageformat": "^7",
     "lit-element": "^2",
+    "lodash-es": "^4.17.20",
```
dlockhart, line 72: Typically we'll just specify `^4` unless there's something we absolutely need in minor/patch releases.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+				flex-wrap: wrap;
+			}
+			:host .d2l-overflow-group-container ::slotted(d2l-button),
+			:host .d2l-overflow-group-container ::slotted(d2l-button-icon),
+			:host .d2l-overflow-group-container ::slotted(d2l-link),
+			:host .d2l-overflow-group-container ::slotted(span),
+			:host .d2l-overflow-group-container ::slotted(d2l-dropdown:not(.d2l-overflow-dropdown)),
+			:host .d2l-overflow-group-container ::slotted(d2l-dropdown-button),
+			:host .d2l-overflow-group-container ::slotted(d2l-dropdown-button-subtle),
+			:host .d2l-overflow-group-container ::slotted(.d2l-overflow-group-custom-item) {
```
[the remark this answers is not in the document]
GeorgePlukov: I removed the references to the `custom-item`! I didn't do a search for that specifically.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,495 @@
...
+ * @slot - Buttons, dropdown buttons, links or other items to be added to the container
+ */
+class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
+
+	static get properties() {
+		return {
+			/**
+			 * Use predefined classes on slot elements to set min and max buttons to show
+			 */
+			autoShow: {
```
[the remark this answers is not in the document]
GeorgePlukov: Yeah when I looked into it I saw that it was using it to style based on primary buttons. I didn't know whether or not we should leave that functionality but I thought it could be useful. These functions just set new values for `minToShow` and `maxToShow` so removing it is rather easy.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,487 @@
...
+		this._mini = this.openerType === OPENER_TYPE.ICON || swapToMiniButton;
+		this._chompIndex = this._overflowMenuHidden ? null : showing.count;
+
+		this.dispatchEvent(new CustomEvent('d2l-overflow-group-updated', { bubbles: true, composed: true }));
+	}
+	_getLayoutItems(filteredNodes) {
+
+		const items = filteredNodes.map((node) => {
+			const computedStyles = window.getComputedStyle(node);
+			// if () {
```
dbatiste, line 409: Looks like some leftover code to be removed?

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,487 @@
...
+			.d2l-dropdown-opener-text {
+				margin-right: 0.3rem;
+				vertical-align: middle;
+			}
+			:host([dir="rtl"]) .d2l-dropdown-opener-text,
+			:host([opener-style="subtle"][dir="rtl"]) .d2l-dropdown-subtle-opener-text {
+				margin-left: 0.3rem;
+				margin-right: 0;
+			}
+			.d2l-overflow-group-container ::slotted([data-is-chomped]) {
```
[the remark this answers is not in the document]
GeorgePlukov: Interesting! That sounds pretty nice, we haven't discussed adding animations at this point but I can keep this in mind for the future

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,487 @@
...
+		this.minToShow = 1;
+		this.openerStyle = OPENER_STYLE.DEFAULT;
+		this.openerType = OPENER_TYPE.DEFAULT;
+		this._mini = this.openerType === OPENER_TYPE.ICON;
+
+		this._slotItems = [];
+	}
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('resize', this._throttledResize);
```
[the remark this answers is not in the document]
GeorgePlukov: I'll switch over to using resize observer, I didn't account for other elements resizing on the page but this makes adding the observers even easier. This should also be more flexible in the future, I left the throttle in for now not sure if you think it's necessary still

components/overflow-group/README.md
```diff
@@ -0,0 +1,44 @@
+# Overflow Groups
+
+## d2l-overflow-group
+The `d2l-overflow-group` element can be used to add responsiveness to a set of buttons, links or menus. Items added to this container element will no longer wrap on to a second line when the container becomes to small but will be added to a dropdown menu with configurable styling.
```
dlockhart, line 4:
```suggestion
The `d2l-overflow-group` element can be used to add responsiveness to a set of buttons, links or menus. Items added to this container element will no longer wrap onto a second line when the container becomes too small but will be added to a dropdown menu with configurable styling. 
```

components/overflow-group/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `auto-show` | Boolean | Automatically determine the min and maximum number of items to show based on which elements have classes `d2l-button-group-show` and `d2l-button-group-no-show`. Please consult the design team when using this attribute. |
+| `min-to-show` | Number | The minimum number of elements to always show. Please consult the design team when using this attribute. |
+| `max-to-show` | Number | The maximum number of elements to show |
+| `opener-style` | String | Set the style of the oferflow menu `default` renders a `d2l-button` while `subtle` will render a `d2l-button-subtle`|
+| `opener-type` | String | Set the opener type to `default` or `icon`, which will permanently render the dropdown menu symbol as `...` |
```
dlockhart, line 42: Wondering if the "..." style should be called "more" since that's what we call the dropdown opener type -- `<d2l-dropdown-more>`? Small nit though.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,468 @@
...
+		super.firstUpdated();
+
+		// selected elements
+		this._buttonSlot = this.shadowRoot.querySelector('slot');
+
+		this._container = this.shadowRoot.querySelector('.d2l-overflow-group-container');
+
+		this._getItems();
+
+		if (this.autoShow) {
```
dlockhart, line 247: `_autoDetectBoundaries` gets called in the `updated()` callback, so I'm wondering if this is necessary in `firstUpdated`? Seems like it might get called twice.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,468 @@
...
+			this._autoDetectBoundaries(this._slotItems);
+		}
+
+		// this._availableWidth = this._container.clientWidth;
```
dlockhart, line 251: Can be removed.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,468 @@
...
+
+		// Slight hack to get the overflow menu width the first time it renders
+		if (!this._overflowMenuWidth) {
+			// this action needs to be deferred until first render of our overflow button
+			requestAnimationFrame(() => {
+				this._chomp();
+			});
+		}
+	}
+	_autoDetectBoundaries(items) {
```
dlockhart, line 294: Not for this PR: I'd still like to see when we port this over to the LMS if we can remove the need for it altogether. Seeing as it's just used once and maybe unnecessarily, it's an extra iteration over all the elements that we can maybe avoid.

dlockhart (approved):
Really like where this ended up! Very clear & clean rendering logic. 🎉

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,468 @@
...
+				itemLayoutOverflowing.isChomped = true;
+			}
+		}
+		const overflowDropdownOverflowing = (showing.width + this._overflowMenuWidth >= this._availableWidth);
+		const swapToMiniButton = overflowDropdownOverflowing && !this._overflowMenuHidden;
+
+		this._mini = this.openerType === OPENER_TYPE.ICON || swapToMiniButton;
+		this._chompIndex = this._overflowMenuHidden ? null : showing.count;
+
+		this.dispatchEvent(new CustomEvent('d2l-overflow-group-updated', { bubbles: true, composed: true }));
```
dbatiste, line 388: Does this event need to bubbled the composed DOM, or is it ok to just bubble the DOM context in which it is dispatched (`this`)?  I know lots of our earlier usages of custom events bubbled the composed DOM but since then we've adopted the more conservative practice of not.  It's fine if we do need it though.

components/overflow-group/overflow-group.js
```diff
@@ -0,0 +1,468 @@
...
+
+		this._getItems();
+
+		if (this.autoShow) {
+			this._autoDetectBoundaries(this._slotItems);
+		}
+
+		this._chomp();
+	}
+	_removeEventListeners() {
```
dbatiste, line 464: Looks like a left-over unused method.

components/overflow-group/overflow-group.js
dlockhart: Agreed yeah. If we can limit this to just the context in which it was dispatched that's ideal.

PR #1186, opened 2021-03-04
adjusting demo to be more like the list use case

dlockhart:
As discussed, this changes the demo a bit to be more like the use cases we're trying to test -- where you have a list of things that can be added/removed and then the whole list itself can be shown or hidden.

Screen Shot 2021-03-04 at 4 26 44 PM

Interestingly the "Add Item" animation is a little clunky at the end, so maybe something to look into there (or I'm using it wrong!).


dlockhart:
Realized the unit tests are using these as well -- will fix.

mixins/demo/visibility-test-list.js
```diff
@@ -0,0 +1,73 @@
...
+	static get properties() {
+		return {
+			_changeTracker: { type: Number, reflect: false },
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: block;
```
AndyWangD2L, line 18: I'm noticing a snap when the list shows/hides. Changing this to be `grid` seems to prevent this from happening.

mixins/demo/visibility-test-item.js
```diff
@@ -0,0 +1,46 @@
...
+		return {
+			number: { type: Number, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				align-items: center;
+				border-bottom: 1px solid #cccccc;
```
AndyWangD2L, line 17: Looks like this is also causing a jump, probably because it's not included in scrollHeight. I'll need to modify the mixin to also include borderTop/Bottom.

mixins/demo/visibility-test-item.js
```diff
@@ -0,0 +1,46 @@
...
+			}
+			d2l-button-icon {
+				flex: 0 0 auto;
+			}
+		`;
+	}
+
+	render() {
+		return html`<span>Item ${this.number}</span>
+			<d2l-button-icon icon="tier1:delete" text="Remove" @click="${this._remove}"></d2l-button-icon>`;
```
AndyWangD2L, line 33: Looks like the finalDummy height isn't accounting for the `d2l-button-icon` in the place it is currently calculating it. I think I'll need to shift the calculation to immediately before the assigning happens.

mixins/demo/visibility-test.js
```diff
@@ -1,19 +1,51 @@
...
+
 	render() {
-		return html`<div>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</div>`;
+		const animateValue = this._listVisibilityChanged ?
+			(this._listVisibility ? 'show' : 'hide') : undefined;
+		const toggleText = this._listVisibility ? 'Hide List' : 'Show List';
+		return html`
+			<d2l-visibility-test-list animate="${ifDefined(animateValue)}"></d2l-visibility-test-list>
+			<d2l-button @click="${this._handleAddItem}">Add Item</d2l-button>
+			<d2l-button @click="${this._handleToggleList}">${toggleText}</d2l-button>
```
AndyWangD2L, line 36: Should the `Add Item` button be disabled while the list is hidden? Adding while the list is hidden leaves a new dummy with `height: 0`, wrapped around a new item with the next number.

mixins/demo/visibility-test-list.js
```diff
@@ -0,0 +1,73 @@
...
+import './visibility-test-item.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { VisibilityMixin } from '../visibility-mixin.js';
+
+export class VisibilityTestList extends VisibilityMixin(LitElement) {
+
+	static get properties() {
+		return {
+			_changeTracker: { type: Number, reflect: false },
```
AndyWangD2L, line 11: Is this used somewhere? It seems like it's only being assigned to, but not assigned to anything else.

AndyWangD2L (commented):
New list demo is looking good!

mixins/demo/visibility-test-list.js
dlockhart: I'm using it to trigger a re-render whenever an item is added or removed. I'll switch it to just manually call `requestUpdate()` to make that clearer.

mixins/demo/visibility-test.js
dlockhart: Agreed -- fixed!

PR #1188, opened 2021-03-04
Remove missing file warning

bearfriend:
Remove the missing file warning to decrease confusion and noise, and fix Firefox.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,5 @@
 {
   "extends": "brightspace/lit-config",
-  "parser": "@babel/eslint-parser",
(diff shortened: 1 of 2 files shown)


.eslintrc.json
```diff
@@ -1,9 +1,5 @@
 {
   "extends": "brightspace/lit-config",
-  "parser": "@babel/eslint-parser",
-  "parserOptions": {
-    "requireConfigFile": false
-  },
```
bearfriend, line 6: This is now in `browser-config` which `lit-config` extends.

dlockhart (approved):

PR #1193, opened 2021-03-05
Bump up z-index for skeleton to provide room for low-level stacking.

dbatiste:
Some components use low-level z-index for their own purposes, and sometimes the value can be slightly higher than the skeleton z-index.  In order to provide some room for components to do their stacking, while still ensuring the skeleton stacks on top, this PR increases the z-index slightly.

### DIFF
--- a/components/skeleton/skeleton-mixin.js
+++ b/components/skeleton/skeleton-mixin.js
@@ -20,7 +20,7 @@ export const skeletonStyles = css`
 		position: absolute;
 		right: 0;
 		top: 0;
-		z-index: 1;
+		z-index: 4;
 	}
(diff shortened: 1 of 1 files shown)


dlockhart (approved):

dbatiste:
The other way I was thinking of resolving this is to have those components initialize a stacking context at their root with `z-index: 0;` so that any children that stack will be relative to `0`.  Although I am not 100% sure this would be sufficient, and this kinda pushes the trickiness onto the consumer and they might not know how to resolve, leading back to adhoc z-index overrides.

```javascript
<some-parent::before (skeleton) z-index: 1>
<some-parent z-index: 0>
    <some-child z-index: 2>
</some-element>
```

PR #1203, opened 2021-03-10
Animation Directive

dlockhart:
The builds on @AndyWangD2L's excellent work [in this pull request](https://github.com/BrightspaceUI/core/pull/1181/) to create a reusable mixin for animating a component when it's shown or hidden. But instead of a mixin, this [uses a directive](https://lit-html.polymer-project.org/guide/creating-directives).

show-hide

Directives have one huge advantage over a mixin: they can be applied to _any_ element, not just custom elements. So whereas with the mixin each web component that needs to be animated would need to extend the mixin, with a directive we can just drop it on whatever element we want -- like ` <div>` or a `<button>`.

There are also lots of changes coming to directives with the next version of lit-html. One of the nicest will be that we'll be able to drop the property name/binding. So using this will become `<div ${hide()}>hide me</div>` instead of `<div .animate="${hide()}">hide me</div>`.

I haven't included a few things from Andy's original pull request:

1. `remove()` functionality: I originally had this, but got worried that by directly manipulating the DOM and removing an element that Lit was tracking, the next time the component re-renders things would get in a weird state. I think we're better off just hiding those DOM nodes and not worrying about cleaning them up. I can be convinced otherwise though!
2. Tests: I'll be honest... I'm not sure how to test this. Visual diff tests don't run animations, so they'd be pretty useless. Also tests run with `prefers-reduced-motion` and as you'll see the code takes a very simple path if that preference is set, so again not sure what tests would accomplish. Open to suggestions here.
3. Events: it seemed like the events were just there for the tests so I haven't included them here. But if folks think there are legit reasons to fire events I can add them back.

I've added some basic focus management though, so that when hiding things that have focus, focus will be moved to the next focusable thing instead of losing it:

focus


directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+		const marginsH = (parseInt(style.marginLeft) || 0) + (parseInt(style.marginRight) || 0);
+		const marginsV = (parseInt(style.marginTop) || 0) + (parseInt(style.marginBottom) || 0);
+
+		let cloneHeight = 0;
+		if (this.clone !== null) {
+			const cloneRect = this.clone.getBoundingClientRect();
+			cloneHeight = cloneRect.height;
+		}
+
+		return {
```
dbatiste, line 104: Hmmm, are the dimensions here all good regardless of whether the element's `box-sizing` is `border-box` or `content-box`?  Do we end up accounting for the border and padding twice if it's `border-box`?

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+		this.state = 'unknown';
+
+	}
+
+	async animate(info) {
+
+		if (this.clone === null) {
+			this.clone = document.createElement('div');
+		}
+		this.elem.parentNode.insertBefore(this.clone, this.elem);
```
dbatiste, line 31: I think consumers ought to be conscious of this, since this might cause their `slotchange` event to be dispatched.  I suspect there is some code that would not be resilient to this... so something we need to be mindful of when building components that assume certain slot content not contain extraneous elements.

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+
+		const id = ++this.id;
+
+		let outerResolve;
+		const onTransitionEnd = (event) => {
+			if (event.target === this.clone) {
```
dbatiste, line 37: When would the target not be the clone?  If it will always be the clone, perhaps we can use the `{once: true}` option which will cleanup the event handler automatically.  Note however, `transitionend` will be dispatched once for each property being animated, but it looks like we only care to run this handler once.

I'm also wondering if we need to consider the same cleanup for the `transitioncancel` event case.  🤔   For that case, I think it would leave the clone in the DOM, and the outer promise would never resolve.  I kinda want to play with this a little bit, since I'm not certain of the behavior of the `show` is run, but then `hide` is run before `show` completes, and vice versa.

directives/animate/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+These directives can be used to animate the showing or hiding of elements in your application.
+
+The animation will slide the element in (or out) of its final position, while transitioning its height and opacity:
+
+![animation](./screenshots/show-hide.gif?raw=true)
+
+## Using the directives
+
+Import the directives you'd like to use and bind them to any element's `animation` property.
```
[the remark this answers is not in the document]
dlockhart: I meant to document this, thanks for the reminder.

directives/animate/README.md
```diff
@@ -0,0 +1,39 @@
+# animate directives
```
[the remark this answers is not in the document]
dlockhart: Yep, fixed.

directives/animate/animate.js
dlockhart: Agreed yeah. I don't see a way around this, since the clone is the thing that has to be present in the DOM taking up space as the host is transitioned in/out. Might just be a case of "fix places that break as they come up".

directives/animate/animate.js
dlockhart: The target-clone check was borrowed from Andy's code, not sure if it's necessary.

The code here that's checking the `id` value is meant to ensure that only the most recent transition's code will run the cleanup. So you can hide/show like crazy (I tested this by slowing the animations down) and the cleanup doesn't happen until the last one finishes.

There may be better ways to handle this with the `transitioncancel` and maybe I should be checking for the different properties. I can look into that a bit more.

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+			this.elem.style.display = 'grid'; // forces margins to collapse during size calculation
+		}
+
+		const rect = this.elem.getBoundingClientRect();
+
+		let top = 0;
+		let left = 0;
+		const offsetParent = this.elem.offsetParent;
+		let e = this.elem;
+		while (e && e !== document && !(e instanceof DocumentFragment)) {
```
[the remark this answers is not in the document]
dlockhart: The purpose of this code is to travel up the DOM searching for the `offsetParent`. `DocumentFragment`s [have no parent](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment), so if we find one we need to stop traveling up in search of the offsetParent. Our `getOffsetParent()` DOM helper does something similar. It will not stop for a shadowRoot.

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+			if (e === offsetParent) {
+				break;
+			}
+			top += e.offsetTop - (e.scrollTop || 0);
+			left += e.offsetLeft - (e.scrollLeft || 0);
+			e = getComposedParent(e);
+		}
+
+		this.elem.style.removeProperty('display');
```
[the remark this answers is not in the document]
dlockhart: Yeah true. I'll try and make this more resilient. There's a bunch of inline styles that this sets and then removes that could likely use a similar treatment.

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+
+		const paddingH = (parseInt(style.paddingLeft) || 0) + (parseInt(style.borderLeftWidth) || 0) + (parseInt(style.paddingRight) || 0) + (parseInt(style.borderRightWidth) || 0);
+		const paddingV = (parseInt(style.paddingTop) || 0) + (parseInt(style.borderTopWidth) || 0) + (parseInt(style.paddingBottom) || 0) + (parseInt(style.borderBottomWidth) || 0);
+		const marginsH = (parseInt(style.marginLeft) || 0) + (parseInt(style.marginRight) || 0);
+		const marginsV = (parseInt(style.marginTop) || 0) + (parseInt(style.marginBottom) || 0);
+
+		let cloneHeight = 0;
+		if (this.clone !== null) {
+			const cloneRect = this.clone.getBoundingClientRect();
```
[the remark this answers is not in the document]
dlockhart: So this particular check will only run if the animation is being changed midway through a run. We need to get the current size of the clone so we can animate its height from that to 0 or full height.

I spent several hours trying to tweak the various read/writes and this was where I landed. The good news is that in almost all scenarios, people aren't going to be animating things when the page first loads -- it's almost always after a user interaction. So the repaints/reflows don't hurt initial page rendering performance.

directives/animate/animate.js
dlockhart: Good catch -- it's broken. Will fix.

directives/animate/animate.js
```diff
@@ -0,0 +1,250 @@
...
+
+	}
+
+}
+
+async function helper(part, action, opts) {
+	if (!stateMap.has(part)) {
+		stateMap.set(part, new AnimationState(part));
+	}
+	const state = stateMap.get(part);
```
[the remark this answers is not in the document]
dlockhart: Yeah I wasn't too worried about the `O(1)` lookup for `WeakMap`, but I can move this around to optimize.

directives/animate/animate.js
AndyWangD2L: Originally, the target-clone check was in the mixin code because the clone wrapped around the content. `transitionend` events fired from the content would bubble up into the clone's listener, and cause the clone to do things when the content's transition ended, rather than when the clone's transition ended. This was especially apparent because of the slight delay between clone and content transitions.

directives/animate/animate.js
dlockhart: Glad you noticed this. I didn't read the [getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) documentation clearly enough:

> In the standard box model, this would be equal to the width or height property of the element + padding + border-width. But if box-sizing: border-box is set for the element this would be directly equal to its width or height.

So it will always return the right thing, meaning we don't need to worry about padding or border. I've removed that code.

directives/animate/animate.js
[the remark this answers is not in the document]
dbatiste: That's a good point about the timing of these animations.  👍

directives/animate/animate.js
dbatiste: Sweet!  👍

directives/animate/animate.js
```diff
@@ -0,0 +1,251 @@
...
+		if (reduceMotion || opts.skip === true) {
+			this.state = 'hidden';
+			this.elem.setAttribute('hidden', '');
+			return;
+		}
+
+		const info = await this.getElemInfo();
+
+		return this.animate({
+			clone: {
```
AndyWangD2L, line 143: Especially with a lot of elements, the content is overflowing during the hide transition. This was solved in the mixin by using `overflow: hidden` since the clone was wrapping the content, but in this case they are siblings... not sure what to do about that here. The `position: absolute` might also cause the content to have a brief period of overlap with other elements while the transition is running & content is overflowing.

directives/animate/animate.js
dbatiste: Cool, yeah.  I think this is something we should just be mindful of when writing/reviewing code that wires up to `slotchange`.  Not that everything needs to handle this case, but just thinking about how the such handlers will deal with these elements if there is a chance the consumer would use these directives in those slots.

directives/animate/animate.js
dbatiste: I think sometimes `transitionend` may not run is `transitioncancel` runs instead... but I'm not 100% clear on when `transitioncancel` runs.  I know it's not an event that we wire up very often though... I think we've sorta been ignoring it.  😬

directives/animate/animate.js
dlockhart: I've tried to capture this behaviour here, but had to slow the animation down 10x to make it really obvious:

overflow

Andy, do you have an example of it looking weird at regular speed?

We could go back to making the clone a wrapper to solve this. My concern there would be then we'd definitely run into slotchange events with parents reacting to the expected children being removed and then re-added each time the animation runs. At least with the clone being a sibling, the actual element doesn't move in the DOM so they just need to ignore the clone.

AndyWangD2L: Yeah, you're right, I'm probably just nit-picking pretty hard, but I think it starts to get more noticeable here. The overall transition happens so quickly that it doesn't cause any issues and people wouldn't notice unless they really looked for it, though.
animate directive 9

If it's easier to solve the slotchange events issues by having them as siblings than with the clone as a wrapper, this overflow probably isn't big enough of an issue to warrant the change.

dlockhart: Cool. Ok let's leave this for now as a known issue to watch for, and if it becomes a bigger problem we can weigh the options again.

directives/animate/animate.js
dlockhart: Here's what I've learned about `transitioncancel`:

- `transitioncancel` is only possible if the animation is delayed or has a duration (ours has a duration)
- only one of `transitioncancel` OR `transitionend` will run
- `transitioncancel` only runs if the animation is stopped before it completes. The scenarios under which this could happen are typically where the transition itself is only being applied under certain scenarios -- like while the user is hovering or while a particular element has a CSS class. As soon as those stop being true, if the tranistion hasn't finished, it'll be cancelled.

Since we're the ones controlling when the `transition` attribute is present, and we only remove it once everything is finished, I think we're OK to ignore it.

directives/animate/animate.js
```diff
@@ -0,0 +1,264 @@
...
+const showTransitionDuration = 300;
+const hideTransitionDuration = 200;
+const moveYValue = 20;
+
+class AnimationState {
+
+	constructor(propertyPart) {
+
+		if (!(propertyPart instanceof PropertyPart) || propertyPart.committer.name !== 'animate') {
+			throw new Error('animation directives must be used with "animation" property');
```
AndyWangD2L, line 16: Should this be `'animation directives must be used with "animate" property'`?

directives/animate/animate.js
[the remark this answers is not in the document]
dlockhart: I've kept this line to undo the setting of `display: grid` from above, but adjust the other code to just restore any inline `style` attribute that may have been present. This is slightly better than simply removing those properties, since previously inline styles would have been lost.

directives/animate/test/animate.test.js
```diff
@@ -0,0 +1,107 @@
...
+		forceFocusVisible(this.shadowRoot.querySelector('#first'));
+	}
+
+	_dispatchEvent() {
+		this.dispatchEvent(new CustomEvent('d2l-animate-test-focus-animate-complete'));
+	}
+
+}
+customElements.define('d2l-animate-test-focus', FocusTestElem);
+
```
dlockhart, line 36: I've added some tests in here to attempt to test some of the weirder edge cases that we'd like to not regress on in the future.

directives/animate/animate.js
```diff
@@ -0,0 +1,268 @@
...
+	async hide(opts) {
+
+		if (this.state === 'hiding' || this.state === 'hidden') {
+			return;
+		}
+		this.state = 'hiding';
+
+		// if focus is inside and was placed by keyboard, move it to next focusable
+		const activeElem = getComposedActiveElement();
+		if (activeElem && activeElem.classList.contains('focus-visible')) {
```
dlockhart, line 141: Slight tweak here to only move focus if focus is visible. This will limit the places that it'll work, but I was finding it weird to click on the "Remove" button and have focus move on me, vs. when using the keyboard having it move was required.

directives/animate/animate.js
```diff
@@ -0,0 +1,268 @@
...
+
+		await new Promise((r) => requestAnimationFrame(r));
+		Object.assign(this.clone.style, info.clone.end);
+		Object.assign(this.elem.style, info.elem.end);
+
+		return new Promise((resolve) => outerResolve = resolve);
+
+	}
+
+	dispatchEvent() {
```
dlockhart, line 69: I've added an event back in, but its name is more generic as I didn't need anything more specific for testing. Not sure how consumers would use this or if they'd need it at all.

dlockhart (commented):
Ok I think I've addressed everything!

directives/animate/animate.js
dlockhart: Yep -- thanks, fixed.

directives/animate/animate.js
```diff
@@ -0,0 +1,273 @@
...
+		Object.assign(this.elem.style, info.elem.end);
+
+		return new Promise((resolve) => outerResolve = resolve);
+
+	}
+
+	dispatchEvent(timeout = 0) {
+		// need a timeout as when motion is reduced
+		// event fires too quickly to listen for
+		// it right after render
```
dlockhart, line 72: It appears as though the unit tests for Edge (Chromium) at least in the BrowserStack environment are running with `prefers-reduced-motion` turned on. In that scenario, the event is firing before the tests can even wire up an event listener. So adding a short timeout that we'll only use in the reduced motion case.

directives/animate/animate.js
```diff
@@ -0,0 +1,273 @@
...
+					position: 'absolute',
+					top: `${info.elem.top}px`,
+					transitionProperty: 'height, opacity, transform',
+					transitionDuration: `${hideTransitionDuration}ms`,
+					transitionTimingFunction: 'ease-out',
+					width: `${info.elem.width}px`
+				},
+				end: {
+					opacity: '0',
+					transform: `translateY(-${moveYValue}px)`
```
bearfriend, line 188: Would it be as simple as `translate${direction}(...` to make this work horizontally as well? I'm thinking about carousels, if that's the right term, like this one in Wave, but probably plenty of use cases.

animate-x

bearfriend: I suppose that would require the ability to specify the value as negative or positive explicitly as well.

bearfriend: ... and possibly events if you wanted to start the fade-in of the next element after the fade-out of the previous element was done.

dlockhart: This directive is only complex due to the craziness we have to do to "slide" the thing in or out while growing/shrinking the clone to fill its space while that's happening.

If all we needed was a simple translation in one direction, I'm not sure you'd need a directive? I'd implement that carousel just using a container that doesn't allow overflow and then when the user arrows translate everything inside it left/right, which could be animated.

bearfriend: Oooh, that makes sense. Ok, so the carousel is a bad example, but I wonder if there would be a need to do it horizontally. I suppose we can cross that bridge when we get to it.

dlockhart: Yeah totally. We could easily extend this in the future by either adding more options to these existing directives (e.g. `show({direction: "right"})`) or yeah by introducing new ones like `slide-in` or whatever.

I think for most animations it should be easy enough for teams to just do them using regular CSS transitions in their components. But for cases where there's a complex series of transitions like this where we want to centralize that logic, creating new animation directives could make sense.

components/list/list.js
```diff
@@ -37,6 +37,9 @@ class List extends LitElement {
 			:host {
 				display: block;
 			}
+			:host([hidden]) {
+				display: none;
+			}
```
dlockhart, line 42: I've changed the demo to just use list directly (removing 2 demo files), but list wasn't handling `hidden` properly so this fixes that.

dbatiste (approved):

PR #1204, opened 2021-03-12
Fix a11y for d2l-input-percent.

dbatiste:
This PR includes a couple of a11y/usability fixes:

* announce "Percent" as part of the label for `d2l-input-percent`.  This required defining a `label-secondary` property on `d2l-input-text` and `d2l-input-number` that is rendered offscreen but is part of the label.

* focus on the input when the invalid icon is clicked because user expectation is that the focus would be applied when clicking in the visible input bounding box.  This is especially noticeable for small inputs, such as number and percent where there is a higher chance the user clicks the invalid icon.


components/inputs/docs/input-text.md
```diff
@@ -27,6 +27,7 @@ The `<d2l-input-text>` element is a simple wrapper around the native `<input typ
 | `disabled` | Boolean | Disables the input |
 | `input-width` | String, default: `100%` | Restricts the maximum width of the input box without impacting the width of the label |
 | `label-hidden` | Boolean | Hides the label visually (moves it to the input's `aria-label` attribute) |
+| `label-secondary` | String | Secondary offscreen label for accessibility |
```
dlockhart, line 30: I guess `description` doesn't quite work the way you want it? It also creates an offscreen thing that hooks it up using `aria-describedby`.

dlockhart (approved):

components/inputs/docs/input-text.md
dbatiste: Yeah, I'd consider description briefly, but it's also leveraged by the tooltip when invalid.  It could potentially be used, but not perfectly since one would have to take precedence.

dbatiste: I've done a bit more experimenting with this using VO and NVDA.  We could actually use the `description` and similarly insert the user-specified value into the validation message for the invalid case.  

Reasons to use `label-secondary`:
1. `aria-describedby` value is not displayed in the list of form controls, for example the VO Rotor.  In this case, having this part of the label is helpful, but not essential. 
2. the value doesn't get announced until after the screen reader announces a bunch of other stuff... which works, but not ideal.

Reasons to use `description`:
1. it doesn't add a seemingly redundant property to the component API
2. consumers might mis-use `label-secondary`, by putting verbose descriptions that should be put into the aria description 

I think I prefer the UX of `label-secondary`, but I am more worried about consumers using it incorrectly, so I _think_ I am leaning towards changing this to just use `description`.  I'd also considered always rendering the user-specified description as I have done here (offscreen part of label), but I don't think we want potentially verbose descriptions showing up as part of labels in the list of form fields such as in VO Rotor.

dlockhart: Cool yeah seems like you've considered the trade-offs -- I'm good with either approach. I do think this could become an interesting pattern that we could use in many places... inputs that have their units displayed right in the input like this. Other examples that come to mind: entering CSS values like `px` or `em` (in a branding UI), currency input, HEX colour values...

Which is making me think -- could the `%` thing just be in a different slot that we introduce? The slot could end up inside the label in the DOM, but we just position it to the right/left similar to the existing slot you've used. That way you may not need to introduce a new lang term for "percent" it would just be "label value %". 🤷

dbatiste:
This turned out to be way more involved than I'd hoped.  In the end, I think this is in a better position in terms of a11y and also flexibility.

components/inputs/input-percent.js
```diff
@@ -74,19 +74,6 @@ class InputPercent extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Rt
 				:host([hidden]) {
 					display: none;
 				}
-				:host([disabled]) [slot="right"] {
```
dbatiste, line 77: No longer needed since we aren't leveraging the slot for this.

components/inputs/input-percent.js
```diff
@@ -169,9 +156,5 @@ class InputPercent extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Rt
 		}
 	}
 
-	_handleSymbolClick() {
```
dbatiste, line 172: No longer needed here, but this gets moved to d2l-input-text for the same usability reason.

components/inputs/input-text.js
```diff
@@ -307,17 +329,17 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 						title="${ifDefined(this.title)}"
 						type="${this._getType()}"
 						.value="${this.value}">
-					<div id="first-slot"><slot name="${firstSlotName}" @slotchange="${this._onSlotChange}"></slot></div>
```
dbatiste, line 310: Rename these divs and include the unit within them, which makes measurement for the padding offset easier.

components/inputs/input-text.js
```diff
@@ -307,17 +329,17 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
...
+					${ (!isValid && !this.hideInvalidIcon && !this._focused) ? html`<div class="d2l-input-text-invalid-icon" style="${styleMap(invalidIconStyles)}" @click="${this._handleInvalidIconClick}"></div>` : null}
+					${ this.validationError ? html`<d2l-tooltip for=${this._inputId} state="error" align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>` : null }
+				</div><div id="after-slot" class="d2l-skeletize" ?hidden="${!this._hasAfterContent}"><slot name="after" @slotchange="${this._handleAfterSlotChange}"></slot></div>
 			</div>
 			${offscreenContainer}
 		`;
 		if (this.label && !this.labelHidden) {
 			return html`
-				<label class="d2l-input-label d2l-skeletize" for="${this._inputId}">${this.label}</label>
+				<label class="d2l-input-label d2l-skeletize" for="${this._inputId}">${this.label}${this.unit ? html`<span class="d2l-offscreen"> ${this.unit}</span>` : ''}</label>
```
dbatiste, line 342: Got some crazy whitespace between the label and the required * unless I removed all whitespace here.

components/inputs/input-text.js
```diff
@@ -332,6 +354,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 				this.requestValidate(false);
 				this.setFormValue(this.value);
 				this._prevValue = (oldVal === undefined) ? '' : oldVal;
+			} else if (prop === 'unit') {
+				if (this.unit && this.unit !== '%') {
```
dbatiste, line 358: As discussed, when/if we support other units, we will either translate them internally for the label or expose another property allowing the consumer to specify.  For "%", the symbol is announced correctly by screen readers.

components/inputs/input-text.js
```diff
@@ -393,6 +420,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		return 'text';
 	}
 
+	_handleAfterSlotChange(e) {
```
dbatiste, line 423: I took the opportunity to rename handlers so we have consistent name, at least within this file.

components/inputs/input-text.js
```diff
@@ -444,27 +480,29 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		this._hovered = false;
 	}
 
-	_onAfterSlotChange(e) {
-		const afterContent = e.target.assignedNodes({ flatten: true });
-		this._hasAfterContent = (afterContent && afterContent.length > 0);
-	}
-
-	_onSlotChange(e) {
```
dbatiste, line 452: By measuring the container instead of the slotted content individually, we are able to avoid `getComputedStyle`, but it also makes the width calculation simpler, not just for the unit case, but also for the simpler case that was originally supported.

dlockhart (approved):

PR #1212, opened 2021-03-15
Ignore - Switch card border to red

svanherk:
Triggering visual-diff to see if the latest token is bad.


svanherk:
Working as expected (for now... 🤨)

PR #1218, opened 2021-03-17
Allow slot inception for dropdown openers.

dbatiste:
This PR allows the contents of a dropdown to be distributed from nested slots.  For example:

```html
<special-dropdown>
  --shadowDOM
    <d2l-dropdown>
      --shadowDOM
        <slot></slot>
      <button class="d2l-dropdown-opener">cool stuff</button>
      <slot></slot>
    </d2l-dropdown>
  <d2l-dropdown-menu>...</<d2l-dropdown-menu> or <d2l-dropdown-content>
</special-dropdown>
```

This is needed for the HtmlEditor.

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -100,7 +100,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__getContentElement() {
-		return this.shadowRoot.querySelector('slot').assignedNodes().filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
+		return this.shadowRoot.querySelector('slot').assignedNodes({ flatten: true })
+			.filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
 	}
 
 	__onClosed() {


dlockhart:
In your example, doesn't the dropdown-menu or dropdown-content still need to be inside `<d2l-dropdown>`?

dbatiste:
> In your example, doesn't the dropdown-menu or dropdown-content still need to be inside `<d2l-dropdown>`?

Yep.  So with `flatten: true` the consumer can write:
```
<special-dropdown>
  <d2l-dropdown-menu>...</<d2l-dropdown-menu>
</special-dropdown>
```
and the `d2l-dropdown-menu` will get distributed to the slot where it belongs.

dbatiste:
An alternative is for `<some-dropdown>` to extend `DropdownOpenerMixin` like core's built in openers, but this should be a viable way to use `d2l-dropdown` as well.

dbatiste:
Hmmm, I actually like using the mixin better I think, since it eliminates a web component from the tree.

dbatiste:
We may want to eventually add this, but closing this PR and opening a different one for a fix to the content slot resolution.

PR #1219, opened 2021-03-17
Add getContentSlot method that mixin consumer can override.

dbatiste:
Consumers of the mixin that have more than one slot need the ability to resolve the slot containing the dropdown content for the opener.  Otherwise, the opener-mixin may try to use the wrong slot, and the dropdown will not open as expected.

In the same way the consumer can override the default for resolving the opener element, they can provide an override to get the slot.

This is alternate solution to this other PR: https://github.com/BrightspaceUI/core/pull/1218

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -79,6 +79,10 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		opener.focus();
 	}
 
+	getContentSlot() {
+		return this.shadowRoot.querySelector('slot');
+	}
+
 	getOpener() {
 		return this;
 	}
@@ -100,7 +104,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__getContentElement() {
-		return this.shadowRoot.querySelector('slot').assignedNodes().filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
+		return this.getContentSlot().assignedNodes()
+			.filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
 	}
 
 	__onClosed() {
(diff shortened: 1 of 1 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -100,7 +112,8 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__getContentElement() {
-		return this.shadowRoot.querySelector('slot').assignedNodes().filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
+		return this.getContentSlot().assignedNodes()
+			.filter(node => node.hasAttribute && node.hasAttribute('dropdown-content'))[0];
```
dlockhart, line 116: Sorry to suggest yet another alternate, but could this code not just iterate over ALL the slots, looking for the `dropdown-content` attribute? Like use `querySelectorAll('slot')`.

dbatiste: Yep it could, I did consider that option as well, but wasn't super keen on repeatedly calling `assignedNodes` looking for the content when the subclass knows best how to provide it.  I don't feel strongly either way though, since it will rarely get used in either case.

dlockhart: Yeah like in the current cases where there's only 1 slot it'll behave exactly the same and only where there's multiple slots will it do a bit more work. Was just thinking it'd be simpler for folks to implement and less things to document, but good either way.

Alternately, can it assume that it's always the main slot? So you could look for `slot:not([name])` to find the one without a name. 🤷

dbatiste: Yeah I did consider that option briefly too.  Initially I was apprehensive about the potential for breaking a current consumer, however it looks like the only usages of `DropdownOpenerMixin` are the core dropdown openers, and they all used unnamed slots so I _think_ it would be a safe change. 
 It would require that any `DropdownOpenerMixin` consumer reserve the unnamed slot for the dropdown content, which is probably an acceptable constraint.  I think I'd be more inclined going this route (even though it limits the consumer) than iterating all the slots.  🤷‍♂️

dbatiste: Ok, I went the `slot:not([name])` route.

dlockhart (approved):

PR #1221, opened 2021-03-18
Scy/de42729/improve template separator label

Scy-D2L:
Previously with no explicit labels screen readers on windows read "separator {pixel width of secondary pane}" when focusing on the divider between the panes and didn't offer any indication that it was interactive or why it was a tabstop.

With these changes it should read "`adjustable split-view` separator {pixel-width of secondary pane} `arrow left or right to adjust the size of the view panels`" (or up and down when vertical).


templates/primary-secondary/primary-secondary.js
```diff
@@ -988,6 +985,17 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(LitEle
 				}
 			}
 		}
+		if (!this._secondary) {
```
Scy-D2L, line 988: This approach resolves the elements not existing during `connectCallback` - but does try to re-add the listeners on every update. This might not be an issue because [duplicate event listeners by reference won't stack up](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#multiple_identical_event_listeners) and there doesn't seem to be a way to check if they exist, but if we're concerned we could add a boolean that flips when they're actually added and on `disconnectedCallback`

vix633: Seems legit. @dbatiste I pointed Scy in this direction, let us know if its smelly.

Scy-D2L:
I can't test what this sounds like using VO on mac, but I know @dbatiste said that before the changes it was much more verbose and had its own description of the keybinds (not supplied by us). If someone can check to make sure it isn't doing anything crazy with the label and the description before this proceeds it would be appreciated ;)

templates/primary-secondary/primary-secondary.js
dbatiste: Yeah this is definitely a little precarious.  If one of those resizer objects is ever modified, to use an anonymous function for one of its handlers, to move the `bind` (which creates a new function) out of the constructor, or if a new resizer object is created, then the function instances would be different and the browser would not discard them as described.  This code really relies on someone changing that code with seemingly safe changes (which is somewhat abstracted from here) to know this code depends on them not making those changes.

I am certain the original author had reasons for writing it this way, however it would be much preferred to wire/unwire the `window` handlers in `connectedCallback`/`disconnectedCallback`, the local handlers in `render` with the `@` directive, and any handlers for `this` in `firstUpdated`.  Unfortunately it looks like that may be a little more involved.

It looks like the original code was relying on rAF to delay wire-up.  I think this was kind of risky already.  Did you ever get to the root of why adding the Localizer mixin caused the event wire-up to break in the connectedCallback?  Perhaps we just need to wait a task using `setTimeout` instead of rAF.

Aside from that, I see what you are doing now as one option.  Another option is to just do the wire-up in `firstUpdated` and if the component ever does need to support being removed and then re-added to the DOM, we cross that bridge then.  Neither of those options are amazing.

dbatiste:
> I can't test what this sounds like using VO on mac, but I know @dbatiste said that before the changes it was much more verbose and had its own description of the keybinds (not supplied by us). If someone can check to make sure it isn't doing anything crazy with the label and the description before this proceeds it would be appreciated ;)

Yep, I can test this out this afternoon.

templates/primary-secondary/primary-secondary.js
Scy-D2L: I couldn't definitively find what was causing the problem when I went through it, but there were a few tests (nothing reproducible) where it magically disappeared, which makes me think it might be a race condition. I didn't get too far into debugging, though, it's pretty involved with the number of things happening here. I don't know the impact the localization logic has on when the dom loads but it does some asynchronous things in its connectedCallback and doesn't explicitly wait for them. I'm hoping not to sink too much more time into it right now though.

If we're ok with it not supporting connecting/disconnecting for now (that doesn't seem like an intuitive thing to do with a template like this anyway?) then firstUpdated seems like the cleanest quick resolution for now. Otherwise I can throw in the boolean I mentioned to allow it to reconnect this way without duplicating handlers, which is a bit messier but doesn't compromise the functionality. Any preference @dbatiste, since future work is more likely to land with the platform team?

dbatiste:
> Yep, I can test this out this afternoon.

Ok, I've done some testing with this.  VO is actually really frustrating to use make this work.  Typically, to operate the divider, the user is supposed to use Ctrl+Opt+Shift+Down, then use Ctrl+Opt followed by left/right arrow keys to move it.  However, I've spent about 20-30min messing with it, and I am not unable to get it to work at all with the original code (while it worked ok a couple days ago), and only about 10-20% of the time with the new code depending on where I can manage to get the virtual "focus", and only after using Ctrl+Opt+Shift+Down.  With VO active, the arrow keys don't function as they would normally... hence these extra keys they need to press in order to interact with the divider.

Aside from hardly being able to get it to work, the message no longer tells the user they need to use Ctrl+Opt+Shift+Down, which could be problematic for VO users unless they are experienced/expert VO users.  I wonder if Carin has a Mac she can use... she might know some magic tricks to make it do what it's supposed to.  Also, I am happy to screenshare if it helps.

dbatiste:
Actually correction... it does still announce the Ctrl+Opt+Shift+Down/Up after everything else. So, it's probably no more frustrating that it already was.  🤷‍♂️

templates/primary-secondary/primary-secondary.js
dbatiste: Yeah that sounds fine.

templates/primary-secondary/primary-secondary.js
```diff
@@ -889,26 +892,20 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(LitEle
...
-		const divider = this.shadowRoot.querySelector('.d2l-template-primary-secondary-divider');
+		this._namespace = 'templates.primary-secondary';
+		this._keyboardDescId = getUniqueId();
 
-		this._desktopKeyboardResizer.connect(divider);
-		this._desktopMouseResizer.connect(divider);
-		this._mobileKeyboardResizer.connect(divider);
-		this._mobileMouseResizer.connect(divider);
-		this._mobileTouchResizer.connect(secondaryPanel);
+		this._secondary;
```
dbatiste, line 899: Unnecessary to define these here... they will be undefined regardless.

dbatiste (approved):
Just a couple variables in constructor without assigned that can be removed.

templates/primary-secondary/primary-secondary.js
```diff
@@ -936,12 +933,14 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(LitEle
 		const scrollClasses = {
 			'd2l-template-scroll': isWindows
 		};
+		const keyboardHelpText = this._isMobile ? this.localize(`${this._namespace}.keyboardVertical`) : this.localize(`${this._namespace}.keyboardHorizontal`);
```
dlockhart, line 936: As handy as it is, can you not use `_namespace` and just use the full string? It helps when we're doing searches for terms to see if they're used if we can search for the full term name.

Scy-D2L: Sure, I was following the pattern I'd seen elsewhere in core but I'll switch it up. :)

dlockhart: Oh no! Haha now I need to stamp those out.

dlockhart (approved):

PR #1236, opened 2021-03-25
Fix overflow group test flakiness, use waitForTimeout to delay test execution

GeorgePlukov:
This just delays the test by a few hundred milliseconds so that the icons will definitely have time to load in before we start taking screen shots. 

There is probably a better fix we can put in here to avoid this being an issue with icons in general. I think the reason we are seeing these tests being flaky and not the dropdown-openers ones as well is because the dropdown menu gets rendered later (once we determine its necessary for the overflow group to have it).

### DIFF
--- a/components/overflow-group/test/overflow-group.visual-diff.js
+++ b/components/overflow-group/test/overflow-group.visual-diff.js
@@ -12,6 +12,8 @@ describe('d2l-overflow-group', () => {
 		browser = await puppeteer.launch();
 		page = await visualDiff.createPage(browser);
 		await page.goto(`${visualDiff.getBaseUrl()}/components/overflow-group/test/overflow-group.visual-diff.html`, { waitUntil: ['networkidle0', 'load'] });
+		// some tests were being excecuted before icons had time to load asnycronously causing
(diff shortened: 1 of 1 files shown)


components/overflow-group/test/overflow-group.visual-diff.js
```diff
@@ -12,6 +12,7 @@ describe('d2l-overflow-group', () => {
 		browser = await puppeteer.launch();
 		page = await visualDiff.createPage(browser);
 		await page.goto(`${visualDiff.getBaseUrl()}/components/overflow-group/test/overflow-group.visual-diff.html`, { waitUntil: ['networkidle0', 'load'] });
+		await page.waitForTimeout(500);
```
dlockhart, line 15: Could you throw a comment next to this just explaining that it's to avoid flake with the icons not being loaded? Thinking about future-us seeing this and being like "hmm.... delete delete delete".

GeorgePlukov: Yeah, added a comment! This does seem ripe for deleting without realizing what its doing.

dlockhart (approved):

PR #1237, opened 2021-03-25
Minor fixes for overflow group

GeorgePlukov:
Changes: 
- Pass click handlers through so menu-items fire their respective buttons click handlers through the dropdown menu.
- change the way throttle is being imported to prevent other packages loading 
- changes incorrect class name in the demo page

### DIFF
--- a/components/overflow-group/overflow-group.js
+++ b/components/overflow-group/overflow-group.js
@@ -34,6 +34,7 @@ function createMenuItem(node) {
 	const childText = node.text || node.firstChild && (node.firstChild.label || node.firstChild.text || node.firstChild.textContent.trim());
 	const disabled = node.disabled;
 	return html`<d2l-menu-item 
+		@click=${node.onclick}
 		?disabled=${disabled}
 		text="${childText}">
 	</d2l-menu-item>`;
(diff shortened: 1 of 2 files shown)


components/overflow-group/overflow-group.js
```diff
@@ -15,7 +15,7 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
-import { throttle } from 'lodash-es';
+import throttle from 'lodash-es/throttle';
```
dlockhart, line 18: Interesting. So before was it important all the things?

components/overflow-group/overflow-group.js
```diff
@@ -34,6 +34,7 @@ function createMenuItem(node) {
 	const childText = node.text || node.firstChild && (node.firstChild.label || node.firstChild.text || node.firstChild.textContent.trim());
 	const disabled = node.disabled;
 	return html`<d2l-menu-item 
+		@click=${node.onclick}
```
dlockhart, line 37: I think you want the `d2l-menu-item-select` event, as the click event likely won't work with the keyboard. Also just passing `node.onclick` likely won't work in all scenarios like if a click event handler was added in JS. The [old code](https://github.com/BrightspaceUI/button-group/blob/master/d2l-button-group-responsive-behavior.js#L311) might be the best way.

components/overflow-group/overflow-group.js
GeorgePlukov: Yeah not sure how I didn't notice this. For the tree shaking to work properly you have to import the individual function file, once I started testing in the lms it became obvious

dlockhart (approved):

PR #1240, opened 2021-03-26
feat: adding trailing-zeroes functionality to input-number

dlockhart:
This adds an undocumented "trailing zeroes" feature to our numeric input, allowing someone to type in as many zeroes as they'd like beyond the decimal (but below `max-fraction-digits`):

Screen Shot 2021-03-26 at 5 19 58 PM

The only place that will use this is [Arithmetic and Significant Figures quiz question types](https://search.d2l.dev/search?full=ShowTrailingZeroes&project=lms), which can be configured to require that the user give a number to a specified precision. If they don't type enough digits -- which can include zeroes -- they get the answer wrong. So `1.29` could be incorrect but `1.2900` would be correct.

Normally we'd use the `min-fraction-digits` to have the numeric input always include a certain number of trailing zeroes, like for the price of something ($`1.00`). The difference here is that we want to show as many trailing zeroes as the user types in -- that is, we don't know how many we want ahead of time.

This will get used by setting two attributes:
```html
<d2l-input-number trailing-zeroes value-trailing-zeroes="2.00">
```

I needed to introduce a new `value-trailing-zeroes` attribute because `value` is a `Number` so as soon as it gets set (via attribute or property) the original string is lost and I can't see how many zeroes there were. I need the additional `trailing-zeroes` Boolean so that the input can be value-less and still work when the user interacts with it.


components/inputs/input-number.js
```diff
@@ -1,20 +1,52 @@
...
 import { FormElementMixin } from '../form/form-element-mixin.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 
-function formatValue(value, minFractionDigits, maxFractionDigits) {
+function formatValue(value, minFractionDigits, maxFractionDigits, showTrailingZeroes, numDecimalDigits) {
+	let adjustedMinFractionDigits = minFractionDigits;
+	if (showTrailingZeroes && numDecimalDigits > 0) {
```
dlockhart, line 12: To make trailing zeroes work, we pass in an overridden `minFractionDigits` to our number formatting code based on how many digits they typed in beyond the decimal. This code here just makes sure that we're still not exceeding the `max-fraction-digits` setting (if set).

components/inputs/input-number.js
```diff
@@ -1,20 +1,52 @@
...
+	}
 	const options = {
 		maximumFractionDigits: maxFractionDigits,
-		minimumFractionDigits: minFractionDigits
+		minimumFractionDigits: adjustedMinFractionDigits
 	};
 	return formatNumber(value, options);
 }
 
+function countDecimalDigits(value) {
```
dlockhart, line 28: This counts the number of digits beyond the decimal separator.

components/inputs/input-number.js
```diff
@@ -78,13 +114,20 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+					this.trailingZeroes,
+					this._numDecimalDigits
+				);
 				this._value = parseNumber(this._formattedValue);
 			}
 		} catch (err) {
 			this._formattedValue = '';
 			this._value = undefined;
 		}
+		this._numDecimalDigits = 0;
```
dlockhart, line 130: We need to reset this so that subsequent calls to set the `value` property don't magically sprout the previous value's trailing zeroes count.

components/inputs/input-number.js
GeorgePlukov: Okay cool, I thought this was potentially running on formatted user input values. I didn't realize you have to pass `notation: 'scientific'` to NumberFormat for it to return the `e+` format or in this case `Exx`

margaree (approved):

PR #1247, opened 2021-03-30
Update dialogs so that descendants can be made fullscreen.

dbatiste:
This PR updates dialogs so that descendants can be made fullscreen.  The approach taken here is to keep track of the number of descendants that are fullscreen, and then making the dialog itself fullscreen. Descendants signal that they want fullscreen by dispatching a `d2l-fullscreen-within` event with the new state. 

Alternative sol'n attempted to update to `position: static` and `transform: none`, and while this approach also worked, there is a delay and the user can see the position of the dialog change which is undesirable.


components/dialog/dialog-mixin.js
```diff
@@ -260,6 +262,11 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		this._focusFirst();
 	}
 
+	_handleFullscreenWithin(e) {
+		if (e.detail.state) this._fullscreenWithin += 1;
```
dbatiste, line 266: This code weakly keeps track of whether or not there are fullscreen elements within.  It assumes that those descendants will only dispatch the event once per state.  If this proves unreliable, we can track references to the nodes themselves.  I don't expect this to be used very often - the editor might end up being the only case.

dlockhart (approved):
Only other approach I thought of would be to include a promise in the initial event that resolves when it no longer wants to be fullscreen. Seems similar though.

dbatiste:
> Only other approach I thought of would be to include a promise in the initial event that resolves when it no longer wants to be fullscreen. Seems similar though.

🤔 That's an interesting option.  I'm not sure which approach I prefer.  I think I will go with this for the moment and may change it later.

PR #1256, opened 2021-04-06
Overflow group causing warnings when running tests in other d2l repositories

GeorgePlukov:
When running unit tests in other repos occasionally the tests would fail with the error: 
`Error: ResizeObserver loop limit exceeded (http://localhost:9876/context.html?v=home%2C0&cvhf=0%2C1&caf=&undefined=&cgf=20%2C40&daf=7&tcgf=bottomRight#:0)`

The error essentially means that the ResizeObserver was not able to deliver all of its observations within the same animation frame. This is a non-fatal error but would cause the tests to fail, I did a bit of investigating and it seems like replacing throttle with requestAnimationFrame is suitable in this scenario. They should have a very similar effect in this case since rAF targets 60fps which is about 16ms (very similar to the throttle value). 

There is another option of ignoring the errors within the tests but I thought it would be bothersome to have to do that for every repo we started seeing this in.

### DIFF
--- a/components/overflow-group/overflow-group.js
+++ b/components/overflow-group/overflow-group.js
@@ -226,7 +226,7 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		super();
 		this._handleResize = this._handleResize.bind(this);
 
-		this._throttledResize = throttle(this._handleResize, 15);
+		this._throttledResize = (entries) => requestAnimationFrame(this._handleResize(entries));
 
 		this._overflowHidden = false;
 		this.autoShow = false;
@@ -249,8 +249,8 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 
 		this._getItems();
 
-		this.resizeObserver = new ResizeObserver(this._throttledResize);
-		this.resizeObserver.observe(this._container);
+		this._resizeObserver = new ResizeObserver(this._throttledResize);
+		this._resizeObserver.observe(this._container);
 	}
 	render() {
 		const overflowMenu = this._getOverflowMenu();
@@ -449,7 +449,7 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		return filteredNodes;
 	}
 	_handleResize(entries) {
-
+		console.log(true)
 		this._availableWidth = Math.ceil(entries[0].contentRect.width);
 
 		this._chomp();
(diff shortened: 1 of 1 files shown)


components/overflow-group/overflow-group.js
```diff
@@ -15,7 +15,6 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
-import throttle from 'lodash-es/throttle';
```
GeorgePlukov, line 18: Should I just remove the lodash-es import until we have another use case for it?

margaree:
So we do have a tool to suppress this error which got added here https://github.com/BrightspaceUI/core/pull/1101/files and is in the BrightspaceUI initializer, so it would get used in new repos. If this strategy works though and seems safe then that sounds alright to me.

components/overflow-group/overflow-group.js
dlockhart: I think yeah, might as well remove it entirely. We can always re-introduce it later if it's needed.

components/overflow-group/overflow-group.js
```diff
@@ -249,8 +248,8 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 
 		this._getItems();
 
-		this.resizeObserver = new ResizeObserver(this._throttledResize);
-		this.resizeObserver.observe(this._container);
+		this._resizeObserver = new ResizeObserver(this._throttledResize);
+		this._resizeObserver.observe(this._container);
```
dlockhart, line 252: If `this._resizeObserver` is never referenced outside of here, could it just be a local variable?

dlockhart (commented):
Cool -- we end up seeing this error a lot in production as well, to the point where we ignore it from our error log tracking metrics. If simply wrapping things in a rAF solves it, that's good to know.

GeorgePlukov:
Yeah I was thinking it would probably get caught in whatever production js error logging we had and could get annoying. Didn't realize it was already a thing happening! It seems like the consensus is that this should prevent the error from firing in the first place, but it's been difficult to reproduce it with `npm link` locally

package.json
```diff
@@ -69,7 +69,6 @@
     "focus-visible": "^5",
     "intl-messageformat": "^7",
     "lit-element": "^2",
-    "lodash-es": "^4",
```
dlockhart, line 72: I think you'll need to update `package-lock` too.

dlockhart (approved):

dbatiste:
Was it a specific browser that was failing for this?  Legacy Edge doesn't support ResizeObserver so it would be relying on the ponyfill, which hacks together some things to provide an RO API.

PR #1257, opened 2021-04-06
Awang/animation directive fixes

AndyWangD2L:
This PR aims to fix some issues that I noticed [while trying to use this directive with Discover's rule picker](https://github.com/BrightspaceHypermediaComponents/foundation-components/commits/awang/us126118-animation-directive-rule-picker), as well as some other issues that I found.

- At the end of the show transition, the "Add Condition" button would snap back. This was because the absolute positioned elem used a `top` value of `elem.offsetTop`, but  `offsetTop` is the distance to the border, whereas `top` position refers to the top of the margin. Subtracted additional `marginTop` from `top` to fix this 
- I removed the use of `display: grid` for calculating the clone's height. I found that in certain situations, this was actually causing an issue. For example, this forces children (i.e. grid items) to have un-collapsed margins, despite being originally collapsed. This caused improper calculations in height. This has been been replaced with using temporary padding on the elem to force margins to but un-collapsed, and then calculating the top and bottom margins individually from there, and applying that to the clone's height. This solution avoids changing the display style, so that we know exactly what will happen to the margins/padding when calculating clone height.
- In the rule picker, it was possible to add a condition, and exit the dialog fast enough such that the `transitionend` event won't fire. This PR adds a 'transitioncancel' event to handle that scenario.
- `styleAttr` would act very weirdly if we hide -> show while still in hide transition. The elem would still have inline styling from the hide animation, so show's styleAttr would be based on that. Handled this case.

With these out of the way, the rule picker looks mostly great. An issue I noticed but wasn't able to fix (and might not affect the rule picker, but would possibly affect other things):
- If there is margin collapse between sibling elements (for example, in the demo's case, if there is `margin-bottom` on the checkbox, and `margin-top` on the list), snapping still occurs. Setting the `margin-bottom` to 0 on the checkbox, and letting the list have any value for `margin-top` seems to avoid snapping, so I suspect that sibling margin-collapse is the issue here.

### DIFF
--- a/directives/animate/animate.js
+++ b/directives/animate/animate.js
@@ -4,8 +4,8 @@ import { isComposedAncestor } from '../../helpers/dom.js';
 
 const stateMap = new WeakMap();
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
-const showTransitionDuration = 3000;
-const hideTransitionDuration = 2000;
+const showTransitionDuration = 300;
+const hideTransitionDuration = 200;
 const moveYValue = 20;
 
 class AnimationState {
--- a/directives/animate/animate.js
+++ b/directives/animate/animate.js
@@ -118,7 +118,7 @@ class AnimationState {
 		if (paddingTopOriginal) {
 			this.elem.style.paddingTop = `${paddingTopOriginal}px`;
 		} else {
-			this.elem.style.removeProperty('padding-top')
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Tried this out and this looks good to me! There are so many possible scenarios to trip things up... I think the best we can hope for is for it to work in the most common ones and we can document any known limitations. Hopefully folks can modify their usages to avoid them.

PR #1260, opened 2021-04-06
fix: properly handle min/max fraction digits

dlockhart:
Lots of tricky stuff at play here, so would definitely appreciate feedback.

This fixes some major issues with how `value` (and `valueTrailingZeroes`), `minFractionDigits` and `maxFractionDigits` work together. These were problems before the trailing-zeroes stuff, but we just weren't testing/looking for it.

The problem is this: while the element is initializing, property setters get called out of order. So initially, the constructor sets default values -- which may be `undefined`. Then one by one, if an attribute is set, the property setter will get called with that attribute. But if you have properties that are interdependent on each other -- like is the case here where value gets rounded based on `maxFractionDigits`, this is problem.

As an example, I noticed that `<d2l-input-number value="1.23456" max-fraction-digits="10">` would cause the value to immediately get trimmed to `1.235` since the default `maxFractionDigits` is `3`. Then `maxFractionDigits` would get set to `10` but it would be too late -- we've already rounded `value` and lost the original precision and can't get it back.

To fix this, `_value` (and `_valueTrailingZeroes` if that crazy setup is being used) will always store the full precision. It's only in the getter that we do the rounding. And then in the setters for max/min fraction digits we just have to update the formatted value.

I tried several iterations of this fix, including storing separate `_value` and `_valuePrecise` and this is the one I landed on. But definitely open to other ideas!

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -40,6 +40,15 @@ function countDecimalDigits(value, useLocaleDecimal) {
 
 }
 
+// from: https://medium.com/swlh/how-to-round-to-a-certain-number-of-decimal-places-in-javascript-ed74c471c1b8
+function roundPrecisely(val, maxFractionDigits) {
+	const isNegative = val < 0;
+	const positiveValue = Math.abs(val);
+	// eslint-disable-next-line prefer-template
+	const roundedValue = Number(Math.round(positiveValue + `e${maxFractionDigits}`) + `e-${maxFractionDigits}`);
+	return isNegative ? -Math.abs(roundedValue) : roundedValue;
+}
+
 class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitElement))) {
 
 	static get properties() {
@@ -63,7 +72,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 			trailingZeroes: { type: Boolean, attribute: 'trailing-zeroes' },
 			unit: { type: String },
 			value: { type: Number },
-			valueTrailingZeroes: { String, attribute: 'value-trailing-zeroes' },
+			valueTrailingZeroes: { type: String, attribute: 'value-trailing-zeroes' },
 			_formattedValue: { type: String }
 		};
 	}
@@ -100,6 +109,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 	}
 
 	get maxFractionDigits() {
+		// emulate Intl's default maxFractionDigits behaviour
 		return this._maxFractionDigits ?? Math.max(this.minFractionDigits, 3);
 	}
 	set maxFractionDigits(val) {
@@ -129,8 +139,7 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 
 	get value() {
 		if (this._value === undefined) return undefined;
-		const precision = Math.pow(10, this.maxFractionDigits);
-		return Math.round((this._value + Number.EPSILON) * precision) / precision;
+		return roundPrecisely(this._value, this.maxFractionDigits);
 	}
 	set value(val) {
 		const oldValue = this.value;
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -7,43 +7,36 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 
-function formatValue(value, minFractionDigits, maxFractionDigits, showTrailingZeroes, numDecimalDigits) {
-	let adjustedMinFractionDigits = minFractionDigits;
-	if (showTrailingZeroes && numDecimalDigits > 0) {
-		adjustedMinFractionDigits = Math.max(minFractionDigits || 0, numDecimalDigits);
-		if (maxFractionDigits !== undefined) {
-			adjustedMinFractionDigits = Math.min(
-				adjustedMinFractionDigits,
-				maxFractionDigits
-			);
-		}
+function formatValue(value, options, numDecimalDigits) {
+	if (value === undefined) return '';
+	if (numDecimalDigits > 0) {
+		options.minimumFractionDigits = Math.min(
+			Math.max(options.minimumFractionDigits, numDecimalDigits),
+			options.maximumFractionDigits
+		);
 	}
-	const options = {
-		maximumFractionDigits: maxFractionDigits,
-		minimumFractionDigits: adjustedMinFractionDigits
-	};
 	return formatNumber(value, options);
 }
 
-function countDecimalDigits(value) {
+function countDecimalDigits(value, useLocaleDecimal) {
 
 	if (value === undefined || value === null) {
-		return undefined;
+		return 0;
 	}
 
-	const descriptor = getNumberDescriptor();
+	const decimalSymbol = useLocaleDecimal ? getNumberDescriptor().symbols.decimal : '.';
 
 	let numDecimalDigits = 0;
 	for (let i = value.length - 1; i > -1; i--) {
 		const c = value.charAt(i);
-		if (c === descriptor.symbols.decimal) {
+		if (c === decimalSymbol) {
 			return numDecimalDigits;
 		} else {
 			numDecimalDigits++;
 		}
 	}
 
-	return undefined;
+	return 0;
 
 }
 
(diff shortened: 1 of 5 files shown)


components/inputs/input-number.js
```diff
@@ -99,42 +92,84 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
 
 		this._formattedValue = '';
 		this._inputId = getUniqueId();
-		this._numDecimalDigits = 0;
+		this._trailingZeroes = false;
+		this._valueTrailingZeroes = '';
+	}
+
+	get maxFractionDigits() {
+		return this._maxFractionDigits ?? Math.max(this.minFractionDigits, 3);
```
dlockhart, line 103: I updated these to always return a value, instead of previously where they'd default to `undefined`. I felt like this better reflected what was going on, since the Intl library would still be rounding to 3 decimals by default.

components/inputs/input-number.js
```diff
@@ -99,42 +92,84 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+	}
+	set maxFractionDigits(val) {
+		if (isNaN(val) || val < 0 || val > 20 || val < this.minFractionDigits) {
+			throw new RangeError('maxFractionDigits must be between 0 and 20 and >= minFractionDigits');
+		}
```
dlockhart, line 108: Also added some exceptions here for invalid min/max fraction digits. I don't believe this is going to break anything that wasn't already broken, but this allows us to avoid some `try`-`catch` scenarios elsewhere and also helps consumers realize that they're doing something weird. Previously it would have just not done anything.

components/inputs/input-number.js
```diff
@@ -99,42 +92,84 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+		this._maxFractionDigits = val;
+		this._updateFormattedValue();
```
dlockhart, line 110: Not calling `requestUpdate()` because changing min/max fraction digits on its own isn't the thing that should trigger a re-render. It's if the formatted value changes that should.

components/inputs/input-number.js
```diff
@@ -99,42 +92,84 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
-					this.minFractionDigits,
-					this.maxFractionDigits,
-					this.trailingZeroes,
-					this._numDecimalDigits
-				);
-				this._value = parseNumber(this._formattedValue);
-			}
-		} catch (err) {
-			this._formattedValue = '';
-			this._value = undefined;
```
dlockhart, line 128: This used to take a `Number`, format it to a string and then parse that. That seemed overly complicated so I simplified it, which has the side effect that `elem.value = '80';` no longer works -- which was a test case in the `input-percent` tests. I think that should be OK though... setting a `Number` property to a string and having it work is bad IMO.

components/inputs/test/input-number.test.js
```diff
@@ -33,7 +31,7 @@ async function setInnerInputValue(elem, value) {
 	setTimeout(() => dispatchEvent(inputTextElement, 'change'));
 }
 
-describe.only('d2l-input-number', () => {
+describe('d2l-input-number', () => {
```
dlockhart, line 34: Uh... I guess this sneaked in earlier! Oops.

components/inputs/test/input-number.test.js
```diff
@@ -58,90 +56,145 @@ describe.only('d2l-input-number', () => {
 			});
 		});
 
-		['autocomplete', 'max', 'maxFractionDigits', 'min', 'minFractionDigits', 'placeholder', 'title', 'value'].forEach((name) => {
+		['autocomplete', 'max', 'min', 'placeholder', 'title', 'value'].forEach((name) => {
 			it(`should default "${name}" property to 'undefined' when unset`, async() => {
-				expect(elem.value).to.equal(undefined);
+				expect(elem[name]).to.equal(undefined);
```
dlockhart, line 61: Haha so this test wasn't really testing what it thought it was before...

components/inputs/test/input-number.test.js
```diff
@@ -58,90 +56,145 @@ describe.only('d2l-input-number', () => {
...
-				expectInnerValue: '1.234'
-			}
-		].forEach((test) => {
-			it(test.name, async() => {
-				const elem = await fixture(test.fixture);
-				elem.value = test.value;
-				await elem.updateComplete;
-				expect(elem.value).to.equal(test.expectedValue);
-				expect(getInnerInputValue(elem)).to.equal(test.expectInnerValue);
-			});
```
dlockhart, line 116: The way these were working (by setting the value _after_ initialization) was what caused us to not notice the problems. So kept all the tests but just rewrote them a different way.

components/inputs/input-number.js
```diff
@@ -99,42 +92,84 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+			this.maxFractionDigits
+		);
+		let valueTrailingZeroes = this.value.toString();
+		const decimalDiff = (numDecimals - countDecimalDigits(valueTrailingZeroes, false));
+		if (decimalDiff > 0) {
+			if (decimalDiff === numDecimals) {
+				valueTrailingZeroes += '.';
+			}
+			valueTrailingZeroes = valueTrailingZeroes.padEnd(valueTrailingZeroes.length + decimalDiff, '0');
+		}
```
dlockhart, line 160: The logic for the `valueTrailingZeroes` is essentially:
- Count how many decimals were in the thing the user typed in, but then don't let it be more than `maxFractionDigits`
- Count how many decimals are in `this.value` (the `Number`), which is already rounded properly
- If there are more decimals in `valueTrailingZeroes`, the user must have typed in a bunch of zeroes, so add that many to the end (and a `.` decimal symbol if needed)

components/inputs/input-number.js
```diff
@@ -239,21 +264,49 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+
 		const value = e.target.value;
 		this._formattedValue = value;
 		await this.updateComplete;
-		const oldValue = this.value;
+
+		let dispatchEvent = false;
 		if (this.trailingZeroes) {
-			this._numDecimalDigits = countDecimalDigits(value);
+			const oldValueTrailingZeroes = this.valueTrailingZeroes;
```
dlockhart, line 274: This code runs when the user-entered values changes:
- Parse it into a `Number`
- Count how many decimals there were in the value they typed in (in their locale)
- Use the `Intl.NumberFormat` utility to convert that back into a string, but preserve the number of decimals
- Optionally trigger a `change` event

margaree (commented):
I think this approach seems good. One thing I noticed is that this line is missing `type` (though seems to work fine anyways) https://github.com/BrightspaceUI/core/blob/e3b4fee65c791cf1942f271f2cb975233cd055c1/components/inputs/input-number.js#L66 

Is there any concern around if my language is French and then I try to type a number with `.` as the decimal instead of `,` everything after the `.` disappears?

dlockhart:
I just added a bunch more tests of the trailing-zeroes stuff, which uncovered another issue with rounding precision and especially rounding negative values. Apparently `Math.round(-0.5)` rounds to `-0` and not `-1`, whee!

> One thing I noticed is that this line is missing `type`
Fixed, thanks!

> Is there any concern around if my language is French and then I try to type a number with . as the decimal instead of , everything after the . disappears?

I share your concerns about this, yeah. Looking at the MVC/legacy number inputs in the monolith, if you typed `1.5` while using the French locale, the validation would complain when you tried to submit the page (it doesn't do any live formatting on-change).

Not sure if the "Invalid number" validation is better or worse than trimming the value... `d2l-input-number` behaves such that the value inside is always a "valid" number or empty -- it might be over/under the max/min, but it's always valid in that sense.

One thing I could look at to make this a little nicer for users is in Intl's number parsing code, if the locale doesn't use `.` as the group separator and we don't see any `,` we could treat a period as the decimal symbol. I'm worried that may open up a whole new can of worms though.

margaree:
Hmm yeah the treating the `.` as the decimal does sound like it could end up complicated.

Imagining being a user in that scenario I feel like it could get really frustrating wondering why your post-decimal content is being erased with no indication as to why and so I'd lean towards the "Invalid number" validation though maybe Jeff should weigh in more on that.

dlockhart:
Ok I'll start a separate thread about this as it's not something introduced with this particular change.

margaree (approved):

components/inputs/input-number.js
```diff
@@ -7,46 +7,51 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
...
+function roundPrecisely(val, maxFractionDigits) {
+	const strValue = new Intl.NumberFormat(
+		'en-US',
+		{
+			maximumFractionDigits: maxFractionDigits,
+			minimumFractionDigits: 0,
+			useGrouping: false
+		}
+	).format(val);
+	return parseFloat(strValue);
```
dlockhart, line 52: I had to change the rounding algorithm as the other one was failing on large numbers with the huge amount of decimal precision that the monolith requires (16 digits). A number larger than `300,000` plus 16 digits of precision was overflowing the Number boundaries.

This is actually better/simpler anyway... just use the browser's formatter and have it do the rounding, then parse that!

PR #1265, opened 2021-04-09
fix: no longer showing group separators

dlockhart:
To improve the usability of our numeric input, we're going to no longer show thousand group separators inside the input box. These can be confusing to users when they go to edit the value, and simply aren't something that's commonly done elsewhere in the industry.


svanherk:
Codeowners file not adding the new group? 🤔

dlockhart:
Yeah I just made a fix to it... maybe it was case sensitive. Changed it to `@BrightspaceUI/gaudi-dev` from `@BrightspaceUI/Gaudi-Dev`.

karma.sauce.conf.js
```diff
@@ -7,17 +7,17 @@ const customLaunchers = {
...
 		base: 'SauceLabs',
 		browserName: 'firefox',
-		platform: 'OS X 10.15'
+		platform: 'macOS 11.00'
 	},
 	safari: {
 		base: 'SauceLabs',
 		browserName: 'safari',
-		platform: 'OS X 10.15'
+		platform: 'macOS 11.00'
```
dlockhart, line 20: Switched to Big Sur while I was in here to pick up Safari 14.

margaree (approved):

PR #1272, opened 2021-04-12
hierarchical view fixes for better custom view support

dbatiste:
This PR addresses two issues:

* ancestor views stealing focus away from nested custom views
* click and key events propagating outside the custom view

To prevent ancestor views from stealing focus, we temporarily remove `tabindex` from composed ancestors when drilled into custom views, and we restore them when drilling out.  Focusable items nested within other focusable items is kinda of a recipe for disaster.

For the click & key events, we simply call `stopPropagation` as they bubble to the custom view container.

I added some comments to the focus-capture code in effort to save some headaches of others that may read this code later, since it cost us both some time to figure it out.

### DIFF
--- a/components/hierarchical-view/hierarchical-view-mixin.js
+++ b/components/hierarchical-view/hierarchical-view-mixin.js
@@ -303,9 +303,9 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 
 		/* The purpose of this logic is to direct focus to the correct element, so that user
 		is never focused in a view that is not visible. It works based on the premise that all
-		elements are "display: none" except for the elements between the root view and the active
-		view, and the elements within the active view. Also worth noting, only the root view
-		captures focus with this handler. */
+		elements within the hierarchy are "display: none" except for the elements between the
+		root view and the active view, and the elements within the active view. Also worth noting,
+		only the root view captures focus with this handler. */
 
 		const parentView = this.__getParentViewFromEvent(e);
 
--- a/components/hierarchical-view/hierarchical-view-mixin.js
+++ b/components/hierarchical-view/hierarchical-view-mixin.js
@@ -300,8 +300,16 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 	}
 
 	__focusCapture(e) {
+
+		/* The purpose of this logic is to direct focus to the correct element, so that user
+		is never focused in a view that is not visible to the user. It works based on the premise
+		that all elements are "display: none" except for the elements between the root of the
+		hierarchy and the active view, and the elements within the active view. Also worth noting,
+		only the root-view captures focus with this handler. */
+
 		const parentView = this.__getParentViewFromEvent(e);
 
+		// the parent view of the element receiving focus is active, so nothing special to do
 		if (parentView.isActive()) return;
 
 		const relatedTarget = e.relatedTarget;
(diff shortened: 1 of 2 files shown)


dbatiste:
Note: I need to make a fix to only `stopPropagation` from child-views, so that the Esc key still closes the dropdown when pressed within the root view.

components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -472,4 +505,28 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
...
+					node.setAttribute(targetAttribute, tabIndex);
+					node.removeAttribute(sourceAttribute);
+				}
+			}
+			node = getComposedParent(node);
+		}
+	}
+
+	__updateFocusablesHide(view) {
+		this.__updateFocusables(view, 'd2l-hierarchical-view-tabindex', 'tabindex');
```
dlockhart, line 525: If these elements could include native elements, would it be better to throw this into a `data-*` attribute? Just to avoid having linters and HTML validators complaining at us for using unsupported attribute names on `<div>`s and such.

dbatiste: Thanks, yeah,  This was meant to use a data attribute.

dlockhart (approved):

components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -472,4 +505,28 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
...
+		}
+	}
+
+	__updateFocusablesHide(view) {
+		this.__updateFocusables(view, 'data-d2l-hierarchical-view-tabindex', 'tabindex');
+	}
+
+	__updateFocusablesShow(view) {
+		this.__updateFocusables(view, 'tabindex', 'data-d2l-hierarchical-view-tabindex');
+	}
```
svanherk, line 530: The names seem backwards to me here, but maybe I'm not understanding.  `__updateFocusablesHide` adds/shows the focus back, and `__updateFocusablesShow` removes/hides the focus again?

svanherk: Or maybe the names are strictly tied to updating the focus because of what is happening? I.e. when we are now showing a new view, we run `__updateFocusablesShow` (and hide the old ones), and when we hide the view we run `__updateFocusablesHide` (which shows the new ones).

dbatiste: >Or maybe the names are strictly tied to updating the focus because of what is happening?

Yeah, `__updateFocusablesHide` is for when a view is being hidden (drilling up), whereas `__updateFocusablesShow` is for when a view is being shown (drilling in).  I think different names would be better... to reflect what it is doing as opposed to what is happening.  Maybe, `enableFocusableAncestors`/`disableFocusableAncestors`.  I originally considered those, but thought enable/disabled wasn't quite right.  Still better in hindsight though.  Thoughts?

svanherk: Maybe `reenable/disable` or `readd/remove`?  Or a better word that doesn't include "re", but means "again"?

svanherk (approved):
Tried this out in the filter and it appears to be working well!  Adding the propagation fix stops the scroll jumping issue caused by the parent menu-item handling the click, and adding the tabindex fix stops the search-input focus craziness.

components/hierarchical-view/hierarchical-view-mixin.js
dbatiste: Ok, I updated to `__removeAncestorTabIndicies` and `__resetAncestorTabIndicies`... I _think_ that is clearer/less confusing.

svanherk: Yup I like it! 👍

PR #1276, opened 2021-04-13
feat: suppressing non-numeric input and showing hints

dlockhart:
This includes several usability improvements to our numeric input. It now suppresses any typed character that isn't one we're expecting (i.e. numbers or decimals or the negative symbol). Additionally, in certain scenarios when a suppression occurs we'll show the user a hint to help them understand why.

If the decimal symbol is entered in a field that's integer-only:
Screen Shot 2021-04-13 at 4 10 51 PM

Or if the user enters the wrong decimal symbol for their chosen locale:
Screen Shot 2021-04-13 at 4 07 46 PM

Or if the user enters a duplicate decimal:
Screen Shot 2021-04-13 at 4 07 55 PM

We also suppress the negative symbol if the cursor isn't at the start of the input (or the end in RTL languages), but we don't provide any hints here.

Paste is a special case we're also now handling, where we treat it similar to the `onchange` case and do a full parse & validate. This way if anything gets stripped out or reformatted the user sees it right away instead of waiting for the blur event. We decided not to handle any of the [other inputTypes](https://rawgit.com/w3c/input-events/v1/index.html#interface-InputEvent-Attributes) like drop/undo/redo as we didn't want to mess with the value before the user was finished interacting with it.


dlockhart:
Of course Legacy-Edge is the only thing that's failing the tests. Looking into it.

components/inputs/input-number.js
```diff
@@ -313,6 +350,56 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+	}
+
+	_handleKeyPress(e) {
+		const key = e.key;
+		let prevent = false;
+		let hintType = HINT_TYPES.NONE;
+		const hasDecimal = e.target.value.indexOf(this._descriptor.symbols.decimal) > -1;
+		if (key === this._descriptor.symbols.negative) {
+			// negative symbol must be at the end
+			if (this._descriptor.patterns.decimal.negativePattern === '{number}-') {
```
GeorgePlukov, line 369: Maybe I'm blanking, but what use case would the negative symbol at the end be for?

components/inputs/input-number.js
```diff
@@ -275,6 +288,30 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 		return [...childErrors, ...errors];
 	}
 
+	_getTooltip() {
+		if (this.disabled) return null;
+		if (this.validationError && this.childErrors.size === 0) {
+			return html`<d2l-tooltip announced for="${this._inputId}" state="error" align="start">${this.validationError}</d2l-tooltip>`;
+		}
+		let lang = '';
```
GeorgePlukov, line 296: nit: maybe this should just be named hint? Also might be a bit nicer to just default it to null and  just do `if (lang) {` on line 306

GeorgePlukov (approved):
Looks good to me!

components/inputs/input-number.js
dlockhart: In Arabic and other RTL locales, negative numbers would look like this: `92-` instead of `-92`.

components/inputs/input-number.js
```diff
@@ -199,21 +210,23 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
-		const tooltip = !this.disabled && this.validationError && this.childErrors.size === 0 ? html`<d2l-tooltip announced for="${this._inputId}" state="error" align="start">${this.validationError}</d2l-tooltip>` : null;
 		return html`
 			<d2l-input-text
 				autocomplete="${ifDefined(this.autocomplete)}"
 				?noValidate="${this.noValidate}"
 				?autofocus="${this.autofocus}"
 				@blur="${this._handleBlur}"
 				@change="${this._handleChange}"
+				@input="${this._handleInput}"
+				@keypress="${this._handleKeyPress}"
```
dbatiste, line 229: Was there a reason to use `keypress` instead of `keydown`/`keyup`, given that [`keypress` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event)?

components/inputs/input-number.js
```diff
@@ -313,6 +350,56 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
...
+			// looking for an integer only
+			} else if (this.maxFractionDigits === 0) {
+				hintType = HINT_TYPES.INTEGER;
+				prevent = true;
+			}
+		} else if ((key === '.' || key === ',') && !hasDecimal && this.maxFractionDigits > 0) {
+			// incorrect decimal
+			hintType = key === ',' ? HINT_TYPES.DECIMAL_INCORRECT_PERIOD : HINT_TYPES.DECIMAL_INCORRECT_COMMA;
+			prevent = true;
+		} else if (['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].indexOf(key) === -1) {
```
dbatiste, line 393: Is `e` a valid character?  The reason I ask is - the native `input type="number"` appears to suppress input, but it allows the `e` character.  I'm guessing we can't use the native type="number" due to our crazy localization rules, ya?  The native number input includes up/down arrows, which undoubtedly would throw a little wrench into our nice layout. Nevertheless, mobile devices may provide a nicer keyboard (showing numbers only) for native number inputs.  But, there may be some browser inconsistency with native number input also, as described [in this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1398528).

components/inputs/input-number.js
dlockhart: Mostly was following examples of suppressing characters, and `keydown` fires for everything (shift, tab, etc.) so separating out the things I cared about is more difficult with it. But yeah deprecated is no good so I'll try to fix this to work with `keydown`.

components/inputs/input-number.js
dlockhart: Yeah native number input isn't in the cards for us right now for the reasons you listed. It's totally inconsistently implemented across browsers and more importantly it doesn't support different decimal separators. Had we never supported those we might be able to get away with it, but changing that behaviour at this point would be very difficult.

dbatiste: Sounds good.  Do we need to allow the `e` character though?

components/inputs/input-number.js
dbatiste: Sounds good.  I think we do have some other stuff that relies on `keypress`.

components/inputs/input-number.js
dlockhart: We've never allowed scientific notation in our numeric inputs -- the "e" character results in an "Invalid Decimal" error with our old version -- so I don't think we need to worry about it at this point.

components/inputs/input-number.js
dlockhart: I looked into switching to `keydown` and... it's complicated. I got something kind-of working, but because it fires for ALL keys -- including Shift/Tab/Arrow/etc. -- it's much more difficult to only suppress "not numbers". I added a bunch of keys to let through, but then ran into cases where "Ctrl-Z" would be suppressed because the key is "z".

So after a conversation with Dave B we decided we'll stick with `keypress` for now, and swap it out for `beforeinput` once legacy-Edge goes away. `beforeinput` exposes `inputType` which lets us check for text insertion events more selectively.

dbatiste (approved):

PR #1280, opened 2021-04-17
List item slim style.

dbatiste:
This PR adds the slim style to the list-item as required.  It also fixes and existing layout issue for the case where items are selectable and separators are extended - previously the extra whitespace would have been incorrectly added between the checkbox and the content, rather than the outside margin of the checkbox. 

There are a lot of permutations of settings that are possible - I have included screenshots that focus on slim.  @svanherk : for the filter, I think you are most interested in the second screenshot which uses extended separators.  

Worth mentioning, `d2l-list-item-content` does not provide any API for controlling the wrapping behaviour of its slotted elements - it leaves this up to the consumer.  I considered adding a `no-wrap` property to it, however I decided against it because we'd actually need three different properties - one for each slot container.  Ultimately, consumers can put whatever they want in these slots, so I currently feel this is best left to them. In this particular case, the ellipsis is accomplished simply with:

```css
d2l-list-item[slim] div {
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

`slim`, `d2l-list-item-content`:
image

`slim`, `extend-separators`, `selectable`, `d2l-list-item-content` (primary only):
image

`slim`, `extend-separators`, `selectable`, `d2l-list-item-content` with `supporting-info` slot:
image


components/list/list-item-mixin.js
```diff
@@ -81,10 +82,20 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 				border-bottom: 1px solid transparent;
 				border-top: 1px solid transparent;
 			}
-			[slot="content"].d2l-list-item-content-extend-separators {
```
dbatiste, line 84: This selector changed because the `d2l-list-item-content-extend-separators` class was moved up to the parent, so that it could be used to select both the `content` and `control` slots when separators are being extended.

components/list/list-item-mixin.js
```diff
@@ -81,10 +82,20 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
...
+			.d2l-list-item-content-extend-separators > [slot="control"] {
+				width: 3rem;
+			}
+			.d2l-list-item-content-extend-separators > [slot="content"] {
 				padding-left: 0.9rem;
 				padding-right: 0.9rem;
 			}
+			:host([selectable]) .d2l-list-item-content-extend-separators > [slot="content"] {
```
dbatiste, line 92: This (along with the change further below to add he whitespace to the outside edge of the checkbox) fixes the issue where the `0.9rem` extra whitespace was incorrectly added in between the content and the checkbox when separators are extended.

components/list/list-item-mixin.js
```diff
@@ -32,6 +32,7 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 			 * Breakpoints for responsiveness in pixels. There are four different breakpoints and only the four largest breakpoints will be used.
 			 */
 			breakpoints: { type: Array },
```
dlockhart, line 34: Comment in here so it'll eventually show up in the auto-generate catalogue stuff?

dlockhart (approved):

svanherk:
Pulled the branch and added to filter - looks nice!
image

svanherk (approved):

m6jones:
Sweet. I need this for work to do I just released. Will update it so next release uses slim

PR #1282, opened 2021-04-20
Refactor html-block rendering for math and @mentions.

dbatiste:
Putting this PR to get feedback on this approach to making `d2l-html-block` rendering extensible. Currently, `d2l-html-block` contains rendering logic specific to MathJax, and we will need to extend it further to support `@mentions`, and likely also `prismjs` for code samples, et.al..

The general idea is to define a common interface for renderer implementations to replace parts of the user-authored HTML.  The `d2l-html-block` component may know of renderers built into core, but would be able to request additional domain-specific renderers using `requestInstance`.

The interface is simple - an asynchronous `prepare` method for loading dependencies or performing checks; and a synchronous `render` method for doing the actual replacement.

### DIFF
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -9,6 +9,60 @@
 		<script type="module">
 			import '../../demo/demo-page.js';
 			import '../html-block.js';
+			import { provideInstance } from '../../../mixins/provider-mixin.js';
+
+			let profileCardAccess;
+			const hasProfileCardAccess = () => {
+				if (profileCardAccess) return profileCardAccess;
+
+				profileCardAccess = new Promise(resolve => {
+
+					resolve(true); // temp
+
+					if (!D2L.LP) {
+						resolve(false);
+						return;
+					}
+
+					D2L.LP.Web.UI.Rpc.Connect(
+						D2L.LP.Web.UI.Rpc.Verbs.GET,
+						new D2L.LP.Web.Http.UrlLocation('/d2l/lp/profilecardsaccess/AccessCheck'),
+						null,
+						{ success: response => resolve(response) }
+					);
+				});
+
+				return profileCardAccess;
+			};
+
+			provideInstance(document, 'html-block-renderers', [{
+				prepare: async() => {
+					return hasProfileCardAccess();
+				},
+				render: elem => {
+
+					const mentions = elem.querySelectorAll('[data-mentions-id]');
+					if (mentions.length === 0) return elem;
+
+					mentions.forEach(mention => {
+
+						const userId = mention.getAttribute('data-mentions-id');
+						if (!userId) return;
+
+						const anchor = document.createElement('a');
+						anchor.href = `/d2l/lp/profilecardsaccess/UserProfile?userId=${parseInt(userId)}`;
+						anchor.target = '_blank';
+						anchor.innerText = mention.innerText;
+
+						mention.innerText = '';
+						mention.appendChild(anchor);
+
+					});
+
+					return elem;
+				}
+			}]);
+
 		</script>
 		<script>
 			if (window.location.search.indexOf('latex=true') !== -1) {
@@ -440,6 +494,7 @@
 							</math> and other symbols.
 						</div>`,
 						'The wizard quickly jinxed the gnomes before they vaporized.',
+						'The wizard (<span data-mentions-id="0">Elmer Fudd</span>) quickly jinxed the gnomes before they vaporized.',
 						'A quick movement of the enemy will jeopardize six gunboats.',
 						'Grumpy wizards make a toxic brew for the jovial queen.',
 						'Painful zombies quickly watch a jinxed graveyard.',
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -1,6 +1,16 @@
 import '../colors/colors.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
-import { loadMathJax } from '../../helpers/mathjax.js';
+import { htmlBlockMathRenderer } from '../../helpers/mathjax.js';
+import { requestInstance } from '../../mixins/provider-mixin.js';
+
+let renderers;
+
+const getRenderers = () => {
+	if (renderers) return renderers;
+	const tempRenderers = requestInstance(document, 'html-block-renderers');
+	renderers = (tempRenderers ? [ htmlBlockMathRenderer, ...tempRenderers ] : [ htmlBlockMathRenderer ]);
+	return renderers;
+};
 
 /**
  * A component for displaying user-authored HTML.
@@ -113,6 +123,11 @@ class HtmlBlock extends LitElement {
 		`;
 	}
 
+	constructor() {
+		super();
+		this._renderers = getRenderers();
+	}
+
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		if (this._templateObserver) this._templateObserver.disconnect();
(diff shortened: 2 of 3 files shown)


components/html-block/demo/html-block.html
```diff
@@ -9,6 +9,60 @@
 		<script type="module">
 			import '../../demo/demo-page.js';
 			import '../html-block.js';
+			import { provideInstance } from '../../../mixins/provider-mixin.js';
+
+			let profileCardAccess;
```
dbatiste, line 14: The code to support `@mentions` is basically copied from the LMS and cleaned up a bit.  I placed it here in this demo file for purposes of temporary demo only... the renderer for `@mentions` perhaps better belongs in [BrightspaceUILabs/user-profile-card](https://github.com/BrightspaceUILabs/user-profile-card), and BSI would pull in the domain specific renderers and provide them via `provideInstance`.

components/html-block/demo/html-block.html
```diff
@@ -9,6 +9,60 @@
...
+
+					const mentions = elem.querySelectorAll('[data-mentions-id]');
+					if (mentions.length === 0) return elem;
+
+					mentions.forEach(mention => {
+
+						const userId = mention.getAttribute('data-mentions-id');
+						if (!userId) return;
+
+						const anchor = document.createElement('a');
```
dbatiste, line 52: This is currently what is rendered in Spark, but in theory we can render a fancy user profile card.

components/html-block/html-block.js
```diff
@@ -1,6 +1,16 @@
 import '../colors/colors.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
-import { loadMathJax } from '../../helpers/mathjax.js';
+import { htmlBlockMathRenderer } from '../../helpers/mathjax.js';
+import { requestInstance } from '../../mixins/provider-mixin.js';
+
+let renderers;
+
+const getRenderers = () => {
```
dbatiste, line 8: Currently, this assumes a single set of renderers provided at the `document` simply as an optimization.  If required, we could request a different set of context-specific renderers for each instance of `d2l-html-block`.

components/html-block/html-block.js
```diff
@@ -129,22 +144,18 @@ class HtmlBlock extends LitElement {
...
-					const mathJaxConfig = { renderLatex: isLatexSupported };
-					await loadMathJax(mathJaxConfig);
-
-					this._renderContainer.innerHTML = `<mjx-doc><mjx-head></mjx-head><mjx-body>${fragmentHTML}</mjx-body></mjx-doc>`;
-					window.MathJax.typesetShadow(this.shadowRoot);
-				} else {
-					this._renderContainer.innerHTML = fragmentHTML;
-				}
+
+				// parallelize prepare (loading dependencies)
```
dbatiste, line 150: The current approach iterates over each of the renderers, loads the dependencies, and invokes their render methods if necessary. The one thing I don't like about this approach is that each renderer ends up searching the HTML.

An alternative approach that I have considered is to search once for fragments that should be replaced with something like `elem.querySelectorAll('[data-d2l-html-block-renderer="renderer key"]')`.  For each element returned, request the required renderer using `requestInstance`, and the proceed with loading dependencies and doing the replacements.  The problem with this approach is - it will not work for math, where we need to query for `math`, `\(`, `\[`, `$$` instead of a common attribute like `data-d2l-html-block-renderer`.

helpers/mathjax.js
```diff
@@ -1,5 +1,26 @@
 let mathJaxLoaded;
 
+export const htmlBlockMathRenderer = {
+	prepare: async elem => {
```
dbatiste, line 4: `prepare` performs necessary checks, loads dependencies, and returns `true`/`false` indicating whether replacements (in this case, math typesetting) are required.  

The reason why this is done separately from `render` is - so that all renderers can load their dependencies or perform their async checks in parallel.  In contrast, `render` must be done in series since they operate on the same DOM.

components/html-block/html-block.js
```diff
@@ -113,6 +123,11 @@ class HtmlBlock extends LitElement {
 		`;
 	}
 
+	constructor() {
+		super();
+		this._renderers = getRenderers();
```
dlockhart, line 128: Could maybe be a `static` property (or no property at all) to avoid every instance having a reference to the array.

components/html-block/html-block.js
dlockhart: Interesting. Like the ideal case is that if there's no fragment for a particular renderer, no work gets done and nothing gets loaded. The way you have it here, `prepare()` always gets called for all renderers, which is going to result in an API call for the mentions stuff even if there are no mentions.

I don't think I'd be too worried about the multiple `querySelectorAll` calls -- they're pretty quick. Could `prepare` and `render` be combined, making `render` async? That way the plugins could query the HTML and bail out if it doesn't contain anything of interest. If it does contain something, then they could do a 1-time "prepare" to load any dependencies and then finally do the actual replacement.

components/html-block/html-block.js
dbatiste: I can give that a try.  Originally I had this outside the class but the timing didn't work out, since the demo page currently imports this file before the provider registered the @mentions renderer.

components/html-block/html-block.js
dbatiste: > The way you have it here, `prepare()` always gets called for all renderers, which is going to result in an API call for the mentions stuff even if there are no mentions.

That's a good point.  The @mentions `prepare` could query for the attribute and return `false` before doing the API check.
 
> Could `prepare` and `render` be combined, making `render` async? That way the plugins could query the HTML and bail out if it doesn't contain anything of interest. If it does contain something, then they could do a 1-time "prepare" to load any dependencies and then finally do the actual replacement.

I was originally thinking along those lines too but then thought there would be issues due to (the math render in particular) returning a different outer element.  I think this may be possible as long as the outer element (`div` passed into each `render` call) is exactly the same, which is the same element ultimately referenced on line 158. I don't think it should matter, but the order they run could become inconsistent.

dlockhart: I guess I'm just questioning whether having a separate `prepare()` makes sense given that it may need to look at the HTML anyway (so it would need to be passed the node) to see if it even needs to do anything. So if you're passing the node to both `prepare()` and `render()`, why not just combine them and for plugins that need to do some kind of preparation (import stuff, do API calls) they can be smart enough to just do that work the first time.

grant-cleary: I wonder about the cost of forcing the prep work to be synchronous (if it's folded into `render()`). I know those query selectors are fast, but I'm less sure about the regex the math stuff needs to test, especially on what could potentially be very large HTML files (potentially MB in size). I guess it doesn't matter, since the rendering can't happen until the prep is done and the user's browser is going to incur that time/processing cost regardless. In those cases, I imagine perhaps the math `prepare()` could end up dwarfing the other providers and it wouldn't cost much extra time to make it synchronous. :shrug:

dbatiste: Yeah I agree that doing it on one if possible would be ideal.  Keeping them async requires that the top-level element reference doesn't change though (which the math renderer currently does replace).  I am going to try and rejig thing slightly to avoid that.

dbatiste: Ok, I've thought about this some more and I think my original concern is legit.  Take these two renderers:

```javascript
// render 1
async render1(elem) {
    const stuffToReplace = elem.querySelector('[some-attr]');
    if (!stuffToReplace) return;

    await loadStuffDependencies();

    stuffToReplace.innerHTML = '...' / stuffToReplace.replaceChild(...);
}

//render 2 (akin to what mathjax does)
async render2(elem) {
    const fooToReplace = elem.querySelector('some-elem');
    if (!fooToReplace) return;

    await loadFooDependencies();

    elem.innerHTML = fooProcessor(elem.innerHTML);
}

await Promise.all(
    render1(elem),
    render2(elem)
);
```
This issue is that the replacement may be done in any order, and one replacement may invalidate node references within another render's closure.  In the above example, if `render2` dependencies are loaded first, its replacement would run first and completely invalidate `stuffToReplace`.  To some extent, this can be mitigated by not using `innerHTML`, but I think there is still opportunity of one replacement to invalidate node references in another render call.

From the beginning, I was trying to be optimistic about loading the dependencies (in parallel), perhaps to the peril of the code at hand, when ultimately, the only scenario when it actually matters is when a user would be seeing rich text containing these things all together, and for the very first time (uncached).  So perhaps we just make these calls synchronous, which avoid this issue altogether, and we can do it all in one method safely.

dlockhart: Right, I see the concern now. I'm less worried about the dependency loading step (which is cacheable like you said) and more concerned about that API call to check if the user has permission. That's likely _not_ being cached, so would literally run once per page load... and it's likely not fast either, probably 100-200ms.

So bailing early (doing the querySelector and if there's nothing, bailing) will be key here.

components/html-block/demo/html-block.html
```diff
@@ -9,6 +9,61 @@
...
+				return profileCardAccess;
+			};
+
+			provideInstance(document, 'html-block-renderers', [
+				async elem => {
+
+					const mentions = elem.querySelectorAll('[data-mentions-id]');
+					if (mentions.length === 0) return elem;
+
+					const hasAccess = await hasProfileCardAccess();
```
dbatiste, line 44: Currently, this call would block rendering of the rest of the user-authored HTML.  Perhaps what would actually be preferred is if this would render a web component that can render itself asynchronously and conditionally based on the access API result, so this code does not have to `await` the API.  In this case this could happen at most once per page.

components/html-block/html-block.js
dbatiste: Ok, I've updated to make this process rely on a single method per "plugin".

Re the API call that @mentions uses - maybe we can optimize beyond just checking for @mentions as per my [comment below](https://github.com/BrightspaceUI/core/pull/1282#discussion_r618832761).  I have added the short-circuit however.

PR #1283, opened 2021-04-20
Update Visual Diff Goldens

svanherk:
New goldens needed after the puppeteer bump (reports [here](https://github.com/BrightspaceUI/core/pull/1284)).


GeorgePlukov (approved):
Visual diffs all looked good to me!

PR #1285, opened 2021-04-20
Migrate d2l-table to Lit

dlockhart:
Ignore this for now! This will eventually be the branch/PR to migrate `d2l-table` and `d2l-scroll-wrapper` to Lit, but for now it just imports the old table stuff in an attempt to create a visual diff baseline to ensure I don't break things (too badly).


components/table/test/table.visual-diff.html
```diff
@@ -0,0 +1,28 @@
...
+			import '../../typography/typography.js';
+			import './table-test-visual-diff.js';
+			import './table-test-sticky-visual-diff.js';
+		</script>
+		<title>d2l-table</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+		<style>
+			html {
+				--d2l-input-position: static; /* emulates what grades is doing in the LMS */
```
svanherk, line 19: Should we be testing without this as well to verify `relative` inputs work as well?  Or is it expected that they won't and you need to set this for inputs in a table?

svanherk (approved):

components/table/test/table.visual-diff.html
dlockhart: Relatively positioned things will float on top of the sticky stuff. The only way around _that_ is to use z-indexes on the sticky things, but then they themselves can't use z-indexes which they're already relying on for the sticky columns to go overtop of the rest. So yeah... no relative things.

dlockhart: It shouldn't, unless the input needed relative positioning for things like the search button or date icon -- which means we can never use those with sticky headers. Mostly I added this so that we wouldn't accidentally remove this CSS variable later because we forgot why it was there.

PR #1287, opened 2021-04-21
build: switching to use ESM for visual-diff

dlockhart:
Converts core to use ES modules for our visual diff tests. Mostly straightforward... the helpers obviously needed to be converted over to modules.

The one remaining ugly part is for tests that use `forceFocusVisible`, they still need to import that into the HTML file and expose it as a global. Since it's a function, it can't be imported into the JS file and serialized and passed into an `eval` call. Maybe we could port it over into the visual diff helpers such that it could take an `ElementHandle` or a CSS query, but that'll be a separate change regardless.

### DIFF
--- a/components/alert/test/alert-toast-helper.js
+++ b/components/alert/test/alert-toast-helper.js
@@ -1,26 +1,21 @@
-
-module.exports = {
-
-	async getRect(page, selector) {
-		const rect = await page.$eval(selector, (toast) => {
-			const elem = toast.shadowRoot.querySelector('.d2l-alert-toast-container');
-			return {
-				x: elem.offe,
-				y: elem.offsetTop,
-				width: elem.offsetwidth,
-				height: elem.offsetHeight
-			};
-		});
+export async function getRect(page, selector) {
+	const rect = await page.$eval(selector, (toast) => {
+		const elem = toast.shadowRoot.querySelector('.d2l-alert-toast-container');
 		return {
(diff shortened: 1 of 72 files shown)


components/alert/test/alert-toast-helper.js
```diff
@@ -1,26 +1,21 @@
...
-		});
+export async function getRect(page, selector) {
+	const rect = await page.$eval(selector, (toast) => {
+		const elem = toast.shadowRoot.querySelector('.d2l-alert-toast-container');
 		return {
-			x: 0,
-			y: rect.y - 10,
-			width: page.viewport().width,
-			height: page.viewport().height - rect.y + 10
+			x: elem.offe,
```
dbatiste, line 5: This `offe` property looks a bit weird.  Is/was it `undefined`, and we were returning `0` before?  The `offsetwidth` one looks a bit strange too, since the native API is `offsetWidth`.

dlockhart: I agree this looks wrong, but this didn't actually change, I just removed the indent (the diff looks weird but if you look at side-by-side it'll be clearer).

dlockhart: The code appears to have been always like this since Allan introduced it. The good news: it's not actually used, so I've updated it a bit to only expose the used stuff.

dbatiste (approved):

PR #1289, opened 2021-04-23
Make d2l-html-block rendering extensible.

dbatiste:
This PR makes `d2l-html-block` rendering extensible by requesting (via `requestInstance`) renderers for replacing parts of the user-authored HTML.  It further refactors the MathJax rendering from this component to its respective helper.

To extend the rendering, implementors simply provide (wired up in BSI via `provideInstance`) a rendering method that takes the user-authored HTML, loads it's dependencies, does the replacement, and returns a promise for element.

```javascript
async elem => {

    // query for stuff to replace
    const elemsToReplace = elem.querySelectorAll(...);

    // bail early if nothing to do
    if (elemsToReplace.length === 0) return elem;

    // load any dependencies required
    ...

    // do the replacements
    ...

    return elem;
}
```

As a best practice, implementations should not block on non-cached checks or dependencies.

Corresponding user-profile / @mentions PR: https://github.com/BrightspaceUILabs/user-profile-card/pull/44


helpers/mathjax.js
```diff
@@ -1,5 +1,21 @@
 let mathJaxLoaded;
```
dlockhart, line 2: Could you see the MathJax renderer moving into the LMS like the others? Might have to do something for iFRAs, but I think they'll need to be handled either way for the other renderers.

dbatiste: Hmmm, I think it could but yeah, I think it would introduce some difficulties with iFRAs, and also math wouldn't work in app component demos.  I think we would also like to use MathJax to render math in the HtmlEditor's design-view if possible, but it remains to be seen if that's doable, and also what we'd need specifically.  Ultimately we would like a single loader and configuration for MathJax, since unfortunately we've seen some apps go on their own with it, which is now causing us some issues.  I could see perhaps moving the MathJax loader and renderer into its own repo if we didn't want it part of core.  In the spirit of that, we will likely have a prismjs loader/renderer as well.

dlockhart: Gotcha cool that makes sense. Just trying to figure out where we'd draw the line between putting something in core vs. an LMS plugin.

dlockhart (approved):
Looks good! Seems like something that could be somewhat easy to add unit tests for?

grant-cleary (approved):

dlockhart (approved):

PR #1299, opened 2021-04-26
d2l-scroll-wrapper port to Lit

dlockhart:
This ports `<d2l-scroll-wrapper>` to Lit. Since it previously lived in the `d2l-table` repo, this is first going into my "table to Lit" branch before all of it can merge together.

Old source here:
https://github.com/BrightspaceUI/table/blob/master/d2l-scroll-wrapper.js

This is dramatically simpler -- 215 lines instead of 573. Some notes:

- No need for the "sticky polyfill" since all the browsers we care about support it natively 🎉 
- I've removed the `<d2l-sticky-element>` and `<d2l-table-circle-button>` elements and just replaced them with native elements. They were never used outside the project, [except for some UI automation](https://search.d2l.dev/xref/lms/uiautomation-tests/D2L.Automation.UI.PageObjects/Dropbox/DropboxFoldersListPage.cs?r=58509bd2) that I'll have to fix once this all merges.
- The `is-sticky` attribute was set by table when sticky headers were used in order to disable all the functionality. I'm going to just conditionally use `<d2l-scroll-wrapper>` from table once I port it instead, so I've removed it from here.

More notes inline...


components/demo/demo-snippet.js
```diff
@@ -34,10 +34,13 @@ class DemoSnippet extends LitElement {
...
 				flex: 1 1 auto;
-				padding: 18px;
+				overflow: hidden;
 				position: relative;
 			}
-			:host([no-padding]) .d2l-demo-snippet-demo {
+			.d2l-demo-snippet-demo-padding {
+				padding: 18px;
+			}
+			:host([no-padding]) .d2l-demo-snippet-demo-padding {
```
dlockhart, line 43: Some tweaks here to the demo snippet so that things that would overflow get capped.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
+import '../colors/colors.js';
+import '../icons/icon.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { FocusVisiblePolyfillMixin } from '@brightspace-ui/core/mixins/focus-visible-polyfill-mixin.js';
+import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+const RTL_MULTIPLIER = navigator.userAgent.indexOf('Edge/') > 0 ? 1 : -1; /* legacy-Edge doesn't reverse scrolling in RTL */
```
dlockhart, line 8: Previously this logic was... quite complex. Legacy-Edge now appears to be the only browser that has weird RTL scrolling behaviour, so I've limited the hackiness to just check for it.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
...
+		this._checkScrollThresholds();
+	}
+
+	scrollDistance(distance, smooth) {
+		if (this._container) {
+			if (this._dir === 'rtl') distance = distance * RTL_MULTIPLIER;
+			if (this._container.scrollBy) {
+				this._container.scrollBy({ left: distance, behavior: smooth ? 'smooth' : 'auto' });
+			} else {
+				// legacy-Edge doesn't support scrollBy
```
dlockhart, line 192: More fun hacks for legacy-Edge. Hopefully not long lived!

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
...
+				</button>
+				<button class="d2l-scroll-wrapper-button d2l-scroll-wrapper-button-right" @click="${this._scrollRight}" aria-hidden="true">
+					<d2l-icon icon="tier1:chevron-right"></d2l-icon>
+				</button>
+			</div>` : null;
+		return html`
+			${actions}
+			<div class="d2l-scroll-wrapper-container" @scroll="${this._checkScrollThresholds}"><slot></slot></div>
+		`;
+	}
```
dlockhart, line 177: I've removed the `scroll()` method since a) it's a native method on all DOM nodes so overriding it is weird and b) nothing was using it.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
...
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this._container = this.shadowRoot.querySelector('.d2l-scroll-wrapper-container');
+		new ResizeObserver(
+			() => requestAnimationFrame(
+				() => this.checkScrollbar()
+			)
+		).observe(this._container);
+		requestAnimationFrame(() => this.checkScrollbar());
```
dbatiste, line 160: Is this needed? I did a quick check in Chrome and Safari and the observer seems to be called initially in both cases, but maybe I did not hit the scenario where it's needed.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
...
+
+	checkScrollbar() {
+		if (!this._container) return;
+		const hScrollbar = Math.abs(this._container.offsetWidth - this._container.scrollWidth);
+		this._hScrollbar = (hScrollbar > 0);
+		this._checkScrollThresholds();
+	}
+
+	scrollDistance(distance, smooth) {
+		if (this._container) {
```
dbatiste, line 187: Nit... looks like we could bail earlier here similar to the other methods.

components/scroll-wrapper/demo/scroll-wrapper.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
```
dbatiste, line 1: By the way... I took it for a spin in the primary-secondary template, and it seems to work nicely there.  😄 

image

components/demo/demo-snippet.js
dlockhart: These are breaking the floating buttons visual diffs (preventing them from working in the demo snippet). I'll look into this more.

components/scroll-wrapper/scroll-wrapper.js
dlockhart: Yeah I'm seeing the same. I'll try it in Firefox & legacy-Edge and if it's always called initially I'll remove it. 👍

dbatiste (approved):
Not sure if that one call is needed in first updated, but otherwise all looks good to me.

components/scroll-wrapper/README.md
```diff
@@ -0,0 +1,22 @@
...
+<d2l-scroll-wrapper show-actions>
+	<!-- content which may overflow horizontall -->
+</d2l-scroll-wrapper>
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `show-actions` | Boolean, default: `false` | Whether to show left/right scroll buttons |
```
svanherk, line 20: It's probably not worth updating everywhere using table to switch it, but it's always felt weird to me that this functionality defaults to `false`, instead of using a `hide-actions` property.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	checkScrollbar() {
+		if (!this._container) return;
+		const hScrollbar = Math.abs(this._container.offsetWidth - this._container.scrollWidth);
+		this._hScrollbar = (hScrollbar > 0);
+		this._checkScrollThresholds();
+	}
+
+	scrollDistance(distance, smooth) {
```
svanherk, line 181: Is the ability to pass in the `smooth` parameter just to make the animation faster for tests?  Will the browser automatically know to "ignore" `smooth` in `scrollBy` if reduced motion is on?

components/scroll-wrapper/scroll-wrapper.js
dlockhart: Confirmed this wasn't needed -- removed.

components/scroll-wrapper/README.md
dlockhart: I agree, and I've toiled with flipping this around as well. It wouldn't affect places using `d2l-table`, just those using `d2l-scroll-wrapper` directly... which according to code search is:
- 5 places that are setting `show-actions`
- 3 places that are not (2 of which are in the monolith and have been broken since last summer)

I think I will do this... it can be pretty safe as I can go into those 3 places right now and add `hide-actions` and then things will "just work" when the switchover happens.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -0,0 +1,214 @@
...
+		if (!this._container) return;
+		if (this._dir === 'rtl') distance = distance * RTL_MULTIPLIER;
+		if (this._container.scrollBy) {
+			this._container.scrollBy({ left: distance, behavior: smooth ? 'smooth' : 'auto' });
+		} else {
+			// legacy-Edge doesn't support scrollBy
+			this._container.scrollLeft = distance;
```
svanherk, line 188: Been a while since I was in scrolling land and can't easily checkout the branch at the moment - does this scroll smooth, or do we need `scroll-behavior: smooth;` in the css?

components/scroll-wrapper/scroll-wrapper.js
dlockhart: That's my understanding, yes. Safari doesn't support `smooth` at all, and obviously legacy-Edge isn't using it. I tested this out in Chrome and Firefox and neither seem to do anything differently when `prefers-reduced-motion` is chosen. 

I'm not super worried about this, since it's not really an animation so much as it's just normal scrolling, which also doesn't seem to behave differently with that setting. 🤷

components/scroll-wrapper/scroll-wrapper.js
dlockhart: It does not, and I didn't care to spend more time / add more hacks for legacy-Edge since it's going away possibly before this even gets released.

svanherk: Fair enough!

components/scroll-wrapper/README.md
dlockhart: Alright, I've flipped this around to `hide-actions` and opened a [PR into the LMS](https://github.com/Brightspace/lms/pull/9856) and [manager view](https://github.com/Brightspace/manager-view-fra/pull/2123) to add that attribute into the 2 places that wanted the actions hidden.

svanherk (approved):

PR #1301, opened 2021-04-28
US122200 - First pieces of the filter component with minimal functionality

svanherk:
In this PR:
- Basic start to a README
- The main `d2l-filter`, as well as the first dimension type (`d2l-filter-dimension-set` and its `d2l-filter-dimension-set-value`)
- Simple demo
- Unit and axe tests for current functionality

The point of this PR is to get the basic pieces setup, and then more functionality will be iteratively added on top (keeping track of counts, searching, special single dimension layout, responsive layout, etc)


components/filter/filter-dimension-value.js
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		const changes = new Map();
+		changedProperties.forEach((oldValue, prop) => {
+			if (oldValue === undefined) return;
+
+			if (prop === 'selected' || prop === 'text') {
```
svanherk, line 31: Key is not changeable after the fact, since that would defeat the purpose of a unique identifier.  The consumer cannot render these components until they have enough info to create a key.

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+						${dimensions}
+					</d2l-menu>
+				</d2l-dropdown-menu>
+			</d2l-dropdown-button-subtle>
+		`;
+	}
+
+	_buildDimension(dimension) {
+		let dimensionHTML;
+		switch (dimension.tagName.toLowerCase()) {
```
svanherk, line 125: Setting things up here to eventually have more than one type of dimension.  A lot of function names have been chosen to group things nicely once that happens as well.

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+
+						<div class="d2l-filter-dimension-value-text">${item.text}</div>
+					</d2l-list-item>
+				`)}
+			</d2l-list>
+		`;
+	}
+
+	_dispatchChangeEvent(eventDetail) {
+		this.dispatchEvent(new CustomEvent('d2l-filter-change', { bubbles: false, composed: false, detail: eventDetail }));
```
svanherk, line 173: This could eventually be more complicated, adding a slight delay to batch quick changes together.  Future story.

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+
+		return html`
+			<d2l-dropdown-button-subtle
+				@d2l-dropdown-close="${this._handleDropdownClose}"
+				@d2l-dropdown-open="${this._stopPropagation}"
+				@d2l-dropdown-position="${this._stopPropagation}"
+				text="${this.localize('components.filter.filters')}"
+				?disabled="${this.disabled}">
+				<d2l-dropdown-menu min-width="285" max-width="420" no-padding-header>
+					<div slot="header">${header}</div>
```
dlockhart, line 108: Worth not rendering the `<div>` either if there's no header? i.e. moving `<div slot="header"></div>` up into where `header` gets set.

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+		this._dispatchChangeEvent({ dimension: dimensionKey, value: { key: valueKey, selected: selected } });
+	}
+
+	_handleDataChangeListDimension(e) {
+		const dimension = this.shadowRoot.getElementById(e.target.key);
+		const changes = e.detail.changes;
+		if (changes.has('text')) {
+			const menuItem = dimension.parentNode;
+			if (menuItem.text !== changes.get('text')) {
+				this.requestUpdate();
```
svanherk, line 197: If a dimension's text changes, there's a number of places we'd need to update, so we just call a re-render here.  But we do try to determine if we need to first.  This logic is fairly specific, and will be specific for each data value and dimension type (you can see more examples in `_handleDataChangeListDimensionValue` directly below).  I was struggling to determine how far to go with this.  For this review I went all the way (determine if we need to change a value, and then try to change it without a re-render if there are no bad consequences).  We could also:
1) Try to determine if we need it, and then just always re-render
2) Try to determine if we need it if it's easy/reusable, and then re-render
3) Just always re-render

As we move down the list the logic will get much simpler, at the cost of re-rendering.  Thoughts?

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+
+	_handleDataChangeListDimensionValue(e) {
+		const dimension = this.shadowRoot.getElementById(e.target.parentNode.key);
+		const value = dimension.querySelector(`[key="${e.target.key}"]`);
+		const changes = e.detail.changes;
+
+		if (changes.has('text')) {
+			const text = value.querySelector('.d2l-filter-dimension-value-text');
+			if (text.innerText !== changes.get('text')) {
+				text.innerText = changes.get('text');
```
dlockhart, line 211: This ends up modifying the DOM outside of `render()`, which we try to avoid. Could we not just `requestUpdate` here which will cause a re-render and the `buildDimension` methods will re-create everything? Lit should then be smart enough to know if there were actually any changes and only make those that were necessary.

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+		this.shadowRoot.querySelector(`#${this._activeDimensionKey}`).hide();
+		this._activeDimensionKey = null;
+	}
+
+	_handleDimensionShowComplete() {
+		const dimension = this.shadowRoot.querySelector(`#${this._activeDimensionKey}`);
+		const focusable = getFirstFocusableDescendant(dimension);
+		if (focusable) {
+			focusable.focus();
+		}
```
svanherk, line 231: This function is currently setup to be used for every dimension type so it is very generic in finding the focus element.  But we could create a custom one for each type if this feels "bad"

components/filter/filter.js
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	_handleSlotChange() {
+		const nodes = this._dimensionsSlot.assignedNodes();
+		const filteredNodes = nodes.filter((node) => {
+			const isNode = node.nodeType === Node.ELEMENT_NODE && node.tagName.toLowerCase() !== 'template';
+			return isNode;
+		});
+
+		this._dimensions = new Map(filteredNodes.map(node => [node.key, node]));
```
svanherk, line 250: I convert this to a map so that when I do a number of lookups above it's cleaner and faster, but not sure if it's worth it overall

lang/en.js
```diff
@@ -46,8 +47,8 @@ export default {
 	"components.input-time-range.startTime": "Start Time",
 	"components.list-item-drag-handle.default": "Reorder item action for {name}",
 	"components.list-item-drag-handle.keyboard": "Reorder item, current position {currentPosition} out of {size}. To move this item, press up or down arrows.",
-	"components.menu-item-return.return": "Returns to previous menu.",
-	"components.menu-item-return.returnCurrentlyShowing": "Returns to previous menu. You are viewing {menuTitle}.",
+	"components.menu-item-return.return": "Return to previous menu.",
+	"components.menu-item-return.returnCurrentlyShowing": "Return to previous menu. You are viewing {menuTitle}.",
```
svanherk, line 51: Jeff requested this update, which I think sounds better.  We were both on the fence about whether filter should reuse the `components.menu-item-return.returnCurrentlyShowing` langtterm or make its own.  Pros and cons to both, so if someone feels strongly either way let me know.

lang/en.js
```diff
@@ -7,6 +7,7 @@ export default {
 	"components.calendar.selected": "Selected.",
 	"components.calendar.show": "Show {month}",
 	"components.dialog.close": "Close this dialog",
+	"components.filter.filters": "Filters",
```
svanherk, line 10: The more complicated langterm discussed will be coming once I add the ability to count and show num selected (future story)

components/filter/README.md
```diff
@@ -0,0 +1,86 @@
...
+| Property | Type | Description |
+|--|--|--|
+| `disabled` | Boolean, default: `false` | Disables the dropdown opener for the filter |
+
+**Events:**
+* `d2l-filter-change`: dispatched when any filter value has changed (may contain info about multiple changes)
+
+## Filter Dimension Types
+
+### d2l-filter-dimension and d2l-filter-dimension-value
```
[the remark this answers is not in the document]
dlockhart: We could try to use a word that's less tied to HTML semantics like "set"?

dlockhart (commented):
Mostly interested in ways we can keep the render loop clean -- i.e. not reading or writing DOM nodes outside of `render()`.

One way (which I think you explored?) would be to keep actual state in `_dimensions` instead of DOM nodes. So instead of looking at those DOM nodes and their tag names and internals, you'd look at JSON -- an array of dimensions, each with a "type" and then based on the type they may have children with values, etc.

That method is going to involve a lot more state management code where the sub-components fire events when they boot up and then again when their stuff changes, but it may let you just listen for those events, update your state and then re-render. Then you can rely on Lit's smart internals to only update things that need updating.

components/filter/README.md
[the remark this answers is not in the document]
svanherk: Hmm...

- `d2l-filter-dimension-set` and `d2l-filter-dimension-set-value`
- `d2l-filter-dimension-group` and `d2l-filter-dimension-group-value`
- `d2l-filter-dimension-collection` and `d2l-filter-dimension-collection-value`

`set` is definitely nice and short.

components/filter/filter.js
svanherk: Yeah that's what this wordy comment was trying to ask: https://github.com/BrightspaceUI/core/pull/1301#discussion_r622330195
Sounds like you're leaning towards 1 - avoid modifying the DOM outside of render and let Lit handle things in a smart well, but still check if we can avoid it altogether?

svanherk:
> One way (which I think you explored?) would be to keep actual state in _dimensions instead of DOM nodes. So instead of looking at those DOM nodes and their tag names and internals, you'd look at JSON -- an array of dimensions, each with a "type" and then based on the type they may have children with values, etc.

Yeah I investigated storing just the state versus storing the nodes early on before all the syncing code was in place, and storing the state was overkill.  But looking at where things ended up, I do think that will end up cleaner overall - I'll try switching back to that and see how it feels

components/filter/filter.js
dlockhart: I think so yeah. As you get the events, you should still be able to compare the new value against what's already in state and then only call `requestUpdate` if they're different. Checking an internal JSON object should be a lot cleaner/quicker than searching through DOM!

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+				text="${this.localize('components.filter.filters')}"
+				?disabled="${this.disabled}">
+				<d2l-dropdown-menu min-width="285" max-width="420" no-padding-header>
+					${header}
+					<d2l-menu label="${this.localize('components.filter.filters')}">
+						<slot
+							@d2l-filter-dimension-data-change="${this._handleDimensionDataChange}"
+							@d2l-filter-dimension-slot-change="${this._handleSlotChange}"
+							@slotchange="${this._handleSlotChange}"
+						></slot>
```
dlockhart, line 100: I guess this slot doesn't need to go inside the menu? Just wondering if it could live... somewhere else so that it isn't confusing menu with children that it doesn't understand.

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+	_handleDimensionShowStart(e) {
+		this._activeDimensionKey = e.detail.sourceView.id;
+	}
+
+	_handleDropdownClose(e) {
+		this._activeDimensionKey = null;
+		this._stopPropagation(e);
+	}
+
+	_handleSlotChange() {
```
dlockhart, line 233: Do we still need to do any of this DOM crawling when slots change? Was hoping we could rely solely on the events fired from the dimensions themselves to maintain our state. That might mean firing an even when they're disconnected to remove them.

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+	_handleDimensionDataChange(e) {
+		const dimension = this._dimensions.find(dimension => dimension.key === e.detail.dimensionKey);
+		const value = e.detail.valueKey && dimension.values.find(value => value.key === e.detail.valueKey);
+		const toUpdate = value ? value : dimension;
+		const changes = e.detail.changes;
+
+		changes.forEach((newValue, prop) => {
+			if (toUpdate[prop] !== newValue) {
+				toUpdate[prop] = newValue;
+				this.requestUpdate();
```
dlockhart, line 206: May want to only call `requestUpdate` once and not inside a loop. I'm sure it's smart enough to dedupe/wait, but just in case.

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+		const dimensionKey = e.composedPath()[0].parentNode.id;
+		const valueKey = e.detail.key;
+		const selected = e.detail.selected;
+
+		// Update the corresponding d2l-filter-dimension-set-value to keep them in sync
+		const dimension = this.querySelector(`d2l-filter-dimension-set[key="${dimensionKey}"`);
+		const slot = dimension.shadowRoot.querySelector('slot');
+		const items = this._getSlottedNodes(slot);
+		const item = items.find(item => item.key === valueKey);
+		item.selected = selected;
```
dlockhart, line 192: This still doesn't quite feel right. Like I get why it's necessary... querying into the DOM and setting properties on the dimension values to reflect the menu's state so that if the consumer reads those properties they're accurate... it just doesn't feel right.

I'm wondering how necessary it is? Like maybe we could say they're only used to build up the initial state, and then after that consumers need to use the `@change` event on the filter to react to changes? That would be similar to how `<input type="checkbox" checked>` is only correct initially and after that you use the value. I dunno, just seems like a lot of work.

components/filter/filter.js
svanherk: Hmm, good point - I think it was more important in previous iterations trying to keep event flow more natural, but I think I can move it to the bottom

dlockhart: Cool, yeah not a big deal but it mostly made me do a double-take... like "Wait, why is there a slot in the menu, aren't we rendering the menu items... oh I see"

components/filter/filter.js
svanherk: So _maybe_ we can get away with that... basically, the user would select an option, the filter fires an event, and the consumer needs to then update the data it uses to render the `?select=<whatever>` portion if it's going to need to care about that value after the initial render.  The main places I see this being important are the HM filter framework (where the filter accepts an href that understands the filters applied and updates the UI component accordingly, every time a filter is applied) and with the filter tags (where they can remove a filter tag, and the corresponding value is unselected in the filter component.

But I _think_ they would naturally keep their own data up-to-date anyways, so maybe I am pre-fixing a problem we may not see.  I could remove it for now and see if it comes up replacing the HM filter or writing the filter tags

components/filter/filter.js
svanherk: Yeah I figured it was smart enough, but it was an assumption - I can loop through and then `requestUpdate` at the end if there are any updates

components/filter/filter.js
dlockhart: I think they're going to need to keep internal state anyway, since the next time their component re-renders they'll also re-render the filter and blow away anything that changed anyway right? So us updating that state ourselves isn't really helping them if they're just going to overwrite it the next time their component renders.

dlockhart: Hmm... unless Lit gets confused and thinks things have or haven't changed when they re-render -- which was maybe the original motivation for doing this.

components/filter/filter.js
svanherk: Hmm, not sure I'm following - are you thinking instead of a `slotchange` event which tells us nothing about _what_ nodes changed, we'd just listen for `d2l-filter-dimension-added` and `d2l-filter-dimension-removed` events (and likewise, `d2l-filter-dimension-set` would need to listen for `d2l-filter-dimension-set-value-added` and `d2l-filter-dimension-set-value-removed`, etc)?

dlockhart: Yeah exactly, that way you could avoid reaching into the DOM entirely. You may even be able to avoid the initial `*-added` events and just assume if you see a `change` with a key you don't have yet then it's new. But you would need `*-removed` events.

components/filter/filter.js
svanherk: Yeah, my worry was that they would try to set something to the same value and Lit wouldn't recognize anything as different and `updated` would never get called.  _But in theory_, if they kept things up-to-date every time, they wouldn't ever get into a state where things are off.  I was trying to foolproof things for them, but I can't remember if I actually _saw_ this as a problem, or figured it would be

components/filter/filter.js
svanherk: Do we know that the order will always be as expected?  As in, Lit will go through in the order they appear and I'll get the events in the same order as the DOM nodes?

components/filter/filter.js
dlockhart: I'm not sure either what's needed. Might be worth doing a quick experiment to see if keeping these properties up-to-date ourselves helps or hinders.

components/filter/filter.js
dlockhart: The event ordering is my biggest worry with this approach... the DOM order but also that the events might fire before the parent is even listening. I think we've had issues with that before. So walking the DOM may be unavoidable.

svanherk: Looks like I'd need to hookup the event listener in `connectedCallback`.  The order is as expected for my little test, but I'm less confident once we get HM calls involved.  And what if someone slotted in another "optional" dimension between two others after the fact, when a HM call completed and they determined they did want that dimension?

The `slotchange` event really _shouldn't_ be firing very often - twice at the beginning, and then never again unless you add or remove dimensions or options.  Adding options is probably the most likely case if loading/paging.  What I _could_ do to try to help that is put each component more in charge of it's own data.  The `dimension-set-value` maintains its `info` object that the `dimension-set` points to, and it maintains its own `info` object that the `_dimensions` array in filter points to.  Then, if something changes, the component updates it's own info and tells filter to `requestUpdate` - that way, the `slotchange` event doesn't cause a dom crawl of _all_ the dimension nodes when an option changes.

dlockhart: Right, so maybe just better to leave things as you have them. I'm not super worried about crawling the whole DOM if it only happens rarely, was mostly just trying to avoid _any_ DOM crawling... but seems like that's not worth it given the ordering issues you outlined.

components/filter/demo/filter.html
```diff
@@ -0,0 +1,79 @@
...
+						<d2l-filter-dimension-set-value key="admin" text="Admin"></d2l-filter-dimension-set-value>
+						<d2l-filter-dimension-set-value key="instructor" text="Instructor"></d2l-filter-dimension-set-value>
+						<d2l-filter-dimension-set-value key="student" text="Student"></d2l-filter-dimension-set-value>
+					</d2l-filter-dimension-set>
+				</d2l-filter>
+			</template>
+		</d2l-demo-snippet>
+	</d2l-demo-page>
+
+	<script type="module">
```
dbatiste, line 69: Nit: I think this could become:

```javascript
<script>
	document.addEventListener('d2l-filter-change', e => {
		console.log(`Filter selection changed for dimension "${e.detail.dimension}":`, e.detail.value);
	});
</script>
```

The `<script>` could also be placed in the demo-snippet's `template` for purpose of example, but in that case care must be taken to wire up to the correct `d2l-filter`.

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+				@d2l-list-selection-change="${this._handleChangeSetDimension}"
+				extend-separators>
+
+				${dimension.values.map(item => html`
+					<d2l-list-item
+						key="${item.key}"
+						selectable
+						?selected="${item.selected}"
+						slim>
+
```
dbatiste, line 165: Does this whitespace get rendered?  I am assuming it gets ignored... 🤔

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+		`;
+	}
+
+	_dispatchChangeEvent(eventDetail) {
+		this.dispatchEvent(new CustomEvent('d2l-filter-change', { bubbles: false, composed: false, detail: eventDetail }));
+	}
+
+	_getSlottedNodes(slot) {
+		const nodes = slot.assignedNodes();
+		return nodes.filter((node) => node.nodeType === Node.ELEMENT_NODE && node.tagName.toLowerCase() !== 'template');
```
dbatiste, line 179: Is the `template` check just to prevent breakage if they are using unexpectedly, or is there a scenario where we expect them to include a `template`?

components/filter/filter.js
dbatiste: I think it should be ok either way - it'll do a single update after the current task.

components/filter/filter.js
```diff
@@ -0,0 +1,270 @@
...
+		this._activeDimensionKey = e.detail.sourceView.id;
+	}
+
+	_handleDropdownClose(e) {
+		this._activeDimensionKey = null;
+		this._stopPropagation(e);
+	}
+
+	_handleSlotChange() {
+		const dimensionNodes = this._getSlottedNodes(this._dimensionsSlot);
```
dbatiste, line 234: This can just be:
```javascript
const dimensionNodes = this._getSlottedNodes(e.target);
```
... which means the `_dimensionsSlot` variable currently isn't needed.

components/filter/filter.js
dbatiste: The initial dispatch of the event could be delayed if necessary.  For the case where a dimension is inserted afterwards, the handler could determine its index and update the filter accordingly.  However, I think the `slotchange` is probably safer from an ordering/timing perspective.

components/filter/filter.js
dbatiste: Yeah I think it would be best to move this slot out of the menu.  `d2l-menu` does wire up to its `slotchange` event, and while it might tolerate non-menu-item elements, it does some work that is ideally avoided.

components/filter/filter.js
dlockhart: I was going to mention this as a "nit" too... I'm always paranoid about whitespace in HTML templates as they can result in extra text nodes, but likely not a real issue.

components/filter/filter.js
svanherk: So I copied this from `menu-item` and `overflow-group` and _thought_ it was related to handling `dom-if` and `dom-repeat` in Polymer, but from some testing it looks like those do just show up as `dom-repeat`.  It seems like I actually need to filter out the `dom-repeat`s if we want someone to be able to use this in Polymer - is this expected?  How have we not run into this before? 🤔

dbatiste: I suspect that the template check was there for menus because that the time, `is` was supported by Polymer and the consumer could have used `<template is="dom-repeat">`, etc.  I think that `d2l-menu` looks specifically for elements with the menu-item roles and ignores the rest. It's certainly worth considering whether consumers will use those, but ideally this component can just ignore elements it doesn't expect - I _think_ we know what those elements will be ya?

components/filter/demo/filter.html
svanherk: So I actually had it this way because I currently have the event set not to bubble: https://github.com/BrightspaceUI/core/pull/1301/files#diff-371eee884788627e4694af0044b256de9257505f6ee6430c4302574842f8b441R174
I figured I'd start off more restrictive with `bubbles: false` and `composed: false` and we can always open it up later.  This may make things annoying when I get to the coordinator of filter and filter tags, depending on how we do that.  Do you think these should just be true from the start?

components/filter/filter.js
svanherk: I can remove it lol, I just hate the look of having the indented properties and then a new component at the same level.  It means I can't tell the hierarchy _quite_ as quick. This would solve that, but I don't think we do this anywhere and it's also not very nice looking:
```
	<d2l-list-item
		key="${item.key}"
		selectable
		?selected="${item.selected}"
		slim
	>
		<div class="d2l-filter-dimension-set-value-text">${item.text}</div>
	</d2l-list-item>
```

components/filter/filter.js
svanherk: Yeah that's what I was thinking to, I was expecting the old school `is` version (already forgetting how to Polymer 😛 ).  You're right, I'll just look for the elements I expect, since we don't want anything else anyways.

components/filter/demo/filter.html
dbatiste: Ah I see.  So `composed: false` for sure.  I tend to be less restrictive with the bubbling of events unless I have a reason to not bubble.  Most native browser events bubble and I think people usually expect them to.   As an example, `focus` does not (not sure why they made that choice... but it was decades ago), and after many years of requests, they ended up adding new `focusin`/`focusout` events that _do_ bubble.  But yeah not a huge deal if you decide not to... but changing it to bubble later would be potentially breaking. 

For the demo, even if it doesn't bubble, the demo script could capture:

```html
<script>
	document.addEventListener('d2l-filter-change', e => {
		console.log(`Filter selection changed for dimension "${e.detail.dimension}":`, e.detail.value);
	}, { capture: true });
</script>
```
But it's just demo code, so nbd.

components/filter/filter.js
svanherk: Looked at two current but different HM usages (the shared HM filter and My Courses) and both are already keeping their own data up-to-date and updating the component upon receiving the event anyways.  Removing this until we see a use case, but I think you're right - it's not hard for the consumer to do this if they need to, and it seems to fit the natural mental model.

components/filter/filter.js
svanherk: I updated this anyways to make the intent clearer, even if Lit would handle it as expected

components/filter/filter.js
```diff
@@ -0,0 +1,253 @@
...
+		switch (dimension.type) {
+			case 'd2l-filter-dimension-set':
+				dimensionHTML = this._createSetDimension(dimension);
+				break;
+		}
+		return html`
+			<d2l-hierarchical-view
+				@d2l-hierarchical-view-show-complete="${this._handleDimensionShowComplete}"
+				@d2l-hierarchical-view-show-start="${this._handleDimensionShowStart}"
+				data-key="${dimension.key}">
```
svanherk, line 116: I previously used `id` here, but in the HM filter framework guids are used as the keys for a dimension, and an `id` can't start with a number.  Is this a good use case for data attributes instead?  I figured this was cleaner than still using `id` and adding `dim-` to the front of it, and then striping that off all the places below where I use this value to grab dimension data (places can be seen in https://github.com/BrightspaceUI/core/pull/1301/commits/96c9cb5054fc0090ff6315dd2726ad677ed7c49e)

dlockhart: I think so, yeah. Like `id`s have all kinds of other restrictions on them (uniqueness, format, etc.) and the browser ends up treating them specially for CSS/creating hashes for quick lookup... so if we don't actually _need_ to reference the thing by id, no need to use one.

svanherk:
Ok, this is ready for review!  The axe test failure is expected, until we deal with the list label issue.  If we're ready to do that now I can work on that, or I can skip the test until we are.

components/filter/filter-dimension-set-value.js
```diff
@@ -0,0 +1,43 @@
...
+		changedProperties.forEach((oldValue, prop) => {
+			if (oldValue === undefined) return;
+
+			if (prop === 'selected' || prop === 'text') {
+				changes.set(prop, this[prop]);
+			}
+		});
+
+		if (changes.size > 0) {
+			this.dispatchEvent(new CustomEvent('d2l-filter-dimension-set-value-data-change', { detail: { valueKey: this.key, changes: changes }, bubbles: true, composed: false }));
```
dlockhart, line 37: Nit: split out onto multiple lines to improve readability

components/filter/filter.js
```diff
@@ -0,0 +1,253 @@
...
+ * @fires d2l-filter-change - Dispatched when a dimension's value(s) have changed
+ */
+class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
+
+	static get properties() {
+		return {
+			/**
+			 * Disables the dropdown opener for the filter
+			 */
+			disabled: { type: String, reflect: true },
```
dlockhart, line 29:
```suggestion
			disabled: { type: Boolean, reflect: true },
```

components/filter/filter.js
```diff
@@ -0,0 +1,253 @@
...
+			_activeDimensionKey: { type: String },
+			_dimensions : { type: Array }
```
dlockhart, line 31: I'm not totally sure what the default is here, but you may want `attribute: false` for these just to make sure people can't go `<d2l-filter _dimensions="blah">` and mess stuff up.

components/filter/filter.js
```diff
@@ -0,0 +1,253 @@
...
+
+	static get styles() {
+		return [bodyStandardStyles, css`
+			div[slot="header"] {
+				padding: 18px 6px;
+			}
+			.d2l-filter-dimension-header {
+				align-items: center;
+				display: flex;
+				width: 100%;
```
dlockhart, line 43: `<div>` should be 100% width by default since they're block elements, unless the `flex` messes this up. You're also using flexbox here but not setting any `flex` on the items inside... is it just needed to get that nice center alignment?

components/filter/filter.js
```diff
@@ -0,0 +1,253 @@
...
+			return html`<d2l-menu-item text="${dimension.text}">
+				${builtDimension}
+			</d2l-menu-item>`;
+		});
+	}
+
+	_buildHeader() {
+		if (!this._activeDimensionKey) return null;
+
+		const dimensionText = this._dimensions.find(dimension => dimension.key === this._activeDimensionKey).text;
```
dlockhart, line 132: Not sure if it's possible for the dimension to not be found... maybe they dynamically removed a dimension? 🤷

dlockhart (commented):
This is looking really good! Submitting some comments now but will have another look in the morning with fresh 👀 .

components/filter/filter.js
svanherk: Yeah like there _should_ always be a dimension here, because if we're rendering a header they _just_ clicked that dimension.  But then maybe it got dynamically removed after that?  There were a couple places where I ran into this conceptual problem - there _should_ be a dimension, so is it better to let it error out and show up in our js logs so we can fix the place that's adding/removing dimensions weirdly?  Or should we err on the side of trying to gracefully handle those weird scenarios?

components/filter/filter.js
svanherk: I want GitHub to add the 🤦 emoji to their core 8 reactions

components/filter/filter.js
dlockhart: Yeah seems like not worth worrying about unless you notice it as an issue in practice. 👍

components/filter/filter.js
svanherk: Yeah looks like the `width: 100%;` isn't needed here.  I didn't need to set any flex values inside because all the defaults worked as intended when I had `width: 100%;` on the `.d2l-filter-dimension-header-text` div, but I can switch that to `flex-grow: 1;` which reflects the intention better. There's a lot more stuff that will be in the header eventually so I think some of this is "leftover" from having that in the POC and removing it for now.

components/filter/test/filter.axe.js
```diff
@@ -0,0 +1,52 @@
...
+		it(`${test.name} opened`, async() => {
+			const elem = await fixture(test.fixture);
+			const dropdown = elem.shadowRoot.querySelector('d2l-dropdown-button-subtle');
+			dropdown.toggleOpen();
+			await oneEvent(dropdown, 'd2l-dropdown-open');
+			await expect(elem).to.be.accessible();
+		});
+	});
+
+	it.skip('Multiple dimensions drilling in', async() => {
```
svanherk, line 41: Confirmed that @dbatiste's changes fix this test.  Once this PR merges, I'll update Dave's to turn this test on.

dbatiste: Sounds good.

dlockhart (approved):
This looks like a good initial state for this -- I'm OK to merge and then continue from there!

PR #1310, opened 2021-05-04
more visual diff tests

dlockhart:
Discovered a few more scenarios that I'd like to visual diff test, so this adds some new tests to the baseline.

Expecting visual diff to fail until the new goldens are in.


components/table/test/table.sticky.visual-diff.html
```diff
@@ -91,7 +91,6 @@
 				<table class="d2l-table">
 					<thead>
 						<tr>
-							<th rowspan="2" style="width: 1%" class="top"><input type="checkbox"></th>
```
dlockhart, line 94: Taking selection out of the "multi-row" header tests and creating 3 separate tests for selection below.

components/table/test/table.sticky.visual-diff.html
```diff
@@ -229,6 +356,23 @@
 				</table>
 			</d2l-table-wrapper>
 		</div>
+		<div class="visual-diff">
+			<d2l-table-wrapper sticky-headers id="one-column">
```
dlockhart, line 360: This one is actually broken (no bottom border), so I'll try to fix it with the new code.

svanherk (approved):

PR #1313, opened 2021-05-05
Wire-up for pending-state in render for d2l-dialog and d2l-dialog-fullscreen.

dbatiste:
This PR fixes and issue where async dialogs could remain in their initial loading state forever if the dialog's local DOM is re-rendered.  In such a case, the async container would no longer have a handler for the `pending-state` event because it was previously being wired up in the `AsyncContainerMixin` `firstUpdated` method, which only runs once. The fix is to wire up for `pending-state` in `render`.

### DIFF
--- a/mixins/async-container/async-container-mixin.js
+++ b/mixins/async-container/async-container-mixin.js
@@ -24,17 +24,14 @@ export const AsyncContainerMixin = superclass => class extends superclass {
 		super();
 		this._initializeAsyncState();
 		this.asyncPendingDelay = 0;
-		if (!this.asyncContainer) {
-			this.addEventListener('pending-state', this._handleAsyncItemState.bind(this));
+		this._handleAsyncItemState = this._handleAsyncItemState.bind(this);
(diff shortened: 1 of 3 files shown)


mixins/async-container/async-container-mixin.js
```diff
@@ -24,17 +24,14 @@ export const AsyncContainerMixin = superclass => class extends superclass {
...
-			this.addEventListener('pending-state', this._handleAsyncItemState.bind(this));
+		this._handleAsyncItemState = this._handleAsyncItemState.bind(this);
+		if (!this.asyncContainerCustom) {
+			this.addEventListener('pending-state', this._handleAsyncItemState);
 		}
 	}
 
-	firstUpdated() {
-		if (!this.asyncContainer) return;
-		const container = this.asyncContainer();
```
dbatiste, line 34: Only `d2l-dialog` and `d2l-dialog-fullscreen` were providing a `asyncContainer` method, so this is safe to remove.

dlockhart (approved):

PR #1316, opened 2021-05-06
Migrate table-wrapper to Lit

dlockhart:
This is the bulk of the work to migrate `d2l-table-wrapper` (and `d2l-table-col-sort-button`) to Lit from Polymer.

Things I'm going to add in follow-up PRs:
- README
- The visual diffs are complete, but I want to improve the demo
- I still haven't decided if I'm going to replace some of the random attribute names like `selected` that you have to put on native elements with CSS classes. It'll mean more CSS but a cleaner API. TBD.

If you're comparing this to the old code, I've made some pretty major changes to how the everything works. This has resulted in 415 fewer lines of CSS (5.3 KB down from 26.7 KB) and 170 fewer lines of JS (6.9 KB down from 11.1 KB).

Summary of the major changes:
- `d2l-table-col-sort` is gone as it was only used in the LMS and [I'm removing that usage](https://github.com/Brightspace/lms/pull/10297)
- All the `d2l-table`, `d2l-td`, `d2l-tr`, etc. components did not get ported as these were just for IE11
- The global CSS properties defined in `d2l-table-shared-styles.js` have moved into `d2l-table-wrapper`. I searched for and fixed any places that were referencing them for use outside of a table.
- Speaking of those variables, I removed the `--d2l-table-light-*` ones
- I've switched the horizontal lines on rows from the top to the bottom. This allowed the sticky and non-sticky CSS to share the same code (deleting a LOT of CSS), but the caveat is we need to put a CSS class on the previous sibling row of selected rows in JavaScript. If the selected row is the first one, this results in a grey AND blue border -- I think it's worth it.
- Just lots of CSS refactoring, like removing duplicates where `[type="default"]` and `[type="light"]` were repeated

Based on local testing, I'm expecting about 90 visual diff "failures" that are actually what we want:
- `d2l-table-col-sort-button` tests don't exist in the base branch as I just couldn't get that component working with Polymer inside core due to import problems with `d2l-icon`
- Various 0.5px -> 2px diffs from anti-aliasing of the rounded corners or input corners -- not sure why these happen, it's very frustrating
- In multi-row headers with row spans, I've fixed a double border
- Because the horizontal borders have moved from the top to the bottom, some of the diffs that are in a scrolled state no longer show one of the borders
- In the 1-column test, previously there was a bug where the bottom border wasn't rendering -- now it does
- In the 1-cell and no-header tests for the "light" style, previously it rendered a top border and now we don't (this first row in light style tables shouldn't have a top border... although it not having a header is a weird edge case anyway)
- Tests that show selection of the first row now have an extra border so are `1px` taller -- this is a result of the simplified approach described above.
- RTL tests the vertical lines moved over one cell as a result of a simplified approach


components/demo/demo-snippet.js
```diff
@@ -35,12 +34,8 @@ class DemoSnippet extends LitElement {
 			}
 			.d2l-demo-snippet-demo {
 				flex: 1 1 auto;
-				overflow: hidden;
 				position: relative;
 			}
-			:host([allow-overflow]) .d2l-demo-snippet-demo {
-				overflow: visible;
-			}
```
dlockhart, line 43: Going back on these changes as they clipped tooltips and stuff -- just made a change to the scroll-wrapper demo to do this a different way.

components/table/demo/table.html
```diff
@@ -0,0 +1,27 @@
...
+		</script>
+	</head>
+	<body unresolved>
+
+		<d2l-demo-page page-title="d2l-table">
+
+			<h2>Demo 1</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-test-table></d2l-test-table>
```
dlockhart, line 21: I'll improve the demos later -- I spent all my time in the visual diffs.

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+		height: var(--d2l-table-cell-height);
+		padding: var(--d2l-table-cell-padding);
+		text-align: left;
+		vertical-align: middle;
+	}
+	d2l-table-wrapper[dir="rtl"] .d2l-table > * > tr > * {
+		text-align: right;
+	}
+
+	/* default cells */
```
dlockhart, line 32: Was able to save a lot of CSS to reverse things like borders in RTL by using the `:not([dir="rtl"])` -- very handy!

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+	.d2l-table > tbody > tr[selected] > *,
+	.d2l-table > * > tr.d2l-table-selected-previous > * {
+		border-bottom-color: var(--d2l-table-row-border-color-selected);
+	}
+	.d2l-table > * > tr.d2l-table-selected-first > * {
+		border-top: 1px solid var(--d2l-table-row-border-color-selected);
+	}
+
+	/* no-column-border */
+	d2l-table-wrapper[type="default"]:not([dir="rtl"]) .d2l-table[no-column-border] > tbody > tr > *:not(.d2l-table-cell-last) {
```
dlockhart, line 96: This `no-column-border` thing is something I want to move to `<d2l-table-wrapper>` -- will investigate as a future change.

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+				--d2l-table-row-background-color-selected: var(--d2l-color-celestine-plus-2);
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			:host([type="light"]) {
+				--d2l-table-cell-height: 1.15rem; /* min-height to be 48px including border */
+				--d2l-table-cell-padding: 0.6rem;
+				--d2l-table-border-color: var(--d2l-color-gypsum);
+				--d2l-table-header-background-color: #ffffff;
```
dlockhart, line 200: Was able to leverage CSS variables here to override some of the values for "light" tables, saving a lot of CSS.

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		if (this._tableObserver) this._tableObserver.disconnect();
+	}
+
+	render() {
+		const slot = html`<slot @slotchange="${this._handleSlotChange}"></slot>`;
+		if (this.stickyHeaders) {
```
dlockhart, line 219: Previously it would tell scroll-wrapper that it was "sticky" and then scroll wrapper would try to basically just do nothing. Easier to just not render it.

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+		}
+	}
+
+	updated(changedProperties) {
+		// hack: grades/groups/outcomes in the LE use this CSS class on the
+		// body to apply special CSS to the page when tables are sticky
+		// Ideally they should be adding this class to the body.
+		if (changedProperties.has('stickyHeaders')) {
+			if (this.stickyHeaders) {
+				document.body.classList.add('d2l-table-sticky-headers');
```
dlockhart, line 232: I definitely hate this. I may try to see if I can get the CSS class added in the LMS only, but for now this keeps things from breaking.

components/table/test/table-test-sticky-visual-diff.js
```diff
@@ -0,0 +1,368 @@
+import '../../../components/inputs/input-number.js';
+import '../../../components/inputs/input-text.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { tableStyles } from '../table-wrapper.js';
+
+const url = new URL(window.location.href);
+const type = url.searchParams.get('type') === 'light' ? 'light' : 'default';
+
```
dlockhart, line 8: All the visual-diff stuff had to move into Lit components so that they could apply the table styles. The tests themselves are the same.

components/table/test/table.visual-diff.js
```diff
@@ -1,6 +1,20 @@
...
+		return {
+			x: elem.offsetLeft - leftMargin,
+			y: elem.offsetTop - topMargin,
+			width: elem.offsetWidth + (leftMargin * 2),
+			height: elem.offsetHeight + (topMargin * 2)
+		};
+	}, id);
+}
+
 describe('d2l-table', () => {
```
dlockhart, line 18: Just some refactoring in here to deal with the sticky tests coming in here and things being in shadow roots. The tests themselves haven't changed.

components/table/demo/table-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+	static get properties() {
+		return {
+			type: { type: String },
+			stickyHeaders: { attribute: 'sticky-headers', type: Boolean }
+		};
+	}
+
+	static get styles() {
+		return tableStyles;
```
dbatiste, line 15: By the way, I did confirm that the imported styles end up in their own constructed/adopted stylesheet that is shared across usages.  And FWIW, it is also possible to provide a constructed stylesheet here as well - I _think_ it was Lit-element version 2.4 that added that capability.

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+			}
+		}
+	}
+
+	async _applyClassNames(table) {
+
+		await new Promise((resolve) => requestAnimationFrame(() => resolve()));
```
dbatiste, line 239: I think this could just be (brackets could also be removed but that's purely stylistic thing):
```javascript
await new Promise((resolve) => requestAnimationFrame(resolve));
```

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+
+		// offsetParent causes reflow/paint so do them all at once
+		const rows = Array.from(table.rows);
+		let firstRow = null;
+		let lastRow = null;
+		rows.forEach((r) => {
```
dbatiste, line 245: Nit: this could become:
```javascript
rows.forEach((r) => {
	if (r.offsetParent === null) return;
	if (firstRow === null) {
		firstRow = r;
	}
	lastRow = r;
});
```

components/table/table-wrapper.js
```diff
@@ -0,0 +1,318 @@
...
+			});
+
+			prevRow = r;
+			skipFirst = Math.max(0, --skipFirst);
+
+		});
+
+		if (this.stickyHeaders && topHeaderHeight > -1) {
+			const offset = this.type === 'default' ? -3 : 1; // default: -5px top + 2px border, light: 0 top + 1px border
+			const ths = Array.from(table.querySelectorAll('tr[header]:not(:first-child) th, thead tr:not(:first-child) th'));
```
dbatiste, line 289: Do we ever need to worry about nested tables?

dlockhart: We've tried to worry about them in the sense that our CSS just targets the rows/cells that are directly in the `.d2l-table` table.

But you're right that this particular sticky-header query is going to find nested table rows as well. It's the same as it was (so I haven't made it worse!) at least. This approach is also pretty broken... like it works for a sticky header with two rows (like grades), but if you had 2+ rows it wouldn't work.

dbatiste (approved):

dlockhart:
For posterity, here's the visual-diff report of the "approved" changes that got merged in here:
https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/table/2021-05-07-18.38.42.902-540bc7945bde39724023d3c15a5550e4dfefdf53-report.html

components/table/table-wrapper.js
```diff
@@ -0,0 +1,319 @@
...
+		// hack: grades/groups/outcomes in the LE use this CSS class on the
+		// body to apply special CSS to the page when tables are sticky
+		// Ideally they should be adding this class to the body.
+		if (changedProperties.has('stickyHeaders')) {
+			if (this.stickyHeaders) {
+				document.body.classList.add('d2l-table-sticky-headers');
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
dlockhart: Yeah I noticed that too and put some notes in `US127495` (the switchover story) to deal with them. I'm not quite sure how yet... I could stub it in here, remove them all, then remove the stub?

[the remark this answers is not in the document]
svanherk: That seems easiest, compared to reworking the old component to not need consumers to call that function and removing it now (assuming the function _is needed_ right now)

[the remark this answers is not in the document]
dlockhart: I still hope to remove the calls at some point during cleanup... but we'll see!

PR #1318, opened 2021-05-07
Updating Translations 2105 with fr-ON TELO lang pack custom values 07/05/2021

ghost:


svanherk:
This is identical to `fr.js`

cbellini:
**tldr; closing this as `core` components have an exemption for support fr-ON, due to the langterms being generic/non-customizable.**

Serge.io automatically creates PRs for all repos new content is imported. There likely won't be future changes to `fr-ON` through Serge after this initial inclusion, but it `core` should be excluded from all future translations.

PR #1326, opened 2021-05-10
Table improvements

dlockhart:
A few table improvements:

1. README
2. Better demo page/element
3. Switched `no-column-borders` to be an attribute on the table wrapper instead of a non-standard attribute on the `<table>`
4. Switched the non-standard `header` attribute on header rows outside of `thead` to be a CSS class
5. Switched the non-standard `sticky` attribute on non-header cells to be a CSS class

I've left the non-standard `selected` attribute as-is until Dave B and I can sort out how selection is going to work.


components/table/table-wrapper.js
```diff
@@ -44,12 +44,14 @@ export const tableStyles = css`
 
 	/* header cells */
 	.d2l-table > thead > tr > th,
+	.d2l-table > * > tr.d2l-table-header > th,
 	.d2l-table > * > tr[header] > th {
```
dlockhart, line 48: Still going to support the old way, at least until I can go through and fix them all... or possibly forever if that's too much work.

components/table/table-wrapper.js
```diff
@@ -94,26 +102,30 @@ export const tableStyles = css`
...
 	}
 
 	/* sticky-headers */
 
 	/* all header cells */
-	d2l-table-wrapper[sticky-headers] .d2l-table > * > tr[header] > *,
-	d2l-table-wrapper[sticky-headers] .d2l-table > thead > tr > th {
+	d2l-table-wrapper[sticky-headers] .d2l-table > thead > tr > th,
+	d2l-table-wrapper[sticky-headers] .d2l-table > * > tr.d2l-table-header > *,
+	d2l-table-wrapper[sticky-headers] .d2l-table > * > tr[header] > * {
```
dlockhart, line 119: Just flipped these around a bit so that it goes in order of goodness: `thead`, then the CSS class, then the legacy attribute.

components/table/demo/table-test.js
```diff
@@ -1,61 +1,103 @@
...
 	}
 
 	render() {
 		const type = this.type === 'light' ? 'light' : 'default';
+		const sorted = data.sort((a, b) => {
+			if (this._sortDesc) {
+				return b.fruit[this._sortField] - a.fruit[this._sortField];
+			}
+			return a.fruit[this._sortField] - b.fruit[this._sortField];
+		});
```
dlockhart, line 49: Went the extra mile here and actually added client-side sorting to the demo!

components/table/demo/table.html
```diff
@@ -15,13 +15,35 @@
...
+			<d2l-demo-snippet>
+				<template>
+					<d2l-test-table type="light"></d2l-test-table>
+				</template>
+			</d2l-demo-snippet>
+
+			<h2>Light, Sticky</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-test-table type="light"></d2l-test-table>
```
svanherk, line 42: Does this need `sticky-headers`?

components/table/test/table.sticky.visual-diff.html
```diff
@@ -1,384 +0,0 @@
-<!DOCTYPE html>
```
svanherk, line 1: Was this file deletion just missed in a previous refactor?

dlockhart: Yeah I definitely deleted it, but it may have snuck back in by accident as a conflict in one of the many rebases I've had to do to keep all these branches in sync.

components/table/test/table.visual-diff.js
```diff
@@ -87,15 +94,23 @@ describe('d2l-table', () => {
...
+						before(async() => {
+							await page.$eval('d2l-test-table-visual-diff', elem => elem.setAttribute('hidden', 'hidden'));
+							await page.$eval('d2l-test-table-sticky-visual-diff', elem => elem.removeAttribute('hidden'));
+						});
+
 						[
 							'one-row-thead',
-							'one-row-no-thead',
+							'one-row-no-thead-class',
+							'one-row-legacy-attr',
```
svanherk, line 105: I think this needs to be `one-row-no-thead-attr`?  And `multi-row-legacy-attr` should be `multi-row-no-thead-attr`.

svanherk (approved):

PR #1328, opened 2021-05-11
Re-activate axe tests

GeorgePlukov:
A bunch of aXe tests were incorrectly calling `expect(elem).to.be.accessible` instead of `expect(elem).to.be.accessible()`. I fixed the places where it was being called incorrectly and also did a global code search but it seems to be isolated to core.

### DIFF
--- a/components/inputs/test/input-checkbox.axe.js
+++ b/components/inputs/test/input-checkbox.axe.js
@@ -7,15 +7,19 @@ describe('d2l-input-checkbox', () => {
 		const elem = await fixture(html`<d2l-input-checkbox aria-label="basic"></d2l-input-checkbox>`);
 		await expect(elem).to.be.accessible();
 	});
-	// failing
+
 	it('checked', async() => {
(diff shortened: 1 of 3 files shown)


components/inputs/test/input-checkbox.axe.js
```diff
@@ -5,17 +5,22 @@ describe('d2l-input-checkbox', () => {
...
 	});
 
 	it('checked', async() => {
-		const elem = await fixture(html`<d2l-input-checkbox checked></d2l-input-checkbox>`);
-		await expect(elem).to.be.accessible;
+		const elem = await fixture(html`<d2l-input-checkbox aria-label="basic" checked></d2l-input-checkbox>`);
+		await expect(elem).to.be.accessible();
+	});
+
+	it('is accessible when provided a label', async() => {
```
GeorgePlukov, line 16: I added a new test here as I believe its valid to provide a label through a slot or the `aria-label` attribute

PR #1330, opened 2021-05-12
fix: NVDA reads buttons instead of dialog text

semibran:
Rally: [US125198](https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fuserstory%2F502279656304)

EDIT - Using `autofocus` is unnecessary here, culprit appears to be a RAF that occurs prior to initial focus

~~Overrides~~ Addresses an accessibility defect with the `dialog-confirm` component where NVDA will occasionally read off the one element focused by `_focusFirst` instead of the actual dialog text. This issue is present in all dialogs that use the dialog mixin.

~~The native `<button>` autofocus attribute (accessible through d2l-button) appears to perform the same functionality while avoiding DOM traversal and retaining screen reader compatibility.~~

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -315,10 +315,21 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		// native dialog backdrop does not prevent body scrolling
 		this._bodyScrollKey = preventBodyScroll();
 
+		// focus first focusable child prior to auto resize (fixes screen reader hiccups)
+		this._focusInitial();
+
 		requestAnimationFrame(async() => {
 			await this._updateSize();
 			this._state = 'showing';
-			this._focusInitial();
+
+			// edge case: no children were focused, try again after one redraw
+			const activeElement = getComposedActiveElement();
+			if (!activeElement
+			|| !isComposedAncestor(dialog, activeElement)
+			|| !activeElement.classList.contains('focus-visible')) {
+				this._focusInitial();
+			}
+
 			if (!reduceMotion) await animPromise;
 			this.dispatchEvent(new CustomEvent(
 				'd2l-dialog-open', { bubbles: true, composed: true }


components/dialog/dialog-confirm.js
```diff
@@ -82,17 +82,7 @@ class DialogConfirm extends DialogMixin(LitElement) {
 	}
 
 	_focusInitial() {
-		const footer = this.shadowRoot.querySelector('.d2l-dialog-footer-slot');
-		const nodes = footer.assignedNodes();
-		for (let i = 0; i < nodes.length; i++) {
-			const node = nodes[i];
-			if (node.nodeType !== Node.ELEMENT_NODE) continue;
-			if (!node.hasAttribute('primary')) {
-				forceFocusVisible(node);
```
dlockhart, line 91: Does this still result in a visible focus ring now that the browser's autofocus is used instead of our `forceFocusVisible`?

semibran: It does once the user presses Tab or Shift - I agree there's a bit of a tradeoff in that respect, but it's between having a visible focus ring and potentially hindering visually-impaired users from using the dialog effectively

dlockhart: Hmm, I see. This change also effectively breaks all existing confirmation dialogs and requires that they all switch over to applying `autofocus` on the primary button... plus all new usages will need to do that as well -- something that could be easily missed or applied inconsistently to the wrong button.

Plus as you mentioned, this doesn't fix the underlying NVDA problem with all non-confirmation dialogs.

Add to that, `autofocus` itself is a little bit of a mystery to me... like [the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) describe it as:
> This Boolean attribute specifies that the button should have input focus when the page loads. Only one element in a document can have this attribute.

Both "when the page loads" and "only one element in a document" parts worry me. Showing/hiding the dialog doesn't constitute a "page load" (there are no iframes here) and a page could easily have multiple dialogs... it's unclear how the browser will handle this.

All this leads me to... I'm not sure if this is the best approach. The [dialog design pattern](https://www.w3.org/TR/wai-aria-practices/#dialog_modal) states:
> In all circumstances, focus moves to an element contained in the dialog.

If NVDA occasionally isn't handling this properly, that's a 3rd party bug with them and we should raise it there -- not try to work around it in our stuff.

semibran: That's interesting, I'm curious if autofocus is working because the browser considers the dialog's shadow DOM instance to be a separate page or if it's just because the element is being patched in.

As far as NVDA is concerned, the "incorrect" reading seems like expected behavior, i.e. the dialog opens -> it's focused with JS -> NVDA skips over the dialog text to read the focused element. It's more of a mystery to me how dialogs are read correctly at all within the current setup, though I imagine it's some kind of race condition gimmick.

I definitely see how this could be a maintainability nightmare though. I can try and see if there's a way to add autofocus to incoming slot children automatically.

dlockhart: Do any of the [W3C examples](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) work properly? Because if they don't... then yeah I'm sorry but this is an NVDA thing.

The reason this normally works properly is that when focus moves into the dialog, the AT is supposed to notice that it's now inside an ARIA dialog (or now the native `<dialog>` element) and inform the user of that. It would be the same as other roles like menus where when you open it you move focus to the first menu item.

semibran: NVDA does seem to be reading the modal dialog example correctly..... after doing some digging the incorrect read seems to be a byproduct of [the RAF autosizing fix](https://github.com/BrightspaceUI/core/blob/master/components/dialog/dialog-mixin.js#L318-L326), which focuses the button a split second after the dialog opens.

dlockhart: A timing issue like this seems a lot more probable to me. @dbatiste that changed was [introduced here](https://github.com/BrightspaceUI/core/commit/e606c636873314d4fac3e5f3df4b5cd72165bbd0) to address some sizing issues. Not sure if we could move the focus bits outside the of rAF?

dbatiste: It's possible that focus may not get applied correctly since the dialog `opened` (which changes the dialog's `display` property) state may not be reflected in the DOM yet, but it's worth a try.

components/dialog/dialog-mixin.js
```diff
@@ -315,10 +316,19 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		// focus first focusable child prior to auto resize (fixes screen reader hiccups)
+		this._focusInitial();
+
 		requestAnimationFrame(async() => {
 			await this._updateSize();
 			this._state = 'showing';
-			this._focusInitial();
+
+			// edge case: no children were selected, try again after one redraw
+			if (!dialog.contains(document.activeElement)
```
dbatiste, line 327: Does this check assume that there would be no other slots defined by the dialog?

Note, I think the composed active element is what this logic would be interested in, because if the dialog is wrapped in another web component and focus is within it, the `document.activeElement` would be that component, in the context of the outer document.

A safer approach that I think avoids these assumptions is to use the `getComposedActiveElement` and `isComposedAncestor` helpers - since it would be operating on the actual focused element, and it wouldn't matter whether it's within a slot or not.

components/dialog/dialog-mixin.js
```diff
@@ -315,10 +316,19 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+			&& !(footer && footer.assignedNodes().includes(document.activeElement))) {
+				this._focusInitial();
+			}
```
dlockhart, line 330: Are you finding that `dialog.contains` is working with shadow roots? Maybe the fact that `document.activeElement` is going to hide the shadow roots is making this work. Curious what it's returning actually.

We have a helper that might make this easier in `helpers/focus.js` called `getComposedActiveElement()` which will return that actual composed element (the `<button>` element inside `d2l-button` inside the footer). You could then check that it's a descendant of the dialog using `isComposedAncestor` helper in `helpers/dom.js`.

semibran: In this case `document.activeElement` is evaluating to the `d2l-button` component instead of the actual composed element, which is why the contains check fails and I assumed the footer slot check would be necessary - though @dbatiste is right about that being less future-proof.

components/dialog/dialog-mixin.js
semibran: > ...if the dialog is wrapped in another web component and focus is within it, the document.activeElement would be that component

Good catch, this is the behavior I'm currently seeing. I've switched them out in favor of the existing helpers ([0a571485](https://github.com/BrightspaceUI/core/pull/1330/commits/0a57148588f91caadb556252cd234158b3cfb6de)).

components/dialog/dialog-mixin.js
```diff
@@ -322,9 +322,11 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			await this._updateSize();
 			this._state = 'showing';
 
-			// edge case: no children in footer slot were selected, try again after one redraw
+			// edge case: no children were focused, try again after one redraw
 			const activeElement = getComposedActiveElement();
-			if (!activeElement || !isComposedAncestor(dialog, activeElement)) {
+			if (!activeElement
+			|| !isComposedAncestor(dialog, activeElement)
+			|| !activeElement.classList.contains('focus-visible')) {
```
semibran, line 329: This entire conditional is becoming a pretty nasty hack, but the last expression seems to be required since the ARIA dialog autofocusing the close button without the focus ring. May or may not be the optimal solution

components/dialog/dialog-mixin.js
dbatiste: Cool yeah, if this were wrapping in another shadowDOM `document.activeElement` would resolve to the element in the outer context, so it all depends on the context in which is it used.

dbatiste (approved):
I think this looks ok to me, but let's wait to see if Dave has anything more to add.

dlockhart (approved):
I'm good with where this ended up. Thanks Brandon for the contribution and sticking with it... I think we ended up solving the root cause here. 💯

emil-furniss:
Hey folks, sorry for bringing this up but it looks like this broke the full-screen dialog when it's opened via `D2L.LP.Web.UI.Legacy.MasterPages.Dialog.OpenFullscreen`. It was working as expected BSI v20.21.6-153 with core v1.131.2, and then it's broken in BSI v20.21.6-153 with this core release v1.131.3. It renders slightly higher than expected exposing the page below and hiding the top of the dialog, and there's a console error related to the focus-trap. As seen here in the FACE assignments "Manage Special Access" dialog:
 
image (5)

dialog-fullscreen-focus-error

I've been looking into it and can't repro in the `dialog-fullscreen` component demo, so I'm thinking it's something to do with how the LMS page handles the render.

dlockhart:
Thanks for letting us know @emil-furniss. Yeah it looks like now that call to `_focusInitial` is happening so much earlier, that in this particular case the body of the dialog doesn't exist so it tries to focus on the focus trap. But the focus trap hasn't even rendered yet so it blows up.

Maybe a `null` check in focus-trap's `focus()` method would avoid the exception and wouldn't actually focus anything, which would allow the second pass in rAF to work as it did before?

emil-furniss:
> Maybe a `null` check in focus-trap's `focus()` method would avoid the exception

Yeah that would solve the error problem. Might there be an impact on the original NVDA screen-reader issue?

dlockhart:
> Might there be an impact on the original NVDA screen-reader issue?

Possibly, although I'm not sure what we can do... if the dialog literally hasn't finished rendering yet then we can't move focus anywhere. I'll look into a fix for this.

PR #1338, opened 2021-05-13
Ignore

svanherk:
Just seeing if we can get to the bottom of the flakiest test


svanherk:
Closing - hopefully https://github.com/BrightspaceUI/core/pull/1357 does the trick.

PR #1340, opened 2021-05-14
US127139: Edit Fullscreen dialog to have a margin to preserve context

annabelsegalld2l:
Made the following changes to the existing d2l-dialog-fullscreen component, consistent with the design at https://xd.adobe.com/view/84fce6d7-52ec-427b-a259-693a98d6bca9-d1af/screen/e7dd9b0a-3d5f-452a-b200-65579f8a68d5?fullscreen.


components/dialog/dialog-fullscreen.js
```diff
@@ -129,8 +164,8 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 			<div class="d2l-dialog-inner">
 				<div class="d2l-dialog-header">
 					<div>
-						<h2 id="${this._titleId}" class="d2l-heading-3">${this.titleText}</h2>
-						<d2l-button-icon icon="tier1:close-small" text="${this.localize('components.dialog.close')}" @click="${this._abort}"></d2l-button-icon>
+						<h3 id="${this._titleId}" class="d2l-heading-3">${this.titleText}</h3>
```
annabelsegalld2l, line 167: I'm not sure if it's possible to have this icon change from the big icon to the small one when the screen is mobile-sized. Any tips for this? I'm thinking we could put both the small close icon and the big one here and toggle visibility based on screen size.

dbatiste: I am assuming the goal is to have a larger button?  To achieve a larger button, I think we need to override these CSS variables that are defined in `d2l-button-icon`.  Depending on the size, it may also be appropriate to bump up the radius for the corners. @geurts fyi... might need specifics.

```css
--d2l-button-icon-min-height
--d2l-button-icon-min-width
--d2l-button-icon-border-radius
```

Ref: https://github.com/BrightspaceUI/core/blob/master/components/button/button-icon.js#L49

Note: the `d2l-button-icon` demo page has a demo (with custom styles) that you can try if you need an example.

dbatiste: And those overrides can be applied in a media-query as required to have a different size for different size viewports.  If we want to change the actual icon, then I think things might become slightly more involved since there may be some focus code that looks for the abort button.

annabelsegalld2l: I believe we do want to change the button as the design doc specifies the thicker version of the close button versus the previous thin version. Not sure if @geurts has a preference if we actually need the thicker button or a larger size would suffice? Let me know 😄

dbatiste: I think for larger viewports we want the larger button and thicker icon, which we can accomplish using the icon currently used in this PR, plus the CSS variables.  

However the design mentions that for smaller devices it should be indistinguishable from the regular dialog, which uses a smaller button and the smaller close icon.  There are a variety of ways to accomplish that, but I would use `matchMedia` in combination with conditionally rendering the button with the appropriate icon.

dlockhart: So I get how we can use the CSS variables to change the size of the button responsively, but how do with change the icon? I don't think we want two buttons that we hide/show for the reasons you mentioned above around dealing with cases where focus is on one of them.

svanherk: Can the icon string itself (`tier1:close-large-thick`) be stored in property that causes a re-render when it changes (and we adjust it on resize)?  Or will that break with the way icons are dynamically imported as needed?

dbatiste: Yeah I think we can just use `matchMedia` (wiring up to its event) to change a property that controls which icon to use.

components/dialog/dialog-fullscreen.js
```diff
@@ -95,12 +160,23 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 		this.async = false;
 		this._autoSize = false;
 		this._hasFooterContent = false;
+		this._handleResize = this._handleResize.bind(this);
```
dlockhart, line 163: Instead of re-calculating every time the window resizes, as Dave B suggested elsewhere you can use `matchMedia` to test for a match but also register to be notified if the match changes. So you could use that in combination with a private property which when changed will automatically trigger a re-render.

```javascript
constructor() {
  this._mediaQueryList = window.matchMedia('(max-width: 616px)');
  this._handleResize();
}

connectedCallback() {
  this._mediaQueryList.addEventListener('change', this._handleResize);
}

disconnectedCallback() {
  this._mediaQueryList.removeEventListener('change', this._handleResize);
}

_handleResize() {
  this._iconType = this._mediaQueryList.matches ? 'tier1:close-large-thick' : 'tier1:close-small'
}
```

svanherk: Dave L beat me too it, was just writing this up 😛

I also just wanted to point out that an important part of Lit components is making `render` a function of the element's properties. By storing the icon type in a property (like `_hasFooterContent`) instead of calculating it inside `render`, we meet that goal and Lit will only call `render` when that value actually changes.  This limits the amount or re-renders and it will handle it automatically for us (no need for us to call `this.requestUpdate();`).

From my experience, this is an easy thing to know and understand and then accidentally overlook when deep in the weeds of a problem, so it's good to keep checking for this along the way (in your own stuff, in pull requests, etc)

annabelsegalld2l: Aha, so that's how you do it! Thanks all. I've made those changes now 😄

components/dialog/dialog-fullscreen.js
```diff
@@ -95,12 +161,26 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 		this.async = false;
 		this._autoSize = false;
 		this._hasFooterContent = false;
+		this._iconType = 'tier1:close-large-thick';
+		this._handleResize = this._handleResize.bind(this);
+		this._mediaQueryList = window.matchMedia('(max-width: 616px)');
+		this._handleResize();
```
dlockhart, line 167: Nit: this line is going to set `this._iconType` anyway, so I don't think you need to set it 2 lines up.

annabelsegalld2l: The reason I added the setting iconType line was because the linter was complaining that the html could be undefined.. I can set it to an empty string?

components/dialog/dialog-fullscreen.js
dbatiste: This API is certainly the preferred one, but I would just test the different browsers to make sure that works as expected.  Safari might have been late to the party supporting `addEventListener` and quite a whole ago when I was first trying to use those for handling color scheme, I had some trouble with it.  I'm not sure if we have anything running previous versions of Safari though.  You _might_ need to include the old method for backwards compatibility... https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#methods

components/dialog/dialog-fullscreen.js
dlockhart: Oh interesting... hmm yeah or just leaving it how you have it is good! 👍

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +25,134 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 			 * Whether to render a loading-spinner and wait for state changes via AsyncContainerMixin
 			 */
 			async: { type: Boolean },
-			_hasFooterContent: { type: Boolean, attribute: false }
+			_hasFooterContent: { type: Boolean, attribute: false },
+			_iconType: { type: String, attribute: false }
```
dbatiste, line 29: Nit: I'd just call this `_icon`, to be consistent with how we name this property elsewhere, and also since it's not really a type of icon.

components/dialog/dialog-fullscreen.js
```diff
@@ -95,12 +161,26 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 		this.async = false;
 		this._autoSize = false;
 		this._hasFooterContent = false;
+		this._iconType = 'tier1:close-large-thick';
+		this._handleResize = this._handleResize.bind(this);
+		this._mediaQueryList = window.matchMedia('(max-width: 616px)');
```
dbatiste, line 166: While we aren't expecting a lot of dialogs on the page, the 
```javascript
this._mediaQueryList = window.matchMedia('(max-width: 616px)');
```
could be moved outside of the class, since we don't really need a different instance for every dialog.

components/dialog/dialog-fullscreen.js
dlockhart: Good memory Dave.

> Prior to Safari 14, MediaQueryList is based on EventTarget, so you must use addListener() and removeListener() to observe media query lists.

IMO as long as it works on first-render in Safari < 14 (which I think it should based on how this code works), I'd be satisfied with that and just stick with the better API.

annabelsegalld2l:
Here are some gifs I recorded. They look way smoother on my machine so I guess the screencap lowered the quality significantly.

d2l-dialog-fullscreen on an LMS page:
activity_fullscreen
Responsive behaviour: 
firefox_responsive
Regular fullscreen dialog:
fullscreen_chrome
Async dialog:
fullscreen_chrome_async
On mobile:
mobile_chrome

dlockhart (approved):
This is looking great to me!

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +27,138 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
 			}
 
-			:host([dir="rtl"]) .d2l-dialog-header > div > d2l-button-icon {
-				margin-left: -15px;
-				margin-right: 15px;
-			}
+			@media (min-width: 616px) {
+				
+				.d2l-dialog-header {
+					border-bottom: 1px solid var(--d2l-color-galena);
```
dbatiste, line 49: Just took this for a spin... the borders (header and footer) looks dark.  I think this is supposed to be gypsum.

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +27,138 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
+				dialog::backdrop {
+					transition: opacity 200ms ease-out;
+				}
+
+				:host([_state="showing"]) dialog::backdrop {
+					transition-duration: 400ms;
+				}
+
+				.d2l-dialog-footer {
+					border-top: 1px solid var(--d2l-color-galena);
```
dbatiste, line 99: I think this one needs to be updated as well.

components/dialog/dialog-fullscreen.js
svanherk: Yeah I also thought these looked dark, but the design _says_ galena which may be a mistake... @geurts?

annabelsegalld2l: Yep, it should be gypsum.

svanherk: Ah, and I see the design has already be updated now... nevermind Jeff! 😄

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +27,149 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
+					margin-left: -15px;
+					margin-right: 15px;
+				}
+				
+				dialog.d2l-dialog-outer,
+				div.d2l-dialog-outer {
+					border-radius: 8px;
+					height: calc(100% - 3rem);
+					margin: 1.5rem;
+					max-width: 1170px;
```
annabelsegalld2l, line 77: The dialog should max out at 1170px, the same as normal page content. Beyond that, center the dialog on the screen.

dbatiste (approved):

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +27,151 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
+				.d2l-dialog-footer.d2l-footer-no-content {
+					padding: 0 0 5px 0;
+				}
+
+				.d2l-dialog-content > div {
+					/* required to properly calculate preferred height when there are bottom
+					margins at the end of the slotted content */
+					border-bottom: 1px solid transparent;
+					/* required to render full height in an i-Frame */
+					height: calc(100% - 1px);
```
annabelsegalld2l, line 154: We need the -1px so that iFrames can render at the full height, but only display the scrollbar when there is an overflow (100% will always display a scrollbar, even with short contents)

components/dialog/dialog-fullscreen.js
```diff
@@ -25,68 +27,151 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
-			.d2l-dialog-content-loading {
-				text-align: center;
-			}
+				.d2l-dialog-content > div {
+					/* required to properly calculate preferred height when there are bottom
+					margins at the end of the slotted content */
+					border-bottom: 1px solid transparent;
+					box-sizing: border-box;
+					height: 100%;
+					padding-bottom: 1rem;
```
annabelsegalld2l, line 64: This will prevent the border from bumping up next to the text.

dlockhart (approved):
The visual diffs look correct to me, as do the rest of the changes in here.

dbatiste (approved):
Looks great!

PR #1342, opened 2021-05-14
Selection and actions for lists.

dbatiste:
This purpose of this PR is to introduce more comprehensive set of selection components for use with lists and potentially other element types such as tables.  It further enhances our list components to support using these new selection components.

Selection Components:

* new `d2l-selection-action` component for actions with optional `requires-selection`
* new `d2l-selection-select-all` component
* new `d2l-selection-checkbox` component
* new `d2l-selection-summary` component
* new `SelectionMixin` for controlling/notifying observers & selectables (used by `d2l-list`)

List Changes:

* new `d2l-list-header` component that provides layout and overflow-group behaviour
* updates `ListItemCheckboxMixin` to render `d2l-selection-checkbox` instead of `input`
* updates `ListItemCheckboxMixin` to not use a label redirect for labelling the checkbox
* updates `ListItemCheckboxMixin` to extend `LabelledMixin`
* updates `d2l-list` to extend `SelectionMixin`
* updates `d2l-list` to include a `header` slot

(the `ListItemChecboxMixin` changes accomplish a couple of things - resolves axe error that existed with the original label redirect approach that was used; enables the use of a shared `d2l-selection-checkbox` component)

Other Changes:

* new `description` property for `d2l-input-checkbox` for use by `d2l-selection-select-all` to provide more info
* updates `d2l-overflow-group` to observe `disabled` and `text` changes to reflect them to overflow menu items
* updates `d2l-overflow-group` to look for `d2l-selection-action` components within its slot


components/selection/selection-mixin.js
```diff
@@ -0,0 +1,119 @@
+/*
+export const selectionTypes = {
```
dbatiste, line 2: Will reintroduce this after I do my spike on single selection.

components/selection/selection-mixin.js
```diff
@@ -0,0 +1,119 @@
...
+			if (selectable.checked) keys.push(selectable.key);
+		});
+
+		let state = selectionStates.none;
+		if (keys.length > 0) {
+			if (keys.length === this._selectionSelectables.size) state = selectionStates.all;
+			else state = selectionStates.some;
+		}
+
+		return {
```
dbatiste, line 62: I am considering wrapping this up in a concrete `SelectionInfo` class.

components/selection/selection-checkbox.js
```diff
@@ -0,0 +1,109 @@
...
+ * @fires d2l-selection-change - Dispatched when the checked state changes
+ */
+class Checkbox extends SkeletonMixin(LitElement) {
+
+	static get properties() {
+		return {
+			/**
+			 * State of the checkbox
+			 */
+			checked: { type: Boolean },
```
dbatiste, line 16: Wondering whether this should be named `selected` instead.

dlockhart: Yeah I like `selected` 🤷 . Or... does this even need to be exposed to consumers? Just wondering if they ever really need to set this state, or if it should always be set by the selection mixin.

components/selection/selection-mixin.js
```diff
@@ -0,0 +1,119 @@
...
+
+	_handleSelectionCheckboxUnsubscribe(e) {
+		e.stopPropagation();
+		const target = e.composedPath()[0];
+		this._selectionSelectables.delete(target);
+	}
+
+	_handleSelectionSelectAllChange(e) {
+		const checked = e.detail.checked;
+		this._selectionSelectables.forEach(selectable => selectable.checked = checked);
```
dlockhart, line 87: If this didn't fire an event when set programatically like this, would that help avoid some of the complex throttling in `_updateSelectionSubscribers`? Like it could iterate through them all here and check them, then call `_updateSelectionSubscribers` itself.

components/selection/selection-mixin.js
```diff
@@ -0,0 +1,119 @@
...
+	_updateSelectionSubscribers() {
+		if (!this._selectionSubscribers || this._selectionSubscribers.size === 0) return;
+
+		// debounce the updates for select-all case
+		if (this._updateSubscribersRequested) return;
+
+		this._updateSubscribersRequested = true;
+		setTimeout(() => {
+			const info = this.getSelectionInfo();
+			this._selectionSubscribers.forEach(subscriber => subscriber.selectionInfo = info);
```
dlockhart, line 114: Curious... I think this is always going to trigger a re-render since the property will always change -- even if the underlying info hasn't. Maybe that's OK though since we're just talking about the summary and the select-all checkbox... which likely always change anyway.

components/selection/selection-summary.js
```diff
@@ -0,0 +1,42 @@
...
+ * A summary showing the current selected count.
+ */
+class Summary extends LocalizeCoreElement(SelectionSubscriberMixin(LitElement)) {
+
+	static get properties() {
+		return {
+			/**
+			 * Text to display if no items are selected.
+			 */
+			noSelectionText: { type: String, attribute: 'no-selection-text' }
```
dlockhart, line 17: Should we provide a nice default for this?

dlockhart:
This looks really good to me. Nicely broken up into reusable bits but also low effort for devs to use. I think this will work well with table too, but will confirm once I get some time to prototype it.

components/list/list-header.js
```diff
@@ -0,0 +1,79 @@
...
+	render() {
+		return html`
+			<div class="d2l-list-header-container">
+				<d2l-selection-select-all></d2l-selection-select-all>
+				<d2l-selection-summary
+					aria-hidden="true"
+					class="d2l-list-header-summary"
+					no-selection-text="${this.localize('components.selection.select-all')}">
+				</d2l-selection-summary>
+				<div class="d2l-list-header-actions">
```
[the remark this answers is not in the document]
dbatiste: That's a good point.  I wasn't thinking about whether we'd always want this overflow behavior, but we probably do.  I'll try to add this - we might run into some slot inception issues depending on whether overflow-group is flattening nodes.

components/list/list-item-checkbox-mixin.js
```diff
@@ -18,6 +18,10 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 			 * Value to identify item if selectable
 			 */
 			key: { type: String, reflect: true },
+			/**
+			 * Label for the checkbox if selectable
+			 */
+			label: { type: String },
```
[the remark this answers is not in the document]
dbatiste: Yep for sure - we definitely should.

components/selection/selection-checkbox.js
dbatiste: Sounds good - I will name it selected.  

I don't have concrete use-case, but I think consumers will need the ability to define list items to be initially selected.  We could keep it private and make it public if/when that comes up.  I don't feel super strongly about it, other than it's nice knowing the subscribes are correctly notified in this scenario. If I didn't have this in the demo page during dev, I might have introduced a couple pitfalls that might have made supporting that later trickier/riskier.

components/list/list-item-checkbox-mixin.js
```diff
@@ -84,33 +89,29 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 	}
 
 	_renderCheckbox() {
-		const disabled = this.disabled || this.skeleton;
 		return this.selectable ? html`
-			<input
+			<d2l-selection-checkbox
+				@d2l-selection-change="${this._onCheckboxChange}"
```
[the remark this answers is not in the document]
dbatiste: Currently the `SelectionMixin` listens for this event, and there could be other elements that also dispatch a `change` event from within the selection control.  There is a similar event for select-all.  We can use `change` but we will need to validate the target (`e.target.composedPath()[0].tagName === 'D2L-SELECTION-CHECKBOX'` I think) in the handlers.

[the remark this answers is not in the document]
dlockhart: Right -- doesn't seem worth it.

components/list/list.js
```diff
@@ -1,17 +1,14 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { SelectionMixin, selectionStates } from '../selection/selection-mixin.js';
 
-export const listSelectionStates = {
-	none: 'none',
-	some: 'some',
-	all: 'all'
-};
```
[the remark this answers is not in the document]
dbatiste: Good question.  Yeah I am not sure if we will need it or not.  Related to that is the question of what "select all" means - that is, is it all on the page, or all for all pages?  I think `all-except` is only really useful if select-all applies to _all_ all. For that scenario with paging, consumers need to figure out what to do on the backend since the client doesn't have all the keys.
 
I think this probably needs some more discussion to figure out what we want design-wise, and what the new APIs are supporting. IIRC, the MVC controls provide a way to select all, and also _all_ all.

components/selection/selection-mixin.js
dbatiste: I think so yes, however I _think_ that if the item state changes as a result of select-all then consumers should be notified if they've wired up for the event.

components/selection/selection-mixin.js
dbatiste: That's a good point, but yeah unless something's gone wrong, I think the info would always be different.  I could add a custom setter to the `SelectionSubscriberMixin` to not re-render in such a case.

components/selection/selection-summary.js
dbatiste: I think there is a bug in the rendering actually - it should default to the existing message `0 selected` or the localized likes thereof.  The `noSelectionText` is just to override that, which `d2l-list-header` does since is places the summary next to the select-all and says "Select All".

dbatiste:
I think I've captured the key "provider" changes that are required from your experimental branch.  I also renamed selection-subscriber to selection-observer for clarity (... I _think_ that's a little clearer), since "subscriber" is a little overloaded in this context.

I still need to make further changes to `d2l-overflow-group` regarding the disabled state and mutation observer that will be required, so the non-interactive stuff is still here for the moment.

components/list/README.md
```diff
@@ -32,7 +32,6 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
 - `getListItemCount`: gets the length of the items within the list
```
dlockhart, line 32: There's a Future Enhancement at the bottom of this page that talks about header that can now be removed. 🎉

components/list/list-header.js
```diff
@@ -0,0 +1,80 @@
...
+ * A header for list components containing select-all, etc.
+ */
+class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
+
+	static get properties() {
+		return {
+			/**
+			 * Whether to render a header with reduced whitespace.
+			 */
+			slim: { reflect: true, type: Boolean }
```
dlockhart, line 18: Should default to `false` in constructor.

components/selection/selection-action.js
```diff
@@ -0,0 +1,95 @@
...
+	static get properties() {
+		return {
+			/**
+			 * Preset icon key (e.g. "tier1:gear")
+			 */
+			icon: { type: String, reflect: true },
+			/**
+			 * Whether the action requires one or more selected items
+			 */
+			requiresSelection: { type: Boolean, attribute: 'requires-selection', reflect: true },
```
dlockhart, line 27: Should have a default value set to `false` in constructor.

components/selection/selection-action.js
```diff
@@ -0,0 +1,95 @@
...
+		const elem = this.shadowRoot.querySelector('d2l-button-subtle');
+		if (elem) elem.focus();
+	}
+
+	_handleActionClick(e) {
+		e.stopPropagation();
+
+		if (this.requiresSelection && this.selectionInfo.state === SelectionInfo.states.none) return;
+
+		this.dispatchEvent(new CustomEvent('d2l-selection-action-click', {
```
dlockhart, line 87: Could we just use the `click` event like subtle button does? I'm worried folks will accidentally wire up to `@click` anyway...

components/selection/selection-checkbox.js
```diff
@@ -0,0 +1,122 @@
...
+			}
+			:host([hidden]) {
+				display: none;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this._provider = null;
```
dlockhart, line 48: Ditto on the default property values here.

components/selection/selection-checkbox.js
```diff
@@ -0,0 +1,122 @@
...
+		if (!this.label || this.label.length === 0) console.warn('d2l-selection-checkbox component requires label text.');
+	}
+
+	render() {
+		return html`
+			<d2l-input-checkbox
+				aria-label="${this.label}"
+				@change="${this._handleChange}"
+				?checked="${this.selected}"
+				class="${ifDefined(this.hovering ? 'd2l-hovering' : undefined)}"
```
dlockhart, line 103: If you just applied the `checkboxStyles` to a native checkbox here I think you could constrain the "hovering" hackiness to just lists + selection-checkbox, but I'm not sure what other goodies you'd lose (like skeletons).

components/selection/selection-mixin.js
dlockhart: Update here?

components/selection/selection-mixin.js
```diff
@@ -0,0 +1,132 @@
...
+		};
+	}
+
+}
+
+export const SelectionMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			//selectionType: { type: String, attribute: 'selection-type' }
```
dlockhart, line 40: Remove until we know more?

components/selection/selection-summary.js
```diff
@@ -0,0 +1,42 @@
...
+				display: none;
+			}
+		`];
+	}
+
+	render() {
+		const summary = (this.selectionInfo.state === SelectionInfo.states.none && this.noSelectionText ?
+			this.noSelectionText : this.localize('components.selection.selected', 'count', this.selectionInfo.keys.length));
+		return html`
+			<div class="d2l-body-compact">${summary}</div>
```
dlockhart, line 36: Nit: semantically a `<p>` is likely more accurate

components/selection/test/selection.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+	it('dispatches d2l-selection-action-click event when clicked', async() => {
+		const el = await fixture(html`<d2l-selection-action></d2l-selection-action>`);
+		setTimeout(() => el.shadowRoot.querySelector('d2l-button-subtle').click());
+		await oneEvent(el, 'd2l-selection-action-click');
+	});
+
+	it('dispatches d2l-selection-action-click event when d2l-button-ghost-click is dispatched', async() => {
+		const el = await fixture(html`<d2l-selection-action></d2l-selection-action>`);
+		setTimeout(() => el.dispatchEvent(new CustomEvent('d2l-button-ghost-click')));
```
dlockhart, line 24: This is a overflow-group thing right?

components/inputs/input-checkbox.js
```diff
@@ -38,7 +40,8 @@ export const checkboxStyles = css`
 	}
 	input[type="checkbox"].d2l-input-checkbox:hover,
 	input[type="checkbox"].d2l-input-checkbox:focus,
-	input[type="checkbox"].d2l-input-checkbox.d2l-input-checkbox-focus {
+	input[type="checkbox"].d2l-input-checkbox.d2l-input-checkbox-focus,
+	:host(.d2l-hovering) input[type="checkbox"].d2l-input-checkbox {
```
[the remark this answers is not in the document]
dbatiste: Yeah I agree.  It took me a while to figure out how this was working before... since indeed the `:hover` selector was being used while hovering the item without hovering the actual checkbox.

components/list/demo/list-header.html
```diff
@@ -0,0 +1,147 @@
...
+			import '../list-header.js';
+			import '../list.js';
+			import '../../selection/selection-select-all.js';
+			import '../../selection/selection-action.js';
+		</script>
+	</head>
+	<body unresolved>
+
+		<d2l-demo-page page-title="d2l-list-header">
+
```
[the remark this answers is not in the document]
dbatiste: Yeah there is definitely some overlap between them.  Currently they are broken out by actions (selection, link items, button items, secondary actions), layouts (separator options, responsive breakpoints, slim, item-content, illustrations), header (all the stuff this PR adds), and then drag & drop is a whole other thing, and then just a general list demo page.  And of course, in some cases, actions affect layouts.  I'll see what I can do to reduce the duplication.

components/list/README.md
```diff
@@ -168,6 +169,7 @@ Where the parameters correspond to the slots of `d2l-list-item`:
 - `disabled` (Boolean): Whether or not the checkbox is disabled
 - `draggable` (Boolean): Whether or not the item is draggable
 - `key` (String): Value to identify item if selectable
+- `label` (String): The hidden label for the checkbox if selectable
 - `selectable` (Boolean): Indicates a checkbox should be rendered for selecting the item
 - `selected` (Boolean): Whether the item is selected
 - `slim` (Boolean): Whether to render the item with reduced whitespace
```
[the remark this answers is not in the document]
dbatiste: I will add something for this.  I kinda think this doc needs to be completely rewritten to make it easier for consumers to follow, and maybe that happens when we start setting up the readmes for the new doc stuff that @margaree and @GeorgePlukov are working on.  This might be a good example a complicated component to document.

components/selection/selection-mixin.js
dbatiste: Yeah... will remove this stuff and add back in the near future.

components/selection/test/selection.test.js
dbatiste: Yep, that's correct.

components/selection/selection-checkbox.js
dbatiste: That's true - I think it is just the skeleton that we'd lose, but it could be handled as well.  I'll fix the other stuff first.  😄  I don't feel super strongly about this one, although I don't want anyone else to use this.

components/list/README.md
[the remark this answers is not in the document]
dlockhart: Agreed yeah. I'm good if you want to wait for that refactor or do it separately, but it's definitely hard to follow as it is now.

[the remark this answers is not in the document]
dbatiste: Sounds good.  I added a section for the `d2l-list-header` since I'd forgotten that.

components/list/demo/list-header.html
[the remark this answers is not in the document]
dbatiste: I moved these demos into the `actions` demo page since that's where they currently fit best (i.e. there was already some selection demos there).

dlockhart (approved):

PR #1344, opened 2021-05-14
Update d2l-dropdown-button opener so that it can be made 100% width.

dbatiste:
This PR updates the width of the `d2l-button` inside of `d2l-dropdown-button` to be 100% width so that consumers can make their opener 100% by doing:
```html
<d2l-dropdown-button style="width: 100%;">
    ...
</d2l-dropdown-button>
```
This is apparently requested in the mobile FACE layout.


dbatiste:
Gonna add a VD test for this.

dlockhart (approved):
Seems reasonable to me!

PR #1348, opened 2021-05-17
Component Catalog: Draft of button and input-date-time documents

margaree:
This draft PR is more to check if these document changes seem on track. If they are generally correct (things will probably be tweaked along the way) then I'll close this PR and will work off of this branch to convert other core components to the same format and test in the component catalog.

Does the following:
- Updates wording of input-date/time* component jsdoc comments to be consistent
- Add custom-elements (temporarily for testing purposes in documentation repo) - this will be removed before a real PR
- Adds `/design` directory, which will contain corresponding design documentation. The pattern I created here was that in `button`, since that has a Best Practices section, the `<!-- docs: design -->` comment causes that design documentation to be put in the correct spot. Open to other ideas here. This could also change as time goes on (e.g., maybe there will be different design documentation that should go in other places)
- Updates button's README file to contain the content seen in Jeff's design for that page
- Also updates input-date-time.md to be able to be used in the catalog
- Catalog prep involves removing the properties, slots, and events tables since those would be generated from custom-elements.json in the catalog, and related to that involves ensuring the wording in the jsdoc comments is correct/consistent


design/buttons.md
```diff
@@ -0,0 +1,14 @@
...
+* Keep button text short - see "Writing" guidelines
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don't use multiple Primary buttons - the primary action should always be clear and obvious
+* Avoid icon-only buttons with unfamiliar icons, only use these for familiar concepts
+* Don't use buttons for navigation, use a link instead
+* Don't open menus with buttons - use a dropdown instead
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
```
dlockhart, line 14: Remind me again why we wanted the design tips stuff in a separate file vs. just inline? Seems like having this stuff visible in the repo is a nice feature of having it inline. Was it just so that we could set up separate designer owners?

margaree: It was, yeah. Then they can review any changes made and also only have to really be concerned with these files vs. a situation like always being pulled into changes on the readme.

dlockhart: Ok. I'm not sure if the pros outweigh the cons here then? The extra infrastructure to piece these back together again, the fact that these design tips won't be visible in the actual repo documentation vs. how often do we realistically expect this stuff to change and not only that but have a designer notice (after months or years of getting no notifications on it) and have their review catch a problem.

margaree: Agreed - especially since these are relatively small pieces I think it would be fine to keep them in the same file.

dlockhart (commented):
These changes look really good to me! All the redundant stuff (property tables, screenshots, "looking for an enhancement?") are gone which is amazing.

Seems in a good state to start applying elsewhere!

svanherk:
I'm kinda surprised to be losing the screenshots and gifs in the README.  They definitely don't make sense in the catalog with the live demos, and I'm not sure where we landed on if things can be easily opted out of (or if it's opt in).  Is the thinking that I'd need to go to the catalog anyways to see properties, events, etc, so I'm probably not using the actual README at all anymore?

margaree:
Yup, exactly - we are expecting the catalog to be more the source of truth and I think we'll be including a link to the applicable page in each document. It is easy to opt out so if we did want to leave in the screenshots for showing up in GitHub but not in the catalog that would be an option.

svanherk:
> Yup, exactly - we are expecting the catalog to be more the source of truth and I think we'll be including a link to the applicable page in each document. It is easy to opt out so if we did want to leave in the screenshots for showing up in GitHub but not in the catalog that would be an option.

Yeah that makes sense, screenshots can get out of date while the live demos won't, and it'll be nice not to have to make gifs.  I just hate when other component repos don't have an image in their README 😛 But probably as long as I can see it _somewhere_ without checking out the repo, that's enough

PR #1351, opened 2021-05-18
Check for null in focus-trap focus() method

emil-furniss:
Putting this up just in case this ends up being the appropriate fix for [this dialog issue](https://github.com/BrightspaceUI/core/pull/1330#issuecomment-843181215). Else just close it.

### DIFF
--- a/helpers/focus.js
+++ b/helpers/focus.js
@@ -1,4 +1,4 @@
-import { getComposedChildren, getComposedParent, getNextAncestorSibling, getPreviousAncestorSibling } from './dom.js';
+import { getComposedChildren, getComposedParent, getNextAncestorSibling, getOffsetParent, getPreviousAncestorSibling } from './dom.js';
 
 const focusableElements = {
 	a: true,
@@ -173,7 +173,7 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
 
 	if (_isFocusable && !includeHidden) {
 		// only perform visibility check if absolutely necessary
-		if (nodeName !== 'body' && node.offsetParent === null) return false;
+		if (nodeName !== 'body' && getOffsetParent(node) === null) return false;
 	}
 
 	return _isFocusable;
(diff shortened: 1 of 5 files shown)


dlockhart:
Thanks! I'll check out this branch and add some tests that hopefully expose the error and prevent it from happening again. Also want to have some better tests for that initial focus placement.

dlockhart:
The main issue here is dialogs that are opened as their initial state (i.e. `<d2l-dialog opened>`), and that initial focus happening before anything has rendered -- including the focus trap. Emil's fix prevents that error, and focus then later gets properly applied after rendering has completed. Wondering though whether that should be happening in `firstUpdated` instead of after a `rAF`. 🤷 

I've added some unit tests for dialog's focus management code, since the logic around which element should be initially getting focus is pretty clear. I've confirmed that the one test would have caught this.

dlockhart:
Looking into the Safari failures... I have a feeling initial focus handling may have never been working here and we just didn't notice.

helpers/focus.js
```diff
@@ -173,7 +173,7 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
 
 	if (_isFocusable && !includeHidden) {
 		// only perform visibility check if absolutely necessary
-		if (nodeName !== 'body' && node.offsetParent === null) return false;
+		if (nodeName !== 'body' && getOffsetParent(node) === null) return false;
```
dlockhart, line 176: Needed to use the helper here as in Safari, dialog button `offsetParent` was always `null`, so focus was never properly moving to them.

dbatiste (approved):
Looks good.  I think we've needed that node check in other places as well... I think it's become more common practice to check for it before trying to focus.

helpers/focus.js
dlockhart: And of course this helper doesn't return `false` for hidden elements so is entirely useless here. So need to figure out what's up with Safari in this particular case.

dlockhart:
Too many issues with the new dialog focus management tests in Firefox & Safari, so I'm going to skip them for now so we can get the fix to the JS error that's breaking FACE released. Then we can tackle those issues -- which aren't new AFAIK -- later.

PR #1353, opened 2021-05-19
[Fix] Overflow Group - Add aria labels to overflowing menu items

GeorgePlukov:
Screen readers were not properly calling out some menu items, this explicitly adds aria labels to the menu items and prioritizes the descriptions (like button/button subtle do).

Tested with NVDA and voiceover

### DIFF
--- a/components/overflow-group/overflow-group.js
+++ b/components/overflow-group/overflow-group.js
@@ -11,6 +11,7 @@ import '../menu/menu-item.js';
 import '../menu/menu-item-separator.js';
 import '../menu/menu-item-link.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
@@ -39,7 +40,8 @@ function createMenuItem(node) {
 	return html`<d2l-menu-item
 		?disabled=${disabled}
 		@d2l-menu-item-select=${handleItemSelect}
-		text="${childText}">
+		text="${childText}"
+		aria-label="${ifDefined(node.description || node.ariaLabel)}">
 	</d2l-menu-item>`;
 }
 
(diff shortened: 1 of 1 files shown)


components/overflow-group/overflow-group.js
```diff
@@ -39,7 +40,8 @@ function createMenuItem(node) {
 	return html`<d2l-menu-item
 		?disabled=${disabled}
 		@d2l-menu-item-select=${handleItemSelect}
-		text="${childText}">
+		text="${childText}"
+		aria-label="${ifDefined(node.description || node.ariaLabel)}">
```
dlockhart, line 44: So our menu-items don't explicitly list `aria-label` as a thing they support, but this likely still works since the `menuitem` role is getting applied to the `<d2l-menu-item>` directly.

I'm wondering... is the issue you're seeing limited to menus inside overflow groups, or menu-items more generally? If the latter, maybe menu-item itself should be reflecting its text to `aria-label` instead of requiring that consumer set it like this.

GeorgePlukov: This was specifically happening with menu-items inside overflow-groups (except for menu-items with sub menus for some reason). I was wondering myself if it was a wider spread issue with menu-items but they seem to be reading out the containing text properly still.

GeorgePlukov: Is it ideal to be setting the `aria-label` based off the `text `attribute? I'm imagining that would work by overriding the `aria-label` property in the `updated` function and set it based on the changed `text`. That approach seems like it would run into issues if we wanted to explicitly override the `ariaLabel` text for some reason unless we added a second flag to say we were overriding it

dlockhart: Like if you have `<div role="menuitem">text</div>` one would hope that AT would use "text" as the label -- which it seems like the do here most of the time. Using `<div role="menuitem" aria-label="text">text</div>` would be redundant, which is why it's weird that it fixes things in your case here.

Is there anything else weird about the menu DOM in overflow groups that differs from when menus are used normally? Maybe it's a timing things where the AT is reading through the virtual DOM before all our nodes are built and it's missing the label.

GeorgePlukov: The overflow group is rendering the dropdown later once it has determined its necessary, so it's definitely possible that it is being missed on a first pass.  Im not exactly sure how the different screen readers parse the page but it seems like adding the explicit `aria-labels` is causing them take a look at the property when the item is actually read

 I downloaded some aria dev tools extensions and it seems like even with the explicitly added aria labels the dev tools can't parse the labels properly (although screen readers all manage to read them). These dev tools also seem to have issues finding labels on the dropdown-button demo code which is very similar to the method overflow-groups uses to create its dropdown. menu-items within the dropdown-button demo code still get read by voiceover however

PR #1354, opened 2021-05-19
DE43639 Cert 21.06: Check for null in focus-trap focus() method

emil-furniss:
Cert for #1351


emil-furniss:
Closed in favor of #1355

PR #1355, opened 2021-05-19
DE43639 Cert 21.06: Check for null in focus-trap focus() method

emil-furniss:
Cert for #1351, porting only the actual fix and excluding the test and package-lock updates.


dlockhart (approved):
Looks good, thanks for cert fixing this!

PR #1357, opened 2021-05-19
Separate tooltip tests and don't reset focus before them

svanherk:
Trying to mitigate the effects of https://github.com/puppeteer/puppeteer/issues/7251 on this especially flakey test.


svanherk:
> Thanks for spending so much time investigating this Stacey. Hopefully this reduces the flake!

🤞 It's passed 5 times in a row now for me, but we'll have to see once it gets out into the wild.  Other flakes may also fit this criteria (if the component is responding to a resize event) and we could try fixing them this way.

PR #1358, opened 2021-05-21
Description and Aria-label property for menu-item

GeorgePlukov:
Adds a description property to `menu-item-mixin` to allow overriding of the `aria-label` text. This PR also makes it so that menu-items always have an `aria-label` property applied to them and don't rely on screen readers recognizing the text within. 

This also fixes an issue with `overflow-group` menu items being read as `menu item` by adding the menu-item description and defaulting it to the items text.


components/menu/menu-item-link.js
```diff
@@ -70,14 +70,6 @@ class MenuItemLink extends MenuItemMixin(LitElement) {
 		`;
 	}
 
-	updated(changedProperties) {
-		super.updated(changedProperties);
-		if (changedProperties.has('text')) {
-			// Voiceover + iOS can't find label inside <a>
-			this.setAttribute('aria-label', this.text);
```
dlockhart, line 77: Haha what a familiar change this is... made by me 2 months ago!

dlockhart (approved):
This looks good, and I now realize is just building off the link menu item stuff that I added back in March.

components/menu/menu-item-link.js
GeorgePlukov: Hahaha I didn't realize this was added so recently, it feels kinda nice to remove the setAttribute and just move this directly over to a property though!

components/menu/README.md
```diff
@@ -35,6 +35,7 @@ To make your usage of `d2l-menu` accessible, use the following property:
 | Attribute | Description |
 |--|--|
 | `label` | **REQUIRED** Acts as a primary label for the menu |
+| `description` | Overrides aria-label to provide extra information to screen reader users |
```
dbatiste, line 38: I think this belongs further down... in the menu-items section.

PR #1365, opened 2021-05-26
fix: labeling scroll-wrapper's buttons (fixes #1362)

### LINKED ISSUE #1362: d2l-table causes aria-hidden-focus accessibility violation
```
Rule: aria-hidden-focus
      Impact: serious
      ARIA hidden element must not contain focusable elements (https://dequeuniversity.com/rules/axe/4.2/aria-hidden-focus?application=axeAPI)
      
      Issue target: d2l-insights-assessment-quality-dashboard,d2l-table-wrapper,d2l-scroll-wrapper,.d2l-scroll-wrapper-button-left
      Context: <button class="d2l-scroll-wrapper-button d2l-scroll-wrapper-button-left" aria-hidden="true">     
                                        <d2l-icon icon="tier1:chevron-left"></d2l-icon>
                                </button>
      Fix all of the following:
        Focusable content should be disabled or be removed from the DOM

      Issue target: d2l-insights-assessment-quality-dashboard,d2l-table-wrapper,d2l-scroll-wrapper,.d2l-scroll-wrapper-button-right
      Context: <button class="d2l-scroll-wrapper-button d2l-scroll-wrapper-button-right" aria-hidden="true">    
                                        <d2l-icon icon="tier1:chevron-right"></d2l-icon>
                                </button>
      Fix all of the following:
        Focusable content should be disabled or be removed from the DOM
```

Steps to reproduce:
1) Run `d2l-insights-engagement-dashboard.accessibility` test in [this repo](https://github.com/Brightspace/insights-assessment-quality-dashboard). Comment out `aria-hidden-focus` inside `ignoredRules` block

### LINKED ISSUE #1363: d2l-table causes scrollable-region-focusable accessibility violation
```
 Rule: scrollable-region-focusable
      Impact: moderate
      Ensure that scrollable region has keyboard access (https://dequeuniversity.com/rules/axe/4.2/scrollable-region-focusable?application=axeAPI)

      Issue target: d2l-insights-assessment-quality-dashboard,d2l-table-wrapper,d2l-scroll-wrapper,.d2l-scroll-wrapper-container
      Context: <div class="d2l-scroll-wrapper-container"><slot></slot></div>
      Fix any of the following:
        Element should have focusable content
        Element should be focusable
```

Steps to reproduce:
1) Run `d2l-insights-engagement-dashboard.accessibility` test in [this repo](https://github.com/Brightspace/insights-assessment-quality-dashboard). Comment out `scrollable-region-focusable` inside `ignoredRules` block

dlockhart:
aXe was complaining that because we're trying to hide the scroll buttons to screen readers using `aria-hidden="true"` but still make them focusable to keyboard users.

Initial approach: ~I don't see a good way to accomplish both outcomes, so showing and labeling the buttons seems like the only option, even though screen reader users don't care about scrolling.~

Updated approach: I've made the scroll buttons clickable only and otherwise hidden from assistive technologies and keyboard users. For keyboard users I've made the scrollable area focusable and placed the visible outline there.

@Vitalii-Misechko I'm not sure if this will also fix #1363 as I couldn't reproduce it with the aXe tests. Let me know and if it's still happening we may need to also make the scrollable `<div>` focusable.


lang/ar.js
```diff
@@ -54,6 +54,8 @@ export default {
 	"components.more-less.less": "أقل",
 	"components.more-less.more": "المزيد",
 	"components.overflow-group.moreActions": "مزيد من الإجراءات",
+	"components.scroll-wrapper.scrollBack": "scroll back",
+	"components.scroll-wrapper.scrollForward": "scroll forward",
```
dlockhart, line 58: Tried to make these RTL-agnostic by going with "back" and "forward" instead of "scroll left" and "scroll right"

Vitalii-Misechko:
> @Vitalii-Misechko I'm not sure if this will also fix #1363 as I couldn't reproduce it with the aXe tests. Let me know and if it's still happening we may need to also make the scrollable `<div>` focusable.

Sure thing. Thanks

Vitalii-Misechko (approved):

dlockhart:
Updated with a new approach: the buttons are now just for mouse users, similar to how the Polymer scroll wrapper worked. This will hide them completely from AT. For keyboard users, the scrollable area is focusable with a visible outline.

Vitalii-Misechko (approved):

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -162,16 +152,16 @@ class ScrollWrapper extends FocusVisiblePolyfillMixin(RtlMixin(LitElement)) {
...
+				</div>
+				<div class="d2l-scroll-wrapper-button d2l-scroll-wrapper-button-right" @click="${this._scrollRight}">
 					<d2l-icon icon="tier1:chevron-right"></d2l-icon>
-				</button>
+				</div>
 			</div>` : null;
 		return html`
 			${actions}
-			<div class="d2l-scroll-wrapper-container" @scroll="${this._checkScrollThresholds}"><slot></slot></div>
+			<div class="d2l-scroll-wrapper-container" @scroll="${this._checkScrollThresholds}" tabindex="0"><slot></slot></div>
```
dbatiste, line 164: I feel like this is a dumb question, but why does the scroll-wrapper need to be focusable?

dlockhart: So that keyboard users have a way to scroll it left/right. The old version of this they simply couldn't scroll it at all. I changed it so that the buttons were focusable but then aXe started complaining that aria-hidden things can't be focusable (the issue Vitalii raised). So just making the scrollable area focusable is the compromise I've arrived at.

dbatiste: Ah I see.  So the slot that's within the scroll wrapper could contain other focusable elements, which would mean that we would have focusable elements inside of this focusable element?  I think that might be ok, although I've sometimes run into problems with that sorta thing in the past.  Most recently with nested menus - that scenario is more complicated though because there is a lot of focus management code that tries to redirect focus if focus events fire on focusable parents when clicking on focusable children.  That's just one example of weirdness.  Long story short... I think we should tread carefully.

dbatiste (approved):

components/scroll-wrapper/scroll-wrapper.js
dlockhart: Ok thanks I'll be on the lookout for issues!

PR #1369, opened 2021-05-28
Created focus helper to traverse up the tree to find focusable ancestor 

annabelsegalld2l:
Traversing up the tree will allow us to restore focus to the first visible element when the original opening element is hidden, as in the case of a dropdown. Currently, if the original opener is hidden, the focus is lost 😢

### DIFF
--- a/components/dialog/demo/dialog-fullscreen.html
+++ b/components/dialog/demo/dialog-fullscreen.html
@@ -14,6 +14,11 @@
 			import '../../button/button.js';
 			import '../dialog-fullscreen.js';
 			import './dialog-async-content.js';
+			import '../../dropdown/dropdown.js';
+			import '../../dropdown/dropdown-more.js';
+			import '../../dropdown/dropdown-menu.js';
+			import '../../menu/menu.js';
+			import '../../menu/menu-item.js';
 		</script>
 		<style>
 			.d2l-typography p:first-child { margin-top: 0; }
@@ -29,6 +34,13 @@
 			<d2l-demo-snippet>
 				<template>
 					<d2l-button id="open">Show Dialog</d2l-button>
+					<d2l-dropdown-more text="Open!">
+						<d2l-dropdown-menu>
+							<d2l-menu label="Dialog">
+								<d2l-menu-item id="dropdown" text="See dialog"></d2l-menu-item>
+							</d2l-menu>
+						</d2l-dropdown-menu>
+					</d2l-dropdown-more>
 					<d2l-dialog-fullscreen id="dialogFullscreen" title-text="Fullscreen Title">
 						<p>Deadlights jack lad schooner scallywag dance the hempen jig carouser broadside cable strike colors. Bring a spring upon her cable holystone blow the man down spanker</p>
 						<p>Shiver me timbers to go on account lookout wherry doubloon chase. Belay yo-ho-ho keelhaul squiffy black spot yardarm spyglass sheet transom heave to.</p>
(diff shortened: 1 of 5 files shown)


helpers/focus.js
```diff
@@ -179,3 +179,12 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
 	return _isFocusable;
 
 }
+
+export function tryApplyFocus(elem) {
+	const focusable = findComposedAncestor(elem, (node) => (isFocusable(node) || getFirstFocusableDescendant(node) !== null));
```
annabelsegalld2l, line 184: In the case of dropdowns, the open button is not an ancestor of the menu items. So going up the tree, the button would never be found to focus on. Adding the check to see if any descendants are focusable allows the button to be focused on.

components/dialog/demo/dialog-fullscreen.html
```diff
@@ -29,6 +34,13 @@ <h2>Fullscreen Dialog</h2>
 			<d2l-demo-snippet>
 				<template>
 					<d2l-button id="open">Show Dialog</d2l-button>
+					<d2l-dropdown-more text="Open!">
```
annabelsegalld2l, line 37: Just added this bit to make testing easier. Not sure if we want to keep it.

helpers/focus.js
dlockhart: If `elem` is itself focusable, wouldn't this go up the ancestor tree first and potentially find a focusable sibling? Just wondering if we should first check that `elem` itself isn't focusable before resorting to an ancestor search.

helpers/focus.js
```diff
@@ -179,3 +179,12 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
...
+	if (focusable) {
+		forceFocusVisible(focusable);
```
dlockhart, line 186: I _think_ we always want to force visible focus here -- at least we do from the dialog case... so maybe this is fine. If not, we could always make it an option in the future.

annabelsegalld2l: I believe that focusable can never be null here. If we traverse up the DOM, we eventually hit the body, which is focusable (but doesn't help much). Effectively, we will force visible focus in every scenario, and return false only if something very weird happens...

helpers/focus.js
dbatiste: Right.  The actual opener that gets the focus is not a direct ancestor eh?  The `getFirstFocusableDescendant` I think isn't exactly cheap and is being called recursively, but it's not the end of the work as long at `tryApplyFocus` is only being called once.

annabelsegalld2l: tryApplyFocus is only called once, correct. Just when we close the dialog to find its opener to focus on. I double-checked that buttons should be caught by checking the element itself, so getFirstFocusableDescendant won't be run in those cases.

dbatiste (approved):
I think we'll want to add this helper to the readme - just s simple one-liner explanation for it like the others.  Also a couple of simple unit tests I think would be prudent.

dlockhart (approved):

helpers/test/focus.test.js
```diff
@@ -358,4 +376,32 @@ describe('focus', () => {
...
 	});
 
+	describe('tryApplyFocus', () => {
+
+		it('returns true on single focusable element', async() => {
+			const elem = await fixture(focusableFixture);
+			expect(tryApplyFocus(elem.querySelector('button'))).to.be.true;
+		});
+
+		it('returns false on non-existent element', async() => {
```
annabelsegalld2l, line 386: I think this the only test we can do to produce a false result. Since we keep checking the ancestors, we eventually bubble up to the focusable body in any non-null case.

helpers/test/focus.test.js
```diff
@@ -358,4 +376,32 @@ describe('focus', () => {
...
+			const elem = await fixture(focusableFixture);
+			expect(tryApplyFocus(elem.querySelector('#nonExistentDiv'))).to.be.false;
+		});
+
+		it('returns true on element with focusable parent', async() => {
+			const elem = await fixture(focusableAncestorFixture);
```
dlockhart, line 392: Nit: one thing you could add to these tests is to not only check for the `true`/`false` return value, but also ensure that it's actually focusing on the element you expect. To do that, you could give them `id`s and ensure that `elem.id` matches what you're expecting.

PR #1370, opened 2021-05-28
Move definition of mediaquerylist inside constructor of dialog

annabelsegalld2l:
When the dialog is used by other classes, the mediaQueryList definition does not get passed with it. Moving the definition inside the constructor will ensure that it will always be defined. 
Example of the current error: https://github.com/Brightspace/d2l-my-courses-ui/pull/892/checks?check_run_id=2695755713#step:8:42


dbatiste:
Oh this is just because of how the test fixture works?

PR #1371, opened 2021-05-28
Only add event listener if method exists (prevents issues on safari)

annabelsegalld2l:
Safari 13 does not have the addEventListener method, specifically it can not listen for the 'change' event. An alternative is to use addListener, but it is a deprecated API. Instead, if the browser does not support addEventListener, no listener will be added at all. The typography will stay the same from when it was loaded (ie. a mobile page will stay mobile-styled).


dlockhart (approved):
Perfect! I like not using the old API just to support post-load changes in Safari < 14. 👍

svanherk (approved):

PR #1373, opened 2021-05-28
fix: use IntersectionObserver to apply classes

dlockhart:
Now that extra 👀  are on tables, it was noticed that in a few places in the LMS when the code to apply the first/last row and cell CSS classes runs, the rows aren't visible yet so nothing gets applied. This results in missing borders and no rounded corners:

image (2)

This actually isn't a new problem -- I went back to the May release and the old Polymer tables had the same issue. In this instance it's a timing issue -- the table is in our legacy tabs and there's an HTML editor on the page and things just take a while to load up. But in theory the table could be legitimately hidden (maybe in a dialog that isn't open yet or an inactive tab) and that would result in the same broken behaviour.

The fix is to use an `IntersectionObserver` to detect when the table enters the viewport for the first time. ~This has the added benefit of deferring the paint/reflows to when the table is actually about to be rendered.~ Nope, that causes an initial jitter as you scroll them into view and broke the visual diff tests.

🎩 -tip to @Dan-DeAraujo for finding this!

### DIFF
--- a/components/table/table-wrapper.js
+++ b/components/table/table-wrapper.js
@@ -261,7 +261,9 @@ export class TableWrapper extends RtlMixin(LitElement) {
 		}
 	}
 
-	async _applyClassNames(table) {
+	async _applyClassNames(table, count) {
+
+		if (count === 10) return;
 
 		await new Promise((resolve) => requestAnimationFrame(resolve));
 
@@ -274,6 +276,10 @@ export class TableWrapper extends RtlMixin(LitElement) {
 			firstRow = firstRow || r;
 			lastRow = r;
 		});
+		if (rows.length > 0 && firstRow === null) {
+			setTimeout(() => this._applyClassNames(table, ++count), 300);
+			return;
+		}
 
 		const topHeader = table.querySelector('tr.d2l-table-header:first-child th:not([rowspan]), tr[header]:first-child th:not([rowspan]), thead tr:first-child th:not([rowspan])');
 		const topHeaderHeight = topHeader ? topHeader.clientHeight : -1;
@@ -328,14 +334,14 @@ export class TableWrapper extends RtlMixin(LitElement) {
 
 		// observes mutations to <table>'s direct children and also
 		// its subtree (rows or cells added/removed to any descendant)
-		this._tableObserver = new MutationObserver(() => this._applyClassNames(table));
+		this._tableObserver = new MutationObserver(() => this._applyClassNames(table, 0));
 		this._tableObserver.observe(table, {
 			attributes: true, /* required for legacy-Edge, otherwise attributeFilter throws a syntax error */
 			attributeFilter: ['selected'],
 			childList: true,
 			subtree: true
 		});
-		this._applyClassNames(table);
+		this._applyClassNames(table, 0);
 
 	}
 
--- a/components/table/table-wrapper.js
+++ b/components/table/table-wrapper.js
@@ -233,12 +233,14 @@ export class TableWrapper extends RtlMixin(LitElement) {
 		this.noColumnBorder = false;
 		this.stickyHeaders = false;
 		this.type = 'default';
-		this._tableObserver = null;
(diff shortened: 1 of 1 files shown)


components/table/table-wrapper.js
```diff
@@ -263,8 +265,6 @@ export class TableWrapper extends RtlMixin(LitElement) {
 
 	async _applyClassNames(table) {
 
-		await new Promise((resolve) => requestAnimationFrame(resolve));
```
dlockhart, line 266: Don't need this delay anymore as the intersection observer handles it.

components/table/table-wrapper.js
```diff
@@ -328,14 +328,34 @@ export class TableWrapper extends RtlMixin(LitElement) {
...
+							requestIdleCallback(() => this._applyClassNames(table));
+						}
+					});
+				});
+			} else {
+				this._tableIntersectionObserver.disconnect();
+			}
+			this._tableIntersectionObserver.observe(table, { root: document.body });
+		} else {
+			requestAnimationFrame(() => this._applyClassNames(table));
```
dlockhart, line 357: Fallback for older browsers to what it was doing before.

components/table/table-wrapper.js
```diff
@@ -328,14 +328,33 @@ export class TableWrapper extends RtlMixin(LitElement) {
...
+							requestIdleCallback(() => this._applyClassNames(table));
+						}
+					});
+				});
+			} else {
+				this._tableIntersectionObserver.disconnect();
+			}
+			this._tableIntersectionObserver.observe(table);
+		}
+		requestAnimationFrame(() => this._applyClassNames(table));
```
dbatiste, line 357: Is this rAF in case `IntersectionObserver` is not supported?  I did a quick test and it seems that the observer gets called initially, not just after page-load. I seem to recall some flakiness with Safari when using it to detect `display` changes a year or two ago... unable to reproduce now though so maybe they fixed some stuff.

components/table/table-wrapper.js
```diff
@@ -233,12 +233,14 @@ export class TableWrapper extends RtlMixin(LitElement) {
...
 		this.type = 'default';
-		this._tableObserver = null;
+		this._tableIntersectionObserver = null;
+		this._tableMutationObserver = null;
 	}
 
 	disconnectedCallback() {
 		super.disconnectedCallback();
-		if (this._tableObserver) this._tableObserver.disconnect();
+		if (this._tableIntersectionObserver) this._tableIntersectionObserver.disconnect();
```
dbatiste, line 242: There maybe an edge case here that would not work - if this component is disconnected then the observers would be disconnected here, but I don't _think_ they would be reconnected if this component is reconnected to the DOM.  I suspect the `slotchange` event handler would not be rerun on reconnect to the DOM.  In order for this to be a problem, it would have to get disconnected, reconnected, and the table would also have to be hidden.  I am wondering if we even need to disconnect the observers here since the observer and the table being observed are all in this elements subtree.  I certainly feel better if we do though.

dbatiste (approved):
Just the couple of thoughts there.  Otherwise looks good!

components/table/table-wrapper.js
dlockhart: I originally only had this in the `else` branch -- so it would only run if `IntersectionObserver` wasn't supported. But that lead to it not running right away if a table was outside the viewport... which initially seems OK but was causing most of the visual diff tests to fail (since they're long pages and not in view) but also some initially jitter when you scroll them into view.

components/table/table-wrapper.js
dlockhart: Yeah agreed... we're definitely inconsistent and maybe a bit over-zealous with this stuff generally. I did some testing with this and you're correct... if I move the `d2l-table-wrapper` DOM node from one place in the document to somewhere else, the observers get disconnected and not re-connected. So I've removed this code that disconnects them altogether. That way when things get re-connected they run again and everything seems to work nicely.

PR #1375, opened 2021-05-28
US122202 - Handle single dimension case

svanherk:
When we have just one dimension, we don't need to render the menu - opening the dropdown will show that dimension.
image

Design: https://xd.adobe.com/view/819a26ff-a0ba-4f8e-b405-ac677da84804-d4c9/screen/b66ee85a-4d6b-42b5-8344-c1bfd3aefe17

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -91,7 +91,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		const dimensions = this._buildDimensions(singleDimension);
 
 		const buttonText = singleDimension ? this._dimensions[0].text : this.localize('components.filter.filters');
-		const description = singleDimension ? this.localize('components.filter.singleDimensionDescription', 'filterName', this._dimensions[0].text) : this.localize('components.filter.filters');
+		const description = singleDimension ? this.localize('components.filter.singleDimensionDescription', { filterName: this._dimensions[0].text }) : this.localize('components.filter.filters');
 
 		const dropdownContent = singleDimension ? html`
 				<d2l-dropdown-content min-width="285" max-width="420" no-padding-header no-padding>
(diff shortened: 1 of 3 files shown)


components/filter/filter.js
```diff
@@ -78,34 +86,52 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 
-		return html`
-			<d2l-dropdown-button-subtle
-				@d2l-dropdown-close="${this._handleDropdownClose}"
-				@d2l-dropdown-open="${this._stopPropagation}"
-				@d2l-dropdown-position="${this._stopPropagation}"
-				text="${this.localize('components.filter.filters')}"
-				?disabled="${this.disabled}">
+		const buttonText = singleDimension ? this._dimensions[0].text : this.localize('components.filter.filters');
+		const description = singleDimension ? this.localize('components.filter.singleDimensionDescription', 'filterName', this._dimensions[0].text) : this.localize('components.filter.filters');
```
svanherk, line 94: These will eventually also incorporate the counts

components/filter/filter.js
```diff
@@ -125,12 +154,13 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		});
 	}
 
-	_buildHeader() {
```
svanherk, line 128: Most of the changes in this function are setting things up for future changes, where there will be a clear button, search bar, etc.

components/inputs/input-checkbox.js
```diff
@@ -19,6 +19,7 @@ export const checkboxStyles = css`
 		display: inline-block;
 		height: 1.2rem;
 		margin: 0;
+		outline: none;
```
svanherk, line 22: Noticed this when testing on Firefox:
image

components/filter/filter.js
dlockhart: This way is still supported because of Polymer, but I think we prefer the `{ filterName: this._dimensions[0].text }` way.

dlockhart (approved):

PR #1380, opened 2021-05-31
Tooltip position improvements

dbatiste:
This PR enhances the tooltip positioning logic to respect the boundaries of containers that do not allow overflow.


templates/primary-secondary/primary-secondary.js
```diff
@@ -571,7 +571,6 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
 				-webkit-overflow-scrolling: touch;
 				overflow-x: hidden;
 				overflow-y: scroll;
-				position: relative;
```
dbatiste, line 574: This is unnecessary and causes all sorts of trouble with tooltip positioning.  It was added to zupport using z-index [here](https://github.com/BrightspaceUI/core/commit/63cacb0e6a5408969d487e77e2dee62defd1d0ff#diff-06cfe552d3f9e7117999c057f42fd8b028a379c2a88a4004b6cc8ab40b63285fR568), and was later removed [here](https://github.com/BrightspaceUI/core/commit/9953f658b02ea91dbde5e4040d95b49fb014dab3#diff-06cfe552d3f9e7117999c057f42fd8b028a379c2a88a4004b6cc8ab40b63285fL568), but the `position: relative` was left behind by accident.

components/tooltip/tooltip.js
```diff
@@ -578,12 +579,23 @@ class Tooltip extends RtlMixin(LitElement) {
 	}
 
 	_computeSpaceAround(offsetParent, targetRect) {
-		const spaceAround = {
+
+		const boundingContainer = getBoundingAncestor(this);
+		const bounded = (boundingContainer !== document.documentElement);
+		const boundingContainerRect = boundingContainer.getBoundingClientRect();
+
+		const spaceAround = (bounded ? {
```
dbatiste, line 587: This is almost the same as dropdown, except the treatment of margins is different.

dlockhart (approved):
Seems good!

PR #1382, opened 2021-06-01
Add disabled tooltip to buttons

dbatiste:
* adds property to buttons for optional tooltip in disabled state
* updates default tooltip delay

### DIFF
--- a/components/button/README.md
+++ b/components/button/README.md
@@ -19,6 +19,7 @@ The `d2l-button` element can be used just like the native button element, but al
 |--|--|--|
 | `description` | String | A description to be added to the `button` for accessibility |
 | `disabled` | Boolean | Disables the button |
+| `disabledTooltip` | String | Tooltip text when disabled |
 | `primary` | Boolean | Styles the button as a primary button |
 
 **Accessibility:**
@@ -53,6 +54,7 @@ The `d2l-button-subtle` element can be used just like the native `button`, but f
 | `text` | String, required | Text for the button |
 | `description` | String | A description to be added to the `button` for accessibility for additional context |
 | `disabled` | Boolean | Disables the button |
+| `disabledTooltip` | String | Tooltip text when disabled |
 | `h-align` | String | Possible values are undefined (default) or `text`. If `text`, aligns the button content to the leading edge of text. |
 | `icon` | String | [Preset icon key](../icons#preset-icons) (e.g. `tier1:gear`) |
 | `icon-right` | Boolean | Render the icon on the right of the button |
@@ -87,6 +89,7 @@ The `d2l-button-icon` element can be used just like the native `button`, for ins
 | `icon` | String, required | [Preset icon key](../icons#preset-icons) (e.g. `tier1:gear`) |
 | `text` | String, required | Accessible text for the buton |
 | `disabled` | Boolean | Disables the button |
+| `disabledTooltip` | String | Tooltip text when disabled |
 | `h-align` | String | Possible values are undefined (default) or `text`. If `text`, aligns the button content to the leading edge of text. |
 | `translucent` | Boolean | Indicates to display translucent (ex. on rich backgrounds) |
 
--- a/components/button/button.js
+++ b/components/button/button.js
@@ -102,7 +102,7 @@ class Button extends ButtonMixin(LitElement) {
 				type="${this._getType()}">
 				<slot></slot>
 			</button>
-			${this.disabled && this.disabledTooltip ? html`<d2l-tooltip for="${this._buttonId}" delay="200">${this.disabledTooltip}</d2l-tooltip>` : ''}
+			${this.disabled && this.disabledTooltip ? html`<d2l-tooltip for="${this._buttonId}">${this.disabledTooltip}</d2l-tooltip>` : ''}
 		`;
 	}
 }
--- a/components/button/button.js
+++ b/components/button/button.js
@@ -85,6 +85,7 @@ class Button extends ButtonMixin(LitElement) {
 	render() {
 		return html`
 			<button
+				aria-disabled="${ifDefined(this.disabled && this.disabledTooltip ? 'true' : undefined)}"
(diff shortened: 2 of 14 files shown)


components/button/README.md
```diff
@@ -19,6 +19,7 @@ The `d2l-button` element can be used just like the native button element, but al
 |--|--|--|
 | `description` | String | A description to be added to the `button` for accessibility |
 | `disabled` | Boolean | Disables the button |
+| `disabledTooltip` | String | Tooltip text when disabled |
```
dbatiste, line 22: If anyone has preference, we could alternatively call this `disabledText`.  Jeff and I talked briefly about both options but leaned towards `disabledTooltip` since it makes it clearer what it's for, even though it does tie it to tooltip.

components/tooltip/tooltip.js
```diff
@@ -366,7 +366,7 @@ class Tooltip extends RtlMixin(LitElement) {
 
 		this.announced = false;
 		this.closeOnClick = false;
-		this.delay = 0;
+		this.delay = 300;
```
dbatiste, line 369: Jeff and I discussed this a fair amount and he ultimately decided we should update the default delay for all tooltips to make them less jarring.  There are a couple of places that provide their own delay... most usages rely on the default.

components/button/button.js
```diff
@@ -83,17 +90,19 @@ class Button extends ButtonMixin(LitElement) {
 				aria-label="${ifDefined(this.description || this.ariaLabel)}"
```
dlockhart, line 90: So you decided to not apply `aria-disabled`?

components/button/button.js
```diff
@@ -83,17 +90,19 @@ class Button extends ButtonMixin(LitElement) {
...
 				?autofocus="${this.autofocus}"
 				class="d2l-label-text"
-				?disabled="${this.disabled}"
+				?disabled="${this.disabled && !this.disabledTooltip}"
```
dlockhart, line 93: Should we be wrapping the `click` event here and preventing bubble / preventing default if we're disabled?

components/button/README.md
dlockhart: I'm OK with this. Normally I don't love tying components together like this, but "tooltip" is also a semantic thing so I'm good with it. If we were to add normal tooltips to buttons (which we wanted to do), I think we'd call it `tooltip` so that tracks.

components/button/button.js
dbatiste: The mixin already stops the propagation of the click if `disabled`, so I _think_ that's handled for us.  I will add a unit test for this specific case.

dlockhart: Ah ok, nice! I'm a little surprised TBH because the native `<button>` would have been doing that already when it's disabled. Hmm.

components/button/button.js
dbatiste: Just following up with Jeff about that.  This is what gets announced in VO and NVDA though.  Note the addition of "dimmed" in the VO case, and "unavailable" in the NVDA case.

VO: "[button text], dimmed, button." [tooltip text] You are currently on a button.  This item is dimmed.
NVDA: "clickable [button text] button unavailable [tooltip text]"

dlockhart: How does that compare to:
1. A normal disabled button
2. The exact same setup but without `aria-disabled`

dbatiste: 1. A normal disabled button - nothing to compare to, since it's not possible to focus on a disabled button.

2. Same setup without `aria-disabled`

> VO: [button text], button. [tooltip text] You are currently on a button. To click this button, press Control-Option-Space.
NVDA: clickable [button text] button [tooltip text]

So yeah, without the `aria-disabled="true"`, VO keeps announcing "To click this button, press Control-Option-Space." like it would for an enabled button, which is definitely misleading.

I spoke with Jeff about the dimmed/unavailable this as well, and unless you disagree, I think we're all in favour of using `aria-disabled="true"`.

components/tooltip/tooltip.js
```diff
@@ -24,6 +24,17 @@ const contentBorderRadius = 6;
 const contentBorderSize = 1;
 const contentHorizontalPadding = 15;
 
+/* once a user shows a tooltip, ignore delay if they hover adjacent target within this timeout */
```
dbatiste, line 27: This is some usability extra sugar that Jeff wanted added to tooltips.  The basic idea here is - once a tooltip has been displayed for a target, if the user hovers over another target within 1s of showing the first tooltip, then we should ignore the delay and just show the tooltip immediately.

components/button/button.js
dlockhart: > nothing to compare to, since it's not possible to focus on a disabled button

You must be able to access it though using arrow keys or by pulling up a list of buttons/form inputs? It's not invisible... just not focusable by tabbing through the page.

But yeah, I'm 100% in agreement that we want this -- I can't think of a single reason not to.

dbatiste: To summarize:

Virtual cursor, normal disabled button:

>VO: [button text], dimmed, button. You are currently on a button. This item is dimmed.
>NVDA: [button text] selected

Virtual cursor, `aria-disabled` with tooltip:

>VO: [button text], dimmed, button. [tooltip text] You are currently on a button. This item is dimmed.
>NVDA: [button text] selected

Focusing, normal disabled button: N/A. 

Focusing, `aria-disabled` with tooltip:

>VO: [button text], dimmed, button. [tooltip text] You are currently on a button. This item is dimmed.
>NVDA: clickable [button text] button unavailable [tooltip text]

Focusing, regular button:

>VO: [button text], button. [tooltip text] You are currently on a button. To click this button, press Control-Option-Space.
>NVDA: clickable [button text] button [tooltip text]

Conclusion: use `aria-disabled="true"` for our simulated disabled state.  Of interest, NVDA only announces the tooltip text when focusing - it does not announce it when using the virtual cursor.

dlockhart: Amazing -- agreed!

dlockhart (approved):

PR #1390, opened 2021-06-04
Experimenting with a reactive controller approach

dlockhart:
Here's what things might look like if we used Lit 2's "reactive controller" approach for the subscriber only.

I'm following their docs here which talks about mixins being appropriate for "is a" situations (like the list "is a" selectable collection) and controllers should be used for "has a" situations like `selection-summary` "has a" selection info.

I've also switched this to not use events during disconnection since the nodes are removed before `disconnectedCallback` is called and the events were going nowhere. More comments inline.


components/overflow-group/overflow-group.js
```diff
@@ -32,7 +32,7 @@ const OPENER_STYLE = {
 
 function createMenuItem(node) {
 	const childText = node.text || node.firstChild && (node.firstChild.label || node.firstChild.text || node.firstChild.textContent.trim());
-	const disabled = node.disabled || node.nonInteractive;
+	const disabled = node.disabled;
```
dlockhart, line 35: I wasn't really sure what was going on here and why overflow-group cared about `nonInteractive` (or why the action had a non-interactive attribute). Maybe I'm missing something though.

components/selection/selection-action.js
```diff
@@ -7,24 +7,20 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
...
 	static get properties() {
 		return {
 			/**
 			 * Preset icon key (e.g. "tier1:gear")
 			 */
 			icon: { type: String, reflect: true },
-			/**
-			 * Indicates whether the action is only focusable
-			 */
-			nonInteractive: { type: Boolean, attribute: 'non-interactive', reflect: true },
```
dlockhart, line 27: Again, wasn't quite sure why this needed to be a public thing that was set as opposed to just a computed internal thing based on `disabled` + `requires-selection`.

components/selection/selection-action.js
```diff
@@ -47,55 +43,47 @@ class Action extends LocalizeCoreElement(SelectionSubscriberMixin(ButtonMixin(Rt
...
-	get selectionInfo() {
-		return super.selectionInfo;
-	}
-
-	set selectionInfo(value) {
-		super.selectionInfo = value;
-		this.nonInteractive = (this.requiresSelection && this.selectionInfo.state === SelectionInfo.states.none);
+	constructor() {
+		super();
+		this._selectionInfo = new SelectionInfoController(this);
```
dlockhart, line 48: This is how controllers work -- you `new` them in the constructor and pass them a reference to the host element and any additional options.

components/selection/selection-action.js
```diff
@@ -47,55 +43,47 @@ class Action extends LocalizeCoreElement(SelectionSubscriberMixin(ButtonMixin(Rt
...
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
 		this.addEventListener('d2l-button-ghost-click', this._handleActionClick);
+		this._selectionInfo.hostConnected(); // remove with Lit 2
```
dlockhart, line 54: Need to manually do this, but with Lit 2 all the lifecycle callbacks will be handled for us.

components/selection/selection-action.js
```diff
@@ -47,55 +43,47 @@ class Action extends LocalizeCoreElement(SelectionSubscriberMixin(ButtonMixin(Rt
...
 	}
 
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		this.removeEventListener('d2l-button-ghost-click', this._handleActionClick);
+		this._selectionInfo.hostDisconnected(); // remove with Lit 2
 	}
 
 	render() {
+		const nonInteractive = (this.requiresSelection && this._selectionInfo.state === SelectionInfo.states.none);
```
dlockhart, line 64: Moved this out to a computed value since it didn't seem to need to be a public property?

components/selection/selection-checkbox.js
```diff
@@ -62,18 +67,19 @@ class Checkbox extends SkeletonMixin(LitElement) {
...
 		super.connectedCallback();
-		this.dispatchEvent(new CustomEvent('d2l-selection-checkbox-subscribe', {
+		const evt = new CustomEvent('d2l-selection-checkbox-subscribe', {
 			bubbles: true,
-			composed: true
-		}));
+			composed: true,
+			detail: {}
+		});
+		this.dispatchEvent(evt);
```
dlockhart, line 75: The parent selection collection (the list or table that has the mixin) will populate `detail.provider`, which we store so that we can later tell it to unsubscribe us when we're disconnected.

components/selection/selection-info-controller.js
```diff
@@ -0,0 +1,51 @@
...
+		this._provider = null;
+		this._state = 'none';
+	}
+
+	get keys() {
+		return this._keys;
+	}
+	set keys(value) {
+		this._keys = value;
+		this._host.requestUpdate();
```
dlockhart, line 22: Setting either of these will cause the host to re-render.

components/selection/selection-checkbox.js
dlockhart: We could actually use a different controller for the selection checkboxes and select-all checkbox... maybe a "selection modifier controller"? It could have methods like `selectAll()`, `unselectAll()` and `select(key), `unselect(key)`. Under the covers it would fire these events, but that would make it a bit cleaner to add more UI bits that modify selection down the road.

components/selection/selection-info-controller.js
```diff
@@ -0,0 +1,51 @@
+export class SelectionInfoController {
+
+	static get properties() {
+		return {
+			keys: { type: Object },
+			state: { type: String }
+		};
+	}
```
dlockhart, line 8: I'm not sure if these actually need to be defined as properties like this. All the Lit docs use TypeScript 😠

components/overflow-group/overflow-group.js
dbatiste: So yeah, maybe `nonInteractive` isn't needed on the action as-is.  The `nonInteractive` property wasn't meant to be set by a consumer, but rather internally depending on the selection state.  The `disabled` property is one that consumers usually expect to able to set, so I felt that wasn't quite right.  I was on the fence as to whether the property used should be public or private.  Maybe we can use `disabled` and provide a getter but no setter.

The `d2l-non-interactive-change` event was being used to toggle the `disabled` state of the corresponding menu-item inside the overflow menu.  This is something that `d2l-overflow-group` doesn't handle so I think we'll need to deal with that one way or another.  

The event is convenient but limited.  An alternative could be to wire-up `MutationObserver`s to the slotted buttons/actions.  However, `MutationObserver`s do not detect composed subtree changes - they only detect subtree changes within the context of the target.  I think to make that work, it would be necessary to wire-up the observers directly to the assigned (and flattened) nodes of the slot, watching for `disabled` attribute changes, and do this in conjunction with `slotchange`.  That might be a good sol'n that would also address this current gap with `d2l-overflow-group` and `disabled` state of buttons.

components/selection/selection-action.js
dbatiste: I think this action component needs to expose this computed state, so that `d2l-overflow-group` can properly reflect the state in it's corresponding overflow menu-items.

components/overflow-group/overflow-group.js
dlockhart: Yeah I was thinking about this more over the weekend and landed on the `MutationObserver` approach as well. There are other things about the buttons that could change (like the text) that in theory overflow-group should be handling.

components/selection/selection-action.js
dbatiste: Hmmm... no I remember why I chose to use a new property - `disabled` is defined and provided by the `ButtonMixin` as a public property, and I wanted a property that did not conflict with that because the consumer could specify the button to be `disabled`, and we would not want to unset that when consumers select items.

dlockhart:
Closing this as we're sticking with the non-controller approach for now.

PR #1394, opened 2021-06-08
Add LabelMixin and LabelledMixin.

dbatiste:
This PR adds two mixins, `LabelMixin` and `LabelledMixin` that enables elements within custom element's shadowDOM to be labelled by elements outside that shadowDOM.  For instance, it enables the following scenarios as showing in the demo page:

```html
<table>
	<tr>
		<td><d2l-demo-input labelled-by="topping-1"></d2l-demo-input></td>
		<td><span id="topping-1" class="d2l-label-text">Caramel</span></td>
	</tr>
	<tr>
		<td><d2l-demo-input label="Sprinkles"></d2l-demo-input></td>
		<td><span class="d2l-label-text">Sprinkles</span></td>
	</tr>
	<tr>
		<td><d2l-demo-input labelled-by="topping-3"></d2l-demo-input></td>
		<td><d2l-demo-text id="topping-3" text="Hard Chocolate Coating"></d2l-demo-text></td>
	</tr>
</table>
```

The idea is that a labelled custom element containing some element that requires a label, extends the `LabelledMixin` which locates the labelling element to resolve the label.  It further wires up a `MutationObserver` to detect label changes.  If the labelling element is a custom element, then it is required to extend the `LabelMixin`, and call its `updateLabel()` method as necessary - this in turn reflects the label to a private `_label` property that is observed by `LabelledMixin`, which then applies the label to the element requiring the label.

@zinabat : this PR is a follow-up from our previous conversation.  I am interested in your thoughts/confirmation that this would be usable for [d2l-activity-name](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/components/activity/name/d2l-activity-name.js).  I think that it would need to extend `LabelMixin` and call `updateLabel()` as described.  However `d2l-activity-name` wraps `d2l-hc-name`, so somehow `d2l-activity-name` would need to know when to call `updateLabel`, and with what value. The alternative we discussed is that 
[d2l-activity-editor-collection-add](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/master/components/activity/editor/collection/d2l-activity-editor-collection-add.js#L140) would need to be updated to render a custom list item instead, so that a `label` could be provided for the list-items.

### DIFF
--- a/mixins/demo/labelled-mixin.html
+++ b/mixins/demo/labelled-mixin.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="../../components/demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '../../components/demo/demo-page.js';
+			import '../../components/inputs/input-checkbox.js';
+			import { css, html, LitElement } from 'lit-element/lit-element.js';
+			import { LabelledMixin, LabelMixin } from '../labelled-mixin.js';
+			import { ifDefined } from 'lit-html/directives/if-defined.js';
+
+			class DemoInput extends LabelledMixin(LitElement) {
+				static get styles() {
+					return css`
+						:host {
+							disply: inline-block;
+						}
+						d2l-input-checkbox {
+							margin-bottom: 0;
+						}
+					`;
+				}
+				render() {
+					return html`
+						<d2l-input-checkbox aria-label="${ifDefined(this.label)}"></d2l-input-checkbox>
+					`;
+				}
+			}
+			customElements.define('d2l-demo-input', DemoInput);
+
+			class DemoText extends LabelMixin(LitElement) {
+				static get properties() {
+					return {
+						text: { type: String }
+					};
+				}
+				static get styles() {
+					return css`
+						:host {
+							disply: inline-block;
+						}
+					`;
+				}
+				constructor() {
+					super();
+					this.text = '';
+				}
+				render() {
+					return html`
+						<span>${this.text}</span>
+					`;
+				}
+				updated(changedProperties) {
+					super.updated(changedProperties);
+					this._label = this.text;
+				}
+			}
+			customElements.define('d2l-demo-text', DemoText);
+
+		</script>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+	</head>
+	<body unresolved>
+		<d2l-demo-page page-title="LabelledMixin">
+
+			<h2>Default</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<table>
+						<tr>
+							<td><d2l-demo-input labelled-by="topping-1"></d2l-demo-input></td>
+							<td><span id="topping-1" class="d2l-label-text">Caramel</span></td>
+						</tr>
+						<tr>
+							<td><d2l-demo-input label="Sprinkles"></d2l-demo-input></td>
+							<td><span class="d2l-label-text">Sprinkles</span></td>
+						</tr>
+						<tr>
+							<td><d2l-demo-input labelled-by="topping-3"></d2l-demo-input></td>
+							<td><d2l-demo-text id="topping-3" class="d2l-label-text" text="Hard Chocolate Coating"></d2l-demo-text></td>
+						</tr>
+					</table>
+				</template>
+			</d2l-demo-snippet>
+
+		</d2l-demo-page>
+	</body>
+</html>
--- a/mixins/demo/labelled-mixin.html
+++ b/mixins/demo/labelled-mixin.html
@@ -64,7 +64,7 @@
 		<meta charset="UTF-8">
 	</head>
 	<body unresolved>
-		<d2l-demo-page page-title="LabelledMixin">
+		<d2l-demo-page page-title="LabelMixin & LabelledMixin">
 
 			<h2>Default</h2>
 
--- a/mixins/demo/labelled-mixin.html
+++ b/mixins/demo/labelled-mixin.html
@@ -53,7 +53,7 @@
 				}
 				updated(changedProperties) {
 					super.updated(changedProperties);
-					this._label = this.text;
+					this.updateLabel(this.text);
 				}
 			}
 			customElements.define('d2l-demo-text', DemoText);
--- a/mixins/labelled-mixin.js
+++ b/mixins/labelled-mixin.js
@@ -1,10 +1,40 @@
 
+const getCommonAncestor = (elem1, elem2) => {
+
+	const labelledPath = new WeakMap();
+	let elem = elem1;
+	while (elem) {
+		labelledPath.set(elem, elem);
+		elem = elem.parentNode;
+	}
+
+	let ancestorElem = elem2.parentNode;
+	while (ancestorElem) {
+		if (labelledPath.has(ancestorElem)) return ancestorElem;
+		ancestorElem = ancestorElem.parentNode;
+	}
+
+};
+
+const isCustomElement = elem => {
+	return elem.tagName.includes('-');
+};
+
+const getLabel = labelElem => {
+	if (isCustomElement(labelElem)) return labelElem._label;
+	else return labelElem.textContent;
+};
+
 export const LabelMixin = superclass => class extends superclass {
 
 	static get properties() {
 		return {
 			_label: { type: String, reflect: true }
 		};
+	}
+
+	updateLabel(text) {
+		this._label = text;
 	}
 
 };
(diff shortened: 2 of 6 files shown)


mixins/labelled-mixin.js
```diff
@@ -0,0 +1,87 @@
...
+
+				if (labelElem) {
+					this.label = getLabel();
+				} else {
+					console.warn('LabelledMixin with specified labelledBy but no such element exists.');
+					this.label = undefined;
+				}
+
+			});
+			this._labelObserver.observe(ancestor, { attributes: true, characterData: true, childList: true, subtree: true });
```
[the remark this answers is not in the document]
dlockhart: Do we know at this point if we could use one mechanism or the other? i.e. do we know here that we're dealing with a web component vs. not? If so, then it might be worth the performance boost to have a different config for each of those two cases.

Also, could `attributeFilter: ['_label']` be used to narrow that down? I think legacy-Edge still needs `attributes: true` [as I discovered here](https://github.com/BrightspaceUI/core/blob/1a9b1ed4818432f54c5e62846e83d8db32c9e262/components/table/table-wrapper.js#L327), but at least we could comment that and then remove it in a month.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,87 @@
...
+
+			this.label = getLabel();
+		}
+
+	}
+
+	_getCommonAncestor(labelledByElem) {
```
dlockhart, line 70: Nit: could pull this out of the mixin class (if `this` was passed in).

mixins/demo/labelled-mixin.html
```diff
@@ -0,0 +1,92 @@
...
+					this.text = '';
+				}
+				render() {
+					return html`
+						<span>${this.text}</span>
+					`;
+				}
+				updated(changedProperties) {
+					super.updated(changedProperties);
+					this._label = this.text;
```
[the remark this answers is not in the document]
dlockhart: Could this be a method like `updateLabel(this.text)` that under the covers updates the `_label` property? Just wondering if that could save us down the road if we want to change the approach without updating 100 places.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const LabelledMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			labelledBy: { type: String, reflect: true, attribute: 'labelled-by' },
+			label: { type: String }
+		};
+	}
```
dlockhart, line 19: Wondering: if we had a setter/getter for `label` (that set a private `_label`), would that give us more info about whether they explicitly set the label vs. if we set it programatically based on `labelled-by`? That could let us have one always take precedence over the other. i.e. if they've set `label` explicitly maybe we just ignore `labelled-by`.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,87 @@
...
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (!changedProperties.has('labelledBy')) return;
+
+		if (this._labelObserver) this._labelObserver.disconnect();
+
+		if (this.labelledBy) {
```
dlockhart, line 28: Nit: could invert this I think and just `return` to reduce nesting.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,87 @@
...
+							break;
+						}
+					}
+
+					if (!labelElem && mutations[i].addedNodes.length > 0) {
+						labelElem = this.getRootNode().querySelector(`#${this.labelledBy}`);
+						break;
+					}
+
+				}
```
dlockhart, line 53: Not sure if this works, but `forEach` and an `indexOf` might simplify the looping logic a bit.

```suggestion
				mutations.forEach((m) => {

					if (m.removedNodes.indexOf(labelElem)) {
						labelElem = null;
						return;
					}

					if (!labelElem && m.addedNodes.length > 0) {
						labelElem = this.getRootNode().querySelector(`#${this.labelledBy}`);
						break;
					}

				});
```

dlockhart (commented):
I really like this... very simple and yet powerful!

mixins/labelled-mixin.js
dbatiste: Do you mean that we would have a separate private `_label` property?  The consumer needs to rerender itself when the label changes, and currently that rerender is triggered by changing the `label` property.  Having that property would keep things separate and allow us to make the consumer-specified `label` take precedence, but I'm not sure how just a setter for `label` would achieve that.  But yeah, currently if they were to specify both `labelledBy` and `label`, the `labelledBy` would take precedence, which I think would be ok since it would be strange to specify both.

mixins/labelled-mixin.js
dbatiste: I'll include this in my changes... need a couple modifications since `NodeList` does not support `indexOf`.  Would not want the `return` statement, and the `break` statement should become `return`.

mixins/labelled-mixin.js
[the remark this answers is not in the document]
dbatiste: Yeah that's a good point.  I will update.

mixins/labelled-mixin.js
dlockhart: Yeah nevermind... thinking about it more I think it just adds more complexity than it's worth.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,108 @@
...
+					labelElem = this.getRootNode().querySelector(`#${this.labelledBy}`);
+					return;
+				}
+
+			});
+
+			if (labelElem) {
+				this.label = getLabel(labelElem);
+			} else {
+				console.warn('LabelledMixin with specified labelledBy but no such element exists.');
```
dlockhart, line 81: Nit: might be useful to include the `labelled-by` value here to give folks a hint as to what's not working.

Something like:

```suggestion
				console.warn(`LabelledMixin: element with labelled-by="${this.labelledBy}", but no such element exists.`);
```

mixins/test/label-mixin.test.js
```diff
@@ -0,0 +1,162 @@
...
+			`;
+		}
+		updated(changedProperties) {
+			super.updated(changedProperties);
+			this.updateLabel(this.text);
+		}
+	}
+);
+
+describe.only('LabelMixin', () => {
```
dlockhart, line 35:
```suggestion
describe('LabelMixin', () => {
```

mixins/test/label-mixin.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+	it('reflects label value change', async() => {
+		elem.text = 'new label value';
+		await nextFrame();
+		expect(elem.getAttribute('_label')).to.equal('new label value');
+	});
+
+});
+
+describe.only('LabelledMixin', () => {
```
dlockhart, line 57: Finally me catching this and not causing it! 🥇 

```suggestion
describe('LabelledMixin', () => {
```

mixins/test/label-mixin.test.js
```diff
@@ -0,0 +1,162 @@
...
+		`);
+	});
+
+	it('reflects label value', async() => {
+		expect(elem.getAttribute('_label')).to.equal('the label value');
+	});
+
+	it('reflects label value change', async() => {
+		elem.text = 'new label value';
+		await nextFrame();
```
dlockhart, line 51: What's the difference between `nextFrame` and `await elem.updateComplete`? I think I usually do the latter.

mixins/test/label-mixin.test.js
```diff
@@ -0,0 +1,162 @@
...
+					<tr>
+						<td><${labelledTag} labelled-by="label2"></${labelledTag}></td>
+						<td><${labelTag} id="label2" text="custom elemnt"></${labelTag}></td>
+					</tr>
+				</table>
+				<span id="label3">other element</span>
+			</div>
+		`);
+	});
+
```
dlockhart, line 82: For all three of these cases, it might be good to test some of the negative cases -- like `id`s referencing elements that don't exist, or elements that are not in the same DOM scope. Just to make sure that your error handling/logging code isn't blowing up.

mixins/test/label-mixin.test.js
dbatiste: So at least for these tests, waiting for `updateComplete` is not sufficient for the browser to have applied the updates, which is necessary to assert the attribute values.  If these tests are changed to assert the property values (instead of the attribute values) then waiting for `updateComplete` is sufficient.  In some of these tests, I think checking the attribute value is preferred due to the way the mixins depend on attribute changes.  

I think this is due to the Lit timing for property changes and when those changes are reflected to the DOM, and the `nextFrame` helper uses rAF (task) so it's waiting longer than just `updateComplete` (microtask I think).

dlockhart: Gotcha makes sense thx.

mixins/labelled-mixin.js
```diff
@@ -0,0 +1,108 @@
...
+
+		this._labelObserver = new MutationObserver(mutations => {
+
+			mutations.forEach(mutation => {
+
+				if (mutation.removedNodes.length > 0 && Array.from(mutation.removedNodes).indexOf(labelElem) !== -1) {
+					labelElem = null;
+				}
+
+				if (mutation.addedNodes.length > 0) {
```
dbatiste, line 71: Previously I was using this optimization... `if (!labelElem) && mutation.addedNodes.length > 0` which was causing problems with legacy Edge where labels are not getting updated correctly.  This was because legacy Edge triggers two mutation records when a node is replaced, and we are notified of the added record before the removed record.

dlockhart (approved):

PR #1404, opened 2021-06-16
US123512 - Add selected counts to the opener and menu items of filter

svanherk:
This adds the logic to keep track of how many selected filter values we have, at a dimension level to display on the menu items, and at a top level to display on the dropdown opener.  Lots of the details around what to say when can be found in the design: https://xd.adobe.com/view/819a26ff-a0ba-4f8e-b405-ac677da84804-d4c9/screen/b66ee85a-4d6b-42b5-8344-c1bfd3aefe17/
image
image

The blue status indicator badge for the menu items (shown in the design) does not exist yet - still waiting on more design details.  For now, I'm just showing a plain text number and offscreen text that will eventually be swapped for the official component.


components/filter/filter.js
```diff
@@ -150,6 +163,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			const builtDimension = this._buildDimension(dimension);
 			return html`<d2l-menu-item text="${dimension.text}">
 				${builtDimension}
+				<div slot="supporting">
+					<span>${this._formatFilterCount(dimension.appliedCount, dimension.maxCount)}</span>
+					<d2l-offscreen>${this.localize('components.filter.filterCountDescription', { number: dimension.appliedCount })}</d2l-offscreen>
+				</div>
```
svanherk, line 169: This will eventually be swapped out for the new status indicator component, which I expect will take `text` and some type of `aria-label`/`description` for screenreaders.  Wanted to get the terminology of these fleshed out now.

components/filter/filter.js
```diff
@@ -60,6 +63,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				white-space: nowrap;
 			}
 
+			.d2l-filter-dimension-set-value-text {
+				line-height: unset;
+			}
+
```
svanherk, line 69: This is unrelated - I noticed the checkboxes weren't centering properly with the text since applying the `d2l-body-compact` styles last PR.

components/filter/filter.js
```diff
@@ -90,8 +99,12 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		const header = this._buildHeader(singleDimension);
 		const dimensions = this._buildDimensions(singleDimension);
 
-		const buttonText = singleDimension ? this._dimensions[0].text : this.localize('components.filter.filters');
-		const description = singleDimension ? this.localize('components.filter.singleDimensionDescription', { filterName: this._dimensions[0].text }) : this.localize('components.filter.filters');
+		const filterCount = this._formatFilterCount(this._totalAppliedCount, this._totalMaxCount);
+		let buttonText = singleDimension ? this._dimensions[0].text : this.localize('components.filter.filters');
```
dbatiste, line 103: Nit: seems like we can avoid some string concatenation.  I think it could be avoided in a single statement as a `const`, but also in two statements if readability is desired.  I think the two statement version would be something like:
```javascript
let buttonText = singleDimension ? this._dimensions[0].text : this.localize('components.filter.filters');
if (filterCount) buttonText = `${buttonText} (${filterCount})`;
```

components/filter/filter.js
```diff
@@ -150,6 +163,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			const builtDimension = this._buildDimension(dimension);
 			return html`<d2l-menu-item text="${dimension.text}">
 				${builtDimension}
+				<div slot="supporting">
+					<span>${this._formatFilterCount(dimension.appliedCount, dimension.maxCount)}</span>
+					<d2l-offscreen>${this.localize('components.filter.filterCountDescription', { number: dimension.appliedCount })}</d2l-offscreen>
```
dlockhart, line 168: I don't feel strongly about this, but there's also the `offscreenStyles` that you can use with the `d2l-offscreen` CSS class. In general it's probably better to use fewer web components/elements if possible.

dlockhart (approved):

components/filter/filter.js
svanherk: I'm actually gonna move this to the description of the menu-item for now, so that it definitely gets read out.  NVDA is ignoring this.  Not sure how this will work once the status-indicator is in - if it also will be ignored and we'll need to keep the accessible text on the menu-item anyways, or if it will be done in a way that the screenreader sees it.

PR #1408, opened 2021-06-16
fix: importing rIC shim for Safari

dlockhart:
Just noticed errors in Safari as it doesn't support `requestIdleCallback`. It doesn't seem to break anything in most scenarios as the class names get applied by the mutation observer.


dbatiste (approved):
Maybe one day Safari will finally release this from behind its flag!

PR #1409, opened 2021-06-16
fix: importing rIC shim for Safari

dlockhart:
Cert version of #1408.


dlockhart:
I had to pick up the Node version change and associated golden updates as well. I just cherry-picked Margaree's changes from `master`.

dbatiste (approved):
Looks good.  A lot of those diffs remind me of a diff that @grant-cleary and I have been seeing in the HtmlEditor - where there is just one pixel somewhere on the border that comes up different.  It always seems to be the same pixel, but not for every run.

dlockhart:
These diffs are from the Puppeteer upgrade that happened in the visual-diff action automatically, but because this branch is cert and back in time it doesn't have the changes from `master` that Margaree merged yesterday.

PR #1414, opened 2021-06-17
US129213: Feat: Add mobile-tray left and right styling 

annabelsegalld2l:
Design spec: https://xd.adobe.com/view/5205ec02-86e7-48b0-ad0a-04aa76db7759-0756/screen/eb611fa3-3250-40f2-8cfc-4d11464911f7/?fullscreen

This PR adds the mobile-tray options of left and right, only appearing when the mobile breakpoint hits. 
If the mobile-tray option is not specified, the dropdown should have the same look on mobile as it does with a wider screen.

GIFs: 
dropdowns_gif
dropdowns_responsive


components/dropdown/dropdown-content-mixin.js
```diff
@@ -706,6 +720,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 					</div>
 				</div>
 			</div>
+			<d2l-backdrop for-target="d2l-dropdown-wrapper" ?shown="${specialMobileStyle}"></d2l-backdrop>
```
annabelsegalld2l, line 723: The target can only be an id, so I needed to add the "d2l-dropdown-wrapper' id to the width class, which defines the height and width of the dropdown.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -47,6 +51,14 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 				type: Number,
 				attribute: 'max-height'
 			},
+			/**
+			 * Override default mobile dropdown style. Specify one of 'tray' or 'dialog'.
+			 */
+			mobileTray: {
```
dbatiste, line 57: Looks like we need to update the comment here -i.e. looks like these are not the accepted values.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -164,6 +180,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 		this.noAutoClose = false;
 		this.noAutoFit = false;
 		this.noAutoFocus = false;
+		this.mobileTray = 'none';
```
dbatiste, line 183: Is it possible to avoid sprouting an attribute by letting this remain `undefined` in the default case?

components/dropdown/dropdown-content-mixin.js
dbatiste: Is `bottom` going to be implemented as well?  I noticed some empty CSS selector further below.

components/dropdown/dropdown-content-styles.js
```diff
@@ -197,4 +225,104 @@ export const dropdownContentStyles = css`
...
+	@-webkit-keyframes d2l-dropdown-mobile-tray-right {
+		0% { opacity: 0; -webkit-transform: translateX(100%); }
+		100% { opacity: 1; -webkit-transform: translate(0, 0); }
+	}
+
+	@keyframes d2l-dropdown-mobile-tray-right-close {
+		0% { opacity: 1; transform: translate(0, 0); }
+		100% { opacity: 0; transform: translateX(100%); }
+	}
+	@keyframes d2l-dropdown-mobile-tray-right-close {
```
annabelsegalld2l, line 242: This second one should be webkit-keyframes for cross-browser compatibility.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -244,7 +275,18 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 	}
 
 	close() {
-		this.opened = false;
+		const hide = () => {
+			this._closing = false;
+			this.shadowRoot.querySelector('.d2l-dropdown-content-width').removeEventListener('animationend', hide);
```
dbatiste, line 280: According to [caniuse](https://caniuse.com/?search=once), it should be possible to avoid the explicit removal if the `once` parameter is passed when adding.

```javascript
this.shadowRoot.querySelector('.d2l-dropdown-content-width')
    .addEventListener('animationend', hide, { once: true });
```

However, Mozilla seems to indicate that was added in Edge 18, and it might be hard to find Edge 17.

components/dropdown/dropdown-content-mixin.js
annabelsegalld2l: Yes, that's a TODO for now, The first story is just the left/right styles since they are very similar.

components/dropdown/dropdown-content-mixin.js
annabelsegalld2l: Looks like the once parameter is being used in other spots in the lms like https://search.d2l.dev/xref/BrightspaceUI/htmleditor/htmleditor.js?r=fd11e9a4#297, so it seems safe to use. I can make that change.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -662,25 +705,64 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
 		};
 
 		const contentStyle = {
 			...contentWidthStyle,
-			maxHeight: this._contentHeight ? `${this._contentHeight}px` : 'none',
+			maxHeight: this._contentHeight && !specialMobileStyle ? `${this._contentHeight}px` : 'none',
 			overflowY: this._contentOverflow ? 'auto' : 'hidden'
 		};
 
+		const closeButtonVisibility = {
```
dbatiste, line 760: Nit: I think `closeButtonStyles` name better reflects it's purpose.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -694,18 +776,24 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
+						<slot name="footer" @slotchange="${this.__handleFooterSlotChange}">
+						</slot>
+						<d2l-button
+						style=${styleMap(closeButtonVisibility)} 
+						@click=${this.close}>
+						${this.localize('components.dropdown.close')}</d2l-button>
 					</div>
 				</div>
 			</div>
+			<d2l-backdrop for-target="d2l-dropdown-wrapper" ?shown="${specialMobileStyle && this.opened && !this._closing}" ></d2l-backdrop>
```
dbatiste, line 796: I'd be interested in avoiding rendering `d2l-backdrop` in cases where it would not be used... i.e. when there is no `specialMobileStyle`.  So, it would become something like:

```javascript
const dropdown = html`
   <div class="d2l-dropdown-content-position" style=${styleMap(positionStyle)}>
      ...
   </div>
`;

if (specialMobileStyle) return html`${dropdown} ${html`<d2l-backdrop ...></d2l-backdrop>`}`;
else return dropdown;
```

components/dropdown/dropdown-content-styles.js
```diff
@@ -30,17 +35,40 @@ export const dropdownContentStyles = css`
 		opacity: 0.9;
 	}
 
-	:host([opened]) {
-		-webkit-animation: var(--d2l-dropdown-animation-name) 300ms ease;
-		animation: var(--d2l-dropdown-animation-name) 300ms ease;
-		display: inline-block;
+	@media (min-width: 616px) {
+		:host([opened]) {
+			-webkit-animation: var(--d2l-dropdown-animation-name) 300ms ease;
```
dbatiste, line 40: I _think_ the `-webkit-animation` property was only needed for Safari 8 and earlier.  I think some of the styles in here are quite old, and are probably being copied in a multitude of places.  Same goes for `-webkit-keyframes` and `-webkit-transform`. It might be nice to avoid them.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -26,6 +26,7 @@ describe('d2l-dropdown', () => {
 	beforeEach(async() => {
 		dropdown = await fixture(normalFixture);
 		content = dropdown.querySelector('d2l-dropdown-content');
+		await content.updateComplete;
```
annabelsegalld2l, line 29: There are some tests that reference the dropdown-content that need to have the await updateComplete line added. This is because using the LocalizeCoreElementMixin causes longer render times for the component. I had many flaky tests and race conditions, as the content was not rendered before the tests went to query it. The updateComplete will ensure that the content is fully loaded each time.

components/dropdown/test/dropdown-content.visual-diff.html
```diff
@@ -397,6 +397,62 @@ <h1>Header Larger Than Max</h1>
 			</d2l-dropdown-content>
 		</d2l-dropdown>
 
+		<d2l-dropdown id="mobile-right-tray">
```
annabelsegalld2l, line 400: Added some tests for mobile screens. Looks like all the tests previously were using a 400*800 screen.

components/filter/test/filter.axe.js
```diff
@@ -32,6 +32,8 @@ describe('d2l-filter', () => {
 		it(`${test.name} opened`, async() => {
 			const elem = await fixture(test.fixture);
 			const dropdown = elem.shadowRoot.querySelector('d2l-dropdown-button-subtle');
+			const dropdownContent = elem.shadowRoot.querySelector('d2l-dropdown-content') || elem.shadowRoot.querySelector('d2l-dropdown-menu');
```
annabelsegalld2l, line 35: Single and multiple dimension filters use content and menu, respectively. Waiting for the update of the content here, like other tests.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -244,7 +276,20 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 	}
 
 	close() {
-		this.opened = false;
+		const hide = () => {
+			this._closing = false;
```
annabelsegalld2l, line 280: I've added a state called _closing to allow the fade-out animation to happen. Previously, the dropdowns just disappeared without transitions, so this state allows the dropdown to be "open" but still have a closing animation.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -662,25 +710,64 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			}
 		}
 
+		const specialMobileStyle = mediaQueryList.matches && this.mobileTray;
+		const mobileTrayRightLeft = mediaQueryList.matches && (this.mobileTray === 'right' || this.mobileTray === 'left');
```
annabelsegalld2l, line 714: Lots of overrides happening here. Luckily I was able to use the same logic for left and right trays. For left and right trays;
-minimum width must be 285px 
-max width is 420px OR the total width - 30px for the backdrop
-width of the dropdown is set to the maximum width unless previously let by some other logic (but it must be within mix-max-widths.

components/dropdown/dropdown-content-styles.js
```diff
@@ -2,9 +2,14 @@ import '../colors/colors.js';
 import { css } from 'lit-element/lit-element.js';
 
 export const dropdownContentStyles = css`
+
 	:host {
 		--d2l-dropdown-above-animation-name: d2l-dropdown-above-animation;
 		--d2l-dropdown-animation-name: d2l-dropdown-animation;
+		--d2l-dropdown-mobile-tray-right-name: d2l-dropdown-mobile-tray-right;
```
dbatiste, line 9: I think I must be missing something - wondering why these are variables instead of just specifying these values where they are used...?

components/dropdown/dropdown-content-styles.js
```diff
@@ -30,17 +35,30 @@ export const dropdownContentStyles = css`
 		opacity: 0.9;
 	}
 
-	:host([opened]) {
-		-webkit-animation: var(--d2l-dropdown-animation-name) 300ms ease;
```
dbatiste, line 34: It looks like there is some other `-webkit-` stuff in here.  Although not required for your changes, I think it would be good to clean that up since we are excluding them for the new styles being added.

components/dropdown/dropdown-content-styles.js
annabelsegalld2l: They are so you can override the animation depending on other css variables, ie. override the default dropdown animation for dark themes. You are right, I can remove these.

dbatiste (approved):
Awesome!  Looks great.

components/dropdown/dropdown-content-styles.js
dbatiste: Cool yeah... I'd only expose stuff as variables that can be overridden if needed.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -129,6 +151,10 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			_bottomOverflow: {
 				type: Boolean
 			},
+			_closing: {
```
dlockhart, line 154: I think this can actually have `attribute: false` and be called `_closing` since you're not actually using the attribute on the dropdown -- it's being bound to a `<div>` which actually has the attribute.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -244,7 +276,20 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
+			this._closing = false;
+			this._showBackdrop = false;
+			this.opened = false;
+		};
+
+		if (!reduceMotion && mediaQueryList.matches && this.mobileTray && isVisible(this)) {
+			this.shadowRoot.querySelector('.d2l-dropdown-content-width')
+				.addEventListener('animationend', hide, { once: true });
+			this._closing = true;
+			this._showBackdrop = false;
```
dlockhart, line 289: `hide()` also sets `_showBackDrop` to `false` -- is it still needed here? I guess it depends whether you want the backdrop to go away before/while it's hiding.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -348,7 +396,7 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			return;
 		}
 
-		this.opened = false;
+		this.close();
```
dlockhart, line 399: Does this mean that consumers need to call `open()`/`close()` instead of setting `opened = true` or `opened = false`?

components/dropdown/dropdown-content-mixin.js
```diff
@@ -692,9 +779,9 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
 			'd2l-dropdown-content-footer': this._hasFooter
 		};
 
-		return html`
+		const dropdown =  html`
 			<div class="d2l-dropdown-content-position" style=${styleMap(positionStyle)}>
-				<div class="d2l-dropdown-content-width" style=${styleMap(widthStyle)}>
+				<div  id="d2l-dropdown-wrapper" class="d2l-dropdown-content-width" style=${styleMap(widthStyle)} ?closing="${this._closing}">
```
dlockhart, line 784: Technically we shouldn't be putting non-standard attributes like `closing` on native elements like a `div`. I'd recommend using something like `?data-closing="${this._closing}"`.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -703,10 +790,22 @@ export const DropdownContentMixin = superclass => class extends RtlMixin(supercl
...
 				</div>
 			</div>
 		`;
+		if (this.mobileTray) return html`
+			${dropdown} 
+			<d2l-backdrop 
+				for-target="d2l-dropdown-wrapper" 
+				?shown="${this._showBackdrop}" >
+			</d2l-backdrop>`;
+		else return html`${dropdown}`;
```
dlockhart, line 808: Nit: totally the same outcome, but you could use a conditional `return` here.

```suggestion
		return this.mobileTray ? html`
			${dropdown} 
			<d2l-backdrop 
				for-target="d2l-dropdown-wrapper" 
				?shown="${this._showBackdrop}" >
			</d2l-backdrop>` : html`${dropdown}`;
```

components/dropdown/dropdown-content-mixin.js
annabelsegalld2l: Yep, we actually need the backdrop,shown to be false, not just remove it from the DOM. The reasoning behind this has to do with the backdrop preventing background scroll when opened. On backdrop.shown = true, the scrolling is locked, and only shown = false can release it (not just deleting from the DOM). I spent a bunch of time debugging this as the screen seemed to 'lock' in place after using the dropdown, because you do manually need to set the backdrop to go away.

dbatiste: We might be able to add a `disconnectedCallback()` to `d2l-backdrop` to do some cleanup for that scenario.  It get strange if the backdrop instance is removed from the DOM and then re-added however.  I think this was originally purpose built for dialog and then refactored, and this scenario was probably avoided in the interest if scope, and not wanting to deal with those scenarios.

PR #1417, opened 2021-06-18
Add search-input default placeholder

svanherk:
Adding `Search...` as the default placeholder in the `search-input` if none is set.  Grabbed the langterms from the many, many places setting this as their placeholder (could try to clean these up as we see them).  I suspect serge will fix any up as needed.


components/inputs/input-search.js
```diff
@@ -142,7 +142,7 @@ class InputSearch extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					@input="${this._handleInput}"
 					@keypress="${this._handleInputKeyPress}"
 					maxlength="${ifDefined(this.maxlength)}"
-					placeholder="${ifDefined(this.placeholder)}"
+					placeholder="${this.placeholder || this.localize('components.input-search.defaultPlaceholder')}"
```
svanherk, line 145: If I tried to set this as a default in the `constructor`, the localize stuff wasn't ready yet.  This seemed like a clean way around that, thoughts?

dlockhart: Yeah looks good!

dlockhart (approved):

PR #1419, opened 2021-06-18
fix: Make sure table classnames are applied every time the table becomes visible

jorge-ramirez-arredondo:
Currently, if a table is hidden when mutating it, the class names are only applied after it becomes visible for the first time. If the table is hidden and made visible again after the first time, the class names won't re-apply properly.

This PR changes it so the class names are re-applied every time the table becomes visible after being hidden.

Additionally, I added some logic to make sure the mutation observer and the intersection observer are properly disconnected when relevant.

### DIFF
--- a/components/table/table-wrapper.js
+++ b/components/table/table-wrapper.js
@@ -238,6 +238,13 @@ export class TableWrapper extends RtlMixin(LitElement) {
 		this._tableMutationObserver = null;
 	}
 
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
This looks really good to me and matches our discussions. Thanks so much for getting to the root of this and putting together a fix!

Issue #1398, opened 2021-06-14
Dropdown Auto Close Not Working When Wrapped In Another Component


svanherk:
@JamieTran is seeing this in Consistent Eval with the labs filter and in FACE with a normal dropdown.  They both use primary-secondary template but he was able to reproduce this outside of that, and I created a simple reproduction inside the dropdown demo: https://github.com/BrightspaceUI/core/compare/demonstrate_broken_auto_close

![broken auto close](https://user-images.githubusercontent.com/13419300/121941081-d74f2a00-cd1c-11eb-98da-4f75ea9cc56f.gif)

It seems like https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-content-mixin.js#L354 isn't triggered when expected.

svanherk:
I'm gonna delete the branch above, which should be fine because:
1. This should be fixed 🎉 
2. @margaree added a visual-diff test to catch if this breaks again

PR #1423, opened 2021-06-21
Issue 1398: fix dropdown auto-close not working when wrapped in another component

margaree:
The problem seemed to be with the `focus` event on `document.body` not being triggered for this scenario. If I instead listen for the blur event on the component then that does seem to work as expected, but I'm not sure if there is a reason that it wasn't done that way initially that I'm missing.

Fixes https://github.com/BrightspaceUI/core/issues/1398


dbatiste:
> The problem seemed to be with the `focus` event on `document.body` not being triggered for this scenario. If I instead listen for the blur event on the component then that does seem to work as expected, but I'm not sure if there is a reason that it wasn't done that way initially that I'm missing.

Yeah, unfortunately I don't recall if there was a specific reason for `focus` over `blur`.  I think there were a lot of scenarios, and I don't necessarily think they are well captured in tests unfortunately.  Things like clicking out of the browser, and then moving focus back into the browser.  I vaguely recall some problems with the browser's Find/Search as well. I would just suggest doing lots of manual unit testing.

margaree:
Thanks for that additional context. I've been playing around with it this morning and the change looks fairly safe (with the addition of the change mentioned below in "Dropdown in card"). 

I tested the following in addition to the issue case on Chrome, Safari, Firefox, and Edge (non-legacy):
- Clicking outside of the browser window then returning focus to the window (dropdown does not close)
- Find/search then exiting (behaves as before - with Safari and Edge pressing escape to close the search also closes the dropdown)
- Menu (e.g., variety of keyboard and click interactions)
- Dropdown in card (I did find a bit of a problem with this where if I clicked on the dropdown to open it, then tabbed the dropdown wouldn't close until focus left the card. If I keep `document.body.addEventListener('focus', this.__onAutoCloseFocus, true);` in addition to the new `this.addEventListener('blur', this.__onAutoCloseFocus, true);` then that problem goes away. Is there any problem with having both?)

Something that seemed buggy that I noticed that did also exist before this change was that in Firefox, I can't tab within the dropdown, focus just leaves it (e.g., there are two links in the dropdown, opening the dropdown with enter causes focus to go to the first link, tabbing then causes focus to leave the dropdown. With any other browser focus will go to the second link).

dbatiste:
> Dropdown in card (I did find a bit of a problem with this where if I clicked on the dropdown to open it, then tabbed the dropdown wouldn't close until focus left the card. If I keep document.body.addEventListener('focus', this.__onAutoCloseFocus, true); in addition to the new this.addEventListener('blur', this.__onAutoCloseFocus, true); then that problem goes away. Is there any problem with having both?)

That does sound a bit peculiar, but I think it should be ok to have both.  If it were the other way around I can imagine scenarios where a focus or blur event would not be observed from the perspective of the body, but I don't _think_ that's the case here.

Card does have some special logic for tracking the dropdown open/closed state but I think it's just to control `z-index` so that dropdowns stack correctly when the user hovers over adjacent cards.  That should be unrelated to this as well.

dlockhart (approved):
This seems reasonable to me. I actually don't understand how this was working previously, since the [focus event isn't supposed to bubble](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event) at all... which means it should _never_ be reaching the `<body>` element. Hmm.

dbatiste:
> This seems reasonable to me. I actually don't understand how this was working previously, since the [focus event isn't supposed to bubble](https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event) at all... which means it should _never_ be reaching the `<body>` element. Hmm.

It was using `capture: true`.

PR #1425, opened 2021-06-21
fix: Guard against race conditions when communicating with ifrau host dialog service

jorge-ramirez-arredondo:
Found a couple of race conditions that can cause the ifrau host dialog service not to properly open or close the host backdrop.

First, if the dialog in the iframe is being re-created instead of being updated when opening/closing, the `_ifrauDialogService` can still be `undefined` when it comes time to try to show/hide the host backdrop. This PR addresses that by making sure to `await` on the promise for getting `_ifrauDialogService` when trying to access it.

Secondly, if there is one dialog in the iframe that is instantiated open and one that is instantiated closed, the closed one can trigger the host backdrop to hide after the open one already triggered it to show (which means we'll end up with the dialog open but the host backdrop hidden). This race condition is based on dialog rendering order. This PR addresses this issue by keeping a registry of all open dialogs in an iframe and only triggering the backdrop show/hide when appropriate.

Related  defect: [DE43960](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F601630684655&fdp=true?fdp=true)

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -73,6 +73,8 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		this._useNative = (window.D2L.DialogMixin.hasNative && window.D2L.DialogMixin.preferNative);
 		this._updateSize = this._updateSize.bind(this);
 		this._updateOverflow = this._updateOverflow.bind(this);
+		this._isIframed = false;
+		this._instanceId = Symbol('Instance ID');
 	}
 
 	async connectedCallback() {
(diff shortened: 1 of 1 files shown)


dbatiste:
If I understand correctly, the second issue stems from a single `d2l-backdrop` element being shared in the host page - is that correct?  If so, perhaps that should not be the case, but rather, we should pass an Id through to the backdrop service and it can determine whether to hide or show a particular backdrop.  Or, if we still need to utilize a single backdrop in the host (ex. to avoid double backdrops in a nested dialog scenario), we can track the dialog ids on the host backdrop.

Ideally, we don't want to have any conditional logic baked into the dialog web components to accommodate FRAs.  While that ship has already sailed (i.e. there's already some logic present), we ideally don't want to add more FRA logic if we can avoid it.  One day when FRAs are gone, we can just stop supporting iFrau, and there is little or no changes to be made to the core components.

jorge-ramirez-arredondo:
> If I understand correctly, the second issue stems from a single `d2l-backdrop` element being shared in the host page - is that correct? If so, perhaps that should not be the case, but rather, we should pass an Id through to the backdrop service and it can determine whether to hide or show a particular backdrop. Or, if we still need to utilize a single backdrop in the host (ex. to avoid double backdrops in a nested dialog scenario), we can track the dialog ids on the host backdrop.
> 
> Ideally, we don't want to have any conditional logic baked into the dialog web components to accommodate FRAs. While that ship has already sailed (i.e. there's already some logic present), we ideally don't want to add more FRA logic if we can avoid it. One day when FRAs are gone, we can just stop supporting iFrau, and there is little or no changes to be made to the core components.

Ah, okay. I see what you're saying. Yeah, the problem comes from sharing one backdrop and having some dialogs open it while others close it, so what you're saying makes sense.

I think for now I'd prefer to keep it with one shared backdrop and just move the multiple dialog tracking logic to the dialog service like you suggested. It'll probably have less of an impact on both the web component and the dialog service itself.

I'll take a quick crack at it. Hopefully, it doesn't take me long to make that change.

dlockhart:
@jorge-ramirez-arredondo not sure what the status of this PR is... should it be closed for now until more work can be done? Our team gets pinged about it daily. :)

dbatiste:
Closing this PR per discussion - the first issue is addressed by more recent changes, and the second issue will be addressed with a different solution.

PR #1426, opened 2021-06-21
US127106 - Ability to lazy load filter values

svanherk:
Was working on adding the search input, but it actually made sense to get this in first so I can hook into the loading flow when a search is occurring.

This PR does a few things:
- In the multiple dimension scenario, instead of finding the first focusable item in the hierarchical view we now focus on the return button.  This helps with the case where the contents are still loading, and is actually what the design calls for and I just missed it.
- Allow the consumer to set a `loading` property to tell us to show that dimension's loading state (which for the set type is just a loading spinner)
- Fire an event when a dimension is opened for the first time.  This will be used by consumers to tell them to set the `loading` property and start loading that dimension's values if they are using lazy-loading.

### DIFF
--- a/components/filter/README.md
+++ b/components/filter/README.md
@@ -27,7 +27,7 @@ The `d2l-filter` component allows a user to filter on one or more dimensions of 
 
 **Events:**
 * `d2l-filter-change`: dispatched when any filter value has changed (may contain info about multiple changes)
-* `d2l-filter-dimension-open`: dispatched when a dimension is opened (if there is only one dimension, this will be dispatched when the dropdown is opened)
+* `d2l-filter-dimension-first-open`: dispatched when a dimension is opened for the first time (if there is only one dimension, this will be dispatched when the dropdown is first opened)
 
 ## Filter Dimension Types
 
--- a/components/filter/README.md
+++ b/components/filter/README.md
@@ -27,6 +27,7 @@ The `d2l-filter` component allows a user to filter on one or more dimensions of 
 
 **Events:**
 * `d2l-filter-change`: dispatched when any filter value has changed (may contain info about multiple changes)
+* `d2l-filter-dimension-open`: dispatched when a dimension is opened (if there is only one dimension, this will be dispatched when the dropdown is opened)
 
 ## Filter Dimension Types
 
@@ -59,6 +60,7 @@ The `d2l-filter-dimension-set` component is the main dimension type that will wo
 | Property | Type | Description |
 |--|--|--|
 | `key` | String, required | Unique identifier for the dimension |
+| `loading` | Boolean | Whether the values for this dimension are still loading and a loading spinner should be displayed |
 | `text` | String, required | Text for the dimension in the menu |
 
 #### d2l-filter-dimension-set-value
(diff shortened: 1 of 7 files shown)


components/filter/test/filter.test.js
```diff
@@ -28,40 +28,89 @@ describe('d2l-filter', () => {
 		runConstructor('d2l-filter');
 	});
 
-	describe('change event', () => {
```
svanherk, line 31: This looks a lot better if you hide whitespace changes

components/filter/README.md
```diff
@@ -27,6 +27,7 @@ The `d2l-filter` component allows a user to filter on one or more dimensions of
 
 **Events:**
 * `d2l-filter-change`: dispatched when any filter value has changed (may contain info about multiple changes)
+* `d2l-filter-dimension-open`: dispatched when a dimension is opened (if there is only one dimension, this will be dispatched when the dropdown is opened)
```
dlockhart, line 30: Wondering if `d2l-filter-dimension-first-open` would be easier for folks to use in a lazy load scenario? Otherwise they might have to track this themselves or remove their event listener.

components/filter/filter.js
```diff
@@ -292,29 +303,35 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	}
 
 	_handleDimensionShowComplete() {
-		const dimension = this.shadowRoot.querySelector(`[data-key="${this._activeDimensionKey}"]`);
-		const focusable = getFirstFocusableDescendant(dimension);
-		if (focusable) {
-			focusable.focus();
-		}
+		const returnButton = this.shadowRoot.getElementById('d2l-filter-dimension-header-return');
```
dlockhart, line 306: Would `this.shadowRoot.querySelector('d2l-button-icon')` work as well? Just avoids an `id`.

components/filter/README.md
svanherk: I could!  For the labs component, they currently just listen for the "tab-selected" event which is dispatched every time, so they're used to having to handle it.  But this would be a nice enhancement.  Maybe a `openEventsSent` array that contains the keys we've notified for?  That way a slot change doesn't blow away that info?

components/filter/filter.js
svanherk: Yeah that _should_ work - the clear buttons won't be icon buttons, and I can narrow it down to the header

components/filter/README.md
dlockhart: Yeah either the keys or you could have a `Map` of the element nodes themselves. That way if the slot changed and they replaced the node with another empty one you'd fire the event again. But maybe I'm thinking too much into this and it's more complicated than it's worth. Could save this for if it actually comes up.

components/filter/filter.js
dlockhart: Or even `.querySelector('d2l-button-icon[icon="tier1:chevron-left"]')`.

components/filter/README.md
svanherk: Looking at current usages, there is one place that is using the repeat notification as a feature.  They display/hide a button in the footer based on being on that tab.  The new filter won't let them do this, so that will need to be solved a different way regardless.

dlockhart (approved):
This looks good -- if you want to get this in and then we can figure out the loading/ARIA stuff after that's cool. 👍

PR #1430, opened 2021-06-23
feat: Add tier2:classes icon

krshnpatel:
## Summary
A while back, I realized that Wave used the `tier3:classes` icon for the nav bar instead of a tier2 icon to match the others. ~So our UX designer, Nick Pagee, created a new tier2 sized SVG for the `classes` icon.~ Using the icon created by Jeff Geurts.

## Notes
I tested this icon on the Wave app as well as the icons demo page on `core` to make sure other attributes of `d2l-icon` work well with this SVG.

I see that other icons have the following format:
```html
<svg>
  <path/>
</svg>
```
So I'm wondering if this SVG needs to follow that **exact** format.


components/icons/images/tier2/classes.svg
```diff
@@ -0,0 +1,181 @@
...
+                  </g>
+                </g>
+              </g>
+            </g>
+          </g>
+        </g>
+      </g>
+    </g>
+  </g>
+</svg>
```
dlockhart, line 181: This is pretty huge! Make sure you [follow the instructions](https://github.com/BrightspaceUI/core/tree/master/components/icons#updating-or-contributing-new-icons) for contributing. I have a feeling after putting this through SVGOmg it'll look a little better (no need for lots of `id`s and `data-name`s.

krshnpatel: So I tried to run it through SVGOMG but it doesn't seem to help much or at all 😞

dlockhart: Haha I see that. I don't think any of the `data-name` attributes are needed, so you could try removing them. Also likely most of the `id`s aren't needed either. You could search for each one and if it's only referenced once, then remove it.

After that, re-run it through SVGOmg and it may be able to collapse a lot of the `<g>`s together. My guess is whatever editor was used to make this is leaving artifacts in the SVG for itself. You could also ask Nick if maybe there's a different way it could be exported to exclude all that.

@geurts I'm not sure if you've found a good technique to get cleaner SVGs out of these apps.

krshnpatel: So I removed `date-name` and `id` attributes wherever I could. I'm wondering if it's possible to replace the `clipPaths` with something else 🤔

dlockhart: Agreed, those seem like overkill for something that's literally 9 boxes. I wonder if the tier3 one (which is just a single path) could just be tweaked to be slightly smaller.

geurts: I actually recreated this back in April for Wave, I guess it got lost in the shuffle! I'll attach it here. When I hack this into Wave via Chrome dev tools, it works well and looks good 👍 

Oh, GitHub won't let me drop an SVG in here. I'll send it over Slack.

components/icons/images/tier2/classes.svg
```diff
@@ -0,0 +1,13 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+  <g fill="none" fill-rule="evenodd" stroke="#494C4E" stroke-width="2">
```
dlockhart, line 2: @krshnpatel did you try this new version out on the demo page? I'm worried that the use of "stroke" instead of "fill" isn't going to work with the demo where it changes the colour.

krshnpatel: Ahhh yes, I did try it out on the demo page but missed that the colour actually didn't change for this logo 😓

dlockhart: A quick fix you could try would be to use `fill` of `#494C4E` on the individual `<rect>`s.

krshnpatel: @geurts gave me another SVG that works perfectly on the demo page (i.e. colour override, size override) 😄

dlockhart (approved):

PR #1432, opened 2021-06-23
US122213 - Configure filter empty state

svanherk:
This PR does two things:
- Enhance the screen reader experience a bit more for the loading state (until we add more to the loading spinner itself)
- Handle a filter dimension that has no values, and utilize the same screen reader enhancements to announce this as well

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -68,17 +68,17 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				line-height: unset;
 			}
 
-            .d2l-filter-dimension-info-message {
-                padding: 0.9rem 0;
-                text-align: center;
-            }
+			.d2l-filter-dimension-info-message {
+				padding: 0.9rem 0;
+				text-align: center;
+			}
 
-            /* Needed to "undo" the menu hover styles */
-            :host(:hover) .d2l-filter-dimension-info-message,
-            :host(:hover) .d2l-filter-dimension-set-value-text {
-                color: var(--d2l-color-ferrite);
-                cursor: default;
-            }
+			/* Needed to "undo" the menu hover styles */
+			:host(:hover) .d2l-filter-dimension-info-message,
+			:host(:hover) .d2l-filter-dimension-set-value-text {
+				color: var(--d2l-color-ferrite);
+				cursor: default;
+			}
 
 			d2l-loading-spinner {
 				padding-top: 0.6rem;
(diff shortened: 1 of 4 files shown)


components/filter/filter.js
```diff
@@ -207,13 +214,22 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+                <div class="d2l-filter-dimension-info-message d2l-body-small" role="alert">
+                    ${this.localize('components.filter.noFilters')}
+                </div>
+            `;
+		}
+
 		return html`
 			<d2l-list
 				@d2l-list-selection-change="${this._handleChangeSetDimension}"
+				data-key="${dimension.key}"
```
svanherk, line 232: This simplifies the `_handleChangeSetDimension` logic, especially now that there's a `div` between the list and the `hierarchical-view`

components/filter/filter.js
```diff
@@ -361,6 +376,15 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		this._setFilterCounts();
 	}
 
+	_isDimensionEmpty(dimension) {
```
svanherk, line 379: This will eventually be used by the search input and clear button as well, and be a tad more complicated comparison once search is in

components/filter/test/filter.test.js
```diff
@@ -42,6 +42,13 @@ describe('d2l-filter', () => {
 		});
 	});
 
+	describe('info messages', () => {
```
svanherk, line 45: Will probably remove this once visual-diff is in

components/filter/filter.js
```diff
@@ -207,13 +214,22 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		if (dimension.loading) {
 			return html`
 				<d2l-loading-spinner></d2l-loading-spinner>
-				<div class="d2l-offscreen">${this.localize('components.filter.loading')}</div>
+				<div class="d2l-offscreen" aria-busy="true" role="alert">${this.localize('components.filter.loading')}</div>
```
dlockhart, line 217: Nit: semantically this and the "empty" `<div>` might be better off a `<p>` tags. 🤷  Note that paragraph tags come with built-in top/bottom margins which may or may make your top/bottom padding unnecessary.

dlockhart (approved):

components/filter/filter.js
svanherk: Yeah that makes sense.  Looks like I still need the padding because `d2l-body-small` adds `margin: auto;`, which removes `p`'s default margins.

PR #1434, opened 2021-06-24
Add the search-input to the filter

svanherk:
To Do:
- Some perf testing of this approach
- Tests
- README

### DIFF
--- a/components/filter/filter-dimension-set-value.js
+++ b/components/filter/filter-dimension-set-value.js
@@ -9,14 +9,28 @@ class FilterDimensionSetValue extends LitElement {
 
 	static get properties() {
 		return {
+			/**
+			 * Whether the value is hidden or not (used when applying search results)
+			 */
+			hidden: { type: Boolean, reflect: true },
+			/**
+			 * REQUIRED: Unique key to represent this value in the dimension
+			 */
 			key: { type: String },
+			/**
+			 * Whether this value in the filter is selected or not
+			 */
 			selected: { type: Boolean, reflect: true },
+			/**
+			 * REQUIRED: The text that is displayed for the value
+			 */
 			text: { type: String }
 		};
 	}
 
 	constructor() {
 		super();
+		this.hidden = false;
 		this.selected = false;
 		this.text = '';
 	}
@@ -28,7 +42,7 @@ class FilterDimensionSetValue extends LitElement {
 		changedProperties.forEach((oldValue, prop) => {
 			if (oldValue === undefined) return;
 
-			if (prop === 'selected' || prop === 'text') {
+			if (prop === 'hidden' || prop === 'selected' || prop === 'text') {
 				changes.set(prop, this[prop]);
 			}
 		});
--- a/components/filter/filter-dimension-set.js
+++ b/components/filter/filter-dimension-set.js
@@ -19,6 +19,15 @@ class FilterDimensionSet extends LitElement {
 			 */
 			loading: { type: Boolean },
 			/**
+			 * Whether to hide the search input, fire an event on search, or perform a simple text search
+			 * @type {'none'|'event'|'simple'}
+			 */
+			searchType: { type: String, attribute: 'search-type' },
+			/**
+			 * Value for the search input
+			 */
+			searchValue: { type: String, attribute: 'search-value' },
+			/**
 			 * REQUIRED: The text that is displayed for the dimension title
 			 */
 			text: { type: String }
@@ -28,6 +37,7 @@ class FilterDimensionSet extends LitElement {
 	constructor() {
 		super();
 		this.loading = false;
+		this.searchType = 'simple';
 		this.text = '';
 		this._slot = null;
 	}
@@ -46,9 +56,17 @@ class FilterDimensionSet extends LitElement {
 
 		const changes = new Map();
 		changedProperties.forEach((oldValue, prop) => {
+			if (prop === 'searchValue') {
+				if (this.searchType === 'simple') {
+					requestAnimationFrame(() => {
+						this._performSearch();
+					});
+				}
+			}
+
 			if (oldValue === undefined) return;
 
-			if (prop === 'text' || prop === 'loading') {
+			if (prop === 'searchValue' || prop === 'text' || prop === 'loading') {
 				changes.set(prop, this[prop]);
 			}
 		});
@@ -66,6 +84,17 @@ class FilterDimensionSet extends LitElement {
 		}));
 	}
 
+	_getSearchType() {
+		switch (this.searchType) {
+			case 'none':
+				return;
+			case 'event':
+				return 'event';
+			case 'simple':
+				return 'on';
+		}
+	}
+
 	_getSlottedNodes() {
 		if (!this._slot) return [];
 		const nodes = this._slot.assignedNodes({ flatten: true });
@@ -76,6 +105,7 @@ class FilterDimensionSet extends LitElement {
 		const valueNodes = this._getSlottedNodes();
 		const values = valueNodes.map(value => {
 			return {
+				hidden: value.hidden,
 				key: value.key,
 				selected: value.selected,
 				text: value.text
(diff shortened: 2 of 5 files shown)


components/filter/filter-dimension-set.js
```diff
@@ -66,6 +84,17 @@ class FilterDimensionSet extends LitElement {
 		}));
 	}
 
+	_getSearchType() {
```
svanherk, line 87: This seems kinda unnecessary, but I was trying to make it easy for other filter types who may not fit into these three categories perfectly (i.e. they just return in their `_getSearchType` function because search doesn't make sense for them, or they could have multiple search types and just tell the filter to make sure search is "on")

components/filter/filter-dimension-set-value.js
```diff
@@ -9,14 +9,28 @@ class FilterDimensionSetValue extends LitElement {
 
 	static get properties() {
 		return {
+			/**
+			 * Whether the value is hidden or not (used when applying search results)
+			 */
+			hidden: { type: Boolean, reflect: true },
```
dlockhart, line 15: All HTML elements get the `hidden` attribute/property, so no need to specify it here.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

components/filter/filter-dimension-set.js
```diff
@@ -96,6 +126,22 @@ class FilterDimensionSet extends LitElement {
...
+		valueNodes.forEach(value => {
+			if (this.searchValue === '') {
+				value.hidden = false;
+			} else {
+				if (value.text.toLowerCase().indexOf(this.searchValue.toLowerCase()) > -1) {
+					value.hidden = false;
+				} else {
+					value.hidden = true;
+				}
+			}
```
dlockhart, line 141: Wondering out loud here, but would it make sense for the hiding/showing  to be done totally by the top-level filter on its list items that it renders, and not worry about reflecting anything down to the dimension-set or dimension-set-value?

Instead of how it's working now where the top-level filter tells the dimension-set a search has happened, which then hides its own values which then gets communicated back up to the top-level filter to hide the rendered list items.

components/filter/filter-dimension-set.js
```diff
@@ -18,6 +18,15 @@ class FilterDimensionSet extends LitElement {
 			 * Whether the values for this dimension are still loading and a loading spinner should be displayed
 			 */
 			loading: { type: Boolean },
+			/**
+			 * Whether to hide the search input, fire an event on search, or perform a simple text search
+			 * @type {'none'|'event'|'simple'}
+			 */
+			searchType: { type: String, attribute: 'search-type' },
```
dlockhart, line 25: Thoughts about `none`, `manual` and `automatic`?

components/filter/filter.js
```diff
@@ -43,10 +47,28 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+			.d2l-filter-dimension-header,
+			.d2l-filter-dimension-header-actions {
 				align-items: center;
 				display: flex;
 			}
+
+			d2l-input-search {
+				flex-grow: 1;
+				padding-left: 0.3rem;
+				padding-right: 0.6rem;
```
dlockhart, line 64: Do margins still get you the same effect? I feel weird about setting padding on a thing where that doesn't impact the space inside its borders.

components/filter/filter-dimension-set.js
```diff
@@ -18,6 +18,15 @@ class FilterDimensionSet extends LitElement {
...
+			/**
+			 * Value for the search input
+			 */
+			searchValue: { type: String, attribute: 'search-value' },
```
dlockhart, line 29: Is there a use case for them needing to specify an initial value to search on? Seems a bit weird.

components/filter/filter.js
```diff
@@ -350,14 +410,44 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 	}
 
+	_handleSearch(e) {
+		e.stopPropagation();
+		const dimension = !this._activeDimensionKey ? this._dimensions[0] : this._dimensions.find(dimension => dimension.key === this._activeDimensionKey);
+		const dimensionNode = !this._activeDimensionKey ? this._dimensionNodes[0] : this._dimensionNodes.find(node => node.key === this._activeDimensionKey);
+		dimension.searchValue = e.detail.value;
+		dimensionNode.searchValue = e.detail.value;
+
+		if (dimension.searchType === 'event') {
```
dlockhart, line 420: Do we have a use case for this type of asynchronous search right away? Just wondering if it's worth implementing all this extra logic where the simple search should hopefully meet a lot of use cases. Or would it be for places that might have a lot of values and require some form of paging?

svanherk: So My Courses uses HM calls to search.  It doesn't do paging, but was maybe setting itself up to one day do so. Or maybe it searches deeper than just the department name (to the description)?  I've never had to dig into that api, so that's just a guess

components/filter/filter-dimension-set.js
svanherk: So the old labs filter allows this, but from my searching it doesn't _seem_ like it's been used anywhere.  I think the plan was to use it for Quick Eval, but it was decided to allow the filters to be set up in a default state on page load, but have the search still clear.  So I could probably remove this and keep things simple for now

components/filter/filter.js
svanherk: Yup I can switch to margins here

components/filter/filter-dimension-set.js
svanherk: Yeah I like that, I think I could then cleanup that mapping function too to be simpler

components/filter/filter-dimension-set.js
svanherk: Yeah, this was my main "?" when implementing this.  My thinking was - the current implementation is kinda overkill for if they use the `automatic` option, because we could have just kept it all in the filter's internal components.  _But_ if they do `manual`, I have to setup the same system anyway, where I coordinate between the two and keep things in sync.

So I figured it was cleaner (code-wise) to have both go through the same workflow, and had the benefit of keeping the searching logic with the dimension type instead of being another thing filter needed to know (although we can still do that somewhat either way, with a searching function).  If we remove the `search-value` stuff though, _maybe_ that cleans up the logic a bit and it's worth having the two separate ways to take advantage of just modifying the internals when we can.  I already made a decision that I wasn't allowing you to switch search-type halfway through - you should know this from the get-go.

components/filter/filter-dimension-set.js
dlockhart: Right. Maybe for like a "saved search" type scenario where the search filters are reflected in the URL (for bookmarking, sending links) and then the state would need to be restored that way.

components/filter/filter-dimension-set.js
dlockhart: I guess it depends on how manual gets implemented. If they just re-render only the values that match the search vs. re-rendering all the values and setting `hidden` on the ones which don't match.

I was under the impression that the only time you'd want manual search would be if there was some kind of paging implemented... in which case returning "all" wouldn't make sense anyway.

components/filter/filter.js
dlockhart: Right, so possibly a combination of searching multiple fields (which may or may not even be rendered) and paging.

components/filter/filter-dimension-set.js
svanherk: So doing a _quick_ look at all the current usages of filter, it seems like there are three methods in use:
1. Simple case-insensitive text compare and then hide options
2. Simple case-insensitive text compare and then remove options
3. Call an api to get the results and then remove options (My Courses)

So maybe we could simplify this a lot... automatic search is applied to the local elements and the consumer doesn't care and can't control that at all.  And for automatic, we'll send an event and it's up to the consumer to physically remove the options they want to hide, which triggers a `slotchange` event and re-render of that filter dimension.

components/filter/filter-dimension-set-value.js
svanherk: If I don't add this though, seems like it won't show up as a `changedProperty` in `updated` (https://github.com/BrightspaceUI/core/pull/1434/files#diff-065ae76afdeb4d78f535d93bcd59d608978e7705d8c0671bf026def23ee3e641R45).  Might not matter, I'm trying out teh simplified version now.

dlockhart: Ah yes, the built-in Lit property stuff. Hopefully not needed!

svanherk:
Closing this PR in favour of https://github.com/BrightspaceUI/core/pull/1451

PR #1435, opened 2021-06-25
Switch to latest d2l-license-checker

devpow112:
Also switch to only analyse production dependencies.


dlockhart (approved):
Yay thanks cleaning this up!

PR #1443, opened 2021-06-28
fix: US129213 code review comments

annabelsegalld2l:
Addressing some outstanding comments from https://github.com/BrightspaceUI/core/pull/1414

- add ability for consumers to call opened = true/false in addition to open()/close() 
- note: the closing animation will not run if opened = false is used. At that point, it's too late to invoke the closing state to animate the fading out. However, the backdrop and dropdown do get removed without the scroll lock issues. I don't think this should be a big issue, as most consumers are calling toggleOpen() which uses the open() and close() functions. 
- remove closing attribute, use data-closing attribute instead
- Also adding a fix for body scrolling here to prevent any future issues where the backdrop is unexpectedly removed from the DOM and scrolling is blocked. 
- Adding a timeout of 10ms on mobile dropdown tests to reduce flakiness. 
- Waiting for content.updateComplete in newly added axe test.


dlockhart (approved):

components/backdrop/backdrop.js
```diff
@@ -75,6 +75,12 @@ class Backdrop extends LitElement {
 		this._state = null;
 	}
 
+	disconnectedCallback() {
+		// allow body scrolling if backdrop is removed from the DOM
```
annabelsegalld2l, line 79: Also adding a fix for body scrolling here to prevent any future issues where the backdrop is unexpectedly removed from the DOM and scrolling is blocked. This will only release the scroll if it had been blocked by this particular backdrop (ie. shouldn't affect any other backdrop instances on a page, it can only access its personal scroll key, if it exists)

components/dropdown/test/dropdown-content.visual-diff.js
```diff
@@ -71,6 +71,7 @@ describe('d2l-dropdown-content', () => {
 			await page.setViewport({ width: 600, height: 500 });
 			const selector = `#${testName}`;
 			await open(page, selector);
+			await new Promise(r => setTimeout(r, 10));
```
annabelsegalld2l, line 74: Added this 10ms for the mobile tests only. There seem to be some rare flakes like https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/dropdown-content/2021-06-28-15.10.29.73-807132d1a625dff191a065988df4e5c6f962b887-report.html and https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/dropdown-content/2021-06-28-12.58.33.311-4238a938811ac8fcd921a1ea1f6999a3769eac25-report.html that fail with only a couple of different pixels. Adding the delay was the only solution I found (no failures after 30 re-runs)

components/backdrop/backdrop.js
annabelsegalld2l: I would agree. If the backdrop gets disconnected we would want to reveal any elements hidden by it.

dlockhart (approved):
This LGTM!

PR #1446, opened 2021-06-29
feat: US129214 Add bottom tray option to mobile dropdown

annabelsegalld2l:
Design : https://xd.adobe.com/view/5205ec02-86e7-48b0-ad0a-04aa76db7759-0756/?fullscreen
Visual diff changes: https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/dropdown-content/2021-06-29-21.07.31.34-8f2920715bef3d77af598bfa71e63e51314e161f-report.html

Added a mobile-tray="bottom" option to slide in from the bottom of the screen. 
bottom_tray

In addition to adding the dropdown, I had some other fixes for all the mobile trays:

- Fixing RTL behaviour to make sure the 'close' button is centered on the tray
- Adding the footer line, even if there is no footer content (other than the close button) 
- If there is a close button:
- with NO other footer contents: pad 12px around as in design
- WITH other footer contents: pad top of close button with 12px, but adhere to dropdown style of 20px footer padding


components/dropdown/dropdown-content-mixin.js
```diff
@@ -739,36 +744,95 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+		}
+
+		let maxHeightOverride;
+		if (!specialMobileStyle) {
+			maxHeightOverride = this._contentHeight ? `${this._contentHeight}px` : 'none';
+		}
+		if (mobileTrayRightLeft) maxHeightOverride = '';
+		if (mobileTrayBottom) {
+			// default maximum height for tray (42px margin)
+			const mobileTrayMaxHeightDefault = window.innerHeight - 42;
```
annabelsegalld2l, line 770: Similar to the left/right style, we want to always leave 42px of the backdrop visible at the top, and also adhere to the maxHeight if given by the consumer. The maxwidth is ignored, as it is set to 100%

components/dropdown/dropdown-content-mixin.js
```diff
@@ -62,7 +62,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 				attribute: 'no-mobile-close-button'
 			},
 			/**
-			 * Override default mobile dropdown style. Specify one of 'left' or 'right'.
+			 * Override default mobile dropdown style. Specify one of 'left','right', or 'bottom'.
```
dlockhart, line 65: I believe the way to get these options to show up properly in the new Daylight site is:

```suggestion
			 * Mobile dropdown style.
			 * @type {'left'|'right'|'bottom'}
```

Should we default this to something like `'none'` instead of `undefined`?

components/dropdown/dropdown-content-mixin.js
dlockhart: Should we move that `42` value up into a constant at the top of the file that they can all share? Might make changing it in the future easier.

components/dropdown/dropdown-content-mixin.js
annabelsegalld2l: I was thinking to leave this as undefined. This way, we prevent sprouting an attribute that says mobile-tray="none" that the consumer was not aware of.

annabelsegalld2l:
Closing this PR since the merge commit made some css changes. I've made a new PR with the same changes and the correct merge.

PR #1450, opened 2021-06-30
Daylight: Button

margaree:


components/button/README.md
```diff
@@ -32,34 +54,24 @@ To make your `d2l-button` accessible, use the following properties when applicab
 | `aria-haspopup` | [Indicate clicking the button opens a menu](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-haspopup). Example: [d2l-dropdown](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-opener-mixin.js#L46). |
 | `description` | Use when text on button does not provide enough context. |
 
-## d2l-button-subtle
+## Subtle Button [d2l-button-subtle]
```
margaree, line 57: Subtle button and icon button have the `hAlign` property which aligns them with text. Should there be a separate demo that shows that behaviour (non-interactive), since playing around with it in the interactive demo wouldn't really show that (though we could also add some text in there to have that be demonstrated, it just could clutter up the demo)?

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
+* Don't use buttons for navigation, use a link instead
+* Don't open menus with buttons - use a dropdown instead
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
+
+## Button [d2l-button]
+
+The `d2l-button` element can be used just like the native button element, but also supports the `primary` attribute for denoting the primary button.
 
-| Property | Type | Description |
```
margaree, line 18: This deletes the properties tables. I can add a note for now that properties can be found within the code, since I believe we still aren't advertising Daylight. They can also still remain and we can wrap them in the  comments. We would just probably then want to come back and remove that, then add a note about using Daylight, in order to only maintain the properties in the code.

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
-| `disabledTooltip` | String | Tooltip text when disabled |
-| `primary` | Boolean | Styles the button as a primary button |
+```html
+<!-- docs: live demo
+name: d2l-button
+-->
+<script type="module">
+  import '@brightspace-ui/core/components/button/button.js';
+</script>
+<d2l-button $attributes>My Button</d2l-button>
```
margaree, line 44: I'm thinking I'll add a hidden-from-docs-site note about ignoring the `$attributes` piece in the demo code.

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
+## Button [d2l-button]
+
+The `d2l-button` element can be used just like the native button element, but also supports the `primary` attribute for denoting the primary button.
 
-| Property | Type | Description |
-|--|--|--|
-| `description` | String | A description to be added to the `button` for accessibility |
-| `disabled` | Boolean | Disables the button |
-| `disabledTooltip` | String | Tooltip text when disabled |
-| `primary` | Boolean | Styles the button as a primary button |
```
margaree, line 23: `primary` is located in ButtonMixin which causes it to appear in attributes for Subtle and Icon buttons even though it does nothing for them. I haven't figured out a way around this. We can add `@ignore` and then also put primary in button but that seems hacky and there is at least one other usage of ButtonMixin that uses primary so that could have downstream affects. Another option is add some sort of property exclusion functionality to the daylight site code. Thoughts on this?

components/button/button-mixin.js
```diff
@@ -5,37 +5,64 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 	static get properties() {
 		return {
 			/**
-			 * Indicate expansion state of a collapsible element
```
margaree, line 8: I ignored a lot of the properties in `ButtonMixin` since they don't need to be in the properties table/weren't currently appearing in the README properties table (the accessibility ones generally appear in the `Accessibility Properties` table as well). An alternative to this is if we went with the exclusion functionality mentioned previously we could leave them with jsdoc comments and still have them excluded from the properties table

components/button/README.md
```diff
@@ -69,68 +81,32 @@ To make your `d2l-button-subtle` accessible, use the following properties when a
...
 
 | Attribute | Description |
 |--|--|
+| `text` | **REQUIRED**. Acts as a primary label and tooltip. |
 | `aria-expanded` | [Indicate expansion state of a collapsible element](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-expanded). Example: [d2l-more-less](https://github.com/BrightspaceUI/core/blob/f9f30d0975ee5a8479263a84541fc3b781e8830f/components/more-less/more-less.js#L158). |
 | `aria-haspopup` | [Indicate clicking the button opens a menu](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-haspopup). Example: [d2l-dropdown](https://github.com/BrightspaceUI/core/blob/master/components/dropdown/dropdown-opener-mixin.js#L46). |
-| `text` | Acts as a primary label and tooltip and is **REQUIRED**. |
 | `aria-label` | Acts as a primary label. If `text` AND `aria-label` are provided, `aria-label` is used as the primary label, `text` is used as the tooltip. |
 
-## d2l-floating-buttons
```
margaree, line 107: Floating buttons have been moved into a separate document but have not been prepped as a Daylight site page.

components/button/README.md
dlockhart: Hmm, seems like it doesn't belong there eh. Could `primary` be removed from the mixin and then just copied to the places using the mixin that also needed it?

margaree: We could do that. I believe it's just Button and an IPSIS usage so should be straightforward.

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
+<d2l-button-subtle text="Subtle Button" icon="tier1:gear"></d2l-button-subtle>
+<d2l-button-icon text="Icon Button" icon="tier1:gear"></d2l-button-icon>
 ```
 
-**Properties:**
+## Best Practices
+<!-- docs: start dos -->
+* Use buttons to cause an action or launch a workflow
+* Keep button text short - see "Writing" guidelines
```
dlockhart, line 22: I guess we'll have to link to these articles once they exist?

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don't use multiple Primary buttons - the primary action should always be clear and obvious
+* Avoid icon-only buttons with unfamiliar icons, only use these for familiar concepts
+* Don't use buttons for navigation, use a link instead
+* Don't open menus with buttons - use a dropdown instead
```
dlockhart, line 29: Should link to the dropdown documentation here if possible.

components/button/README.md
dlockhart: Right. If it's going to be a couple months before the Daylight Site is ready, it might be nice to leave these in for now and add the comments to hide them?

components/button/README.md
dlockhart: I'm not too worried about the demo always being able to exercise every single property... a simple demo that makes it easy to understand the basic usage is nice.

Then we can decide if it's worth adding a separate demo just for the horizontal alignment stuff.

components/button/README.md
svanherk: Yeah, I can't seem to find a way that we could "comment" this out so it's not rendered in the readme when it's inside a code block (assuming we could update the the parser to understand that new syntax).

components/button/README.md
```diff
@@ -1,28 +1,50 @@
...
+The `d2l-button` element can be used just like the native button element, but also supports the `primary` attribute for denoting the primary button.
 
-| Property | Type | Description |
-|--|--|--|
-| `description` | String | A description to be added to the `button` for accessibility |
-| `disabled` | Boolean | Disables the button |
-| `disabledTooltip` | String | Tooltip text when disabled |
-| `primary` | Boolean | Styles the button as a primary button |
+```html
+<!-- docs: live demo
```
svanherk, line 38: I haven't been following along with the Daylight site PRs and how the parsing works, so this may be more difficult than it's worth.  But I assume if we moved this comment above the ` ```html`, then it wouldn't be displayed in the readme at all, like the other docs comments throughout.  Would that really mess up the parser though?

margaree: It would, yeah. Text/comments vs code blocks are parsed out as separate blocks so it would complicate the parsing for sure. We could probably replace the comment with the opening tag for the demo component but it's a bit complicated by the fact that we don't treat the content strictly as a slot. I'll do some experimenting and see.

svanherk: Ok cool, yeah not trying to derail things too much.  Would be a nice-to-have, especially where we repeat the attributes already there in the `defaults` section

components/button/README.md
dlockhart: Could the docs site make an assumption that the attributes go on the first `<d2l-*>` element inside the code block? That would likely cover 95% of cases and would allow us to avoid the confusion here.

dbatiste: I haven't been following this stuff super closely, so I might be a little late to the party here.  If so, and this is already figured out, then ignore.  😏 One concern I would have about assuming the first `d2l-*` element is that some elements require some parent element, so the element of interest would not be the first element.  Consider:
```html
<d2l-list>
	<d2l-list-header slot="header">
		<d2l-selection-action icon="..." text="..." requires-selection></d2l-selection-action>
		<d2l-selection-action icon="..." text="..."></d2l-selection-action>
	</d2l-list-header>
	<d2l-list-item selectable key="..." label="...">
		<img slot="illustration" src="...">
		<d2l-list-item-content>...</d2l-list-item-content>
	</d2l-list-item>
	<d2l-list-item selectable key="..." selected label="...">
		<img slot="illustration" src="...">
		<d2l-list-item-content>...</d2l-list-item-content>
	</d2l-list-item>
	<d2l-list-item selectable key="..." label="...">
		<img slot="illustration" src="...">
		<d2l-list-item-content>...</d2l-list-item-content>
	</d2l-list-item>
</d2l-list>
```
To demo some of these behaviours, these components must be used together... i.e. they can't just be demo'd in isolation.  Do we already have a plan for how to deal with more complex components?

svanherk: Or maybe that could be specified in the `<!-- docs: live demo` snippet?  Something like:
```
<!-- docs: live demo
name: d2l-button
attributes: d2l-button
-->
```
And that's actually a selector, so in Dave B 's case you could use something like:
```
<!-- docs: live demo
name: d2l-list
attributes: d2l-list-item:first-of-type
-->
```
(This might not make sense though with how the parsing/page creation works, not sure at which points we use JS or not)

margaree: So I think we could likely use what's in `name` for this if we wanted to go this route since `name` corresponds to whatever the component is that's being used for the properties table and then I think we could match that up to the first usage. @GeorgePlukov do you have any thoughts/concerns around this?

GeorgePlukov: It could be possible to move this into the demo comment, originally I did it using the `$attributes` for the reason Dave B mentioned where we may have embedded items that we want to be changing the properties on, not just the top level `d2l-*` element. 

Right now the demo code is replacing the $attribute symbol with the attributes as text and updating the playground-project to re-render the demo, so using a css selector wouldn't really work. If we were to use a name property would we just add the attributes to any tag with that name? I was thinking that there could be cases like demoing `d2l-list-item` where we would want a subset of all the items with that tag to update with the attributes

margaree: I'm a bit on the fence for if it should affect 1 or all. I think all could be useful, since we'd be selective with our demos, e.g., with the `menu-item` specific demo, we might have the first item be a simple menu item and the second be a nested menu, and so having toggling the `disabled` property in the table affect both `d2l-menu-item`s could be neat to see. However this could be complex with the case of `menu-item-radio` where if there were 2 radio items, we wouldn't want `selected` being toggled in the table to affect both. Could also add an option in the comment block for these cases for does it affect all or just first.

GeorgePlukov: Yeah thats an interesting case too, Im struggling to think of a better way to go with this so that may be our best option

dlockhart (approved):

components/button/README.md
```diff
@@ -1,28 +1,65 @@
...
+<!-- docs: start dos -->
+* Use buttons to cause an action or launch a workflow
+* Keep button text short - see "Writing" guidelines
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don't use multiple Primary buttons - the primary action should always be clear and obvious
+* Avoid icon-only buttons with unfamiliar icons, only use these for familiar concepts
+* Don't use buttons for navigation, use a [link](https://github.com/BrightspaceUI/core/tree/master/components/link) instead
+* Don't open menus with buttons - use a [dropdown](https://github.com/BrightspaceUI/core/tree/master/components/dropdown) instead
```
svanherk, line 29: Will these eventually link to the pages in the Daylight site instead?

svanherk (approved):

components/button/README.md
margaree: Yup they will. I'll update them once those pages exist.

PR #1452, opened 2021-06-30
DE44176 - Support begin, ref, and eqref LaTeX delimiters

grant-cleary:
This is now in use, so it's important we get this back in sync with the LMS.


dbatiste (approved):

dlockhart (approved):

grant-cleary:
@dbatiste, @dlockhart... Any objections if this goes in as a patch? I guess it's an academic distinction since most apps will lock on major version anyways, but this is definitely a bug from the LMS' point of view even if it feels feature-y from the component's point of view.

dlockhart:
Fine with me.

dbatiste:
Yep... no objections here.

PR #1453, opened 2021-07-01
visual-diff: try to un-skip skipped tests

margaree:
The dialog-fullscreen tests not seem to behave as expected and are not flaky from several test runs. I also tweaked the flaky `input-time-range` test so it's hopefully less flaky.


dlockhart (approved):
Nice!

svanherk:
> I also tweaked the flaky `input-time-range` test so it's hopefully less flaky.

🎉 🤞

Issue #1458, opened 2021-07-08
In Safari, d2l-input-text cursor always jumps to the end


wongvincent:
Tested the `d2l-input-text` demo on Mac Safari. When the cursor is not at the end, and we type a character, the cursor jumps to the end of the input field.

This bug is not reproducible on Mac Chrome, so it appears to be a Safari thing.

I don't have an iOS device to test if it happens on that too.

I found this jsfiddle demo to also test on: https://jsfiddle.net/zdennis/xpvt214o/178966
I think it has something to do with how the value keeps getting set.

PR #1459, opened 2021-07-08
fix: don't re-render unnecessarily

dlockhart:
Apparently [Safari has a bug](#1458) (thanks @wongvincent) where if you set the `value` property (even to the same thing it already was), it will always move focus to the end of the input. This is a problem, because we're listening to the `oninput` event on the input and internally reflecting that to our `value` property (so consumers can access it) which in turn causes a re-render which then moves the cursor to the end -- not a good experience as the user types.

The solution is to create our own getter/setter for it, so setting `_value` from `oninput` won't cause a re-render.

While testing this I noticed that hovering while focused also caused an unnecessary re-render so I changed that logic a bit as well.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -123,7 +123,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 			title: { type: String },
 			/**
 			 * The type of the text input
-			 * @type {'text'|'email'|'number'|'password'|'tel'|'url'}
+			 * @type {'text'|'email'|'number'|'password'|'search'|'tel'|'url'}
 			 */
 			type: { type: String },
 			/**
@@ -221,20 +221,26 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		this.readonly = false;
 		this.required = false;
 		this.type = 'text';
-		this.value = '';
+		this._value = '';
 
+		this._descriptionId = getUniqueId();
+		this._firstSlotWidth = 0;
 		this._focused = false;
 		this._hasAfterContent = false;
 		this._hovered = false;
 		this._inputId = getUniqueId();
-		this._firstSlotWidth = 0;
 		this._lastSlotWidth = 0;
-		this._descriptionId = getUniqueId();
+		this._prevValue = '';
 
 		this._handleBlur = this._handleBlur.bind(this);
 		this._handleFocus = this._handleFocus.bind(this);
 		this._handleMouseEnter = this._handleMouseEnter.bind(this);
 		this._handleMouseLeave = this._handleMouseLeave.bind(this);
+	}
+
+	get value() { return this._value; }
+	set value(val) {
+		this._setValue(val, true);
 	}
 
 	get selectionStart() {
@@ -279,6 +285,8 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
+
+		this._setValue(this.value, true);
 
 		const container = this.shadowRoot.querySelector('.d2l-input-text-container');
 		if (!container) return;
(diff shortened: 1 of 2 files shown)


components/inputs/input-text.js
```diff
@@ -298,7 +306,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 	}
 
 	render() {
-		const isFocusedOrHovered = !this.disabled && (this._focused || this._hovered);
+		const isFocusedOrHovered = !this.disabled && (this._focusedOrHovered);
```
dlockhart, line 309: The re-render only really cares if they're focused OR hovered -- it doesn't matter if they hover and then move away if they're still focused.

components/inputs/input-text.js
```diff
@@ -464,10 +473,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 
 	_handleFocus() {
 		this._focused = true;
+		this._focusedOrHovered = true;
 	}
 
 	_handleInput(e) {
-		this.value = e.target.value;
+		this._value = e.target.value; // set directly to avoid re-render which causes cursor jumping in Safari
```
dlockhart, line 480: This bypasses the logic in `updated()` for validation/form stuff... working on a better fix.

dbatiste:
Looks good.  It looks like `d2l-input-search` also has this problem, and it doesn't use `d2l-input-text` so I think it needs a special fix.  The `d2l-input-number` (which `d2l-input-percent` uses) uses `d2l-input-text` so this sol'n should already cover those cases.

components/inputs/input-text.js
```diff
@@ -367,8 +378,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 						style="${styleMap(inputStyles)}"
 						tabindex="${ifDefined(this.tabindex)}"
 						title="${ifDefined(this.title)}"
-						type="${this._getType()}"
-						.value="${this.value}">
```
dlockhart, line 371: I've removed the binding to the input's `value` altogether. Otherwise anytime a re-render is triggered by anything -- validation state change, value change, mouse hover, etc. -- the cursor will jump to the end in Safari.

The workaround is instead to manually reach into the DOM and set the input's value when we initially render (in `firstUpdated`) and whenever the property setter is set (including via attribute).

components/inputs/test/input-text.test.js
```diff
@@ -324,6 +324,14 @@ describe('d2l-input-text', () => {
 			expect(errors).to.be.empty;
 		});
 
+		it('should validate when input value changes', async() => {
+			const elem = await fixture(html`<d2l-input-text label="label" minlength="5" value="123456"></d2l-input-text>`);
+			getInput(elem).value = '123';
+			dispatchEvent(elem, 'input', true);
+			const errors = await elem.validate();
+			expect(errors).to.contain('label must be at least 5 characters');
+		});
```
dlockhart, line 333: We weren't testing the validation logic was re-running when the user interacted with the input.

wongvincent (commented):
LGTM, thanks @dlockhart!

components/inputs/input-text.js
```diff
@@ -261,7 +267,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 
 	get validity() {
 		const elem = this.shadowRoot.querySelector('.d2l-input');
-		if (!elem.validity.valid) {
+		if (elem && !elem.validity.valid) {
```
dbatiste, line 270: I am guessing this was causing a JS error in some case?

components/inputs/input-text.js
```diff
@@ -280,6 +286,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
+		const input = this.shadowRoot.querySelector('.d2l-input');
+		if (input) {
```
dbatiste, line 290: I think this element will always exist at this point.  Just wondering if I am missing something...?

dbatiste (approved):

components/inputs/input-text.js
dlockhart: It was, but your comment made me think a bit more about it. The reason it was causing an error was that it was running earlier than before (when the property initializes and before things have rendered vs. after `updated()`).

I've adjusted it so that things should run at the same times as they did before.

components/inputs/input-text.js
dlockhart: Correct. I've changed this around so that this just calls into `_setValue` which will do this same thing and also set up the validation state the first time.

PR #1460, opened 2021-07-08
input-date: improve logic for opening direction

margaree:
This is to fix a bug where calendar would open upwards for a month with 5 rows, but then once the user navigated to a month with 6 rows, the header would no longer be visible.

I added a `min-height` to dropdown which would be used in the calculation of whether to open above or below.


components/dropdown/test/dropdown-content.visual-diff.html
```diff
@@ -119,6 +120,14 @@
 			</d2l-dropdown-content>
 		</d2l-dropdown>
 
+		<d2l-dropdown id="min-height">
+			<div class="d2l-dropdown-opener"></div>
+			<d2l-dropdown-content min-height="1000" no-auto-fit>
```
margaree, line 125: With just `no-auto-fit` this would have opened above, but by adding `min-height` this should open below.

components/inputs/input-date.js
```diff
@@ -98,7 +99,8 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 			:host([disabled]) d2l-icon {
 				opacity: 0.5;
 			}
-			.d2l-input-date-hidden-content {
+			.d2l-input-date-hidden-text,
```
margaree, line 102: `d2l-input-date-hidden-content` was just renamed

components/inputs/input-date.js
```diff
@@ -182,12 +185,19 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
 
-		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-date-hidden-content');
+		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-date-hidden-text');
 		this._hiddenContentResizeObserver = new ResizeObserver(() => {
 			const width = Math.ceil(parseFloat(getComputedStyle(hiddenContent).getPropertyValue('width')));
 			this._hiddenContentWidth = `${width}px`;
 		});
 		this._hiddenContentResizeObserver.observe(hiddenContent);
+
+		const hiddenCalendar = this.shadowRoot.querySelector('.d2l-input-date-hidden-calendar');
```
margaree, line 195: This is the same idea as with the hidden text used for calculation of text width. I did choose to remove the calendar after initial render, since if language changes the font-size heights generally seem to be the same and this doesn't seem like it would need to be recalculated.

components/inputs/input-date.js
```diff
@@ -260,6 +271,13 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 				</d2l-input-text>
 				${dropdownContent}
 			</d2l-dropdown>
+			${!this._dropdownFirstOpened ? html`<div aria-hidden="true" class="d2l-input-date-hidden-calendar">
```
margaree, line 274: I put this after rather than with the other hidden content because there are some pieces that target `d2l-calendar` and I didn't want to accidentally break anything that would be looking for the first `d2l-calendar`

components/dropdown/test/dropdown-content.visual-diff.html
dlockhart: And I guess `<d2l-dropdown-content style="min-height: 1000px">` doesn't quite work? Kind of unfortunate to have attributes that mirror existing CSS things.

margaree: Sadly it doesn't. The dropdown actually disappears when that's set which seems a bit weird.

dlockhart (approved):

PR #1468, opened 2021-07-09
fix: switch search inputs to use text input

dlockhart:
In fixing #1458 with text-inputs, Dave noticed that search inputs would also be affected and because it wasn't using `d2l-input-text` would need the same fix applied. That got me thinking... why aren't search inputs using text inputs?

I think it was either a chicken & egg situation, or initially because text inputs didn't supports the right/left slots and search needed them search did its own thing. Well that's no longer true, and switching this over yields some pretty nice code duplication savings!

### DIFF
--- a/components/inputs/input-search.js
+++ b/components/inputs/input-search.js
@@ -1,7 +1,7 @@
 import '../button/button-icon.js';
 import '../colors/colors.js';
+import './input-text.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
-import { classMap } from 'lit-html/directives/class-map.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputStyles } from './input-styles.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
@@ -42,9 +42,7 @@ class InputSearch extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			/**
 			 * Value of the input
 			 */
-			value: { type: String },
-			_focussed: { type: Boolean, attribute: false },
-			_hovered: { type: Boolean, attribute: false }
+			value: { type: String }
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


components/inputs/input-text.js
```diff
@@ -376,8 +376,8 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 						tabindex="${ifDefined(this.tabindex)}"
 						title="${ifDefined(this.title)}"
 						type="${this._getType()}">
-					<div class="d2l-input-inside-before">${this.dir === 'rtl' ? unit : ''}<slot name="${firstSlotName}" @slotchange="${this._handleSlotChange}"></slot></div>
-					<div class="d2l-input-inside-after">${this.dir !== 'rtl' ? unit : ''}<slot name="${lastSlotName}" @slotchange="${this._handleSlotChange}"></slot></div>
+					<div class="d2l-input-inside-before" @keypress="${this._suppressEvent}">${this.dir === 'rtl' ? unit : ''}<slot name="${firstSlotName}" @slotchange="${this._handleSlotChange}"></slot></div>
+					<div class="d2l-input-inside-after" @keypress="${this._suppressEvent}">${this.dir !== 'rtl' ? unit : ''}<slot name="${lastSlotName}" @slotchange="${this._handleSlotChange}"></slot></div>
```
dlockhart, line 380: One bug this change exposed is that because search input was listening for `keypress` on the input to detect when ENTER was pressed, when you focus on the search/clear buttons in the slots and hit ENTER it was interpreting that the same way. So we want to prevent these events from bubbling up.

components/inputs/input-search.js
```diff
@@ -54,54 +52,23 @@ class InputSearch extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					display: inline-block;
 					width: 100%;
 				}
-				.d2l-input-search-container {
-					position: relative;
-				}
 				:host([hidden]) {
 					display: none;
 				}
-				.d2l-input {
```
dbatiste, line 63: Nice to get id of some fo this.  🙌

dlockhart: Yeah there's a lot of stuff handling the input styling, the slotted clear/search buttons and the "showing focus styles on hover" that is removed.

components/inputs/input-text.js
dbatiste: Another way this could be handled would be to have `d2l-input-search`'s `_handleInputKeyPress` check the target and bail if it were any of its buttons.  I could see reasons for going either way.

dbatiste (approved):
Wow... this really does tidy things up nicely!  😄

components/inputs/input-text.js
dlockhart: I considered that, but figured this fix belonged more in `d2l-input-text` since consumers would be expecting `@keypress` and `@input` events to be originating from the text input, not stuff in the slots.

PR #1469, opened 2021-07-09
fix: restoring a test that was skipped due to Firefox bug

dlockhart:
The Firefox bug that caused this test to be skipped has been resolved.


dlockhart:
Hmm... I guess they lied and didn't really fix it. 👎

PR #1472, opened 2021-07-12
Focus visible for card

GeorgePlukov:
- Removes the internal `_active` property on cards. Which doesn't seem to be necessary anymore
- Adds focus visible polyfill

### DIFF
--- a/components/card/test/card.visual-diff.js
+++ b/components/card/test/card.visual-diff.js
@@ -33,7 +33,13 @@ describe('d2l-card', () => {
 		{ name: 'no-link-focus', selector: '#header-content', action: (selector) => page.$eval(selector, (elem) => elem.focus()) },
 		{ name: 'subtle', selector: '#subtle' },
 		{ name: 'link', selector: '#link' },
-		{ name: 'link-focus', selector: '#link', action: (selector) => page.$eval(selector, (elem) => elem.focus()) },
+		{ name: 'link-focus', selector: '#link', action: (selector) => page.$eval(selector, (elem) => {
+			elem.addEventListener('blur', () => {
+				elem.classList.remove('focus-visible');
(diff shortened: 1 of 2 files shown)


components/card/test/card.visual-diff.js
```diff
@@ -33,7 +33,13 @@ describe('d2l-card', () => {
...
 		{ name: 'subtle', selector: '#subtle' },
 		{ name: 'link', selector: '#link' },
-		{ name: 'link-focus', selector: '#link', action: (selector) => page.$eval(selector, (elem) => elem.focus()) },
+		{ name: 'link-focus', selector: '#link', action: (selector) => page.$eval(selector, (elem) => {
+			elem.addEventListener('blur', () => {
+				elem.classList.remove('focus-visible');
+			}, { once: true });
+			elem.classList.add('focus-visible');
+			elem.focus();
+		}) },
```
dlockhart, line 42: There's a `forceFocusVisible` helper that does exactly this (I think you might have copied this from there?). To use it from visual diff tests is a bit tedious though as you have to [import it from the HTML file](https://github.com/BrightspaceUI/core/blob/master/components/button/test/button-icon.visual-diff.html#L9) and expose it as a global on the window.

GeorgePlukov: In this case since card itself isn't focusable the `forceFocusVisible` method attempts to find a focusable child which is the `a` tag within the card. The method adds the `focus-visible` class to that instead of the host element which causes the styles to not appear properly

dlockhart: Interesting, and it looks like the polyfill puts `.focus-visible` on both the host AND the `<a>` which I wasn't quite expecting. 👍

dlockhart (approved):

PR #1474, opened 2021-07-12
Make sure ifrau dialog service has loaded before we try to check for it in iframed FRAs

grant-cleary:
This PR has two goals:
1. Make sure we're always waiting for the dialog service to be created before checking for it (presuming we're in an FRA).
2. Lazy-load the service so we only connect one ifrau client (and only create one dialog service) for all dialog components.

The root of this issue was the discovery (thanks @dbatiste for helping me debug this!) that Lit does not wait for `connectedCallback()` to finish before moving on to `updated()`. So it was possible (in fact, likely) for the check for `this._ifrauDialogService` to run before the service was actually created. This led to dialogs opening in the wrong location in FRAs if they opened too soon after first being attached to the DOM, as they would size themselves against the document.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -21,10 +21,11 @@ if (window.D2L.DialogMixin.preferNative === undefined) {
 let ifrauDialogService;
 
 async function getIfrauDialogService() {
-	if (!ifrauDialogService && window.ifrauclient) {
-		const ifrauClient = await window.ifrauclient().connect();
-		ifrauDialogService = await ifrauClient.getService('dialogWC', '0.1');
-	}
+	if (!window.ifrauclient) return;
+	if (ifrauDialogService) return ifrauDialogService;
+	
+	const ifrauClient = await window.ifrauclient().connect();
+	ifrauDialogService = await ifrauClient.getService('dialogWC', '0.1');
 
 	return ifrauDialogService;
 }
@@ -102,9 +103,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		super.updated(changedProperties);
 		if (!changedProperties.has('opened')) return;
 
-		let dialogService;
-		if (window.ifrauclient) dialogService = await getIfrauDialogService();
-
+		const dialogService = await getIfrauDialogService();
 		if (this.opened) {
 			if (dialogService) {
 				this._ifrauContextInfo = await dialogService.showBackdrop();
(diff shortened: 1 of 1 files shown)


grant-cleary:
Hmm, uncertain why what I changed would cause that test to fail. I'll need to poke at it a bit more.

grant-cleary:
Well... The failing test passed and now a new test is failing. I'm pretty confident these are unrelated, so I suppose we now play the "re-run until we win" game. I'll wait for reviews before doing that though; if I end up committing more I would rather not create extra CI jobs beforehand.

components/dialog/dialog-mixin.js
```diff
@@ -18,6 +18,17 @@ if (window.D2L.DialogMixin.preferNative === undefined) {
 	window.D2L.DialogMixin.preferNative = true;
 }
 
+let ifrauDialogService;
+
+async function getIfrauDialogService() {
+	if (ifrauDialogService === undefined && window.ifrauclient) {
```
dbatiste, line 24: This is a minor point, but I would tend to just check
```javascript
if (!ifrauDialogService && window.ifrauclient)
```
Partly out of laziness, but in this case, I would expect to treat `null` and `undefined` the same... and create a new service in such a case, so omitting the explicit check seems more resilient.

components/dialog/dialog-mixin.js
```diff
@@ -93,14 +101,18 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 	async updated(changedProperties) {
 		super.updated(changedProperties);
 		if (!changedProperties.has('opened')) return;
+
+		let dialogService;
```
dbatiste, line 105: Could this just be...
```javascript
const dialogService = await getIfrauDialogService();
if (this.opened) {
    if (dialogService) {
        ...
    }
    this._open();
} else {
    if (dialogService) {
        ...
    }
    this._close();

}
```

dbatiste: We probably would need to update the helper to include.
```javascript
if (!window.ifrauclient) return;
```

grant-cleary: The helper already checks `window.ifrauclient`, so we're good there, but I think we might want to switch the order of the checks just in case (somehow) the service gets set without ifrauclient loaded.

grant-cleary: Ah, but then if it gets set without ifrauclient loaded, it would have some value so we can't just return it.

dbatiste (approved):

PR #1475, opened 2021-07-13
Add Ability to override dropdown mobile breakpoint, use 767px breakpoint for filter

annabelsegalld2l:
According to https://xd.adobe.com/view/819a26ff-a0ba-4f8e-b405-ac677da84804-d4c9/screen/37f0464b-51e6-4072-aeef-b0b6331f6d50?fullscreen, we want the mobile tray to appear at viewports <768px. We currently have the breakpoint set to 616px (the mobile typography breakpoint). I found a way to override the breakpoint even though CSS media queries do not support variables. I also applied the mobile tray in the filter component. 

Here's a GIF of the wider screen with the tray. The width is still capped at 420px so it doesn't look super wide.
filter_tray


components/dropdown/dropdown-content-mixin.js
```diff
@@ -197,6 +204,8 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this.noPaddingFooter = false;
 		this.noPaddingHeader = false;
 		this.noPointer = false;
+		this.mobileBreakpointOverride = 616;
```
annabelsegalld2l, line 207: Setting the default to 616px. If mobile-breakpoint is never specified, it will use the default value.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -801,6 +812,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 					</div>
 				</div>
 			</div>
+			<style>
+				${dropdownContentStyles(this.mobileBreakpointOverride)}
```
annabelsegalld2l, line 816: Adding the styles using the <style> attribute on each render. We need to load it for each render rather than using the static styles() function (shared among all instances).

components/dropdown/dropdown-content-styles.js
```diff
@@ -1,285 +1,288 @@
...
+			--d2l-dropdown-above-animation-name: d2l-dropdown-above-animation-dark;
+			--d2l-dropdown-animation-name: d2l-dropdown-animation-dark;
+			--d2l-dropdown-background-color: #333536; /* ferrite @ 70% */
+			--d2l-dropdown-border-color: var(--d2l-color-ferrite);
+			--d2l-dropdown-foreground-color: var(--d2l-color-sylvite);
+			--d2l-dropdown-shadow-color: rgba(0, 0, 0, 1);
+			opacity: 0.9;
+		}
+
+		@media (min-width: ${breakpoint}px) {
```
annabelsegalld2l, line 36: We sub in the provided breakpoint in these media queries, rather than a css variable. Note that I didn't change any content here, even though the diff looks pretty crazy...

components/dropdown/dropdown-content.js
```diff
@@ -13,10 +12,6 @@ import { dropdownContentStyles } from './dropdown-content-styles.js';
  */
 class DropdownContent extends DropdownContentMixin(LitElement) {
 
-	static get styles() {
```
annabelsegalld2l, line 16: We can delete the styles from the classes that inherit from the mixin, since the styles are already defined in there.

components/filter/demo/filter.html
```diff
@@ -11,7 +11,7 @@
 		import '../../filter/filter-dimension-set.js';
 		import '../../filter/filter-dimension-set-value.js';
 	</script>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1.0">
```
annabelsegalld2l, line 14: I had to add that minimum scale because mobile emulators weren't working properly with the filter demo page (ie. there was a horizontal scroll bar being shown rather than using the correct scaling)

components/dropdown/dropdown-content-mixin.js
dlockhart: The downside to this is that browser who supported constructible stylesheets won't be able to reuse these styles across all dropdowns.

An alternate but similar approach that might work:

Instead of using a media query in `dropdown-content-styles` (i.e. `@media (min-width: 616px) {`), could you use the presence of a CSS class or `data-` attribute on the host to trigger the mobile styles to kick in?

For example:

```css
:host([data-mobile][opened]) {
    animation: var(--d2l-dropdown-animation-name) 300ms ease;
    display: inline-block;
}
```

And then you could use the `mediaQueryList` to detect when things are a match, but you'd need to add an event listener on it like we did in fullscreen dialogs to be notified when it changes.

annabelsegalld2l: Thanks Dave. I'll try that!

dbatiste: Thanks Dave.  Yeah this is the route that I was thinking as well.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -205,6 +216,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this.noPaddingFooter = false;
 		this.noPaddingHeader = false;
 		this.noPointer = false;
+		this.mobileBreakpointOverride = 616;
+		this.mediaQueryList = window.matchMedia(`(max-width: ${this.mobileBreakpointOverride - 1}px)`);
```
dlockhart, line 220: Does this need to be set here, or since it's getting set again in `connectedCallback` is that enough?

components/dropdown/dropdown-content-styles.js
```diff
@@ -174,7 +170,7 @@ export const dropdownContentStyles = css`
 
 	@media (prefers-reduced-motion: reduce) {
 		:host([opened]), :host([opened-above]) {
-			animation: none;
+			animation: none !important;
```
dlockhart, line 173: Nit: now that legacy-Edge support is dropping, we could actually start inverting these so that we only apply the animation if `prefers-reduced-motion: no-preference` is true. Would avoid an `!important` here.

dlockhart (approved):

PR #1480, opened 2021-07-13
fix: DE44358: Force dialog content to fit rather than horizontal scrolling

annabelsegalld2l:
This issue was discovered in the MyCourses widget, where a long list of tabs was rendering without the overflow button. The fullscreen dialog currently allows its content to be as wide as it wants. This change will force the content to a fixed width, forcing an overflow button. 
See: [DE44358](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?Iteration=%2Fiteration%2F450584641844&Name=saddsfd&Project=%2Fproject%2F15545167705&Release=u&cpoid=15545167705&detail=%2Fdefect%2F603823819031&fdp=true&fdp=true&rankScope=%2Fiteration%2F450584641844&rankTo=BOTTOM&typeDef=7362609683&fdp=true?fdp=true): [Fullscreen Dialog] View All Courses Dialog in MCW Expands the Course Cards

Here's an example of the change in FACE rubrics. On small screens, a scrollbar used to appear. Now, the overflow arrow will appear to indicate more content.
faceui_scroll

faceui_overflow

NOTE: The least hacky solution required removing the padding from the bottom of the content. The content should still be visible, just without the extra padding at the end. This is the reason for the two scroll-top-shadow visual diff failures.


components/dialog/demo/dialog-fullscreen.html
```diff
@@ -42,6 +44,66 @@ <h2>Fullscreen Dialog</h2>
...
+				<template>
+					<d2l-button id="openOverflow">Show Dialog</d2l-button>
+					<d2l-dropdown-more text="Open!">
+						<d2l-dropdown-menu>
+							<d2l-menu label="Dialog">
+								<d2l-menu-item id="dropdownOverflow" text="See dialog"></d2l-menu-item>
+							</d2l-menu>
+						</d2l-dropdown-menu>
+					</d2l-dropdown-more>
+					<d2l-dialog-fullscreen id="dialogFullscreenOverflow" title-text="Fullscreen Title">
```
annabelsegalld2l, line 82: Added another demo here with a long tab list and overflowing text.

components/dialog/dialog-fullscreen.js
```diff
@@ -53,17 +53,18 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
+					padding-bottom: 21px;
+					padding-top: 1rem;
 				}
 
 				.d2l-dialog-content > div {
 					/* required to properly calculate preferred height when there are bottom
 					margins at the end of the slotted content */
 					border-bottom: 1px solid transparent;
 					box-sizing: border-box;
+					display: contents;
```
annabelsegalld2l, line 66: Removing display: grid and adding display: contents here allowed the content width to be fixed. It seems like everything is working fine with this, but I'll need to do some regression testing within the lms to make sure.

components/dialog/demo/dialog-fullscreen.html
dlockhart: Would adding this as a visual diff test instead (or in addition) maybe help us catch a regression to this in the future?

annabelsegalld2l: We do have tests for the vertical overflow. I'll add one with the tabs to make sure we have horizontal overflow as well.

components/dialog/dialog-fullscreen.js
dlockhart: Something worrying from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/display):

> Due to a bug in browsers this will currently remove the element from the accessibility tree — screen readers will not look at what's inside.

The [cross-browser support](https://caniuse.com/css-display-contents) is also a little sketchy... specifically legacy-Edge doesn't support it and there are some scary yellow yellow boxes for Safari and relatively recent Chrome/Firefox.

annabelsegalld2l: That's not great 😢 I'll keep looking for another solution. I'm having trouble getting the content to stay a fixed width, and keeping the top/bottom padding.

dlockhart (approved):

PR #1486, opened 2021-07-15
fix: introducing a basic polyfill for escaping CSS

dlockhart:
In the LMS, we use `$` symbols to split up scopes in relative IDs, like `data$HomePageIsShared`. Unfortunately, while valid in HTML IDs, the dollar symbol (and others) are not valid in calls to `elem.querySelector()`. 🤦 

The native API to do this is called `CSS.escape` and it's [supported almost everywhere](https://caniuse.com/mdn-api_css_escape) we care about... I'd be fine to use it except for legacy-Edge right now. So in a couple months we can likely clean this up and just call it directly.


helpers/dom.js
```diff
@@ -1,3 +1,11 @@
+export function cssEscape(val) {
+	if (window.CSS && window.CSS.escape) {
+		return window.CSS.escape(val);
+	}
+	val = val.replace(/\$/g, '\\$');
```
dlockhart, line 5: There's a [real polyfill](https://github.com/mathiasbynens/CSS.escape) for this, but since this is really just a temporary thing unti legacy-Edge is gone in a couple months, I think we're good with a specific fix for dollar signs.

svanherk:
> So in a couple months we can likely clean this up and just call it directly.

Maybe add a note about that in the `cssEscape` function so we don't forget? `// Needed for Legacy Edge, after just use directly`

jordankay13 (approved):

svanherk (approved):

PR #1488, opened 2021-07-15
Component analyzer: Switch to @custom-elements-manifest/analyzer

margaree:
I actually don't think we should make the switch yet as I have a few times run the analyzer and noticed that there was at least one missing component, and we can't have that. I'll work on a way to figure out when that's happening. There also still seems to be a lot of active development going on.

If we don't want to make the switch over yet, the changes here do still work with the web-component-analyzer, and I've been able to make the changes necessary to documentation in such a way that it supports both formats, so we can be updating the jsdoc comments as we go in hopes of using this analyzer while still having Daylight work.

### DIFF
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -20,17 +20,18 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 			/**
 			 * Aligns the leading edge of text if value is set to "text"
 			 * @type {'text'|''}
-			 * @default ""
 			 */
 			hAlign: { type: String, reflect: true, attribute: 'h-align' },
 
 			/**
 			 * REQUIRED: Preset icon key (e.g. "tier1:gear")
+			 * @type {string}
 			 */
 			icon: { type: String, reflect: true },
 
 			/**
 			 * REQUIRED: Accessible text for the button
+			 * @type {string}
 			 */
 			text: { type: String, reflect: true },
 
@@ -141,6 +142,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 	constructor() {
 		super();
 		this.translucent = false;
+
+		/** @internal */
 		this._buttonId = getUniqueId();
 	}
 
--- a/components/button/button-mixin.js
+++ b/components/button/button-mixin.js
@@ -77,10 +77,10 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 	/**
 	 * @attr disabled - Disables the button
 	 */
-	get disabled() { return this.__disabled; }
+	get disabled() { return this._disabled; }
 	set disabled(value) {
-		const oldValue = this.__disabled;
-		this.__disabled = value;
+		const oldValue = this._disabled;
+		this._disabled = value;
 		this.requestUpdate('disabled', oldValue);
 	}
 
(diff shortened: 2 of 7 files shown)


components/button/button-icon.js
```diff
@@ -20,17 +20,18 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 			/**
 			 * Aligns the leading edge of text if value is set to "text"
 			 * @type {'text'|''}
-			 * @default ""
```
margaree, line 23: `@default` doesn't seem to work. It more just looks at if there's an initial value set in the constructor. This case is weird anyways since the default isn't really `""`, it's more undefined.

components/button/button-icon.js
```diff
@@ -20,17 +20,18 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
...
 			 */
 			hAlign: { type: String, reflect: true, attribute: 'h-align' },
 
 			/**
 			 * REQUIRED: Preset icon key (e.g. "tier1:gear")
+			 * @type {string}
```
margaree, line 28: `type` will only get to string in `custom-elements.json` if the value is actually initialized in the constructor, otherwise we need to set `@type` here

components/button/button-icon.js
```diff
@@ -141,6 +142,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 	constructor() {
 		super();
 		this.translucent = false;
+
+		/** @internal */
```
margaree, line 146: Internal variables get added to `custom-elements.json` without this.

components/button/button-mixin.js
```diff
@@ -65,12 +65,36 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 
 	constructor() {
 		super();
-		this.autofocus = false;
 		this.disabled = false;
-		this.primary = false;
```
margaree, line 70: I missed moving this to be just in `button.js` in the previous PR.

components/button/button-mixin.js
```diff
@@ -65,12 +65,36 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
...
+
+		/** @ignore */
+		this.autofocus = false;
+
+		/** @ignore */
 		this.type = 'button';
 	}
 
+	/**
```
margaree, line 77: So this is the most annoying part. This is the only way I was able to get mixin attributes to "work" and show up in `custom-elements.json` with all the necessary pieces of information (e.g., default value, description, actual attribute name). `disabled` technically doesn't need this, it also works to just add the description above the `this.disabled = false` line in the constructor, but I thought I'd include both for consistency. If we did that though I believe it doesn't get included in the `attributes` array in `custom-elements.json`, which we aren't currently using (there's another array called members that is more detailed that I picked to use instead), but if we ever wanted to switch then we might want to not have to worry about some attributes not being present.

dlockhart: That's... unfortunate. I wonder if we were using the TypeScript syntax for components if it'd pick those up better...

dlockhart (commented):
Boo, seems like a fair number of workarounds that weren't necessary before... I wonder why they made those changes. Probably it was built for generic custom elements where you need setters/getters anyway and the Lit plugin needs some work.

components/button/button-mixin.js
margaree: Based on some examples I was looking at I think that it would.

dlockhart (approved):

GeorgePlukov (approved):

PR #1490, opened 2021-07-16
Revert "Focus visible for card"

GeorgePlukov:
Reverts BrightspaceUI/core#1472

The card changes aren't working in the lms as they are within our demo page. Since branching is occurring today thought It would be best to revert this while working on a fix.


dlockhart (approved):
Makes sense!

Issue #1406, opened 2021-06-16
[visual-diff] Overflow-group tests: re-enable failing tests


margaree:
With the new version of puppeteer the tests `more-than-max-to-show` and `all-item-types` were failing. I found that when I commented those tests out `auto-show-small` also began to fail so it is also commented out. Get these tests working again. They were commented out in this PR https://github.com/BrightspaceUI/core/pull/1402.

<img width="1230" alt="Screen Shot 2021-06-16 at 3 01 59 PM" src="https://user-images.githubusercontent.com/6804600/122277494-f59b5e00-ceb3-11eb-8b6b-d2ef0fdc9586.png">
<img width="1236" alt="Screen Shot 2021-06-16 at 3 02 07 PM" src="https://user-images.githubusercontent.com/6804600/122277497-f633f480-ceb3-11eb-912d-a9d89ea54ad9.png">

PR #1497, opened 2021-07-19
Try "captureBeyondViewport: false" fix on overflow-group tests

svanherk:
Noticed that this overflow-group test seems to be flakey: https://github.com/BrightspaceUI/core/pull/1492.  Verified it had the same page resizes going on as the `switch` tests.  I'm applying the `captureBeyondViewport: false` fix, which led to a few other adjustments. I also tried to fix up the other tests that were turned off, which fixes https://github.com/BrightspaceUI/core/issues/1406.

I re-ran ci for this PR 5 times and the overflow tests didn't fail (just an `input-date` one).


components/overflow-group/test/overflow-group.visual-diff.js
```diff
@@ -84,32 +87,30 @@ describe('d2l-overflow-group', () => {
...
 		...autoShow,
 		...iconType,
 	].forEach((test) => {
 		it(test.name, async function() {
 			const selector = `#${test.name}`;
 			const containerSelector = `#${test.name}-container`;
-			await page.evaluate(() => {
-				return new Promise(resolve => setTimeout(resolve, 0));
-			});
+			await page.$eval(containerSelector || selector, (elem) => elem.scrollIntoView());
```
svanherk, line 108: Because we are no longer capturing beyond the viewport (i.e. no longer resizing the page randomly for every test), we need to make sure we've scrolled the test into view before we take the screenshot.

components/overflow-group/test/overflow-group.visual-diff.js
```diff
@@ -69,6 +69,9 @@ describe('d2l-overflow-group', () => {
 	const iconType = [
 		{
 			name: 'opener-type-mini-menu',
+			action: async() => {
+				await page.mouse.move(0, 0);
```
svanherk, line 73: This was a weird one... the mouse was randomly hovering over the third button for this test, I think because we click the third button in another test and then scroll to this one.  This fixes the problem, but I wasn't sure if there's a better way.

svanherk:
@GeorgePlukov @margaree Do these screenshots look as you'd expect?  I tried to match the goldens from https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/overflow-group/2021-06-16-18.05.59.136-9180c6eac16d4a926d20af00c9430fb1a79fec44-report.html before they were turned off... the only difference seems to be hovering on teh first item in teh dropdown, which I think was another mouse placement coincidence

components/overflow-group/test/overflow-group.visual-diff.js
dlockhart: Crazy! Could put a helper in the core visual-diff library like our resetFocus one, but probably worth waiting to see if the pops up in more places.

components/overflow-group/test/overflow-group.visual-diff.js
dlockhart: Alternately, I think we used to just set the viewport to be super tall before this `captureBeyondViewport` was a thing.

dlockhart (approved):
Awesome!

components/overflow-group/test/overflow-group.visual-diff.js
svanherk: Ah, that does sounds familiar!  Confirmed that also works, do you have thoughts on which is better?  Seems like the large viewport is a _tad_ faster for the test that needs to scroll, but this way means you don't need to increase the viewport if you add tests.  The mouse placement issue is only an issue in this case, not the large viewport case (cause it doesn't scroll to a specific place and move the mouse with it)

margaree (approved):
Nice! The screenshots LGTM

components/overflow-group/test/overflow-group.visual-diff.js
dlockhart: I don't have feelings either way, no.

svanherk: Gonna leave it like this for now since that's the way I tested it, and we'll see if it's stable or still has issues

GeorgePlukov:
Sorry I think I got here a couple minutes late but the one screenshot for `d2l-overflow-group more-than-max-to-show` looks like its missing the down arrow on the "More Actions" menu. That was being flakey at one point in the past and I think I added a timeout so that the icon would load before the tests were taken

svanherk:
> Sorry I think I got here a couple minutes late but the one screenshot for d2l-overflow-group more-than-max-to-show looks like its missing the down arrow on the "More Actions" menu. That was being flakey at one point in the past and I think I added a timeout so that the icon would load before the tests were taken

I think it's there, no?  The report I linked above was to show what I was using as a reference to the "good" goldens since they'd since been removed, here's the actual report for this PR: https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/overflow-group/2021-07-19-15.50.42.751-a11645e9fb7af31da6c943fae62d4104bcf80457-report.html

GeorgePlukov:
My bad! I think those look all good, misunderstood what you said in your previous comment!

svanherk:
> My bad! I think those look all good, misunderstood what you said in your previous comment!

No problem, I definitely could have worded it better!

PR #1503, opened 2021-07-20
Daylight: date & time inputs

margaree:
Changes:
- prep `input-date-time.md` for Daylight
- component file tweaks to wording to get things like tense consistent
- `@ignore` for some properties coming from mixins, and some events (this is also affecting the old analyzer)

### DIFF
--- a/components/inputs/docs/input-date-time.md
+++ b/components/inputs/docs/input-date-time.md
@@ -9,7 +9,6 @@ size:xlarge
 <script type="module">
   import '@brightspace-ui/core/components/inputs/input-date.js';
   import '@brightspace-ui/core/components/inputs/input-time.js';
-  import '@brightspace-ui/core/components/inputs/input-date-time.js';
 </script>
 <d2l-input-date label="Date Input"></d2l-input-date>
 <d2l-input-time label="Time Input"></d2l-input-time>
@@ -29,8 +28,8 @@ Note: All `*value` properties should be in ISO 8601 calendar date format (`YYYY-
 name:d2l-input-date
 align:flex-start
 autoSize:false
-size:xlarge
 defaults:{ "label": "Birthdate" }
+size:xlarge
 -->
 ```html
 <script type="module">
@@ -82,8 +81,8 @@ Note: All `*value` properties should be in ISO 8601 calendar date format (`YYYY-
 name:d2l-input-date-range
 align:flex-start
 autoSize:false
-size:xlarge
 defaults:{ "label": "Availability Range" }
+size:xlarge
 -->
 ```html
 <script type="module">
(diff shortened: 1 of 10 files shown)


components/inputs/docs/input-date-time.md
```diff
@@ -1,24 +1,46 @@
...
+size:xlarge
+-->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/inputs/input-date.js';
 </script>
-<d2l-input-date
-  label="Start Date"
-  value="2020-11-20">
+<d2l-input-date>
```
dlockhart, line 38: It's unfortunate that when viewing on GitHub the inline code example isn't going to contain any of the attributes you need to get it to work.

Would it be possible for the docs site to just ignore any attributes that are set here? Probably not easy eh.

dlockhart: Or I guess even better: use the attributes provided here as the defaults so that `defaults: { "label": "Birthdate" }` didn't need to be passed in as part of the metadata.

components/inputs/input-date-range.js
```diff
@@ -47,45 +47,52 @@ class InputDateRange extends SkeletonMixin(FormElementMixin(RtlMixin(LocalizeCor
...
 			minValue: { attribute: 'min-value', reflect: true, type: String },
 			/**
-			 * Indicates that a value is required
+			 * Indicates that values are required
 			 */
 			required: { type: Boolean, reflect: true },
 			/**
-			 * Label for the start date input
+			 * Accessible label for the start date input. Defaults to localized "Start Date".
+			 * @type {string}
```
dlockhart, line 89: For later: I think it'd be useful to have a step-by-step guide somewhere in the Daylight Site on how to properly document a component so that everything shows up properly. When to include `@type`, how to mark something as required, show the default value, ignore stuff, hide stuff from the Daylight site, and so on.

components/inputs/input-date.js
```diff
@@ -362,6 +367,7 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		if (this._calendar) this._calendar.reset();
 		this._dropdownOpened = false;
 		this._textInput.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
+		/** @ignore */
```
dlockhart, line 370: Cool (and crazy?) that it finds these!

dlockhart (approved):

components/inputs/docs/input-date-time.md
margaree: I'll take a look at that. It's a bit more difficult since they're initially just strings and not code we can pull the attributes out of, but there's likely some regex we could use.

components/inputs/input-date-range.js
margaree: For sure!

PR #1504, opened 2021-07-20
feat: US129214 Add bottom tray option to mobile dropdown 

annabelsegalld2l:
This is a copy of https://github.com/BrightspaceUI/core/pull/1446. Had to delete that branch this the merge with the filter PR messed up some css.

Design : https://xd.adobe.com/view/5205ec02-86e7-48b0-ad0a-04aa76db7759-0756/?fullscreen
Visual diff changes: https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/dropdown-content/2021-06-29-21.07.31.34-8f2920715bef3d77af598bfa71e63e51314e161f-report.html

Added a mobile-tray="bottom" option to slide in from the bottom of the screen.
bottom_tray

In addition to adding the dropdown, I had some other fixes for all the mobile trays:

Fixing RTL behaviour to make sure the 'close' button is centered on the tray
Adding the footer line, even if there is no footer content (other than the close button)
If there is a close button:
with NO other footer contents: pad 12px around as in design
WITH other footer contents: pad top of close button with 12px, but adhere to dropdown style of 20px footer padding


components/dropdown/dropdown-content-styles.js
```diff
@@ -237,35 +258,47 @@ export const dropdownContentStyles = css`
...
+
 	:host([data-mobile][mobile-tray="left"][opened]) > .d2l-dropdown-content-position > .d2l-dropdown-content-width > .d2l-dropdown-content-container,
 	:host([data-mobile][mobile-tray="right"][opened]) > .d2l-dropdown-content-position > .d2l-dropdown-content-width > .d2l-dropdown-content-container {
 		height: 100vh;
 	}
 
 	:host([data-mobile][mobile-tray="left"]) > .d2l-dropdown-content-pointer,
-	:host([data-mobile][mobile-tray="right"]) > .d2l-dropdown-content-pointer {
+	:host([data-mobile][mobile-tray="right"]) > .d2l-dropdown-content-pointer,
+	:host([data-mobile][mobile-tray="bottom"]) > .d2l-dropdown-content-pointer {
```
dlockhart, line 273: If the value of `mobile-tray` doesn't matter, you could rewrite a lot of these to be:

```css
:host([data-mobile][mobile-tray]) > .d2l-dropdown-content-pointer {
}
```

Hmm, unless that causes the specificity of the CSS rule to be lower and therefore not take precedence?

dlockhart (approved):

components/dropdown/dropdown-content-mixin.js
```diff
@@ -759,6 +763,10 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			}
 		}
 
+		if (mobileTrayBottom) {
+			maxWidthOverride = '100vw';
+		}
```
dlockhart, line 768: Nit: this is getting a little bit unwieldily with all the `if (mobileTrayThis) { .. }` spread across the next ~100 lines or so. There might be an opportunity to refactor it a bit, or maybe move some of it out into helper functions.

It's tricky because it's nice to have things laid out like you do where you declare a variable (e.g. `widthOverride`) and then right after you calculate it with an `if - elseif - else`. But on the flip side it'd be a lot less code to just have `if (mobileTrayBottom) { thing1 = whatever; thing2 = that; } else if (mobileTrayLeft) { thing1 = that; }`.

Anyway, it's understandable how it is... just getting a bit complex and potentially fragile. :)

annabelsegalld2l: Yep, totally agree. I'll work on refactoring it this afternoon.

dlockhart (approved):
LGTM 💯

components/dropdown/dropdown-content-mixin.js
dlockhart: Looks a lot better now!

PR #1506, opened 2021-07-21
Daylight: Menu

margaree:


GeorgePlukov (approved):
LGTM! Didn't notice anything out of place, it does really suck that we have to ignore properties both in the declaration and constructor though

PR #1509, opened 2021-07-21
Daylight: Switch

GeorgePlukov:


components/switch/switch-mixin.js
```diff
@@ -194,6 +215,9 @@ export const SwitchMixin = superclass => class extends RtlMixin(FocusVisiblePoly
 	updated(changedProperties) {
 		super.updated(changedProperties);
 		if (!changedProperties.has('on') || changedProperties.get('on') === undefined) return;
+		/**
+		 *  Dispatched when there is a change to the switch value. "on" corresponds to the current state of the switch.
```
GeorgePlukov, line 219: Documented this in line as doing it at the top of the file didn't seem to pass it through, might be due to the fact that this is a mixin. Doing it like this makes it appear in both `switch` and `switch-visibility` variants

components/switch/README.md
```diff
@@ -25,25 +35,31 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 | `on` | Boolean | Whether the switch is "on" or "off" |
 | `text-position` | String | Valid values are: `start`, `end` (default), and `hidden` |
 | `tooltip` | String | Text to display in a tooltip for the switch |
+## Events
```
margaree, line 38: This should also be an h3

components/switch/README.md
```diff
@@ -25,25 +35,31 @@ The `d2l-switch` element is a generic switch with on/off semantics.
...
 | Attribute | Description |
 |--|--|
 | text | **REQUIRED** [Acts as a primary label on the switch](https://www.w3.org/WAI/tutorials/forms/labels/). Visible unless text-position is `hidden`. |
 
-**Events:**
-
-- `change`: dispatched when the `on` property is updated
-
-## d2l-switch-visibility
+## Switch Visibility [d2l-switch-visibility]
```
margaree, line 51: Probably want to call this "Visibility Switch"

components/switch/README.md
```diff
@@ -2,21 +2,31 @@
...
 <script type="module">
   import '@brightspace-ui/core/components/switch/switch.js';
+  import '@brightspace-ui/core/components/switch/switch-visibility.js';
 </script>
-
-<d2l-switch text="Dark Mode" on></d2l-switch>
+<d2l-switch></d2l-switch>
 ```
+![Switch](./screenshots/switch.png?raw=true)
```
margaree, line 27: I think for consistency we'd want to leave this where it was above the demo and then wrap i the hidden comments there.

components/switch/README.md
```diff
@@ -52,6 +68,7 @@ The `d2l-switch-visibility` element is a variant of the generic switch configure
 <d2l-switch-visibility></d2l-switch-visibility>
 ```
 
 **Properties:**
```
margaree, line 72: This and events should be switched to h3s

components/switch/README.md
GeorgePlukov: Okay cool, I was just trying to conglomerate some of the hidden items but makes sense to keep a consistent format

components/switch/README.md
```diff
@@ -2,21 +2,34 @@
...
+  import '@brightspace-ui/core/components/switch/switch-visibility.js';
 </script>
-
-<d2l-switch text="Dark Mode" on></d2l-switch>
+<d2l-switch></d2l-switch>
 ```
 
-**Properties:**
+
```
margaree, line 31: nit: extra line here

components/switch/README.md
```diff
@@ -2,21 +2,34 @@
...
 
+<!-- docs: demo live name:d2l-switch -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/switch/switch.js';
+  import '@brightspace-ui/core/components/switch/switch-visibility.js';
 </script>
-
-<d2l-switch text="Dark Mode" on></d2l-switch>
+<d2l-switch></d2l-switch>
```
margaree, line 27: Should this demo also have `text`?

margaree (approved):

PR #1511, opened 2021-07-21
fix: US130144: dialogs getting cut off in iframes due to incorrect height settings

annabelsegalld2l:
A couple of the dialog styles were overriding dialog-mixin and adding a height attribute. The problem occurs with iFrames, where a 100vh or 100% is interpreted as the height of the iFrame, rather than the viewport. This was causing super tall dialogs with the footer cut off and buttons inaccessible. 

This PR adds an override to the height of d2l-dialog-inner with the calculated height when given by the ifrauContext (minus required padding). The d2l-dialog-outer height setting is removed. 
When the dialog is NOT in an iframe, the d2l-dialog-outer height is set as previously, with no override for d2l-dialog-inner. Unfortunately, I had to add an attribute, in-iframe, for the CSS to style based on whether or not its in an iframe. Open to naming suggestions here or a way to not have to use an attribute. 

Here is a photo of the issue happening on a test page: 
image+(3)

Here is the fixed version: 
fixed_dialog_iframe

### DIFF
--- a/components/dialog/dialog-fullscreen.js
+++ b/components/dialog/dialog-fullscreen.js
@@ -29,7 +29,7 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 			async: { type: Boolean },
 			_hasFooterContent: { type: Boolean, attribute: false },
 			_icon: { type: String, attribute: false },
-			_headerStyle: { type: String, attribute: false },
+			_headerStyle: { type: String, attribute: false }
 		};
 	}
 
--- a/components/dialog/dialog-fullscreen.js
+++ b/components/dialog/dialog-fullscreen.js
@@ -76,7 +76,6 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 				dialog.d2l-dialog-outer,
 				div.d2l-dialog-outer {
 					border-radius: 8px;
-					height: calc(100% - 3rem);
 					margin: 1.5rem;
 					max-width: 1170px;
 					opacity: 0;
@@ -163,9 +162,7 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 				
 				dialog.d2l-dialog-outer,
 				div.d2l-dialog-outer {
-					height: calc(var(--d2l-vh, 1vh) * 100 - 42px);
 					margin: 0 !important;
-					min-height: calc(var(--d2l-vh, 1vh) * 100 - 42px);
 					min-width: calc(var(--d2l-vw, 1vw) * 100);
 					top: 42px;
 				}
@@ -200,6 +197,19 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 
 	render() {
 
+		const heightOverride = {} ;
+		if (this._ifrauContextInfo) {
+			// in iframes, use calculated available height from dialog mixin minus padding
+			heightOverride.height = mediaQueryList.matches
+				? `${this._ifrauContextInfo.availableHeight - 42}px`
+				: `${this._ifrauContextInfo.availableHeight - 60}px`;
+		} else {
+			heightOverride.height = mediaQueryList.matches
+				? 'calc(var(--d2l-vh, 1vh) * 100 - 42px)'
+				: 'calc(100% - 3rem)';
+		}
+		heightOverride.minHeight = heightOverride.height;
+
 		let loading = null;
 		const slotStyles = {};
 		if (this.async && this.asyncState !== asyncStates.complete) {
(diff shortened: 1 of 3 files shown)


components/dialog/dialog.js
```diff
@@ -164,6 +193,10 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 		this._hasFooterContent = (footerContent && footerContent.length > 0);
 	}
 
+	_handleResize() {
+		this._autoSize = !mediaQueryList.matches;
```
annabelsegalld2l, line 197: We need to turn off the autosize attribute to get the dialog to show up as a "fullscreen" in mobile mode. Otherwise, it will just fit the content and not extend to the bottom of the screen as required. The code for the mediaQueryList and listener is the same as d2l-dialog-fullscreen. I also made sure to keep the check for the addEventListener function to avoid that Safari 13 issue with compatibility.

components/dialog/dialog.js
```diff
@@ -128,7 +157,7 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 
 		if (!this._titleId) this._titleId = getUniqueId();
 		const inner = html`
-			<div class="d2l-dialog-inner">
+			<div class="d2l-dialog-inner"  style=${styleMap(heightOverride)}>
```
annabelsegalld2l, line 160: I had to override the dialog-inner class rather than the outer class because the d2l-dialog-outer class is controlled by the dialogMixin. I didn't want to add the height override there, as the mixin would then have to know about the fullscreen-specific styling and dialog mobile styles. I decided to keep the mixin generic and keep the component-specific styling here.

annabelsegalld2l:
Looking into the failed dialog ifrau visual diff test. It seems to be passing on my machine. I have a feeling it might be a race condition issue where the dialog is still resizing when the screenshot is taken.

components/dialog/dialog-fullscreen.js
```diff
@@ -200,6 +210,16 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 
 	render() {
 
+		const heightOverride = {} ;
+		if (this._ifrauContextInfo) {
+			this._inIframe = true;
```
dlockhart, line 215: It's not great to be setting properties from inside `render()`, since this is going to trigger an unnecessary re-render afterwards.

One option would be to move the `_inIframe` into the dialog mixin and set it to `true` when `_ifrauContextInfo` gets set. That works pretty well actually since the normal dialog needs it as well and the property is duplicated there.

dlockhart: Oh and actually a good side effect of that will be that when the `_ifrauContextInfo` connects, it'll set `_inIframe` to `true` which will cause a re-render. As it stands now, technically I think it could connect at a random time and nothing would cause a re-render for these overrides to be set.

svanherk:
> Looking into the failed dialog ifrau visual diff test. It seems to be passing on my machine. I have a feeling it might be a race condition issue where the dialog is still resizing when the screenshot is taken.

This may be the same resize issue we've been seeing.  If you add this to the visual-diff html file:
```
	<script type="module">
		window.addEventListener('resize', () => {
			console.log('resize', document.documentElement.clientWidth); // or window.innerWidth
		});
	</script>
```

And this to the JS file:
```
const delay = async(page) => {
	await page.evaluate(() => {
		return new Promise(resolve => {
			setTimeout(resolve, 2000);
		});
	});
};
```

And then throw `await delay(page);` before your test takes the screenshot.  Do you see lots of resizing to 0?

(The delay helps make the race condition happen a lot more often)

svanherk:
Oh, and this or you won't actually see the console errors:
```
const page = await browser.newPage();
page.on('console', msg => console.log(msg.text()));
```

annabelsegalld2l:
> Oh, and this or you won't actually see the console errors:
> 
> ```
> const page = await browser.newPage();
> page.on('console', msg => console.log(msg.text()));
> ```

Thanks Stacey, this helped me figure out the real problem. No race conditions, just needed to call the resize function once autoSize gets turned on, otherwise the dialog does not re-render.

components/dialog/dialog-fullscreen.js
```diff
@@ -29,7 +29,7 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 			async: { type: Boolean },
 			_hasFooterContent: { type: Boolean, attribute: false },
 			_icon: { type: String, attribute: false },
-			_headerStyle: { type: String, attribute: false }
+			_headerStyle: { type: String, attribute: false },
```
dlockhart, line 32:
```suggestion
			_headerStyle: { type: String, attribute: false }
```

components/dialog/dialog.js
```diff
@@ -31,7 +33,7 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 			 * The preferred width (unit-less) for the dialog
 			 */
 			width: { type: Number },
-			_hasFooterContent: { type: Boolean, attribute: false }
+			_hasFooterContent: { type: Boolean, attribute: false },
```
dlockhart, line 36:
```suggestion
			_hasFooterContent: { type: Boolean, attribute: false }
```

dlockhart (approved):

PR #1513, opened 2021-07-21
Daylight: loading spinner

GeorgePlukov:


components/loading-spinner/README.md
```diff
@@ -1,18 +1,22 @@
...
 <script type="module">
   import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 </script>
 
 <d2l-loading-spinner></d2l-loading-spinner>
 ```
 
-**Properties:**
+### Properties:
```
margaree, line 19: Hmm so our properties table defaults to properties being an h3, but in this case it should actually be an h2 since `Loading Spinner` is an h1. Can either restructure the document for something like

```
h1: Loading Spinner
blurb about loading spinner
h2: Loading Spinner [d2l-loading-spinner]
other technical stuff
h3 Properties
```
or we could add another property on the comment to be something like `<!-- docs: demo live tableHeadingLevel:2 ... `.

GeorgePlukov: I discussed this with Jeff this morning and looks like we are going to try and keep the format similar for single/multi variant pages (pending some more discussion). I'll switch it to use that formatting for now!

margaree (approved):

PR #1514, opened 2021-07-22
fix: remove use of nullish coalescing operator (workaround for `d2l-rubric`)

devpow112:
This isn't really a defect but it is causing issues with `d2l-rubric` CI and current workflows. Sadly, `d2l-rubric` is still using polymer and uses polymer related tooling for some of it's CI and development workflow. We are planning to upgrade to lit but for now this is a workaround to allow us to use the new `d2l-input-number` component in polymer.

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -124,7 +124,11 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 
 	get maxFractionDigits() {
 		// emulate Intl's default maxFractionDigits behaviour
-		return this._maxFractionDigits ?? Math.max(this.minFractionDigits, 3);
+		if (!this._maxFractionDigits) {
+			return Math.max(this.minFractionDigits, 3);
+		}
+
+		return this._maxFractionDigits;
 	}
 	set maxFractionDigits(val) {
 		if (isNaN(val) || val < 0 || val > 20 || val < this.minFractionDigits) {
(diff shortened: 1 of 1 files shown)


components/inputs/input-number.js
```diff
@@ -124,7 +124,11 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 
 	get maxFractionDigits() {
 		// emulate Intl's default maxFractionDigits behaviour
-		return this._maxFractionDigits ?? Math.max(this.minFractionDigits, 3);
+		if (!this._maxFractionDigits) {
+			return Math.max(this.minFractionDigits, 3);
+		}
+
+		return this._maxFractionDigits;
```
dlockhart, line 131: This won't work if `maxFractionDigits` is zero, since `!0 === true`.

```suggestion
		return (this._maxFractionDigits === undefined || this._maxFractionDigits === null) ?
				Math.max(this.minFractionDigits, 3) : this._maxFractionDigits;
```

components/inputs/input-number.js
```diff
@@ -135,7 +139,11 @@ class InputNumber extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(Lit
 	}
 
 	get minFractionDigits() {
-		return this._minFractionDigits ?? 0;
+		if (!this._minFractionDigits) {
+			return 0;
+		}
+
+		return this._minFractionDigits;
```
dlockhart, line 146: Ditto here.

```suggestion
		return (this._minFractionDigits === undefined || this._minFractionDigits === null) ?
				0 : this._minFractionDigits;
```

dlockhart (changes requested):

components/inputs/input-number.js
devpow112: Yeah working on changing that now 👍

devpow112:
@dlockhart any idea what's up with BrowserStack? I see a random disconnect after a timeout which I don't see how my change could have caused that

dlockhart:
We get legacy-Edge disconnects like that pretty frequently, normally not too many times in a row though. We're a few days away from turning off legacy-Edge tests completely haha. (I just restarted that step for you)

dlockhart (approved):

PR #1515, opened 2021-07-22
US130144: add top override for iframe dialogs to continuously move dialog down the page

annabelsegalld2l:
The last PR, https://github.com/BrightspaceUI/core/pull/1511, got the dialogs to appear the correct height. Now, the override is needed for the "top" of the dialog. Otherwise, it will stay fixed to the top of the page.

### DIFF
--- a/components/dialog/dialog-fullscreen.js
+++ b/components/dialog/dialog-fullscreen.js
@@ -209,12 +209,18 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 	render() {
 
 		const heightOverride = {} ;
+		let topOverride = null;
 		if (this._ifrauContextInfo) {
 			// in iframes, use calculated available height from dialog mixin minus padding
 			heightOverride.height = mediaQueryList.matches
 				? `${this._ifrauContextInfo.availableHeight - 42}px`
 				: `${this._ifrauContextInfo.availableHeight - 60}px`;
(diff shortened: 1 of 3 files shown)


components/dialog/dialog-fullscreen.js
```diff
@@ -209,12 +209,18 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
...
 		const heightOverride = {} ;
+		let topOverride = null;
 		if (this._ifrauContextInfo) {
 			// in iframes, use calculated available height from dialog mixin minus padding
 			heightOverride.height = mediaQueryList.matches
 				? `${this._ifrauContextInfo.availableHeight - 42}px`
 				: `${this._ifrauContextInfo.availableHeight - 60}px`;
 			heightOverride.minHeight = heightOverride.height;
+			const iframeTop = this._ifrauContextInfo.top < 0
+				? -this._ifrauContextInfo.top
```
annabelsegalld2l, line 220: If the contextInfo.top is positive, that means we are scrolled above the iframe. So, set the top of the iframe to 0 (we can't go above it). Otherwise, when the context.top is negative, we flip the sign to get the correct top.

dlockhart: Haha iframes are fun eh.

dlockhart (approved):
LGTM, although I'm sure we haven't seen the last of these pesky issues!

PR #1518, opened 2021-07-22
DE44163

jeroach:
[DE44163](https://rally1.rallydev.com/#/42960320374d/dashboard?detail=%2Fdefect%2F602532638451&fdp=true?fdp=true): [Consistent Eval] Loading the Consistent Eval page is causing a 404 error in the Browser console

Slack thread for context: https://d2l.slack.com/archives/C0PHG3QB0/p1626899269461000

- move `en` lang code before `en-us`

### DIFF
--- a/mixins/localize-mixin.js
+++ b/mixins/localize-mixin.js
@@ -124,6 +124,9 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass 
 	_generatePossibleLanguages() {
 		const langs = new Set();
 
+		langs.add('en');
+		langs.add('en-us');
+
 		let docLang = this.__documentLocaleSettings.language;
 		if (docLang) {
 			docLang = docLang.toLowerCase();
@@ -145,9 +148,6 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass 
 				langs.add(baseDocFallbackLang);
 			}
 		}
-
-		langs.add('en-us');
-		langs.add('en');
 
 		return Array.from(langs);
 	}
--- a/mixins/localize-mixin.js
+++ b/mixins/localize-mixin.js
@@ -124,13 +124,15 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass 
 	_generatePossibleLanguages() {
 		const langs = new Set();
 
-		langs.add('en');
-		langs.add('en-us');
+		const enVariants = ['en-us', 'en-ca', 'en-gb'];
 
 		let docLang = this.__documentLocaleSettings.language;
 		if (docLang) {
 			docLang = docLang.toLowerCase();
-			langs.add(docLang);
+
+			if (!enVariants.includes(docLang)) {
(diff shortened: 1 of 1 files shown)


mixins/localize-mixin.js
```diff
@@ -124,6 +124,9 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
 	_generatePossibleLanguages() {
 		const langs = new Set();
 
+		langs.add('en');
+		langs.add('en-us');
```
jeroach, line 128: Any thoughts on if we still need `en-us`?

I don't think web components typically provide a US specific lang term for EN.

mixins/localize-mixin.js
```diff
@@ -124,6 +124,9 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
 	_generatePossibleLanguages() {
 		const langs = new Set();
 
+		langs.add('en');
```
jeroach, line 127: Moved this block before the `docLang` and `docFallbackLang` blocks as those seem to also add `en-us` - which causes the same issue as before.

mixins/localize-mixin.js
bearfriend: They definitely shouldn't. If someone feels confident enough in their search skills to determine if any en-us files exist, I'd be happy to remove it if it's not necessary.

mixins/localize-mixin.js
bearfriend: Hmmm. Ok, I don't think we can do that, as it defeats the purpose of checking the document settings. Let me think about this a little bit more.

jeroach: Thanks :) Yeah it seems to be causing issues with the unit tests as well - I'm poking around

jeroach: Related to this - do we need the other `en` lang options here: 
https://github.com/BrightspaceUI/intl/blob/30c5729f28eb856a15c55bb36eecf314a2efa6bb/lib/common.js#L3

This is where those document language settings are pulled from.

jeroach: Okay so I think I have something that works - on the assumption that we don't use `en` variants in web components

Thoughts? @bearfriend @dlockhart @svanherk 

Thanks :)

svanherk: I'm not as familiar with our localize mixins, so someone feel free to jump in - but couldn't we definitely theoretically need an `en-us.js` file?  If a langterm included "favourite" in it and we need an `en-us` file just for those terms?
Also, why is this something that would only happen in dev?  I figure we're seeing this because the LMS defaults to `en-us`, and wouldn't that be a valid setting for all our US customers?

jeroach: Yeah tackling the issue from that side would work as well - adding `en-us` and `en-gb` language files to all web components.

As for why this only happens in Dev - I'm not sure - do the web components have access to a `en-us` in Prod environments? If so that would explain a lot.

PR #1522, opened 2021-07-23
Daylight: breadcrumbs

margaree:


components/breadcrumbs/README.md
```diff
@@ -2,8 +2,68 @@
...
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
+
+## Responsive Behavior
+
+There are various options to truncate the breadcrumb when in horizontally constrained spaces in order to keep the breadcrumb on one line.
+
+### Limited Width
+
+![screenshot of d2l-breadcrumbs component limited width](./screenshots/limited-width.png)
```
GeorgePlukov, line 42: Do we want to show these screenshots initially? Seems alright to leave it like this for now but I'd guess a lot of the screenshots/copy text will change when Jeff starts going through the pages

margaree: Good point. This one might be useful as a demo with code actually, and we can probably hide the compact mode one.

GeorgePlukov (approved):
LGTM. Up to you if you want to switch up those screenshots now or when we are reviewing pages (might be better to do it all at once)

PR #1523, opened 2021-07-23
DE44538: Wait for content to fully render before getting the contentContainer

annabelsegalld2l:
Dave Lockhart found this issue during the testing of d2l-dropdown. In some cases, the _openedChanged function runs before the content inside the dropdown has a chance to fully render. This is causing "this.__getContentContainer()" to return null in some cases, causing Javascript errors in the console. 

Note that this only occurs on Firefox, and only on some dropdowns in the LMS. The dropdown still appears and is clickable despite the error. Adding an await to wait for all updates to complete ensures the content can be found in all cases.

I'll also be certfixing this into 20.21.8, as the console error occurs there as well.

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -475,6 +475,8 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 	async __openedChanged(newValue) {
 
+		await this.updateComplete;
+
 		this.__previousFocusableAncestor =
 			newValue === true
 				? getPreviousFocusableAncestor(this, false, false)


svanherk (approved):

components/dropdown/dropdown-content-mixin.js
```diff
@@ -475,6 +475,8 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 	async __openedChanged(newValue) {
 
+		await this.updateComplete;
```
dbatiste, line 478: I know this is already merged so I wouldn't worry too much about this, but in the absence of a test to cover the error case, I might be inclined to add a code comment even though we like the code to speak for itself.  An unfamiliar dev might think to remove this line and it might seem ok if all tests pass, and it's not obvious from the code what erroneous behaviour might result.

PR #1524, opened 2021-07-23
Fix filter counts for the manual search type

svanherk:
When consumers decide to use the "manual" `search-type`, they have to physically remove options from the filter on search.  This avoids a lot of syncing work on our end and matches how all the consumers currently using HM search work anyways.  But when this happens, we no longer "know" how many options there are and how many are selected.  So we don't want to recount in this scenario.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -375,11 +375,11 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					this._totalAppliedCount--;
 				}
 			} else if (prop === 'values') {
-				shouldRecount = true;
+				if (dimension.searchType !== 'manual' || !dimension.searchValue) shouldRecount = true;
 			}
 		});
 
-		if (shouldRecount) this._setFilterCounts();
+		if (shouldRecount) this._setFilterCounts(dimension);
 		if (shouldUpdate) this.requestUpdate();
 	}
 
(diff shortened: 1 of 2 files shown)


dlockhart (approved):

components/filter/filter.js
```diff
@@ -375,11 +375,11 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					this._totalAppliedCount--;
 				}
 			} else if (prop === 'values') {
-				shouldRecount = true;
+				if (dimension.searchType !== 'manual' || !dimension.searchValue) shouldRecount = true;
```
svanherk, line 378: If there's no search value then we should have all the options and can safely recount.  We want to recount as often as we can in case set values are loaded in asynchronously.  A scenario that shouldn't happen but will break the counts is if after a search, the consumer returns us new values we didn't originally have before searching.  This is easily avoidable though - don't finish loading and pass us the values until you have at least all the keys.  The filter doesn't support paging/load more.

components/filter/test/filter.test.js
```diff
@@ -433,8 +433,10 @@ describe('d2l-filter', () => {
 
 			const newDim = document.createElement('d2l-filter-dimension-set');
 			newDim.key = 'newDim';
+			newDim.text = 'New Dim';
```
svanherk, line 436: Unrelated, this fixes come console warnings after @dbatiste  added the LabelMixin stuff

PR #1525, opened 2021-07-26
Add square corners to slide out mobile trays depending on direction

annabelsegalld2l:
During testing we found that all corners are rounded in the dropdown mobile trays. This PR makes sure only the "sliding" corners are rounded, and the corners on the edge of the screen are straight. The visual diffs for all the trays are affected by this change. As you can see, only the corners are affected.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -475,6 +475,8 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 	async __openedChanged(newValue) {
 
+		// DE44538: wait for dropdown content to fully render,
```
annabelsegalld2l, line 478: Adding this comment as a follow-up to https://github.com/BrightspaceUI/core/pull/1523

dlockhart (approved):

PR #1534, opened 2021-07-26
Daylight: Progress meters (linear, radial, circle)

GeorgePlukov:


components/meter/README.md
```diff
@@ -1,91 +1,138 @@
...
+* Use to provide feedback when a user completes an action that results in a change in data.
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don't use a lot of progress indicators on a single screen because they can overwhelm the user.
+* Don't use motion unless the data is the primary content on the screen.
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
+
+## Light Foreground
```
GeorgePlukov, line 35: I wasn't 100% sure if this should be left on the page but it seems to fit best as an H2 if we were to leave it. Might be something that gets cut in a design review

GeorgePlukov: Actually now that I think about it, this should probably be defined on the mixin as a property. Seems like all of the meter components use this property (and I would think it should show up in the attributes table)

components/meter/README.md
```diff
@@ -1,91 +1,138 @@
...
-- `percent` (Boolean): Shows a percentage instead of `value/max`.
-- `text-inline` (Boolean): Keeps the meter to a single line.
-- `text` (String): Context information about what the meter is about.
-	- `{%}` in the string will be replaced with percentage value
-	- `{x/y}` in the string will be replaced with fraction with the proper language support
-	- **DEPRECATED** `{x}` in the string will be replaced with `value`
-	- **DEPRECATED** `{y}` in the string will be replaced with `max`
+* `value` (required, Number): Current number of completed units. A positive, non-zero number that is less than or equal to `max`.
+* `value` (required, Number): Current number of completed units. A positive, non-zero number that is less than or equal to `max`.
+* `value` (required, Number): Current number of completed units. A positive, non-zero number that is less than or equal to `max`.
```
margaree, line 66: `value` x 3😱

margaree (approved):

components/meter/README.md
GeorgePlukov: Oh no hahaha, I was using `cmd` + `opt` to switch from `-` to `*` and must have hit the wrong button 😆

components/meter/README.md
GeorgePlukov: So one problem with adding this as an attribute is that we need a dark background to display the light versions properly.  I may leave the attribute undocumented in jsdocs, but add a separate demo that shows the `foreground-light` version. Will need to add some configuration for a dark background in our iframe along with this, maybe a `darkmode:true` or something similar

margaree: Cool yeah the separate demo probably makes sense for now.

PR #1537, opened 2021-07-28
Daylight: Calendar: Attributes should be on single line

margaree:
Attributes should be on single line or else they don't seem to update (something potentially to look into in documentation as well, not sure on fix complexity).


GeorgePlukov (approved):

GeorgePlukov:
Very weird!

PR #1540, opened 2021-07-28
Updating Translations 2108 28/07/2021

annabelsegalld2l:
Moved this PR to point to main instead of master. There are some missing translations that will hopefully get added in the next translations PR. Also, looks like there are some merge conflicts that need to be resolved.


annabelsegalld2l:
We are good to close this PR since the "Filters" term has already been translated in the lang files.

PR #1543, opened 2021-07-28
Daylight: Tab and Tab-panel

GeorgePlukov:
I added a section for tab-panel here so that you would be able to interact with those properties, possible that we remove this since its just selected, text, and the no-padding properties

### DIFF
--- a/components/tabs/README.md
+++ b/components/tabs/README.md
@@ -1,18 +1,15 @@
+
 # Tabs
+Tabs are used to present related information in mutually exclusive panels, allowing users to view just one panel at a time.
 
-## d2l-tabs
-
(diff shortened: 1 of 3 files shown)


components/tabs/README.md
```diff
@@ -23,24 +61,54 @@ The `d2l-tabs` element is a web component for tabbed content. It provides the `d
 </d2l-tabs>
 ```
 
-**Tabs Properties:**
+
```
margaree, line 64: nit: looks like an extra line got added

components/tabs/tab-panel-mixin.js
```diff
@@ -7,6 +7,7 @@ export const TabPanelMixin = superclass => class extends superclass {
 		return {
 			/**
 			 * Opt out of default padding/whitespace around the panel
+			 * @type {Boolean}
```
margaree, line 10: Types are generally lower cased, and I believe our `demo-tables` code looks for lower cased values.

margaree (approved):

PR #1547, opened 2021-07-29
Daylight: Table

margaree:
This one was a bit complex to get working in Daylight due to not really being a component. I ended up utilizing the demo test-table for the live demo, partially because it is difficult to create a component in the documentation repo and have that imported into a demo. I added a few rows to the demo in order to better showcase sticky-headers and also added `no-column-border`.

I just kept the gifs for sortable column buttons and selection since that was simpler. Those are also shown in the live demo, though their properties can't be interacted with since it's more a demo of the wrapper.

### DIFF
--- a/components/table/README.md
+++ b/components/table/README.md
@@ -1,8 +1,32 @@
-# Table
+# Tables
+
+Tables are used to display tabular data in rows and columns of cells. They can allow users to select rows and sort by columns.
+
+## Best Practices
+<!-- docs: start best practices -->
+<!-- docs: start dos -->
+* Do use tables to display complex data sets.
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don’t use tables to display a simple list of objects or entities; consider using the list component for this instead.
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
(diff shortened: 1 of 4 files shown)


components/table/README.md
```diff
@@ -91,23 +137,22 @@ When tabular data can be sorted, the `<d2l-table-col-sort-button>` can be used t
 </table>
 ```
 
-**Properties:**
+### Properties
 
 |--|--|--|
```
GeorgePlukov, line 143: Might require testing again but I think if we want table styles to apply properly to inline markdown tables they need to have 3 dashes to specify the header line 
```suggestion
|---|---|---|
```

components/table/table-wrapper.js
```diff
@@ -184,6 +185,7 @@ export class TableWrapper extends RtlMixin(LitElement) {
 			},
 			/**
 			 * Whether header row is sticky
```
GeorgePlukov, line 187: Did we want to update property descriptions like this? Not super critical but some of these arent particularly descriptive of what the property actual controls

GeorgePlukov (approved):

PR #1549, opened 2021-07-29
US130416 - Allow selection components to be used outside the collection

svanherk:
This went through a few iterations, but I think the one below is the cleanest.  The selection components can now be used outside of the selection they want to subscribe to, as long as they're in the same DOM scope.  The subscribers handle they're own subscribing/unsubscribing, including the cases where they are added before the collection, after the collection, or want to change the collection they are subscribing to.

I did add this to filter to make sure it all works as expected - that will be added in a follow-up PR.


components/selection/selection-mixin.js
```diff
@@ -60,6 +60,10 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 		this.addEventListener('d2l-selection-select-all-change', this._handleSelectionSelectAllChange);
 		this.addEventListener('d2l-selection-observer-subscribe', this._handleSelectionObserverSubscribe);
 		this.addEventListener('d2l-selection-input-subscribe', this._handleSelectionInputSubscribe);
+
+		const externalSubscribers = this.getRootNode().querySelectorAll(`[subscribed-to="${this.id}"]`);
```
svanherk, line 64: Originally I was using a mutation observer similar to the `LabelledMixin`, but it was being added too late in the filter scenario and the selection components had to register themselves anyways.  So what I have currently is a two-way registration system.  If the components try to register themselves but the collection (ex. `list`) they want to subscribe to doesn't exist yet, this code will add them once it is connected.  Likewise, if a component is added after this code has run, [it will subscribe itself](https://github.com/BrightspaceUI/core/pull/1549/files#diff-a7412e30193c0f86945842eaebc7ea7709b8f1b5a541f5b42a7bd56026f5735aR39).

components/selection/selection-mixin.js
```diff
@@ -87,6 +91,19 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 		return new SelectionInfo(keys, state);
 	}
 
+	setSelectionForAll(selected) {
```
svanherk, line 94: Exposing this for:
- the `selection-select-all` component, since it can't fire an event if it's external to the collection
- For filter's "clear" button to use
  - From a convo in #wc-list, Jeff doesn't think other use cases will need a clear button (rather than just uncheck the select all box).  So the mixin will still control how to clear, but there won't be a specific `selection-action-clear` control (at least not yet).

components/selection/selection-observer-mixin.js
```diff
@@ -7,7 +7,12 @@ export const SelectionObserverMixin = superclass => class extends superclass {
 			/**
 			 * The selection info (set by the selection component).
 			 */
-			selectionInfo: { type: Object }
+			selectionInfo: { type: Object },
+			/**
+			 * Id of the collection this component wants to subscribe to (if not located with that collection)
+			 */
+			subscribedTo: { type: String, reflect: true, attribute: 'subscribed-to' },
+			_provider: { type: Object, attribute: false }
```
svanherk, line 15: Added this for the case where the collection later registers the components, so that this will cause a re-render of the `selection-select-all` and the `selection-summary` components.  They hide themselves if the collection has the `selection-single` attribute.

components/selection/selection-select-all.js
```diff
@@ -51,6 +51,7 @@ class SelectAll extends LocalizeCoreElement(SelectionObserverMixin(LitElement))
 			composed: true,
 			detail: { checked: e.target.checked }
 		}));
+		if (this.subscribedTo && this._provider) this._provider.setSelectionForAll(e.target.checked);
```
svanherk, line 54: My thinking here is that we always want to send the `d2l-selection-select-all-change` event since we advertise that for consumers, but if we're external to the collection we'll need to trigger the selection ourselves.

components/selection/selection-observer-mixin.js
```diff
@@ -21,13 +26,18 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
+			if (!this.subscribedTo) {
+				const evt = new CustomEvent('d2l-selection-observer-subscribe', {
+					bubbles: true,
+					composed: true,
+					detail: {}
+				});
+				this.dispatchEvent(evt);
+				this._provider = evt.detail.provider;
+			} else {
+				this._provider = this.getRootNode().querySelector(`#${this.subscribedTo}`);
```
dlockhart, line 38: Can you use the new `cssEscape` DOM helper around the id?

components/selection/selection-observer-mixin.js
```diff
@@ -21,13 +26,18 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
-			this._provider = evt.detail.provider;
+			if (!this.subscribedTo) {
+				const evt = new CustomEvent('d2l-selection-observer-subscribe', {
+					bubbles: true,
+					composed: true,
+					detail: {}
+				});
+				this.dispatchEvent(evt);
+				this._provider = evt.detail.provider;
+			} else {
```
dlockhart, line 37: I think it'd be better to move this part into the `updated()` callback so that if the `subscribed-to` value changes you can unsubscribe and re-subscribe to the new provider. It's an edge case though.

components/selection/selection-select-all.js
dlockhart: For simplicity then would it be worth changing the provider so that it doesn't listen for the event but instead just expects this method to be called?

components/selection/selection-mixin.js
dlockhart: Interesting. Where were you adding the mutation observer?

svanherk: I created it here in the `connectedCallback` and had it watching all children of the `this.getRootNode()`.  I tried moving it to the `constructor` and it still seemed to miss it, but I can confirm that again

components/selection/demo/selection.html
```diff
@@ -91,6 +110,48 @@ <h2>Single Selection</h2>
...
+							<div class="d2l-selection-header">
+								<d2l-selection-select-all subscribed-to="collection-2"></d2l-selection-select-all>
+								<d2l-selection-action subscribed-to="collection-2" text="Add Note" icon="tier1:add-message"></d2l-selection-action>
+								<d2l-selection-summary subscribed-to="collection-2"></d2l-selection-summary>
+							</div>
+							<d2l-demo-selection id="collection-2" selection-single>
+								<ul>
+									<li><d2l-selection-input key="let" label="Lettuce"></d2l-selection-input>Lettuce</li>
+									<li><d2l-selection-input key="tom" label="Tomato"></d2l-selection-input>Tomato</li>
+									<li><d2l-selection-input key="onion" label="Onion"></d2l-selection-input>Onion</li>
```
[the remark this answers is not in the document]
svanherk: Then you must use the form on the left, in my overly-complicated-but-very-limited sandwich shop app 😛

components/selection/selection-select-all.js
svanherk: Yeah, if we need to do this anyways that would simplify things.  If no one is using the event yet, do we even still want it?  I'm working under the assumption this is an event we expect consumers to want to listen for and then do something with

components/selection/selection-mixin.js
dbatiste: It is possible the subscriber was added before the connectedCallback for this component...?

components/selection/selection-mixin.js
dbatiste: If a consumer wants to set the selection - how would they go about it?  I think my original thinking was that they would set the `selected` property on the items individually.  I am just contemplating whether we would even want a `setSelectionInfo` (akin to the existing `getSelectionInfo`), and if so, would that method be used for this purpose instead.  Just thinking out loud.

components/selection/selection-observer-mixin.js
```diff
@@ -7,7 +7,12 @@ export const SelectionObserverMixin = superclass => class extends superclass {
 			/**
 			 * The selection info (set by the selection component).
 			 */
-			selectionInfo: { type: Object }
+			selectionInfo: { type: Object },
+			/**
+			 * Id of the collection this component wants to subscribe to (if not located with that collection)
+			 */
+			subscribedTo: { type: String, reflect: true, attribute: 'subscribed-to' },
```
dbatiste, line 14: I don't feel strongly either way on the naming of this property.  If we wanted to hide the "subscribe" implementation detail from the naming, we _could_ opt for something like `for`.  🤔

components/selection/selection-mixin.js
dlockhart: One place I was thinking for the observer was actually in the subscriber. So it would do a `querySelector` for it in `updated()` and if it didn't find it, it could add the mutation observer.

components/selection/selection-observer-mixin.js
dlockhart: Since it's a mixin I worry more about name collisions... haha I'd almost go the other way and call it `selection-subscribed-to` or `selection-for`.

components/selection/selection-mixin.js
svanherk: Hmm that could work, and then cleans up these lines so the subscriber is completely in charge of subscribing.  Is that "heavier" though... to be listening to all mutations for when the list arrives, rather than have the list tell me when it's arrived?  It means I could handle the case where the list suddenly disappears, which I'm realizing isn't really handled right now.

components/selection/selection-mixin.js
svanherk: > I think my original thinking was that they would set the selected property on the items individually.

That was my thought as well, since they should be aware of their collection and can set the `input-selection` or `list-item` value directly.  Whereas the selection components don't know much about the list or how to traverse it or update it.  We could offer a `setSelectionInfo` to make that cleaner, but I still think `select-all` would need a `setSelectionInfoForAll` function (that could call `setSelectionInfo`), since it doesn't know all the keys?

components/selection/selection-observer-mixin.js
svanherk: Seems like `selection-for` handles both concerns (doesn't give away implementation details, and is a bit more unique)

components/selection/selection-mixin.js
dlockhart: Ideally mutation observer is supposed to be performant, but yeah it could be a bit heavy. You can limit the things you listen for -- so in your case you only care about additions/removals I think. And then yeah it would then suddenly cover the case where the list was removed entirely and you could act accordingly.

components/selection/selection-select-all.js
```diff
@@ -51,6 +51,7 @@ class SelectAll extends LocalizeCoreElement(SelectionObserverMixin(LitElement))
 			composed: true,
```
svanherk, line 51: Still not sure if we want this event or not, now that the mixin doesn't listen for it.

components/selection/selection-mixin.js
```diff
@@ -87,6 +85,21 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
...
```
svanherk, line 88: Related to @dbatiste's original comment here, still contemplating whether we want some kind of `setSelectionInfo` function.  It could take the selected value and then a key/node/whatever, and if no key is passed we could select/clear everything.  But we don't really have a use-case for that yet, so I'm leaning towards leaving it like this for now

svanherk (commented):
Handling the registration completely in the observer simplified other parts nicely.  I think I addressed the comments, still pondering the below while I work on tests

components/selection/demo/selection.html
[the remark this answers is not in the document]
svanherk: I actually ended up changing this lol

components/selection/selection-select-all.js
dlockhart: I always lean towards not exposing things until we absolutely need them.

components/selection/README.md
```diff
@@ -66,6 +90,7 @@ The `d2l-selection-action` is an optional component that provides a button for a
 |--|--|--|
 | `icon` | String | Preset icon key (e.g. "tier1:gear"). |
 | `requires-selection` | Boolean | Whether the action requires one or more selected items. If no items are selected, the action button will be focusable and a hint displayed in a tooltip. |
+| `selection-for` | String | Id of the corresponding `selectionMixin` component, if not placed within it. |
```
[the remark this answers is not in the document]
svanherk: Yup good call!  Missed that

components/selection/selection-select-all.js
svanherk: Removed the event.  Will be easy to add back if it turns out a consumer does need it

components/selection/selection-observer-mixin.js
```diff
@@ -33,8 +41,31 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
+		super.updated(changedProperties);
+
+		if (!changedProperties.has('selectionFor')) return;
+
+		if (this._selectionForObserver) this._selectionForObserver.disconnect();
+		if (this._provider) this._provider.unsubscribeObserver(this);
+
+		this._provider = this.getRootNode().querySelector(`#${cssEscape(this.selectionFor)}`);
+		if (this._provider) this._provider.subscribeObserver(this);
+
```
dlockhart, line 57: GitHub lost my comment that I put in here from yesterday, so I'll try to re-create it...

I think we now need to handle the case where the provider is removed from the DOM but the observer is not. I think that would be easiest to do from the SelectionMixin's `disconnectedCallback`? That way if _both_ the provider and observer are removed, they still get detached.

And then once that's happening, I'm wondering if we could refactor this observer code... does it always run at least once? If so, I think you could just let the observer code do the initial fetching of the provider. If not, then can we avoid wiring it up at all if the provider is found right away?

Lastly, the observer should be disconnected in this component's `disconnectedCallback`.

svanherk: > I think we now need to handle the case where the provider is removed from the DOM but the observer is not. I think that would be easiest to do from the SelectionMixin's `disconnectedCallback`? That way if both the provider and observer are removed, they still get detached.

So that is actually handled, because the mutation observer runs whenever any node is added or deleted.  So if the provider is deleted, the code runs and [checks that it can still find it's provider](const selectionComponent = this.getRootNode().querySelector(`#${cssEscape(this.selectionFor)}`);), and if not it'll [set the its provider to null](https://github.com/BrightspaceUI/core/pull/1549/files#diff-a7412e30193c0f86945842eaebc7ea7709b8f1b5a541f5b42a7bd56026f5735aR61).  And then if the provider is added back, the same code will run and add it back.  I definitely missed adding a test for that case though, so I'll add that now!

I originally _did_ have the provider's `disconnectedCallback` setting all its subscribers' `_provider`s to null, but it was overkill when the `mutationObserver` code was also handling that case.  But we could change it to more like what you're saying - try to get the provider in `updated` when `selectionFor` is set or changed.  If we can, all good - don't setup a `mutationObserver` until our provider tells us it's going away in its `disconnectedCallback`.  If we can't find it, setup a `mutationObserver` to keep checking every time a node is added or deleted until we get one.

I liked that this way didn't involve the `SelectionMixin` having to care at all, but the way above does seem like it'd be more performant with less `mutationObserver`s running

dlockhart: Apologies I missed how that would set things to `null` -- I'm good with how you have it set up. 👍 

That last question remains: does the mutation observer always fire at least once when you set it up, or if it finds the provider right away does it never fire? Just wondering if you could remove the duplicate 2 lines that query for the provider and subscribe.

dlockhart (approved):

components/selection/selection-observer-mixin.js
svanherk: > does the mutation observer always fire at least once when you set it up

From what I can tell, it does not - it will only fire for added/removed nodes after `observe` is called.  Meaning it never fires for the demo page, and only fires in specific situations on the filter page (like loading, where the list pops in later).  I added a bit of code to those two lines to handle clearing the selection info though, so I wrapped it in a function.

svanherk: > Lastly, the observer should be disconnected in this component's disconnectedCallback.

Done, and I also added it to the `LabelledMixin` I was using as an example

dlockhart: Cool that addresses my code duplication concern regardless.

dlockhart (approved):

components/selection/selection-observer-mixin.js
```diff
@@ -33,8 +41,39 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
 
+	_updateProvider() {
+		const selectionComponent = this.getRootNode().querySelector(`#${cssEscape(this.selectionFor)}`);
+		if (this._provider === selectionComponent) return;
+
+		this._provider = selectionComponent;
+		if (this._provider) {
+			this._provider.subscribeObserver(this);
+		} else {
+			this.selectionInfo = new SelectionInfo();
```
svanherk, line 76: If we lose our provider, clear our `selectionInfo`

PR #1550, opened 2021-07-29
feat: US130326, US129220 Add a focus trap to dropdowns, use mobile-tray in date-picker

annabelsegalld2l:
This PR adds a new attribute to dropdown-content, trap-focus. If the property is true, a d2l-focus-trap will automatically wrap the dropdown, creating a focus loop. 

The focus trap was required for the date-picker to use a mobile tray, which is why I put the PR's together. If the date-picker controls its own focus-trap, it can never reach the mobile close button via tabbing (due to its focus loop). If the date-picker uses the trap-focus attribute instead, it can wrap both the calendar AND the close button. The date picker listens to the dropdown's focus-enter event to shift focus to the correct calendar elements. 

Also added some date picker visual diffs to show the mobile tray style.


components/dropdown/README.md
```diff
@@ -221,11 +221,13 @@ To make your usage of `d2l-dropdown-more` accessible, use the following property
 | `no-padding-header` | Boolean: default: `false` | Render the header with no padding (if it has content) |
 | `no-padding-footer` | Boolean, default: `false` | Render the footer with no padding (if it has content) |
 | `no-pointer` | Boolean, default: `false` | Render without a pointer |
+| `trap-focus` | Boolean, default: `false` | Optionally render a d2l-focus-trap around the dropdown content |
 | `vertical-offset` | Number or String | Provide custom offset, positive or negative |
 
 **Events:**
 * `d2l-dropdown-open`: dispatched when the dropdown is opened
 * `d2l-dropdown-close`: dispatched when the dropdown is closed
+* `d2l-dropdown-focus-enter`: dispatched when the 'trap-focus' attribute is applied and the focus-trap is entered
```
dlockhart, line 230: If we think this is a weird edge case, we _could_ hide this event from the documentation. Good either way.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -735,6 +745,21 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+			const focusable = getFirstFocusableDescendant(content);
+			if (focusable) {
+				// bumping this to the next frame is required to prevent IE/Edge from crazily invoking click on the focused element
+				requestAnimationFrame(() => focusable.focus());
+			} else {
+				content.setAttribute('tabindex', '-1');
+				content.focus();
+			}
+		}
+		this.dispatchEvent(new CustomEvent('d2l-dropdown-focus-enter', { bubbles: true, composed: true }));
```
dlockhart, line 760: Does this need to bubble and be composed? Seems like the consumer could just wire up directly to the dropdown-content if they care about it.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -338,4 +338,35 @@ describe('d2l-dropdown', () => {
 
 	});
 
+	describe('trap-focus option', () => {
+
+		it('no focus trap when trap-focus is false', async() => {
+			//content.setAttribute('trap-focus', false);
```
dlockhart, line 344:
```suggestion
```

components/dropdown/dropdown-content-styles.js
```diff
@@ -108,6 +108,13 @@ export const dropdownContentStyles = css`
 		width: 100vw;
 	}
 
+	d2l-focus-trap {
+		max-width: 370px;
+		min-width: 70px;
+		position: absolute;
+		width: 100vw;
+	}
```
dlockhart, line 116: Interesting -- why was this needed?

components/inputs/input-date.js
```diff
@@ -376,7 +376,10 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 	}
 
 	_handleDropdownOpen() {
-		if (this._dropdown && !this._dropdown.openedAbove) this.shadowRoot.querySelector('d2l-focus-trap').scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
+		if (this._dropdown && !this._dropdown.openedAbove) {
+			this._dropdown.shadowRoot.querySelector('d2l-focus-trap')
+				.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
```
dlockhart, line 381: Hmm, the reaching into dropdown's shadow root to do the scrolling is a little unfortunate. Do we know what this was being done for initially? Wondering if we still need it now with the mobile tray...

components/inputs/test/input-date.test.js
```diff
@@ -55,7 +55,7 @@ describe('d2l-input-date', () => {
 			await dropdownContent.updateComplete;
 			dropdown.toggleOpen();
 			await oneEvent(dropdown, 'd2l-dropdown-open');
-			const focusTrap = getChildElem(elem, 'd2l-focus-trap');
+			const focusTrap = getChildElem(dropdownContent, 'd2l-focus-trap');
```
dlockhart, line 58: Could probably just delete these two tests since technically this behaviour now belongs to dropdown.

components/dropdown/dropdown-content-styles.js
annabelsegalld2l: Ah, some of those styles were left over from previous debugging, I can remove them. The position: absolute is needed, though. In the non-focus case, .d2l-dropdown-content-width has the same position. It also needs to be applied to the focus-trap, or else the positioning is wrong and the pointer does not line up with the container.

components/dropdown/README.md
annabelsegalld2l: I think we can keep it here. It's useful if any consumers want the focus trap added, but want to override the initial focus location.

components/inputs/input-date.js
annabelsegalld2l: Did some digging here. It's used in the case of short screens (not necessarily mobile) where the dropdown gets scrolled into view. Looks like replacing focus-trap with calendar works as well, since they are the same size.

dbatiste:
Something to be careful/aware of - if there is ever more than one active focus trap on the page, I think it can get into an infinite loop where there are fighting over focus.  For example, if such a dropdown were used inside of a `d2l-dialog`, I think this situation might arise but I am not 100% sure.  Nesting them _might_ be fine, like nested dialogs, but active sibling focus traps would not be fine.  Might be a scenario to check to ensure it's ok.

annabelsegalld2l:
> Something to be careful/aware of - if there is ever more than one active focus trap on the page, I think it can get into an infinite loop where there are fighting over focus. For example, if such a dropdown were used inside of a `d2l-dialog`, I think this situation might arise but I am not 100% sure. Nesting them _might_ be fine, like nested dialogs, but active sibling focus traps would not be fine. Might be a scenario to check to ensure it's ok.

I did run into the infinite loop issue with a focus-trapped dropdown inside a dialog. The date-picker did not have the focus loop issue since it listens to the dropdown's focus-enter event. I was able to fix this by having the dialog mixin ignore focus enter events not targeted towards it. This will stop any nested focus traps inside the dialogs from having this issue.

dlockhart (approved):
Looks great! 🎉

PR #1552, opened 2021-07-29
Daylight: Overflow-group

GeorgePlukov:


components/overflow-group/README.md
```diff
@@ -1,10 +1,40 @@
 # Overflow Groups
+The `d2l-overflow-group` element can be used to add responsiveness to a set of buttons, links or menus. 
 
-## d2l-overflow-group
-The `d2l-overflow-group` element can be used to add responsiveness to a set of buttons, links or menus. Items added to this container element will no longer wrap onto a second line when the container becomes to small but will be added too a dropdown menu with configurable styling.
+<!-- docs: demo size:medium autoSize:false display:block -->
```
margaree, line 4: super nit: I've been trying to alphabetize the properties (other than if `name` is there I usually put that first)

components/overflow-group/README.md
```diff
@@ -1,10 +1,40 @@
...
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/overflow-group/overflow-group.js';
```
margaree, line 7: Would this need the other imports? Like dropdown button etc.

components/overflow-group/README.md
```diff
@@ -1,10 +1,40 @@
...
+		</d2l-dropdown-menu>
+	</d2l-dropdown>
+	<d2l-button>Copy</d2l-button>
+	<d2l-button>Import</d2l-button>
+	<d2l-button>Delete</d2l-button>
+</d2l-overflow-group>
+```
+
+## Overflow Group [d2l-overflow-group]
+Items added to this container element will no longer wrap onto a second line when the container becomes to small but will be added too a dropdown menu with configurable styling.
```
margaree, line 31:
```suggestion
Items added to this container element will no longer wrap onto a second line when the container becomes too small, but will be added to a dropdown menu with configurable styling.
```

margaree (approved):

components/overflow-group/README.md
GeorgePlukov: The buttons/dropdowns that get used are imported within the component so we don't technically need to add them here.

PR #1554, opened 2021-07-30
Add tier1 share-hollow icon

jameswklassen:
A new tier1 share-hollow icon created to use in Portfolio

## Context
Previously we were using the tier2 share icon and forcing the size down to `18px`.

Comparison of new tier1 icon (left) vs. tier2 (right)

image

## Checks
- [x] Regenerated catalogue.md with `npm run build:icons`
- [x] Tested using demo page to verify behaviour

## Screenshots
share-hollow

share-hollow-rtl


dlockhart (approved):
Looks great!

PR #1563, opened 2021-08-04
Daylight: Checkbox & Radio

margaree:
I investigated combining these into 1 page, but they both have quite a bit of content and so I decided to leave them separate at least for now.

Checkbox is fairly straightforward. Radio is a bit more complex, and so I took a strategy similar to that used with table where I used example components for demos, and added a couple additional demos.

There are a couple small tweaks to input-date-time and table which I put in here as well.


components/inputs/docs/input-checkbox.md
```diff
@@ -24,43 +72,49 @@ The `<d2l-input-checkbox>` element can be used to get a checkbox and optional vi
...
 
 | Attribute | Description |
-|--|--|
+|---|---|
 | `aria-label` | Use when text on checkbox does not provide enough context |
 | `description` | Use when label on input does not provide enough context. |
 
+### Usage
+
 **Methods:**
```
GeorgePlukov, line 97: Are we including bold titles like this within daylight pages? Maybe we remove the `Methods` title or switch the H3 above it to `Usage Methods`

components/inputs/docs/input-radio.md
```diff
@@ -1,19 +1,52 @@
...
 
+## Best Practices
+<!-- docs: start best practices -->
+<!-- docs: start dos -->
+* Use as an input for traditional forms
+* Use when there are 3 or more mutually exclusive options
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don’t use 2 radios if a 1 checkbox works better
```
GeorgePlukov, line 31:
```suggestion
* Don’t use two radio buttons if a checkbox works better
```

components/inputs/docs/input-radio.md
```diff
@@ -66,10 +99,28 @@ class MyElem extends LitElement {
...
+  import '@brightspace-ui/core/components/inputs/demo/input-radio-solo-test.js';
+</script>
+<d2l-test-input-radio-solo></d2l-test-input-radio-solo>
+```
+
+## Radio Spacer [d2l-input-radio-spacer]
 
 To align related content below radio buttons, the `d2l-input-radio-spacer` element can be used:
 
+<!-- docs: demo display:block -->
```
GeorgePlukov, line 116: Why do these need `display: block`?

components/inputs/docs/input-checkbox.md
margaree: I believe we'd discussed having a separate "usage" or "how to use" section in components where necessary and so calling it "usage" would keep that consistent, and then here within that is methods. I'll confirm with Jeff though.

components/inputs/docs/input-radio.md
margaree: I feel like there was some weird spacing happening, but yeah they do seem fine. I'll re-test and remove.

GeorgePlukov: Yeah no problem to leave it, just expected these to work well with `display: flex`

GeorgePlukov (approved):

PR #1565, opened 2021-08-04
Lit 2.0: run-async tests

dlockhart:
The Lit 2.0 migration of directives is pretty significant, so I wrote some unit tests around the existing Lit 1.0 directive to help avoid regressions.

I'll leave this open until @dbatiste has had a chance to review.

To make room for the tests, I also decided to move the directive into its own directory. Technically this is a breaking change, but a quick code search revealed that we're only using this internally inside core, so seems safe-ish.

### DIFF
--- a/directives/run-async/test/run-async.test.js
+++ b/directives/run-async/test/run-async.test.js
@@ -50,7 +50,6 @@ describe('run-async directive', () => {
 			expect(elem.innerText).to.equal('initial');
 		});
 
-		// TODO: was this actually desired behaviour, or should nothing be rendered?
 		it('should render failure template if InitialStateError occurs but no initial template is provided', async() => {
 
 			const elem = await fixture(html`<p>${runAsync('key', task, {
--- a/directives/run-async/test/run-async.test.js
+++ b/directives/run-async/test/run-async.test.js
@@ -70,7 +70,7 @@ describe('run-async directive', () => {
 			expect(elem.innerText).to.equal('failure');
 		});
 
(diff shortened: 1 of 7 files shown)


directives/run-async/test/run-async.test.js
```diff
@@ -0,0 +1,112 @@
...
+	describe('templates', () => {
+
+		it('should render initial template if InitialStateError occurs', async() => {
+			const elem = await createFixture();
+			taskReject(new InitialStateError());
+			await elem.updateComplete;
+			expect(elem.innerText).to.equal('initial');
+		});
+
+		// TODO: was this actually desired behaviour, or should nothing be rendered?
```
dlockhart, line 53: I wasn't totally sure if this was intended, but it reflects [how the original was written](https://github.com/PolymerLabs/async-demos/blob/master/packages/await-io/src/run-async.ts). Essentially if an `initial` template isn't provided and an `InitialStateError` happens it's going to render the `failure` template instead.

dbatiste: That seems reasonable.

dlockhart: Cool thanks, I've removed the comment. This PR is now fully ready for review.

dbatiste (approved):
Looks good!  Besides the unit tests, for the conversion, I would also be tempted to manually check the demo for the async container mixin, and also the dialog demo.  Those demos happen to contain usages of `runAsync`.

PR #1566, opened 2021-08-04
Lit 2.0: convert run-async directive to class-based syntax

dlockhart:
Building on the new unit tests, this converts the directive to follow the new Lit 2.0 class-based syntax:
https://lit.dev/docs/templates/custom-directives/#async-directives

The big change with the class-based directives is that a `WeakMap` is no longer needed for state, which can instead move into the class. This simplifies things a bit in that you always have access to that state in the callbacks and don't need to keep grabbing it and confirming that it's valid.

Things also get split out into `render()` (which can run on the server) and `update()` which does not. So that's where I put the firing of the `pending-state` event.

### DIFF
--- a/directives/run-async/run-async.js
+++ b/directives/run-async/run-async.js
@@ -1,4 +1,9 @@
 /**
+ * Originally based on the Polymer run-async Lit 1.0 directive:
+ * https://github.com/PolymerLabs/async-demos/blob/master/packages/await-io/src/run-async.ts
+ * Ported to Lit 2.0
+ */
+/**
  * @license
  * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
  * This code may only be used under the BSD style license found at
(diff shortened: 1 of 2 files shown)


dlockhart:
From Dave B in #1565:
> Besides the unit tests, for the conversion, I would also be tempted to manually check the demo for the async container mixin, and also the dialog demo. Those demos happen to contain usages of runAsync.

Yep for sure, tested these a lot while developing this.

directives/run-async/run-async.js
```diff
@@ -11,16 +16,107 @@
...
+					}
+				}
+			});
+		}
+
+		if ((this.state === 'pending') && typeof pending === 'function') {
+			return pending();
+		}
+
+		return noChange;
```
dbatiste, line 98: Interesting... https://lit.dev/docs/api/custom-directives/#noChange. 😄

dbatiste (approved):
Looks good!  So for the most part, it looks like it's following the original logic fairly closely with the key differences as you highlighted.

I wonder if it still needs to wait a microtask for the initial render to complete before dispatching the event.  If that's not needed, it looks like the `async() => { ... })();` wrapper would no longer be needed.

dlockhart:
> I wonder if it still needs to wait a microtask for the initial render to complete before dispatching the event. If that's not needed, it looks like the async() => { ... })(); wrapper would no longer be needed.

I just played around with this, and it looks like it can be removed. Although now the first time it runs happens before there's a host so `part.parentNode.host` isn't set yet. I added some checking for that and things seem to be working.

directives/run-async/run-async.js
```diff
@@ -11,16 +16,103 @@
...
+
+	}
+
+	update(part, [key, task, templates, options]) {
+		const renderResult = this.render(key, task, templates);
+		const directiveOptions = Object.assign({ pendingState: true }, options);
+		if (directiveOptions.pendingState && this.key === key && this.state === 'pending') {
+			const element = part.parentNode.nodeType === Node.ELEMENT_NODE ? part.parentNode : part.parentNode.host;
+			if (element) {
```
dbatiste, line 107: Could `element` ever be `null` or `undefined` and `update` not run again, such that the event never gets dispatched?

dbatiste: Or, what if the async task immediately returns a resolve promise and the state goes from pending -> success?  Could that happen before this `update` method runs, where dispatching the event relies on the state change?  I'm not sure what the behaviour of that was before.

dlockhart: I think that's what's happening in Firefox & legacy-Edge in the failing tests. :(

dbatiste (approved):

PR #1569, opened 2021-08-05
Daylight: Text & Textarea

margaree:


components/inputs/docs/input-text.md
```diff
@@ -60,43 +141,134 @@ To make your usage of `d2l-input-text` accessible, use the following properties
...
-**Events:**
+## Applying styles to native text input
 
-The `d2l-input-text` dispatches the `change` event when an alteration to the value is committed (typically after focus is lost) by the user. To be notified immediately of changes made by the user, use the `input` event.
+As an alternative to using the `<d2l-input-text>` custom element, you can style a native text input inside your own element. Import `input-styles.js` and apply the `d2l-input` CSS class to the input:
 
-**Slots:**
-* `left`: Slot within the input on the left side. Useful for an `icon` or `button-icon`.
-* `right`: Slot within the input on the right side. Useful for an `icon` or `button-icon`.
+```javascript
```
GeorgePlukov, line 148: Does using `javascript` here have the correct code styling appear in the demo? I know I exposed a language property from the `demo-snippet` component but I'm not sure if its wired up to work with fences like this

GeorgePlukov (approved):

components/inputs/docs/input-text.md
margaree: Ah that's a good point. It only really adds colour to the html pieces in the code, but not the rest. I'll look into wiring that up.

PR #1571, opened 2021-08-05
US129220 Add mobile-tray to d2l-input-time component

annabelsegalld2l:
Added the bottom mobile tray to d2l-input time component, to match the date-picker. 
- Added visual diff tests to show the tray for both input-time and dropdown-menu
- I needed to alter the focus in the menu logic to ensure that the button is tabbable. Otherwise, there seems to be a focus loop and the "close" button can not be tabbed to. I've added a comment at that line so we can remember the reasoning in the future.
- I also made the header/footer for the tray a minimum height of 5px when there is no content. The auto footers defaulted to be invisible, which is why the rounded corners were not showing up with the squared menu items


components/inputs/input-time.js
```diff
@@ -177,6 +180,9 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
 				:host([hidden]) {
 					display: none;
 				}
+				:host([data-mobile]) .d2l-input-time-menu {
+					text-align: center;
```
annabelsegalld2l, line 184: Centering the times in the mobile mode so they aren't left-adjusted. Otherwise, the times look really weird on wider screens with a bunch of empty space.

components/dropdown/dropdown-menu.js
```diff
@@ -69,6 +70,13 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 	}
 
 	_onFocus() {
+		if (this._useMobileStyling && !this.noMobileCloseButton && this._menuFocused) {
+			// if menu is focused and tab is clicked, move focus to the close button
+			this._menuFocused = false;
+			this.shadowRoot.querySelector('d2l-button').focus();
+			return;
+		}
+		this._menuFocused = true;
```
dlockhart, line 79: I'm still a bit curious why this was necessary. Once focus moves into the menu, [it manages its own key presses](https://github.com/BrightspaceUI/core/blob/main/components/menu/menu.js#L260), but I don't see anything in there that should be suppressing the TAB key. So TAB when in the menu should move focus to the next focusable element beyond it, which _should_ be the button in this case. But maybe there's a bug in there somewhere?

components/inputs/input-time.js
dlockhart: I wonder if this is something menus (or in this case dropdown-menus) should be doing themselves? That way any menu that goes into a tray would have centred items. Maybe a question for Jeff.

components/inputs/input-time.js
```diff
@@ -306,8 +326,9 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
 					@d2l-dropdown-close="${this._handleDropdownClose}"
 					@d2l-dropdown-open="${this._handleDropdownOpen}"
 					no-padding-footer
-					max-height="${ifDefined(this.maxHeight)}"
-					min-width="195">
+					max-height="${this._dataMobile ? null : ifDefined(this.maxHeight)}"
```
dlockhart, line 329: Again, should `d2l-dropdown-menu` maybe just be ignoring its `max-height` if `mobile-tray` is set and it's put into mobile mode?

components/dropdown/dropdown-menu.js
annabelsegalld2l: I think there is some sort of bug between the dropdown menu and the mobile tray specifically. In non-mobile mode, you can tab the button normally and exit out of the dropdown. In mobile, there is some kind of loop happening where you can only tab inside the dropdown without clicking escape. This seemed to fix it by overriding the focus to the close button... not sure what the underlying cause is for specifically the tray to have this issue ( and only with menu, trays without menus work fine).

components/inputs/input-time.js
annabelsegalld2l: I agree with you on the text alignment, I'll ask Jeff about that. And maybe we might want it for bottom trays only, since left and right aren't supposed to be symmetrical anyways. Not sure about ignoring the maximum height for menus with trays, there may be some scenario where a consumer wants a short bottom tray on purpose? I'll start a discussion on slack 😄

dlockhart: Cool sounds good!

components/dropdown/dropdown-menu.js
```diff
@@ -22,11 +22,20 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
...
 		this.noAutoFocus = true;
 		this.noPadding = true;
+		this._menuFocused = false;
+		this._maxHeightNonTray = this.maxHeight;
 	}
 
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
+		this._maxHeightNonTray = this.maxHeight;
```
annabelsegalld2l, line 32: I decided to ignore the max-height for all menus in trays. For cases like the user profile card, we do want to set that maximum height in a tray (since the card is quite "short" and would look odd with the extra space), so I decided to keep the max-height in the dropdown mixin.

components/inputs/input-time.js
annabelsegalld2l: I am actually thinking to keep the centering here in input time rather than a menu default setting. I realized that the filter, which uses a menu, wants a left-centered list of items as per the spec. I think it makes sense to keep that center formatting in the time picker only for now.

components/dropdown/dropdown-menu.js
```diff
@@ -68,11 +77,19 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 		menu.show({ preventFocus: true });
 	}
 
-	_onFocus() {
+	_onFocus(e) {
+		if (e.srcElement === this) return;
```
annabelsegalld2l, line 81: I've now got a hypothesis for why adding the close button causes that focus loop! The close button is NOT in the footer slot, so its focus event bubbles up the DOM to the dropdown menu when the button gets focused on. This is picked up by the onFocus method of dropdown-menu , which proceeds to override the focus to the menu, causing that loop. We need to ignore focus events coming from the dropdown element. I tested the existing dropdown menus with this change, and this return case is only hit with the close button.

dlockhart: Nice, that makes sense!

components/dropdown/dropdown-menu.js
```diff
@@ -22,11 +22,20 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 		super();
 		this.noAutoFocus = true;
 		this.noPadding = true;
+		this._menuFocused = false;
```
dlockhart, line 25: I think `_menuFocused` can go away now?

components/dropdown/dropdown-menu.js
```diff
@@ -98,6 +115,7 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 
 		if (this.__applyFocus) {
 			menu.focus();
+			this._menuFocused = true;
```
dlockhart, line 118: Ditto here setting it.

components/inputs/test/input-time.visual-diff.js
```diff
@@ -10,7 +10,7 @@ describe('d2l-input-time', () => {
 
 	before(async() => {
 		browser = await puppeteer.launch();
-		page = await visualDiff.createPage(browser, { viewport: { width: 300, height: 800 } });
+		page = await visualDiff.createPage(browser, { viewport: { width: 650, height: 800 } });
```
annabelsegalld2l, line 13: Needed to change the width of the viewport to 650px > 616px to get the normal styling. This caused the text in the time-picker footer to expand. Those are the only visual diff changes in the time-picker screenshots.

dlockhart (approved):
Looks great!

PR #1574, opened 2021-08-06
Tooltip doc tweaks

margaree:
Related PR https://github.com/BrightspaceUI/documentation/pull/270

Also tweaked input-text to use correct method formatting.

### DIFF
--- a/components/tooltip/README.md
+++ b/components/tooltip/README.md
@@ -12,10 +12,17 @@ Tooltips display additional information when users focus or hover on a point of 
   import '@brightspace-ui/core/components/button/button.js';
   import '@brightspace-ui/core/components/tooltip/tooltip.js';
 </script>
-
+<script>
+  window.addEventListener('load', function () {
+    setTimeout(function() {
+      var tooltip = document.querySelector('#tooltip');
+      tooltip.showing = true;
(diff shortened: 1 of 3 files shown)


components/tooltip/README.md
```diff
@@ -112,20 +119,35 @@ This can be done with the `boundary` attribute that allows any of the tooltip's
...
+    padding-top: 60px;
+    position: relative;
+    width: 350px;
+  }
+</style>
+
+<div class="offset-parent boundary">
+  <d2l-button id="tooltip-boundary">Tooltip boundary</d2l-button>
+  <d2l-tooltip for="tooltip-boundary"
+    boundary="{&quot;top&quot;:0, &quot;bottom&quot;:0, &quot;left&quot;:0, &quot;right&quot;:0}">
```
GeorgePlukov, line 149: Do the `&quot` get rendered as `"` in the text view? I can play around with this a bit to see if there a better way to write it

margaree: They don't, they get rendered as is. Based on the demo, this seems to be how we want this to be used so my thought is we would want to show it in the text like this

GeorgePlukov (approved):

Issue #1544, opened 2021-07-28
404 errors thrown in console (dev only; not prod) related to `en-us` language


jeroach:
404 errors are being thrown in the console for `dev` builds when loading web components that make use of the dynamic mixin.

Slack thread for context: https://app.slack.com/client/T025EV233/C0PHG3QB0

Closed PR which attempted to solve the issue: https://github.com/BrightspaceUI/core/pull/1518

PR #1576, opened 2021-08-06
localize-dynamic-mixin: Don't request unsupported langpacks in dev

bearfriend:
Fixes https://github.com/BrightspaceUI/core/issues/1544

### DIFF
--- a/mixins/localize-dynamic-mixin.js
+++ b/mixins/localize-dynamic-mixin.js
@@ -2,7 +2,7 @@ import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js
 import { LocalizeMixin } from './localize-mixin.js';
 
 const fallbackLang = 'en';
-const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'es', 'es-es', 'fi', 'fr', 'fr-fr', 'fr-on', 'hi', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-cn', 'zh-tw'];
+const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'es', 'es-es', 'fr', 'fr-fr', 'fr-on', 'hi', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-cn', 'zh-tw'];
 
 export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(superclass) {
 
--- a/mixins/localize-dynamic-mixin.js
+++ b/mixins/localize-dynamic-mixin.js
@@ -2,6 +2,7 @@ import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js
 import { LocalizeMixin } from './localize-mixin.js';
 
 const fallbackLang = 'en';
+const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'es', 'es-es', 'fi', 'fr', 'fr-fr', 'fr-on', 'hi', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-cn', 'zh-tw'];
 
 export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(superclass) {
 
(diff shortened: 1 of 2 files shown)


jeroach (approved):
Thanks Daniel :)

mixins/localize-dynamic-mixin.js
```diff
@@ -7,6 +7,12 @@ export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(su
 
 	static async getLocalizeResources(langs, { importFunc, osloCollection }) {
 
+		// in dev, don't request unsupported langpacks
+		if (!importFunc.toString().includes('switch')) {
+			const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'es-es', 'es', 'fi', 'fr-fr', 'fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh-cn'];
```
dlockhart, line 12: I wonder if we could somehow leverage [these lists of supported locales](https://github.com/BrightspaceUI/intl/blob/main/lib/common.js#L3) from Intl... they're not exactly what you need though unfortunately. 🤔

mixins/localize-dynamic-mixin.js
```diff
@@ -7,6 +7,12 @@ export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(su
 
 	static async getLocalizeResources(langs, { importFunc, osloCollection }) {
 
+		// in dev, don't request unsupported langpacks
+		if (!importFunc.toString().includes('switch')) {
```
dlockhart, line 11: Seems _slightly_ fragile... maybe a couple unit tests to cover both cases so we don't accidentally break this in the future?

mixins/localize-mixin.md
```diff
@@ -83,7 +83,9 @@ class MyComponent extends LocalizeDynamicMixin(LitElement) {
 }
 ```
 
-If your build system does not support variable dynamic imports, you'll need to manually set up imports for each language:
+When using this method, depending on various user settings, it's possible that a language file that does not exist will be requested, resulting in a network error (404). In production, your build system should prevent this by transpiling the variable dynamic import. BSI does this for you.
```
dlockhart, line 86: I'd rather not reference BSI in core if we can avoid it, since core is really more uh... "core" than BSI and could be used by things other than Brightspace. BSI also may not always be around (fingers crossed).

```suggestion
When using this method, depending on various user settings, it's possible that a language file that does not exist will be requested, resulting in a network error (404). In production, your build system should prevent this by transpiling the variable dynamic import into a `switch` statement.
```

mixins/localize-dynamic-mixin.js
svanherk: Similarly, we also have https://github.com/BrightspaceUI/actions/blob/master/prepare-translations-for-npm/prepare-translations.js#L4, https://github.com/BrightspaceUI/create/blob/main/src/generators/localization/templates/configured/_element.serge.json, https://github.com/Brightspace/serge-localize/blob/master/static/fra.common.serge.inc#L7... probably not for this PR to solve, but seems we could try to localize this to one source of truth.  For example, does each repo really need to define all this in their serge config, or can we just force a standard naming convection?  And could the `create` repo dynamically generate all the `lang` files based on the most up-to-date list? etc.

mixins/localize-dynamic-mixin.js
```diff
@@ -2,11 +2,17 @@ import { getLocalizeOverrideResources } from '../helpers/getLocalizeResources.js
 import { LocalizeMixin } from './localize-mixin.js';
 
 const fallbackLang = 'en';
+const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'es', 'es-es', 'fi', 'fr', 'fr-fr', 'fr-on', 'hi', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-cn', 'zh-tw'];
```
devpow112, line 5: Do we need `fi`?

dlockhart:
@bearfriend @svanherk I'm wondering if we can resurrect this PR? Now that core is using the `LocalizeDynamicMixin`, other repos that consume core in their tests/demos have pages with a default language of `en-us`, which causes an extra 404 to appear in logs.

mixins/localize-dynamic-mixin.js
svanherk: If My Courses (which actually has a `fi` file with real translations) ever moved to use the dynamic mixin, would that just mean you can't test `fi` locally (but in production it'd work fine)?

dlockhart: Yeah exactly.

devpow112: I'm assuming my courses should probably remove their `fi`.

devpow112: Interesting. Feels weird to have `fi` support for some components/clients but not all.

svanherk (approved):
I still want to try to get down to one source of truth for the languages we support, but until that happens this looks good to me

PR #1582, opened 2021-08-09
upgrading to eslint-config-brightspace version 0.16

dlockhart:
Now that the `ignorePackages` is in version 0.16, this upgrades core to it. I made all the fixes previously, but this reverts the `visual-diff` one since it's no longer needed with that option.

### DIFF
--- a/components/tooltip/test/tooltip-helper.js
+++ b/components/tooltip/test/tooltip-helper.js
@@ -1,4 +1,4 @@
-import { oneEvent } from '@brightspace-ui/visual-diff/helpers/index.js';
+import { oneEvent } from '@brightspace-ui/visual-diff/helpers';
 
 export async function hide(page, selector) {
 	const hideEvent = oneEvent(page, selector, 'd2l-tooltip-hide');
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,35 +5,35 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.12.13",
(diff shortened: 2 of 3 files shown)


components/tooltip/test/tooltip-helper.js
```diff
@@ -1,4 +1,4 @@
-import { oneEvent } from '@brightspace-ui/visual-diff/helpers/index.js';
+import { oneEvent } from '@brightspace-ui/visual-diff/helpers';
```
svanherk, line 1: Is this because it _knows_ it's a package due to the file name being `index.js`?  Or is it just checking that it's not a relative path and `import {customElement} from 'lit/decorators'` would work as well?

dlockhart: Hmm, good question. Looking [at the source](https://github.com/import-js/eslint-plugin-import/blob/master/src/rules/extensions.js#L160), it appears to just treat anything starting with `@something` or any characters (i.e. not a `../` relative path) as a module.

So... that's not good enough actually, because `import { ifDefined } from 'lit-html/directives/if-defined';` both doesn't get caught by the linter and also doesn't work in Lit 2.0.

Back to the drawing board I think. Like this change still catches _some_ of the issues, just not all of them. So I think we can leave it in the eslint config for now and fix these places in core... but we'll need something more.

svanherk (approved):
Looks good, agree that we might as well leave this in for now to catch some of the issues

PR #1585, opened 2021-08-10
US130430 - Batch filter change events

svanherk:
This is helpful for:
- When someone is clicking around quickly in filter and undoing changes they just did
- For bulk events (clearing and select all)

If the same filter is changed multiple times within a batch, only the final state is sent to minimize work for the consumer.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -112,6 +112,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	constructor() {
 		super();
 		this.disabled = false;
+		this._changeEventsToDispatch = new Map();
 		this._dimensions = [];
 		this._openedDimensions = [];
 		this._totalAppliedCount = 0;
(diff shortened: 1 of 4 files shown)


components/filter/filter.js
```diff
@@ -347,7 +360,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			this._totalAppliedCount--;
 		}
 
-		this._dispatchChangeEvent({ dimension: dimensionKey, value: { key: valueKey, selected: selected } });
+		this._dispatchChangeEvent(`${dimensionKey}-${valueKey}`, { dimension: dimensionKey, value: { key: valueKey, selected: selected } });
```
svanherk, line 363: This key value will be different for each type of filter dimension.  For example - for date filter, the dimension key alone will probably be enough.

components/filter/filter.js
```diff
@@ -301,8 +302,20 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+		if (!this._changeEventTimeout) {
+			this._changeEventTimeout = setTimeout(() => {
+				this.dispatchEvent(new CustomEvent('d2l-filter-change', {
+					bubbles: true,
+					composed: false,
+					detail: { changes: Array.from(this._changeEventsToDispatch.values()) }
+				}));
+				this._changeEventsToDispatch = new Map();
+				this._changeEventTimeout = null;
+			}, 200);
```
svanherk, line 317: This 200 value was based on what "felt" good - nothing real scientific.  This was about the fastest I could click two separate filters to have them grouped, or clicking/unclicking the same option about two or three times.  Longer would be more performant (especially if someone keeps toggling the same option), but might have a noticeable delay on the user side before anything starts loading.

The _main_ purpose here is to batch the clear/select all case anyways, so we could go lower if we were worried.

dlockhart (approved):
Looks good!

PR #1588, opened 2021-08-12
Allow the d2l-selection-select-all to be disabled

svanherk:
Just exposing the `disabled` ability on the `input-checkbox`.  I want to set this to disabled in the filter when the options are loading - any concerns?


dlockhart (approved):
Makes sense to me!

PR #1590, opened 2021-08-13
US130429 - Add select all to filter

svanherk:
Adding the ability to turn on a Select All checkbox in the header for the set dimension.

image
image

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -10,6 +10,8 @@ import '../list/list-item.js';
 import '../loading-spinner/loading-spinner.js';
 import '../menu/menu.js';
 import '../menu/menu-item.js';
+import '../selection/selection-select-all.js';
+import '../selection/selection-summary.js';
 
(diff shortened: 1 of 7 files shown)


components/filter/filter.js
```diff
@@ -222,9 +239,23 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			</d2l-input-search>
 		`;
```
svanherk, line 241: The select all checkbox and summary hides when we search, and is disabled if we're loading or there are no values.  When there are no values selected, we want it to say "Select All" instead of the default "0 selected".

components/filter/test/filter.test.js
```diff
@@ -29,16 +29,18 @@ describe('d2l-filter', () => {
 	});
 
 	describe('loading', () => {
```
svanherk, line 31: Enhanced the loading test because we can't easily test that with visual diff, but the rest will be tested in the visual diff tests or is tested in the selection component tests

dlockhart (approved):
LGTM!

PR #1591, opened 2021-08-13
Better cleanup of some leaking menu-item styles

svanherk:
Looked into this a bit more after @dlockhart saw a weird axe flake.  Went with a more complete cursor fix and a more consistent colour fix.


dlockhart (approved):
👍

PR #1594, opened 2021-08-16
Daylight: status indicator fixes

GeorgePlukov:
Adds two section from design review, updates some other minor issues on the page

### DIFF
--- a/components/status-indicator/README.md
+++ b/components/status-indicator/README.md
@@ -1,4 +1,4 @@
-## Status Indicators
+# Status Indicators
 Status Indicators are used to communicate the status of an item. They are non-interactive and assert prominence on state.
 
 <!-- docs: start hidden content -->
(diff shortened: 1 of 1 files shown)


components/status-indicator/README.md
```diff
@@ -1,4 +1,4 @@
-# Status Indicators
+## Status Indicators
```
margaree, line 1: Shouldn't this stay an h1?

components/status-indicator/README.md
```diff
@@ -65,6 +65,139 @@ Status Indicators are used to communicate the status of an item. They are non-in
...
+    <d2l-status-indicator state="success" text="complete" bold></d2l-status-indicator>
+    <d2l-status-indicator state="alert" text="overdue" bold></d2l-status-indicator>
+    <d2l-status-indicator state="none" text="closed" bold></d2l-status-indicator>
+  </div>
+</div>
+```
+The state is used to apply a meaningful colour to the status indicator to assist with scannability. The visible label should still have a clear meaning so that users who can't see the colour can still understand the state.
+
+* Default - the state is important but not urgent 
+* Success - the state is considered positive or complete.
```
margaree, line 122: nit: should be consistent about ending these lines with periods or not

components/status-indicator/README.md
```diff
@@ -65,6 +65,139 @@ Status Indicators are used to communicate the status of an item. They are non-in
...
+		</div>
+	</div>
+
+</div>
+```
+
+The text label should be kept short; one or two words at most. If more information is needed to explain the state or give extra context, the information should be included after the label and should use the same color to associate the information with the state. When placing additional text on the same line as the status indicator, ensure that the bottom of the text is even across all elements.
+
 <!-- docs: start hidden content -->
 ### Variants
```
margaree, line 202: Should we get rid of this section given that the one above explains it so it just adds duplication?

margaree (approved):
Thoughts on if the hidden Properties section should be moved to before the new `### Bold` section, since so much content was added it becomes a bit buried?

GeorgePlukov:
> Thoughts on if the hidden Properties section should be moved to before the new `### Bold` section, since so much content was added it becomes a bit buried?

I think this makes sense!

margaree (approved):

PR #1595, opened 2021-08-19
fix: set seconds on manual EOD to 59

mdulat:
Fixes https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fdefect%2F605123292669&fdp=true

The issue is that when an EOD time (i.e. 11:59 PM) is entered manually, the seconds value is set to 0, and when the default value of END_OF_DAY is used, the seconds value is set to 59. The causes problems when comparing times.

### DIFF
--- a/components/inputs/input-time.js
+++ b/components/inputs/input-time.js
@@ -224,8 +224,22 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
 		if (this.enforceTimeIntervals) {
 			time = getTimeAtInterval(this.timeInterval, time);
 		}
+
+		const isManualEndOfDay = time.getHours() === 23 && time.getMinutes() === 59 && time.getSeconds() === 0;
+		if (isManualEndOfDay) {
+			time = END_OF_DAY;
+		}
+
 		this._value = formatDateInISOTime(time);
 		this._formattedValue = formatTime(time);
+
+		if (isManualEndOfDay) {
+			this.dispatchEvent(new CustomEvent(
+				'change',
+				{ bubbles: true, composed: false }
+			));
+		}
+
 		this.requestUpdate('value', oldValue);
 	}
 


components/inputs/input-time.js
```diff
@@ -224,8 +224,22 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
...
+		const isManualEndOfDay = time.getHours() === 23 && time.getMinutes() === 59 && time.getSeconds() === 0;
+		if (isManualEndOfDay) {
+			time = END_OF_DAY;
+		}
+
 		this._value = formatDateInISOTime(time);
 		this._formattedValue = formatTime(time);
+
+		if (isManualEndOfDay) {
+			this.dispatchEvent(new CustomEvent(
```
mdulat, line 237: This is necessary so that any currently saved EOD times with the seconds value of 0 must be changed to have seconds = 59, and that change must be dispatched so the client can be updated. Otherwise we have the same problem of compared times potentially being 59 seconds off, but in reverse of the original issue.

dlockhart: Hmm... I don't love putting that logic here, where it's going to live forever. Also, we try not to dispatch `change` events like this unless there was a user interaction, so it's weird to do so only in this one scenario.

I'm wondering if you might be able to make a change to the backend code that's saving these values? If it sees 11:59:00 when an activity is saved, it could automatically convert it to 11:59:59 maybe?

dlockhart:
Could you open this from a branch in the repo instead of a fork? That way the visual diff and unit tests will have access to the repo tokens. You should have Write access as a member of the BrightspaceUI organization.

components/inputs/input-time.js
mdulat: I can do it on the backend, but won't this problem come up everywhere this component is used? I thought it would be better to fix it in one place.

mdulat:
replaced by https://github.com/BrightspaceUI/core/pull/1598

components/inputs/input-time.js
dlockhart: Maybe I'm misunderstanding the issue. This is client-side validation (or is it client and server?) that's comparing a manually entered 11:59:00 to a chosen "end of day" (11:59:59) time elsewhere on the page (due date vs. end date) right?

IMO it seems pretty specific to FACE having multiple date/times and validating that they come before each other.

Another option would be to fix it backend before the page loads (instead of after saving), which would prevent validation errors happening right away.

And then yet another option would be to write a migration script to run on the database that would fix all the values. Advantage there is it's a one-time fix that'll resolve it for everyone, no matter whether they visit the edit page or not. Disadvantage would be... hmm yeah it might be hard to actually know from the UTC dates in the DB whether they intended it to be 11:59 or not, since that's dependant on whoever created the activity's timezone.

mdulat: The validation is done on the server, but I can have the client fix the times before sending them for validation, then actually save the fixed values when the activity is saved. If this is really just a FACE thing (comparing multiple dates on a page) then seems reasonable to put the fix in there.

PR #1596, opened 2021-08-19
Prevent switch from firing change event when on value is set programatically 

GeorgePlukov:
Firing an event on any update to the on value is causing an issue in the daylight site demo snippet that makes it difficult to clear fields by setting them back to a default value programmatically. This change will fire the event when toggling the state of the switch and not any time the switch on value is changed.

Also adds a variable to control the background color of the switch (as it needs to be overridden in the daylight site)


components/switch/switch-mixin.js
```diff
@@ -50,7 +50,7 @@ export const SwitchMixin = superclass => class extends RtlMixin(FocusVisiblePoly
 			}
 
 			.d2l-switch-container {
-				background-color: #ffffff;
+				background-color: var(--d2l-switch-container-background-color, '#fff');
```
dlockhart, line 53: Could you add a basic visual diff test to exercise this variable? Just so it can fail if we ever accidentally break this.

dlockhart: I don't think we need to document it though since we don't really want folks to know about it. 🤫

components/switch/test/switch.test.js
```diff
@@ -9,9 +9,10 @@ describe('d2l-switch', () => {
 		runConstructor('d2l-switch');
 	});
 
-	it('dispatches change event', async() => {
+	it.only('dispatches change event', async() => {
```
dlockhart, line 12: I've done this so many times lol...

```suggestion
	it('dispatches change event', async() => {
```

GeorgePlukov: Nice catch! 😮‍💨

dlockhart (approved):

PR #1597, opened 2021-08-19
Daylight: Select Lists: add demo to top of page

margaree:
I seem to have missed adding a demo at the top of this page. It is currently the same demo as the one at the bottom, which might be why I left it off, but the page does feel empty without it.

### DIFF
--- a/components/inputs/docs/input-select-styles.md
+++ b/components/inputs/docs/input-select-styles.md
@@ -2,6 +2,14 @@
 
 A Select List allows the user to select a single option out of a relatively large number of items, or to reduce the visual prominence of an option selection.
 
+<!-- docs: demo -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/inputs/demo/input-select-test.js';
+</script>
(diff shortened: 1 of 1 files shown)


components/inputs/docs/input-select-styles.md
```diff
@@ -2,6 +2,14 @@
 
 A Select List allows the user to select a single option out of a relatively large number of items, or to reduce the visual prominence of an option selection.
 
+<!-- docs: demo -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/inputs/demo/input-select-test.js';
+</script>
+<d2l-test-input-select></d2l-test-input-select>
```
dlockhart, line 10: Wondering if the demos might be more useful if they showed the "applying styles to native select elements" technique? Just because showing a test element doesn't really convey anything.

GeorgePlukov (approved):

margaree:
Will reopen as a new PR with changes to this and other pages with demos that involved the applying styles to native elements.

PR #1600, opened 2021-08-19
Add d2l-dialog-abort event to dialog-mixin

bearfriend:


dlockhart:
You'll want to rebase from `main` as we've removed Browserstack tests. Also there might be a good place to add a unit test for this now that we're more confident in the approach.

jorge-ramirez-arredondo:
So, I was doing some local testing for this and I'm noticing that in some cases, `_close` gets called before `_handleClose` and in some cases its the opposite.
It looks like this is due to the native `dialog` triggering its own `cancel` event when the `esc` key is not caught by our own key handlers.

Would it make sense to add a listener to the dialog `cancel` event using our own `_close` instead of the default behaviour there?
Ex:
```
_handleCancel(e) {
	// Override the default logic for closing the native dialog with our own
	e.preventDefault();
	this._close(abortAction);
}
```
```
<dialog
	...
	@cancel="${this._handleCancel}"
></dialog>
```

This way, I think `_close` will consistently be called before `_handleClose`.

dlockhart:
> Would it make sense to add a listener to the dialog cancel event using our own _close instead of the default behaviour there?

I think that makes sense, and is a nice behaviour anyway to ensure that all the internal state and other logic in our own close is applied consistently in the same order in the native vs. not case.

dbatiste:
> Hmm, in what scenarios does the `close` event trigger when the `cancel` event doesn't?

Based on the docs, it sounds like the `cancel` event will only run when pressing Esc, or a button with `type="reset"`.  Everything else would be relying on `close` event.

I am a bit concerned that the dismissible stack can could get out of step when the abort/dismiss is cancelled.  We may need further changes to address that.

dbatiste:
I think bullets 2 & 3 can be summarized as Esc.  We prevent the default behaviour of native dialogs, and we normalize the behaviour by handling Esc through the dismissible helper regardless of native vs custom impl.

There are also the cases where the consumer removes the `opened` attribute or sets the `opened` property to `false`, or the user clicks on a dialog button which has the `data-dialog-action` attribute.  I'm not sure what happens if a consumer uses `<button type="reset">` as the docs mention that results in a `cancel` event like Esc.  The docs sort of imply that button in a native dialog will close the dialog, but I don't think we've ever explored that.

dlockhart:
This is looking pretty stale. @bearfriend @jorge-ramirez-arredondo is this still something that's needed, or can we close this off?

dlockhart:
Closing for now due to age -- feel free to re-open later if necessary.

jorge-ramirez-arredondo:
Sorry for leaving this hanging for so long. I should've mentioned that I ended up implementing a workaround instead of this due to time constraints.
I'd still like to get back to this at some point, since I think something like this makes sense to have (for the sake of making the open state fully controllable), but without an actual use-case to test it with, it's hard to justify working on it now. I'll re-open this if I run into another scenario where this makes sense to have.

PR #1603, opened 2021-08-20
Daylight: Meter review fixes

GeorgePlukov:
Updates demos to use different values and text property and changes a bunch of the copy text

### DIFF
--- a/components/meter/README.md
+++ b/components/meter/README.md
@@ -113,16 +113,6 @@ Circle meters display data in a compact circle format, so they're useful when ho
 <d2l-meter-circle value="1" max="6"></d2l-meter-circle>
 ```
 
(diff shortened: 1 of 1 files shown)


components/meter/README.md
```diff
@@ -139,6 +123,21 @@ Circle meters will show the progress as a full circle.
 	* `\{\%\}` in the string will be replaced with percentage value
 	* `{x/y}` in the string will be replaced with fraction with the proper language support
 
+## Light Foreground
```
margaree, line 126: Looks like this would now be hidden. Is that what's desired?

GeorgePlukov: Think I put it one block too low, nice catch thanks!

margaree (approved):

PR #1604, opened 2021-08-20
US130990 - Add support for reading context object from the document root

grant-cleary:
Because we're fetching more complicated checks (outside of a feature flag), we need to be able to pass more information into the component than we did before. This PR hooks up the context look-up we need to do and adjusts the existing renderers to extract the information they need out of the context object.

Because we also need the host in the case of FRAs (specifically for @mentions URL resolution), we unfortunately still do need ifrau for this. :(


components/html-block/html-block.js
```diff
@@ -3,6 +3,43 @@ import { css, LitElement } from 'lit-element/lit-element.js';
 import { htmlBlockMathRenderer } from '../../helpers/mathjax.js';
 import { requestInstance } from '../../mixins/provider-mixin.js';
 
+let contextPromise;
+
+const getContext = () => {
+	if (contextPromise) return contextPromise;
+
+	contextPromise = new Promise(async resolve => { // eslint-disable-line no-async-promise-executor
+		const context = JSON.parse(document.documentElement.getAttribute('data-hb-context'));
```
grant-cleary, line 12: I recently added a new feature to ifrau to sync html element data attributes automatically. For FRAs to take advantage of this particular feature within this component, they'll need to update ifrau to v0.36.0.

components/html-block/html-block.js
```diff
@@ -3,6 +3,43 @@ import { css, LitElement } from 'lit-element/lit-element.js';
...
+		if (window.ifrauclient) {
+			const ifrauClient = await getIfrauClient();
+			context.host = await ifrauClient.request('valenceHost');
```
grant-cleary, line 15: I wish I could not use ifrau at all, but @mentions requires links to profile cards within the LMS, which are unfortunately LMS routes and thus within FRAs definitely need to have the host pre-pended.

Right now @mentions isn't available in any FRAs, but that may change.

helpers/mathjax.js
```diff
@@ -21,6 +25,9 @@ export function loadMathJax(mathJaxConfig) {
 	if (mathJaxLoaded) return mathJaxLoaded;
 
 	window.MathJax = {
+		chtml: {
+			scale: (mathJaxConfig && mathJaxConfig.outputScale) || 1
```
grant-cleary, line 29: This has been supported in the LMS for some time now; it just wasn't propagated here originally.

components/html-block/html-block.js
grant-cleary: Also note: This context object isn't set in the LMS yet. I'm just trying to lay down the groundwork in the components first.

grant-cleary:
Crap, forgot to run linting. 🤦

components/html-block/html-block.js
grant-cleary: One thing maybe worth considering is that if we slapped the `valenceHost` as a data attribute on the host, we could use the new marshaling to pull that into the iframe without needing to explicitly invoke the client.

components/html-block/html-block.js
```diff
@@ -3,6 +3,43 @@ import { css, LitElement } from 'lit-element/lit-element.js';
 import { htmlBlockMathRenderer } from '../../helpers/mathjax.js';
 import { requestInstance } from '../../mixins/provider-mixin.js';
 
+let configuredIfrauClient;
+
+const getIfrauClient = async() => {
+	if (configuredIfrauClient === undefined && window.ifrauclient) {
+		configuredIfrauClient = await window.ifrauclient({
```
dbatiste, line 10: I think there is a "slim" client that we should be looking into using for our core components that require such a check.  I don't think it's documented though, so I think we might need to look into the code a little to figure out what exactly it does, but I think it is meant to avoid some of the issues we've had with client's clobbering each other.

components/html-block/html-block.js
dbatiste: What does this yield if the data attribute doesn't exist?  Should we be defaulting to an empty object?

components/html-block/html-block.js
```diff
@@ -3,6 +3,43 @@ import { css, LitElement } from 'lit-element/lit-element.js';
...
+
+	return configuredIfrauClient;
+};
+
+let contextPromise;
+
+const getContext = () => {
+	if (contextPromise) return contextPromise;
+
+	contextPromise = new Promise(async resolve => { // eslint-disable-line no-async-promise-executor
```
dbatiste, line 23: Seems like we should be able to avoid this by moving the ifrau bit out of this promise handler (i.e. move it above).

components/html-block/html-block.js
```diff
@@ -150,8 +187,9 @@ class HtmlBlock extends LitElement {
 				let temp = document.createElement('div');
 				temp.appendChild(fragment);
 
+				const context = await getContext();
 				for (const render of HtmlBlock._renderers) {
-					temp = await render(temp);
+					temp = await render(temp, context);
```
dbatiste, line 192: Hmmmm. I think it would be preferred if the renderer can get the context themselves, as opposed to depending on html-block passing in an object that contains everything it needs.  Was this something that was considered?

grant-cleary: Yeah, I didn't have any strong reason for doing it this way. It felt slightly cleaner to centralize things, but I figured you would have ideas on this when you got back, since I don't really know the long-term vision for this. I can easily make this change. :)

components/html-block/html-block.js
grant-cleary: Yeah, that's a good point. If we're only fetching the host we should be able to use that.

components/html-block/html-block.js
grant-cleary: Yeah, I haven't updated this PR since I made the changes to the HTML editor context stuff, but I was intending on propagating those changes here (or wherever we move the context fetching).

components/html-block/html-block.js
grant-cleary: Oh, you're right. I can do that in the editor too I think.

helpers/mathjax.js
```diff
@@ -1,11 +1,16 @@
 let mathJaxLoaded;
 
 export async function htmlBlockMathRenderer(elem) {
-	const isLatexSupported = window.D2L && window.D2L.LP && window.D2L.LP.Web.UI.Flags.Flag('us125413-mathjax-render-latex', true);
+	const context = JSON.parse(document.documentElement.getAttribute('data-mathjax-context')) || {};
```
grant-cleary, line 4: This provider hasn't been created yet, so an LMS PR will be coming to handle that shortly. This also introduces an implicit dependency on FRAs - specifically, FRAs who want this functionality will need to update ifrau to at least 0.36.0 or they won't be able to take advantage of the `data-` attribute syncing.

dbatiste (approved):

dmd2l (approved):

helpers/framed.js
```diff
@@ -21,7 +21,7 @@ export async function isFramed() {
 	framed = await Promise.race([
 		new Promise(resolve => {
 			const handleIsFramedResponse = evt => {
-				if (!evt || !evt.data || !evt.data.isFramed === undefined) return;
+				if (!evt || !evt.data || evt.data.isFramed === undefined) return;
```
grant-cleary, line 24: Ran into this while testing. Accidentally inverted this logic when I switched this from `!evt.data.isFramed` to an explicit `undefined` check. 😬

dbatiste: Maybe this was intended to be `if (!evt || !evt.data || !evt.data.isFramed)`?

dbatiste: Oh, that's what it used to be.  Was the explicit check needed?

grant-cleary: No, it just felt... More correct? The host page will only ever return `true` or `undefined`, but it still feels more correct to explicitly resolve to `false` if `isFramed` happens to be false somehow.

dbatiste: Sometimes I write it to be specific, but other times not, to be more tolerant of both `undefined` and `null` possible values.  It's not an issue as long a `null` is never a possible value, so more just wondering.

dbatiste (approved):

helpers/framed.js
grant-cleary: Fair enough. Yeah, this was a bit of a weird one.

grant-cleary:
Going to hold off on merging this until the LMS providers are merged; otherwise, this could break LaTeX rendering in other components that may be consuming the HTML block component.

PR #1606, opened 2021-08-23
Fix skeleton stacking context

bearfriend:
The skeleton `z-index` was set to `4` to give room for components to use `z-index` for other purposes without rendering on top of the sekeleton, but because there is no guaranteed stacking context up the tree, the skeleton itself can now render on top of things outside the host component that it shouldn't.

This PR creates a guaranteed stacking context on any host that uses skeletons, and _only while skeletons are visible_.

Using `opacity` to create the stacking context (again, only while the skeletons are visible) seems fairly safe, but there are other options like `transform: perspective(0)`, which seem pretty safe as well.

### DIFF
--- a/components/skeleton/skeleton-mixin.js
+++ b/components/skeleton/skeleton-mixin.js
@@ -10,6 +10,9 @@ export const skeletonStyles = css`
 		75% { background-color: var(--d2l-color-sylvite); }
 		100% { background-color: var(--d2l-color-sylvite); }
 	}
+	:host([skeleton]) {
+		opacity: .999;
+	}
 	:host([skeleton]) .d2l-skeletize::before {
 		animation: loadingPulse 1.8s linear infinite;
 		background-color: var(--d2l-color-sylvite);
@@ -20,7 +23,7 @@ export const skeletonStyles = css`
 		position: absolute;
 		right: 0;
 		top: 0;
-		z-index: 4;
+		z-index: 999;
 	}
 	@media (prefers-reduced-motion: reduce) {
 		:host([skeleton]) .d2l-skeletize::before {


bearfriend:
Here's an example of the issue.

Screen Shot 2021-08-23 at 1 13 35 PM

svanherk:
@dlockhart knows the most about the skeleton mixin, but he's off this week (FYI for when you're back).

Creating a stacking context here feels like a good idea, although I'm surprised the opt in flyout has such a low z-index and would be seeing this 🤔

svanherk:
Looks like [it's set to 900](https://github.com/BrightspaceUILabs/opt-in-flyout/blob/master/internal/flyout-impl.js).  Is it being placed inside something else with a lower `z-index` than 4?

components/skeleton/skeleton-mixin.js
```diff
@@ -20,7 +23,7 @@ export const skeletonStyles = css`
 		position: absolute;
 		right: 0;
 		top: 0;
-		z-index: 4;
+		z-index: 999;
```
svanherk, line 26: Checking I understand correctly - fixing the specific flyout issue is solved by setting the opacity and creating the stacking context.  This doesn't actually _need_ to raise here because `4` has worked so far (meaning everything implementing skeleton doesn't set its internal components that should skeletize to anything higher than that).  But because we have a stacking context now, we can set this to a much higher in case anything ever does?

bearfriend: That's exactly my thinking!

bearfriend:
> Looks like [it's set to 900](https://github.com/BrightspaceUILabs/opt-in-flyout/blob/master/internal/flyout-impl.js#L39). Is it being placed inside something else with a lower `z-index` than 4?

I took a cursory look at the flyout just to see if I could get it to stack overtop of the skeleton another way, but all I could see (in dev tools) was a `z-index: 10000` set on `.d2l-flyout-fixed`, so obviously there's another context in there somewhere that I couldn't determine.

Barring any unforeseen issues, this seems like the better approach anyway.

svanherk:
> Barring any unforeseen issues, this seems like the better approach anyway.

Yeah, handling this here seems like a win.  I imagine when/if the flyout is moved to `core` we'd take a careful look into the z-index settings and adjust it to play nicely with everything - I think there's other bugs with it and image-based nav.  But if skeleton can avoid ever being the issue, that feels like the right path!

Could you add a simple visual-diff test for this?  I'm thinking something that would fail if the opacity was removed, or set to 1 (I can help with questions if you haven't had to write one before)

dlockhart:
Great idea -- thanks for this!

PR #1612, opened 2021-08-27
fix: demo snippet styling

devpow112:
This allows elements that want to use scroll wrapper and overflow to display correctly in demo snippets. It will hide the overflow and allow scrolling with arrow buttons when `overflow-hidden` is supplied. This also makes the code view always be the same width as the demo snippet when it is part of the snippet. It will still have the maximum width by itself.

These changes were mainly done to support rubrics so we can switch to using the standard demo snippet. It was tested with core demo pages as well as rubrics.

| Before | After |
|---|---|
|image|image|


dlockhart (approved):
LGTM

PR #1616, opened 2021-08-30
Adding logic for dropdown mobile trays to work in Iframes/ non-responsive screens

annabelsegalld2l:
Currently, the trays work by being pinned to one side by using something like bottom: 0. However, this doesn't work in iframes, where the inner height and width of the window is actually the height and width of the iframe (much larger than the screen). In those cases, pinning to the bottom requires users to scroll to the bottom of the iframe, rather than having the tray pop up in the viewport. 

This PR leverages the existing IfrauDialogService to add/remove backdrops on iframed pages as needed. I've pulled out the logic into a helper, since it's identical to the dialog-mixin usage.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -4,6 +4,7 @@ import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
 import { forceFocusVisible, getComposedActiveElement, getNextFocusable, tryApplyFocus } from '../../helpers/focus.js';
 import { classMap } from 'lit-html/directives/class-map.js';
+import { getIfrauBackdropService } from '../../helpers/ifrauBackdropService.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { html } from 'lit-element/lit-element.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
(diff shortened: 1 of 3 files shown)


helpers/ifrauBackdropService.js
```diff
@@ -0,0 +1,9 @@
+export async function getIfrauBackdropService(ifrauBackdropService) {
+	if (!window.ifrauclient) return;
+	if (ifrauBackdropService) return ifrauBackdropService;
+
+	const ifrauClient = await window.ifrauclient().connect();
+	ifrauBackdropService = await ifrauClient.getService('dialogWC', '0.1');
```
dlockhart, line 6: I think you'll want to define `ifrauBackdropService` in this file instead of passing it in?

helpers/ifrauBackdropService.js
```diff
@@ -0,0 +1,9 @@
+export async function getIfrauBackdropService(ifrauBackdropService) {
```
dlockhart, line 1: Since this method can find nothing, should we call it something like `tryGetIfrauBackdropService()`? Then it can return `null` is there's no `ifrauclient` (instead of `undefined` right now).

components/dialog/dialog-mixin.js
```diff
@@ -105,16 +96,16 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		super.updated(changedProperties);
 		if (!changedProperties.has('opened')) return;
 
-		const dialogService = await getIfrauDialogService();
+		ifrauDialogService = await getIfrauBackdropService(ifrauDialogService);
```
dlockhart, line 99: Related to the comment down in the helper method, I think you can remove the local `let ifrauDialogService` at the top of this file and change this to:

```javascript
const ifrauBackdropService = tryGetIfrauBackdropService();
```

annabelsegalld2l: The reason I wanted to pass in and keep the ifrauDialogService is so that it only needs to be fetched once. The getIfrauBackdropService function will return the ifrauDialogService if it's already defined. I'm not sure if there would be a performance implication with fetching the service every time the dialog is opened. 🤔

dlockhart: Understood, but the helper itself can do that by having the local variable in its file. That way it can dedupe calls across dialogs and dropdowns (and anything else we add this to in the future).

annabelsegalld2l: Aha, makes sense! Will do.

dlockhart (approved):
Cool this LGTM, excited to see what the future refactoring looks like!

PR #1617, opened 2021-08-30
US126584 - Add selection-single to d2l-filter-dimension-set

svanherk:
This PR adds the ability to turn on `selection-single` for the set dimension, so that only one value can be selected from the list.  When the value selected changes, the batched change event will have two changes in the array - one turning on the new filter value, and one turning off the old filter value.  Therefore no special function/handling is needed for a `selection-single` list versus a multi-select one.

This PR doesn't adjust filter counts for this case.  After talking to Jeff, we decided we're going to remove the "All" language and keeping track of the total.  It has the potential to cause more confusion than help, and is not super intuitive in the single selection case anyways.  It also simplifies integrating the future `count-badge`.  This functionality will be removed in the next PR.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -316,7 +316,8 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			<d2l-list
 				id="${SET_DIMENSION_ID_PREFIX}${dimension.key}"
 				@d2l-list-selection-change="${this._handleChangeSetDimension}"
-				extend-separators>
+				extend-separators
+				?selection-single="${dimension.selectionSingle}">
 				${dimension.values.map(item => html`
 					<d2l-list-item
 						?hidden="${item.hidden}"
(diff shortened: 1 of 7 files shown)


components/filter/filter.js
```diff
@@ -496,7 +497,8 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			switch (type) {
 				case 'd2l-filter-dimension-set': {
 					info.searchType = dimension.searchType;
-					if (dimension.selectAll) info.selectAllIdPrefix = SET_DIMENSION_ID_PREFIX;
+					info.selectionSingle = dimension.selectionSingle;
+					if (dimension.selectAll && !dimension.selectionSingle) info.selectAllIdPrefix = SET_DIMENSION_ID_PREFIX;
```
svanherk, line 501: In the `selection-single` case, the select all checkbox and summary are hidden automatically.  But it seems like they still take up a bit of space, and I also want to hide my container with margins anyways

dlockhart (approved):

PR #1618, opened 2021-08-31
input-date: add open function

margaree:
In Daylight we want the input-date/time* demos to be open by default so this adds that functionality to `input-date`. If this pattern seems good then I'll add it to the other input-date/time components.


components/inputs/input-date.js
```diff
@@ -313,6 +313,14 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		if (this._textInput) this._textInput.focus();
 	}
 
+	async open() {
+		if (!this._dropdownFirstOpened) await this._handleFirstDropdownOpen();
+		this._dropdown.open();
+		await this._handleChange();
+		this._calendar.focus();
+		this._setFormattedValue();
+	}
```
dlockhart, line 322: What do you think about making this an `opened` attribute/property similar to how dialogs work? Then internally it could be refactored a bit to just use that.

IMO we should always be relying on properties to store state like this... it just makes the component so much easier to use vs. having to get a reference to it in JS and call a method on it.

margaree: Yup that makes sense to me

components/inputs/test/input-date.visual-diff.html
```diff
@@ -13,14 +13,29 @@
 
 			const elems = document.querySelectorAll('d2l-input-date');
 			elems.forEach(elem => elem._handleFirstDropdownOpen());
+
+			window.addEventListener('load', () => {
+				setTimeout(() => {
+					const dateInput = document.querySelector('#open-by-default');
+					dateInput.opened = true;
```
dlockhart, line 20: Could this one just have the `opened` attribute set in the HTML markup?

components/inputs/input-date.js
```diff
@@ -297,14 +301,18 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
+					this.setFormValue(this.value);
+					this.setValidity({
+						rangeUnderflow: !this.noValidateMinMax && this.value && this.minValue && getDateFromISODate(this.value).getTime() < getDateFromISODate(this.minValue).getTime(),
+						rangeOverflow: !this.noValidateMinMax && this.value && this.maxValue && getDateFromISODate(this.value).getTime() > getDateFromISODate(this.maxValue).getTime()
+					});
+					this.requestValidate(false);
+				}
+			} else if (prop === 'opened') {
+				if (this.opened) this._open();
+				else if (!this.opened && this._dropdown.opened) this._close();
```
dlockhart, line 315: Nit: the check for `this._dropdown.opened` could move into `_close()` just in case we want to call `_close()` from elsewhere maybe?

components/inputs/input-date.js
```diff
@@ -381,7 +393,6 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		setTimeout(() => {
 			this._textInput.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
 		}, 150);
-		this._dropdownOpened = true;
```
dlockhart, line 384: Guessing this isn't needed anymore since we're always the ones who open it now?

components/inputs/input-date.js
```diff
@@ -445,6 +448,21 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		this.focus();
 	}
 
+	async _open() {
+		if (this.disabled || this.skeleton) return;
```
dlockhart, line 452: So what happens if it's in initially `<d2l-dropdown opened skeleton>` but then afterwards `skeleton` is removed -- will it still open? Similar for if it's initially opened and disabled.

dlockhart (commented):
This seems much cleaner!

components/inputs/input-date.js
margaree: Yup exactly

components/inputs/input-date.js
margaree: That's a good point - I would presume then in those cases we would want to open since opened would be set and that would be inconsistent if it wasn't actually open.

dlockhart: Yeah that was my thinking as well.

components/inputs/test/input-date.test.js
```diff
@@ -45,35 +45,87 @@ describe('d2l-input-date', () => {
 
 	});
 
-	describe('focus trap', () => {
```
margaree, line 48: I moved these tests to just be combined with the `interacting with dropdown` tests since the setup was the same

components/inputs/test/input-date.visual-diff.js
```diff
@@ -10,9 +10,13 @@ describe('d2l-input-date', () => {
 
 	before(async() => {
 		browser = await puppeteer.launch();
-		page = await visualDiff.createPage(browser, { viewport: { width: 800, height: 900 } });
```
margaree, line 13: Some of these tests could be a bit too reliant on state caused by a previous test rather than working in isolation, so there was work done here to improve that and to pull out some shared code (e.g., opening with key and click functions). This PR can probably be ignored until I deal with how the visual-diff PR turns out since there should be some changes, mostly naming, and some new goldens.

margaree:
This ended up messy due to the multiple visual-diff PRs, so now that things seem stable I'm going to open a new PR, which should make reviewing the visual-diff changes easier.

PR #1619, opened 2021-08-31
US126584 - Remove max count functionality in filter

svanherk:
We're no longer going to say "All" when every filter has been selected, since this implies having everything selected is the same as having nothing selected (which might not be true).

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -41,8 +41,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			disabled: { type: Boolean, reflect: true },
 			_activeDimensionKey: { type: String, attribute: false },
 			_dimensions: { type: Array, attribute: false },
-			_totalAppliedCount: { type: Number, attribute: false },
-			_totalMaxCount: { type: Number, attribute: false }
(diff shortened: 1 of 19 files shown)


components/filter/filter.js
```diff
@@ -356,16 +354,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		}
 	}
 
-	_formatFilterCount(count, max) {
```
svanherk, line 359: This function will go away once we are using the count-badge, which will handle the formatting and max value truncating for us

lang/en.js
```diff
@@ -9,7 +9,6 @@ export default {
 	"components.dialog.close": "Close this dialog",
 	"components.dropdown.close": "Close",
 	"components.filter.loading": "Loading filters",
-	"components.filter.filterCount": "{number, select, max {99+} all {All} other {{number}}}",
```
dlockhart, line 12: Haha too bad this didn't remove `filterCountDescription` that we had all the translation issues with.

dlockhart (approved):
LGTM

lang/en.js
svanherk: Yeah unfortunately we're still on the hook for those lol

PR #1622, opened 2021-09-01
US131350 - Setup Basic Count Badge with some styling

annabelsegalld2l:
This PR sets up a basic count badge with a small and large size, and "notification" and "count" types. The notification type truncates numbers larger than 99. The count type will be used in the filter component.


components/count-badge/README.md
```diff
@@ -0,0 +1,28 @@
...
+<d2l-count-badge badge-size="small" badge-type="notification" number="100"></d2l-count-badge>
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `number` | Number, required | The number to display on the badge. |
+| `badge-size`, default: `small` | String | The size of the badge. Valid options are `"small"` and `"large"`. |
+| `badge-type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and truncate numbers above 99 to `99+`, while count badges are grey and do not truncate the number. |
```
dlockhart, line 24: Nit: could probably go with just `size` and `type` as attribute names here, since we know it's for the badge.

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,110 @@
...
+		:host([badge-size="small"]) .d2l-count-badge-background {
+			border-radius: 0.45rem;
+		}
+
+		:host([badge-size="large"]) .d2l-count-badge-background {
+			border-radius: 0.6rem;
+		}
+
+		:host([badge-type="notification"]) .d2l-count-badge-background {
+			background: var(--d2l-color-carnelian-minus-1);
```
dlockhart, line 79: Nit: I like being really specific here since `background` is one of those complex attributes that can take a lot of stuff.
```suggestion
			background-color: var(--d2l-color-carnelian-minus-1);
```

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,110 @@
...
+
+		:host([badge-type="count"]) .d2l-count-badge-background {
+			background: var(--d2l-color-gypsum);
+		}
+		`];
+	}
+
+	constructor() {
+		super();
+		this.number = null;
```
dlockhart, line 90: Any particular reason to use `null` vs. leaving it `undefined`?

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		:host([badge-size="large"]) .d2l-count-badge-number {
+			font-size: 0.8rem;
+			line-height: 1.2rem;
+			padding-left: 0.4rem;
+			padding-right: 0.4rem;
+		}
+
+		.d2l-count-badge-background {
```
dlockhart, line 64: Could the background just be applied to the `:host` element and remove the need for this extra `<div>`?

components/filter/filter.js
```diff
@@ -356,16 +357,8 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
-		if (count === 0) return;
-
-		let number = count;
-		if (count === max) {
-			number = 'all';
-		} else if (count >= 100) {
-			number = 'max';
-		}
-		return this.localize('components.filter.filterCount', { number: number });
+	_formatFilterCount(count) {
```
dlockhart, line 360: This method seems kind of pointless now -- could it just be removed?

components/count-badge/demo/count-badge.html
```diff
@@ -0,0 +1,37 @@
...
+	</head>
+	<body unresolved>
+		<d2l-demo-page page-title="d2l-count-badge">
+
+			<h2>Small Notification Badge</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-count-badge badge-size="small" badge-type="notification" number="1"></d2l-count-badge>
+					<d2l-count-badge badge-size="small" badge-type="notification" number="10"></d2l-count-badge>
+					<d2l-count-badge badge-size="small" badge-type="notification" number="100"></d2l-count-badge>
```
dlockhart, line 22: This may be coming in one of the later stories, but what's the accessibility plan in terms of providing both a better label for the count (e.g. "100 ungraded activities") and also optionally linking the count with other elements (maybe via `aria-describedby` or something)?

dlockhart (commented):
Looking good!

components/filter/filter.js
svanherk: I think you need to get the latest version of this file, this function changed recently.  And then for now can we leave this function as is and just use this above (no need for the `span`):
````
<d2l-count-badge number="${dimension.appliedCount}"></d2l-count-badge>
```
This will leave the filter button working for now, and once I switch that out we can delete this function

components/count-badge/demo/count-badge.html
annabelsegalld2l: I haven't put much thought into the accessibility plan, it will be in a later story for sure 😄 Stacey mentioned that the filter does have accessible text within the menu-item, so it's not needed urgently for filter.

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,128 @@
+import '../colors/colors.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+class CountBadge extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			/**
+			 * The number to be displayed on the badge. If badge-type is "notification", the number will be truncated to three digits ("99+").
```
annabelsegalld2l, line 10: This should be two digits. I've gone by the maximum number of digits in the number portion, and not including the + sign.

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,128 @@
...
+		super();
+		this.type = 'count';
+		this.size = 'small';
+		this.noRenderZero = false;
+	}
+
+	render() {
+		let numberString = `${this.number}`;
+		if (this.noRenderZero && this.number === 0) {
+			numberString = '';
```
annabelsegalld2l, line 114: In the noRenderZero case, the count badge will be rendered on the page, just without any content within the div, so it will be effectively invisible.

components/count-badge/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `number` | Number, required | The number to display on the badge. |
+| `size`, default: `small` | String | The size of the badge. Valid options are `"small"` and `"large"`. |
+| `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and truncate numbers above 99 to `99+`, while count badges are grey and do not truncate the number. |
+| `digit-limit`, default: `no limit` | Number | For `"count"` type badge, optionally specify a digit limit, after which numbers are truncated. |
+| `no-render-zero`, default: `false` | Boolean | Optionally choose to not render the count badge when the number is zero. |
```
svanherk, line 26: I wonder if this should be `max-digits` to somewhat match the naming scheme we use in https://github.com/BrightspaceUI/core/blob/main/components/inputs/docs/input-numeric.md.  And maybe `hide-zero`?

components/count-badge/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+![Notification Badge](./screenshots/count-badge-notification-small.png?raw=true)
+![Count Badge](./screenshots/count-badge-count-large.png?raw=true)
+
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/count-badge/count-badge.js';
+</script>
+<d2l-count-badge badge-size="small" badge-type="notification" number="100"></d2l-count-badge>
```
dlockhart, line 15:
```suggestion
<d2l-count-badge size="small" type="notification" number="100"></d2l-count-badge>
```

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,128 @@
...
+			line-height: 1.2rem;
+			padding-left: 0.4rem;
+			padding-right: 0.4rem;
+		}
+
+		:host {
+			border: 2px white;
+			display: inline-block;
+			min-width: 0.9rem;
+		}
```
dlockhart, line 84: Once you set the `display` of the host, you also need to include this:

```css
:host([hidden]) {
	display: none;
}
```

components/count-badge/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+**Properties:**
+
+| Property | Type | Description |
+|--|--|--|
+| `number` | Number, required | The number to display on the badge. |
+| `size`, default: `small` | String | The size of the badge. Valid options are `"small"` and `"large"`. |
+| `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and truncate numbers above 99 to `99+`, while count badges are grey and do not truncate the number. |
+| `max-digits`, default: `no limit` | Number | For `"count"` type badge, optionally specify a maximum number of digits, after which numbers are truncated. |
```
dlockhart, line 25: I don't love having attributes that only do something if another attribute is set to a certain value... usually it's an indicator that the component could maybe be broken up into two.

For `max-digits` though, the `notification` type also kind-of supports it in that it truncates 999 to "99+"... so I'm wondering if maybe we could support it and it just defaults to `2` when the type is `notification` and nothing otherwise. That's also a little weird though.

annabelsegalld2l: That sounds like a good idea to me, we can default the notification type to 2 and make it overridable. I can add a descriptive comment in the readme so hopefully its not too confusing.

svanherk: Jeff had mentioned "...I think Notifications should force the 2 digit limit if that’s possible without hackery/awkwardness in the code"

svanherk: Oh, but then also said "If it’s annoying to implement that way, then yeah, it can just be the default for notifications and consumers can override it"

components/count-badge/count-badge.js
```diff
@@ -0,0 +1,136 @@
...
+		this.type = 'count';
+		this.size = 'small';
+		this.hideZero = false;
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		if (!this.maxDigits && this.type === 'notification') {
+			// default to two digits for notification type
+			this.maxDigits = 2;
```
svanherk, line 119: I think this is fine to set here, because they shouldn't be changing `type` or `maxDigits` after the fact - and if they do, I don't think we want to try to adjust things on them.  Just calling this out in case someone has a use case I'm not thinking about.

components/count-badge/test/count-badge.visual-diff.html
```diff
@@ -0,0 +1,29 @@
...
+	<script type="module">
+		import '../count-badge.js';
+		import '../../typography/typography.js';
+	</script>
+	<title>d2l-count-badge</title>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
+	<meta charset="UTF-8">
+</head>
+<body class="d2l-typography">
```
svanherk, line 15: Might seem silly, but would probably be good to have a test or two for the `hide-zero` property.  Maybe one where the property is set but not the number isn't zero so it shows, and one where it is 0 so it doesn't show

components/filter/filter.js
```diff
@@ -216,7 +217,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 			return html`<d2l-menu-item text="${dimension.text}" description="${dimensionDescription}">
 				${builtDimension}
 				<div slot="supporting">
-					<span>${this._formatFilterCount(dimension.appliedCount)}</span>
+					<d2l-count-badge number="${dimension.appliedCount}" max-digits=2 hide-zero></d2l-count-badge>
```
svanherk, line 220:
```suggestion
					<d2l-count-badge number="${dimension.appliedCount}" max-digits="2" hide-zero></d2l-count-badge>
```
Pretty sure it works fine without the quotes in this case, but for consistency we use them everywhere

svanherk (approved):
Nice, looks good to me!

components/count-badge/count-badge.js
dlockhart: Late to the party, but another option would be to do this in the getter for the property. Se default it to `undefined` but then the getter returns `2` if the type is notification no matter what they set it to.

PR #1631, opened 2021-09-02
feat: input-date: add opened property 

margaree:
Adding `opened` in order to more easily be able to change the opened state of the dropdown

This would technically be slightly breaking if anyone was relying on `_dropdownOpened` but that does not seem to be the case.

Previous PR https://github.com/BrightspaceUI/core/pull/1618

### DIFF
--- a/components/inputs/test/input-date.visual-diff.js
+++ b/components/inputs/test/input-date.visual-diff.js
@@ -2,7 +2,7 @@ import { getRect, getRectTooltip, open, reset } from './input-helper.js';
 import puppeteer from 'puppeteer';
 import VisualDiff from '@brightspace-ui/visual-diff';
 
-describe.only('d2l-input-date', () => {
+describe('d2l-input-date', () => {
 
 	const visualDiff = new VisualDiff('input-date', __dirname);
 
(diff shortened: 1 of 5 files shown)


components/inputs/test/input-date.visual-diff.js
```diff
@@ -118,193 +180,113 @@ describe('d2l-input-date', () => {
...
 				const rect = await getRect(page, '#min-max');
 				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 			});
 
 			describe('out of range date typed', () => {
 				// min-value="2018-02-13" max-value="2018-02-27"
-				before(async() => {
-					await page.$eval('#min-max', (elem) => {
+
+				async function setValueBlur(page, selector, value) {
```
margaree, line 192: This is used here because the `blur` is what triggers the validation styling

components/inputs/test/input-date.visual-diff.js
```diff
@@ -315,55 +297,74 @@ describe('d2l-input-date', () => {
 			});
 		});
 
-		describe('with placeholder', () => {
-			afterEach(async() => {
-				await reset(page, '#placeholder');
+		describe('with empty-text', () => {
```
margaree, line 300: These were added because the state is different between open with enter vs with click and I wanted to capture that

margaree:
The changes to the goldens in the corresponding PR are expected and look correct to me.

dlockhart (approved):

PR #1634, opened 2021-09-02
US125557 - Add clear button to filter

svanherk:
This PR adds a clear button to every dimension.  Because the "manual" search-type exists where values are physically removed from the filter on search, we also need to be able to handle sending an event to tell the consumer to clear all their values for that dimension.  This is also needed for those using the HM filter framework, which has a "clear" action for dimensions.  Without this, it's fairly slow (each filter is removed one api call at a time).

Opening this as a draft to get feedback, and opening the clear all one as well to add context to some of the decisions below.

To Do:
- [ ] Tests


components/filter/filter.js
```diff
@@ -228,6 +231,15 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 
 		const dimension = singleDimension ? this._dimensions[0] : this._dimensions.find(dimension => dimension.key === this._activeDimensionKey);
 
+		const clear = html`
+			<d2l-button-subtle
```
svanherk, line 235: I felt it was easiest to just use a `d2l-button-subtle` here rather than a `d2l-selection-action` if it's a list dimension and a `d2l-button-subtle` if it's not.  I have the `dimension.appliedCount` readily available to do the same `require-selection` check, and I also want to disable the button if values are loading.

components/filter/filter.js
```diff
@@ -276,6 +289,40 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+	_clearDimension(dimension) {
+		switch (dimension.type) {
+			case 'd2l-filter-dimension-set': {
+				this.shadowRoot.querySelector(`#${SET_DIMENSION_ID_PREFIX}${dimension.key}`).setSelectionForAll(false);
+				break;
+			}
+		}
+	}
+
+	_clearDimensionData(dimension, sendClearEvent, sendIndividualEvents) {
```
svanherk, line 301: This function does things that aren't utilized in this PR, but are in the Clear All PR: https://github.com/BrightspaceUI/core/pull/1635/files#diff-371eee884788627e4694af0044b256de9257505f6ee6430c4302574842f8b441R474-R478

components/filter/filter.js
```diff
@@ -489,6 +550,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 
 			switch (type) {
 				case 'd2l-filter-dimension-set': {
+					info.clearType = dimension.searchType === 'manual' ? 'manual' : dimension.clearType;
```
svanherk, line 553: If they've chosen to handle search manually, they also have to handle clear manually, since we aren't able to clear values that have been searched away.

components/filter/filter.js
```diff
@@ -276,6 +289,40 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+		this._totalAppliedCount = this._totalAppliedCount - dimension.appliedCount;
+		dimension.appliedCount = 0;
+
+		switch (dimension.type) {
+			case 'd2l-filter-dimension-set': {
+				dimension.values.forEach(value => {
+					if (value.selected && sendIndividualEvents) {
+						this._dispatchChangeEvent(`${dimension.key}-${value.key}`, { dimension: dimension.key, value: { key: value.key, selected: false } });
+					}
+					value.selected = false;
```
svanherk, line 311: For the "manual" clear type, we can't just send the event and clear our counts because when the consumer clears _their_ values, we'll decrease our counts again.  If instead we don't decrease our counts, we'll still be off for all hidden values until the search is cleared.  So we need to decrease the counts and clear all the values we have, and then when the consumer clears them they won't be counted again (because they're trying to set the value to false, which it already is).

components/filter/filter.js
```diff
@@ -306,8 +353,6 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					${this.localize('components.filter.searchResults', { number: count })}
 				</p>
 			`;
-
-			if (count === 0) return searchResults;
```
svanherk, line 310: This was a perf enhancement to not render a list where everything is hidden.  But we want to do that now so that we can clear all values.  This happens when a user has selected values in a list and then searches away all values.  When they click the clear button, we still want to clear all those hidden items.

components/filter/filter-dimension-set.js
```diff
@@ -10,6 +10,13 @@ class FilterDimensionSet extends LitElement {
 
 	static get properties() {
 		return {
+			/**
+			 * Whether to automatically handle clearing or send an event for this dimension
+			 * This should only be set to "manual" for use cases with specific perf enhacements in place for this
+			 * Note: If you choose "manual" for your `search-type`, your `clear-type` will also be "manual" because we cannot update the values removed by searching
+			 * @type {'automatic'|'manual'}
+			 */
+			clearType: { type: String, attribute: 'clear-type' },
```
dlockhart, line 19: Just to confirm: there are known valid cases where they'd want a manual search-type but an automatic clear-type or vice versa? Just want to make sure we can't merge these into one.

svanherk: Yeah, the one I know of is Quick Eval - they are fine with the "automatic" searching, but would want the clear events to apply their HM "clear" action instead of individually unapplying each filter.  They could do that but then they're not taking advantage of that whole functionality of the api, and I worry about perf (which was a problem back in the day at least).

One thing I hadn't really considered, but is technically an option - we could have them do a check for whether they can use the clear action for that dimension.  They'd look at all the events they got back, and see if any would fully clear a dimension.  That still doesn't solve things for those using `manual` search though, and if we have to handle sending clear events anyways...

svanherk: So _another_ thing I was considering was adjusting the structure of the `changes` array in the change event to be grouped by their dimension, and I _could_ add something like `cleared: true|false`.  That way you don't sign up for clear events - if you use manual search you'll _have_ to look for them, and if you want to optimize you can check for that value.

I think in that case I'd need to make certain to send off any applied filter events first in their own event.  Or maybe something like this?
```
detail: {
  changes: [
    { dimension: role, allCleared: true, cleared: [ 'admin', 'instructor'] },
    { dimension: course, allCleared: false, cleared: [ 'biology'], applied: ['art] }
  ]
}
```

dlockhart: That definitely sounds interesting! Mostly I was just poking at it since having the two types _seems_ complicated to wire up, but if they both end up being necessary that's cool.

svanherk: It's definitely more complicated than I wanted, I went back and forth a lot while developing trying to make it less so.  Figured I'd spend the afternoon poking at the new idea of just including this info in the events.  This PR includes both clear and clear all: https://github.com/BrightspaceUI/core/pull/1640

It's definitely less code, and there's no new attributes introduced.  The bulk of the "logic" shifts from being about the different possible combinations with search + clear + clear-all, and moves to making sure the change event is accurate and sent immediately after processing.

svanherk:
Replaced by https://github.com/BrightspaceUI/core/pull/1724

PR #1636, opened 2021-09-03
Daylight: Floating buttons

GeorgePlukov:


components/button/floating-buttons.md
```diff
@@ -1,24 +1,65 @@
...
+  <d2l-button primary>Primary</d2l-button>
+  <d2l-button>Secondary</d2l-button>
+</d2l-floating-buttons>
+```
+
+<!-- docs: start hidden content -->
 ![Floating Buttons](./screenshots/floating-buttons.png?raw=true)
+<!-- docs: end hidden content -->
 
+## Floating button [d2l-floatin-button]
```
margaree, line 32:
```suggestion
## Floating Button [d2l-floating-buttons]
```

margaree: This one's a bit weird with our pattern of non-plural components since the component is actually pluralized🤔

GeorgePlukov: Very good point, it doesn't really make sense to have it as `Floating Button` but otherwise we would have a duplicate level 1/2 headers. I think I'll change it to the plural for now and see what Jeff thinks during the review

margaree (approved):

PR #1637, opened 2021-09-03
fix: avoid layout thrashing inside a loop

dlockhart:
This is pretty bad! We're doing the one thing you're never supposed to do -- read something that would force layout to be calculated and then immediately write inside a loop. It's [the main example](https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing#avoid_layout_thrashing) of layout thrashing. Oops!

### DIFF
--- a/components/table/table-wrapper.js
+++ b/components/table/table-wrapper.js
@@ -1,6 +1,5 @@
 import '../colors/colors.js';
 import '../scroll-wrapper/scroll-wrapper.js';
-import '../../helpers/requestIdleCallback.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
@@ -269,6 +268,11 @@ export class TableWrapper extends RtlMixin(LitElement) {
 
 	async _applyClassNames(table) {
 
+		// wait until start of next frame to read
+		await new Promise(resolve => {
+			requestAnimationFrame(resolve);
+		});
+
 		// offsetParent causes reflow/paint so do them all at once
 		const rows = Array.from(table.rows);
 		let firstRow = null;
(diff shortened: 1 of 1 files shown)


components/table/table-wrapper.js
```diff
@@ -317,7 +321,7 @@ export class TableWrapper extends RtlMixin(LitElement) {
 			const offset = this.type === 'default' ? -3 : 1; // default: -5px top + 2px border, light: 0 top + 1px border
 			const ths = Array.from(table.querySelectorAll('tr.d2l-table-header:not(:first-child) th, tr[header]:not(:first-child) th, thead tr:not(:first-child) th'));
 			ths.forEach((th) => {
-				th.style.top = `${topHeader.clientHeight + offset}px`;
+				th.style.top = `${topHeaderHeight + offset}px`;
```
dlockhart, line 324: This is the main fix here -- awesomely I already stored this in a variable... just needed to actually use it.

components/table/table-wrapper.js
```diff
@@ -362,7 +366,8 @@ export class TableWrapper extends RtlMixin(LitElement) {
 			this._tableIntersectionObserver.observe(table);
 		}
 
-		requestAnimationFrame(() => this._applyClassNames(table));
+		this._applyClassNames(table);
```
dlockhart, line 369: I moved the `requestAnimationFrame` inside `_applyClassNames` so this isn't needed.

components/table/table-wrapper.js
```diff
@@ -352,7 +356,7 @@ export class TableWrapper extends RtlMixin(LitElement) {
 				this._tableIntersectionObserver = new IntersectionObserver((entries) => {
 					entries.forEach((entry) => {
 						if (entry.isIntersecting) {
-							requestIdleCallback(() => this._applyClassNames(table));
+							this._applyClassNames(table);
```
dlockhart, line 359: Decided to remove the `requestIdleCallback` here since it's already waiting for an animation frame. The intersection observer already runs really late, so this isn't causing any issues.

dbatiste (approved):

PR #1638, opened 2021-09-03
fix: input-date: fix incorrect focus behaviour when max-value month same as focus date month

margaree:
This was a bit of a lucky find. The input-date demo page has a demo where the max value is September and I noticed when I used set to today (which is out of range) then clicked the left arrow to go back to range, focus wasn't on the correct max date, it stayed on the number that was where today was on the other month. This is because date focus only updates when month changes and not when year changes.


components/calendar/calendar.js
```diff
@@ -764,20 +765,18 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		const oldFocusDate = new Date(this._focusDate);
 		if (!isDateInRange(possibleFocusDate, getDateFromISODate(this.minValue), getDateFromISODate(this.maxValue))) {
 			// if date is not in range but we are in a dialog, _focusDate should become min or max date if possible
-			if (this._dialog) {
```
margaree, line 767: Nothing changed here - I just was initially in here looking for the problem and cleaned it up a bit while here.

dlockhart (approved):

PR #1635, opened 2021-09-02
US122201 - Add Clear All button to filter

svanherk:
Putting this up early to help give some context for https://github.com/BrightspaceUI/core/pull/1634
"Clear all" calls each dimension and mimics their regular clear behavior (either individual change events or the `d2l-filter-dimension-clear` event).  Or, the consumer can tell the filter to instead send them a "clear all" event - again, I only see this being necessary for the HM filter framework.

Still to do:
- [ ] Clear searches when "Clear All" is pressed
- [ ] Tests

PR #1640, opened 2021-09-03
Alternate Approach - Clear data available in all filter change events

svanherk:
Alternative approach to https://github.com/BrightspaceUI/core/pull/1634 and https://github.com/BrightspaceUI/core/pull/1635.  Instead of consumers turning on and subscribing to clear events, we add that data in to the regular change events always.


svanherk:
Replaced by https://github.com/BrightspaceUI/core/pull/1724 and https://github.com/BrightspaceUI/core/pull/1764

PR #1641, opened 2021-09-03
fix: avoid and defer measuring slot sizes to improve performance

dlockhart:
Our text inputs have some code that measures the widths of its slots in order to apply that same amount of padding, essentially reserving room for the absolutely positioned slots.

This measuring via `getBoundingClientRect()` forces the browser to paint the screen -- possibly before it's ready to do so. In performance speak, this is called a [forced synchronous layout](https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing#avoid_forced_synchronous_layouts).

When the page just has one or two of these components on a page, it's not super noticeable... but the grades "spreadsheet view" could potentially have _thousands_ of them, and the more elements on the screen the longer it takes to paint, which creates an exponentially slower page.

The solution is multi-pronged:

**Avoid measuring completely if we can**

We can actually look inside the slots and if they have no nodes inside them, we can safely assume that their width is zero and then do absolutely no work. This is best-case scenario and actually applies to the grades page which isn't using slots. Win!

**If the slots are in use, move the measuring inside a `requestAnimationFrame`**

This will defer the callback until the next time things have painted, so chances are the measure operation will be free. A bunch of sequential measures can also be batched this way. This solution isn't perfect, as rAF tries to run 60 times per second so with thousands of inputs on a page they're not all going to render quick enough to batch these calls.

**Wait until it's visible on the screen to measure**

The code was already using a `MutationObserver` to re-run the calculation for the case where the input was hidden initially and then shown later. We can turn that into an `IntersectionObserver` instead which will accomplish the same thing but also won't run for visible inputs until they appear in the viewport.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -8,7 +8,6 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
 import { inputStyles } from './input-styles.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
-import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 import { styleMap } from 'lit-html/directives/style-map.js';
@@ -256,6 +255,8 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		this._hasAfterContent = false;
 		this._hovered = false;
 		this._inputId = getUniqueId();
+		this._intersectionObserver = null;
+		this._isIntersecting = false;
 		this._lastSlotWidth = 0;
 		this._prevValue = '';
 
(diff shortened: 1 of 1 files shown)


components/inputs/input-text.js
```diff
@@ -533,14 +543,16 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
...
-
-		// requestUpdate needed for legacy Edge
-		this.requestUpdate();
-		this.updateComplete.then(() => {
-			this._updateInputLayout(container);
-		});
+	_handleSlotChange() {
+		// requestUpdate needed for legacy-Edge
+		if (navigator.userAgent.indexOf('Edge/') > -1) {
+			this.requestUpdate();
```
dlockhart, line 549: This was causing a double-render in all browsers, so putting in a temporary user agent check here avoids it. We can delete this soon.

dlockhart:
And of course all the deferring is causing the visual diff tests to run too soon -- will investigate this next week.

components/inputs/input-text.js
```diff
@@ -331,14 +333,21 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		container.addEventListener('mouseover', this._handleMouseEnter);
 		container.addEventListener('mouseout', this._handleMouseLeave);
 
-		if (!this.offsetParent) {
```
dlockhart, line 334: Checking `offsetParent` was forcing a paint.

dbatiste (approved):

PR #1644, opened 2021-09-07
docs: prefix visual diff commands with "npx"

dlockhart:
Since `mocha` isn't installed globally by the command above, you need `npx` in front for it to work.


annabelsegalld2l (approved):
Thanks for adding this! I was having the same issue and wondering if it was just me.

PR #1645, opened 2021-09-07
testing performance of text input

dlockhart:
This adds some experimental performance testing of our components to core, starting with `<d2l-input-text>`.

How this works:

- Components add a "performance monitor" controller. Once we go to Lit 2, this can be a proper Reactive Controller and all the hooks will work automatically.
- The performance monitor calculates the amount of time that passes between when the component is connected and when it renders. It then fires an event that we can listen for elsewhere.
- I've created a special Chai plugin so you can create a fixture and then assert that `.it.is.performant(val)`. The value you pass in there is the expected average render time of the component. Coming up with this value is going to take a bit of trial & error. I'm hoping the variance in CI isn't too high.

For now I'll make this GitHub Action not be merge-blocking, since I'm worried there will be variance and reliability issues.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -8,6 +8,7 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { inputLabelStyles } from './input-label-styles.js';
 import { inputStyles } from './input-styles.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
+import { PerfMonitor } from '../../helpers/perfMonitor.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 import { styleMap } from 'lit-html/directives/style-map.js';
@@ -264,6 +265,7 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		this._handleFocus = this._handleFocus.bind(this);
 		this._handleMouseEnter = this._handleMouseEnter.bind(this);
 		this._handleMouseLeave = this._handleMouseLeave.bind(this);
+		this._perfMonitor = new PerfMonitor(this);
 	}
 
 	get value() { return this._value; }
@@ -310,6 +312,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		return super.validity;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this._perfMonitor.hostConnected();
+	}
+
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		if (this._intersectionObserver) this._intersectionObserver.disconnect();
@@ -351,6 +358,9 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 	}
 
 	render() {
+
+		this._perfMonitor.hostUpdated();
+
 		const isFocusedOrHovered = !this.disabled && (this._focused || this._hovered);
 		const inputClasses = {
 			'd2l-input': true,
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -312,11 +312,6 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		return super.validity;
 	}
 
-	connectedCallback() {
-		super.connectedCallback();
-		this._perfMonitor.hostConnected();
-	}
-
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		if (this._intersectionObserver) this._intersectionObserver.disconnect();
--- a/components/inputs/test/input-text.perf.js
+++ b/components/inputs/test/input-text.perf.js
@@ -1,0 +1,24 @@
+import '../input-text.js';
+import { expect, html } from '@open-wc/testing';
+
+describe('d2l-input-text', () => {
+
+	it('normal', async() => {
+		await expect(html`<d2l-input-text label="label"></d2l-input-text>`)
+			.to.be.performant(1500);
+	});
+
+	it('slots', async() => {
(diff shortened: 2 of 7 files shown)


components/inputs/input-text.js
```diff
@@ -310,6 +312,11 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		return super.validity;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this._perfMonitor.hostConnected();
```
dlockhart, line 317: This won't be necessary when we move to Lit 2.

components/inputs/input-text.js
```diff
@@ -351,6 +358,9 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 	}
 
 	render() {
+
+		this._perfMonitor.hostUpdated();
```
dlockhart, line 362: Neither will this in Lit 2.

helpers/perfMonitor.js
```diff
@@ -0,0 +1,25 @@
...
+		this.constructedTime = window.performance.now();
+		this.host = host;
+	}
+
+	hostConnected() {
+		this.connectedTime = window.performance.now();
+	}
+
+	hostUpdated() {
+		if (!window.d2lPerfTestInProgress || this.renderedTime !== undefined) return;
```
dlockhart, line 13: Checking for `d2lPerfTestInProgress` means that in production and in normal use the component isn't firing events all over the place.

tools/perf-test-helper.js
```diff
@@ -0,0 +1,45 @@
...
+			total += e.detail.value;
+			if (count === numComponents) {
+				const avg = Math.round(total / count);
+				delete window.d2lPerfTestInProgress;
+				resolve(avg);
+			}
+		});
+
+		for (let i = 0; i < numComponents; i++) {
+			fixture(element);
```
dlockhart, line 23: We render the component 1000 times on the page, then wait for all the events to fire and calculate the average render time.

tools/perf-test-helper.js
```diff
@@ -0,0 +1,45 @@
...
+		}
+
+	});
+}
+
+export const chaiPerf = (_chai, utils) => {
+
+	utils.addMethod(_chai.Assertion.prototype, 'performant', function perfTest(baseline) {
```
GeorgePlukov, line 31: This is awesome!

components/inputs/test/input-text.perf.js
```diff
@@ -0,0 +1,24 @@
...
+		await expect(html`<d2l-input-text label="label">
+			<span slot="left">left</span>
+			<span slot="right">right</span>
+			<span slot="after">after</span>
+		</d2l-input-text>`).to.be.performant(1500);
+	});
+
+	it('unit', async() => {
+		await expect(html`<d2l-input-text label="label" unit="%"></d2l-input-text>`)
+			.to.be.performant(1500);
```
dbatiste, line 21: If I follow correctly, this `1500` is milliseconds, and that would be the max average time to render?

helpers/perfMonitor.js
```diff
@@ -0,0 +1,25 @@
...
+		setTimeout(async() => {
+			await this.host.updateComplete;
+			this.renderedTime = window.performance.now();
```
dbatiste, line 16: Should this be `_renderedTime`?

helpers/perfMonitor.js
```diff
@@ -0,0 +1,25 @@
+export class PerfMonitor {
+
+	constructor(host) {
+		this.constructedTime = window.performance.now();
+		this.host = host;
+	}
+
+	hostConnected() {
```
dbatiste, line 8: Is this being used?  The PR description mentions "connected" but it looks like the "constructed" time is being used in the calc below... which one did you want to use?

components/inputs/test/input-text.perf.js
dlockhart: Yeah exactly. It renders it 1000 times and calculates the average render time and then compares it to this value.

helpers/perfMonitor.js
dlockhart: Ah you're right. Technically we could use this instead of constructed time... which... yeah I'm not sure which one would be more accurate. I'll remove this for now.

dbatiste:
I am liking this.

What about a case were we want to test the perf of a state change after the initial render?  For instance, opening a dialog, or a date-time picker?  If we wanted to test that, would we just do that in a special test that uses the native performance API directly to capture timestamps before/after the property change?

helpers/perfMonitor.js
dlockhart: I just used non "private" variables in here since this isn't a mixin -- it's a controller so we don't need to worry about variable name collisions.

dlockhart:
> What about a case were we want to test the perf of a state change after the initial render?

Yeah this doesn't handle anything like that. But as you said, you likely wouldn't need any special instrumentation in the component to test that -- you could just measure, then perform an action and wait for an event and then measure again.

But yeah... doing that over and over again and averaging could be tricky. So maybe in the future we could extend the Chai plugin somehow to handle that.

I'm far less worried about the performance of things that happen when the user interacts with them vs. the initial page load though.

tools/perf-test-helper.js
```diff
@@ -0,0 +1,45 @@
...
+	return new Promise((resolve) => {
+
+		let total = 0;
+		let count = 0;
+		window.d2lPerfTestInProgress = true;
+		document.body.addEventListener('d2l-component-perf', (e) => {
+			count++;
+			total += e.detail.value;
+			if (count === numComponents) {
+				const avg = Math.round(total / count);
```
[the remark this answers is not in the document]
dlockhart: We talking about milliseconds here, so really the difference between 1500 and 1501 doesn't worry me too much... especially when the variance I was seeing in Sauce Labs was anywhere from 800 -> 1400.

dbatiste (approved):

PR #1646, opened 2021-09-08
feat: US131353 - add a11y features to count-badge

annabelsegalld2l:
Required Accessibility features added: 

1. Text - to be used as an aria-label on the count badge. I've configured it so only the description is read by screen readers, NOT the number text. This allows consumers to specify their own labels and optionally change how numbers are displayed to screen readers, such as saying "100 filters applied" rather than "99+".

Optional:
2. Tab-stop - Optionally allow the badge to be tabable. 
3. Announce-changes - add an aria-live region to the badge. I decided to keep this as default off, as in most cases we don't need to announce new changes to the badge. 
4. Has-tooltip. Optionally adds a tooltip that says the text property.


components/count-badge/README.md
```diff
@@ -25,6 +25,17 @@ The `d2l-count-badge` element is a web component to display a number count, eith
 | `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
 | `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
 
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `description` | Optionally choose to add a description, to be the aria-label on the badge. |
```
dlockhart, line 34: Wondering: without this, a screen reader user is just going to happen upon a random number when combing through the page. Should we rename this to `label` and make it required?

components/count-badge/count-badge.js
```diff
@@ -134,7 +164,11 @@ class CountBadge extends RtlMixin(LitElement) {
 			numberString = `${'9'.repeat(this.maxDigits)}+`;
 		}
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div class="d2l-count-badge-number" 
+				tabindex="${this.tabStop ? '0' : '-1'}" 
+				role="${this.announceChanges ? 'status' : 'presentation' }"
```
dlockhart, line 169: Hmm, `presentation` things don't end up as anything semantic in the accessibility API, which I'm not sure we want. Maybe just either `status` or leave it undefined and use the `ifDefined` directive? That way it just won't set the attribute at all.

components/count-badge/count-badge.js
```diff
@@ -134,7 +164,11 @@ class CountBadge extends RtlMixin(LitElement) {
 			numberString = `${'9'.repeat(this.maxDigits)}+`;
 		}
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div class="d2l-count-badge-number" 
+				tabindex="${this.tabStop ? '0' : '-1'}"
```
dlockhart, line 168: If you don't need to programmatically focus on the element, you could leave it `undefined` + the `ifDefined` directive to just set it to `0` when `tabStop` is `true`.

```html
tabindex="${ifDefined(this.tabStop ? '0' : undefined)}" 
```

components/count-badge/count-badge.js
```diff
@@ -51,6 +51,33 @@ class CountBadge extends RtlMixin(LitElement) {
...
+				reflect: true,
+				attribute: 'tab-stop'
+			},
+			/**
+			 * Optionally choose to announce changes to the badge with an aria-live region. If the number property is changed, the description will be read by screenreaders. Defaults to false.
+			 * @type {boolean}
+			 */
+			announceChanges: {
+				type: Boolean,
+				reflect: true,
```
dlockhart, line 78: I've noticed recently a non-zero performance cost to reflecting properties to attributes. So likely if you don't need a CSS selector, we could leave reflection off.

components/count-badge/README.md
annabelsegalld2l: In its current state, without the tab stop, the label just never gets read. There are some cases, like card-footer-link, where screen readers don't read the count badge at all, so no label is required.

components/count-badge/count-badge.js
annabelsegalld2l: Yep, leaving out the role if we don't need it makes sense. I'm still playing around between using role and the aria-live attribute, both are failing the axe tests currently :(

components/count-badge/README.md
dlockhart: What about just arrowing through? It should get read then. Like most non-interactive things on the page (headings, paragraphs, lists, etc) don't get tab stops either -- the screen reader user navigates to them using other means.

annabelsegalld2l: I see what you mean now. Then it does make sense to make it required, I'll send an update with the *hopefully* fixed tests.

components/count-badge/README.md
```diff
@@ -25,6 +25,17 @@ The `d2l-count-badge` element is a web component to display a number count, eith
...
 
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `description`, required | A description of the badge for screen-readers, to be used as an aria-label. NOTE: Only the description will be read by screen-readers (not the number). |
+| `tab-stop` | Optionally choose to make the badge a tab stop. |
+| `announce-changes` | Optionally choose to announce changes to the badge with an aria-live region. If the number property is changed, the description will be read by screenreaders. |
```
dlockhart, line 36: I think we typically still include accessibility-related properties in the property list above, and then just call them out again in these lists with extra guidance. Although I'm not sure if @margaree has thoughts here or if we've changed our approach with the Daylight site?

components/count-badge/README.md
```diff
@@ -25,6 +25,17 @@ The `d2l-count-badge` element is a web component to display a number count, eith
 | `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
 | `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
 
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `description`, required | A description of the badge for screen-readers, to be used as an aria-label. NOTE: Only the description will be read by screen-readers (not the number). |
```
dlockhart, line 34: To me, `description` still sounds a little... optional. Wondering if `text` might be more appropriate? That would line up nicely with `<d2l-button-icon>`, which is kind of a similar component.

components/count-badge/count-badge.js
```diff
@@ -134,7 +160,13 @@ class CountBadge extends RtlMixin(LitElement) {
...
 		}
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div
+			tabindex="${ifDefined(this.tabStop ? '0' : undefined)}" 
+			role="${ifDefined(this.announceChanges ? 'status' : undefined)}"
+			aria-label="${this.description}">
+				<div class="d2l-count-badge-number" aria-hidden="true">${numberString}</div>
+				<span class="d2l-offscreen">"${this.description}"</span>
+			</div>`;
```
dlockhart, line 169: This might be more of a question for @geurts, but what do you think about adding a `<d2l-tooltip>` with the description/text that they're required to provide? It would have the advantage of showing the full text ("1 unread discussion") to visual users on-hover, and it would automatically hook up as `aria-labelledby`.

If we did that though, I think we'd want there to always be a tab stop so that keyboard users could see it.

components/count-badge/README.md
margaree: Yup that's still correct

components/count-badge/count-badge.js
annabelsegalld2l: I know for filter we did not want to add a tab stop, since it messes with the tab order of the menu. I like the idea of adding an optional tooltip though. There are cases, like the card-footer-link on the course cards that uses a separate tooltip; it has different text than the count badge (related to the link rather than the number).

annabelsegalld2l: I chatted with Jeff, he is on board with an optional tooltip 😄 It will be opt-in since we only have a few use cases where it's needed.

components/count-badge/count-badge.js
```diff
@@ -134,7 +172,15 @@ class CountBadge extends RtlMixin(LitElement) {
...
 		}
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div
+			id="${this._badgeId}"
+			tabindex="${ifDefined(this.tabStop || this.tooltip ? '0' : undefined)}" 
+			role="${ifDefined(this.announceChanges ? 'status' : undefined)}"
+			aria-label="${this.text}">
+				<div class="d2l-count-badge-number" aria-hidden="true">${numberString}</div>
+			${this.tooltip ? html`<d2l-tooltip for="${this._badgeId}">${this.text}</d2l-tooltip>` : html`<span class="d2l-offscreen">"${this.text}"</span>`}
```
annabelsegalld2l, line 181: We only need one of the tooltip or offscreen so I just have a ternary check here to show the tooltip if required, otherwise use d2l-offscreen.

components/count-badge/test/count-badge.visual-diff.js
```diff
@@ -14,6 +15,18 @@ describe('d2l-count-badge', () => {
 		await page.bringToFront();
 	});
 
+	async function getRect(page, selector) {
+		return page.$eval(selector, (elem) => {
```
annabelsegalld2l, line 19: The tooltip wasn't getting into the rect clip for some reason, so I just added my own getRect function for the tooltip test.

components/filter/filter.js
```diff
@@ -213,11 +214,13 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		}
 		return this._dimensions.map((dimension) => {
 			const builtDimension = this._buildDimension(dimension);
-			const dimensionDescription = `${dimension.text}. ${this.localize('components.filter.filterCountDescription', { number: dimension.appliedCount })}`;
-			return html`<d2l-menu-item text="${dimension.text}" description="${dimensionDescription}">
+			const dimensionDescription = `${dimension.text}.`;
+			const countBadgeId = getUniqueId();
```
annabelsegalld2l, line 218: Didn't know about this handy feature. All the Ids here need to be unique so the aria-describedby tag can read from the correct badge.

components/count-badge/count-badge.js
```diff
@@ -48,14 +50,45 @@ class CountBadge extends RtlMixin(LitElement) {
...
+				attribute: 'tab-stop'
+			},
+			/**
+			 * Optionally choose to announce changes to the badge with an aria-live region. If the number property is changed, the text will be read by screenreaders. Defaults to false.
+			 * @type {boolean}
+			 */
+			announceChanges: {
+				type: Boolean,
+				attribute: 'announce-changes'
+			},
```
dlockhart, line 78: Nit: typically we alphabetize properties in this list and in the constructor.

components/count-badge/count-badge.js
```diff
@@ -134,7 +172,15 @@ class CountBadge extends RtlMixin(LitElement) {
...
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div
+			id="${this._badgeId}"
+			tabindex="${ifDefined(this.tabStop || this.tooltip ? '0' : undefined)}" 
+			role="${ifDefined(this.announceChanges ? 'status' : undefined)}"
+			aria-label="${this.text}">
+				<div class="d2l-count-badge-number" aria-hidden="true">${numberString}</div>
+			${this.tooltip ? html`<d2l-tooltip for="${this._badgeId}">${this.text}</d2l-tooltip>` : html`<span class="d2l-offscreen">"${this.text}"</span>`}
+			</div>
```
dlockhart, line 182: The tooltip has a mode where it can act as the label for a thing, and when used, it makes using `aria-label` redundant. But of course then when it's not used, you still need the label. One thing's for sure though: you don't need `aria-label` AND the offscreen text.

I'd suggest something like this:

```suggestion
        	<div
			id="${this._badgeId}"
			tabindex="${ifDefined(this.tabStop || this.tooltip ? '0' : undefined)}" 
			role="${ifDefined(this.announceChanges ? 'status' : undefined)}"
			aria-label="${ifDefined(this.tooltip ? undefined : this.text)}">
				<div class="d2l-count-badge-number" aria-hidden="true">${numberString}</div>
			${this.tooltip ? html`<d2l-tooltip for="${this._badgeId}">${this.text}</d2l-tooltip>` : null}
			</div>
```

components/count-badge/count-badge.js
```diff
@@ -48,14 +50,45 @@ class CountBadge extends RtlMixin(LitElement) {
...
+			 */
+			announceChanges: {
+				type: Boolean,
+				attribute: 'announce-changes'
+			},
+			/**
+			 * Optionally add a tooltip on the badge. Defaults to false.
+			 * @type {boolean}
+			 */
+			tooltip: {
```
dlockhart, line 83: Nit: could this be `has-tooltip`? Just want the name to clearly be "boolean" as we use `tooltip` on `<d2l-switch>` as a String.

components/count-badge/count-badge.js
```diff
@@ -48,14 +50,45 @@ class CountBadge extends RtlMixin(LitElement) {
...
 				attribute: 'hide-zero'
 			},
+			/**
+			 * The text that will show as an aria-label on the badge. NOTE: Only the text will be read by screen-readers (not the number).
+			 * @type {string}
+			 */
+			text: {
+				type: String,
+				attribute: 'text'
+			},
```
dlockhart, line 62: You don't need to set `attribute` if the desired name matches the property name.

```suggestion
			text: {
				type: String
			},
```

components/count-badge/count-badge.js
```diff
@@ -48,14 +50,45 @@ class CountBadge extends RtlMixin(LitElement) {
 			 */
 			hideZero: {
 				type: Boolean,
-				reflect: true,
 				attribute: 'hide-zero'
 			},
+			/**
+			 * The text that will show as an aria-label on the badge. NOTE: Only the text will be read by screen-readers (not the number).
```
dlockhart, line 56: Since this is now the tooltip text as well, maybe something like:
```suggestion
			 * Descriptive text for the badge which will act as an accessible label and tooltip text when tooltips are enabled. Only the text will be read by screen-readers (not the number), so include the number in the text.
```

The bit about how screen readers don't read the number could maybe even just move down to the "Accessibility Properties" section in the README.

components/count-badge/count-badge.js
```diff
@@ -134,7 +173,15 @@ class CountBadge extends RtlMixin(LitElement) {
 			numberString = `${'9'.repeat(this.maxDigits)}+`;
 		}
 		return html`
-        	<div class="d2l-count-badge-number">${numberString}</div>`;
+        	<div
+			class="d2l-count-badge-number" 
+			id="${this._badgeId}"
+			tabindex="${ifDefined(this.tabStop || this.hasTooltip ? '0' : undefined)}" 
+			role="${ifDefined(this.announceChanges ? 'status' : undefined)}">
+					<div aria-hidden="true">${numberString}</div>
```
annabelsegalld2l, line 181: I decided to get rid of the aria-label altogether, and choose either a tooltip or d2l-offscreen as the accessible label.

dlockhart: Cool. In that case though you'll need to manually link the `span` and the badge together (tooltip does this for you) by giving the `span` a generated `id` and then putting `aria-labelledby` on the badge.

dlockhart: Hmm, actually since it's a child that may not be necessary actually. But then I'm thinking maybe the tooltip shouldn't be a child since it will create that linkage for you.

annabelsegalld2l: It seems to be working without the aria-labelledby and the screenreader gets the text in either case. I'm also applying an aria-describedby to the filter menu items and they can get the text from the d2l-offscreen as well... so I didn't need the aria-label in either case.

components/count-badge/README.md
```diff
@@ -24,6 +24,21 @@ The `d2l-count-badge` element is a web component to display a number count, eith
 | `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and count badges are grey. |
 | `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
 | `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
+| `text`, required | String | Descriptive text for the badge which will act as an accessible label and tooltip text when tooltips are enabled. |
+| `tab-stop`, default: `false` | Boolean | Optionally choose to make the badge a tab stop. |
```
svanherk, line 28: Just wrapping my head around the need for both `tab-stop` and `has-tooltip`... sometimes we want to be able to tab to a count-badge so the accessibility label can be read, but we _don't_ want a tooltip?

I'm wondering if these two should be combined for a keyboard-only user - basically, if you want to add a tab-stop for screenreader users, then keyboard-only users also benefit from the tooltip.  And the reverse seems to already be true - if you add a tooltip, then you're already adding a tab-stop.

annabelsegalld2l: The spec mentions we can do a tab stop without a tooltip. I think one case with a tab stop and no tooltip could be portfolio where the text is already showing something similar to what the tooltip would show (ie. overdue evidence). A screenreader may want to still read out "X pieces of evidence" but not show a tooltip there.

components/count-badge/README.md
```diff
@@ -24,6 +24,21 @@ The `d2l-count-badge` element is a web component to display a number count, eith
...
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `text`, required | Only the text will be read by screen-readers (not the number), so include the number in the text. |
+| `tab-stop` | A tab stop allows keyboard users to easily tab to the badge. Otherwise, keyboard users will need to arrow through to the badge. |
+| `announce-changes` | Use "announce-changes" if screen-reader users should be notified that the count has been updated, such as a new notification. The "text" property will be read as soon as the screen-reader is idle. |
+| `tooltip` | The tooltip will be visible on hover, and read out by screen-readers. |
```
svanherk, line 41:
```suggestion
| `has-tooltip` | The tooltip will be visible on hover/tab stop, and read out by screen-readers. |

```

components/count-badge/README.md
```diff
@@ -24,6 +24,21 @@ The `d2l-count-badge` element is a web component to display a number count, eith
...
+| `has-tooltip`, default: `false` | Boolean | Optionally choose to have a tooltip below the badge. |
+
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
+
+| Attribute | Description |
+|--|--|
+| `text`, required | Only the text will be read by screen-readers (not the number), so include the number in the text. |
+| `tab-stop` | A tab stop allows keyboard users to easily tab to the badge. Otherwise, keyboard users will need to arrow through to the badge. |
```
svanherk, line 39:
```suggestion
| `tab-stop` | A tab stop allows screen-reader users to easily tab to the badge. Otherwise, screen-reader users will need to arrow through to the badge. |
```

Wonder if this is more accurate?  (Since just the tab-stop doesn't really add anything for keyboard users)

components/count-badge/demo/count-badge.html
```diff
@@ -17,20 +17,27 @@
 			<h2>Small Notification Badge</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-count-badge size="small" type="notification" number="1"></d2l-count-badge>
-					<d2l-count-badge size="small" type="notification" number="10"></d2l-count-badge>
-					<d2l-count-badge size="small" type="notification" number="100"></d2l-count-badge>
+					<d2l-count-badge size="small" announce-changes tab-stop text=" 1 new notification." type="notification" number="1"></d2l-count-badge>
```
svanherk, line 20: Just an idea, but might be nice to add a button to this demo that increments this number and updates the text to test the live-region

components/filter/filter.js
```diff
@@ -213,11 +214,13 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		}
 		return this._dimensions.map((dimension) => {
 			const builtDimension = this._buildDimension(dimension);
-			const dimensionDescription = `${dimension.text}. ${this.localize('components.filter.filterCountDescription', { number: dimension.appliedCount })}`;
-			return html`<d2l-menu-item text="${dimension.text}" description="${dimensionDescription}">
+			const dimensionDescription = `${dimension.text}.`;
```
svanherk, line 217: Minor: Probably don't need this variable anymore, we can just use `dimension.text` directly in the `menu-item`.

components/filter/filter.js
svanherk: I don't know the answer to this, maybe @dlockhart... do you know if there's a perf hit with changing this id every re-render? We could also make an id out of a combo of the dimension kay and the value key, but not sure if it matters.

annabelsegalld2l: I did have the combo of the dimension and value key as my previous iteration. I was just worried it would get really messy turning the text into an id, considering it could contain different punctuation, spaces, that would have to be removed.

dlockhart: That's a good point Stacey... yeah this is going to cause a perf issue in the sense that Lit is going to have to update DOM on every re-render, whereas normally if the output doesn't change they don't do anything.

You could either combine dimension key + value key (which should both be numbers or something similar) or generate a unique ID but store it in the map so that you're not re-generating it each time.

svanherk: I guess value doesn't make sense here. If you do want to use the dimension for the id, I'd suggest something like:
```
id="${MENU_ITEM_ID_PREFIX}${dimension.key}"
```
because IDs can't start with a number, and the consumer key might.
Similar to here: https://github.com/BrightspaceUI/core/blob/main/components/filter/filter.js#L316

components/count-badge/README.md
```diff
@@ -24,6 +24,21 @@ The `d2l-count-badge` element is a web component to display a number count, eith
 | `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and count badges are grey. |
 | `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
 | `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
+| `text`, required | String | Descriptive text for the badge which will act as an accessible label and tooltip text when tooltips are enabled. |
+| `tab-stop`, default: `false` | Boolean | Optionally choose to make the badge a tab stop. |
+| `announce-changes`, default: `false` | Boolean | Optionally choose to announce changes to the badge with an aria-live region. If the number property is changed, the text will be read by screen-readers. |
```
svanherk, line 29:
```suggestion
| `announce-changes`, default: `false` | Boolean | Optionally choose to announce changes to the badge with an aria-live region. If the text property is changed, the text will be read by screen-readers. |
```
I think this is technically more accurate with the latest implementation

svanherk (approved):

PR #1648, opened 2021-09-08
fix: US130195 - Lower z-index for primary-secondary footer

annabelsegalld2l:
This is required to get the dropdown trays (specifically user-profile-card" to render correctly on top of the primary-secondary footer. Without this, the footer goes overtop of the close button of the tray making it not clickable. With this change, the page functions normally, and the dropdown can go on top of the footer when it opens.  The z-index was added here for a previous bug (https://github.com/BrightspaceUI/core/pull/770), which looks like it it is still fixed when the footer has a z-index of 1. I took a look at the LMS and the primary-secondary pages appear to look the same.

Here's a before and after:
profile-z-index
user-profile-after

I was wondering why this issue didn't appear with the date-picker, and it seems to do with the HTML order that determines the stacking context on each page.


dbatiste (approved):
This seems ok.  I don't think tooltips would be affected by this assuming they open upwards, and dialogs would be on top of the footer already.

PR #1650, opened 2021-09-09
feat: input-time: add opened property

margaree:


margaree:
The goldens look as expected. There are some height changes due to moving `input-time`s around in order to reflect the same functionality in the tests (i.e., if the new tests were added below instead of above `#dropdown` then it would open down instead of up)

components/inputs/input-time.js
```diff
@@ -426,5 +438,18 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
...
+		if (!this._dropdownFirstOpened) this._dropdownFirstOpened = true;
+
+		const dropdown = this.shadowRoot.querySelector('d2l-dropdown-menu');
+		if (dropdown && dropdown.opened) return;
+
+		await this.updateComplete;
+
+		dropdown.open(true);
+		this.shadowRoot.querySelector('d2l-menu').focus();
+	}
```
dlockhart, line 453: I'm curious... instead of `_open()` and `_close()` could we bind to `d2l-dropdown-menu`'s `opened` in `render()`? Something like:

```javascript
const opened = this.opened && !this.disabled && !this.skeleton;
...
return html`...<d2l-dropdown-menu ?opened="${opened}">...`;
```

margaree: ooo good idea, I think that works!

components/inputs/input-time.js
```diff
@@ -141,6 +141,10 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
 			 * @type {number}
 			 */
 			maxHeight: { type: Number, attribute: 'max-height' },
+			/**
+			 * Whether or not the dropdown is open
+			 */
+			opened: { type: Boolean, reflect: true },
```
dlockhart, line 147: If this doesn't need to reflect for CSS reasons, probably best to not. I'm seeing non-zero performance implications of reflecting.

components/inputs/input-time.js
```diff
@@ -403,15 +412,10 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
...
 
 	async _handleKeydown(e) {
-		const dropdown = this.shadowRoot.querySelector('d2l-dropdown-menu');
 		// open and focus dropdown on down arrow or enter
 		if (e.keyCode === 40 || e.keyCode === 13) {
-			if (!this._dropdownFirstOpened) {
-				this._dropdownFirstOpened = true;
-				await this.updateComplete;
-			}
-			dropdown.open(true);
```
margaree, line 413: When opening, `applyFocus` is true by default so this behaviour should remain the same. The `this.shadowRoot.querySelector('d2l-menu').focus();` also seems to happen by default on key open

components/inputs/input-time.js
```diff
@@ -403,15 +412,10 @@ class InputTime extends SkeletonMixin(FormElementMixin(LitElement)) {
...
-			this.shadowRoot.querySelector('d2l-menu').focus();
```
dlockhart, line 414: That cleans this code up nicely. 💯

dlockhart (approved):

components/inputs/input-time.js
svanherk: Should this be updated? https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-reflect-properties-back-to-attributes

dlockhart: Yeah that's a good call -- will update.

PR #1654, opened 2021-09-09
Replacing "min-height" attribute on floating buttons with a constant

dlockhart:
In prepping for some performance improvements, I noticed that nobody is setting `min-height` on floating buttons -- so we can simplify the logic by removing the property altogether.


components/button/floating-buttons.js
```diff
@@ -15,12 +15,7 @@ class FloatingButtons extends RtlMixin(LitElement) {
 			/**
 			 * Indicates to display buttons as always floating
 			 */
-			alwaysFloat: { type: Boolean, attribute: 'always-float' },
-
-			/**
-			 * The minimum height of the viewport to display floating buttons at (where applicable)
-			 */
-			minHeight: { type: String, attribute: 'min-height' },
+			alwaysFloat: { type: Boolean, attribute: 'always-float', reflect: true },
```
dlockhart, line 18: Noticed that the CSS has an attribute selector for `always-float` so this property needs to reflect.

components/button/floating-buttons.md
```diff
@@ -19,7 +19,7 @@ Floating workflow buttons behavior can be added by using the `<d2l-floating-butt
 <p>
   Lorem ipsum dolor sit amet, consectetur adipisicing elit. Adipisci repellat cum totam! Enim, sunt. Numquam voluptate, velit quisquam ipsa molestias laudantium odit reiciendis nisi corporis voluptatibus, voluptatum sunt natus, accusantium magnam consequatur fugit officiis minima voluptatem consequuntur nam, earum necessitatibus! Cupiditate ullam repellendus, eius iure voluptas at commodi consectetur, quia, adipisci possimus, ex mollitia. Labore harum error consectetur officiis aut optio, temporibus iste nobis ducimus cumque laudantium rem pariatur. Ut repudiandae id, consequuntur quasi quis pariatur autem corporis perferendis facilis eius similique voluptatibus iusto deleniti odio officia numquam tenetur excepturi, aspernatur sunt minima aut fugiat ipsam.
 </p>
-<d2l-floating-buttons min-height="100px">
+<d2l-floating-buttons>
```
dlockhart, line 22: @GeorgePlukov I think this might break the demo in the Daylight site... do you think adding `always-float` would fix things?

GeorgePlukov: I should be able to get the demos working with the `always-float` property. Since `min-height` has a default value set is it not possible this changes the point where buttons start floating on some pages?

dlockhart: Ok thanks -- I've updated it to set `always-float`.

It shouldn't cause any change in behaviour since nobody was setting the attribute anywhere, so it was always `500px`.

dbatiste (approved):

GeorgePlukov (approved):

PR #1656, opened 2021-09-09
Bounding ancestors support and performance fixes

dlockhart:
The changes here accomplish two things:

1. **Bounding Ancestor Support**: In FACE quiz editor, it was noticed that floating buttons didn't work at all inside our legacy dialogs. This is because they have an ancestor element that's bounding -- it uses `overflow: scroll`. This fixes that.
2. **Performance**: The old `_shouldFloat` method does a bunch of measuring and calculating that forces a reflow. It also does those calculations _a lot_ -- every time the user scrolls, a DOM node it added to the page, the window resizes, and so on.

The solution involves using an `IntersectionObserver` to observe a fake element that we insert into the DOM just below the floating buttons. If that element isn't visible in the viewport, we float the buttons, and when it appears in the viewport we stop floating. Pretty simple!


components/button/floating-buttons.js
```diff
@@ -16,11 +18,11 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
 			alwaysFloat: { type: Boolean, attribute: 'always-float', reflect: true },
-			_containerMarginLeft: { type: String },
-			_containerMarginRight: { type: String },
+			_containerMarginLeft: { attribute: false, type: String },
+			_containerMarginRight: { attribute: false, type: String },
 			_floating: { type: Boolean, reflect: true },
-			_innerContainerLeft: { type: String },
-			_innerContainerRight: { type: String }
+			_innerContainerLeft: { attribute: false, type: String },
+			_innerContainerRight: { attribute: false, type: String }
```
dlockhart, line 25: We don't want people to be able to affect these by setting the attributes.

components/button/floating-buttons.js
```diff
@@ -147,75 +165,98 @@ class FloatingButtons extends RtlMixin(LitElement) {
 		`;
 	}
 
-	_calcContainerPosition() {
-		this._floating = this._shouldFloat();
-		if (!this._floating || !this.offsetParent) {
-			return;
+	updated(changedProperties) {
+		if (changedProperties.has('alwaysFloat')) {
+			this._recalculateFloating();
```
dlockhart, line 170: Adding this attribute after things had rendered wasn't working before.

components/button/floating-buttons.js
```diff
@@ -147,75 +165,98 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
 		}
+	}
 
-		const offsetParentBoundingRect = this.offsetParent.getBoundingClientRect();
-		const boundingRect = this.getBoundingClientRect();
+	async _calcContainerPosition() {
```
dlockhart, line 174: Side-by-side diff might be best for this method. Mostly things just got indented inside the `rAF`.

components/button/floating-buttons.js
```diff
@@ -147,75 +165,98 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
-		const left = boundingRect.left;
-		const containerLeft = left - offsetParentLeft - 1;
-		if (containerLeft !== 0) {
-			// only update this if needed - needed for firefox
-			this._containerMarginLeft = `-${containerLeft}px`;
-		}
+			this._containerMarginLeft = '';
+			this._containerMarginRight = '';
+			this._innerContainerLeft = '';
+			this._innerContainerRight = '';
```
dlockhart, line 180: We weren't resetting these before, which left the margins on the containers even when things weren't floating.

components/button/floating-buttons.js
```diff
@@ -147,75 +165,98 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
+				this._innerContainerRight = `${containerRight}px`;
+			}
+
+		});
 	}
 
-	_shouldFloat() {
-		if (this.alwaysFloat) {
-			return true;
+	_getBoundingAncestor() {
```
dlockhart, line 216: When searching, we pick wither the bounding ancestor (`overflow: scroll`) or offsetParent (`position: relative`) that we see first by traveling up the DOM.

components/demo/demo-snippet.js
```diff
@@ -178,9 +178,6 @@ class DemoSnippet extends LitElement {
 	_handleDirChange(e) {
 		this._dir = e.target.on ? 'rtl' : 'ltr';
 		this._applyAttr('dir', this._dir, true);
-		this.dispatchEvent(new CustomEvent(
-			'd2l-dir-update', { bubbles: true, composed: true, detail: { dir: this._dir } }
-		));
```
dlockhart, line 183: Floating buttons don't need this event anymore and nothing else was using it.

components/button/floating-buttons.js
```diff
@@ -99,32 +101,48 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
+		if (mediaQueryList.removeEventListener) mediaQueryList.removeEventListener('change', this._recalculateFloating);
+		if (this._intersectionObserver) this._intersectionObserver.disconnect();
 	}
 
 	firstUpdated() {
 		super.firstUpdated();
 
-		this._calcContainerPosition();
-		this._startObserver();
+		const test = document.createElement('div');
```
dbatiste, line 128: If I understand this correctly, we need something outside this component to test, that doesn't get floated...?

In the code below, is there only ever one entry?  Just wondering if there is more than one entry, are we calling the recalc function more than once?

dlockhart: Right, I meant to put a note in about this. I was originally hoping this fake element could just be output in `render()` and the styles that are getting applied to the host could move to an inner container.

The problem is with sticky positioning -- the way it works is that the parent element always becomes the sticky container and the siblings are the things it "sticks" against. So the host HAS to be the thing that's sticky, so that its parent is the sticky container. That meant manually inserting the fake element as the host's next sibling.

In terms of the observer, you end up with one entry for each target element that you're observing. Since we're only observing 1 element (the fake test element), we'll only ever get 1 entry.

dbatiste: I see, makes sense.  So in this case, if the floating buttons are moved in the DOM (I think unlikely), this would currently stop working.  Also, if the app inserts anything between the test node and the custom element, we could get some odd behaviour.  Not sure how much we care about these scenarios.

dlockhart: Correct. But the moving thing is a valid concern... I'll see if it's easy to handle on disconnect/connect.

dlockhart: I just pushed some changes that handle removing/re-adding the test element when the buttons are disconnected and re-connected.

I also noticed that on really slow pages, the floating buttons would be hidden (and therefore float), then be visible (and un-float), then get pushed down as content loaded and become floating again... which lead to 3-4 re-calculations. So I've moved the call to `observe` into a `requestIdleCallback` with a 5 second timeout. Having the buttons immediately float just isn't that important so I think deferring a bit is fine.

dbatiste (approved):

PR #1658, opened 2021-09-10
feat: input-date-time: opened

margaree:


components/inputs/input-date.js
```diff
@@ -457,7 +457,7 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 
 	async _open() {
 		if (this.disabled || this.skeleton) return;
-		if (!this._dropdownFirstOpened) await this._handleFirstDropdownOpen();
+		if (!this._dropdownFirstOpened || !this._dropdown) await this._handleFirstDropdownOpen();
```
margaree, line 460: There was a console error that could happen here if `opened` was set initially so this prevents that.

margaree:
The goldens look as expected

dlockhart (approved):

PR #1667, opened 2021-09-10
input-date-range: await updateComplete for opened test

margaree:
I'll rerun several times to make sure this is helping

1st failure was `d2l-dropdown-content mobile-right-tray-no-close` (off by 1 px)
2nd failure was `d2l-dropdown-content bottom-right` (dropdown looked to be slightly off in position)


margaree:
I've run 3 times in a row without the opened test flaking. I'll run again.

dlockhart (approved):
I'm seeing about a 50/50 fail rate in my other PRs so 3 times sounds promising!

PR #1671, opened 2021-09-10
Initial check-in for supporting nested lists.

dbatiste:
This PR adds initial layout and selection support for nested lists.

image


components/list/list-item-generic-layout.js
```diff
@@ -143,8 +154,8 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			<slot name="control-action" class="d2l-cell" data-cell-num="3"></slot>
 			<slot name="control" class="d2l-cell" data-cell-num="4"></slot>
 			<slot name="actions" class="d2l-cell" data-cell-num="6"></slot>
-
 			<slot name="content" @focus="${this._preventFocus}"></slot>
+			<slot name="nested"></slot>
```
dbatiste, line 158: This introduces a new slot for the "nested" content for the generic layout and list-item.  This enables the list-item to control the layout of the nested content in terms of the CSS grid.  For instance, it enables the nested content to span underneath the optional secondary actions on the right side - this would not be possible if the nested list was distributed to the default slot.

components/list/list-item-mixin.js
```diff
@@ -295,20 +297,24 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 		});
 	}
 
-	_onFocusIn() {
+	_onFocusIn(e) {
 		this._focusing = true;
+		e.stopPropagation();
```
dbatiste, line 302: These `stopPropagation()` calls are necessary so that outer/parent list-items don't appear focused when a list-item within a nested list is focused.

components/list/list.js
```diff
@@ -74,6 +74,20 @@ class List extends SelectionMixin(LitElement) {
 		return this._getItems().indexOf(item);
 	}
 
+	getSelectionInfo(includeNested) {
```
dbatiste, line 77: This override is necessary because determining the keys is special for lists.  I considered ways of relating the nested selection-inputs to the parents (note they are not actually parents, or even in the same DOM tree) using IDs but it gets pretty messy and fragile. So, I eventually arrived at this sol'n which is pretty simple.

components/list/list-item-checkbox-mixin.js
```diff
@@ -25,7 +26,11 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 			/**
 			 * Whether the item is selected
 			 */
-			selected: { type: Boolean, reflect: true }
+			selected: { type: Boolean, reflect: true },
+			/**
+			 * Private. The selection info (set by the selection component).
+			 */
+			selectionInfo: { type: Object }
```
dlockhart, line 33: Should this have `attribute: false` so that people can't set the attribute in the HTML and have it bind?

components/list/list-item-checkbox-mixin.js
```diff
@@ -59,6 +65,32 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		if (!this.key) this.setSelected(undefined, true);
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		if (this.selectable) {
+			this.shadowRoot.querySelector('slot[name="nested"]').addEventListener('slotchange', e => {
```
dlockhart, line 71: Could also ask consumers to do this wire-up in their `render()`, which would ensure that the event listener gets removed.

dlockhart (commented):
Looking good to me!

dlockhart (approved):

PR #1672, opened 2021-09-10
HierarchicalView performance fix

dlockhart:
In hierarchical view's `connectedCallback`, it's making a call to `this.__autoSize()` (inside an rAF). That method checks `offsetParent` and gets a bounding rectangle -- both of which force layout. But hierarchical view is almost always inside a menu, which is almost always hidden inside a dropdown -- so there's nothing to calculate (yet).

This change adds an intersection observer (my new favourite toy) to determine when the hierarchical view first becomes visible. At that point it can automatically size itself.

### DIFF
--- a/components/hierarchical-view/hierarchical-view-mixin.js
+++ b/components/hierarchical-view/hierarchical-view-mixin.js
@@ -113,7 +113,21 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 
 		this.__isChildView();
 
+		if (typeof(IntersectionObserver) === 'function') {
+			this.__intersectionObserver = new IntersectionObserver((entries) => {
+				entries.forEach((entry) => {
+					if (entry.isIntersecting) {
+						this.__autoSize(this);
+					}
+				});
+			});
+			this.__intersectionObserver.observe(this);
+		}
+
 		requestAnimationFrame(() => {
+			if (typeof(IntersectionObserver) !== 'function') {
+				this.__autoSize(this);
+			}
 			this.__startResizeObserver();
 
 			if (!this.childView) {
@@ -131,7 +145,10 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 		this.removeEventListener('focus', this.__focusCapture);
 		this.removeEventListener('focusout', this.__focusOutCapture);
 		window.removeEventListener('resize', this.__onWindowResize);
-		if (this.__intersectionObserver) this.__intersectionObserver.disconnect();
+		if (this.__intersectionObserver) {
+			this.__intersectionObserver.disconnect();
+			this.__isAutoSized = false;
+		}
 	}
 
 	firstUpdated(changedProperties) {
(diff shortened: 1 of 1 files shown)


components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -152,6 +156,20 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 		this.addEventListener('keypress', stopPropagation);
 
 		this.__isChildView();
+
```
dlockhart, line 159: Since this code is running in `firstUpdated` it's not going to work nicely if the view is disconnected and re-connected. But neither are all the event listeners being wired up above, so I just added to the problem for now. :)

components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -97,9 +97,13 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 	constructor() {
 		super();
 
+		/** @ignore */
+		this.childView = false;
```
dlockhart, line 101: This property wasn't being initialized.

components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -261,16 +279,20 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 	}
 
 	__autoSize(view) {
-		if (this.childView || view.offsetParent === null) return;
+		if (this.__isAutoSized || this.childView) return;
```
dlockhart, line 282: Logic mostly stays the same here, I just moved the short-circuit for `this.childView` up higher, added a new short-circuit so this only runs on the first intersection, and wrapped everything in a `rAF`.

components/hierarchical-view/hierarchical-view-mixin.js
dbatiste: I think the ones immediately above aren't removed in disconnect, and the ones that are removed in the disconnected callback are added in the connected callback.  Two place this might be a concern are cards and list-items where those elements can be drag & dropped.

dbatiste:
> But hierarchical view is almost always inside a menu, which is almost always hidden inside a dropdown -- so there's nothing to calculate (yet).

One place that I recall uses the menu outside of dropdown is the mobile nav menu, but it too is hidden initially.  That would be a good thing to check if you haven't yet.  There is also the "Find it Fast" widget that uses the menu on the home-page, but I'm not sure if that's just an internal Spark thing.

components/hierarchical-view/hierarchical-view-mixin.js
dlockhart: Ok good call. I've moved the code around such that the `IntersectionObserver` now works when it's disconnected and then re-connected. I tested it in drag & drop lists that have menus, as well as the my-courses widget when the tiles move around.

Also tested out the mobile menu -- works well.

dbatiste (approved):

PR #1687, opened 2021-09-14
US129220 - Mobile - Open date-picker on icon click, remove tray from time-picker

annabelsegalld2l:
The testing brought up a couple of issues with the dropdown time and date trays, so I made the following fixes:

- Mobile tray is removed from the time-picker altogether. It was just one line to disable it, so it will be easy to bring back again once we have ideas for a better UX there. 
- Date picker now uses the auto-fit for the tray. This allows the calendar to be scrollable on super tiny screens shorter than 420px. 
- I've changed the opening behaviour for mobile as follows:

1. Clicking the calendar/error icon will open up the calendar and focus on it with NO text input. 
2. Clicking elsewhere on the time input will open a text input keyboard ONLY, no calendar (which would be super small and unusable with the keyboard also on the screen)
3. Using the down key or enter key on the input box will open the calendar and focus on it, with NO text input. The calendar is navigable with arrows. This is consistent with our tooltip that says "Arrow down or press enter to access mini-calendar".


components/inputs/input-date.js
```diff
@@ -438,13 +440,16 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 	_handleKeydown(e) {
 		// open dropdown on down arrow or enter and focus on calendar focus date
 		if (e.keyCode === 40 || e.keyCode === 13) {
+			this._openedOnKeydown = true;
```
dlockhart, line 443: Would be good to default this and `_openCalendarOnly` to `false` in the constructor, otherwise they're `undefined`.

components/inputs/input-date.js
```diff
@@ -438,13 +440,16 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
 			this.opened = true;
 			e.preventDefault();
 		}
 	}
 
-	_handleMouseup() {
+	_handleMouseup(e) {
 		this._inputTextFocusMouseup = true;
+		const icon = this.shadowRoot.querySelector('d2l-icon');
+		this._openCalendarOnly = e.srcElement.tagName === icon.tagName;
```
dlockhart, line 452: Doesn't seem necessary to get the icon using its tag name and then compare the tag name...

```suggestion
		this._openCalendarOnly = e.srcElement.tagName === 'd2l-icon';
```

components/inputs/input-date.js
```diff
@@ -232,7 +234,7 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 				@d2l-dropdown-focus-enter="${this._handleFocusTrapEnter}"
 				max-width="335"
 				min-height="${this._hiddenCalendarHeight}"
-				no-auto-fit
+				?no-auto-fit="${!mediaQueryList.matches}"
```
dlockhart, line 237: Technically I think if `matches` changes (if the user resizes their window) the control isn't going to re-render. That's probably fine though since we really only do this when testing, but something to be aware of.

dlockhart (approved):

PR #1700, opened 2021-09-15
Daylight: dropdown

GeorgePlukov:


components/dropdown/README.md
```diff
@@ -1,10 +1,12 @@
+
 # Dropdowns
-## Dropdown Openers
+A Dropdown is a button that opens a floating container to offer menu items or other content.
 
-### d2l-dropdown
+## Opener: Dropdown [d2l-dropdown]
```
GeorgePlukov, line 5: Not sure how I feel about these titles, I thought some method of grouping openers/content would be ideal. Maybe these get split into a separate page?

components/dropdown/README.md
```diff
@@ -1,10 +1,12 @@
...
 
 `d2l-dropdown` is a generic opener for dropdown content (`d2l-dropdown-content`, `d2l-dropdown-menu` or `d2l-dropdown-tabs`) enabling alternate opener implementation using existing elements/components. Provide and indicate your own opener element with the class attribute value `d2l-dropdown-opener`.  Wire-up is automatic.
 
+<!-- docs: demo live name:d2l-dropdown autoSize:false align:flex-start size:medium -->
```
GeorgePlukov, line 9: The demos are currently static sized (`medium`) but the dropdown's start closed. @margaree To match what you did with date/time do you think that these should begin open (or maybe just the ones dealing with dropdown content)?

margaree: Check with Jeff to be sure but I think we'd likely want to have them open as well and adopt a similar strategy as that page for how `opened` gets handled.

components/dropdown/README.md
margaree: Yeah for sure - I like that you added something to distinguish the type. Can probably leave as is for now and see what Jeff thinks.

components/dropdown/README.md
```diff
@@ -88,28 +85,33 @@ dropdown.addEventListener('click', function() {
...
 | `description` | String | A description to be added to the inner `button` opener for accessibility |
 | `disabled` | Boolean, default: `false` | Disables the dropdown opener |
 | `no-auto-open` | Boolean, default: `false` | Prevents the dropdown from automatically on click or on key press |
+<!-- docs: end hidden content -->
 
-**Accessibility:**
+### Accessibility
 
-To make your `d2l-dropdown-button-subtle` accessible, use the following properties when applicable:
+To make your `d2l-dropdown-button-subtle` accessible, use the following properties when applicable
```
margaree, line 101: nit: probably want to keep the :

components/dropdown/README.md
```diff
@@ -224,34 +231,35 @@ To make your usage of `d2l-dropdown-more` accessible, use the following property
...
+| `header` | Sticky container at the top of the dropdown |
+| `footer` | Sticky container at the bottom of the dropdown |
+| `default` | Anything inside of `d2l-dropdown-content` that isn't in the `header` or `footer` slots appears as regular content |
+<!-- docs: stop hidden content -->
 
-**Methods:**
-* `async resize()`: Call if the size of the content changes due to a change in
-a nested component. The nested component may choose to fire a custom event, which
-the component containing the `d2l-dropdown-content` can catch and call this method.
+### Methods
```
margaree, line 249: For a page I previously worked on Jeff had mentioned preferring leaving the methods as bullet points

components/dropdown/README.md
```diff
@@ -224,34 +231,35 @@ To make your usage of `d2l-dropdown-more` accessible, use the following property
...
-  Some Content...
-  <div slot="footer">Footer Content</div>
-</d2l-dropdown-content>
-```
+| Name | Description |
+|---|---|
+| `async resize()` | Call if the size of the content changes due to a change in a nested component. The nested component may choose to fire a custom event, which the component containing the `d2l-dropdown-content` can catch and call this method. |
 
-### d2l-dropdown-menu
+### Content: Menu [d2l-dropdown-menu]
```
margaree, line 255: I believe this should be an h2

components/dropdown/README.md
```diff
@@ -284,11 +292,14 @@ the component containing the `d2l-dropdown-content` can catch and call this meth
 </d2l-dropdown-button>
 ```
 
-### d2l-dropdown-tabs
+### Content: Tabs [d2l-dropdown-tabs]
```
margaree, line 295: Also h2

components/dropdown/README.md
GeorgePlukov: Reverted

components/dropdown/dropdown-content.js
```diff
@@ -9,6 +9,7 @@ import { dropdownContentStyles } from './dropdown-content-styles.js';
  * @slot footer - Sticky container at the bottom of the dropdown
  * @fires d2l-dropdown-open - Dispatched when the dropdown is opened
  * @fires d2l-dropdown-close - Dispatched when the dropdown is closed
+ * @fires d2l-dropdown-focus-enter - Dispatched when user focus enters the dropdown content
```
annabelsegalld2l, line 12: Can we have some way to specify that this event only fires when the trap-focus option is on? Maybe just a "(trap-focus option only)" at the end of the line. Otherwise consumers might think that it happens whenever the user clicks inside the dropdown.

components/dropdown/README.md
```diff
@@ -21,190 +29,230 @@
...
+<script>
+  window.addEventListener('load', function () {
+    var dropdown = document.querySelector('d2l-dropdown-button');
+	setTimeout(() => {
+		dropdown.toggleOpen();
+	}, 100);
+  });
+</script>
 <d2l-dropdown-button text="Open!" primary>
   <d2l-dropdown-content min-width="150" max-width="400">
```
annabelsegalld2l, line 246: Do we want to add a mobile tray (ie. mobile-tray="bottom") to display the trays somewhere on this page? I was thinking to maybe add one of each tray as a demo, but putting it in with this example should be fine too. Not sure what the default size is for the snippets but it would activate at 616px.

components/dropdown/README.md
```diff
@@ -21,190 +29,230 @@
...
-});
-```
-
-**Properties:**
+<!-- docs: start hidden content -->
+### Properties:
 
 | Property | Type | Description |
-|--|--|--|
+|---|---|
```
margaree, line 36: This table isn't actually working (I think it needs another `|---|`)

components/dropdown/README.md
```diff
@@ -21,190 +29,230 @@
...
-  // fetch some content
-  ...
-  // take control of when the dropdown is actually opened
-  dropdown.toggleOpen();
-});
-```
-
-**Properties:**
+<!-- docs: start hidden content -->
+### Properties:
```
margaree, line 33:
```suggestion
### Properties
```

components/dropdown/README.md
```diff
@@ -21,190 +29,230 @@
 
 If the dropdown is initially empty when it's opened, the dropdown pointer will not be positioned correctly.  In such cases, the `no-auto-open` attribute may be added to the opener, enabling you to take control of when the dropdown is actually opened:
```
margaree, line 30: This sentence becomes a bit strange without the code below

components/dropdown/README.md
```diff
@@ -21,190 +29,230 @@
...
 <script type="module">
   import '@brightspace-ui/core/components/dropdown/dropdown-button.js';
   import '@brightspace-ui/core/components/dropdown/dropdown-content.js';
 </script>
-
+<script>
+  window.addEventListener('load', function () {
+    var dropdown = document.querySelector('d2l-dropdown-button');
+	setTimeout(() => {
+		dropdown.toggleOpen();
```
margaree, line 58: This looks to be using a mix of tabs and spaces

components/dropdown/README.md
```diff
@@ -1,16 +1,24 @@
...
 
 `d2l-dropdown` is a generic opener for dropdown content (`d2l-dropdown-content`, `d2l-dropdown-menu` or `d2l-dropdown-tabs`) enabling alternate opener implementation using existing elements/components. Provide and indicate your own opener element with the class attribute value `d2l-dropdown-opener`.  Wire-up is automatic.
 
+<!-- docs: demo live name:d2l-dropdown autoSize:false align:flex-start size:medium -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/dropdown/dropdown.js';
   import '@brightspace-ui/core/components/dropdown/dropdown-content.js';
 </script>
-
```
margaree, line 13: Thoughts on hiding these code chunks?

components/dropdown/README.md
GeorgePlukov: I added it as default enabled within the live demo, it won't show until you resize the demos but once you hit `616px` it does appear

components/dropdown/README.md
GeorgePlukov: I changed the `:`  to a `.`, I think it makes sense to keep this line in here. Do you think it would be useful to add a code example back?

margaree: I think so, though I think you could combine the two code chunks, e.g.:
```html
<script>

dropdown.addEventListener('click', function() {
  // fetch some content
  ...
  // take control of when the dropdown is actually opened
  dropdown.toggleOpen();
</script>
<d2l-dropdown no-auto-open>
  ...
</d2l-dropdown>
```

GeorgePlukov:
> This is looking mostly good to go. Does it have the issue with jumpiness when the demos load after changing values in the properties table? I had that issue with date/time so I added a check for first load

Added that check to these too, it is a little bit annoying how it blinks open otherwise.

components/dropdown/README.md
```diff
@@ -56,9 +81,11 @@ If the dropdown is initially empty when it's opened, the dropdown pointer will n
 <!-- docs: start hidden content -->
 <script>
   window.addEventListener('load', function () {
-    var dropdown = document.querySelector('d2l-dropdown-button');
+    var demoElem = document.querySelector('#demo-element');
+    if (!demoElem.hasAttribute('data-first-load')) return;
     setTimeout(() => {
-      dropdown.toggleOpen();
+      var dropdown = document.querySelector('d2l-dropdown-button');
+        dropdown.toggleOpen();
```
margaree, line 88: nit: spacing (as well as a few places below)

components/dropdown/README.md
```diff
@@ -378,7 +415,10 @@ To make your usage of `d2l-dropdown-more` accessible, use the following property
 <script>
   window.addEventListener('load', function () {
     var dropdown = document.querySelector('d2l-dropdown-button');
+    const demoElem = document.querySelector('#demo-element');
+    if (!demoElem.hasAttribute('data-first-load')) return;
     setTimeout(() => {
+       var dropdown = document.querySelector('d2l-dropdown-button');
```
margaree, line 421: nit: dropdown is defined twice

margaree (approved):

PR #1701, opened 2021-09-16
Removing duplicate js import

geurts:
@GeorgePlukov, I also noticed the bullet points (Best Practices) somehow get periods added at render time. Any idea how to stop that? Couldn't find anything about it on the web.


GeorgePlukov:
@geurts I'm not seeing periods being added at render time, I think we removed the periods from the status-indicator page in a recent PR so It's possible the site you are viewing is out of date

geurts:
Thanks @GeorgePlukov I just had to rm package-lock & node modules, I didn't realize it wasn't pulling in the latest READMEs. The PR218 site is also showing the periods so it convinced me the issue wasn't local :P

GeorgePlukov (approved):

PR #988, opened 2020-12-09
Experiment 2: Replace karma with web-test-runner

margaree:
Replaces https://github.com/BrightspaceUI/core/pull/945

PR #1703, opened 2021-09-16
Switch to @web/test-runner and dev-server and add Playwright

dlockhart:
I've been working on this for a bit while waiting for other things -- switching away from the deprecated `@open-wc/testing-karma` and dev server to the Modern Web versions. This mostly went smoothly, although the new tooling is much better at surfacing JS errors that get logged during testing (but weren't always failing the tests) so I ended up making a few fixes.

This also adds Playwright to run in parallel with SauceLabs. We can decide after a while based on flake whether it'd be worthwhile to just remove Sauce altogether.

@margaree [started this experiment](https://github.com/BrightspaceUI/core/pull/988/files), so this is attempt 2!


README.md
```diff
@@ -110,7 +110,6 @@ npm test
 npm run test:headless
 
 # debug or run a subset of local unit tests
-# then navigate to `http://localhost:9876/debug.html`
```
dlockhart, line 113: The new CLI tooling has a nice interface that tells you to "Press D to debug in a browser" and it'll automatically open the browser and let you drill into individual tests. So this isn't needed.

components/inputs/input-date.js
```diff
@@ -369,14 +369,18 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 
 	async _handleClear() {
 		await this._updateValueDispatchEvent('');
-		this._dropdown.close();
+		if (this._dropdown) {
+			this._dropdown.close();
+		}
```
dlockhart, line 374: In the tests there were a few places `this._dropdown` wasn't set.

components/inputs/test/input-date.test.js
```diff
@@ -25,7 +25,7 @@ describe('d2l-input-date', () => {
 	describe('accessibility', () => {
 
 		it('should set description when empty-text', async() => {
-			const elem = await fixture('<d2l-input-date empty-text="text description"></d2l-input-date>');
+			const elem = await fixture('<d2l-input-date empty-text="text description" label="label text"></d2l-input-date>');
```
dlockhart, line 28: Was causing warnings to show up in the test output.

components/list/test/list-item-drag-mixin.test.js
```diff
@@ -106,6 +106,9 @@ describe('ListItemDragDropMixin', () => {
 
 		it('should have the list item go back to normal when dragging ends', async() => {
 			const dragArea = element.shadowRoot.querySelector('.d2l-list-item-drag-area');
+			dispatchDragEvent(dragArea, 'dragstart');
+			await oneEvent(dragArea, 'dragstart');
+			await element.updateComplete;
```
dlockhart, line 111: This was causing a JS error but wasn't failing the test. The start event needs to fire first before the end event.

components/tooltip/test/tooltip.test.js
```diff
@@ -194,29 +194,22 @@ describe('d2l-tooltip', () => {
 
 	describe('delay', () => {
 
-		beforeEach(async() => {
-			tooltip.setAttribute('delay', 100);
-			await tooltip.updateComplete;
```
dlockhart, line 199: There's a 300ms delay by default now, so this isn't needed.

components/tooltip/test/tooltip.test.js
```diff
@@ -194,29 +194,22 @@ describe('d2l-tooltip', () => {
...
-		});
-
 		it('should not show if hover is lost before tooltip delay finishes', async() => {
-
-			oneEvent(tooltipFixture, 'd2l-tooltip-show').then(() => {
-				expect.fail('tooltip should not have been shown');
```
dlockhart, line 205: This wasn't working at all and was just failing silently. It required an actual fix to tooltip.js below.

components/tooltip/tooltip.js
```diff
@@ -425,6 +425,7 @@ class Tooltip extends RtlMixin(LitElement) {
 		this._removeListeners();
 		window.removeEventListener('resize', this._onTargetResize);
 		clearDismissible(this._dismissibleId);
+		delayTimeoutId = null;
```
dlockhart, line 428: There's code here that bypasses the delay when you hover on one tooltip and then move to another. But with the unit tests this causes the "delay tests" to fail because the tooltip appears immediately after a previous test run. The fix is to reset the delay bypass when a tooltip is disconnected.

mixins/labelled-mixin.js
```diff
@@ -102,6 +100,9 @@ export const LabelledMixin = superclass => class extends superclass {
 
 		});
 
+		if (!labelElem) return;
+		const ancestor = getCommonAncestor(this, labelElem);
```
dlockhart, line 104: The call to `getCommonAncestor` was throwing an error when `labelElem` was `null` -- which it was intentionally in one of the tests.

web-test-runner.config.mjs
```diff
@@ -0,0 +1,47 @@
...
+	groups: [
+		{
+			name: 'aXe',
+			files: getPattern('axe'),
+		},
+		{
+			name: 'perf',
+			files: getPattern('perf')
+		}
+	],
```
dlockhart, line 17: These groups are sweet!

mixins/labelled-mixin.js
```diff
@@ -43,7 +43,6 @@ export const LabelMixin = superclass => class extends superclass {
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		this.removeEventListener('d2l-label-change', this._handleLabelChange);
-		if (this._labelObserver) this._labelObserver.disconnect();
```
svanherk, line 46: Why is this not needed?  I think I recently added it when I was adding it to https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-observer-mixin.js#L44, which is a similar setup

dlockhart: Sorry I meant to call this out -- it wasn't needed because `LabelMixin` never creates an observer, only `LabelledMixin` does.

svanherk: Ah!  Took me way to long to see that, the way GitHub minimizes the diff - and I obviously missed it the first time when adding it.  Should it be moved down to `LabelledMixin`?

dlockhart: @dbatiste and I talked a bit about that. The problem with disconnecting it is if it gets reconnected later (maybe via a drag & drop) then it won't start observing again. And that change was larger than I wanted to make here where I just happened to notice it.

dlockhart:
The default configuration uses the W3C webdriver capabilities, instead of the JWP ones we're using today. JWP is called out as "legacy" by SauceLabs. So that explains why we're seeing some failures that we weren't before. I'm going to try and go through them each and fix them up, but could be a bit.

web-test-runner.sauce.config.mjs
```diff
@@ -0,0 +1,55 @@
...
+	sauceLabsOptions,
+	sauceLabsCapabilities
+);
+
+const extraOptions = {
+	idleTimeout: 500 // default 90
+};
+
+config.concurrentBrowsers = 4; // concurrent browsers
+config.concurrency = 1; // concurrent tests in a single browser
```
dlockhart, line 27: Huge 🤦  moment: this defaulted to `6` which meant it was running 6 tests at a time in each browser... which was super fast but was also the reason for all the broken tests!

So after two days of unknowingly trying to fix tests so they can run in parallel, just setting this to `1` solves all the problems. It'd be great to run stuff in parallel, and maybe one day we can break our tests out into "isolated" vs. not... but for now I don't think it's realistic as long as we're reusing fixtures across tests for them to be able to run at the same time.

tools/resize-observer-test-error-handler.js
```diff
@@ -1,15 +1,5 @@
...
-		if (err === 'ResizeObserver loop limit exceeded') {
-			console.warn('Ignored: ResizeObserver loop limit exceeded');
-			return false;
-		} else if (err === 'ResizeObserver loop completed with undelivered notifications.') {
-			console.warn('Ignored: ResizeObserver loop completed with undelivered notifications');
-			return false;
-		} else {
-			return e(...arguments);
-		}
-	};
```
dlockhart, line 14: I'd originally deleted this and moved it inline into the configuration, but then realized that the [create initializer](https://github.com/BrightspaceUI/create/blob/main/src/generators/test-unit/templates/static/karma.conf.js#L14) uses it which means lots of folks' projects are using it.

The original way this was written by hooking into the `before` hook no longer worked, but I found this to work equally well.

svanherk (approved):
Love the new progress output!  Those filter test logs will be gone after https://github.com/BrightspaceUI/core/pull/1724 merges

PR #1704, opened 2021-09-16
US131759 - Simplify manual search for filter

svanherk:
Having the consumer remove the values when doing manual search caused lots of complications with single selection, searching, and clearing.  Instead, they will now pass us an array of the keys they want displayed for that search value.

This PR also fixes the issue where selecting + single selection + manual search is broken.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -395,8 +395,9 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 
 		if (!toUpdate) return;
 
-		let shouldUpdate = false;
-		let shouldRecount = false;
+		let shouldUpdate = false,
+			shouldSearch = false,
+			shouldRecount = false;
 		changes.forEach((newValue, prop) => {
 			if (toUpdate[prop] === newValue) return;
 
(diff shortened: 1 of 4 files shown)


components/filter/filter.js
```diff
@@ -412,10 +413,12 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 					this._totalAppliedCount--;
 				}
 			} else if (prop === 'values') {
-				if (dimension.searchType !== 'manual' || !dimension.searchValue) shouldRecount = true;
```
svanherk, line 415: When values are added/removed we can now always recount, and we'll need to re-search if a search was applied.

components/filter/filter.js
```diff
@@ -452,7 +455,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		const searchValue = e.detail.value.trim();
 		dimension.searchValue = searchValue;
 
-		if (dimension.searchType === 'automatic') {
+		if (dimension.searchType === 'automatic' || searchValue === '') {
```
svanherk, line 458: I figured it'd be easier and more performant for the filter to always handle clearing the search, no matter the search-type, since that's just setting everything to `hidden = false`

dlockhart (approved):
Looks like this should work nicely!

PR #1707, opened 2021-09-16
Update d2l-selection-input to not delay dispatching d2l-selection-change.

dbatiste:
This PR updates `d2l-selection-input` to not delay dispatching `d2l-selection-change` event with rAF.  In the single selection scenario, two events are dispatched if a previously selected item becomes unselected as a result of a newly selected item.  By not delaying the dispatch statement, the two events should be dispatched in the same frame, making consumer code less fragile.  Also, I think this also improves some jankiness observed when toggling the selected state.

### DIFF
--- a/components/list/list-item-checkbox-mixin.js
+++ b/components/list/list-item-checkbox-mixin.js
@@ -56,6 +56,19 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		this._checkboxId = getUniqueId();
 	}
 
+	get selectionInfo() {
+		return this._selectionInfo;
+	}
+
+	set selectionInfo(val) {
+		const oldVal = this._selectionInfo;
+		if (oldVal !== val) {
+			this._selectionInfo = val;
+			this.selected = (this._selectionInfo.state === SelectionInfo.states.all);
+			this.requestUpdate('selectionInfo', oldVal);
+		}
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
 		if (this.selectable) {
--- a/components/list/test/list-item-checkbox-mixin.test.js
+++ b/components/list/test/list-item-checkbox-mixin.test.js
@@ -32,10 +32,10 @@ describe('ListItemCheckboxMixin', () => {
 	describe('Does not render checkbox or action area when not selectable', () => {
 		const cases = [{
 			input: '',
-			expected: { selectable: undefined, disabled: undefined, selected: undefined }
+			expected: { selectable: undefined, disabled: undefined, selected: false }
 		}, {
 			input: 'disabled',
-			expected: { selectable: undefined, disabled: true, selected: undefined }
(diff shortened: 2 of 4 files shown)


components/list/list-item-checkbox-mixin.js
```diff
@@ -56,6 +56,19 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
...
 
+	get selectionInfo() {
+		return this._selectionInfo;
+	}
+
+	set selectionInfo(val) {
+		const oldVal = this._selectionInfo;
+		if (oldVal !== val) {
+			this._selectionInfo = val;
+			this.selected = (this._selectionInfo.state === SelectionInfo.states.all);
```
dbatiste, line 67: Not delaying the event caused some trouble with the selection behaviour of list-items where a parent list-item (of a nested list) would become selected and then unselected because the `SelectionMixin` updates `selectionInfo` of observers (i.e. parent list-items in this case) causing a rerender while the selected state is false, thereby undoing the selection state of the `d2l-selection-input` it renders. 

This change ensures that when the `SelectionMixin` updates the `selectionInfo` property of the list-item, the `selected` property is also updated so that when the list-item re-renders, it does not incorrectly set the `d2l-selection-input` `selected` state back to false.

components/list/test/list-item-checkbox-mixin.test.js
```diff
@@ -32,10 +32,10 @@ describe('ListItemCheckboxMixin', () => {
 	describe('Does not render checkbox or action area when not selectable', () => {
 		const cases = [{
 			input: '',
-			expected: { selectable: undefined, disabled: undefined, selected: undefined }
+			expected: { selectable: undefined, disabled: undefined, selected: false }
```
dbatiste, line 35: I'm not totally sure why these didn't fail when I added the initial state assignment in my last change.  It must have something to do with the timing.  Regardless, the expected value is `false` now.

dlockhart (approved):

svanherk:
One thing I'm noticing that I think is a change in behavior - before, if a list with single-selection had two values initially `selected`, it would keep the first one selected and undo the other one.  Whereas now, it keeps them both selected. So this:
```
<d2l-list selection-single>
  <d2l-list-item selectable selected key="1" label="Introductory Earth Sciences">
    Stuff...
  </d2l-list-item>
  <d2l-list-item selectable selected key="2" label="Engineering Materials for Energy Systems">
    Stuff...
  </d2l-list-item>
  <d2l-list-item selectable key="3" label="Geomorphology and GIS">
    Stuff...
  </d2l-list-item>
```
Causes this on page load:
image

I'm not sure why you'd have two `selected` by default, but before (on `main`) it does this:
image

dbatiste:
Ok yeah.  That's an invalid usage, but I'll see if I can get the original behaviour.

svanherk (approved):
Nice!  It looks like the last update also solves the newer race condition issue I was seeing in that test

PR #1709, opened 2021-09-17
Daylight: collapsible containers

margaree:
This combines more-less and expand-collapse into one page which I put in the existing expand-collapse page (though perhaps it makes sense to put it somewhere completely different and then link out both expand-collapse and more-less?).


components/expand-collapse/README.md
```diff
@@ -1,36 +1,76 @@
...
 ```html
 <script type="module">
+  import '@brightspace-ui/core/components/button/button.js';
   import '@brightspace-ui/core/components/expand-collapse/expand-collapse-content.js';
-</script>
 
-<d2l-expand-collapse-content expanded>
+  const button = document.querySelector('d2l-button');
+  button.addEventListener('click', () => {
+    const section = document.querySelector('d2l-expand-collapse-content');
```
GeorgePlukov, line 19: Maybe a good spot to try out `<!-- docs: hidden content-->`?

GeorgePlukov (approved):

components/expand-collapse/README.md
margaree: I think I'll actually leave this one as is for now since it does advice on how to actually implement.

PR #1714, opened 2021-09-20
fix: DE45163 - query the useMobileStyling property rather than the mediaQueryList 

annabelsegalld2l:
This solves the weird behaviour when the data-mobile property sometimes does not get set after the screen changes to mobile, causing some funky styling. See GIFs below. It might also be a performance improvement since we can rely on the property rather than asking the mediaQueryList if we are in mobile mode. This fix should also fix similar problems with the daylight site as well. 

Before:
tray-resize-bug

After:
tray-resize-fixed

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -191,8 +191,8 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			},
 			_useMobileStyling: {
 				type: Boolean,
-				attribute: 'data-mobile',
-				reflect: true
+				reflect: true,
+				attribute: 'data-mobile'
 			},
 			_hasHeader: {
 				type: Boolean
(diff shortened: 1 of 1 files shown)


components/dropdown/dropdown-content-mixin.js
```diff
@@ -787,9 +789,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			}
 		}
 
-		const specialMobileStyle = this.mediaQueryList.matches && this.mobileTray;
-		const mobileTrayRightLeft = this.mediaQueryList.matches && (this.mobileTray === 'right' || this.mobileTray === 'left');
-		const mobileTrayBottom = this.mediaQueryList.matches && (this.mobileTray === 'bottom');
+		const specialMobileStyle = this._useMobileStyling && this.mobileTray;
```
annabelsegalld2l, line 792: There seemed to be some sort of race condition where the _handleMobileResize function detects the screen change and sets the _useMobileStyling attribute to true, but it never gets reflected into an attribute on the dropdown. I've fixed it by changing all the mediaQueryList.matches to checking the property directly, to prevent any of the mismatches that caused the styling issues.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -771,8 +771,10 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this.dispatchEvent(new CustomEvent('d2l-dropdown-focus-enter'));
 	}
 
-	_handleMobileResize() {
+	async _handleMobileResize() {
 		this._useMobileStyling =  this.mediaQueryList.matches;
+		this._showBackdrop = this._useMobileStyling && this.mobileTray;
```
annabelsegalld2l, line 776: If the mobile styling attribute is changed, also check if the backdrop should be shown and if opened, reposition the dropdown.

dlockhart (approved):

PR #1719, opened 2021-09-21
Cleaning up dropdown-content-mixin code

annabelsegalld2l:
- Separating out each of the right/left. bottom tray, and regular dropdown styles into separate functions. It looks a lot cleaner to look at each type specifically and we can also ensure the wrong styles aren't being applied from another dropdown type. It also makes the renderContent function a lot shorter.
- There is more duplication here since each of the types applies a widthStyle/ContentWidthStyle.
- I decided to have a map that stores each styleMap with a key that describes which element it will be applied to.


dlockhart (approved):
Cool -- definitely makes `render()` way cleaner!

PR #1724, opened 2021-09-21
US125557 - Add clear button to filter

svanherk:
Feels like you've seen this fifteen times before, but this is THE PR that adds a clear button.  The filter handles all clearing and the regular change events are sent, but there is an added `cleared` property in the event if teams want to listen to this for perf enhancements.


components/filter/filter.js
```diff
@@ -397,6 +409,20 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 
+	_handleClear() {
+		const dimension = !this._activeDimensionKey ? this._dimensions[0] : this._dimensions.find(dimension => dimension.key === this._activeDimensionKey);
+
+		this._performDimensionClear(dimension);
+		this._dispatchChangeEventNow();
+		this.requestUpdate();
+
+		if (!this._activeDimensionKey) {
+			announce(this.localize('components.filter.clearAnnounceSingle'));
```
annabelsegalld2l, line 420: This announces that the filters are in the process of clearing. Is there going to be some way to know when they are done clearing? Does more need to be done after _performDimensionClear is called? Just making sure I understand the flow.

components/filter/filter.js
```diff
@@ -397,6 +409,20 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		this._dispatchChangeEvent(dimension, { valueKey: valueKey, selected: selected });
 	}
 
+	_handleClear() {
+		const dimension = !this._activeDimensionKey ? this._dimensions[0] : this._dimensions.find(dimension => dimension.key === this._activeDimensionKey);
```
dlockhart, line 413: I see this code repeated... might be worth throwing it in a `_getActiveDimensionOrFirst()` or something.

components/filter/filter.js
svanherk: Yeah I talked to Jeff about this - ultimately it's up to the consumer to announce when the actual data on the page is done loading/updating/clearing.  We don't have a way to tell when this happens and it's probably specific to the page.  But I think it'd be a good idea to this call out - I'll add an accessibility section to remind the consumer of this!  Good point!

components/inputs/test/input-time-range.visual-diff.js
```diff
@@ -70,7 +70,7 @@ describe('d2l-input-time-range', () => {
 		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 	});
 
-	describe('opened behavior', () => {
+	describe.skip('opened behavior', () => {
```
svanherk, line 73: AS discussed in standup, skipping these and opened https://github.com/BrightspaceUI/core/issues/1749

dlockhart (approved):

PR #1726, opened 2021-09-22
US131354 - Add optional icon property to count-badge

annabelsegalld2l:
This PR adds the icon property, which takes in an icon string to be rendered with the number. I also updated the focus-visible states for the number-only badge to add the D2L focus ring.

I've now got a CountBadgeMixin that contains all the number rendering. The plain count badge and icon count badge now inherit this mixin. I've also split out the number rendering and the tooltips/offscreen rendering into two functions to add some flexibility for the consumers to place the tooltips inside or beside their wrappers.


components/count-badge/count-badge.js
```diff
@@ -34,6 +36,22 @@ class CountBadge extends RtlMixin(LitElement) {
...
+			 */
+			icon: {
+				type: String,
+				reflect: true
+			},
+			/**
+			 * Optionally specify a size for the height/width of the icon.
+			 * @type {string}
+			 */
+			iconSize: {
```
annabelsegalld2l, line 51: This size fixing is required for d2l-card-footer-link to cap all the icons at 18*18px. I don't expect any consumers to use this property directly. Any thoughts on making this private and not exposing the functionality?

components/count-badge/count-badge.js
```diff
@@ -172,19 +226,37 @@ class CountBadge extends RtlMixin(LitElement) {
...
+				${this.icon ? html`<d2l-icon role="presentation" id="${this._badgeId}" icon="${ifDefined(this.icon)}" class="d2l-button-icon" style=${styleMap(iconSizeStyle)}></d2l-icon></div>` : null }
+				${this.hasTooltip ?
+		html`<d2l-tooltip id="${this._textId}" ?force-show="${this._forceTooltipOn}" aria-live="${this.announceChanges ? 'polite' : 'off'}" for="${this._badgeId}">${this.text}</d2l-tooltip>`
+		: html`<span id="${this._textId}" aria-live="${this.announceChanges ? 'polite' : 'off'}" class="d2l-offscreen">"${this.text}"</span>`}
+			<div>
 			`;
 	}
+
+	_onBlur() {
+		this._forceTooltipOn = false;
```
annabelsegalld2l, line 254: Forcing the tooltip on is required when the icon property is set, as we want the tooltip to be positioned on the icon rather than the number or the whole badge. The tooltip's for property then gets set to the icon, meaning it only shows up when hovering on the icon directly. Forcing the tooltip on focus ensures the user can hover anywhere and focus on the badge and still get the tooltip.

components/count-badge/count-badge.js
```diff
@@ -172,19 +293,37 @@ class CountBadge extends RtlMixin(LitElement) {
 		if (this.maxDigits && this.number.toString().length > this.maxDigits) {
 			numberString = `${'9'.repeat(this.maxDigits)}+`;
 		}
+		const iconSizeStyle = {};
+		if (this.icon && this._fixIconSizeTier1) {
```
annabelsegalld2l, line 297: This size fixing is required for d2l-card-footer-link to cap all the icons at 18*18px. I don't expect any consumers to use this property directly. Any thoughts on keeping this private and not exposing the functionality in the readme?

components/count-badge/count-badge.js
```diff
@@ -172,19 +293,37 @@ class CountBadge extends RtlMixin(LitElement) {
...
 			aria-labelledby="${ifDefined(this.hasTooltip ? undefined : this._textId)}">
-					<div aria-hidden="true">${numberString}</div>
-					${this.hasTooltip ?
-		html`<d2l-tooltip id="${this._textId}" for="${this._badgeId}" aria-live="${this.announceChanges ? 'polite' : 'off'}">${this.text}</d2l-tooltip>`
-		: html`<span id="${this._textId}" class="d2l-offscreen" aria-live="${this.announceChanges ? 'polite' : 'off'}">"${this.text}"</span>`}
-			</div>
+				<div class="d2l-count-badge-number">
+						<div aria-hidden="true">${numberString}</div>		
+				</div>
+				${this.icon ? html`<d2l-icon role="presentation" id="${this._badgeId}" icon="${ifDefined(this.icon)}" class="d2l-button-icon" style=${styleMap(iconSizeStyle)}></d2l-icon>` : null }
```
dlockhart, line 312: Do you need the `ifDefined` check on the `icon` attribute since if it's `undefined` it would already be rendering `null`.

`role="presentation"` also shouldn't be needed -- if it is then we should be doing something internally in `<d2l-icon>` for that.

components/count-badge/count-badge.js
dlockhart: Yikes... that's complicated, and I think is going to mess up the tooltip's automatic adding of `aria-labelledby`... since the icon itself is presentational.

Would it be worth considering what splitting this into 2 components might look like? i.e. `<d2l-count-badge>` and `<d2l-count-badge-icon>`? Seems like we have a bunch of things that shift significantly when there's an icon and certain attribute/behaviours that get forced on. Lots of CSS that's icon-specific as well.

components/count-badge/count-badge.js
dlockhart: I think you can do this purely in CSS:

```css
:host([fixed-icon-size]) d2l-icon {
  --d2l-icon-height: 18px;
  --d2l-icon-width: 18px;
}
```

components/count-badge/count-badge.js
```diff
@@ -92,52 +110,153 @@ class CountBadge extends RtlMixin(LitElement) {
...
+		}
+		:host([size="large"][icon^="tier3"]) d2l-icon {
+			margin-top: -1.6rem;
+		}
+		:host([size="small"][icon][fixed-icon-size]) d2l-icon {
+			margin-top: -1.8rem;
+		}
+		:host([size="large"][icon][fixed-icon-size]) d2l-icon {
+			margin-top: -2.1rem;
+		}
```
dlockhart, line 256: That's a lot of positioning CSS! It's hard to figure out what it's all for... but there must be a better way.

components/count-badge/README.md
```diff
@@ -24,6 +25,8 @@ The `d2l-count-badge` element is a web component to display a number count, eith
 | `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and count badges are grey. |
 | `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
 | `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
+| `icon`, default: no icon | String | Optionally choose an icon to render the badge above. |
+| `icon-size`, default: none | Boolean | Optionally choose a fixed height/width for the icon. Defaults to icon's respective tier size. |
```
dlockhart, line 29: This is `fixed-icon-size`, ya? Could you give some more detail about the scenarios for which the default tier size isn't good enough? Just trying to think of other ways around this.

annabelsegalld2l: Yep, this comment was from before I moved it into a private property. The one case we have to have this for is d2l-card-footer link, which forces all icons to 18*18. That's really the only case, which consumers of the card-footer-link don't have to worry about. That's why I wanted to sort of hide it in a private method, it should really only be used in that case where the icons must be fixed for formatting purposes.

components/count-badge/count-badge.js
annabelsegalld2l: Yeah, that's due to the 6 combos we need to support: the three icon tiers and the large and small numbers 😞 Each combo needs different margins to get the number to render 0.55rem/0.65 above and to the right of the icon.

components/count-badge/README.md
dlockhart: Would we be able to audit the uses of the card footer link and just make sure that they're all tier1 icons and if not, switch them to tier1 icons?

It's just that we don't do this kind of thing anywhere else (e.g. `<d2l-button-icon>`, `<d2l-button-subtle>`) -- we just assume that people are using tier1 icons with them and if they aren't, things look bad.

components/count-badge/count-badge.js
dlockhart: One thing you could experiment with would be to make the number absolutely positioned when an icon is present and the `.d2l-count-badge-wrapper` `position: relative` -- all the more reason to split it into 2 components though.

components/count-badge/README.md
annabelsegalld2l: Looks like they are all tier1 icons, you're right. That will simplify the code a bit!

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,139 @@
+import '../colors/colors.js';
+import '../icons/icon.js';
+import { CountBadgeMixin, countBadgeStyles } from './count-badge-mixin.js';
+import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { offscreenStyles } from '../offscreen/offscreen.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { styleMap } from 'lit-html/directives/style-map';
+
+class CountBadgeIcon extends CountBadgeMixin(RtlMixin(LitElement)) {
```
annabelsegalld2l, line 9: I've tried splitting the badge into the two components and it seems to be working well. Still thinking about how to remove the forceTooltipOn property.

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,196 @@
...
+			numberString = `${'9'.repeat(this.maxDigits)}`;
+			numberString = this.localize('components.count-badge.plus', { number: numberString });
+		}
+		return html`
+			<div 
+			class="d2l-count-badge-wrapper"
+			id="${ifDefined(tooltipElement ? undefined : this._badgeId)}"
+			tabindex="${ifDefined((this.tabStop || this.hasTooltip) && (!hideNumber || tooltipElement) ? '0' : undefined)}" 
+			aria-labelledby="${ifDefined(this.hasTooltip ? undefined : this._textId)}"
+			role="${ifDefined(this.hasTooltip && !tooltipElement ? 'img' : undefined)}">
```
annabelsegalld2l, line 184: The axe tests were failing with the tooltip saying that you can't have aria-labelledby without a role... I've added img here since that seemed to match the most. I also tried status, but that adds a live region and other behaviour we don't want here. Img doesn't add anything to screenreaders other than "graphic" in front of the description.

dlockhart: Cool, I was actually wondering about that and how this "thing" would appear semantically. Although not perfect, image seems reasonable.

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this.addEventListener('blur', this._onBlur);
+		this.addEventListener('focus', this._onFocus);
+	}
+
+	render() {
+		const iconStyle = {
```
annabelsegalld2l, line 85: This line can be removed. I was playing around to ensure that in the cases with hide-zero, that the icon shows up and the number does not, but the aria-labels are still accessible.

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,196 @@
...
+	outline: none;
+}
+
+:host([size="large"]) .d2l-count-badge-wrapper {
+	border-radius: 0.8rem;
+	outline: none;
+}
+`;
+
+export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(RtlMixin(superclass)) {
```
annabelsegalld2l, line 63: All the logic that was previously in the count-badge has been extracted to this Mixin, so we have less duplicated code. There are a few changes that I'll point out below in the render function.

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,196 @@
...
+			numberString = `${'9'.repeat(this.maxDigits)}`;
+			numberString = this.localize('components.count-badge.plus', { number: numberString });
+		}
+		return html`
+			<div 
+			class="d2l-count-badge-wrapper"
+			id="${ifDefined(tooltipElement ? undefined : this._badgeId)}"
+			tabindex="${ifDefined((this.tabStop || this.hasTooltip) && (!hideNumber || tooltipElement) ? '0' : undefined)}" 
+			aria-labelledby="${ifDefined(this.hasTooltip ? undefined : this._textId)}"
+			role="${ifDefined((!tooltipElement || !this.hasTooltip) && !this.announceChanges ? 'img' : undefined)}">
```
annabelsegalld2l, line 184: Adding the img role makes the badge accessible when using the tooltip. It just adds "graphic" to the beginning of the description.

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,196 @@
...
+				<div class="d2l-count-badge-number"
+				style=${styleMap(numberVisibility)}>
+						<div aria-hidden="true">${numberString}</div>		
+				</div>
+				${tooltipElement}
```
annabelsegalld2l, line 189: The icon is passed here as the tooltip element. If tooltipElement is specified, the tooltip will be attached there. The forceTooltipOn parameter controls the force-show property of the tooltip. This is needed for when focus is on the badge wrapper, but we want the tooltip on the icon to show up as well.

components/count-badge/test/count-badge-icon.axe.js
```diff
@@ -0,0 +1,24 @@
+import '../count-badge-icon.js';
+import { expect, fixture, html } from '@open-wc/testing';
+
+const countBadgeTooltip = html`
```
annabelsegalld2l, line 4: For each of the badges, I've added cases for the tooltip, announce-changes, and a basic badge since they have slightly different accessibility labels.

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,108 @@
...
+		:host(.focus-visible) d2l-icon,
+		.d2l-count-badge-wrapper.focus-visible d2l-icon {
+			color: var(--d2l-color-celestine);
+		}
+
+		.d2l-count-badge-number {
+			position: relative;
+		}
+
+		.d2l-count-badge-wrapper {
```
dlockhart, line 39: This and the other rules targeting elements/classes defined by the mixin feels like a little bit of a leaky abstraction... where the consumer of the mixin (count badge icon) knows things about the makup of the mixin itself.

There are a couple ways I can think of to get around this:
1. Expose these as CSS variables on the mixin (e.g. `--d2l-count-badge-wrapper-padding`) and then the consumer here would set that `:host { --d2l-count-badge-wrapper-padding: 0 14px 0 2px; }`.
2. Pass these things in as parameters to the `renderCount()` method.

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,108 @@
...
+		this.addEventListener('focus', this._onFocus);
+	}
+
+	render() {
+		const iconStyle = {
+			visibility: 'visible'
+		};
+		const icon = html`
+			<d2l-icon 
+				id="${this._badgeId}"
```
dlockhart, line 90: Again, a bit of leaking here where the mixin is defining this private variable `_badgeId` and the icon is using it... but also the count uses when there's a tooltip which worries me that it might be getting set on two elements.

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,108 @@
...
+				icon="${this.icon}" 
+				class="d2l-button-icon"
```
dlockhart, line 92: This class shouldn't be needed?

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,108 @@
...
+				icon="${this.icon}" 
+				class="d2l-button-icon" 
+				style=${styleMap(iconStyle)}
+				role="${ifDefined(this.hasTooltip ? 'img' : undefined)}">
+			</d2l-icon>`;
+		return this.renderCount(icon, this._forceTooltipOn);
+	}
+
+	_onBlur() {
+		this._forceTooltipOn = false;
```
dlockhart, line 100: I thought tooltip had some built-in logic that automatically shows it when targets receive focus? Maybe just a matter of making sure the target of the tooltip is the one firing the focus/blur events?

components/count-badge/count-badge-icon.js
dlockhart: Looks like the mixin already mixes in `RtlMixin` so I don't think you need it here (or in the non-icon one).

components/count-badge/demo/count-badge-icon.html
```diff
@@ -0,0 +1,50 @@
...
+		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+		<script type="module">
+			import '../../demo/demo-page.js';
+			import '../count-badge-icon.js';
+			import '../../button/button.js';
+        </script>
+	</head>
+	<body unresolved>
+		<d2l-demo-page page-title="d2l-count-badge">
```
dlockhart, line 16:
```suggestion
		<d2l-demo-page page-title="d2l-count-badge-icon">
```

components/count-badge/README.md
```diff
@@ -40,6 +40,47 @@ To make your `d2l-count-badge` accessible, use the following properties when app
...
+| `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
+| `hide-zero`, default: `false` | Boolean | Optionally choose not to show the count badge when the number is zero. |
+| `text`, required | String | Descriptive text for the badge which will act as an accessible label and tooltip text when tooltips are enabled. |
+| `tab-stop`, default: `false` | Boolean | Optionally choose to make the badge a tab stop. |
+| `announce-changes`, default: `false` | Boolean | Optionally choose to announce changes to the badge with an aria-live region. If the text property is changed, the text will be read by screen-readers. |
+| `has-tooltip`, default: `false` | Boolean | Optionally choose to have a tooltip below the badge. |
+
+### Accessibility Properties
+
+To make your `d2l-count-badge` accessible, use the following properties when applicable:
```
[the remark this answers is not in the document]
margaree: If they are identical then you could just have an `## Accessibility` section at the top and put the accessibility properties table in there. You can also leave it as is for now and George or I can tweak when prepping the document for the site.

components/count-badge/count-badge-icon.js
annabelsegalld2l: It shouldn't get set on two elements as the mixin checks if there is a tooltipElement passed along (the icon). If it has the tooltipElement, it won't set an id there.

components/count-badge/count-badge-icon.js
annabelsegalld2l: The problem with this is we need to focus on the wrapper, not the icon that we want the tooltip to be on. So we need to listen to the wrapper's blurring event to turn on the tooltip. I did explore making the icon focusable so it turns the tooltip on, but then the offscreen label won't be read as it's in the wrapper. We could make the icon labelledby the offscreen span, but that would be another leak.

components/count-badge/count-badge-icon.js
```diff
@@ -0,0 +1,101 @@
...
+			 * REQUIRED: Preset icon key (e.g. "tier1:gear")
+			 * @type {string}
+			 */
+			icon: {
+				type: String,
+				reflect: true
+			},
+			_forceTooltipOn: {
+				type: Boolean
+			},
```
dlockhart, line 23: I think this can go away now?

components/count-badge/count-badge-icon.js
dlockhart: Re-upping this ^ comment about `RtlMixin` not being necessary since `CountBadgeMixin` includes it.

components/count-badge/demo/count-badge-icon.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		<meta charset="UTF-8">
+		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
```
dlockhart, line 8: Don't need the `webcomponents-loader.js` or `<meta http-equiv="X-UA-Compatible" content="ie=edge">` lines anymore now that IE and legacy-Edge are gone.

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,185 @@
...
+			this.maxDigits = 2;
+		}
+	}
+
+	getOffscreenId() {
+		return this.hasTooltip ? undefined : this._textId;
+	}
+
+	renderCount(numberStyles) {
+		let numberString = `${this.number}`;
```
dlockhart, line 157: Wondering if you want to put the number through [formatNumber](https://github.com/BrightspaceUI/intl#number-formatting)? That way "1000000" would render as "1,000,000" (in most English locales anyway).

dlockhart (commented):
This is looking really good!

components/count-badge/count-badge-mixin.js
annabelsegalld2l: Yep, I had this as a separate story to investigate. Great that there's already a helper to do it. 🎊

components/count-badge/count-badge-mixin.js
```diff
@@ -178,8 +182,8 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(R
 	renderTooltips(badgeId) {
 		return html`
 			${this.hasTooltip  ?
-		html`<d2l-tooltip aria-live="${this.announceChanges ? 'polite' : 'off'}" for="${badgeId}" for-type="label">${this.text}</d2l-tooltip>`
-		: html`<span id="${this._textId}" aria-live="${this.announceChanges ? 'polite' : 'off'}" class="d2l-offscreen">"${this.text}"</span>`}
+		html`<d2l-tooltip aria-atomic="true" aria-live="${this.announceChanges ? 'polite' : 'off'}" for="${badgeId}" for-type="label">${this.text}</d2l-tooltip>`
```
annabelsegalld2l, line 185: Fun browser tidbit, chrome defaults to aria-atomic=true, while firefox does not. Firefox was saying some really funky things when announcing the changes (like only saying the last digit that changed?). Aria-atomic makes sure the WHOLE message is repeated when one portion changes.

components/count-badge/count-badge-mixin.js
```diff
@@ -165,7 +166,10 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(R
 		};
 		if (this.maxDigits && this.number.toString().length > this.maxDigits) {
 			numberString = `${'9'.repeat(this.maxDigits)}`;
+			numberString = formatNumber(numberString);
```
dlockhart, line 169: `formatNumber` is expecting a JavaScript `Number` here... so maybe put it through `parseInt` first just to be safe?

components/count-badge/count-badge-mixin.js
```diff
@@ -165,7 +166,10 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(R
...
 			numberString = this.localize('components.count-badge.plus', { number: numberString });
+		} else if (!hideNumber) {
+			numberString = formatNumber(numberString);
```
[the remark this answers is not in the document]
dlockhart: Yay shared libraries!

components/count-badge/count-badge-mixin.js
dlockhart: TIL!

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,189 @@
...
+		if (hideNumber) {
+			numberString = '';
+		}
+		numberStyles = {
+			...numberStyles,
+			visibility: hideNumber ? 'hidden' : 'visible'
+		};
+		if (this.maxDigits && this.number.toString().length > this.maxDigits) {
+			numberString = `${'9'.repeat(this.maxDigits)}`;
+			numberString = formatNumber(parseNumber(numberString));
```
dlockhart, line 169: It's confusing/weird, but `parseNumber` is expecting the number in the user's locale, whereas you're creating a number by just repeating `9` a bunch of times, so you know that it's just a "normal" number -- the built-in native `parseInt()` API is meant for that.
```suggestion
			numberString = formatNumber(parseInt(numberString));
```

components/count-badge/count-badge-mixin.js
```diff
@@ -0,0 +1,189 @@
...
+			numberString = this.localize('components.count-badge.plus', { number: numberString });
+		} else if (!hideNumber) {
+			numberString = formatNumber(parseNumber(numberString));
```
dlockhart, line 172: And then in this case the number is coming from the property directly, so it's already `Number` and shouldn't need parsing.
```suggestion
			numberString = formatNumber(numberString);
```

components/count-badge/count-badge-mixin.js
annabelsegalld2l: Ah I see, will make that change. Too many different ways to write numbers... 🤦

dlockhart (approved):

PR #1728, opened 2021-09-22
Daylight: Templates

GeorgePlukov:


templates/primary-secondary/README.md
```diff
@@ -30,32 +45,34 @@ Mental Models and Page Structure:
 
 If no nodes are assigned to the `footer` slot, the footer is hidden.
 
-**Note:** this template automatically includes `<header>`, `<main>`, `<aside>` and `<footer>` elements, so there's no need to include them inside the various slots.
+Note: this template automatically includes `<header>`, `<main>`, `<aside>` and `<footer>` elements, so there's no need to include them inside the various slots.
```
margaree, line 48: It might make sense to move this and `If no nodes are assigned to the `footer` slot, the footer is hidden.` above it to above the demo since otherwise they end up pushed far below the demo tables

margaree (approved):

GeorgePlukov:
I left the readme within the primary/secondary folder, I imagine if we ever add another template we would either move the readme up a folder and combine all of the templates in one page, or split each template into its own page (and move the top level template content out of this).

PR #1731, opened 2021-09-23
Daylight: lists

margaree:
I did some reorganizing of this document to call out selection and drag & drop lists more obviously near the top. I also copied the accessibility section from design.d2l.


components/list/list-item-checkbox-mixin.js
```diff
@@ -7,28 +7,36 @@ import { nothing } from 'lit-html';
 import { SelectionInfo } from '../selection/selection-mixin.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 
+/**
+ * @property label - The hidden label for the checkbox if selectable
```
margaree, line 11: This overrides the definition in `LabelledMixin`

components/list/list-item-checkbox-mixin.js
```diff
@@ -7,28 +7,36 @@ import { nothing } from 'lit-html';
...
+ */
 export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(LabelledMixin(superclass)) {
 
 	static get properties() {
 		return {
 			/**
-			 * Disables the input
+			 * **Selection:** Disables the input
```
margaree, line 18: I added these for selection and drag & drop so that it's easier to identify properties for a specific state in the giant properties table. I also was thinking we could sort by that (separate PR in documentation to come for that)

components/list/README.md
```diff
@@ -1,46 +1,88 @@
 # Lists
 
-## d2l-list
-
-The `d2l-list` is the container to create a styled list of items using `d2l-list-item` or `d2l-list-item-button`. It provides the appropriate `list` semantics as well as options for displaying separators, etc.
-
-![List](./screenshots/list.png?raw=true)
+A list displays a collection of objects of the same type. A list is primarily used in order to help users navigate to a full-page representation of a single object, or to select several items and execute an action on them.
 
+<!-- docs: demo autoSize:false size:large -->
```
margaree, line 5: I tried to use autoSize but with the lists the sizing comes a bit too late sometimes for the resizing logic in the demo snippet to work correctly. I tried adding timeouts and that did help but I kept having to increase it to capture all cases and so not using autoSize here just seemed safer

components/list/README.md
```diff
@@ -56,100 +98,285 @@ The `grid` attribute will enable a table-like keyboard grid that allows a user t
...
+    - Image: max dimensions: `width: 120px` and `height: 71px` and has `20px margin` from the main content;
+    - default break: `581px < x < 636px` where `x` is the width of the component.
+  - Breakpoint 2
+    - Image: max dimensions: `width: 180px` and `height: 102px` and has `20px margin` from the main content;
+    - default break: `637px < x < 842px`  where `x` is the width of the component.
+  - Breakpoint 3
+    - Image: max dimensions: `width: 216px` and `height: 120px` and has `20px margin` from the main content;
+    - default break: `843px < x`  where `x` is the width of the component.
 
 ## ListItemMixin
```
margaree, line 377: I tried to get this section to be a working demo example but couldn't quite get it to work so leaving mostly as is for now.

components/list/list-item-mixin.js
```diff
@@ -30,10 +30,12 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 		return {
 			/**
 			 * Breakpoints for responsiveness in pixels. There are four different breakpoints and only the four largest breakpoints will be used.
+			 * @type {array}
```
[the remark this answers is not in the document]
margaree: Yup I think that's ok - we'll see what Jeff thinks and can modify if needed

components/list/README.md
```diff
@@ -56,100 +98,285 @@ The `grid` attribute will enable a table-like keyboard grid that allows a user t
...
+
+<!-- docs: demo live name:d2l-list autoSize:false size:medium -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/list/list.js';
+  import '@brightspace-ui/core/components/list/list-item.js';
+  import '@brightspace-ui/core/components/list/list-item-content.js';
+</script>
+
+<d2l-list style="width: 100%">
```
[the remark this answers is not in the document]
margaree: Yeah I'm torn on it - adding in the style just for changing the width adds a lot of bulk to the README and also as you said this allows easier copy/pasting. I believe setting to `display: block` on the demo does give full width, but I think it didn't vertically centre so there was that trade off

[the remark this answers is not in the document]
GeorgePlukov: Yeah thats a good point, maybe for now we say use inline style for single properties but once it goes multi css properties switch to style blocks?

GeorgePlukov (approved):

PR #1736, opened 2021-09-23
DE45328 - Delay hierarchical-view-resize event to prevent ResizeObserver errors

svanherk:
In the filter, the combination of the dropdown header and hierarchical-view content rendering at the same time means Safari sometimes throws the `ResizeObserver loop completed with undelivered notifications` error.  This didn't occur when the filter header was smaller, but adding the clear button means it might trigger a resize.

I tested in `filter`, `menu`, `dropdown` and `hierarchical-view` and don't see any behaviour change with delaying this event a frame.

### DIFF
--- a/components/hierarchical-view/hierarchical-view-mixin.js
+++ b/components/hierarchical-view/hierarchical-view-mixin.js
@@ -333,8 +333,11 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 			composed: true,
 			detail: contentRect
 		};
-		/** @ignore */
-		this.dispatchEvent(new CustomEvent('d2l-hierarchical-view-resize', eventDetails));
+		// To avoid ResizeObserver undelivered notifications error, wait a frame to send the event
+		requestAnimationFrame(() => {
+			/** @ignore */
+			this.dispatchEvent(new CustomEvent('d2l-hierarchical-view-resize', eventDetails));
+		});
 	}
 
 	__focusCapture(e) {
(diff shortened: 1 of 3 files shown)


components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -521,7 +524,10 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 	}
 
 	__onViewResize(e) {
-		this.style.height = `${e.detail.height}px`;
```
svanherk, line 524: The event delay above is the actual fix for the `ResizeObserver` error, but when investigating I also found this was sometimes being set to the same value multiple times

dlockhart (approved):
Thanks for digging into this!

components/menu/test/menu-rtl.visual-diff.js
```diff
@@ -38,7 +38,7 @@ describe('d2l-menu rtl', () => {
 			// this scenario also tests height change when going from 1 menu item to 2 within nested menu
 			await page.$eval('#nested-item', item => {
 				return new Promise((resolve) => {
-					item.addEventListener('d2l-hierarchical-view-show-complete', resolve, { once: true });
+					item.addEventListener('d2l-hierarchical-view-show-complete', () => requestAnimationFrame(resolve), { once: true });
```
svanherk, line 41: Now that the `resize` event is a frame later, we need to wait a frame for the visual-diff tests.  We could also listen for the `d2l-hierarchical-view-resize` event here instead, but I like testing that the resize is only one frame after the `show` event.  This doesn't change the order of events - the `show` event always came before the `resize` one in the `reduced-motion` case.

PR #1739, opened 2021-09-23
Don't wipe out shadowroot styles in non-Chromium browsers

grant-cleary:
In non-Chromium browsers, setting the shadowroot's innerHTML this way wipes out its styles (thanks for looking into that, @dlockhart!). This means the host styles don't get applied and wrapping stops functioning. Instead of replacing the inner HTML, append to it instead to ensure we keep any styles that may be there.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -141,7 +141,7 @@ class HtmlBlock extends LitElement {
 
 		if (this._renderContainer) return;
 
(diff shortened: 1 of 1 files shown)


grant-cleary:
FYI for when you get back, @dbatiste. :)

dlockhart (approved):
Thanks!

PR #1740, opened 2021-09-23
removing web component polyfill from demos

dlockhart:
Now that we're not testing legacy-Edge in SauceLabs, I think it's a good time to stop including the web component polyfill for it in our demos.

I've also removed the `<meta http-equiv="X-UA-Compatible" content="ie=edge">` from our demos, which was a way to tell legacy-Edge to never emulate old versions of IE... which we also don't care about anymore.

Apologies for all the file changes -- they're the same change in every demo/visual-diff HTML file.


annabelsegalld2l (approved):
🎉

PR #1743, opened 2021-09-23
Add helper to detect if we're in an iframed FRA

grant-cleary:
We want a good, performant, and reliable way to indicate whether we're in an FRA, so we can add one in that tries to communicate with the host page to get this information.

### DIFF
--- a/helpers/framed.js
+++ b/helpers/framed.js
@@ -1,0 +1,26 @@
+let framedPromise;
+
+export async function isFramed() {
+	if (framedPromise !== undefined) return framedPromise;
+
+	if ((window.D2L && window.D2L.IsNotAnIframedApp) || window === window.parent) {
+		framedPromise = Promise.resolve(false);
+		return framedPromise;
+	}
+
+	const handleIsFramedResponse = (evt, resolve) => {
+		if (!evt || !evt.data || !evt.data.isFramed) return;
+		window.removeEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+		resolve(evt.data.isFramed);		
+	};
+
+	framedPromise = Promise.race([
+		new Promise(resolve => {
+			window.addEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+			window.parent.postMessage('isFramedRequest', '*');
+		}),
+		new Promise(resolve => setTimeout(() => resolve(false), 100))
+	]);
+
+	return framedPromise;
+}
--- a/helpers/framed.js
+++ b/helpers/framed.js
@@ -1,26 +1,32 @@
-let framedPromise;
+let framed;
 
 export async function isFramed() {
-	if (framedPromise !== undefined) return framedPromise;
+	if (framed !== undefined) return framed;
 
 	if ((window.D2L && window.D2L.IsNotAnIframedApp) || window === window.parent) {
-		framedPromise = Promise.resolve(false);
-		return framedPromise;
+		framed = false;
+		return framed;
 	}
 
 	const handleIsFramedResponse = (evt, resolve) => {
 		if (!evt || !evt.data || !evt.data.isFramed) return;
 		window.removeEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
-		resolve(evt.data.isFramed);
+		framed = evt.data.isFramed;
+		resolve();
 	};
 
-	framedPromise = Promise.race([
+	await Promise.race([
 		new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


helpers/framed.js
```diff
@@ -0,0 +1,26 @@
...
+		window.removeEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+		resolve(evt.data.isFramed);		
+	};
+
+	framedPromise = Promise.race([
+		new Promise(resolve => {
+			window.addEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+			window.parent.postMessage('isFramedRequest', '*');
+		}),
+		new Promise(resolve => setTimeout(() => resolve(false), 100))
```
grant-cleary, line 22: I did some local testing and the message took about 20-35ms to process, in general. 100ms is pretty conservative here, so I'm perfectly fine if we want it to cut it down.

grant-cleary:
Ignore this for now. Dave gave me a good idea of how to make this more resilient to future lookups, so I'm going to make a few changes here. :)

helpers/framed.js
```diff
@@ -0,0 +1,32 @@
...
+	};
+
+	await Promise.race([
+		new Promise(resolve => {
+			window.addEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+			window.parent.postMessage('isFramedRequest', '*');
+		}),
+		new Promise(resolve => {
+			setTimeout(() => {
+				framed = false;
```
grant-cleary, line 25: Ah, this presents a bit of an issue. Will fix.

helpers/framed.js
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	const handleIsFramedResponse = (evt, resolve) => {
+		if (!evt || !evt.data || !evt.data.isFramed) return;
+		window.removeEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
+		framed = evt.data.isFramed;
+		resolve(framed);
+	};
+
+	framed = await Promise.race([
```
grant-cleary, line 18: The gist of this is that we resolve to the first promise that finishes. Thus, if we hit the 75ms timeout waiting for a response from the host page, we set our internal state false. However, if that listener is ever hit, we explicitly set `framed = evt.data.isFramed` so that the internal state will eventually update and future calls will start returning this new value.

helpers/framed.js
```diff
@@ -0,0 +1,31 @@
+let framed;
+
+export async function isFramed() {
+	if (framed !== undefined) return framed;
+
+	if ((window.D2L && window.D2L.IsNotAnIframedApp) || window === window.parent) {
```
dlockhart, line 6: Double-check that accessing `window.parent` cross-origin doesn't throw an exception here. I've needed `try`-`catch` blocks around similar code in the past.

helpers/framed.js
```diff
@@ -0,0 +1,31 @@
...
+		framed = false;
+		return framed;
+	}
+
+	const handleIsFramedResponse = (evt, resolve) => {
+		if (!evt || !evt.data || !evt.data.isFramed) return;
```
dlockhart, line 12: Should the last check be `evt.data.isFramed === undefined`? I guess `isFramed` if it comes back at all is always `true` eh... yeah OK.

helpers/framed.js
```diff
@@ -0,0 +1,31 @@
...
+		window.removeEventListener('message', evt => handleIsFramedResponse(evt, resolve), false);
```
dlockhart, line 13: Double-check that this is actually getting removed. I've only ever seen this done by passing a single argument (e.g. `removeEventListener('eventName', callbackName, false)`) -- never an arrow function. I have a feeling this is creating a new arrow function that isn't going to match the one that was created when it was registered.

helpers/framed.js
grant-cleary: Yeah, this is a weird one. The host will never respond with anything else. I can certainly make the check more specific just for correctness.

helpers/framed.js
grant-cleary: Yeah, I think you're right. This will end up defining a new function.

helpers/framed.js
grant-cleary: This didn't throw when I tested it locally and the docs suggest it shouldn't anymore. For extra safety's sake, however, I can split this checkout and try-catch it anyways. Admittedly I didn't test across our full range of supported browsers so there for sure could be a case where this blows up.

dlockhart (approved):
Looks good!

PR #1748, opened 2021-09-24
Add langterm for "99+" on the count badge since not all languages use the + sign

annabelsegalld2l:
I was looking through portfolio which will use the count badge, and I noticed it had a langterm for "99+". I've added a similar langterm here, calling it just plus instead of "99+" since the badge can have an arbitrary amount of digits.


annabelsegalld2l:
I've got the input-time-range flake here as well. The visual diffs should pass once it's skipped with Stacey's PR.

dlockhart (approved):

Issue #1751, opened 2021-09-24
d2l-input-time-range attribute order matters with non-default interval


dlockhart:
When you don't set a start and/or end value (or have invalid values) in `<d2l-input-time-range>` and you have an interval set to something other than the default (30), the order of the attributes can lead to different results.

This is because when computing the initial start and end values, it uses the current time and rounds up to the nearest interval. But if the value attributes are specified _before_ the interval attribute in the markup, it'll use the default interval of 30 to do this calculate. Then later when the interval property is initialized, the value doesn't recalculate and it's too late to do so.

For example, this functions correctly, resulting in a start time of `15:20` (since 15:15 isn't on the interval):

```html
<d2l-input-time-range
  enforce-time-intervals
  time-interval="10"
  start-value="15:15"
></d2l-input-time-range>
```

However, if you flip the `time-interval` to be set last, it'll incorrectly result in a time of `15:30`, using the default interval of 30:
```html
<d2l-input-time-range
  enforce-time-intervals
  start-value="15:15"
  time-interval="10"
></d2l-input-time-range>
```

The same issue exists with `enforce-time-intervals` since it's also part of the determination.

margaree:
Closing as this appears to be fixed now. Tests were added for this case here https://github.com/BrightspaceUI/core/pull/2335

dlockhart:
> Closing as this appears to be fixed now.

Cool! My guess is Lit 2 quietly fixed this.

PR #1752, opened 2021-09-24
Cleanup legacy-Edge from Tests

dlockhart:
Just some follow up legacy-Edge cleanup (and a bit of IE11 cleanup) after @svanherk removed it from our unit testing.

I've only cleaned up things that would impact our testing code -- there are still a bunch of legacy-Edge hacks/references in the actual code which I think we want to leave in until we get the official "all clear".

### DIFF
--- a/components/inputs/test/input-time-range.test.js
+++ b/components/inputs/test/input-time-range.test.js
@@ -1,4 +1,6 @@
-import { aTimeout, expect, fixture, oneEvent } from '@open-wc/testing';
+import { aTimeout, expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { formatDateInISOTime, getDateFromISOTime } from '../../../helpers/dateTime.js';
+import { getDefaultTime } from '../input-time.js';
 import { getDocumentLocaleSettings } from '@brightspace-ui/intl/lib/common.js';
 import { getShiftedEndTime } from '../input-time-range.js';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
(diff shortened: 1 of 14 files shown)


components/inputs/test/input-time-range.test.js
```diff
@@ -163,33 +162,38 @@ describe('d2l-input-time-range', () => {
 				{ enforceTimeIntervals: false, validStart: false, validEnd: false }
 			].forEach((testCase) => {
 				it(`when enforceTimeIntervals = ${testCase.enforceTimeIntervals}, valid start = ${testCase.validStart} and valid end = ${testCase.validEnd}`, async() => {
+
+					const defaultStartValue = formatDateInISOTime(getDefaultTime(undefined, testCase.enforceTimeIntervals, 'ten'));
+					const defaultEndTime = getDateFromISOTime(defaultStartValue);
+					defaultEndTime.setMinutes(defaultEndTime.getMinutes() + 10);
+					const defaultEndValue = formatDateInISOTime(defaultEndTime);
```
dlockhart, line 169: So sometime between when these tests were skipped because they were timing out in legacy-Edge and now, we changed the behaviour of the code. It now uses the current time when coming up with the default value, not midnight.

components/inputs/test/input-time-range.test.js
```diff
@@ -163,33 +162,38 @@ describe('d2l-input-time-range', () => {
...
 						else if (testCase.validStart && testCase.enforceTimeIntervals) expectedEndTime = '12:30:00';
-						else if (!testCase.enforceTimeIntervals) expectedEndTime = '00:11:00';
-						else expectedEndTime = '00:20:00';
+						else expectedEndTime = defaultEndValue;
 					}
 
-					const caseFixture = `<d2l-input-time-range
+					// Issue #1751: the order of enforce-time-intervals and time-interval matters.
+					// If they're specified after start-value or end-value, their default values will be used
+					// for the calculation
```
dlockhart, line 189: See #1751 as to why I had to re-order the attributes here.

mixins/test/localize-mixin.test.js
```diff
@@ -212,7 +211,7 @@ describe('LocalizeMixin', () => {
 				setTimeout(() => {
 					expect(renderCount).to.equal(1);
 					done();
-				}, 500); // larger timeout for legacy-Edge
+				}, 300);
```
dlockhart, line 214: 300 was the original timeout before we increased it for legacy-Edge.

PR #1758, opened 2021-09-27
DE45295: Fixing date-picker positioning on legacy screens

annabelsegalld2l:
Legacy screens are not responsive, meaning setting bottom: 0 will put the date-picker way at the bottom of the page rather than at the bottom of the screen. I initially tried to remedy this by setting the bottom to be the difference between window and screen sizes, but that hasn't been working consistently as Mark found in his testing. 

I've got a new approach where the bottom is still set to zero (at the bottom of the page), but the calendar is scrolled into view and fully visible. Once the calendar is closed, the focus is shifted back to the text input for the date. I don't think moving to the bottom of the page should be an issue since there is no guarantee we can see the text input anyways, it could be blocked behind the tray. Note that this change will affect the non-responsive pages only, such as the quiz or discussion editing pages. The web component and MVC pages should continue to work as normal.

Previous behaviour, very inconsistent, needs scrolling:
legacy-date-picker-bug

New behaviour: Always down at the bottom: 
legacy-date-picker-bug-fixed

### DIFF
--- a/components/inputs/input-date.js
+++ b/components/inputs/input-date.js
@@ -404,10 +404,15 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 		if (this._dropdown && !this._dropdown.openedAbove && !fullCalendarVisible) {
 			this._dropdown.querySelector('d2l-calendar').scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
 		}
-		// use setTimeout to wait for keyboard to open on mobile devices
-		setTimeout(() => {
-			this._textInput.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
-		}, 150);
+
+		// if keyboard opens (in all cases except mobile calendar view),
+		// ensure text input is in viewport
+		if (!(mediaQueryList.matches && this._dropdown.opened)) {
+			// use setTimeout to wait for keyboard to open on mobile devices
+			setTimeout(() => {
+				this._textInput.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
+			}, 150);
+		}
 		/** @ignore */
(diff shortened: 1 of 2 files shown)


components/inputs/input-date.js
```diff
@@ -404,10 +404,15 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
...
 			this._dropdown.querySelector('d2l-calendar').scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
 		}
-		// use setTimeout to wait for keyboard to open on mobile devices
-		setTimeout(() => {
-			this._textInput.scrollIntoView({ block: 'nearest', behavior: 'smooth', inline: 'nearest' });
-		}, 150);
+
+		// if keyboard opens (in all cases except mobile calendar view),
+		// ensure text input is in viewport
+		if (!(mediaQueryList.matches && this._dropdown.opened)) {
```
margaree, line 410: Just to clarify since I've been out of this code for a bit - would this if statement ever be true? I think `_handleDropdownOpen` is only ever called after `opened` is set to true.

annabelsegalld2l: I made a recent change here that prevents opening the keyboard on mobile if the calendar is open. It's a bit confusing here so I've added that comment on top so we can remember the reasoning. So if we are in mobile AND the calendar is open, the keyboard will never open, so do not scroll the input into view. All other cases (non-mobile, mobile with keyboard), this will be true.

annabelsegalld2l: Ah, you are right in the mobile case that the keyboard and non-calendar opening flow would never get here... because the dropdown would never open 🤦. This code should still work, but it has the redundant this._dropdown.opened

margaree: Ah okay yup that was what I was wondering about - thanks for clarifying! Otherwise LGTM

margaree (approved):

PR #1759, opened 2021-09-28
Update dialog component to use new ifrau SlimClient when necessary

grant-cleary:
Now that ifrau can be imported directly, let's do that! Components can create `SlimClient`s and avoid clobbering ifrau settings set by the parent FRA client.


components/dialog/test/dialog-ifrau.visual-diff.html
```diff
@@ -1,58 +1,65 @@
...
+		};
+
+		window.addEventListener('message', handleIsFramedRequest, false);
+
+		const marshalDialogOpenEvent = evt => {
+			if (!verifyPostMessage(evt, 'ifrau-dialog-open')) return;
+			getContainer().dispatchEvent(new CustomEvent('ifrau-dialog-open'));
+		};
+
+		window.addEventListener('message', marshalDialogOpenEvent, false);
```
grant-cleary, line 37: I have no idea if marshaling the events this way is actually the easiest/best option.

components/dialog/test/dialog-ifrau.visual-diff.js
```diff
@@ -15,19 +15,59 @@ describe('d2l-dialog-ifrau', () => {
...
+				});
+			}),
+			closeEvent
+		]);
+
+		return page.$eval(selector, dialogContainer => {
+			const iframe = dialogContainer.querySelector('iframe');
+			iframe.contentWindow.document.body.querySelector('#open').click();
+		});
+	};
```
grant-cleary, line 55: I'm not 100% sure these tests will pass. They worked locally, but I also don't exactly know what the results were supposed to look like and it's hard to compare locally vs. CI by just looking at static images. I'll let CI tell me whether I need to fix these further. :)

... Also, this repetition of the `open` and `reset` stuff is kind of gross, but the existing helpers don't quite cut it because there's no single selector that can peek into the iframe contents.

grant-cleary:
Grr, I forgot to run linting again.

helpers/ifrauBackdropService.js
```diff
@@ -1,10 +1,13 @@
+import { isFramed } from './framed.js';
+import { SlimClient } from 'ifrau/client/slim.js';
+
 let ifrauBackdropService;
 
 export async function tryGetIfrauBackdropService() {
-	if (!window.ifrauclient) return null;
+	if (!await isFramed()) return null;
```
grant-cleary, line 7: Since `isFramed()` has no dependence on the ifrau client, we no longer need the `window.ifrauclient` fallback check.

grant-cleary:
Hmm, something is weird with the visual diff tests here.

grant-cleary:
Oh that's interesting... Errors that `getDialog` is not defined... Yet it clearly is. This is weird.

grant-cleary:
> Oh that's interesting... Errors that `getDialog` is not defined... Yet it clearly is. This is weird.

Oh, I think this _might_ be a scoping problem since the inner page is within an iframe.

helpers/ifrauBackdropService.js
```diff
@@ -1,10 +1,13 @@
...
 	if (ifrauBackdropService) return ifrauBackdropService;
 
-	const ifrauClient = await window.ifrauclient().connect();
+	const ifrauClient = await new SlimClient().connect();
```
dlockhart, line 10: Would a dynamic import of the slim client work here? That'd avoid making the HTTP request until it's actually needed instead of always.

package.json
```diff
@@ -65,6 +65,7 @@
     "@open-wc/dedupe-mixin": "^1",
     "@webcomponents/shadycss": "^1",
     "focus-visible": "^5",
+    "ifrau": "^0.39.2",
```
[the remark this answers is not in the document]
grant-cleary: Ah, fair. I blindly let VSCode auto-fill it for me. 😄

helpers/ifrauBackdropService.js
grant-cleary: I imagine it certainly would, yes. I can change this.

components/dialog/test/dialog-ifrau.visual-diff.js
```diff
@@ -9,25 +9,62 @@ describe('d2l-dialog-ifrau', () => {
 	let browser, page;
 
 	before(async() => {
-		browser = await puppeteer.launch();
+		browser = await puppeteer.launch({ args: ['--no-sandbox', '--disable-setuid-sandbox'], });
```
grant-cleary, line 12: For heaven's sake. I need this to run locally but sometimes I forget to clear it out of my commits... Sigh.

grant-cleary:
Visual diff tests appear to be hanging on installing dependencies. I'll dig in when the run times out I suppose.

grant-cleary:
What in the what...
```
Need to get 372 MB of archives.
After this operation, 1105 MB of additional disk space will be used.
```
That seems kind of crazy to me. Do these core visual diff tests actually need to download ~1GB of stuff to run? O.o

svanherk:
Hmm, seems like this has to do with downloading the Korean font... sometimes the apt-packages stuff all updates and we download everything, but it should be a one-time thing per machine.  Did you cancel, or did it time out?

grant-cleary:
> Hmm, seems like this has to do with downloading the Korean font... sometimes the apt-packages stuff all updates and we download everything, but it should be a one-time thing per machine. Did you cancel, or did it time out?

I didn't cancel it. I was watching it run until it timed out. I'll give it a re-run and see if it times out again I suppose.

dlockhart:
Cool yeah I didn't do anything... that Korean font thing we're considering ditching because it's SO slow.

svanherk:
> Cool yeah I didn't do anything... that Korean font thing we're considering ditching because it's SO slow.

Yeah, seeming like now is the time...

grant-cleary:
> Yeah, seeming like now is the time...

Yeah, looks like my re-run is timing out again. I think maybe it's not writing anything permanently to disk until it's done the entire update and it times out before that can happen.

svanherk:
I'm gonna put up a PR to remove that step and the tests that need it

svanherk:
Ok merged, should be good to merge in `main` and see if that helps!

grant-cleary:
> Ok merged, should be good to merge in `main` and see if that helps!

Well, the visual diff tests are running and it didn't take 15+ minutes to install dependencies, so... I would say yes! Haha, thanks @svanherk! :)

grant-cleary:
Ok, so it looks like the visual diff results are actually pretty good. All of the goldens are different, but it looks like those are just small offsets because of the div container. I can try to fix these, but I'm not 100% sure it's worth the effort. I think I'm going to try to add a bit of padding to the iframe at least to get rid of some of the cases where the dialog is butting up against the edge of the viewport.

grant-cleary:
Ok, I can confirm what I thought before. The HTML body has a margin built in. Now that we have an iframe in the mix, we get two bodies, and thus two margins! So I think what I'm going to do here is just apply the test styles the way the old one did, both inside and outside the iframe. Then I'll adjust the viewport/iframe size as necessary and we can take the new goldens, assuming they still look correct. The alternative would be to zero out the margins, but that seems like a weird work-around for something that (imo) isn't actually a problem.

grant-cleary:
I'm not 100% sure I understand the SauceLabs test failures here. The test failures look unrelated, but they've now been consistent across multiple test runs. I'll dig in locally tomorrow I think.

svanherk:
> I'm not 100% sure I understand the SauceLabs test failures here. The test failures look unrelated, but they've now been consistent across multiple test runs. I'll dig in locally tomorrow I think.

I was getting the same thing - I couldn't reproduce the Firefox error locally but I added something I'm hoping will help and bumped the Playwright timeout in https://github.com/BrightspaceUI/core/pull/1764.  You can try merging `main` in again!

components/dialog/test/dialog-ifrau.visual-diff.html
```diff
@@ -2,57 +2,68 @@
...
-		<div>Line 1</div>
-		<div>Line 2</div>
-		<div>Line 3</div>
-		<div>Line 4</div>
-		<d2l-button slot="footer" primary>Yes</d2l-button>
-		<d2l-button slot="footer" id="cancel">No</d2l-button>
-	</d2l-dialog>
-
-	<script>
+	<div id="ifrau-dialog-container" style="width: 650px"></div>
```
grant-cleary, line 13: Technically this isn't necessary. We can give the iframe an ID via ifrau and emit events from that element. We can also just set the iframe's parent as the HTML body. I think I'm going to do that.

grant-cleary: Oh... Except without that the page width gets a bit wonky and things end up looking kind of gross. I'll still set the ID on the iframe because it makes some lookups easier in the visual diff tests themselves, but I think I'll keep the parent div.

grant-cleary: Eh, except then we would need to switch over the iframe to emit all of the events anyways, or the test would need to know both about the div and the iframe. Maybe this isn't worth doing.

helpers/ifrauBackdropService.js
```diff
@@ -1,10 +1,13 @@
...
 
 export async function tryGetIfrauBackdropService() {
-	if (!window.ifrauclient) return null;
+	if (!await isFramed()) return null;
 	if (ifrauBackdropService) return ifrauBackdropService;
 
-	const ifrauClient = await window.ifrauclient().connect();
+	const ifrau = await import('ifrau/client/slim.js');
+	const ifrauClient = await new ifrau.SlimClient().connect();
 	ifrauBackdropService = await ifrauClient.getService('dialogWC', '0.1');
```
dbatiste, line 11: It seems a little strange that this backdrop specific helper is defined outside of dialog, even though the iframe service itself is dialog specific.  As written, being dialog specific, I would expect this to live with the dialog component.  If it were changed to be a more generic backdrop service, perhaps it could live with the backdrop component.  🤷‍♂️

grant-cleary: Yeah, I recalled it living in the dialog component at some point and was surprised to find that it wasn't when I went in to make this change. I didn't look into when/why it moved.

grant-cleary: Cool, killing two birds with one stone here then, I guess. :)

grant-cleary:
Ok, new goldens are merged into this PR and everything is passing. Yay!

components/dialog/test/dialog-ifrau-contents.visual-diff.html
```diff
@@ -0,0 +1,42 @@
...
+		window.D2L = { DialogMixin: { preferNative: preferNative } };
+	</script>
+	<script type="module">
+		import '../../typography/typography.js';
+		import '../../button/button.js';
+		import '../dialog.js';
+	</script>
+	<title>d2l-dialog-ifrau</title>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<meta http-equiv="X-UA-Compatible" content="ie=edge">
```
dlockhart, line 16: This line can go away now that legacy-Edge isn't something we care about. I removed it from all existing files earlier, but you didn't have that change when you copied this file. :)

dlockhart (approved):
This LGTM

components/dialog/test/dialog-ifrau-contents.visual-diff.html
grant-cleary: Oh, nice! Yeah, I forgot to merge main before I started working on the visual diff tests so there were a few other things I also needed to nuke. I'll remove this and compare against the existing file in main to make sure I caught all of it.

dbatiste (approved):

PR #1761, opened 2021-09-28
LabelledMixin validation of missing labels

dlockhart:
If the `label` attribute is missing on one of our input elements, or if `labelled-by` is pointing to an element that doesn't exist or has an empty label, it's a P1 accessibility violation. We're pretty inconsistent with how we notify developers of missing labels -- sometimes we do nothing, occasionally we `console.warn` in `firstUpdated`.

This change attempts to centralize the reporting of that error from `LabelledMixin` in a way that doesn't initially blow up for all the existing implementations that might be unknowingly missing labels. It will throw the error but not block rendering, so that we can go through the error logs and find any problem areas.

If this approach looks good, I'll do some follow-up PRs to actually adopt the `LabelledMixin` in more (or all) of our input components. Currently only lists use it.

### DIFF
--- a/components/list/list-item-checkbox-mixin.js
+++ b/components/list/list-item-checkbox-mixin.js
@@ -81,7 +81,7 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		super.connectedCallback();
 		if (this.selectable) {
 			if (!this.key) console.warn('ListItemCheckboxMixin requires a key.');
-			if (!this.label || this.label.length === 0) console.warn('ListItemCheckboxMixin requires a label.');
+			if (!this.label && !this.labelledBy) console.warn('ListItemCheckboxMixin requires a label.');
 		}
 		if (!this.key) this.setSelected(undefined, true);
 	}
--- a/components/list/list-item-checkbox-mixin.js
+++ b/components/list/list-item-checkbox-mixin.js
@@ -81,7 +81,8 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		super.connectedCallback();
 		if (this.selectable) {
 			if (!this.key) console.warn('ListItemCheckboxMixin requires a key.');
-			if (!this.label && !this.labelledBy) console.warn('ListItemCheckboxMixin requires a label.');
+		} else {
+			this._labelRequired = false;
 		}
 		if (!this.key) this.setSelected(undefined, true);
 	}
--- a/components/selection/selection-input.js
+++ b/components/selection/selection-input.js
@@ -80,7 +80,6 @@ class Input extends SkeletonMixin(LabelledMixin(LitElement)) {
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 		if (!this.key || this.key.length === 0) console.warn('d2l-selection-input component requires a key.');
-		if (!this.label || this.label.length === 0) console.warn('d2l-selection-input component requires label text.');
 	}
 
 	render() {
--- a/mixins/demo/labelled-mixin.html
+++ b/mixins/demo/labelled-mixin.html
@@ -21,6 +21,7 @@
 					`;
 				}
 				render() {
+					if (!this.validateLabel()) return;
 					return html`
 						<d2l-input-checkbox aria-label="${ifDefined(this.label)}"></d2l-input-checkbox>
 					`;
--- a/mixins/demo/labelled-mixin.html
+++ b/mixins/demo/labelled-mixin.html
@@ -21,7 +21,6 @@
 					`;
 				}
 				render() {
-					if (!this.validateLabel()) return;
 					return html`
 						<d2l-input-checkbox aria-label="${ifDefined(this.label)}"></d2l-input-checkbox>
 					`;
--- a/mixins/labelled-mixin.js
+++ b/mixins/labelled-mixin.js
@@ -94,6 +94,7 @@ export const LabelledMixin = superclass => class extends superclass {
 	constructor() {
 		super();
 		this._labelElem = null;
+		this._labelRequired = true;
 		this._missingLabelErrorHasBeenThrown = false;
 		this._throwNoLabelExceptionImmediately = false;
 	}
@@ -135,7 +136,7 @@ export const LabelledMixin = superclass => class extends superclass {
 	}
 
 	_throwError(err) {
-		if (this._missingLabelErrorHasBeenThrown) return;
+		if (!this._labelRequired || this._missingLabelErrorHasBeenThrown) return;
 		this._missingLabelErrorHasBeenThrown = true;
 		// we don't want to prevent rendering
 		if (!this._throwNoLabelExceptionImmediately) {
--- a/mixins/labelled-mixin.js
+++ b/mixins/labelled-mixin.js
@@ -1,14 +1,6 @@
 
 import { cssEscape } from '../helpers/dom.js';
 
-class MissingLabelError extends Error {
-	constructor(elementName) {
-		elementName = elementName.toLowerCase();
-		super(`Label missing on element: ${elementName}`);
-		this.elementName = elementName;
-	}
-}
-
 const getCommonAncestor = (elem1, elem2) => {
 
 	const labelledPath = new WeakMap();
@@ -84,10 +76,15 @@ export const LabelledMixin = superclass => class extends superclass {
 	constructor() {
 		super();
 		this._missingLabelErrorHasBeenThrown = false;
+		this._throwNoLabelExceptionImmediately = false;
 	}
 
 	updated(changedProperties) {
 		super.updated(changedProperties);
+
+		if (changedProperties.has('label')) {
+			this._validateLabel();
+		}
 
 		if (!changedProperties.has('labelledBy')) return;
 
@@ -114,14 +111,18 @@ export const LabelledMixin = superclass => class extends superclass {
 
 			if (labelElem) {
 				this.label = getLabel(labelElem);
+				// TODO: how to validate empty label?
 			} else {
-				console.warn(`LabelledMixin: element with labelled-by="${this.labelledBy}", but no such element exists.`);
 				this.label = undefined;
+				this._validateLabel();
 			}
 
 		});
 
-		if (!labelElem) return;
+		if (!labelElem) {
+			this._validateLabel();
+			return;
+		}
 		const ancestor = getCommonAncestor(this, labelElem);
 
 		/* assumption: the labelling element will not change from a native to a custom element
@@ -145,18 +146,24 @@ export const LabelledMixin = superclass => class extends superclass {
 
 	}
 
-	validateLabel(throwImmediately = false) {
+	_validateLabel() {
 
-		if (this.labelledBy) return true;
+		if (this._missingLabelErrorHasBeenThrown) return true;
 
 		const hasLabel = (typeof this.label === 'string') && this.label.length > 0;
-		if (hasLabel || this._missingLabelErrorHasBeenThrown) return true;
+		if (hasLabel) return true;
 
 		this._missingLabelErrorHasBeenThrown = true;
-		const err = new MissingLabelError(this.tagName);
+
+		let err = null;
+		if (this.labelledBy) {
+			err = new Error(`LabelledMixin: "${this.tagName.toLowerCase()}" is labelled-by="${this.labelledBy}", but no such element exists or its label is empty`);
+		} else {
+			err = new Error(`LabelledMixin: "${this.tagName.toLowerCase()}" is missing a required "label" attribute`);
+		}
 
 		// we don't want to prevent rendering
-		if (!throwImmediately) {
+		if (!this._throwNoLabelExceptionImmediately) {
 			setTimeout(() => { throw err; });
 		// just for testing so we can actually catch it
 		} else {
--- a/mixins/labelled-mixin.js
+++ b/mixins/labelled-mixin.js
@@ -96,7 +96,6 @@ export const LabelledMixin = superclass => class extends superclass {
 		this._labelElem = null;
 		this._labelRequired = true;
 		this._missingLabelErrorHasBeenThrown = false;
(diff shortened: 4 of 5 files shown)


mixins/demo/labelled-mixin.html
```diff
@@ -21,6 +21,7 @@
 					`;
 				}
 				render() {
+					if (!this.validateLabel()) return;
```
dlockhart, line 24: This will be the change that controls will need to opt into to get label validation. For new controls or ones where we're certain labels are always provided, they can `return` right away and not render if the label is missing. For existing ones though we'll want to just validate but continue rendering until we're sure we've caught them all.

mixins/test/label-mixin.test.js
```diff
@@ -86,38 +87,19 @@ describe('LabelledMixin', () => {
 
 	let elem;
 
-	beforeEach(async() => {
```
dlockhart, line 89: I re-organized these a bit so that every test doesn't reuse the same giant fixture (which I was about to make bigger). The tests themselves haven't changed.

grant-cleary (approved):
Seems good to me!

dlockhart:
Going to update this to also handle missing elements referenced by `labelled-by`, which is currently just `console.warn`-ing.

mixins/labelled-mixin.js
```diff
@@ -101,14 +111,18 @@ export const LabelledMixin = superclass => class extends superclass {
 
 			if (labelElem) {
 				this.label = getLabel(labelElem);
+				// TODO: how to validate empty label?
```
dlockhart, line 114: The validation in this case wasn't handled before, and still isn't. The case is an element `labelled-by` another element that _exists_, but has an empty label.

The reason we can't validate this case right now is because `LabelMixin` implementations typically call `updateLabel()` inside their own `updated()` callback, which means there's an initial render where the thing hasn't set the label yet, and then it sets it and everything gets re-rendered. So if we validated right here, we'd falsely fail.

The only way I see to get around this is for `LabelMixin` implementations to synchronously call `updateLabel` in the `setter` for whatever is affecting the label -- and we're well aware of the perils in doing that. Alternately, we could delay the validation somehow? @dbatiste interested in your thoughts here.

grant-cleary: This is interesting from the editor's point of view too, I guess. The label syncs down to TinyMCE on init and updates if the label is changed afterwards, but if we can't bank on an explicit `set`, then we would sync an empty title down to TinyMCE and never change that. 🤔

dlockhart: So in almost all scenarios -- even when `labelled-by` is used -- the label is there right away on first render and everything is peachy.

But in some weird scenarios (like one of the tests where it replaces the label element with a new one), the label isn't there on the first render. But even then, the `label` attribute _eventually_ gets set, which would trigger a re-render of the editor.

grant-cleary: Ah, okay, I misunderstood your first comment - we can't validate here because we can't differentiate between the case where the label is going to stay empty vs. it getting set a bit later in the lifecycle.

grant-cleary: Hence we can always bank on it being set later provided the consumer hasn't just gone and tossed a legitimately empty label in.

dlockhart: Exactly yeah. One weird/maybe fragile option would be to just "wait a bit" and come back and if the label is still missing then we error.

dbatiste: I am a little worried about the case where the label data is coming from an HM response which would be delayed.

dlockhart: Is anything using `labelled-by` though yet? A [quick code search](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22labelled-by%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5) doesn't appear to show anything.

I'd really like to encourage teams to not render the list items (or inputs, or whatever) until they have all the data they need, and forcing this from the start before anyone does it in the first place seems like the right approach.

dbatiste: I was poking around earlier and I don't think anything is technically using it at the moment, but there is [one spot that should be](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/foundation-components/components/activity/editor/collection/d2l-activity-editor-collection-add.js?r=dd8f23fe#140).  I am going to put up a PR to add the `labelled-by` attribute here - the mixins are all set up, but I think the wire-up was forgotten.

components/list/list-item-checkbox-mixin.js
```diff
@@ -81,7 +81,7 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		super.connectedCallback();
 		if (this.selectable) {
 			if (!this.key) console.warn('ListItemCheckboxMixin requires a key.');
-			if (!this.label || this.label.length === 0) console.warn('ListItemCheckboxMixin requires a label.');
+			if (!this.label && !this.labelledBy) console.warn('ListItemCheckboxMixin requires a label.');
```
dbatiste, line 84: I am guessing we will want to change this to throw an error?  I needed to update this to alternatively check for labelled-by.

dlockhart: I was hoping the mixin code I added would either catch this case (making this warning unnecessary) or be able to be modified slightly to catch it.

dbatiste: It's fine if you want to exclude this from these changes... I can make this change separately.  It's not specific to the labelled-mixin, but is related.

mixins/labelled-mixin.js
```diff
@@ -27,6 +27,24 @@ const getLabel = labelElem => {
 	else return labelElem.textContent;
 };
 
+const waitForElement = async(contextElement, selector, timeout) => {
```
dlockhart, line 30: I _think_ with some fiddling this function could be made smaller by making it recursive... but not that important.

mixins/labelled-mixin.js
```diff
@@ -73,16 +91,30 @@ export const LabelledMixin = superclass => class extends superclass {
...
 
 		if (this._labelObserver) this._labelObserver.disconnect();
 
 		if (!this.labelledBy) return;
 
+		// try to get the labelling element syncronously first
 		let labelElem = this.getRootNode().querySelector(`#${cssEscape(this.labelledBy)}`);
+		if (!labelElem) {
+			labelElem = await waitForElement(this.getRootNode(), `#${cssEscape(this.labelledBy)}`, 3000);
+		}
```
dlockhart, line 117: I think this can just be simplified now?

```suggestion
		let labelElem = await waitForElement(this.getRootNode(), `#${cssEscape(this.labelledBy)}`, 3000);
```

And then if `labelElem === null` at this point, I think we can just throw an exception? At this point it has waited 3s and not found it.

mixins/labelled-mixin.js
```diff
@@ -73,16 +91,30 @@ export const LabelledMixin = superclass => class extends superclass {
...
 		if (!changedProperties.has('labelledBy')) return;
 
 		if (this._labelObserver) this._labelObserver.disconnect();
 
 		if (!this.labelledBy) return;
 
+		// try to get the labelling element syncronously first
 		let labelElem = this.getRootNode().querySelector(`#${cssEscape(this.labelledBy)}`);
+		if (!labelElem) {
+			labelElem = await waitForElement(this.getRootNode(), `#${cssEscape(this.labelledBy)}`, 3000);
```
grant-cleary, line 116: Is there anything in here that can get clobbered if a later lifecycle event is processed? Lit doesn't seem to await its lifecycle events, so if we're waiting 3 seconds here it seems likely that later lifecycle events (incl. re-renders) might start running before this resolves.

grant-cleary: I have no idea if this is something that might cause a problem in the mixin - it was just my first thought when I saw the long wait. This bit me in the HTML editor a while ago. 😅

dlockhart: Yeah I think Dave B and I are mildly concerned about this too. I'm going to do some testing.

components/list/list-item-checkbox-mixin.js
dbatiste: This scenario is that the item is selectable, but they did not provide a label or a reference to a label.  So currently,   I think the `LabelledMixin` mixin would only catch the scenario where they assigned a `labelled-by` value but it is not valid.

mixins/labelled-mixin.js
dbatiste: Maybe the observer could get replaced but not disconnected.  🤔

mixins/labelled-mixin.js
dbatiste: I was hesitant to do that because by using `waitForElement` I think it will automatically delay by a microtask due to the Promise.  Maybe that's ok to always introduce that delay. I was just being nervous about timing/delays.

I don't think we need to throw a special exception though... since once it waits 3s it will validate after the observer is defined.

mixins/labelled-mixin.js
```diff
@@ -73,49 +91,90 @@ export const LabelledMixin = superclass => class extends superclass {
...
+		} else {
+			throw err;
+		}
+	}
 
-				if (mutation.addedNodes.length > 0) {
-					labelElem = this.getRootNode().querySelector(`#${cssEscape(this.labelledBy)}`);
-					return;
-				}
+	_updateLabelElem(labelElem) {
```
dlockhart, line 150: I've refactored things and it might be easier to see them by just checking out the branch as the diff makes it look crazy.

Essentially I'm now tracking `labelElem` as a private variable. I originally had it as a property, which was really clean  since changing `labelledBy` would update `_labelElem` which would update `label` but that chain for 3 `updated()` calls was gross.

I also simplified the observer such that we just always try to re-fetch the element, which captures it going away or being added.

mixins/test/label-mixin.test.js
```diff
@@ -143,27 +128,41 @@ describe('LabelledMixin', () => {
...
 			const labelledElem = elem.querySelector('[labelled-by="label1"]');
 			labelledElem.labelledBy = 'label3';
 			await nextFrame();
 			expect(labelledElem.shadowRoot.querySelector('input').getAttribute('aria-label')).to.equal('other element');
 		});
 
-		it('does not explode if invalid id reference provided', async() => {
-			const labelledElem = elem.querySelector('[labelled-by="nolabel"]');
-			expect(labelledElem.shadowRoot.querySelector('input').hasAttribute('aria-label')).to.equal(false);
+		it.skip('does not explode if invalid id reference provided', async() => {
```
dlockhart, line 138: The way the exceptions are thrown now (sometimes with that 3 second delay) make it impossible for you to wire Mocha up to actually catch them. So skipping for now so we can review the changes... but we may end up not being able to actually test the error throwing.

mixins/labelled-mixin.js
dbatiste: Sounds good.  I think the original observer logic was trying to optimize it's task based on whether there were removed/added nodes, but it probably wasn't much of an optimization.

dbatiste (approved):
Looks good!

components/list/list-item-checkbox-mixin.js
```diff
@@ -81,7 +81,8 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		super.connectedCallback();
 		if (this.selectable) {
 			if (!this.key) console.warn('ListItemCheckboxMixin requires a key.');
-			if (!this.label || this.label.length === 0) console.warn('ListItemCheckboxMixin requires a label.');
+		} else {
+			this._labelRequired = false;
```
dlockhart, line 85: Since `ListItemCheckboxMixin` is unfortunately _always_ mixed into all list items... we need a way to disable the label validation if selection isn't enabled. This logic may need to move into the `ListItemMixin` if it becomes more complex (e.g. selection OR drag & drop requires a label).

components/selection/selection-input.js
```diff
@@ -80,7 +80,6 @@ class Input extends SkeletonMixin(LabelledMixin(LitElement)) {
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 		if (!this.key || this.key.length === 0) console.warn('d2l-selection-input component requires a key.');
-		if (!this.label || this.label.length === 0) console.warn('d2l-selection-input component requires label text.');
```
dlockhart, line 83: Was able to remove this since now it'll throw in this scenario.

mixins/labelled-mixin.js
```diff
@@ -128,7 +194,13 @@ export const LabelledMixin = superclass => class extends superclass {
 			});
 		}
 
-		this.label = getLabel(labelElem);
+		this.label = getLabel(this._labelElem);
+		this.dispatchEvent(new CustomEvent(
```
dlockhart, line 191: Added this event to help with testing.

mixins/test/label-mixin.test.js
```diff
@@ -172,35 +162,31 @@ describe('LabelledMixin', () => {
...
 			labelElem.parentNode.replaceChild(newLabelElem, labelElem);
-			await nextFrame();
+			await oneEvent(labelledElem, 'd2l-labelled-mixin-label-elem-change');
 			expect(labelledElem.shadowRoot.querySelector('input').getAttribute('aria-label')).to.equal('new label value');
 		});
 
-		it('does not explode if invalid id reference provided', async() => {
-			const labelledElem = elem.querySelector('[labelled-by="invalidlabel"]');
-			expect(labelledElem.shadowRoot.querySelector('input').hasAttribute('aria-label')).to.equal(false);
-		});
```
dlockhart, line 188: I had to delete the tests that would cause exceptions as there's no way to catch them. :(

dlockhart (commented):
Ok, I think everything has been addressed. I added to ability to disable the exceptions, which ListItems need to do if they're not using selection.

dbatiste (approved):

PR #1762, opened 2021-09-28
Stop installing ko for visual-diff tests

svanherk:
Looks like installing `ko` was also giving us `ja`, `zh`, and `zh-tw`.  Removed those tests as well.


dlockhart:
The install step is down to `10s` from around `1m40s` in the last successful run. 🎉

dlockhart (approved):

PR #1764, opened 2021-09-28
US122201 - Add Clear All button to filter

svanherk:
Similar to the clear button, the filter handles all clearing of each dimension and will also clear any search inputs with values.  Regular change events are sent, and there is an added `allCleared` property in the event if teams want to listen to this for perf enhancements.


components/filter/filter.js
```diff
@@ -227,7 +227,18 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	}
 
 	_buildHeader(singleDimension) {
-		if (!this._activeDimensionKey && !singleDimension) return null;
+		if (!this._activeDimensionKey && !singleDimension) {
+			return html`
+				<div slot="header">
```
dlockhart, line 232: Is the `<div>` needed, or could the `slot="header"` go right on the subtle button?

dlockhart (approved):

components/filter/filter.js
svanherk: I had it so this style would get applied: https://github.com/BrightspaceUI/core/blob/main/components/filter/filter.js#L53-L55
But looks like if I remove the `div` there too, everything works properly

dlockhart: Ah ok cool. Yeah fewer elements is always better!

PR #1766, opened 2021-09-30
US131477 - Fix esc behaviour in filter

svanherk:
Thanks to @annabelsegalld2l for catching this!  Pressing `esc` in the header of a nested dimension would close the dropdown (because `dropdown` was handling it), whereas pressing it in the content/list would return to the dimension list (because `hierarchical-view` was handling it).  We want it to always return to the dimension list so I need to handle it manually in the header.  This also fixes a bug where I wasn't setting `this._activeDimensionKey` to `null` if the `esc` method was used instead of the back button, so the header wasn't updating correctly.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -24,6 +24,7 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
+const ESCAPE_KEY_CODE = 27;
 const SET_DIMENSION_ID_PREFIX = 'list-';
 
 /**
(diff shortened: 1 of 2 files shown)


annabelsegalld2l (approved):
Yay, glad it was a quick fix! Would it be worth it to add one test case where another key is clicked to ensure the dropdown doesn't close, or would it just be redundant since escape is the only special key anyways?

dlockhart (approved):

svanherk:
> Yay, glad it was a quick fix! Would it be worth it to add one test case where another key is clicked to ensure the dropdown doesn't close, or would it just be redundant since escape is the only special key anyways?

I think that'd be more useful if I could better simulate a real keyboard click - luckily that should be an easy bug to spot.  I did add a test though to check that we don't mess with the `esc` behaviour when there's no active dimension!

PR #1769, opened 2021-10-01
US131477 - Add left arrow functionality to filter

svanherk:
We also want the user to be able to press the left arrow key in a nested dimension and go back to the dimension list.  So just adjusting the previous behaviour to handle this for `esc`.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -24,6 +24,7 @@ import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
+const ARROWLEFT_KEY_CODE = 37;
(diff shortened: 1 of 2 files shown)


components/filter/filter.js
```diff
@@ -205,6 +206,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				@d2l-hierarchical-view-hide-start="${this._handleDimensionHideStart}"
 				@d2l-hierarchical-view-show-complete="${this._handleDimensionShowComplete}"
 				@d2l-hierarchical-view-show-start="${this._handleDimensionShowStart}"
+				@keyup="${this._handleDimensionHideKeyPress}"
```
svanherk, line 209: The `hierarchical-view` was handling this before for `esc`, but we need to add the listener here for the left arrow key.  The` hierarchical-view` will still handles `esc` and `_handleDimensionHideKeyPress` won't be called in that case.

dlockhart (approved):

PR #1770, opened 2021-10-04
Daylight: off-screen edits

geurts:
Made some minor edits to the intro text. Also removed the demo since I think it's actually clearer without it.


GeorgePlukov (approved):

GeorgePlukov:
After discussing this I think we landed on leaving the live demo but updating the text to be more descriptive about how to explore the hidden content with a screen reader

GeorgePlukov (approved):

PR #1771, opened 2021-10-04
Count-badge: Hidden count badges should still take up space

annabelsegalld2l:
The visibility of the numberString will be hidden, which means it will not be visible but still take up space. This wasn't working since I was setting numberString to empty, so it wasn't rendering at all and causing layout changes with the filter component.


components/count-badge/count-badge-mixin.js
```diff
@@ -157,9 +157,6 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(R
 	renderCount(numberStyles) {
 		let numberString = `${this.number}`;
 		const hideNumber = this.hideZero && this.number === 0;
-		if (hideNumber) {
-			numberString = '';
-		}
 		numberStyles = {
 			...numberStyles,
 			visibility: hideNumber ? 'hidden' : 'visible'
```
annabelsegalld2l, line 162: The number will still be invisible, but take up the correct height for the badge, so if the count goes to zero, the layouts are not affected.

dlockhart (approved):

svanherk (approved):

PR #1777, opened 2021-10-05
Use LabelledMixin in our input components

dlockhart:
`LabelledMixin` is a pretty sweet way to get consistent labelling & missing label exception handling across all our components. It also supports `labelled-by`, which is pretty sweet.

So this PR updates almost all of our inputs to use it: `d2l-input-text`, `d2l-input-textarea`, `d2l-input-date`, `d2l-input-time`, `d2l-input-date-time`, `d2l-input-number` and `d2l-input-percent`.

There are a few like the date/time range components and the checkbox/radio components where it would either be difficult (the ranges have multiple labels) or weird (checkboxes and radios use fieldsets), so I left those for now.

### DIFF
--- a/components/inputs/input-date-time.js
+++ b/components/inputs/input-date-time.js
@@ -221,6 +221,7 @@ class InputDateTime extends LabelledMixin(SkeletonMixin(FormElementMixin(Localiz
 				@mouseover="${this._handleInputTimeFocus}"
 				label="${this.localize('components.input-date-time.time')}"
 				label-hidden
+				.labelRequired="${false}"
 				max-height="430"
 				?required="${this.required}"
 				?skeleton="${this.skeleton}"
@@ -245,6 +246,7 @@ class InputDateTime extends LabelledMixin(SkeletonMixin(FormElementMixin(Localiz
 						id="${this._inputId}"
 						label="${this.localize('components.input-date-time.date')}"
 						label-hidden
+						.labelRequired="${false}"
 						max-value="${ifDefined(this._maxValueLocalized)}"
 						min-value="${ifDefined(this._minValueLocalized)}"
 						?opened="${dateOpened}"
--- a/components/inputs/input-date-time.js
+++ b/components/inputs/input-date-time.js
@@ -17,6 +17,7 @@ import { FormElementMixin } from '../form/form-element-mixin.js';
 import { getDefaultTime } from './input-time.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { LabelledMixin } from '../../mixins/labelled-mixin.js';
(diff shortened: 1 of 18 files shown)


components/inputs/input-date-time.js
```diff
@@ -191,10 +187,6 @@ class InputDateTime extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(R
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		if (!this.label) {
-			console.warn('d2l-input-date-time component requires label text');
-		}
```
dlockhart, line 196: We don't need these warnings anymore since the mixin handles this now!

components/inputs/input-date.js
```diff
@@ -276,7 +268,8 @@ class InputDate extends SkeletonMixin(FormElementMixin(LocalizeCoreElement(LitEl
 					hide-invalid-icon
 					id="${this._inputId}"
 					label="${ifDefined(this.label)}"
-					?label-hidden="${this.labelHidden}"
+					?label-hidden="${this.labelHidden || this.labelledBy}"
+					.labelRequired="${false}"
```
dlockhart, line 272: When a component passes the label down to another component that also uses `LabelledMixin`, we need to turn off the label validation for the nested component. Otherwise when there's an error, a missing label exception will get thrown from both the top-level and lower-level component.

components/inputs/input-text.js
```diff
@@ -312,6 +308,13 @@ class InputText extends FormElementMixin(SkeletonMixin(RtlMixin(LitElement))) {
 		return super.validity;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		if (this.ariaLabel) {
+			this.labelRequired = false;
```
dlockhart, line 314: Maintaining support for the legacy `aria-label` attribute on this component means when it's set we need to _not_ throw a missing label exception.

mixins/labelled-mixin.js
```diff
@@ -93,8 +93,8 @@ export const LabelledMixin = superclass => class extends superclass {
 
 	constructor() {
 		super();
+		this.labelRequired = true;
 		this._labelElem = null;
-		this._labelRequired = true;
```
dlockhart, line 97: I made this property slightly more public so that components can set it... but it's still not a Lit property so I don't think we need to advertise it.

dlockhart:
Firefox tests are pretty angry... looking into them. (and the 1 Safari failure).

components/inputs/input-date.js
margaree: Would this be needed in `input-date-time` then since it does pass a label down to the `input-date` and `input-time` components?

dlockhart: Indeed it would -- thanks! I did `input-date` before I discovered this problem and then didn't circle back. Fixed now.

margaree (approved):

PR #1779, opened 2021-10-05
Fix filter axe tests

svanherk:
The axe tests for the opened filter weren't really doing anything, because everything was happening too fast.  Even if I added contrast failures it didn't find them.  If I actually told it to wait, there were a lot of false positives as it measured things while animations were still happening.  I need to be able to turn off animations for these tests to be useful, and I _think_ the below is fairly reasonable.  I tested that it properly catches contrast failures now.

We use the Chrome runner for axe tests, but I'm gonna put up a draft PR using Playwright just out of curiosity - there's some confusion over whether [it does](https://playwright.dev/docs/api/class-page/#page-emulate-media) or [does not](https://modern-web.dev/docs/test-runner/commands/#emulate-media) support this functionality.

### DIFF
--- a/components/filter/test/filter.axe.js
+++ b/components/filter/test/filter.axe.js
@@ -1,7 +1,7 @@
-import '../filter.js';
 import '../filter-dimension-set.js';
 import '../filter-dimension-set-value.js';
 import { expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { emulateMedia } from '@web/test-runner-commands';
 
 const singleSetDimensionFixture = html`
 	<d2l-filter>
@@ -28,6 +28,17 @@ const multiDimensionFixture = html`
 	</d2l-filter>`;
 
 describe('d2l-filter', () => {
+	before(async() => {
+		await emulateMedia({ reducedMotion: 'reduce' });
+		expect(matchMedia('(prefers-reduced-motion: reduce)').matches).to.be.true;
+		await import('../filter.js');
+	});
+
+	after(async() => {
+		await emulateMedia({ reducedMotion: 'no-preference' });
+		expect(matchMedia('(prefers-reduced-motion: no-preference)').matches).to.be.true;
+	});
+
 	[
 		{ name: 'Single set dimension', fixture: singleSetDimensionFixture },
 		{ name: 'Single set dimension - single selection', fixture: singleSetDimensionSingleSelectionOnFixture },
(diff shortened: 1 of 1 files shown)


components/filter/test/filter.axe.js
```diff
@@ -28,6 +28,17 @@ const multiDimensionFixture = html`
 	</d2l-filter>`;
 
 describe('d2l-filter', () => {
+	before(async() => {
+		await emulateMedia({ reducedMotion: 'reduce' });
```
dlockhart, line 32: Would it make sense to try and do this is [in our helper](https://github.com/BrightspaceUI/visual-diff/blob/main/helpers/createPage.js#L3), or is it too late if we've already loaded the files?

svanherk: I'm confused by what you mean - you mean create a similar helper for axe tests?  Since they aren't puppeteer and don't have an `html` file, I was running into the issue where https://github.com/BrightspaceUI/core/blob/main/components/hierarchical-view/hierarchical-view-mixin.js#L6 was happening before the emulation was happening, and it's never measured again (which is fine outside of tests).  If we want all axe tests to have reduced motion on, I think we could add it here: https://modern-web.dev/docs/test-runner/browser-launchers/chrome/#customizing-browser-context-and-page

dlockhart: Ah right sorry I got aXe and visual diff mixed up. But yeah the sentiment is the same -- to centralize this so that reduced motion is off for all aXe tests... if that's something we want. Probably added bonus would be a lot faster tests and a lot less waiting/timeouts.

svanherk: I can definitely look into it.  I'm not sure it'll actually help many other axe tests, it looks like we've avoided writing any that need to wait for something to animate/open - there's no tests for dropdown or hierarchical-view, or any where inputs are opened.  Which is probably fine if we have tests for the things inside those (menu, calendar, etc).  But filter just has a bit more going on so I wanted to get that working.  Applying it to all will help with anything similar in the future though

dlockhart: Cool. I'm also good with merging this as-is and then if it comes up again and maybe again a 3rd time then we invest in making it more generic.

dlockhart (approved):

svanherk:
Closing, going with https://github.com/BrightspaceUI/core/pull/1803 instead

PR #1780, opened 2021-10-05
WIP - backport count-badge into card-footer-link

annabelsegalld2l:
This PR replaces the card-footer-link's badge logic with the count-badge. Unfortunately, this does cause some breaking changes since the count badge purposefully does not support strings, only numbers (all the localization and truncating logic should live there). 

Some of the big changes:
- Adding a mandatory slot for the tooltip. I talked it over with Jeff and he agreed that the tooltip should be mandatory for the badge. I also experimented with passing the tooltip through to the badge to render it there, but that became super complicated since each card-footer-link currently controls its own tooltip (like forcing an opening direction, adding lists inside the tooltip). The slot for the tooltip is a good compromise, and we can move the tooltip slightly to get it aligned with the icon.
- Secondary-text becomes secondary-count, and takes in a number only. I will need to make some changes in the consumers, but they are pretty minimal (passing the number directly, removing some of the formatting)
- I have made the count-badge aria-hidden here, as all the screen readers really need is the offscreen text below the <a> element to make it clear that it is a link.
- I've also added a focus-ring property to the count-badge-mixin to forcefully put the focus ring on when focusing on the a element. Adding a tab-stop to the icon was my first attempt, but then tabbing to the icon will not allow you to get to the link. This property could be useful in numerous cases, like going through a list and perhaps highlighting the supporting badge.


components/card/card-footer-link.js
```diff
@@ -151,26 +127,26 @@ class CardFooterLink extends RtlMixin(LitElement) {
 				target="${ifDefined(this.target)}"
 				type="${ifDefined(this.type)}">
 				<span class="d2l-offscreen">${this.text}</span>
+				<d2l-count-badge-icon
+					aria-hidden="true"
+					icon="${this.icon}"
+					max-digits="${ifDefined(this.secondaryCountMaxDigits ? this.secondaryCountMaxDigits : undefined)}"
+					number="${noNumber ? 0 : this.secondaryCount}"
```
annabelsegalld2l, line 134: Passing the number 0 will render a zero on the badge. If consumers want no counts to show up, the secondary-count should not be set.

components/card/test/card-footer-link.visual-diff.html
```diff
@@ -14,30 +17,47 @@
...
 
 	<div class="visual-diff">
-		<d2l-card-footer-link id="secondary-count" icon="tier1:gear" text="Settings" href="https://www.d2l.com" secondary-text="99+" secondary-text-type="count"></d2l-card-footer-link>
+		<d2l-card-footer-link id="secondary-count" icon="tier1:gear" tooltip-text="text here" text="Settings" href="https://www.d2l.com" secondary-count="100" secondary-count-max-digits="2" secondary-count-type="count">
+			<d2l-tooltip slot="tooltip" for="secondary-count">100 settings applied</d2l-tooltip>
+		</d2l-card-footer-link>
 	</div>
 
 	<div class="visual-diff">
-		<d2l-card-footer-link id="secondary-notification-rtl" icon="tier1:gear" text="Settings" href="https://www.d2l.com" secondary-text="99+" secondary-text-type="notification" dir="rtl"></d2l-card-footer-link>
```
annabelsegalld2l, line 36: I've removed the rtl tests in card-footer-link and added some in the count-badge directly. Unfortunately, visual diff tests don't play nicely with nested rtl elements.

components/card/card-footer-link.js
```diff
@@ -151,26 +128,26 @@ class CardFooterLink extends RtlMixin(LitElement) {
...
 				type="${ifDefined(this.type)}">
 				<span class="d2l-offscreen">${this.text}</span>
+				<d2l-count-badge-icon
+					aria-hidden="true"
+					icon="${this.icon}"
+					max-digits="${ifDefined(this.secondaryCountMaxDigits ? this.secondaryCountMaxDigits : undefined)}"
+					number="${noNumber ? 0 : this.secondaryCount}" 
+					?hide-zero="${noNumber}"
+					text="${this.text}"
+					type="${this.secondaryCountType}">
```
annabelsegalld2l, line 138: Not sure why the linter is complaining at this line, saying that type accepts only 'count' or 'notification', which I set the definition to in this file.

PR #1783, opened 2021-10-06
Nested List - Selection Fixes

dbatiste:
This PR fixes a couple of selection issues:

* ensure that item selection state is updated before dispatching `d2l-list-item-selected` so that any consumers calling `getSelectionInfo` from within a `d2l-list-item-selected` or `d2l-list-selection-change` handler will get the correct result.
* rolls up the nested keys of items when updating their `selectionInfo` - this is necessary to evaluate the selected items correctly for the header summary.
* adding visual-diff tests for the header in both simple and nested cases, particularly to capture the observed selection state/counts


components/list/list-item-checkbox-mixin.js
```diff
@@ -100,13 +100,26 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
...
+		/* wait for internal state to be updated in case of action-click case so that a consumer
+		 calling getSelectionInfo will get the correct state */
+		await this.updateComplete;
 		this.dispatchEvent(new CustomEvent('d2l-list-item-selected', {
 			detail: { key: this.key, selected: value },
 			bubbles: true
 		}));
 	}
 
+	_getNestedList() {
```
dbatiste, line 113: I originally factored this out so that it could be used to evaluate the selected keys for a tree or sub-tree, but in the end, I found a much more performant method.  No harm in this factoring though, so I kept this change.

components/selection/selection-mixin.js
```diff
@@ -194,7 +193,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 
 		this._updateObserversRequested = true;
 		setTimeout(() => {
-			const info = this.getSelectionInfo();
+			const info = this.getSelectionInfo(true);
```
dbatiste, line 196: This is required so that providers will "roll-up" all the keys for the entire selection tree for observers.

PR #1791, opened 2021-10-08
fix: adding a delay before throwing missing label exceptions

dlockhart:
Hopefully this addresses the issues you were seeing in lists Dave.

The idea is to wait 3 seconds before checking for empty labels and throwing exceptions, in the hopes that any labels that get set not on the first update but maybe on a subsequent update will be present by then. If this is still an issue... we may have larger problems.


dlockhart:
After some offline discussions with @dbatiste, I've optimized this a bit such that it will only validate the label if the label actually changes or on `firstUpdated` in case the label is never set in the first place. Thanks for the suggestion!

PR #1804, opened 2021-10-14
Add custom dirty event to form

jorge-ramirez-arredondo:
This adds a `d2l-form-dirty` event that triggers whenever a native or custom form element fires an `input` or `change` event. Additionally, the `this._dirty` is now set to `true` at the same time as the new event fires. Before, `this._dirty` was only set to `true` when a native form element fired an `input`, `change`, or `focusout` event.

For context: `this._dirty` is being used with the `track-changes` attribute to show a native browser warning dialog when the user tries to navigate away from the page (i.e. when the window unloads). So, it's not built to support a save dialog for other scenarios (like a dialog closing). Another thing worth noting is that `this._dirty` can only be set to `false` by submitting the form.

### DIFF
--- a/components/form/docs/form.md
+++ b/components/form/docs/form.md
@@ -112,6 +112,7 @@ If you're looking to emulate native form element submission, `d2l-form-native` m
 ### Events
 - `d2l-form-submit`: Dispatched when the form is submitted. The form data can be obtained from the `detail`'s `formData` property.
 - `d2l-form-invalid`: Dispatched when the form fails validation. The error map can be obtained from the `detail`'s `errors` property.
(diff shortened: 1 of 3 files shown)


jorge-ramirez-arredondo:
@dlockhart I just realized that I forgot to add versioning to the commits. Should I just amend a commit to add `feat: ` and force push to this branch?

dlockhart:
> Should I just amend a commit to add feat: and force push to this branch?

Nope -- when you squash & merge the PR you can still amend the squash commit to put `feat:` in it (it'll be the top input box).

PR #1813, opened 2021-10-15
Create HTML data attribute context controller

grant-cleary:
The goal here is to create a faux reactive controller that other components can use to verify when the document html data attributes have been set on the page via a mutation observer on the document element. Multiple controllers can subscribe, but if any mutation matches, we disconnect the observer and move on.


grant-cleary:
Ah crap, I meant to open this as a draft...

grant-cleary:
Another open question is how specific this should be. As far as I know, ifrau's various syncing processes are the only ones that can actually lead to this situation occurring, so perhaps it's worth laser-targeting that. This current implementation is more generic.

PR #1824, opened 2021-10-22
Update HTML block to use attribute observer for re-rendering on context changes

grant-cleary:
There are two related PoCs in here:
1. Re-configuring the `HtmlAttributeObserverController` to take in an array of attributes instead of a single attribute (for bulk operations).
2. Integrating that into html-block component renderers.

I think (1) is relatively self-explanatory. (2) works under the principle that the html-block component will orchestrate re-stamping its template if/when one of these context objects is updated.

Note: This PR will not pass CI because tests haven't yet been updated.


helpers/htmlAttributeObserverController.js
```diff
@@ -1,24 +1,25 @@
...
+		this._host = host;
+		this._attributes = attributes;
+		this.values = new Map();
 	}
 
 	hostConnected() {
-		this.value = document.documentElement.hasAttribute(this._attribute)
-			? document.documentElement.getAttribute(this._attribute)
-			: undefined;
+		this._setAttributes(this._attributes);
```
dbatiste, line 22: Should the attributes be specified on the constructor or on the `hostConnected` call?  Could a situation arise where we don't know what attributes to observe until after the component is constructed, or we want to observe a different attribute after being reconnected?  Are these valid scenarios?  🤔

grant-cleary: Hmm, that's a good point. My intent here was to simulate what an actual reactive controller would do (which, as I understand it, would handle the `hostConnected` and `hostDisconnected` under the hood). But I'm also fine changing this behaviour. Certainly right now the attributes we're watching are immutable per controller, but it doesn't have to be that way.

helpers/htmlAttributeObserverController.js
```diff
@@ -27,12 +28,22 @@ export class HtmlAttributeObserverController {
...
-			? document.documentElement.getAttribute(this._attribute)
-			: undefined;
+	_handleContextChange(attributeNames) {
+		const attributes = attributeNames.filter(attr => this._attributes.includes(attr));
+		if (attributes.length === 0) return;
+		this._setAttributes(attributes);
 		this._host.requestUpdate();
 	}
 
+	_setAttributes(attributes) {
```
dbatiste, line 38: I think this name just threw me for a loop.  Should this be named `setAttributeValues` instead?  The way I read it above, I though this was setting the attribute names that should be observed.  I think it would be good to be explicit with `attributeName` and `attributeValue` to avoid any confusion.  Or perhaps `updateAttributeValues` would be even better, since it's taking in a list of attribute names and then obtaining the value from elsewhere, as opposed to a `setter` which would be taking the values as an arg.

components/html-block/html-block.js
```diff
@@ -155,15 +171,53 @@ class HtmlBlock extends LitElement {
...
+
+	updated() {
+		super.updated();
+		if (this._contextObjectHasChanged()) {
+			const template = this.querySelector('template');
+			this._stamp(template);
+		}
+	}
+
+	static get _renderers() {
```
dbatiste, line 194: Do we need to wrap this function call?

helpers/mathjax.js
```diff
@@ -1,5 +1,36 @@
 let mathJaxLoaded;
 
+export class HtmlBlockMathRenderer {
+
+	get contextAttribute() {
+		return 'data-mathjax-context';
```
dbatiste, line 6: Could there be a use-case where a renderer depends on more than one attribute?

helpers/htmlAttributeObserverController.js
```diff
@@ -1,24 +1,25 @@
...
+		mutations.map(mutation => mutation.attributeName)
+	));
 });
 
 export class HtmlAttributeObserverController {
 
-	constructor(host, attribute) {
-		this._host = host;
-		this._attribute = attribute;
+	constructor(host, ...attributes) {
```
dbatiste, line 11: Do you think it is better to pass the attribute names on the `constructor`, or should they be passed as an arg to `hostConnected`?  i.e. could there be use-case where we do not know the attributes to observer until after the component is constructed?

helpers/htmlAttributeObserverController.js
grant-cleary: Oh, good point, yes. I think this could be named to be much more clear. 👍

components/html-block/html-block.js
grant-cleary: What do you mean?

helpers/htmlAttributeObserverController.js
grant-cleary: Hmm, that's a good point. My intent here was to simulate what an actual reactive controller would do (which, as I understand it, would handle the `hostConnected` and `hostDisconnected` under the hood). But I'm also fine changing this behaviour. Certainly right now the attributes we're watching are immutable per controller, but it doesn't have to be that way.

components/html-block/html-block.js
dbatiste: Could the callers of this just call `getRenderers` directly?

helpers/htmlAttributeObserverController.js
dbatiste: If we're expecting this to eventually be a controller derived from something Lit provides, where those methods are called automatically, then I would stick with what you have.

grant-cleary: Fair. I'll double check to make sure I'm not misunderstanding the controller interface. If I'm not, I'll leave it as is.

components/html-block/html-block.js
grant-cleary: Good question. This is existing code (just moved), but I'm not sure if there's a reason to expose this as a static getter. I feel like nothing else needs to use it, so calling the function directly seems perfectly fine to me.

helpers/mathjax.js
grant-cleary: Hmm, that's a good question. I envisioned these (broadly speaking) as single context attributes specific to each renderer, similar to how the HTML editor context object currently works. I'm unsure if there's value in supporting the multiple-attribute case, given that these can be arbitrary objects as it is. 🤔

grant-cleary: That said, it would be trivial to support this now, so perhaps it would be better to just do so.

grant-cleary:
Crap, I committed my custom config changes again. Sigh.

grant-cleary:
Test failures are also interesting... I assume the replacements aren't happening for some reason, but I'll poke at it.

components/html-block/html-block.js
```diff
@@ -155,15 +171,51 @@ class HtmlBlock extends LitElement {
...
+			this._stamp(template);
+
+		});
+		this._renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
+		this._context = { ...this._contextObserverController.values };
+	}
+
+	updated() {
+		super.updated();
+		if (this._contextObjectHasChanged()) {
```
dbatiste, line 188: If none of the renderers depend on attributes, does this mean that the template won't be stamped initially?

components/html-block/html-block.js
```diff
@@ -155,15 +171,51 @@ class HtmlBlock extends LitElement {
...
 
 				let temp = document.createElement('div');
 				temp.appendChild(fragment);
 
-				for (const render of HtmlBlock._renderers) {
-					temp = await render(temp);
+				for (const renderer of getRenderers()) {
+					if (renderer.contextAttributes) {
+						const contextValues = new Map();
+						renderer.contextAttributes.forEach(attr => contextValues.set(attr, this._contextObserverController.values.get(attr)));
```
dbatiste, line 214: Maybe there is a way we can avoid having to do this for every instance of a `d2l-html-block`...?  I hesitate to provide the entire context though.

grant-cleary: So the mutation observer underlying the controller is shared, which is good, but if we did pass the entire context to each renderer we would avoid needing to do this mapping. I think filtering the map needs to be done either here or in the renderer, but I don't have a strong feeling of which is better.

components/html-block/html-block.js
grant-cleary: That's a good question. I feel like that's not the case, since this is strictly additive to what was there before. i.e. The slotchange event handler will still stamp the template. I can test this locally by stripping out the attribute on the math handler and checking if stamping still happens.

components/html-block/html-block.js
dbatiste: There is a single set of renderers in this module outside of the class.  We could keep the values for the attributes per renderer, either in this module or on the renderer which is shared across instances of html-block.  The difference is probably negligible even when there are many html-blocks though, since the collection of attributes should be so small.  If it weren't, I think avoiding the nested looping would be desirable.

grant-cleary: I'm not sure I understand what you mean. If we pass the full context object to the renderer, it can be responsible for pulling out its attributes. But ultimately the controller itself must be specific to each html-block.

dbatiste: Yep, we could pass them all, even though it didn't ask us to.  I am just saying that we could avoid this looping, either by passing them all, or caching the set of values per renderer outside the component so we only need to do it once. It would be a lookup in a Map vs looping over the entire collection for each html-block. But as mentioned, as long as the number of attributes is small, the work should be negligible.

grant-cleary: Ah okay, caching is an interesting option. I think that's probably better than passing the whole thing. I think you're right that this should be negligible because it's always possible to condense attributes if things start to get unreasonable. That said, I do like the idea of maintaining a map outside of the component that we can reference and update this way.

I think we can also get some minor efficiencies this way because the mutation observer will fire if I do something like this: `elem.setAttribute('attr', elem.getAttribute('attr'))` even though it's clearly not going to change anything. So there's probably some efficiency to be had in not triggering a re-render unless that mapped value has actually changed.

components/html-block/html-block.js
grant-cleary: So if no attributes were present we exploded trying to create the context controller. So I pushed a commit to fix that. I can confirm that beyond that though, the initial stamping still happens.

dbatiste:
I think this all makes sense.  Look'n good.

dbatiste (approved):

PR #1829, opened 2021-10-25
US133079 - Create a new DropdownOpenerMixin that also opens on hover events

annabelsegalld2l:
This PR creates the DropdownHoverOpenerMixin (and a DropdownHoverButton for testing purposes). Like the regular dropdownOpenerMixin, it can be extended to work as an opener for any dropdown content. I've added tests to demonstrate the behaviour of the mixin:

- Opens/closes on opener click (as usual)
- Opens on hover, closes on hover away from the opener
- 400ms fade out if hovered away (no fading on clicks away)
- If hovering on the opener and moving to the dropdown content, it will not close. However, after hovering from the content or opener, the dropdown closes. This allows users to hover over the opener to get it to open, and quickly hover over the content. 
- Otherwise, the dropdown should work as usual (ie. clicking escape to close, tab away to close)


components/dropdown/dropdown-content-styles.js
```diff
@@ -172,6 +172,11 @@ export const dropdownContentStyles = css`
 		text-align: right;
 	}
 
+	:host(.d2l-dropdown-content-fading) {
```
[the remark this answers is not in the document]
dbatiste: We should include `transition: none;` for the pefers reduced motion case.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -73,6 +127,43 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this.addEventListener('d2l-dropdown-close', this.__onClosed);
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (!this.openOnHover) return;
```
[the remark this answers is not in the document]
dbatiste: A consequence of doing this in `updated` is that if the dropdown is opened, subsequently disconnected and then reconnected to the DOM (ex. drag & drop), the handlers would be removed (above in `disconnectedCallback`) but they would not be re-added unless `updated` is called for some other reason.  This is perhaps an unlikely scenario though - it might require hovering to open the dropdown and then switching to keyboard to perform such an action.  Another way would be if the dropdown content contained an action that caused this disconnect/reconnect scenario.  Strange, but stranger things have been done.  

Is it necessary to wire-up directly to the opener and content elements - can we just wire up to `this`? In such a case, we avoid accessing the `this.shadowRoot` and the wire-up can be done in `connectedCallback`.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -133,6 +267,68 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
...
+		this._isOpenedViaClick = true;
+		if (!this._isOpen) this.openDropdown(true);
 	}
 
+	/* used by open-on-hover option */
+	_closeTimerStart() {
+		if (this._isOpen) return;
+		clearTimeout(this._setTimeoutId);
+		this._isFading = true;
+		this._setTimeoutId = setTimeout(() => {
```
dbatiste, line 311: Is the purpose of this `setTimeout` to match the opacity transition time, so that we don't actually `display: none` on the content until the animation completes?

annabelsegalld2l: That's correct. If we did have the display: none, it would be impossible to re-open the dropdown via hover, since it would disappear from the screen. This way, users have 400ms to hover to the dropdown before it closes.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -133,6 +267,68 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
...
+	/* used by open-on-hover option */
+	_closeTimerStop() {
+		clearTimeout(this._setTimeoutId);
+		this._isFading = false;
+	}
+
+	/* used by open-on-hover option */
+	_onOutsideClick(e) {
+		if (!this._isOpen) return;
+		const isWithinDropdown = isComposedAncestor(this.__getContentElement(), (e.path || e.composedPath())[0]);
```
dbatiste, line 326: I think I need a reminder - under what circumstances do we not have a `path`?  Normally I end up always using `composedPath` because I am interested in the complete path, but when I inspect these two in the console, they appear identical.  Mozilla only lists `composedPath` in their docs.

components/dropdown/dropdown-opener-mixin.js
dbatiste: Normally we would listen for the `transitionend` (and `transitioncancel`/`transitionrun` when necessary) event to be in sync with the duration (and perhaps delay) that is specified in the CSS.  Note that if the user prefers reduced motion and it's closed immediately, the timing specified here would need to be different as well I think - another consideration.

dbatiste: A follow-up thought on this - if there is no animation and the content is hidden immediately, does that present an issue when the user moves their mouse from the opener to the content?

annabelsegalld2l: My idea for prefer-reduced-motion is actually to keep the content open for the 400ms, just without the fade-out. After those 400ms, it will close with no transition. That will stop the issue of hiding the content immediately and gives the user time to hover to the content. What are your thoughts on that approach?

components/dropdown/dropdown-opener-mixin.js
annabelsegalld2l: Ah, looks like this was a copy-paste from the original user-profile-card code. My best guess is that the path || composedPath check was added before full browser support of composedPath. Looking at our other core code, looks like we are safe to just use composedPath. I'll make that change.

components/dropdown/dropdown-opener-mixin.js
dbatiste: I think that's fine - it sounds like we will need a timeout no matter what to prevent it from closing immediately.  If it were not for that, I would recommend leveraging the events to tie into the animation.

I might be inclined to add a comment in the two spots where the 400ms is specified - "matches setTimeout" and "matches CSS" so if anyone goes to change either one, they know that they need to change the other as well.

dbatiste (approved):
Looks great!

PR #1838, opened 2021-10-28
US132674 - Filter Visual Diff Tests

svanherk:
~This uses the new `allowedPixels` option in https://github.com/BrightspaceUI/visual-diff/pull/54.  Once that merges, going to rerun this a bunch of times to test for flakes.~ Removed the dependency on that PR, as increasing the `deviceScaleFactor` seems to have fixed the issue for now.


annabelsegalld2l (approved):
LGTM! Visual diffs look as expected. I think there is a fair bit of duplication in the diffs, like screenshotting all the dimensions in the multi-select case. It's good to have those while the filter is still in development, but we should revisit later to see if we can cut down on some unneeded tests for the sake of time ⏲️

svanherk:
Yeah I agree, I already cut a bunch but there's so many little combos that I wonder if they're worth the time it takes to run the test.  I'll revisit as we add more dimension types, cause we don't want that to double the tests each time we do that

PR #1844, opened 2021-11-01
Re-connect template observer if html-block is re-connected to the DOM

grant-cleary:
If an `html-block` is detached from the DOM, we disconnect its template observer. However, when re-connecting to the DOM, we never re-attach it. This leads to an interesting problem where if the template contents change, there's no way for the component to pick up and act on that change.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -135,6 +135,14 @@ class HtmlBlock extends LitElement {
 		`];
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		if (!this._templateObserver) return;
+
+		const template = this.querySelector('template');
+		if (template) this._templateObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
+	}
+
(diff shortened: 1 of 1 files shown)


components/html-block/html-block.js
```diff
@@ -135,6 +135,14 @@ class HtmlBlock extends LitElement {
 		`];
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		if (!this._templateObserver) return;
```
grant-cleary, line 140: If the observer doesn't already exist, we don't need to do anything here, since presumably `firstUpdated` hasn't yet run.

components/html-block/html-block.js
```diff
@@ -135,6 +135,14 @@ class HtmlBlock extends LitElement {
...
+
+		const template = this.querySelector('template');
```
grant-cleary, line 142: Not sure if there's a better way to fetch the template here.

dbatiste (approved):

PR #1845, opened 2021-11-02
feat: US133079: Removing the touchstart event, fixing opener check

annabelsegalld2l:
After some testing, looks like the touchstart event is a duplicate of the clicking action. Removing that listener will simplify the code a bit!

Also, I changed the dropdown click behaviour to lock open the dropdown, so that if users click on the dropdown, they can hover away and the dropdown will not close. This is consistent with the old user-profile card. Clicking inside the dropdown, similar to clicking the opener, should allow the dropdown to be kept open. I added a test for this as well.

Also, I fixed the opener check in the listeners. There are generally two cases with the mouseup/leave events:

- Src element is the opener itself (ie. d2l-dropdown-more, this is how most of our buttons work)
- Src element is some element within the button (ie. a span within the opening button)
- We can now cover both those cases. Previously, we were only covering the first case, since the content within a button failed the check to see if it was an ancestor of the button.

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -210,14 +210,14 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	}
 
 	__onKeypress(e) {
-		if (isComposedAncestor(e.srcElement, this.getOpenerElement())) {
+		if (e.srcElement === this || isComposedAncestor(this.getOpenerElement(), e.srcElement)) {
 			this.__onOpenerKeyPress(e);
 		}
 	}
 
 	__onMouseEnter(e) {
 		if (!this.openOnHover) return;
-		if (isComposedAncestor(e.srcElement, this.getOpenerElement())) {
+		if (e.srcElement === this || isComposedAncestor(this.getOpenerElement(), e.srcElement)) {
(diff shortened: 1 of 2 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -204,6 +202,13 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this._closeTimerStart();
 	}
 
+	__onDropdownMouseUp() {
+		this._isOpen = true;
+		this._isFading = false;
+		this._isOpenedViaClick = true;
```
annabelsegalld2l, line 208: Setting _isOpenedViaClick will prevent the dropdown from closing on hovering away.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -172,6 +169,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 			return;
 		}
 		content.toggleOpen(applyFocus);
+		this._isOpen = !this._isOpen;
```
annabelsegalld2l, line 172: This function isn't being used by the hover mixin, but I added this line in case a consumer wants to toggle the dropdown.

dbatiste (approved):

svanherk (approved):

PR #1848, opened 2021-11-03
US133079: More dropdown opener cleanup

annabelsegalld2l:
Just made some misc, fixes and added some tests for the dropdown opener class. The behaviour should stay the same.

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -60,10 +60,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this._isHovering = false;
 		this._isFading = false;
 
-		this.__onWholeKeypress = this.__onKeypress.bind(this);
-		this.__onMouseUp = this.__onMouseUp.bind(this);
-		this.__onMouseEnter = this.__onMouseEnter.bind(this);
-		this.__onMouseLeave = this.__onMouseLeave.bind(this);
+		this._onOutsideClick = this._onOutsideClick.bind(this);
 		this._contentRendered = null;
 		this._openerRendered = null;
 	}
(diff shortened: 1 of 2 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -188,49 +185,50 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
...
 		this._isFading = false;
+		this._isHovering = false;
 		this._isOpenedViaClick = true;
 		this._closeTimerStop();
 	}
 
 	__onKeypress(e) {
 		if (e.srcElement === this || isComposedAncestor(this.getOpenerElement(), e.srcElement)) {
-			this.__onOpenerKeyPress(e);
+			if (e.keyCode !== 13 && e.keyCode !== 32) return;
```
annabelsegalld2l, line 198: Moved the previous _onOpenerKeypress into this function to get rid of an unneeded extra definition there.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -60,10 +60,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this._isHovering = false;
 		this._isFading = false;
 
-		this.__onWholeKeypress = this.__onKeypress.bind(this);
```
annabelsegalld2l, line 63: This was leftover from when I was listening at the window level. The only function that actually needs to get bound is the _onOutsideClick function.

components/dropdown/dropdown-opener-mixin.js
```diff
@@ -325,10 +287,11 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 	_onOutsideClick(e) {
 		if (!this._isOpen) return;
```
annabelsegalld2l, line 288: Added a check to ensure clicks coming from the opener do not trigger a close - had to add this due to the binding function above. Just a note on performance - I've used the composedAncestor functions many times in this function, BUT they will only be called with the hover option and only for open dropdowns.

dbatiste (approved):

PR #1857, opened 2021-11-08
Add compose mixins helper

semibran:
Proposed addition as an aside from some of the [rubrics Lit conversion work](https://github.com/Brightspace/d2l-rubric/pull/1295#discussion_r743976360) that Mango is doing

#### Before
```js
class ExampleComponent extends LocalizeMixin(EntityMixin(LitElement)) {
  // ...
}
```

#### After
```js
class ExampleComponent extends composeMixins(
  LitElement,
  EntityMixin,
  LocalizeMixin,
) {
  // ...
}
```

Using the helper is more readable, allows for cleaner diffs, and makes it easier to add and arrange mixins in a logical fashion.

### DIFF
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -21,6 +21,25 @@ import { AsyncStateEvent } from '@brightspace-ui/core/helpers/asyncStateEvent.js
 const asyncStateEvent = new AsyncStateEvent(pendingPromise);
 ```
 
+## composeMixins
+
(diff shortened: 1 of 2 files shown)


dlockhart (approved):
Very elegant and useful -- thanks! Could you add a small section with a code sample [in the README](https://github.com/BrightspaceUI/core/blob/main/helpers/README.md) for helpers?

dlockhart (approved):

PR #1860, opened 2021-11-09
fix: cleanup extra scripts in an attempt to fix opened issue

margaree:
Also threw in some alphabetizing and removal of Future Enhancements sections

Would need https://github.com/BrightspaceUI/documentation/pull/421


components/dropdown/README.md
```diff
@@ -130,21 +108,10 @@ If the dropdown is initially empty when it's opened, the dropdown pointer will n
 <!-- docs: demo live name:d2l-dropdown-button-subtle align:flex-start autoSize:false size:medium -->
 ```html
 <script type="module">
+  import './pages/assets/open-first-item-helper.js';
```
GeorgePlukov, line 111: I didn't think of this until now but these code snippets become a little less copy/paste-able if this is done as an import. Does the import show up in the code snippet as well, is that intended?

GeorgePlukov (approved):

components/dropdown/README.md
margaree: It currently does but that would be something we could easily filter at the Daylight level

PR #1862, opened 2021-11-09
DE45955 - Make count bubble overflow - do not count in badge area

annabelsegalld2l:
This makes sure that the badges are centered on the icon, not taking into account the count bubble. The bubble can overflow into the DOM. See comments for some questions I'd like feedback for!


components/count-badge/count-badge-icon.js
```diff
@@ -33,23 +33,35 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
...
+			position: relative;
+			padding-left: 0.5rem;
+			padding-right: var(--d2l-count-badge-icon-horizontal-padding);
 		}
 
 		:host([dir="rtl"]) {
-			padding-left: var(--d2l-count-badge-icon-padding);
-			padding-right: 0;
+			padding-left: var(--d2l-count-badge-icon-horizontal-padding);
+			padding-right: 0.5rem;
```
annabelsegalld2l, line 50: Our "best-effort" padding to prevent most overflows is 0.5rem on each side, to keep it symmetrical. The default of --d2l-count-badge-icon-horizontal-padding is also 0.5rem;

components/count-badge/count-badge-icon.js
```diff
@@ -79,18 +91,31 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
...
+		if (excessDigits >= 0) {
+			const extraRem = excessDigits / 2;
+			horizontalPadding = `calc(0.5rem + ${extraRem}rem)`;
+		}
+		this.style.setProperty('--d2l-count-badge-icon-horizontal-padding', `${horizontalPadding}`);
 		const tabbable = (this.tabStop || this.hasTooltip) && !this.skeleton;
 		const innerHtml = html`
-			${this.renderCount(numberStyles)}
+			<div class="d2l-count-badge-number-wrapper">
+				${this.renderCount(numberStyles)}
```
annabelsegalld2l, line 117: To get the overflow to work, I had to add a position: absolute wrapper to the count bubble so it would not be counted in the host's area.

components/count-badge/count-badge-icon.js
```diff
@@ -79,18 +91,31 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
...
-			visibility: this.skeleton ? 'hidden' : undefined
+			visibility: this.skeleton ? 'hidden' : undefined,
+			display: 'inline-block'
 		};
+
+		let horizontalPadding = '0.5rem';
+		const excessDigits = this.getNumberString().length - 4;
+		// large number strings could overflow onto the next badge,
+		// update padding in this case
+		if (excessDigits >= 0) {
```
annabelsegalld2l, line 109: This part is a bit funky... looking for feedback here. 

For languages with long plus strings (ie. Plus de 99 in french), the bubble will overflow wayyy too far and hit other badges in the MCW. 

One option is to cap the width to prevent all overflows and have a vertical stack style. Pro: this will preserve the centering. Cons: looks weird and makes the badge quite tall. 
multiword-stacked

The alternate approach I've gone with here is a best-of-both-worlds? For short strings less than 4 characters, allow a small amount of overflow. This will never get to other badges. This will preserve the centering. 
For long strings, we know it will absolutely overflow. We can add some backup padding in this case. Here, I've got 0.5 for each extra character over 4 characters. The centering will not be preserved in this case, but at least we will be able to read the number. 
extra-padding-long-nums-only

Not sure which approach to go with. We can also scrap the long langterms and go back to 99+. This will still allow overflow for count badges without limits, like in Portfolio.

components/count-badge/count-badge-icon.js
```diff
@@ -33,23 +33,35 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
 			display: inline-block;
 		}
 
+		.d2l-count-badge-number-wrapper {
+			position: absolute;
+			width: max-content;
```
annabelsegalld2l, line 38: This line allows us to overflow. Removing it, and we get the stacked option for multi-word strings.

components/count-badge/count-badge-icon.js
dlockhart: Instead of this extra `<div>`, could the absolute positioning not have been applied to `.d2l-count-badge-number`?

components/count-badge/count-badge-icon.js
dlockhart: I was away when these translations came back, but now that the implication of their choice to not just stick with `99+` is even larger, should we check again that it's actually the correct interpretation for those languages? i.e. could we make these not language terms at all and just use `99+` for ALL languages.

components/count-badge/count-badge-icon.js
```diff
@@ -79,18 +91,31 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
 	}
 
 	render() {
-		const numberPadding = this.size === 'small' ? '0.55rem' : '0.7rem';
+		const numberPadding = this.size === 'small' ? '-0.65rem' : '-0.7rem';
```
dlockhart, line 94: So now that things are positioned absolutely, could we switch to using `top`/`right` to position the number and maybe a transform to slide it up 50% of its height?

Something like this:
https://codepen.io/dlockhart/pen/VwzdxMK

Screen Shot 2021-11-09 at 4 38 47 PM

annabelsegalld2l: Yep that seemed to work! Didn't need the wrapper div either.

components/count-badge/count-badge-icon.js
annabelsegalld2l: Yep that was my thought too. I'll mention it in standup tomorrow. I think it's still good to have the padding function as a backup for the case where we have a super long number. I'm not sure it should be up to the consumers to be monitoring the size of the bubble which would require going into the shadowroot.

dlockhart: So if the positioning is done like my codepen below, it will "overflow" to the left overtop of the icon. Then maybe we could cap it at X characters such it can never be large enough to overflow beyond the left boundary of the icon.

components/count-badge/count-badge-icon.js
```diff
@@ -29,37 +29,35 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
...
+			position: relative;
 		}
 
 		:host([icon*="tier1:"]) {
 			--d2l-count-badge-icon-height: 18px;
+			--d2l-count-badge-icon-padding: calc(-50% + 11px);
 		}
 		:host([icon*="tier2:"]) {
 			--d2l-count-badge-icon-height: 24px;
+			--d2l-count-badge-icon-padding: calc(-50% + 14px);
```
annabelsegalld2l, line 46: This fun math tells us the x transform value to center the count badge over the bubble (used for long numbers beyond 3 digits)

components/count-badge/count-badge-icon.js
```diff
@@ -79,15 +77,32 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
...
 			border: '2px solid white',
-			position: 'relative',
-			left: this.dir === 'rtl' ? 0 : 'var(--d2l-count-badge-icon-padding)',
-			right: this.dir === 'rtl' ? 'var(--d2l-count-badge-icon-padding)' : 0,
-			top: numberPadding,
-			visibility: this.skeleton ? 'hidden' : undefined
+			visibility: this.skeleton ? 'hidden' : undefined,
+			display: 'inline-block',
+			position: 'absolute',
+			width: 'max-content'
```
annabelsegalld2l, line 85: Width: max-content allows the badge to overflow slightly outside the padding if required.

components/count-badge/count-badge-mixin.js
```diff
@@ -156,25 +157,29 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
...
 		super.connectedCallback();
-		if (!this.maxDigits && this.type === 'notification') {
-			// default to two digits for notification type
-			this.maxDigits = 2;
+		if (!this.maxDigits) {
+			// default to two digits for notification type, 5 for count
+			this.maxDigits = this.type === 'notification' ? 2 : 5;
+		} else if (this.maxDigits > 5) {
+			// limit all badges to 5 digits
+			this.maxDigits = 5;
```
annabelsegalld2l, line 165: I added the limit in both connectedCallback and updated so that the limit is enforced from the start for elements that never update (like content tool), and for demo pages where you can update the maxDigits manually.

components/count-badge/count-badge-mixin.js
```diff
@@ -183,9 +188,20 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
 			numberString = formatNumber(numberString);
 		}
 
+		return numberString;
+	}
+
+	renderCount(numberStyles) {
```
annabelsegalld2l, line 194: Extracted this logic into a public function so the icon badge can determine the length of the number string.

annabelsegalld2l:
Closing this PR for now until we hear back from localization for the '99+' langterm.

PR #1864, opened 2021-11-10
fix: html block Latex demos

margaree:


components/html-block/demo/html-block.html
```diff
@@ -39,7 +39,7 @@
 		</script>
 		<script>
 			if (window.location.search.indexOf('latex=true') !== -1) {
-				document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ renderLaTeX: true });
+				document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ renderLatex: true });
```
margaree, line 42: I noticed the html block demo wasn't working when `latex=true` was used. Looks like https://github.com/BrightspaceUI/core/blob/main/helpers/mathjax.js#L5 was updated to use `renderLatex` so this just makes that consistent.

components/html-block/README.md
```diff
@@ -109,26 +109,25 @@ Examples are provided to display how user-authored math can be embedded within y
...
+  D2L.LP.Web = {};
+  D2L.LP.Web.UI = {};
+  D2L.LP.Web.UI.Flags = {
+    Flag: (feature, defaultValue) => {
+      if (feature === 'us125413-mathjax-render-latex') return true;
+      else return defaultValue;
+    }
+  };
+  <!-- docs: start hidden content -->
+  document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ renderLatex: true });<!-- docs: end hidden content -->
```
margaree, line 124: This demo didn't seem to show the Latex rendered properly so I used this to do so

GeorgePlukov (approved):

PR #1865, opened 2021-11-11
feat: Removing IE11 CSS variable setters

annabelsegalld2l:
Legacy edge is gone so we are safe to remove these styleSubtree CSS setters 🎉 
I did test on a current Edge version to make sure it was still working by setting the vertical-offset attribute.


dlockhart (approved):
🎉

PR #1868, opened 2021-11-11
Visual diffs: add a requestAnimationFrame to mobile dropdowns

annabelsegalld2l:
For dropdowns with empty content, the focus should go to the close button on the mobile tray. Sometimes, the tests get flaky because the close button is not getting focused on. The dropdown requests a new frame before focusing, so adding that to the test as will **should** stop the flakiness. 

I'll rerun a couple of times to make sure the flake doesn't happen again.


dlockhart (approved):
👍

PR #1873, opened 2021-11-17
US133625: Add 'Now' button to date-time (range) picker

annabelsegalld2l:
This PR adds a 'now-button' attribute to the date-picker, to be used for date-times only. Clicking the now button will overwrite the existing date and time to the current time. 

I also changed the langterm 'Set to Today' -> 'Today' to save space, and centered the buttons.


components/inputs/input-date.js
```diff
@@ -61,6 +61,11 @@ class InputDate extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeCor
 			 * Disables validation of max and min value. The min and max value will still be enforced but the component will not be put into an error state or show an error tooltip.
 			 */
 			noValidateMinMax: { attribute: 'novalidateminmax', type: Boolean },
+			/**
+			 * @ignore
+			 * Optionally add a 'Now' button to be used in date-time pickers only.
+			 */
+			nowButton: { attribute: 'now-button', type: Boolean },
```
annabelsegalld2l, line 68: Added the ignore tag here since this option should be used for the date and time pickers together, and not for just date-pickers.

components/inputs/input-time.js
```diff
@@ -47,6 +47,11 @@ export function getDefaultTime(time, enforceTimeIntervals, timeInterval) {
 			return END_OF_DAY;
 		case 'startOfDay':
 			return enforceTimeIntervals ? MIDNIGHT : START_OF_DAY;
+		case 'now': {
+			const today = getToday();
```
annabelsegalld2l, line 51: Added a 'now' case for getDefaultTime that will return the current time without rounding to the nearest interval.

components/inputs/input-date.js
dlockhart: Nit: I might suggest naming this something like `has-now` without exposing the fact that we're using a button to implement it. That way if we change the underlying implementation in the future, the attribute still makes sense. Although since it's internal it doesn't _really_ matter.

components/inputs/input-date.js
```diff
@@ -503,10 +516,13 @@ class InputDate extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeCor
...
 
-	async _updateValueDispatchEvent(dateInISO) {
-		if (dateInISO === this._shownValue) return; // prevent validation from happening multiple times for same change
+	async _updateValueDispatchEvent(dateInISO, setToNow) {
+		// prevent validation from happening multiple times for same change,
+		// except for now button that affects time
+		if (!setToNow && dateInISO === this._shownValue) return;
 		this._shownValue = dateInISO;
 		this.value = dateInISO;
+		this.setToNow = setToNow;
```
dlockhart, line 525: Instead of storing this on the element (which becomes meaningless after the event is dispatched), I'd suggest storing it in the event detail (an optional 3rd parameter on `CustomEvent`). Then the listener can check that instead of looking at the target.

components/inputs/input-time.js
dlockhart: Technically I think this means they can go `<d2l-input-time value="now">`, right? Not sure if we actually want that since that's the same as not setting a value at all I believe.

Also, how does `enforce-time-invervals` work with this?

annabelsegalld2l: This affects the default-value attribute in the time and date-time picker, not the value. So consumers could set default-value="now', but that functionality is currently not exposed.

As for `enforce-time-intervals`. the date-time input will always have that property set to false, so we technically couldn't get to a case where we needed to enforce them here. The question is if we do want to expose the now option to consumers or keep it hidden.

dlockhart: Ah right, `default-value`.

Makes sense about `enforce-time-intervals`. I'm good keeping it hidden until a valid use case comes up.

annabelsegalld2l: I ended up just adding a ternary statement to enforce the intervals in the case the now option becomes exposed in the future. We can assume it won't be used for now and always return the date since it's only used in date-time.

components/inputs/input-date.js
```diff
@@ -71,6 +71,11 @@ class InputDate extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeCor
 			 * @type {boolean}
 			 */
 			required: { type: Boolean, reflect: true },
+			/**
+			 * @ignore
+			 * Optionally add a 'Now' button to be used in date-time pickers only.
+			 */
+			setToNow: { attribute: 'has-now', type: Boolean },
```
dlockhart, line 78: I think this should now be `hasNow` to match the attribute name?

dlockhart (approved):
Looks good!

PR #1876, opened 2021-11-19
fix: fixing hovering dropdown focus behaviour on click - should not force focus

annabelsegalld2l:
Just lumped two one-liners into one PR here. 

1. The hovering dropdown was previously forcing focus on click events, similar to keyboard openings. I've changed it to not apply focus, to be consistent with our other dropdown behaviour (note that the PR that George is working on might force focus anyways, but not visibly).
2. Adding an applyFocus detail to the focus-enter event. This will allow consumers to know if the focus was actually applied to the first element. The focus-enter event will always be dispatched regardless of applying focus, so adding the detail lets consumers know when to override that focus (ie. profile card will use this to force focus on another element, but only on keyboard events).

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -472,7 +472,6 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 				|| isComposedAncestor(this.__getOpener(), activeElement)) {
 				return;
 			}
-
 			this.close();
 		}, 0);
 	}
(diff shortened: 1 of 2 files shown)


dlockhart (approved):

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1023,7 +1022,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			}
 		}
 		/** Dispatched when user focus enters the dropdown content (trap-focus option only) */
-		this.dispatchEvent(new CustomEvent('d2l-dropdown-focus-enter'));
+		this.dispatchEvent(new CustomEvent('d2l-dropdown-focus-enter', { detail:{ applyFocus: this.__applyFocus } }));
```
svanherk, line 1025: Minor: `appliedFocus` might make this a bit more clear, although not sure if we often use past tense

svanherk (approved):

PR #1882, opened 2021-11-22
fix: remove optional chaining syntax to prevent transpile errors in activities

annabelsegalld2l:
Older tools like polymer don't recognize the ?. syntax, leading to some transpile errors. See https://github.com/BrightspaceHypermediaComponents/activities/runs/4291411518?check_suite_focus=true for an example.


dlockhart (approved):
Thanks for the quick fix!

PR #1883, opened 2021-11-22
Update to latest stylelint config.

dbatiste:
This PR updates to the latest stylelint config.  It also enables linting styles in HTML files, and fixes those existing errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint": "npm run lint:eslint && npm run lint:style && npm run lint:lit",
     "lint:eslint": "eslint . --ext .js,.html",
     "lint:lit": "lit-analyzer \"{components,directives,helpers,mixins,templates,test,tools}/**/*.js\" --strict",
-    "lint:style": "stylelint \"**/*.js\"",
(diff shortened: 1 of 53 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "lint": "npm run lint:eslint && npm run lint:style && npm run lint:lit",
     "lint:eslint": "eslint . --ext .js,.html",
     "lint:lit": "lit-analyzer \"{components,directives,helpers,mixins,templates,test,tools}/**/*.js\" --strict",
-    "lint:style": "stylelint \"**/*.js\"",
+    "lint:style": "stylelint \"**/*.{js,html}\"",
```
dbatiste, line 17: This enables linting styles in out HTML files.  If we think this will be annoying, we can leave these out, but it did find quite a few existing linting errors so it seems worthwhile even though it's just demo/test files.  I think it might have been excluded before either because a special plugin might have been required, or because we tend to use lots of adhoc non-prefixed class names in those files which would have been flagged as errors before.

dlockhart (approved):
Looks good!

PR #1885, opened 2021-11-24
Support synchronous (non-deferred) rendering in html-block

grant-cleary:
This is a shot at supporting synchronous rendering within the html-block component. There are a couple things to point out here:
- We can't synchronously render math. The MathJax shadowDOM adaptor is incompatible with this, since it itself pulls the rendering elements into a new document, manipulates them, and then renders them back to the DOM. MathJax's APIs seem to expect a full document as input, so I see no way around this.
- Related to the above, we need to prevent any renderers with this requirement from running if we're in non-deferred mode. I figure that's the purpose the `canRenderInline` property on the renderers can serve.
- Given that this should only be used in specific LMS use cases, I've effectively made this behaviour opt-in via replacing the usual `template` with a `div` with a specific class.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -140,7 +140,7 @@ class HtmlBlock extends LitElement {
 			:host([no-deferred-rendering="true"]) div.d2l-html-block-rendered {
 				display: none;
 			}
-			:host([no-deferred-rendering="false"])::slotted(*) {
+			:host([no-deferred-rendering="false"]) ::slotted(*) {
 				display: none;
 			}
 		`];
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -118,6 +118,12 @@ const getRenderers = () => {
  */
 class HtmlBlock extends LitElement {
 
+	static get properties() {
+		return {
+			_deferRendering: { Type: Boolean, attribute: 'defer-rendering', reflect: true }
+		};
+	}
+
 	static get styles() {
 		return [ htmlBlockContentStyles, css`
 			:host {
@@ -131,7 +137,10 @@ class HtmlBlock extends LitElement {
 			:host([hidden]) {
 				display: none;
 			}
-			::slotted(*) {
+			:host([defer-rendering="false"]) div.d2l-html-block-rendered {
+				display: none;
+			}
+			:host([defer-rendering])::slotted(*) {
 				display: none;
 			}
 		`];
@@ -177,7 +186,8 @@ class HtmlBlock extends LitElement {
 			const template = e.target.assignedNodes({ flatten: true })
 				.find(node => (node.nodeType === Node.ELEMENT_NODE && node.tagName === 'TEMPLATE'));
 
-			this._stamp(template);
+			if (template) this._stamp(template);
+			else this._renderInline();
 
 		});
 		this._renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
@@ -188,7 +198,8 @@ class HtmlBlock extends LitElement {
 		super.updated();
 		if (this._contextObserverController && this._contextObjectHasChanged()) {
 			const template = this.querySelector('template');
-			this._stamp(template);
+			if (template) this._stamp(template);
+			else this._renderInline();
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


grant-cleary:
Visual diff failure appears to be legit. Something about the styling is now off and the non-template case is displaying when it shouldn't.

components/html-block/html-block.js
```diff
@@ -118,6 +118,12 @@ const getRenderers = () => {
  */
 class HtmlBlock extends LitElement {
 
+	static get properties() {
+		return {
+			_deferRendering: { Type: Boolean, attribute: 'defer-rendering', reflect: true }
```
grant-cleary, line 123: I need to flip the default here.

components/html-block/html-block.js
```diff
@@ -131,14 +137,18 @@ class HtmlBlock extends LitElement {
 			:host([hidden]) {
 				display: none;
 			}
-			::slotted(*) {
+			:host([no-deferred-rendering="true"]) div.d2l-html-block-rendered {
```
grant-cleary, line 140: I don't really have a solid idea of why this is necessary. When this property is being reflected back to an attribute, it's being explicitly set this way, not being added/removed like I would normally expect for a boolean attribute. 😕

components/html-block/html-block.js
```diff
@@ -131,14 +137,18 @@ class HtmlBlock extends LitElement {
...
+				display: none;
+			}
+			:host([no-deferred-rendering="false"])::slotted(*) {
```
grant-cleary, line 143: Similarly, the visual diff tests are failing because sometimes this attribute doesn't get added at all and sometimes it does, but with the explicit value `false`. I feel like I'm missing something simple here, so I'll circle back tomorrow and see if I can make heads or tails of this.

components/html-block/html-block.js
```diff
@@ -118,6 +118,12 @@ const getRenderers = () => {
  */
 class HtmlBlock extends LitElement {
 
+	static get properties() {
+		return {
+			_noDeferredRendering: { type: Boolean, attribute: 'no-deferred-rendering', reflect: true }
```
grant-cleary, line 123: So... `type` was typo'd as `Type` and Lit was ignoring the extra property and (correctly) assuming it should use its default serialization (which is `type: String`). Fixing the typo makes this behave like a boolean attribute as expected.

dbatiste: Is this designated at "private" because we only want the LMS to use it?

grant-cleary: So right now this isn't what the component uses to determine whether rendering should be deferred. Instead, the component just sets this internally so styling can be applied properly. I was trying to keep to the original pattern of "if it's a template, defer; otherwise don't", with the additional caveat that the consumer needs to provide a specific container `div` for us to actually render.

grant-cleary: So basically as written now, nothing should set this externally because the html-block is just going to overwrite it anyways.

components/html-block/html-block.js
```diff
@@ -188,7 +199,8 @@ class HtmlBlock extends LitElement {
 		super.updated();
 		if (this._contextObserverController && this._contextObjectHasChanged()) {
 			const template = this.querySelector('template');
```
dbatiste, line 201: It is intriguing that above, it queries via `assignedNodes` to get the template, whereas here it just queries the lightDOM using `this.querySelector`.  Technically, the resolution that `assignedNodes({ flatten: true })` is more desirable.  I wonder why this line doesn't do that.

grant-cleary: Ah, mostly because when I originally implemented this, I thought `assignedNodes` was specific to the `slotchange` event and not just an API on the slot itself. I can change this.

components/html-block/html-block.js
```diff
@@ -201,6 +213,30 @@ class HtmlBlock extends LitElement {
 		return false;
 	}
 
+	async _render(elem) {
```
dbatiste, line 216: I might be inclined to name this differently, since this is specifically for calling the delegated renderers, or whatever we want to call them.  Not sure I have a better name in mind at the moment.  The reason I was thinking of this is because the if-else condition above (for calling this._stamp vs this._renderInline) is repeated in two places, and I was _almost_ going to suggest factoring it out into a render method.

grant-cleary: Yeah, I was musing about that myself. I was a bit torn on the idea of combining them into one function to avoid repetition and make potential later changes a bit more pleasant. I'll think a bit about how to name this a bit better and incorporate that.

components/html-block/html-block.js
```diff
@@ -201,32 +206,60 @@ class HtmlBlock extends LitElement {
...
 			} else {
 				this._renderContainer.innerHTML = '';
 			}
 		};
 
+		const template = this._findSlottedElement('TEMPLATE', slot);
+
 		if (this._templateObserver) this._templateObserver.disconnect();
 		if (template) {
+			this._noDeferredRendering = false;
```
dbatiste, line 262: I think this isn't needed anymore, correct?

grant-cleary: Oops, you're right. Missed this spot.

dbatiste (approved):

PR #1887, opened 2021-11-25
fix: Daylight: use autoOpen in some readme demos

margaree:


components/dropdown/README.md
```diff
@@ -1,7 +1,7 @@
 # Dropdowns
 A Dropdown is a button that opens a floating container to offer menu items or other content.
 
-<!-- docs: demo autoSize:false align:flex-start size:medium -->
+<!-- docs: demo align:flex-start autoSize:false size:medium -->
```
GeorgePlukov, line 4: I noticed this one isn't using `autoOpen` is that intended?

GeorgePlukov (approved):

components/dropdown/README.md
margaree: Yup, this one has a more delayed opening defined in the demo script

PR #1888, opened 2021-11-25
DE43318: Focus not on first item when opening dropdowns

GeorgePlukov:
- makes dropdown menus attempt to focus by default 
- use focus visible for styling of focus (to prevent mouse users seeing focus)
- Mouse users will now have the first item selected by default upon entering the menu (no styles will be applied) but clicking enter/space will activate/click links or buttons

I looked into getting rid of the applyFocus property but I think in cases where someone is using `focus-trap` it could potentially be harmful to force users to always have focus move to the menu

NVDA: 

Regarding styling in NVDA, I was able to move focus to the first item in the list (and it is read out correctly) but there is a bug in NVDA that is preventing their styling from appearing on the selected item (no black outline). Still slightly better accessibility for users. 

Outstanding issue: https://github.com/nvaccess/nvda/issues/10727


components/dropdown/dropdown-menu.js
```diff
@@ -18,6 +18,7 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 
 	constructor() {
 		super();
+		this.__applyFocus = true;
```
dlockhart, line 21: Isn't `__applyFocus` in the base mixin `true` by default anyway?

components/dropdown/dropdown-content-mixin.js
```diff
@@ -239,7 +239,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 		this.noAutoClose = false;
 		this.noAutoFit = false;
-		this.noAutoFocus = false;
+		this.noAutoFocus = true;
```
dlockhart, line 242: I don't quite understand this change... isn't this going to cause dropdowns which used to automatically handle focus (because this wasn't set and was `true` before) to suddenly stop doing that?

components/dropdown/dropdown-menu.js
GeorgePlukov: Should dropdown menu always be used with a component that uses the `opener-mixin`? We have a couple of examples for dropdown that just use `button` and this variable is unset for those elements, I was thinking that could be a viable use case. If we don't want to support that then setting this through the mixin should be enough

dlockhart: Maybe I'm missing something. Searching through `dropdown-menu.js`, this is the only occurrence of `__applyFocus`.  It's defined in `DropdownContentMixin` (which this class extends) but in there it's also set to `true` in its constructor. So I don't see what setting it to `true` in the constructor here is doing that isn't already being done in the mixin constructor.

components/menu/menu.js
```diff
@@ -20,7 +21,7 @@ const keyCodes = {
  * @slot - Menu items
  * @fires d2l-menu-resize - Dispatched when size of menu changes (e.g., when nested menu of a different size is opened)
  */
-class Menu extends ThemeMixin(HierarchicalViewMixin(LitElement)) {
+class Menu extends ThemeMixin(HierarchicalViewMixin(FocusVisiblePolyfillMixin(LitElement))) {
```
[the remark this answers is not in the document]
GeorgePlukov: Yeah I wasn't expecting this to be needed here, however when its removed it causes the menus to not apply the correct styling to the first element when entering a sub menu

components/menu/menu-item-return.js
```diff
@@ -1,11 +1,12 @@
 import '../icons/icon.js';
 import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { FocusVisiblePolyfillMixin } from '../../mixins/focus-visible-polyfill-mixin.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { MenuItemMixin } from './menu-item-mixin.js';
 import { menuItemStyles } from './menu-item-styles.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
-class MenuItemReturn extends RtlMixin(LocalizeCoreElement(MenuItemMixin(LitElement))) {
+class MenuItemReturn extends RtlMixin(LocalizeCoreElement(MenuItemMixin(FocusVisiblePolyfillMixin(LitElement)))) {
```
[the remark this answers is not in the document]
GeorgePlukov: Great idea!

components/dropdown/dropdown-menu.js
GeorgePlukov: Yeah this was dumb, one of the `toggleOpen` calls was passing `undefined`  instead of `true` making me think it wasn't getting set properly for some reason. Changing the toggle call makes this unnecessary

components/filter/test/filter.visual-diff.js
```diff
@@ -194,6 +194,27 @@ describe('d2l-filter', () => {
...
+				'multiple-selected'
+			].forEach(type => {
+				it(type, async function() {
+					const selector = `#${type}`;
+					await openWithKeyboard(page, selector);
+					page.waitForTimeout(50);
+					await page.$eval (selector, async(elem)  => {
+						requestAnimationFrame(async() => await elem.updateComplete);
+					});
+					await visualDiff.screenshotAndCompare(page, `${this.test.fullTitle()}-keyboard-focused`, { captureBeyondViewport: false });
```
dlockhart, line 215: Are these testing anything particular interesting about keyboard focus in filter beyond what dropdown itself is testing?

What I mean by that is if we have good dropdown + keyboard + focus tests, then I don't think we need to re-test them for everything that uses dropdown unless it's a special/custom case.

@svanherk FYI

svanherk: Trying to get caught up on the context of this PR - what will the change to filter focus be?

I was gonna check the visual-diffs to see but I think it's very unhappy because `${this.test.fullTitle()}-keyboard-focused` isn't an id in the html files, and the test names are duplicated.  I think you actually want to wrap the above test in another describe like this:
```
describe('keyboard-focused', () => {
		[
				'single-set-empty',
				'single-set-single-selection',
				'single-set-single-selection-select-all',
				'single-set-multi-selection',
				'single-set-multi-selection-no-search',
				'single-set-multi-selection-no-search-select-all',
				'single-set-multi-selection-all-selected',
				'multiple-empty',
				'multiple-selected'
			].forEach(type => {
				it(type, async function() {
					const selector = `#${type}`;
					await openWithKeyboard(page, selector);
					page.waitForTimeout(50);
					await page.$eval (selector, async(elem)  => {
						requestAnimationFrame(async() => await elem.updateComplete);
					});
					await visualDiff.screenshotAndCompare(page, `${this.test.fullTitle()}`, { captureBeyondViewport: false });
				});
		});
});
```

GeorgePlukov: Hey @svanherk no specific changes regarding filter but there are some focus changes where opening dropdowns using a mouse will now cause the focus to move to within the menu (but with no styles). Seemed like filter hit a bunch of focus use cases since it has focus trap + a bunch of internal components. I can remove the tests if its preferred since they are technically testing other functionality, but I though having some snapshots of what keyboard focus vs mouse focus is could be useful.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -518,12 +518,10 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 			await this.__position();
 			this._showBackdrop = this._useMobileStyling && this.mobileTray;
-
 			if (!this.noAutoFocus && this.__applyFocus) {
 				const focusable = getFirstFocusableDescendant(this);
 				if (focusable) {
-					// bumping this to the next frame is required to prevent Legacy-Edge from crazily invoking click on the focused element
-					requestAnimationFrame(() => focusable.focus());
```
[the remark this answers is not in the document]
GeorgePlukov: So I'm still not exactly sure _why_ this is happening but removing these rAF can cause an infinite focus loop when switching from one focus-trap element to another (specifically reproducible with filter). I was having difficulty reproducing it with the focus-trap options for dropdown content so I'm guessing its something specific to that implementation.

dlockhart (approved):

PR #1892, opened 2021-11-26
fix: listen for change to data-timezone in input-date-time

margaree:
This change is related to the behaviour seen in Activity Feed where occasionally UTC timezone is used with date/time on load. It seems that sometimes (on a fast connection approximately 25% of the time, on wifi could not reproduce) the `input-date-time` is rendered before `data-timezone` is synced to the iframe html element. If something (e.g., opening the time dropdown) triggers a re-render, then the date/time is localized correctly.

This change uses `HtmlAttributeObserverController` to listen for changes to `data-timezone`.

`input-date-time` (and `input-date-time-range`) are the only date/time components that deal with localizing times, so this change is not needed for the others.


components/inputs/test/input-date-time-range.visual-diff.html
```diff
@@ -58,7 +58,7 @@
 				<d2l-input-date-time-range id="start-end-label" start-label="A long start date label explanation" end-label="Finish"></d2l-input-date-time-range>
 			</div>
 			<div class="visual-diff">
-				<d2l-input-date-time-range id="start-end-value" start-value="2020-12-02T15:00:00.000Z" end-value="2021-01-12T08:30:00.000Z"></d2l-input-date-time-range>
+				<d2l-input-date-time-range id="start-end-value" start-value="2020-12-02T15:00:00.000Z" end-value="2021-01-12T07:30:00.000Z"></d2l-input-date-time-range>
```
margaree, line 61: I moved this back an hour so that the visual-diffs could capture a change in date due to timezone for the timezone change test

components/inputs/input-date-time.js
```diff
@@ -16,6 +16,7 @@ import { formatDateInISO,
 import { FormElementMixin } from '../form/form-element-mixin.js';
 import { getDefaultTime } from './input-time.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
+import { HtmlAttributeObserverController } from '../../helpers/htmlAttributeObserverController.js';
```
margaree, line 19: @grant-cleary this seems to be working here and was super easy to use. Let me know if I should actually be using this in a different way.

grant-cleary: Neat, I'm glad! This seems entirely fine. I imagine the component is doing something else to fetch the data from `data-timezone`, hence no need to use the `values` map on the controller. I don't know enough about this component to know if there's potentially some efficiencies to be had using the controller value instead of doing that fetch wherever it happens here, but I don't think there's anything wrong with using it this way. :)

components/inputs/input-date-time.js
```diff
@@ -123,6 +124,8 @@ class InputDateTime extends LabelledMixin(SkeletonMixin(FormElementMixin(Localiz
 		this._namespace = 'components.input-date-time';
 		this._preventDefaultValidation = false;
 		this._timeOpened = false;
+
+		this._contextObserverController = new HtmlAttributeObserverController(this, 'data-timezone');
```
dlockhart, line 128: One disadvantage here is that the fact that our Intl library is using the `data-timezone` attribute is leaking into core -- if we want to change how that works down the road, there'll be downstream changes required.

Poking around, Intl actually does have a way to add change listeners, since it's actually already listening for changes itself:
https://github.com/BrightspaceUI/intl/blob/main/lib/common.js#L126

So I think you should be able to import `getDocumentLocaleSettings` and call into that and accomplish the same thing.

margaree: Ah good call - I'll give that a try.

margaree: Currently the change listeners only get called if language or fallback language get changed so I'll have to update `intl` to also call the change listeners called when timezone changes. Thoughts on if the listeners should just be called whenever any attributes we care about in https://github.com/BrightspaceUI/intl/blob/main/lib/common.js#L153 change?

dlockhart: I think that makes sense yeah. And hopefully given the way `ifrau` works they'll all "change" together with a single callback instead of multiple.

grant-cleary: I think that ends up being true: https://github.com/Brightspace/ifrau/blob/master/port/port.js#L43

ifrau schedules all of the sync tasks in one promise resolution, so I imagine the next event loop will pick up all of those changes at once.

dlockhart (approved):
Looks good!

margaree:
FYI @ronan-f this should fix the issue in activity feed with UTC timezone sometimes being used.

PR #1898, opened 2021-11-29
US133950 - Keep temp elements connected to the DOM during rendering so MathJax can find style properties

grant-cleary:
MathJax allows a global scale property to be attached to its output renderers that renders math at some fixed scale (perhaps unsurprisingly) larger or smaller than the surrounding text. This is configurable within the LMS, but it turns out it hasn't been functional in the component.

MathJax creates a test element within the document and tries to compute its style to figure out all of its base measurements (font-size, ex-height, etc.), which it then multiplies by the scaling factor. However, we use two layers of temp elements that aren't attached to the DOM, so `window.getComputedStyle` returns empty attributes when MathJax does this checking. Hence it ignores that scale factor entirely.

Proposed solution: Within the html-block component, attach the temp element to the render container and don't display it. Then in the math renderer, do something similar: attach our temporary shadowRoot to the existing element.

### DIFF
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -42,7 +42,7 @@
 
 		</script>
 		<script>
-			document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ renderLatex: window.location.search.indexOf('latex=true') !== -1 });
+			document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ outputScale: 1.1, renderLatex: window.location.search.indexOf('latex=true') !== -1 });
 		</script>
 	</head>
 	<body unresolved>
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -245,8 +245,10 @@ class HtmlBlock extends LitElement {
 			if (fragment) {
 
 				let temp = document.createElement('div');
+				temp.style.display = 'none';
 				temp.appendChild(fragment);
 
+				this._renderContainer.appendChild(temp);
 				temp = await this._processRenderers(temp);
 				this._renderContainer.innerHTML = temp.innerHTML;
 
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -32,9 +32,11 @@ export class HtmlBlockMathRenderer {
 		await loadMathJax(mathJaxConfig);
 
 		const temp = document.createElement('div');
+		temp.style.display = 'none';
 		temp.attachShadow({ mode: 'open' });
 		temp.shadowRoot.innerHTML = `<div><mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc></div>`;
 
+		elem.appendChild(temp);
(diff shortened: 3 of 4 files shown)


components/html-block/html-block.js
```diff
@@ -245,8 +245,10 @@ class HtmlBlock extends LitElement {
 			if (fragment) {
 
 				let temp = document.createElement('div');
+				temp.style.display = 'none';
 				temp.appendChild(fragment);
 
+				this._renderContainer.appendChild(temp);
```
grant-cleary, line 251: I have no idea if this is the best solution, so if there's something better I can do here, I'm happy to consider alternatives. My immediate thoughts were that we could check the existing document font size directly or something of the sort, but that becomes problematic because MathJax acts iteratively on the document it's provided... In other words, the font size it computes will still depend on the surrounding HTML, so I don't think we can reasonably "fake" it.

grant-cleary: Note that I _do_ believe this should be reasonably safe. Two lines below we overwrite the render container entirely, so this temp element will get wiped out as soon as processing finishes.

helpers/mathjax.js
```diff
@@ -32,9 +32,11 @@ export class HtmlBlockMathRenderer {
 		await loadMathJax(mathJaxConfig);
 
 		const temp = document.createElement('div');
+		temp.style.display = 'none';
 		temp.attachShadow({ mode: 'open' });
 		temp.shadowRoot.innerHTML = `<div><mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc></div>`;
 
+		elem.appendChild(temp);
```
grant-cleary, line 39: Similar comments RE: the note above. Similarly as well, two lines below we ignore this temp element and return only the child's shadowRoot, which we then use to overwrite `elem` entirely. I think we're safe from any long-term side effects here.

components/html-block/demo/html-block.html
```diff
@@ -42,7 +42,7 @@
 
 		</script>
 		<script>
-			document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ renderLatex: window.location.search.indexOf('latex=true') !== -1 });
+			document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({ outputScale: 1.1, renderLatex: window.location.search.indexOf('latex=true') !== -1 });
```
grant-cleary, line 45: This just makes it very easy to see that scaling is happening - `font-size` will be set on the various rendered math elements if scaling is working correctly.

grant-cleary:
Hmm, the test failure is calling me a liar. I think there might be some weirdness going on if the template is empty and/or replacements don't happen.

grant-cleary:
> Hmm, the test failure is calling me a liar. I think there might be some weirdness going on if the template is empty and/or replacements don't happen.

Turns out this isn't quite true. The problem was the empty replacement test wasn't waiting for replacements to finish and now that we're temporarily adding elements to the DOM, they're surfacing in the test.

dbatiste (approved):

PR #1900, opened 2021-11-29
Revalidate inputs when a custom validator is attached

mpharoah-d2l:
Automatically call `requestValidate` when a custom validator is added to prevent race conditions when attempting to validate based on a custom validator.

### DIFF
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -227,7 +227,7 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 
 	validationCustomConnected(custom) {
 		this._validationCustoms.add(custom);
-		if (validateOnInit) {
+		if (this.validateOnInit) {
 			this.requestValidate(true);
 		}
 	}
--- a/components/form/form-element-mixin.js
+++ b/components/form/form-element-mixin.js
@@ -99,6 +99,11 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 			noValidate: { type: Boolean, attribute: 'novalidate' },
 			/**
 			 * @ignore
+			 * Perform validation immediately instead of waiting for the user to make changes.
+			 */
+			validateOnInit: { type: Boolean, attribute: 'validate-on-init' },
+			/**
+			 * @ignore
 			 */
 			validationError: { type: String, attribute: false },
 			/**
@@ -124,6 +129,8 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		this.invalid = false;
 		/** @ignore */
 		this.noValidate = false;
+		/** @ignore */
+		this.validateOnInit = false;
 		/** @ignore */
 		this.validationError = null;
 		/** @ignore */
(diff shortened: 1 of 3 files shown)


components/form/form-element-mixin.js
```diff
@@ -210,6 +210,7 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 
 	validationCustomConnected(custom) {
 		this._validationCustoms.add(custom);
+		this.requestValidate(true);
```
dlockhart, line 213: Inputs don't validate by default initially though, so this is a breaking change in that any input using a custom validator will now be validating immediately when the page loads vs. those without custom validators will not.

I think a change like this needs to be coupled with a new `validate-when-connected` attribute that when present would automatically call `requestValidate` during `firstUpdated` and also whenever a new custom validator is connected.

mpharoah-d2l:
Still need to add test. Just pushing to test CI

components/form/form-element-mixin.js
```diff
@@ -153,6 +160,13 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this._validity;
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		if (this.validateOnInit) {
+			this.requestValidate(true);
+		}
+	}
```
dlockhart, line 168: Doing this is `firstUpdated` should be redundant if you're also doing it in `updated` (which gets called right after).

components/form/form-element-mixin.js
```diff
@@ -171,6 +185,9 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 				this.dispatchEvent(new CustomEvent('invalid-change'));
 			}
 		}
+		if (this.validateOnInit && (changedProperties.has('noValidate') || changedProperties.has('validateOnInit'))) {
```
dlockhart, line 188: Why does `noValidate` changing affect this?

mpharoah-d2l: So if the `noValidate` attribute is removed later, it will immediately revalidate.

dlockhart: Gotcha.

components/inputs/test/input-number.test.js
```diff
@@ -635,5 +649,23 @@ describe('d2l-input-number', () => {
 			});
 		});
 
+		it('should not validate immediately if validate-on-init is not set', async() => {
+			const elem = await fixture(minMaxWithValueFeature);
+			expect(!elem.hasAttribute('invalid'));
+			await elem.validate();
+			expect(elem.hasAttribute('hasAttribute'));
```
mpharoah-d2l, line 656: Uhhhh... well that's a typo. But... how is this test passing?

mpharoah-d2l: Oh, apparently I need to add `.to.be.true` :/

dlockhart (approved):
This LGTM -- thanks for adding this in!

PR #1901, opened 2021-11-29
POC - Adding subscription controllers

svanherk:
This create a new `ProviderController` and `SubscriberController` that can be used in mixins or components to setup a subscription model.  There are pieces I like about it, and pieces I don't.  I'll add more context throughout.

### DIFF
--- a/components/selection/selection-mixin.js
+++ b/components/selection/selection-mixin.js
@@ -1,3 +1,4 @@
+import { ProviderController } from '../../helpers/subscriptionControllers.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
 const keyCodes = {
@@ -49,17 +50,27 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 	constructor() {
 		super();
 		this.selectionSingle = false;
-		this._selectionObservers = new Map();
-		this._selectionSelectables = new Map();
+
+		this._observerController = new ProviderController(this,
+			{ updateSubscribers: this._updateObservers.bind(this) },
+			{ eventName: 'd2l-selection-observer-subscribe' }
+		);
+
+		this._selectablesController = new ProviderController(this,
+			{ onSubscribe: this._subscribeSelectable.bind(this), onUnsubscribe: this._unsubscribeSelectable.bind(this) },
+			{ eventName: 'd2l-selection-input-subscribe' }
+		);
+
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
+		if (this._observerController) this._observerController.hostConnected();
+		if (this._selectablesController) this._selectablesController.hostConnected();
+
 		if (this.selectionSingle) this.addEventListener('keydown', this._handleRadioKeyDown);
 		if (this.selectionSingle) this.addEventListener('keyup', this._handleRadioKeyUp);
 		this.addEventListener('d2l-selection-change', this._handleSelectionChange);
-		this.addEventListener('d2l-selection-observer-subscribe', this._handleSelectionObserverSubscribe);
-		this.addEventListener('d2l-selection-input-subscribe', this._handleSelectionInputSubscribe);
 		requestAnimationFrame(() => {
 			/** @ignore */
 			this.dispatchEvent(new CustomEvent('d2l-selection-provider-connected', { bubbles: true, composed: true }));
@@ -69,24 +80,33 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 
 	disconnectedCallback() {
 		super.disconnectedCallback();
+		if (this._observerController) this._observerController.hostDisconnected();
+		if (this._selectablesController) this._selectablesController.hostDisconnected();
+
 		if (this.selectionSingle) this.removeEventListener('keydown', this._handleRadioKeyDown);
 		if (this.selectionSingle) this.removeEventListener('keyup', this._handleRadioKeyUp);
 		this.removeEventListener('d2l-selection-change', this._handleSelectionChange);
-		this.removeEventListener('d2l-selection-observer-subscribe', this._handleSelectionObserverSubscribe);
-		this.removeEventListener('d2l-selection-input-subscribe', this._handleSelectionInputSubscribe);
+	}
+
+	getController(controllerId) {
+		if (controllerId === 'observer') {
+			return this._observerController;
+		} else if (controllerId === 'input') {
+			return this._selectablesController;
+		}
 	}
 
 	getSelectionInfo() {
 		let state = SelectionInfo.states.none;
 		const keys = [];
 
-		this._selectionSelectables.forEach(selectable => {
+		this._selectablesController.subscribers.forEach(selectable => {
 			if (selectable.selected) keys.push(selectable.key);
 			if (selectable._indeterminate) state = SelectionInfo.states.some;
 		});
 
 		if (keys.length > 0) {
-			if (keys.length === this._selectionSelectables.size) state = SelectionInfo.states.all;
+			if (keys.length === this._selectablesController.subscribers.size) state = SelectionInfo.states.all;
 			else state = SelectionInfo.states.some;
 		}
 
@@ -96,27 +116,12 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 	setSelectionForAll(selected) {
 		if (this.selectionSingle && selected) return;
 
-		this._selectionSelectables.forEach(selectable => {
+		this._selectablesController.subscribers.forEach(selectable => {
 			if (!!selectable.selected !== selected) {
 				selectable.selected = selected;
 			}
 		});
-		this._updateSelectionObservers();
-	}
-
-	subscribeObserver(target) {
-		if (this._selectionObservers.has(target)) return;
-		this._selectionObservers.set(target, target);
-		this._updateSelectionObservers();
-	}
-
-	unsubscribeObserver(target) {
-		this._selectionObservers.delete(target);
-	}
-
-	unsubscribeSelectable(target) {
-		this._selectionSelectables.delete(target);
-		this._updateSelectionObservers();
+		this._observerController.updateSubscribers();
 	}
 
 	_handleRadioKeyDown(e) {
@@ -133,7 +138,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 		if (!e.composedPath()[0].classList.contains('d2l-selection-input-radio')) return;
 		if (e.keyCode < keyCodes.LEFT || e.keyCode > keyCodes.DOWN) return;
 
-		const selectables = Array.from(this._selectionSelectables.values());
+		const selectables = Array.from(this._selectablesController.subscribers.values());
 		let currentIndex = selectables.findIndex(selectable => selectable.selected);
 		if (currentIndex === -1) currentIndex = 0;
 		let newIndex;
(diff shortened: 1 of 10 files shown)


components/selection/selection-mixin.js
```diff
@@ -69,24 +80,33 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
...
+		if (this._selectablesController) this._selectablesController.hostDisconnected();
+
 		if (this.selectionSingle) this.removeEventListener('keydown', this._handleRadioKeyDown);
 		if (this.selectionSingle) this.removeEventListener('keyup', this._handleRadioKeyUp);
 		this.removeEventListener('d2l-selection-change', this._handleSelectionChange);
-		this.removeEventListener('d2l-selection-observer-subscribe', this._handleSelectionObserverSubscribe);
-		this.removeEventListener('d2l-selection-input-subscribe', this._handleSelectionInputSubscribe);
+	}
+
+	getController(controllerId) {
```
svanherk, line 91: I don't love this, but I couldn't think of another great way to do both:
1. Expose the controller without a forced naming convention AND
2. Have the ability for there to be multiple provider controllers in a single component

helpers/subscriptionControllers.js
```diff
@@ -0,0 +1,140 @@
...
+
+	_handleSubscribe(e) {
+		e.stopPropagation();
+		e.detail.provider = this._host;
+		const target = e.composedPath()[0];
+		this.subscribe(target);
+	}
+}
+
+export class SubscriberController {
```
svanherk, line 56: I don't think this contreoller reads super well on how you can use _either_ the event method or the `for` method because the consumer of the controller can set themselves up for both.  Tried to add some comments throughout

svanherk:
Still trying to work through the labelledby mixin to see if I can use any of this, but it's _really_ specific.  I think I'll open that as a separate PR to determine if it's worth it or not.

helpers/subscriptionControllers.js
```diff
@@ -0,0 +1,140 @@
...
+
+		this._forObserver.observe(this._host.getRootNode(), {
+			childList: true,
+			subtree: true
+		});
+	}
+
+	_updateProviders() {
+		let providerIds = this._host[this._forPropertyName];
+		if (typeof(providerIds) === 'string') providerIds = [providerIds];
```
dlockhart, line 123: Shouldn't it always be a string? If we want to support multiple I'd expect it to still be a string but have the IDs be space-separated.

helpers/subscriptionControllers.js
```diff
@@ -0,0 +1,140 @@
...
+	hostUpdated(changedProperties) {
+		if (!changedProperties.has(this._forPropertyName)) return;
+
+		if (this._forObserver) this._forObserver.disconnect();
+		this._providers.forEach(provider => {
+			provider.getController(this._controllerId).unsubscribe(this._host);
+		});
+		this._providers = new Map();
+
+		this._updateProviders();
```
dlockhart, line 109: I forget if the mutation observer get called once always when connected? If so, I think call could be removed.

helpers/subscriptionControllers.js
```diff
@@ -0,0 +1,140 @@
+import { cssEscape } from '../../helpers/dom.js';
+
+export class ProviderController {
```
dlockhart, line 3: For later: I don't love this name since it overlaps with our `RequesterMixin`/`ProviderMixin` and could cause some confusion.

components/selection/selection-observer-mixin.js
```diff
@@ -15,68 +15,36 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
-
-			const evt = new CustomEvent('d2l-selection-observer-subscribe', {
-				bubbles: true,
-				composed: true,
-				detail: {}
-			});
-			this.dispatchEvent(evt);
-			this._provider = evt.detail.provider;
-		});
+		this._subscriberController.hostConnected();
```
dlockhart, line 34: This is a good sign -- once we go to Lit 2, `connectedCallback`, `disconnectedCallback` and `updated` can be removed.

helpers/subscriptionControllers.js
svanherk: Interesting, I automatically assumed it'd be an array of strings.  But string with spaced ids works too

helpers/subscriptionControllers.js
svanherk: I don't _think_ it does, but I'll need to recheck that

helpers/subscriptionControllers.js
dlockhart: I'm curious what Lit's default attribute processor does for arrays. Like would it turn `<d2l-foo ids="one two three">` into an array if that property had an `Array` type? I'm good with the property being an array as long as that's how the parsing works (which matches what HTML does for things like `aria-labelledby`).

helpers/subscriptionControllers.js
svanherk: Yeah I tried out lots of different names and didn't like most of them.  We could do publisher/subscriber, but that's an actual model that this doesn't seem to _really_ follow.  Sender/Receiver maybe, but that seems more like a 1-to-1 relationship.  Sender/Subscriber?

helpers/subscriptionControllers.js
svanherk: Yeah they're fairly different.  The tricky part is that selection components can use either, depending on whether the `for` property is set, and we don't know which method is being used in the constructor.  But attaching both controllers and then just using one of them sounds cleaner, it's probably not a common use case.  Filter would just use the `ForPropertySubscriberController`

dlockhart: Cool I like that.

helpers/subscriptionControllers.js
dlockhart: I don't mind "provider" but maybe just add more context to it? Similar to `ForAttributeSubscriberController` -- it's a controller for subscribing to the "for" attribute of a host.

helpers/subscriptionControllers.js
```diff
@@ -0,0 +1,140 @@
...
+	get provider() {
+		return this._providers.values().next().value;
+	}
+
+	get providers() {
+		return Array.from(this._providers.values());
+	}
+
+	hostConnected() {
+		if (!this._eventName || this._host[this._forPropertyName]) return;
```
svanherk, line 78: The weirdest part about splitting this controller into two is this line.  Should I add a callback to ask the consumer whether they actually want us to fire the event (or if they ended up getting the `for` attribute set instead), or do I just send the event?  Sending the event when we don't need to means:
1. It goes off into space and no one actions it, and we don't care
2. A provider does hear it and adds us as a subscriber, but we basically ignore that provider. There is a hypothetical case where the provider is trying to update us and we don't want it to, but I can specifically protect against that in the selection components

dlockhart: Hmm, I think this highlights that I didn't really understand what was going on here. I assumed that the code was only needed in the `EventSubscriberController` case, but it looks like additionally if it's using BOTH controllers then it should still not fire the event?

svanherk: It's only needed for the event case, but the selection components are weird.  They basically ask you to provide a `selectionFor` property, and if you don't, they use the event case.  If you do, they use the `ForProperty/Attribute` case.So I only evere need to use one or the other, but when I'm setting up controllers in the `constructor`, I don't yet know if we're going to get a `for` property.  I'm assuming I can't setup controllers (the real Lit ones) in the `connected` callback?

svanherk: I do think the divided works much better, I'll push that change so you can see

helpers/subscriptionControllers.js
svanherk: `SubscriptionProviderController`, `EventSubscriberController`, and `IdSubscriberController`?

helpers/subscriptionControllers.js
svanherk: Yeah confirmed, this is needed.  The `mutationObserver` does not run once when you call `observe`

helpers/subscriptionControllers.js
svanherk: So looks like no - if I set `type` to `Array` and pass in a string, things blow up.  It's trying to run `JSON.parse(<string_I_pass_in>)`.  It looks like I should be able to write a custom converter to take in the string and convert it to an array.  Or we just expect a spaced string and handle it, or just expect an array and handle it.

helpers/subscriptionControllers.js
dlockhart: I think those work, ya! Also wondering if we should start a `controllers` top-level directory... they're similar to mixins in that sense.

helpers/subscriptionControllers.js
dlockhart: Yeah I was just thinking about those custom parsers -- if they're easy to write that could be nice.

components/validation/validation-custom-mixin.js
```diff
@@ -11,60 +11,49 @@ export const ValidationCustomMixin = superclass => class extends superclass {
 
 	constructor() {
 		super();
-		this._forElement = null;
+
+		this._eventSubscriberController = new EventSubscriberController(this, {},
+			{ eventName: 'd2l-validation-custom-connected' }
+		);
```
dlockhart, line 17: Is the event subscriber needed? Seems like the custom validator isn't expecting descendent elements to fire this event.

components/validation/validation-custom-mixin.js
```diff
@@ -11,60 +11,49 @@ export const ValidationCustomMixin = superclass => class extends superclass {
...
+			{ onUnsubscribe: this._onUnsubscribe.bind(this) },
+			{ forProperty: 'for' }
+		);
 	}
 
 	get forElement() {
-		return this._forElement;
+		if (this.for) {
+			// Validation custom components only support one for element
+			return this._forPropertySubscriberController.providers[0];
```
dlockhart, line 28: I'd suggest doing an array length check first and return `null` otherwise. Consumers could have been relying on this being `null` if no element was connected and now accessing this property will throw an array index out of bounds exception.

components/validation/validation-custom-mixin.js
```diff
@@ -11,60 +11,49 @@ export const ValidationCustomMixin = superclass => class extends superclass {
...
-		if (this._forElement !== oldForElement) {
-			if (isCustomFormElement(oldForElement)) {
-				oldForElement.validationCustomDisconnected(this);
-			}
-			if (isCustomFormElement(this._forElement)) {
-				this._forElement.validationCustomConnected(this);
-			}
+	_onUnsubscribe() {
+		if (this._forPropertySubscriberController.provider.length === 0) {
+			throw new Error(`validation-custom failed to find element with id ${this.for}`);
```
dlockhart, line 56: When does this end up getting thrown? Seems like before it would throw during `connectedCallback` -- which arguably could be too soon if there was a delay rendering the element it's looking for.

Having some kind of built-in error throwing capability in the controller itself might be a nice future enhancement too... could simplify the `LabelledBy` code as well as it does a lot of its own error throwing.

components/validation/validation-custom-mixin.js
svanherk: As far as I can tell, the event is used by `FormMixin` and `FormElementMixin` to allow for someone to put a `d2l-validation-custom` element _inside_ of one of them without adding a `for` id.  Like, if `d2l-input-text` wanted to instead just put a `d2l-validation-custom` element inside of itself.  There are no examples of this on the demo pages and I had to hack something together to see it happen.  When we actually implement this, I think it's worth trying to figure out if anyone actually ever uses this, and decide if that's something we'd still want to support.

dlockhart: Whoa! Yeah I'd definitely lean towards just removing that functionality altogether.

components/validation/validation-custom-mixin.js
svanherk: Yeah I didn't test this out before/after but wanted to keep it in to remember - I assume it always "just worked" because people are putting the validator right beside the element.  I was looking at adding a "waitTime" to handle `LabelledBy` anyways, I wonder if I could generalize that to always search for three seconds, and then call an error callback.  I don't think we'd setup an error callback for selection or filter because it's not the end of the world if the element doesn't exist, they'll just handle themselves accordingly?

dlockhart: I like the idea of an error callback a lot. 👍

svanherk:
Closing this - https://github.com/BrightspaceUI/core/pull/3819 is an up-to-date version.

PR #1906, opened 2021-11-30
feat: switch to module type by default

dlockhart:
This switches core to be ESM-by-default.

It also uses the latest `visual-diff` action which upgrades Puppeteer among other things.


svanherk (approved):
🎉

PR #1912, opened 2021-12-01
Add support for compact styles to html-block

grant-cleary:
Add a `compact` option to the HTML block for displaying HTML more compactly. Primarily intended for use cases where space is at a premium. It's possible the styles here may change and/or more styles may be added.

Note that this PR as is doesn't provide support for this functionality if `no-deferred-rendering` is set on the HTML block. That's not to say this can't be done, but I wanted to get this PR up with at least this first piece in place.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -5,6 +5,11 @@ import { HtmlBlockMathRenderer } from '../../helpers/mathjax.js';
 import { requestInstance } from '../../mixins/provider-mixin.js';
 
 export const htmlBlockContentStyles = css`
+	.d2l-html-block-compact {
+		font-size: 0.8rem;
+		font-weight: 400;
+		line-height: 1.2rem;
+	}
 	h1, h2, h3, h4, h5, h6, b, strong, b *, strong * {
 		font-weight: bold;
 	}
@@ -49,7 +54,11 @@ export const htmlBlockContentStyles = css`
 	ul, ol {
 		list-style-position: outside;
 		margin: 1em 0;
-		padding-left: 3em;
+		padding-inline-start: 3em;
+	}
+	.d2l-html-block-compact ul,
+	.d2l-html-block-compact ol {
+		padding-inline-start: 1.5em;
 	}
 	ul, ul[type="disc"] {
 		list-style-type: disc;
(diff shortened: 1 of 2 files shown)


components/html-block/html-block.js
```diff
@@ -5,6 +5,11 @@ import { HtmlBlockMathRenderer } from '../../helpers/mathjax.js';
 import { requestInstance } from '../../mixins/provider-mixin.js';
 
 export const htmlBlockContentStyles = css`
+	.d2l-html-block-compact {
```
dlockhart, line 8: I'd suggest making the `compact` property reflect and then using attribute selectors to apply this CSS instead of needing to add and trigger off a CSS class.

```css
:host([compact]) {
  font-size: 0.8rem;
  /* etc */
}
```

grant-cleary: So the reason this is in the style exports is because it's also going to need to be imported into the HTML editor's edit mode, as is done here: https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor-mixin.js#L546

So I was looking for a solution that didn't require making identical style changes in both locations (keeping it common to the export) while also allowing it to remain static CSS.

grant-cleary: If you think the duplication is preferred, then that's also fine. I have no strong opinions here. The HTML editor does already import its own styles as well; there's just this block that's common to both because we want some of our styling to be consistently reflected in both components.

dlockhart: Gotcha, OK that makes sense! I think it's fine then... the only other option would be something like:

```css
:host([compact]), .d2l-html-block-compact {
  /* etc. */
}
```

So one rule for this component and one rule for others to use. Of course if the others just happened to have a `compact` property weird things could ensue! I'm good with whatever.

grant-cleary: Yeah, that's sort of interesting. It's a bit unfortunate that we have to deal with the editor iframe, otherwise I think it would be much easier to do this sort of stuff. 😞

grant-cleary:
Getting this to work in concert with `no-deferred-rendering` will be tricky. When that attribute is set, we don't render the HTML into the html-block's shadowDOM... Instead we modify the slotted content directly. This is problematic in this case because of course these new styles aren't loaded globally and I'm not sure we want to do that. I chatted about it briefly offline with @dbatiste and I don't think we have a fantastic solution to this problem yet, aside from being sad that we need to support `no-deferred-rendering`. 😅

grant-cleary:
@dbatiste Just to double check, did you have any thoughts/opinions on this?

dbatiste:
> @dbatiste Just to double check, did you have any thoughts/opinions on this?

I am good either way.

dbatiste (approved):

PR #1916, opened 2021-12-02
removing unneeded Edge compatibility thing

dlockhart:
Noticed this when doing some searches -- this must have been in your count badge branch Annabel when we scrubbed it from the repo!


annabelsegalld2l (approved):
👍

PR #1918, opened 2021-12-03
Remove flaky test for already opened dropdown

GeorgePlukov:
Just killing that test for already opened dropdowns, seems like the other ones using the same `#opened` fixture are passing consistently for me (probably since they are editing other attributes).


dlockhart (approved):
👍

PR #1924, opened 2021-12-06
Change performance tests to use web-test-runner-performance package

GeorgePlukov:
Switches performance tests to use `web-test-runner-performance` for test execution.

Adds performance tests for:
* d2l-input-number
* d2l-link
* d2l-button
* d2l-button-icon
* d2l-count-badge
* d2l-icon

Migrated tests for:
* d2l-input-text


components/inputs/test/input-number.perf.js
```diff
@@ -0,0 +1,37 @@
...
+	});
+
+	it('min-max', async() => {
+		const element = html`<d2l-input-number label="label" value="18" min="18" max="100"></d2l-input-number>`;
+		const result = await testRenderTime(element, { iterations: 500 });
+		expect(result.duration).to.be.below(1500);
+	});
+	it('min-max-exclusive', async() => {
+		const element = html`<d2l-input-number label="label" value="5" min="5" max="100" min-exclusive max-exclusive></d2l-input-number>`;
+		const result = await testRenderTime(element, { iterations: 500 });
```
GeorgePlukov, line 21: This test was particularly slow so I increased the timeout to 2500. In this file I actually reduced the amount of iterations to 500 to speed things up. Any thoughts on if we want to keep all iteration values consistant and only change timeouts or some other option?

web-test-runner.config.js
```diff
@@ -23,9 +24,21 @@ export default {
 		},
 		{
 			name: 'perf',
-			files: getPattern('perf')
+			files: getPattern('perf'),
```
GeorgePlukov, line 27: I attempted splitting this config into its own file but I couldn't get it working properly.

web-test-runner.config.js
```diff
@@ -23,9 +24,21 @@ export default {
...
+				async createPage({ context }) {
+					const page = await context.newPage();
+					await page.emulateMedia({ reducedMotion: 'reduce' });
+					return page;
+				}
+			})],
 		}
 	],
+	plugins: [
+		renderPerformancePlugin(),
```
GeorgePlukov, line 40: I thought it would be better to define this within the `perf` section but unfortunately some node commands don't get defined if its there.

components/inputs/test/input-text.perf.js
```diff
@@ -1,24 +1,29 @@
 import '../input-text.js';
-import { expect, html } from '@open-wc/testing';
+import { html, testRenderTime } from 'web-test-runner-performance/browser.js';
```
dlockhart, line 2: Curious what the differences are between the two `html`s? i.e. can we keep using the `@open-wc` one or do we need to switch for this to work?

components/link/test/link.perf.js
```diff
@@ -0,0 +1,13 @@
+import '../link.js';
+import { html, testRenderTime } from 'web-test-runner-performance/browser.js';
+import { expect } from '@open-wc/testing';
+
+describe('d2l-link', () => {
+
+	it('normal', async() => {
+		const element = html`<d2l-button>Button</d2l-button>`;
```
dlockhart, line 8: This link looks like a button :)

components/icons/test/icons.perf.js
```diff
@@ -0,0 +1,13 @@
...
+import { html, testRenderTime } from 'web-test-runner-performance/browser.js';
+import { expect } from '@open-wc/testing';
+
+describe('d2l-icon', () => {
+
+	it('normal', async() => {
+		const element = html`<d2l-icon icon="d2l-tier3:assignments"></d2l-icon>`;
+		const result = await testRenderTime(element, { iterations: 1000 });
+		expect(result.duration).to.be.below(1000);
+	});
```
dlockhart, line 11: Could you add a test for `d2l-icon-custom`? It works completely differently to likely worth testing.

components/button/test/button.perf.js
```diff
@@ -0,0 +1,24 @@
...
+		const element = html`<d2l-button primary>Button</d2l-button>`;
+		const result = await testRenderTime(element, { iterations: 1 });
+		expect(result.duration).to.be.below(1500);
+	});
+
+	it('disabled', async() => {
+		const element = html`<d2l-button disabled>Button</d2l-button>`;
+		const result = await testRenderTime(element, { iterations: 1 });
+		expect(result.duration).to.be.below(1500);
+	});
```
dlockhart, line 23: I'm a little torn about where to draw the line in terms of perf testing all kinds of variations vs. sticking with the variations that we know do things "different" enough to be worth it. The more tests we have, the slower it's going to run and the higher likelihood of flake... but also the better coverage we have and the higher the chances of us catching a real performance regression.

Like for button, we know that `disabled` just puts a `disabled` attribute on the internal `<button>` element... so should have no performance implications.

components/button/test/button.perf.js
```diff
@@ -0,0 +1,24 @@
+import '../button.js';
+import { html, testRenderTime } from 'web-test-runner-performance/browser.js';
+import { expect } from '@open-wc/testing';
+
+describe('d2l-button', () => {
+
+	it('normal', async() => {
+		const element = html`<d2l-button>Button</d2l-button>`;
+		const result = await testRenderTime(element, { iterations: 1 });
```
dlockhart, line 9: 1 iteration likely isn't a good enough sample? Looks like the default is 3 though... so should we just always stick with that? I see you've used 1, 100, 500 and 1000... we should probably pick one and stick to it unless there's a good reason to deviate.

components/inputs/test/input-number.perf.js
```diff
@@ -0,0 +1,37 @@
...
+	});
+	it('integer-only', async() => {
+		const element = html`<d2l-input-number label="label" value="18" max-fraction-digits="0" max="150" min="0"></d2l-input-number>`;
+		const result = await testRenderTime(element, { iterations: 500 });
+		expect(result.duration).to.be.below(1500);
+	});
+	it('after-slot', async() => {
+		const element =
+			html`<d2l-input-number label="Help Text">
+				<d2l-button-icon icon="tier1:help" text="help" slot="after"></d2l-button-icon>
```
dlockhart, line 32: I'd suggest just using a `<span>` or something for the slot tests so as not to leak button-icon performance into input-number. i.e. try to keep these as "unit" test-like as possible.

components/inputs/test/input-number.perf.js
dlockhart: And it looks like it's still failing at 2500, what a slow thing! I wonder if you've uncovered an actual performance problem?

components/button/test/button.perf.js
GeorgePlukov: Think I fat fingered this one when I was switching iteration counts

components/button/test/button.perf.js
GeorgePlukov: Yeah I added these for the first couple but realized they are kinda useless and didnt add them for `d2l-input-number` or `d2l-input-text`. I can just get rid of them, it'll make the test suite slightly faster anyway

components/inputs/test/input-text.perf.js
GeorgePlukov: Sorry meant to leave a comment about this, internally this package [exposes lit](https://github.com/coryrylan/web-test-runner-performance/blob/main/src/browser.ts#L4). I first attempted with the `open-wc` one but I took a look at the tests rendering in non-headless mode and everything was getting rendered to the screen as `[object]`

components/icons/test/icons.perf.js
GeorgePlukov: Done!

dlockhart (approved):
Very cool! Hopefully this proves to be more stable than the thing I hacked together haha.

GeorgePlukov:
Haha yeah.. seems like some of the tests are still being on the flaky side, glad these arent required during CI! We should probably take a look at a few of those input-number options that require significantly longer timeouts (and have tests failing occasionally).

PR #1925, opened 2021-12-07
Drag & Drop for Nested Lists

dbatiste:


components/list/demo/list-drag-and-drop-position.js
```diff
@@ -5,7 +5,8 @@ import { html, LitElement } from 'lit-element/lit-element.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { repeat } from 'lit-html/directives/repeat.js';
 
-class ListDemoDragAndDropUsage extends LitElement {
```
dbatiste, line 8: Just renaming for some consistency.  I will be doing another PR shortly to cleanup the demos and the docs.

components/list/demo/list-drag-and-drop.js
```diff
@@ -0,0 +1,174 @@
...
+		};
+
+		return html`
+			<div @d2l-list-items-move="${this._handleListItemsMove}">
+				${renderList(this.items, false)}
+			</div>
+		`;
+	}
+
+	async _handleListItemsMove(e) {
```
dbatiste, line 114: The consuming app is responsible for taking the item info (items being dragged and the item being dropped upon) and updating their data accordingly.  This can be somewhat complicated, but the logic very much depends on the consumer's data structure, and whether they are doing the move operation client or server-side.  In quizzing case, it is being done server-side.

components/list/list-item-checkbox-mixin.js
```diff
@@ -140,6 +130,10 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(L
 		}
 	}
 
+	_onNestedSlotChange() {
```
dbatiste, line 133: This addresses an issue that can arise when the last selectable item is removed from a nested list.  Without this, selection state gets completely messed up.  This took a long time to figure out.  😬

components/list/list-item-drag-drop-mixin.js
```diff
@@ -1,23 +1,27 @@
...
 import { classMap } from 'lit-html/directives/class-map.js';
 import { dragActions } from './list-item-drag-handle.js';
-import { findComposedAncestor } from '../../helpers/dom.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { nothing } from 'lit-html';
+import { SelectionInfo } from '../selection/selection-mixin.js';
 
-export const dropLocation = Object.freeze({
+export const moveLocations = Object.freeze({
```
dbatiste, line 11: I'd like to transition to "move" verbiage for this, to be a bit more semantically correct.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -263,6 +263,11 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 			 * @type {string}
 			 */
 			dragHandleText: { type: String, attribute: 'drag-handle-text' },
+			/**
+			 * **Drag & drop:** Whether the items can be dropped as nested children
+			 * @type {boolean}
+			 */
+			dropNested: { type: Boolean, attribute: 'drop-nested' },
```
dbatiste, line 270: This new property enables the consumer to indicate which items can become parents by dropping items on them.  This is necessary because not all consumers will want to enable dropping on items to create nested lists, and we also need to ensure that existing lists do not suddenly allow this.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -38,20 +42,16 @@ const isDragSupported = () => {
...
 		this._time = 0;
 	}
 
-	get dragTarget() {
-		return this._dragTarget;
-	}
-
-	get dragTargetKey() {
-		return this._dragTarget && this._dragTarget.key;
+	get dragTargets() {
```
dbatiste, line 53: This change is to track drag targets for the multi drag & drop case.  We are not advertising that feature just yet because we need to make some further tweaks to how to the drag image displays, but the logic is otherwise in place and works.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -351,6 +356,10 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 		super.firstUpdated(changedProperties);
 	}
 
+	activateDragHandle() {
```
dbatiste, line 359: This new method enables the consumer to reactivate the drag handle after they've re-rendered the list/list-items.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 		}));
 	}
 
+	_dispatchListItemsMove(sourceItems, targetItem, moveLocation, keyboardActive) {
```
dbatiste, line 371: This `d2l-list-items-move` is the new event we want consumers to use.  They basically get the items being moved, the target item where they are being moved to, and the location with respect to that item - `above`, `below`, `nest`.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+			this._dispatchListItemsMove([this], listItem, moveLocations.nest, true);
+		}
+	}
+
+	_dispatchMoveListItemNext() {
+		const listItem = this._getNextListItemSibling();
+		if (listItem) {
+			const nestedList = listItem._getNestedList();
+			const items = (nestedList ? nestedList.getItems() : []);
+			if (items.length > 0) {
```
dbatiste, line 412:
```
A
B
    <A moves to this location>
    C
    D
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+				this._dispatchListItemsMove([this], items[0], moveLocations.above, true);
+			} else {
```
dbatiste, line 414:
```html
A
B
< A moves to this location >
C
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+				this._dispatchListItemsMove([this], listItem, moveLocations.below, true);
+			}
+		} else {
+			const parentListItem = this._getParentListItem();
+			if (parentListItem) {
```
dbatiste, line 419:
```html
A
    B
    C
<C moves to this location>
D
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+			}
+		}
+	}
+
+	_dispatchMoveListItemPrevious() {
+		const listItem = this._getPreviousListItemSibling();
+		if (listItem) {
+			const nestedList = listItem._getNestedList();
+			const items = (nestedList ? nestedList.getItems() : []);
+			if (items.length > 0) {
```
dbatiste, line 430:
```html
A
    B
    C
    <D moves to this location>
D
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+				this._dispatchListItemsMove([this], items[items.length - 1], moveLocations.below, true);
+			} else {
```
dbatiste, line 432:
```html
A
<C moves to this location>
B
C
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+				this._dispatchListItemsMove([this], listItem, moveLocations.above, true);
+			}
+		} else {
+			const parentListItem = this._getParentListItem();
+			if (parentListItem) {
```
dbatiste, line 437:
```html
<B moves to this location>
A
    B
    C
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+		this._dispatchListItemsMove([this], items[0], moveLocations.above, true);
+	}
+
+	_dispatchMoveListItemLast(moveToRoot) {
+		const list = (moveToRoot ? this._getRootList() : findComposedAncestor(this, node => node.tagName === 'D2L-LIST'));
+		const items = list.getItems();
+		this._dispatchListItemsMove([this], items[items.length - 1], moveLocations.below, true);
+	}
+
+	_dispatchMoveListItemNest() {
```
dbatiste, line 400: Nesting an item (right arrow):
```html
A
    <B moves to this location>
B
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -359,6 +368,85 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+			}
+		} else {
+			const parentListItem = this._getParentListItem();
+			if (parentListItem) {
+				this._dispatchListItemsMove([this], parentListItem, moveLocations.above, true);
+			}
+		}
+	}
+
+	_dispatchMoveListItemUnnest() {
```
dbatiste, line 443: Unnesting an item (left arrow key):
```html
A
    B
<B moves to this location>
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -380,11 +468,26 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 	}
 
 	_onDragEnd(e) {
+
 		const dragState = getDragState();
 		this.dragging = false;
-		if (dragState.shouldDrop(e.timeStamp)) {
-			this._annoucePositionChange(dragState.dragTargetKey, dragState.dropTargetKey, dragState.dropLocation);
+
+		if (e.dataTransfer.dropEffect !== 'none' && dragState.shouldDrop(e.timeStamp)) {
```
dbatiste, line 475: `dropEffect === 'none'` when the user hits Escape while dragging. (`dragend` event always fires to we need to check this before dispatching our custom move event).

components/list/list-item-drag-drop-mixin.js
```diff
@@ -380,11 +468,26 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+
+			const dropTargetList = findComposedAncestor(dragState.dropTarget, node => node.tagName === 'D2L-LIST');
+			const shouldDispatchPositionChange = !dragState.dragTargets.find(dragTarget => {
```
dbatiste, line 478: This check prevents dropping `A` onto `B` in the following example: 
```html
A
    B
```

components/list/list-item-drag-drop-mixin.js
```diff
@@ -427,13 +546,32 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
-		createDragState(this);
+		// getSelectionInfo(false) is fast so we can quickly check the state
+		if (!rootList.dragMultiple || rootList.getSelectionInfo(false).state === SelectionInfo.states.none) {
+			createDragState([this]);
+		} else {
+
+			// get the seelcted items, but do not include selected items of selected items
+			const getDragTargets = list => {
+				let dragTargets = [];
+				list.getItems().forEach(item => {
```
dbatiste, line 562: In this method, we preserve the order to the items being dragged.  Also worth mentioning, we do not list nested items of items being dragged, as it is already implied that their subtree is being dragged intact (i.e. we do not want the consumer to flatten it).

components/list/list-item-drag-drop-mixin.js
```diff
@@ -461,35 +599,47 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 		dragState.setActiveDropTarget(this, dragState.dropLocation);
 	}
 
-	_onDropTargetBottomDrag(e) {
+	_onDropTargetBottomDragEnter(e) {
```
dbatiste, line 602: Just renaming for consistency with the other methods in this group.

components/list/list-item-drag-handle.js
```diff
@@ -221,15 +232,15 @@ class ListItemDragHandle extends LocalizeCoreElement(LitElement) {
 	}
 
 	_onInactiveKeyboard(e) {
-		if (e.type === 'click' || e.keyCode === keyCodes.ENTER || e.keyCode === keyCodes.SPACE || e.keyCode === keyCodes.LEFT) {
```
dbatiste, line 224: This `LEFT` behaviour needed to be removed because `LEFT` and `RIGHT` are now being used as a way to move items into/out of nested lists.

components/list/list-item-generic-layout.js
```diff
@@ -67,6 +67,17 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				grid-column: content-start / end;
 				grid-row: nested-start / nested-end;
 			}
+			:host(.d2l-dragging-over) ::slotted([slot="nested"]) {
```
dbatiste, line 70: We need to make sure the nested list is stacked on top of the outer list-item's drop target when dragging over, otherwise we can never activate the nested list items' drop targets.

components/list/list.js
```diff
@@ -16,6 +16,11 @@ class List extends SelectionMixin(LitElement) {
 
 	static get properties() {
 		return {
+			/**
+			 * Whether to the user can drag multiple items
+			 * @type {boolean}
+			 */
+			dragMultiple: { type: Boolean, attribute: 'drag-multiple' },
```
dbatiste, line 23: Consumers will need to opt into multi drag & drop since their code will need to handle it.

components/list/list.js
```diff
@@ -91,17 +96,36 @@ class List extends SelectionMixin(LitElement) {
...
+		const slot = this.shadowRoot.querySelector('slot:not([name])');
+		if (!slot) return [];
+		return slot.assignedNodes({ flatten: true }).filter((node) => {
+			return node.nodeType === Node.ELEMENT_NODE && (node.role === 'rowgroup' || node.role === 'listitem');
+		});
+	}
+
+	getListItemByKey(key) {
+		const items = this.getItems();
+		for (let i = 0; i < items.length; i++) {
```
dlockhart, line 109: Nit: could use `Array`'s `find()` method.

components/list/list.js
```diff
@@ -91,17 +96,36 @@ class List extends SelectionMixin(LitElement) {
...
+			if (items[i].key === key) return items[i];
+			if (items[i]._selectionProvider) {
+				const tempItem = items[i]._selectionProvider.getListItemByKey(key);
+				if (tempItem) return tempItem;
+			}
+		}
```
dlockhart, line 115: This will return `undefined` if the item isn't found -- is that what you'd like, or `null`?

components/list/demo/list-drag-and-drop.js
```diff
@@ -0,0 +1,174 @@
...
+
+		const sourceListItems = e.detail.sourceItems;
+		if (!sourceListItems || sourceListItems.length === 0) return;
```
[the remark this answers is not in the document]
dbatiste: Yeah, this is just an over-zealous check.  I don't think it's actually possible for this to not be an array with at least one item.  I'll remove the check.

components/list/demo/list-drag-and-drop.js
```diff
@@ -0,0 +1,174 @@
...
+
+		const target = e.detail.target;
+		if (!target || !target.item) return;
```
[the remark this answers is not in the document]
dbatiste: Same thing here - I think this will always be an element.  Will remove the check.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -1,23 +1,27 @@
...
 	first: 3,
 	last: 4,
 	shiftDown: 5,
 	shiftUp: 6,
+	nest: 7,
 	void: 0
 });
 
-const dropTargetLeaveDelay = 1000; //ms
+export const dropLocation = moveLocations; // backwards compatibility
```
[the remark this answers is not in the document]
dbatiste: 👍  Yeah, I will switch it over once this is merged.

components/list/list-item-mixin.js
```diff
@@ -341,18 +388,18 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 			'd2l-list-item-content-extend-separators': this._extendSeparators,
 			'd2l-focusing': this._focusing,
 			'd2l-hovering': this._hovering,
+			'd2l-dragging-over': this._draggingOver
```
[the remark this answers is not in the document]
dbatiste: Yeah I hear ya. We tried pretty hard to keep things separated, but ultimately some of these things need to be aware of each other.  In this case the `d2l-list-item-generic-layout` needs to know when the user is dragging over so that it can stack the nested list on top of the drop target layer of the parent list-item (otherwise their `dragstart` events will never fire).

I could move this class to a method on the drag & drop mixin - say `getDragDropListItemClasses()`.  That would improve the leakiness slightly.  Thoughts on that?  Either way, the `d2l-list-item-generic-layout` still needs to be aware of the class, or `ListItemDragDropMixin` needs to be aware of `d2l-list-item-generic-layout`. I chose to keep all the stacking rules together, in `d2l-list-item-generic-layout` since its much easier to understand what's going on when they are located in the same place.

components/list/list.js
dbatiste: I think the issue with `find` in this case is that the delegate returns `true/false`, not the item itself.  So I think it would work for a flat list, but not a nested list.  In the nested list case, it would end up returning the root list element that is the ancestor of the nested list item.

components/list/list.js
dbatiste: I don't have a strong opinion either way.  Do you have a preference?

components/list/list-item-mixin.js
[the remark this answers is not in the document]
dlockhart: This is clear enough what's going on, and yeah they're all pretty tightly coupled together anyway. I'm good either way.

components/list/list.js
dlockhart: Ah, gotcha!

components/list/list.js
dlockhart: Not really, I think we're inconsistent. Mostly I like to return _something_ (or throw if we're not expecting someone to ever _not_ get something) just so it's clear that I actually remembered this possibility.

dlockhart (approved):

PR #1927, opened 2021-12-09
US134384 - Add new subscription controllers

svanherk:
Cleaned up version of https://github.com/BrightspaceUI/core/pull/1901.  Creates a new `controllers` folder with new subscriber controllers inside, along with tests and a README.  The README was surprisingly difficult to write and I don't love it - feedback welcome

### DIFF
--- a/controllers/subscriber/subscriberControllers.md
+++ b/controllers/subscriber/subscriberControllers.md
@@ -1,0 +1,175 @@
+# Subscriber Controllers
+
+The `SubscriberRegistryController` and the corresponding `*SubscriberController`s can be used to create a subscription system within your app. Components can setup a subscriber registry instance to keep track of all components subscribed to them with the `SubscriberRegistryController`. Whenever it makes sense to do so, they can iterate over their subscribers to perform some action, update them with new data, etc.  Components can subscribe themselves to different registries using the `IdSubscriberController` or the `EventSubscriberController`. This system supports a many-to-many relationship - registry components can contain multiple registry instances with multiple subscribers in each, and subscriber components can subscribe to multiple different registries.
+
+## Usage
+
+Create an instance of the `SubscriberRegistryController` in the component that will be responsible for providing some data or performing some function on all its subscribers:
+
+```js
+import { SubscriberRegistryController } from '@brightspace-ui/core/controllers/subscriber/subscriberControllers.js';
+
+class InterestingFactSubscription extends LitElement {
+	constructor() {
+		super();
+		this._catFactSubscribers = new SubscriberRegistryController(this,
+			{ onSubscribe: this._sendCatsAreBestWelcome.bind(this), updateSubscribers: this._sendCatFact.bind(this) },
+			{ eventName: 'd2l-interesting-fact-subscribe-cat' }
+		);
+
+		this._dogFactSubscribers = new SubscriberRegistryController(this, {},
+			{ updateSubscribers: this._sendDogFact.bind(this) },
+			{ eventName: 'd2l-interesting-fact-subscribe-dog' }
+		);
+
+		// This controller only supports registering by id - no event is needed
+		this._generalFactSubscribers = new SubscriberRegistryController(this,
+			{ updateSubscribers: this._sendGeneralFact.bind(this) }, {});
+	}
+
+	getController(controllerId) {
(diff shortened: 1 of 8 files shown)


controllers/subscriber/test/subscriberControllers.test.js
```diff
@@ -0,0 +1,516 @@
...
+		}
+	}
+);
+
+const idSubscriber = defineCE(
+	class extends LitElement {
+		static get properties() {
+			return {
+				for: {
+					converter: {
```
svanherk, line 121: So here's an example of the converter property to change it from a string of space-delimited ids.  Pretty clean, but means that any component that supports the list (versus just one id) would need to use this for it's `for` property.  Or we can just always expect a string at https://github.com/BrightspaceUI/core/pull/1927/files#diff-647765ef2f64b4feb979d54c882d976462c39916efc668bc36b1e6c85814cbbdR143, and do the string to array parsing there instead. Thoughts?

controllers/subscriber/subscriberControllers.md
```diff
@@ -0,0 +1,175 @@
...
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		if (this._subscriptionController) this._subscriptionController.hostDisconnected();
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (this._subscriptionController) this._subscriptionController.hostUpdated(changedProperties);
+	}
+```
```
dlockhart, line 153: I found the sample code a little hard to follow in terms of tying what the example was trying to do to the core concepts. 🐶  🐱  It's very cute though and I did learn fun facts about tongue prints!

Maybe a metaphor that also uses a subscription pattern could work? Like a "newspaper" <-> "reader" or a "cable company" <-> "viewer" type metaphor? Do people still even subscribe to newspapers?!

controllers/subscriber/test/subscriberControllers.test.js
svanherk: Yeah, or we can just always expect a string and do the array conversion for them in the controller.  That's _probably_ easier...

controllers/subscriber/subscriberControllers.md
```diff
@@ -0,0 +1,175 @@
+# Subscriber Controllers
```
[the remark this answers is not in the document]
svanherk: Good call, did all the development with it in the `helpers` folder and followed the convention there

controllers/subscriber/test/subscriberControllers.test.js
dlockhart: I'm good with that 👍

controllers/subscriber/subscriberControllers.md
svanherk: Yeah I adapted this from the Provider/Requester README, but using a more intuitive subscription example makes sense.  I'll try it with cable channels and viewers

controllers/subscriber/test/subscriberControllers.test.js
```diff
@@ -0,0 +1,517 @@
...
+
+			expect(registry._onSubscribeTargets.length).to.equal(2);
+			expect(registry._onSubscribeTargets[0]).to.equal(elem.querySelector('#id'));
+			expect(registry._onSubscribeTargets[1]).to.equal(elem.querySelector('#event'));
+		});
+
+		it('Additional subscribers can be subscribed manually', () => {
+			expect(registry._eventSubscribers.subscribers.size).to.equal(1);
+			expect(registry._idSubscribers.subscribers.size).to.equal(1);
+			expect(registry._onSubscribeTargets.length).to.equal(2);
```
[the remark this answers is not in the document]
svanherk: Yeah, I like making the tests "readable" but it's definitely redundant.  Can cleanup a bit

controllers/subscriber/subscriberControllers.md
svanherk: Ok, updated - let me know if that works better. It's still a bit of a weird example when looking at the resulting html, but I wanted to show how event subscribers must go inside the registry component, and id subscribers can go outside.

dlockhart: I like this a lot! 👍

dlockhart (approved):
This is looking great to me!

PR #1934, opened 2021-12-11
US134385 - Use the SubscriberRegistryController in filter

svanherk:
Add the new `SubscriberRegistryController` directly into `d2l-filter` to be able to register subscribers interested in getting updates on which filters are active.  The only consumer of this for now will be a new version of the applied filters lab component, but eventually that will be fixed up and moved into `core` once we have an official `tag` component.  PR for that is here: https://github.com/BrightspaceUILabs/facet-filter-sort/pull/93

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -23,6 +23,7 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { SubscriberRegistryController } from '../../controllers/subscriber/subscriberControllers.js';
 
 const ARROWLEFT_KEY_CODE = 37;
 const ESCAPE_KEY_CODE = 27;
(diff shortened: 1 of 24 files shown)


svanherk:
Will need to use Sauce to figure out why Safari is unhappy when I get back.

components/filter/filter.js
```diff
@@ -716,6 +744,37 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+	_updateActiveFilters() {
+		const activeFilters = [];
+
+		this._dimensions.forEach(dimension => {
+			switch (dimension.type) {
+				case 'd2l-filter-dimension-set': {
+					dimension.values.forEach(value => {
+						if (value.selected) {
+							const keyObject = { dimension: dimension.key, value: value.key };
+							const text = dimension.valueOnlyActiveFilterText ? value.text : this.localize('components.filter.activeFilterText', { filterName: dimension.text, activeFilter: value.text });
```
dlockhart, line 756: I'm not totally convinced that this needs to be localized, since it's just one thing before the other separated by a colon. My worry is just that it's one more place where translators could mess it up... and I can't think of any language-specific variances that there would be anyway.

dlockhart (approved):
Looks good to me -- I know you were planning on investigating the Safari Sauce error when you got back.

components/filter/filter.js
svanherk: Yeah I agree, we can always add a langterm later if it gets more complicated but for now it's only asking for trouble

components/filter/filter.js
```diff
@@ -217,6 +234,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		if (opener) opener.focus();
 	}
 
+	getController() {
```
svanherk, line 237: @dlockhart I think I'm going to change this to be `getSubscriberController` instead to make it clearer and avoid name collisions 🤔 .  No one is using the controller yet, so easier to do it now

dlockhart: Yep sounds good!

dlockhart (approved):
LGTM!

PR #1935, opened 2021-12-13
upgrading to use NPM 8 and Node 16

dlockhart:
DRAFT: just testing this out


.github/workflows/ci.yml
```diff
@@ -9,7 +9,6 @@ jobs:
       - uses: Brightspace/third-party-actions@actions/checkout
       - uses: Brightspace/third-party-actions@actions/setup-node
         with:
-          node-version: '14'
```
devpow112, line 12: Know this is a draft/trying it out but just FYI you'll want `node-version-file: .nvmrc` to pick up the `nvmrc` file. This is working to pick up node 16 since the default on runners is now node 16

dlockhart: Ah gotcha -- thanks!

PR #1937, opened 2021-12-13
removing mocha dependency lock

dlockhart:
I don't really remember why we were locking Mocha to version 8, but 1) core doesn't have an explicit dependency on Mocha anymore and b) we're now successfully using Mocha 9 for visual-diff testing so any issues should be resolved.


annabelsegalld2l (approved):
LGTM! Definitely don't need this if we don't have the dependency on mocha anymore.

PR #1939, opened 2021-12-13
Move HTML attribute observer controller to new controllers directory

grant-cleary:
We now have a controllers directory, so it makes more sense for this to live there. No README currently; I want to get the move done early because I need to use this in the HTML editor and would prefer not to have to play the "race BSI when I rename things" game. 😅


components/html-block/html-block.js
```diff
@@ -1,6 +1,6 @@
 import '../colors/colors.js';
 import { css, LitElement } from 'lit-element/lit-element.js';
-import { HtmlAttributeObserverController } from '../../helpers/htmlAttributeObserverController.js';
+import { HtmlAttributeObserverController } from '../../controllers/attributeObserver/htmlAttributeObserverController.js';
```
grant-cleary, line 3: `attributeObserver` is fairly generic, but there were some suggestions that we could extend this controller (or create a new one) to handle the more generic case of observing attributes on an arbitrary element. As such, I figure it might be good to leave room for that.

dlockhart (approved):
Haha good luck with the race! If it proves really difficult, you could temporarily have a stub file in the old location that imports and exports the new thing.

grant-cleary:
> Haha good luck with the race! If it proves really difficult, you could temporarily have a stub file in the old location that imports and exports the new thing.

Thankfully I haven't merged the PR where I implemented this quite yet, so I should be safe there. These come up so often that I've learned to be much more careful about my order of operations. 😅

dlockhart (approved):

PR #1941, opened 2021-12-13
explicitly ignoring dependabot updates

dlockhart:
While reviewing this I noticed we had ignored these upgrades through the PR workflow but should probably list them out explicitly here, along with the reason.

I'm going to make this change to the `npm init` generator too, otherwise new repos will suddenly get Dependabot PRs to upgrade Lit.


annabelsegalld2l (approved):
👍

PR #1943, opened 2021-12-16
animate fixes

dlockhart:
Made some tweaks here that seem to be working for me.

I'll add some comments inline!

### DIFF
--- a/components/demo/code-view.js
+++ b/components/demo/code-view.js
@@ -36,7 +36,9 @@ class CodeView extends LitElement {
 			const path = `/node_modules/prismjs/components/prism-${language}.min.js`;
 			this._dependenciesPromise = import(path);
 		}
-		this._updateCode(this.shadowRoot.querySelector('slot'));
+		if (this.shadowRoot) {
+			this._updateCode(this.shadowRoot.querySelector('slot'));
+		}
 		super.attributeChangedCallback(name, oldval, newval);
 	}
 
--- a/directives/animate/animate.js
+++ b/directives/animate/animate.js
@@ -1,8 +1,8 @@
-import { directive, PropertyPart } from 'lit/directive.js';
+import { directive, PartType } from 'lit-html/directive.js';
 import { getComposedActiveElement, getNextFocusable } from '../../helpers/focus.js';
-import { AsyncDirective } from 'lit/async-directive.js';
+import { AsyncDirective } from 'lit-html/async-directive.js';
 import { isComposedAncestor } from '../../helpers/dom.js';
-import { noChange } from 'lit';
+import { noChange } from 'lit-html';
 
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 const showTransitionDuration = 300;
--- a/directives/animate/animate.js
+++ b/directives/animate/animate.js
@@ -4,6 +4,7 @@ import { AsyncDirective } from 'lit-html/async-directive.js';
 import { isComposedAncestor } from '../../helpers/dom.js';
 import { noChange } from 'lit-html';
 
+const stateMap = new WeakMap();
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 const showTransitionDuration = 300;
 const hideTransitionDuration = 200;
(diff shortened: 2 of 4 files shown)


components/demo/code-view.js
```diff
@@ -36,7 +36,9 @@ class CodeView extends LitElement {
 			const path = `/node_modules/prismjs/components/prism-${language}.min.js`;
 			this._dependenciesPromise = import(path);
 		}
-		this._updateCode(this.shadowRoot.querySelector('slot'));
+		if (this.shadowRoot) {
+			this._updateCode(this.shadowRoot.querySelector('slot'));
+		}
```
dlockhart, line 41: This was unrelated to the animate stuff, but noticed a JS error on the animate demo page -- it's accessing `shadowRoot` before things have rendered.

I think it'd be safer to do this work in `updated()` anyway, so might be worth looking into that.

directives/animate/animate.js
```diff
@@ -1,8 +1,8 @@
-import { directive, PropertyPart } from 'lit/directive.js';
+import { directive, PartType } from 'lit-html/directive.js';
```
dlockhart, line 1: The first error I was getting as that `PropertyPart` wasn't importing at all. The class/type itself didn't end up being needed since the new `PartType` info object has that stuff on it anyway.

directives/animate/animate.js
```diff
@@ -1,8 +1,8 @@
...
 import { getComposedActiveElement, getNextFocusable } from '../../helpers/focus.js';
-import { AsyncDirective } from 'lit/async-directive.js';
+import { AsyncDirective } from 'lit-html/async-directive.js';
 import { isComposedAncestor } from '../../helpers/dom.js';
-import { noChange } from 'lit';
+import { noChange } from 'lit-html';
```
dlockhart, line 5: I switched this over to pull everything from `lit-html` or `lit-element`. The `lit` module is currently just wrapping those two. I think what we'll want to do is switch everything in core over at once to reference `lit`.

directives/animate/animate.js
```diff
@@ -291,50 +289,32 @@ class AnimationState {
...
-		}
+	render() {
 		return noChange;
 	}
 	update(part, [opts]) {
-		/* Any imperative updates to DOM/parts would go here */
 		opts = opts || {};
 		this.state.show(opts);
-		return this.render(part);
+		return this.render(opts);
```
dlockhart, line 317: This change seemed to fix things. I don't even know if we need the `render()` at all actually.

dlockhart (commented):
So this is working with Lit 2 (I think). What I'd like to try next is to switch it back to Lit 1 and make sure the same code is still working -- then we can just merge this today.

directives/animate/animate.js
```diff
@@ -291,50 +289,32 @@ class AnimationState {
...
-		}
+	render() {
 		return noChange;
 	}
-	update(part, opts) {
-		/* Any imperative updates to DOM/parts would go here */
+	update(part, [opts]) {
 		opts = opts || {};
 		this.state.hide(opts);
 		return this.render(part);
```
annabelsegalld2l, line 302: Wondering why this one didn't need to be changed to opts as well? I would think the Hide and Show would be identical with the exception of this.state.hide/show function calls.

directives/animate/animate.js
annabelsegalld2l: Nice. Yeah I was stuck on this one for a while since it just didn't import. Lucky that we can swap in the PartType here.

directives/animate/animate.js
dlockhart: Sorry that was just a mistake on my "part" haha. Render should always take `opts` and `update()` gets passed both the part and the parameters. In our case though `render()` doesn't do anything at all, so that's why this isn't blowing up.

package.json
```diff
@@ -67,9 +67,8 @@
     "focus-visible": "^5",
     "ifrau": "^0.39",
     "intl-messageformat": "^7",
-    "lit": "^2.0.2",
-    "lit-element": "^3.0.2",
-    "lit-html": "^2.0.2",
+    "lit-element": "^2",
+    "lit-html": "^1",
```
dlockhart, line 71: Switched this back to Lit 1 so we could merge this right away.

directives/animate/animate.js
```diff
@@ -290,52 +289,37 @@ class AnimationState {
...
+	render() {
 		return noChange;
 	}
-	update(part, opts) {
-		/* Any imperative updates to DOM/parts would go here */
+	update(part, [opts]) {
 		opts = opts || {};
-		this.state.hide(opts);
-		return this.render(part);
+		let state = stateMap.get(part.element);
```
annabelsegalld2l, line 298: Right, so instead of the class property we now use the stateMap to store the state.

annabelsegalld2l (approved):
Thanks for getting this working 😄

PR #1944, opened 2021-12-16
feat: US132778 - Lit 2 Prep - Animate Directive

annabelsegalld2l:
This PR moves to the Lit 2 -style syntax for directives, which use a class wrapper rather than helper functions.  

For more context, see https://lit.dev/docs/releases/upgrade/#overview-of-directive-api-changes


directives/animate/animate.js
```diff
@@ -10,15 +12,13 @@ const moveYValue = 20;
 
 class AnimationState {
 
-	constructor(propertyPart) {
-
-		if (!(propertyPart instanceof PropertyPart) || propertyPart.committer.name !== 'animate') {
-			throw new Error('animation directives must be used with "animate" property');
+	constructor(partInfo) {
+		if (!(partInfo.type === PartType.PROPERTY) || partInfo.name !== 'animate') {
```
annabelsegalld2l, line 16: In preparation for Lit 2, we swapped out PropertyPart for PartType, and take in a partInfo instead of a propertyPart. They have very similar fields, except partInfo does not have the committer field and you can access the properties directly.

directives/animate/animate.js
```diff
@@ -289,20 +289,37 @@ class AnimationState {
...
+			state = new AnimationState(part);
+			stateMap.set(part.element, state);
+		}
 		state.hide(opts);
+		return this.render();
 	}
 }
 
-export const hide = directive((opts) => async(part) => helper(part, 'hide', opts));
+class Show extends AsyncDirective {
```
annabelsegalld2l, line 308: Show and Hide are very similar classes, just differing on the state.show or state.hide calls. The state is stored in a WeakMap and gets queried/updated by both classes.

directives/animate/animate.js
dlockhart: Once we're 100% on Lit 2, I think we can relax this check a bit. In Lit 2 you can just go `<div ${ hide() }>` and not tie it to a property at all, which is pretty sweet!

dlockhart (approved):
Awesome!

PR #1945, opened 2021-12-17
US134551 - Lit 2 Prep - Add shadowRoot checks

annabelsegalld2l:
This PR adds a guardrail to all of our shadowRoot calls to actually check it exists before querying the shadowRoot. It will not be guaranteed that the shadowRoot is in place if a method is called very early (before firstUpdated). As a precaution, I've added these checks to all the shadowRoot accesses, with the exception of those in the updated()/firstUpdated() calls. It may be a bit overkill, but it will not affect most of the code.  

Another PR will be coming after this one for more Lit 2 prep changes. Figured this PR was big enough on its own 😆

### DIFF
--- a/components/button/button-mixin.js
+++ b/components/button/button-mixin.js
@@ -108,7 +108,7 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 	}
 
 	focus() {
-		const button = this.shadowRoot.querySelector('button');
+		const button = this.shadowRoot ? this.shadowRoot.querySelector('button') : undefined;
 		if (button) button.focus();
 	}
 
--- a/components/button/button-mixin.js
+++ b/components/button/button-mixin.js
@@ -108,7 +108,7 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 	}
 
 	focus() {
-		const button = this.shadowRoot ? this.shadowRoot.querySelector('button') : undefined;
+		const button = this.shadowRoot && this.shadowRoot.querySelector('button');
 		if (button) button.focus();
 	}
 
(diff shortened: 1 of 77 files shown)


components/form/form-element-mixin.js
```diff
@@ -160,6 +157,18 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		return this._validity;
 	}
 
+	connectedCallback() {
```
annabelsegalld2l, line 160: Just calling attention to the only differing check here. The form mixin was adding the shadowRoot listeners in its constructor, which is valid in Lit 1 but not Lit 2. Moving the listeners to the connectedCallback fixes the issue with no functionality changes.

components/button/button-mixin.js
```diff
@@ -108,7 +108,7 @@ export const ButtonMixin = superclass => class extends FocusVisiblePolyfillMixin
 	}
 
 	focus() {
-		const button = this.shadowRoot.querySelector('button');
+		const button = this.shadowRoot ? this.shadowRoot.querySelector('button') : undefined;
```
annabelsegalld2l, line 111: If Dave L can get optional chaining to be allowed in our Polymer builds, it will be very easy to move over to the ?. chaining syntax. Until then we are stuck with these ugly checks 😢 . I don't think we will have these in place for most of the new code written, as we can better audit to know when exactly the code is called.

GeorgePlukov:
```suggestion
		const button = this.shadowRoot && this.shadowRoot.querySelector('button');
```
Any chance you tried this? Its a bit smaller and I _think_ it should work the same

GeorgePlukov: Sorry wasn't implying for you to go change all these, just suggesting a potentially alternate for the future. I try and avoid `?:` at all costs

annabelsegalld2l: Good catch, I will try that! Even if it's just a bit shorter it will help clean things up

dlockhart (approved):
Giving up on getting optional chaining working with Polymer, so unfortunately this is what we'll need to go with for now.

PR #1953, opened 2021-12-21
Make sure MathJax loads even if only a single non-deferred html-block is loaded on the page

grant-cleary:
The problem with not bothering to render math at all when `noDeferredRendering` is set is that we're still reliant on MathJax's global rendering in that case... But if we don't run the renderer we never load MathJax!

So I'm not tied to this solution necessarily, but it feels like the cleanest... Pass `noDeferredRendering` to the renderers as an option and let them figure out how they should support it. The math renderer can still load MathJax if needed and then bail before doing any inline replacements. Other renderers can, in theory, do something entirely different.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -225,14 +225,17 @@ class HtmlBlock extends LitElement {
 
 	async _processRenderers(elem) {
 		for (const renderer of getRenderers()) {
-			if (this.noDeferredRendering && !renderer.canRenderInline) continue;
-
 			if (this._contextObserverController && renderer.contextAttributes) {
 				const contextValues = new Map();
 				renderer.contextAttributes.forEach(attr => contextValues.set(attr, this._contextObserverController.values.get(attr)));
-				elem = await renderer.render(elem, contextValues);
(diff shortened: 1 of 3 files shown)


components/html-block/html-block.js
```diff
@@ -225,14 +225,17 @@ class HtmlBlock extends LitElement {
...
 		for (const renderer of getRenderers()) {
-			if (this.noDeferredRendering && !renderer.canRenderInline) continue;
-
 			if (this._contextObserverController && renderer.contextAttributes) {
 				const contextValues = new Map();
 				renderer.contextAttributes.forEach(attr => contextValues.set(attr, this._contextObserverController.values.get(attr)));
-				elem = await renderer.render(elem, contextValues);
+				elem = await renderer.render(elem, {
+					contextValues: contextValues,
+					noDeferredRendering: this.noDeferredRendering
```
grant-cleary, line 233: This will break the user profile renderer again if it merges as is, so if we're okay with this, I'll commit something there first that's backwards compatible with this change. May as well play it safe since we're in no huge rush for this.

dbatiste (approved):

PR #1955, opened 2021-12-21
Keyboard Instructions for List Nav and Reordering

dbatiste:
This PR introduces two tooltips providing keyboard instructions for navigating lists in `grid` mode, as well as for reordering/moving items within lists.  The tooltips look like:

image

image


components/list/list-item-mixin.js
```diff
@@ -446,6 +476,7 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 				<div class="d2l-list-item-active-border"></div>
 			</div>
 			${this._renderBottomPlacementMarker(html`<d2l-list-item-placement-marker></d2l-list-item-placement-marker>`)}
+			${this._displayKeyboardTooltip && tooltipForId ? html`<d2l-tooltip align="start" announced for="${tooltipForId}" for-type="descriptor">${this._renderTooltipContent()}</d2l-tooltip>` : ''}
```
dbatiste, line 479: We use the `announced` attribute here since this tooltip could actually be pointing to the selection checkbox: `d2l-selection-input`, which contains `d2l-input-checkbox`, which contains `input`. This ensure that the tooltip gets announced even though the input being focused in in a different DOM scope.

components/list/list-item-role-mixin.js
```diff
@@ -19,8 +19,9 @@ export const ListItemRoleMixin = superclass => class extends superclass {
 
 		const separators = parent.getAttribute('separators');
 
-		this.role = parent.grid ? 'rowgroup' : 'listitem';
+		this.role = parent.hasAttribute('grid') ? 'rowgroup' : 'listitem';
```
dbatiste, line 22: This was actually randomly broken it seemed.  Like the `grid` property wasn't always set on the parent in time.

components/list/list-item-drag-handle.js
```diff
@@ -266,17 +288,32 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	_renderKeyboardDragging() {
 		return html`
 			<button
+				aria-label="${this._defaultLabel}"
```
dbatiste, line 291: This is mostly reordering.  The new things here are `id` and `@focusin`.

components/list/list-item-mixin.js
```diff
@@ -351,6 +376,9 @@ export const ListItemMixin = superclass => class extends ListItemDragDropMixin(L
 
 	_onFocusIn() {
 		this._focusing = true;
+		if (this.role !== 'rowgroup' || !tabPressed || hasDisplayedKeyboardTooltip) return;
```
dbatiste, line 379: Only show the tooltip if the list is in `grid` mode, and focus is the result of tabbing into the list, and the tooltip has not already been displayed.

components/list/list-item-drag-handle.js
```diff
@@ -266,17 +288,32 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
+				@focusout="${this._onFocusOutKeyboardButton}"
+				id="${this._buttonId}"
 				@keydown="${this._onPreventDefault}"
-				aria-live="assertive"
-				aria-label="${this._defaultLabel}">
+				@keyup="${this._onActiveKeyboard}">
 				<d2l-icon icon="tier1:arrow-toggle-up" @click="${this._dispatchActionUp}" class="d2l-button-icon"></d2l-icon>
 				<d2l-icon icon="tier1:arrow-toggle-down" @click="${this._dispatchActionDown}" class="d2l-button-icon"></d2l-icon>
 			</button>
+			${this._displayKeyboardTooltip ? html`<d2l-tooltip align="start" for="${this._buttonId}" for-type="descriptor">${this._renderTooltipContent()}</d2l-tooltip>` : ''}
```
dlockhart, line 302: Missing an import for tooltip in this file.

components/list/list-item-mixin.js
```diff
@@ -10,10 +11,23 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
...
 
+let tabPressed = false;
+document.addEventListener('keydown', e => {
+	if (e.keyCode !== 9) return;
+	tabPressed = true;
+});
+document.addEventListener('keyup', e => {
+	if (e.keyCode !== 9) return;
+	tabPressed = false;
+});
```
dlockhart, line 27: I don't think it's a big deal, but these event listeners will get added as soon as this file is included in the page, regardless of whether the page actually has a list or list in grid mode.

The only optimization I can think would be to run this code from the `connectedCallback` and gate it so that it only runs once -- it'd still run for non-grid mode but it's a slight improvement. Something like:

```javascript
let tabPressed = false;
let tabListenerAdded = false;
function addTabListener() {
	if (tabListenerAdded) return;
	tabListenerAdded = true;
	document.addEventListener('keydown', e => {
		if (e.keyCode !== 9) return;
		tabPressed = true;
	});
	document.addEventListener('keyup', e => {
		if (e.keyCode !== 9) return;
		tabPressed = false;
	});
}
```

dbatiste: That's a good point.  I will make that improvement.

dlockhart (approved):

PR #1959, opened 2021-12-22
Ignore broken analyzer warning and fix other easy warnings

svanherk:
This PR ignores the `no-unknown-tag-name` warnings because we know these are broken ([more info here](https://github.com/runem/lit-analyzer/issues/34)).  It also tries to fix a number of real warnings, getting our total down from 99 to 66.

These are the remaining warnings we'd need to fix to get rid of all analyzer warnings:
- `animate` directive:
  - Unknown property `animate` (will go away with Lit 2)
- `FormElementMixin`:
  - `forceInvalid` is ignored
  - `invalid` is ignored
  - `noValidate` is ignored
- `d2l-input-date`:
  - `has-now` is ignored
  - `novalidateminmax` is ignored
- `d2l-input-number`:
  - `trailing-zeroes` is ignored
  - `value-trailing-zeroes` is ignored
- `d2l-selection-input`:
  - `hovering` is ignored
  - `_indeterminate` is ignored
- `d2l-table`:
  - Placing `selected` on `tr`s
  - Placing `sticky` `th`s
  - Placing `no-column-border` on `table`

These values are all hidden for a reason so we would need to develop a new way of hiding them from the Daylight site, but not the analyzer.  And switching the table over to classes or data attributes may be time consuming and tricky.

### DIFF
--- a/components/dialog/test/dialog-confirm.test.js
+++ b/components/dialog/test/dialog-confirm.test.js
@@ -1,3 +1,4 @@
+import '../../button/button.js';
 import '../dialog-confirm.js';
 import { expect, fixture, oneEvent } from '@open-wc/testing';
 import { getComposedActiveElement } from '../../../helpers/focus.js';
@@ -19,8 +20,8 @@ describe('d2l-dialog-confirm', () => {
 		it('should focus on first non-primary button', async() => {
 			const el = await fixture(html`
 				<d2l-dialog-confirm opened>
-					<button slot="footer" primary>Yes</button>
-					<button slot="footer">No</button>
+					<d2l-button slot="footer" primary>Yes</d2l-button>
+					<d2l-button slot="footer">No</d2l-button>
 				</d2l-dialog-confirm>
 			`);
 			await oneEvent(el, 'd2l-dialog-open');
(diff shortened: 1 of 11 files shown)


components/dialog/test/dialog-confirm.test.js
```diff
@@ -19,8 +20,8 @@ describe('d2l-dialog-confirm', () => {
 		it('should focus on first non-primary button', async() => {
 			const el = await fixture(html`
 				<d2l-dialog-confirm opened>
-					<button slot="footer" primary>Yes</button>
-					<button slot="footer">No</button>
+					<d2l-button slot="footer" primary>Yes</d2l-button>
```
svanherk, line 23: This test is skipped anyways, but the analyzer wasn't happen with putting `primary` on a native button so I've just switched them to `d2l-button`.

dlockhart (approved):
Nice! Thanks for fixing these Stacey, looks like this found some actual issues in our tests and holes in the documentation.

components/form/form-element-mixin.js
```diff
@@ -88,7 +88,6 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 			 */
 			invalid: { type: Boolean, reflect: true },
 			/**
-			 * @ignore
```
GeorgePlukov, line 91: Thanks for removing this!

PR #1960, opened 2021-12-22
Don't use templates as html-block content container

grant-cleary:
~I haven't tested this yet, so it's a draft only for now, but this might be a decent path forward for transitioning to using `div`s for the html-block's content container instead of templates.~

The use of templates within the `html-block` component becomes problematic for the Lit 2 migration, since Lit 2 no longer supports rendering dynamic HTML within a template. For example:
```
return html`
   <d2l-html-block>
      <template>
        ${returnsDynamicHtml()}
      </template>
   <d2l-htmlblock>
`;
```
... Will now generate an error, provided `returnsDynamicHtml()` actually does return dynamic HTML. We wish to support this case, which means we need to jettison the use of `<template>` as our html-block container. Instead, we'll allow consumers to just provide the HTML directly into the block's slot, no container required.


helpers/mathjax.js
```diff
@@ -19,6 +19,7 @@ export class HtmlBlockMathRenderer {
 		if (!elem.querySelector('math') && !(isLatexSupported && /\$\$|\\\(|\\\[|\\begin{|\\ref{|\\eqref{/.test(elem.innerHTML))) return elem;
 
 		const mathJaxConfig = {
+			deferTypeset: true,
```
grant-cleary, line 22: This needs to be a config because we'll for sure want to do initial typeset when we load MathJax into content documents or the equation editor's dialog.

helpers/mathjax.js
```diff
@@ -27,7 +28,10 @@ export class HtmlBlockMathRenderer {
 
 		// If we're opting out of deferred rendering, we need to rely
 		// on the global MathJax install for rendering.
-		if (options.noDeferredRendering) return elem;
+		if (options.noDeferredRendering) {
+			window.MathJax.typeset([elem]);
```
grant-cleary, line 32: We can't rely on initial typesetting any longer, so we need to manually typeset. This might be worth wrapping in a try-catch because things may be weird if `elem` is directly in a component's shadowDOM.

helpers/mathjax.js
```diff
@@ -138,15 +142,17 @@ export function loadMathJax(mathJaxConfig) {
 					const InputJax = startup.getInputJax();
 					const OutputJax = startup.getOutputJax();
 					const html = mathjax.document(root, { InputJax, OutputJax });
-					html.render();
+					html.render().typeset();
```
grant-cleary, line 145: When startup typesetting is disabled, we need to explicitly invoke the typeset action on the document.

helpers/mathjax.js
grant-cleary: Note that this will likely need to be changed to:
```
window.MathJax.typeset([elem]);
```
While it's unlikely that we could possibly sneak in before MathJax is done initializing (the initial typeset is usually what takes so long when its startup runs), there's no reason not to guard against that case here. I feel like we may also want to do the same before `typesetShadow` is invoked below.

components/html-block/html-block.js
```diff
@@ -178,16 +178,21 @@ class HtmlBlock extends LitElement {
...
-		if (!this._templateObserver || this.noDeferredRendering) return;
+		if (!this._contentObserver || this.noDeferredRendering) return;
 
 		const template = this._findSlottedElement('TEMPLATE');
-		if (template) this._templateObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
+		if (template) {
+			this._contentObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
+		} else {
+			const container = this._findSlottedElement('DIV');
+			if (container) this._contentObserver.observe(container, { attributes: true, childList: true, subtree: true });
```
grant-cleary, line 188: This is technically different behaviour, since this will trigger the observer if a consumer adds an attribute onto the `div` itself. I suspect this case isn't worth worrying about, compared to the relative icky-ness of needing to attach the observer to every child element of the div just to ensure it and its children can use different options. 🤷

components/html-block/html-block.js
```diff
@@ -271,15 +283,23 @@ class HtmlBlock extends LitElement {
...
+			this._contentObserver = new MutationObserver(() => stampHTML(template, true));
+			this._contentObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
+			stampHTML(template, true);
+			return;
+		}
+
+		const container = this._findSlottedElement('DIV', slot);
+		if (container) {
+			this._contentObserver = new MutationObserver(() => stampHTML(container));
+			this._contentObserver.observe(container, { attributes: true, childList: true, subtree: true });
```
dlockhart, line 299: Very naively: could we not just listen to the `slotchange` event instead of wiring `MutationObserver`s up to the things inside the slot, or are things weird because this doesn't actually use Lit's `render()`? Or is the `MutationObserver` so that we catch any changes the consumer might make to their content _after_ we've initially rendered it?

What might be nice is if consumers didn't even need to put the `<div>` in there (which is kind of a strange API), but instead could just put their HTML right inside. That would mean changing this code slightly though to wire up observers to all children and to call `stampHTML` once for each child -- but would have the advantage that if only part of their HTML changes, you only re-stamp that bit. 🤷

grant-cleary: So yes, I believe this was originally done via an observer because of the second reason you mention, since the `slotchange` event only fires if the slotted element itself changes, not its descendants. I don't believe it has anything to do with Lit's rendering in this case.

Yeah, I don't think it's unreasonable to do something like that and I agree that this API makes much less sense with a div vs. a template. I definitely worry a bit about the cost of stamping once for every immediate child of the html-block... My main concern is the MathJax renderer, which has to create a custom document and do a bunch of wire-up. I _think_ we can get around that by restructuring how the renderers work though. MathJax in particular supports typesetting lists of elements and I would hope it would be able to do so at least somewhat efficiently. 🤔

I do think I like the above idea better from an API point of view though, since fetching the original HTML just becomes a case of pulling the html-block's `innerHTML`.

grant-cleary: I guess my thought is that on initial render, we probably want to stamp all children at once, and then afterwards stamping via the observer could happen one element at a time. I guess my concern there would be if the consumer changed out literally everything (which the LMS for sure does). In that case, we run into a situation where we have little choice but to have MathJax typeset element-by-element, unless we get clever with some way of batching changes within the renderer.

grant-cleary: Though... The slotchange event handler could always batch updates, which would probably solve that problem. If someone changed out all of their HTML, it's the slotchange events that would be firing to indicate that change. 🤔

grant-cleary: I can undo any changes here in the PR easily; I'm going to wire something like this up and if we don't like it, I can back it out.

dbatiste: >So yes, I believe this was originally done via an observer because of the second reason you mention, since the slotchange event only fires if the slotted element itself changes, not its descendants. I don't believe it has anything to do with Lit's rendering in this case.

I can confirm this is the case.

dbatiste: Consumers might try specifying multiple sibling `div` elements into the slot, and if we aren't prepared for it we'd just ignore everything except the first one.  They may or may not notice depending on the breadth of their testing.  Not great.  

So, I'm also interested in enabling them to just put their content in without requiring a single wrapper `div`.  Having said that, we should be considerate of any perf differences, good or bad, since all the plugins would be running that many more times, albeit on smaller bits of HTML.  

Also, be careful of text nodes because they only have a subset of the element DOM API.  Not sure if this would break anything.
```html
<d2l-html-block>
   <div>stuff</div>
   some text
   <!-- some comment -->
   <div>more stuff</div>
</d2l-html-block>
```

dbatiste: ^ how many times would MathJax's typesetting be done in this case, and on what HTML would it be operating in each case?

grant-cleary: Good point RE: text nodes. I think each renderer would need to have some understanding of how to handle them. I think we could straight up ignore comment nodes (I imagine we wouldn't want to mutate them under any circumstances, since I assume whoever wrote the comment wants it to be as is), but text nodes could have math or @mentions in them.

Naively, I wonder about batching the slotchange event. i.e. The html-block stores some internal state mapping handlers with the element they're processing (I think that needs to be a thing lest one event fires twice and we end up rendering duplicates). We then batch those in some reasonable way and only run the renderers once. This would require changing the renderers to operate on a list of elements rather than one at a time though.

And yes, MathJax in particular would typeset once per element, and it would be operating only on that element and its subtree.

grant-cleary: I'm less convinced we need special handling for the observers, though we could do something similar and I guess we may as well. The main reason I'm less concerned is because the slotchange event will already cover the cases where elements are being added/removed, so the main issue there would be if some consumer added an attribute to every element within the html-block or something of that sort.

dlockhart: For that initial render, you could also just take all the children and put them inside your own `<div>` before slamming them into the `<template>`'s `innerHTML` and then typeset that element.

Regardless, I think there's some combination of `slotchange` + mutation observers that should allow arbitrary children to be handled in a performant way... but likely requires some experimentation & testing. :)

grant-cleary: Yeah, it occurs to me that we're using the `slotchange` event to handle initial rendering anyways, so whatever works for the general case would also handle initial rendering. Naively, my thought is to batch via rAF - i.e. bundle up any events that process within a single frame and then render that batch at the end of the frame. I'm just starting to actually write code to test that all out though, so no word on how well it works quite yet. :D

grant-cleary: There's perhaps a broader problem with the idea of allowing rendering of only one or more children though. The MathJax renderer creates a new document and pulls in everything it's going to render. I don't know how feasible it is to adapt that to working on a list of potentially disparate elements scattered around the DOM while maintaining relative DOM order post-render.

components/html-block/html-block.js
```diff
@@ -255,8 +259,15 @@ class HtmlBlock extends LitElement {
...
 	_stamp(slot) {
 		const stampHTML = async template => {
-			const fragment = template ? document.importNode(template.content, true) : null;
-			if (fragment) {
+			let fragment;
+			if (template) fragment = document.importNode(template.content, true);
+			else {
+				fragment = document.createDocumentFragment();
+				this.shadowRoot.querySelector('slot').assignedNodes({ flatten: true })
+					.forEach(node => fragment.appendChild(node.cloneNode(true)));
```
grant-cleary, line 267: I think once we no longer support templates, it will make more sense to just skip constructing the fragment and just append these to the div directly (since the div won't yet be connected to the DOM anyways). However, for now, this makes the code branching a little less complicated right now, so I'm inclined to keep the fragment construction.

grant-cleary:
@dbatiste, @dlockhart (whenever you're back from your respective vacations), I've opened this up for review now as it should now match what we decided on offline. I was a bit more fast and loose with changing up the non-deferred rendering code, since nothing is using that yet. Otherwise, this should be fully backwards compatible.

components/html-block/html-block.js
```diff
@@ -178,24 +175,31 @@ class HtmlBlock extends LitElement {
 		super.connectedCallback();
 		if (this._contextObserverController) this._contextObserverController.hostConnected();
 
-		if (!this._templateObserver || this.noDeferredRendering) return;
+		if (!this._contentObserver || this.noDeferredRendering) return;
 
 		const template = this._findSlottedElement('TEMPLATE');
-		if (template) this._templateObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
+		if (template) this._contentObserver.observe(template.content, { attributes: true, childList: true, subtree: true });
```
dbatiste, line 181: Is this just here for backwards compatibility, and will be removed once consumers are switched?  Currently, it looks like `assignedNodes `needs to be resolved at least twice during initial render.

components/html-block/test/html-block.visual-diff.html
```diff
@@ -24,92 +24,41 @@
...
-				</ol>
-				<div><a href="https://d2l.com">anchor</a></div>
-			</template>
-		</d2l-html-block>
-	</div>
-
-	<div class="visual-diff">
-		<d2l-html-block id="update-template">
-			<template>before update</template>
+			<h1>heading 1</h1>
```
dbatiste, line 29: I know we discussed it previously, but I don't recall what we decided re text nodes.  I think it would be prudent to include a text node here as well.

components/html-block/demo/html-block.html
```diff
@@ -54,28 +54,26 @@ <h2>HTML Block</h2>
...
-								<li>unordered item 1</li>
-								<li>unordered item 2</li>
-							</ul>
-							<ol>
-								<li>ordered item 1</li>
-								<li>ordered item 2</li>
-							</ol>
-							<div><a href="https://d2l.com">anchor</a></div>
-						</template>
+						<h1>heading 1</h1>
```
dbatiste, line 57: Same as for visual-diff, I am thinking it would be good to include a text node as a direct child of `d2l-html-block` as well.

components/html-block/html-block.js
grant-cleary: Yes, this is strictly for backwards compatibility and can go away as soon as consumers have been updated. I acknowledge that it's sort of gross as is (because of that double rendering). I figured it wasn't worth splitting out the `assignedNodes` call just for this brief compatibility window, but if there are significant concerns I can also do that.

components/html-block/test/html-block.visual-diff.html
grant-cleary: Text nodes "just work" with the CSS change above to set `display: none` on the actual `<slot>`. I agree, I'll add an explicit text node to the demo page and the visual diff tests.

components/html-block/html-block.js
dbatiste: I might just check to see how many times that resolution is going to be required even after the template clean-up.  Otherwise I think it is fine.

grant-cleary: Sounds good. It's been a bit since I wrote the code so I'll double check, but my goal was to align things so that the number of resolutions is unchanged from the templated version.

dbatiste:
I think we'll need to update the readme as well, and there is also at least one code comment that would show up in the docs as well.
```html
/**
 * A component for displaying user-authored HTML.
 * @slot - Provide an html template that contains your user-authored HTML
 */
```

Other than that, I think it's looking good.  I would update the PR description to include the reason for this change, so in the future we aren't asking ourselves why we made this change.

grant-cleary:
> I think we'll need to update the readme as well, and there is also at least one code comment that would show up in the docs as well.
> 
> ```
> /**
>  * A component for displaying user-authored HTML.
>  * @slot - Provide an html template that contains your user-authored HTML
>  */
> ```
> 
> Other than that, I think it's looking good. I would update the PR description to include the reason for this change, so in the future we aren't asking ourselves why we made this change.

Good point. I was originally going to wait for the clean-up to update the README, but there's no reason not to do it now. Also good point RE: the PR description. I updated the title, but neglected to update the description accordingly. 😆

components/html-block/html-block.js
```diff
@@ -178,24 +175,31 @@ class HtmlBlock extends LitElement {
...
+		if (this._contentObserver) this._contentObserver.disconnect();
 	}
 
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
 		if (this._renderContainer) return;
-
-		this.shadowRoot.innerHTML += `<div class="d2l-html-block-rendered${this.compact ? ' d2l-html-block-compact' : ''}"></div><slot></slot>`;
+		this.shadowRoot.innerHTML += `<div class="d2l-html-block-rendered${this.compact ? ' d2l-html-block-compact' : ''}"></div><slot ${!this.noDeferredRendering ? 'style="display: none"' : ''}></slot>`;
```
dlockhart, line 202: The [classMap](https://lit.dev/docs/v1/lit-html/template-reference/#classmap) and [styleMap](https://lit.dev/docs/v1/lit-html/template-reference/#stylemap) might be a slightly cleaner way to do the dynamic CSS classes and inline styles.

grant-cleary: Hmm, I don't think that works outside of a Lit template though, does it? I did try that, but all it did was spit out the actual mapping code into the html-block.

dlockhart: Ah right, sorry.

grant-cleary: No worries, I wasn't 100% sure I wasn't missing something. I'm also hopeful that once all of this is done and dusted, I can take a look at switching this over to use Lit's rendering and template this stuff. I'm cautiously optimistic that it may be possible with more selective math rendering, but... No promises. 😅

dbatiste (approved):

dlockhart (approved):

annabelsegalld2l (approved):
Thanks for digging into this Grant! I've confirmed that this change works with html-blocks in Lit 2. The only change for consumers will be to remove the template tag wrapper, which we can add to our list of backwards-compatible changes.

grant-cleary:
> Thanks for digging into this Grant! I've confirmed that this change works with html-blocks in Lit 2. The only change for consumers will be to remove the template tag wrapper, which we can add to our list of backwards-compatible changes.

Sweet. Because the component has so few uses, I sort of figured I would just go through and open PRs to migrate them now-ish to get it out of the way. If we're more comfortable leaving that for the broader Lit 2 work then that's fine, but I also have no problem switching over consumers now.

dlockhart:
If it's easy and quick to do now, let's do it. That way we can clean up the soon-to-be-unneeded code in here now while it's fresh in your mind!

PR #1983, opened 2022-01-05
fix: dismissing on keydown so Esc on native elements like selects works

dlockhart:
Native/OS elements like `<select>` and `<dialog>` dismiss when Escape is pressed on `keydown`. But because our dismiss code is listening for `keyup`, in these scenarios we execute a second dismiss action.

For example, a dialog or dropdown which has a `<select>` in it -- if the user opens the `<select>` and then presses Escape, the OS/browser will dismiss it and prevent default on that `keydown` event. But a `keyup` event still fires, which hits our dismissible code and closes the dialog or dropdown. Not what we want.

By listening to the `keydown` event instead, the `preventDefault` works and our code doesn't run in this scenario.

🎩 -tip to @jguann for raising this and @jorge-ramirez-arredondo for your help with the solution!

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -282,9 +282,11 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 	_handleKeyDown(e) {
 		if (!this.opened) return;
 		if (e.keyCode === 27) {
-			// escape (note: prevent native dialog close so we can: animate it; use setDismissible)
+			// escape (note: prevent native dialog close so we can: animate it;)
 			e.stopPropagation();
 			e.preventDefault();
+			if (!this.opened) return;
+			this._close(abortAction);
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


components/dialog/dialog-mixin.js
```diff
@@ -282,9 +282,11 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 	_handleKeyDown(e) {
 		if (!this.opened) return;
 		if (e.keyCode === 27) {
-			// escape (note: prevent native dialog close so we can: animate it; use setDismissible)
+			// escape (note: prevent native dialog close so we can: animate it;)
 			e.stopPropagation();
 			e.preventDefault();
+			if (!this.opened) return;
+			this._close(abortAction);
```
dlockhart, line 289: Now that the dismissible code is listening for `keydown` as well, it no longer gets called here -- so we need to run the code to close the dialog ourselves.

dbatiste: Do we just need to remove the `e.stopPropagation();` so that the dismissible event handler can run?  Otherwise we're setting up a dismissible callback and not ever using it.

dlockhart: 🤦  Yep that is indeed all we need! Much cleaner now.

dbatiste (approved):

PR #1986, opened 2022-01-06
Remove template support for HTML Block API

grant-cleary:
All consumers have now been updated to avoid using the template container, so we can remove this code. :)


components/html-block/html-block.js
```diff
@@ -177,15 +177,11 @@ class HtmlBlock extends LitElement {
...
-		else {
-			const slot = this.shadowRoot.querySelector('slot');
-			if (slot) {
-				slot.assignedNodes({ flatten: true }).forEach(
-					node => this._contentObserver.observe(node, { attributes: true, childList: true, subtree: true })
-				);
-			}
+		const slot = this.shadowRoot.querySelector('slot');
+		if (slot) {
+			slot.assignedNodes({ flatten: true }).forEach(
```
grant-cleary, line 182: So situationally this can lead to `slot.assignedNodes` being called twice on render, specifically if an html-block gets re-added to the DOM, which will cause this to fire and then also fire the slotchange event. I don't know that this is avoidable though, since we can't exactly cache this (the nodes can certainly change before the slotchange event fires).

grant-cleary:
Well... Ok, I missed one spot. Going to get that merged first.

dbatiste (approved):

PR #1987, opened 2022-01-06
US133011 - Reduce dropdown margin gap to 18px and fix offset bug

svanherk:
- Reduces the margin gap to `18px` for the top and bottom (previously was `20px`/`20px` for bounded content and `30px`/`60px` for unbounded)
- Fixes an issue where we weren't taking custom vertical offsets into account


components/dropdown/dropdown-content-mixin.js
```diff
@@ -616,9 +619,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			if (bounded) {
 				spaceAround = this._constrainSpaceAround({
 					// allow for target offset + outer margin
-					above: targetRect.top - boundingContainerRect.top - 40,
```
svanherk, line 619: The previous 40 included 20 for the offset and 20 for the margin.  I'm changing it to use the actual offset - previously, if someone were to set a large custom offset the dropdown would get cutoff (I don't think anyone has actually done that in practice) or if they set a small offset, the margin would be bigger than needed.

This goes for all the examples below as well.  The previous number - 20 was the old margin.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -616,9 +619,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+					above: targetRect.top - boundingContainerRect.top - this._verticalOffset - 18,
```
dlockhart, line 622: Worth putting the `18` into a constant at the top?

dlockhart (approved):
LGTM

dlockhart (approved):

PR #1989, opened 2022-01-06
DE45955 - truncate all count badges to 5 digits

annabelsegalld2l:
Any numbers beyond 99,999 will be truncated as they do not provide too much useful information to consumers. It also allows us to avoid formatting overflows with giant numbers.


components/count-badge/README.md
```diff
@@ -46,7 +46,7 @@ The `d2l-count-badge` element is a web component to display a number count, depe
 | `number` | Number, required | The number to display on the badge.  Must be a positive integer. |
 | `size`, default: `small` | String | The size of the badge. Valid options are `"small"` and `"large"`. |
 | `type`, default: `count` | String | The type of the badge. Valid options are `"notification"` and `"count"`. Notification badges are orange and count badges are grey. |
-| `max-digits`, default: `2` when `type="notification"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and no limit for `"count"` type.
+| `max-digits`, default: `2` when `type="notification"`, `5` when `type="count"` | Number | Optionally specify a digit limit, after which numbers are truncated. Defaults to two for `"notification"` type and five for `"count"` type. Must be between 1-5.
```
dlockhart, line 49: Is there still a reason to even expose this as an attribute? Like are there valid examples where folks want `4`, or are our defaults of 2 and 5 now good enough? Just trying to keep the components as simple as possible.

annabelsegalld2l: There are cases, like filter, where the consumers want the count-type styling and a 2 digit limit. I think portfolio does the same, so I think we do need to keep it exposed.

dlockhart: Gotcha -- thanks!

components/count-badge/count-badge-mixin.js
```diff
@@ -156,25 +157,29 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
 
 	connectedCallback() {
 		super.connectedCallback();
-		if (!this.maxDigits && this.type === 'notification') {
-			// default to two digits for notification type
-			this.maxDigits = 2;
+		if (!this.maxDigits) {
+			// default to two digits for notification type, 5 for count
+			this.maxDigits = this.type === 'notification' ? 2 : 5;
```
svanherk, line 162: Looks like it'd be good to save 5 at the top as a constant and then use it in these three places. That way if we ever decide that we want 4 or 6 to be the max digits, we only have to update it in one place.  And then it's obvious that the default for count is actually "max"

dlockhart (approved):

PR #2008, opened 2022-01-11
Drag Image for Multi Drag & Drop

dbatiste:
Thus PR introduces a new (internal) accessory component for creating the drag image when multiple items are being dragged.

Note, in the mouse pointer is not displayed in the screenshots below because the screenshot tool doesn't capture it. However, the mouse pointer is displayed next to the count bubble.

When only one item is being dragged:
image

When more than one item is being dragged:
image


components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+			count: { type: Number }
+		};
+	}
+
+	static get styles() {
+		return [ bodySmallStyles, css`
+			:host {
+				display: block;
+				height: 70px;
+				left: -10000px;
```
dbatiste, line 25: The `d2l-offscreen` styles could not be used because they mess up the `dataTransfer.setDragImage` result.

components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+				left: 26px;
+				min-width: 1.4rem;
+				padding: 0.2rem 0.4rem;
+				position: absolute;
+				text-align: center;
+				top: 30px;
+				z-index: 1;
+			}
+			:host([dir="rtl"]) .count {
+				left: 14px;
```
dbatiste, line 96: There appears to be a bug with the `dataTransfer.setDragImage` in RTL - the browser doesn't maintain the position of the mouse pointer... it renders it on the left side.  This CSS is set to position the bubble next to the mouse pointer, even though the mouse pointer doesn't seem to be positioned the way I think it should.

components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+	constructor() {
+		super();
+		this.count = 0;
+	}
+
+	render() {
+		return html`
+			<div class="first">
+				<div class="count d2l-body-small">${formatNumber(this.count)}</div>
+				<svg width="18" height="18" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18">
```
dbatiste, line 110: We need to hard-code the `svg` image here to ensure it is loaded and rendered in time for the drag image to be created.

components/list/test/list-item-drag-image.visual-diff.html
```diff
@@ -0,0 +1,39 @@
...
+			import '../../typography/typography.js';
+		</script>
+		<style>
+			.visual-diff {
+				height: 80px;
+				position: relative;
+				width: 350px;
+			}
+			d2l-list-item-drag-image {
+				left: 0 !important;
```
dbatiste, line 21: This is simply to override the internal offscreen positioning so that we can take a screen shot.

components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+			  	<path fill="#494c4e" d="M8 16v1c0 .5-.4 1-1 1H6c-.6 0-1-.5-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 16v1c0 .5-.4 1-1 1h-1c-.6 0-1-.5-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 11v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 11v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 6v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1V6c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 6v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1V6c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 1v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1V1c0-.5.4-1 1-1h1c.6 0 1 .5 1 1M13 1v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1V1c0-.5.4-1 1-1h1c.6 0 1 .5 1 1"/>
+				</svg>
+				<d2l-input-checkbox disabled skeleton></d2l-input-checkbox>
+				<div class="text">
+					<div class="line-1"></div>
+					<div class="line-2"></div>
```
dlockhart, line 116: The skeleton stuff does have support for blocks of text as well -- I _think_ it should handle your use case here.

components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	render() {
+		return html`
+			<div class="first">
+				<div class="count d2l-body-small">${formatNumber(this.count)}</div>
+				<svg width="18" height="18" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18">
+			  	<path fill="#494c4e" d="M8 16v1c0 .5-.4 1-1 1H6c-.6 0-1-.5-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 16v1c0 .5-.4 1-1 1h-1c-.6 0-1-.5-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 11v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 11v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1v-1c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 6v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1V6c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M13 6v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1V6c0-.6.4-1 1-1h1c.6 0 1 .4 1 1M8 1v1c0 .6-.4 1-1 1H6c-.6 0-1-.4-1-1V1c0-.5.4-1 1-1h1c.6 0 1 .5 1 1M13 1v1c0 .6-.4 1-1 1h-1c-.6 0-1-.4-1-1V1c0-.5.4-1 1-1h1c.6 0 1 .5 1 1"/>
+				</svg>
+				<d2l-input-checkbox disabled skeleton></d2l-input-checkbox>
```
dlockhart, line 113: I'm a little torn on using a real checkbox here vs. just styling a box with rounded corners and a background colour to look like a skeleton checkbox. I guess screen reader users are never going to encounter this thing, so this is probably fine.

components/list/list-item-drag-image.js
```diff
@@ -0,0 +1,126 @@
...
+
+	constructor() {
+		super();
+		this.count = 0;
+	}
+
+	render() {
+		return html`
+			<div class="first">
+				<div class="count d2l-body-small">${formatNumber(this.count)}</div>
```
dlockhart, line 109: Interesting that we're not using the `<d2l-count-badge>` here since it does all this for us (but is grey or orange) -- I'm guess that was a design decision.

components/list/list-item-drag-image.js
dbatiste: Yeah I was on the fence - I just chose the checkbox so I would not have to hardcode the same dimensions.

components/list/list-item-drag-image.js
dbatiste: I'll give it a try.

components/list/list-item-drag-image.js
dbatiste: I'll talk to Jeff & Noah about it.  We could introduce a variable to override the color if needed.

dlockhart: Sounds good. Just seemed weird for us to have gone through and changed a bunch of places that had their own "coloured circle with a number in it" to use count badge and then re-introduce a new variant ourselves. :)

components/list/list-item-drag-image.js
dbatiste: Updated to use the skeleton styles.  I was not able to use the typography specific skeleton styles since there was a bunch of stuff I would have had to override.

components/list/list-item-drag-image.js
dbatiste: Closing the loop on this - for now this will be a one-off since we don't want to introduce a color override or new `type` value on `d2l-count-badge`.  

Aside, when I first tried this it didn't show up because all the internals of the element being used needs to be updated before `e.dataTransfer.setDragImage` is called, and it must be done synchronously with the `dragstart` event or no drag image will be created at all. I think this timing issue could be resolved by always rendering `d2l-list-item-drag-image` for a draggable item, instead of waiting until `dragstart`.

components/list/list-item-drag-image.js
dbatiste: Bugger... the skeleton z-index overlaps the bubble.  🤣   Will have to bump z-index on the bubble.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -555,9 +555,14 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
-			dragImage.count = selectionInfo.keys.length;
-			this.shadowRoot.appendChild(dragImage);
+			let dragImage = this.shadowRoot.querySelector('d2l-list-item-drag-image');
+			if (dragImage) {
+				dragImage.count = selectionInfo.keys.length;
+			} else {
+				dragImage = document.createElement('d2l-list-item-drag-image');
+				dragImage.count = selectionInfo.keys.length;
+				this.shadowRoot.appendChild(dragImage);
+			}
```
dlockhart, line 565: Nit: to remove duplicate code:

```suggestion
			let dragImage = this.shadowRoot.querySelector('d2l-list-item-drag-image');
			if (!dragImage) {
				dragImage = document.createElement('d2l-list-item-drag-image');
				this.shadowRoot.appendChild(dragImage);
			}
			dragImage.count = selectionInfo.keys.length;
```

dlockhart (approved):

PR #2012, opened 2022-01-12
RTL fix for count-badge-icon

annabelsegalld2l:
Currently, the count bubble is appearing on the incorrect side for RTL languages, until the badge is re-rendered (only on focus, which does not always happen). This leads to the bubbles swapping sides as you tab through the badges.

I was able to narrow down the issue to the RTL mixin not adding the dir="rtl" attribute until the count badge has already rendered. To fix this we can call the mixin's private property, _dir, which does get updated in its constructor. A bunch of places are already doing so (I assume because of similar timing issues?), see https://search.d2l.dev/search?full=%22this._dir%22&defs=&refs=&path=&hist=&type=&xrd=&nn=7&si=full&searchall=true&si=full. I don't think this is a great solution as we shouldn't be touching the mixin's private properties 😞 

An alternative solution I thought of was to update the RTLMixin constructor to forcibly set the dir attribute in its constructor, rather than relying on the requestUpdate call that isn't updating the attributes list in time. That way this.getAttribute('dir') will return the correct direction every time. I _think_ it should be fine, but it will require some more testing.

### DIFF
--- a/mixins/rtl-mixin.js
+++ b/mixins/rtl-mixin.js
@@ -2,12 +2,18 @@ import { dedupeMixin } from '@open-wc/dedupe-mixin';
 
 export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 
+	static get properties() {
+		return {
+			dir: { type: String, reflect: true }
+		};
+	}
+
 	constructor() {
 		super();
 		const dir = document.documentElement.getAttribute('dir');
 		// avoid reflecting "ltr" for better performance
 		if (dir && dir !== 'ltr') {
-			this.setAttribute('dir', 'rtl');
+			this.dir = dir;
 		}
 	}
 
--- a/mixins/rtl-mixin.js
+++ b/mixins/rtl-mixin.js
@@ -2,12 +2,6 @@ import { dedupeMixin } from '@open-wc/dedupe-mixin';
 
 export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 
-	static get properties() {
-		return {
-			_dir: { type: String, reflect: true, attribute: 'dir' }
-		};
-	}
-
 	constructor() {
 		super();
 		const dir = document.documentElement.getAttribute('dir');
--- a/mixins/rtl-mixin.js
+++ b/mixins/rtl-mixin.js
@@ -13,7 +13,7 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 		const dir = document.documentElement.getAttribute('dir');
 		// avoid reflecting "ltr" for better performance
 		if (dir && dir !== 'ltr') {
-			this._dir = dir;
+			this.setAttribute('dir', 'rtl');
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


dlockhart:
Yeah I wonder... since the initialization is happening in the constructor, how bad would it be for the `RtlMixin` to just not have a `_dir` and set `this.dir` (and `this.setAttribute('dir')` as you suggested Annabel) directly in the constructor? In other words, kind of bypass Lit's property/attribute reflection entirely since `dir` is a native attribute that all elements have.

dlockhart:
I tried changing `RtlMixin` to this, and it seems to work:

```javascript
export const RtlMixin = dedupeMixin(superclass => class extends superclass {

	constructor() {
		super();
		const dir = document.documentElement.getAttribute('dir');
		// avoid reflecting "ltr" for better performance
		if (dir && dir !== 'ltr') {
			this.setAttribute('dir', 'rtl');
		}
	}

});
```

The `setAttribute` call sets the native property automatically since that's just how the native DOM APIs work.

annabelsegalld2l:
> Yeah I wonder... since the initialization is happening in the constructor, how bad would it be for the `RtlMixin` to just not have a `_dir` and set `this.dir` (and `this.setAttribute('dir')` as you suggested Annabel) directly in the constructor? In other words, kind of bypass Lit's property/attribute reflection entirely since `dir` is a native attribute that all elements have.

Worried about removing the _dir entirely - there are a few spots that are using that private property. 🤔

mixins/rtl-mixin.js
```diff
@@ -13,7 +13,7 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 		const dir = document.documentElement.getAttribute('dir');
 		// avoid reflecting "ltr" for better performance
 		if (dir && dir !== 'ltr') {
-			this._dir = dir;
+			this.setAttribute('dir', 'rtl');
```
dlockhart, line 16: So I was kind of hoping we could remove the `properties()` section altogether and then switch the places referencing `this._dir` to just reference `this.dir` instead.

dbatiste: I think there are certain scenarios where modifying the DOM from within the constructor can cause a JS error.  I'm trying to remember where I first ran into this, and I think Allan ran into it as well.  🤔

dbatiste: I think it was assigning `id` from within the constructor - `createElement()` would fail.  That was one of the reasons (if not _the_ reason) for the constructor tests that we later created.  I don't remember Allan's case - it might have been the same, but different component.

annabelsegalld2l: You are exactly right Dave B. It's failing at this line: https://github.com/BrightspaceUI/core/blob/main/components/menu/menu.js#L167, since document.createElement does not allow for attributes 😢 .

dlockhart: What if we kept `dir` as a property above and right here set `this.dir = 'rtl';` (so the property gets set immediately). Then the attribute reflection code in Lit should handle the rest.

dbatiste (approved):
Looks good to me!

PR #2017, opened 2022-01-13
Alternate RTL fix: add mutationObserver to count-badge-icon to re-render on dir attribute

annabelsegalld2l:
This is an alternate fix for the RTL issues in the count badge. I think this is a pretty clean solution that we can suggest to others that will face this issue in the future. It avoids the issues we had with the attribute setting in the RTL mixin, and the performance impact is minimal since we only re-render for the dir attribute change.

### DIFF
--- a/components/count-badge/count-badge-icon.js
+++ b/components/count-badge/count-badge-icon.js
@@ -77,17 +77,9 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
 
 	connectedCallback() {
 		super.connectedCallback();
-		const self = this;
-		this._mutationObserver = new MutationObserver(mutations => {
-			mutations.forEach(mutation => {
-				if (mutation.type === 'attributes'
-					&& mutation.attributeName === 'dir') {
-					// re-render when dir attribute is added to reposition for RTL
-					self.requestUpdate();
-				}
-			});
-		});
-		this._mutationObserver.observe(this, { attributes: true });
+		// re-render when dir attribute is added to reposition for RTL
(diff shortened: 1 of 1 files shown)


components/count-badge/count-badge-icon.js
```diff
@@ -75,6 +75,26 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
 		this._badgeId = getUniqueId();
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		const self = this;
```
annabelsegalld2l, line 80: Binding the mutation function didn't seem to work, as I kept getting a "this is undefined" error. So I just created a self variable to use in the mutation listener.

components/count-badge/count-badge-icon.js
```diff
@@ -75,6 +75,18 @@ class CountBadgeIcon extends CountBadgeMixin(LitElement) {
 		this._badgeId = getUniqueId();
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		// re-render when dir attribute is added to reposition for RTL
+		this._mutationObserver = new MutationObserver(() => this.requestUpdate());
+		this._mutationObserver.observe(this, { attributeFilter: ['dir'] });
```
annabelsegalld2l, line 82: AttributeFilter ensures we will only request the update if the dir is changed (which should only happen once). This is valid syntax for all our supported browsers (https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe).

PR #2018, opened 2022-01-13
fix: Round padding so we don't lose the cursor in Firefox

svanherk:
David L reached out to me about this issue yesterday, it's an odd one.  So on certain screen resolutions and component locations on the page, the width of the slot that's measured is a crazy decimal number and we set the padding to that as well.  This can cause the cursor to disappear when it's near the padding, as seen in @GeorgePlukov's video below.  I could reproduce this on my external monitor but not my main laptop screen, and I had to adjust the demo page to move the input more to the center of the screen to make it happen.  This happens most consistently in Firefox (although I did get it to happen in Chrome a few times).

https://user-images.githubusercontent.com/13419300/149412179-95589f9b-adec-416c-9aec-7ecb62fa9475.mov

What I'm doing here is just rounding the padding up to always be an integer.  It means we may have slightly more padding than needed, but never by more than 1px.

Thanks @davern for raising this!


components/inputs/input-text.js
```diff
@@ -605,12 +605,14 @@ class InputText extends LabelledMixin(FormElementMixin(SkeletonMixin(RtlMixin(Li
 			|| (this.unit && this.dir !== 'rtl');
 
 		if (firstSlotHasNodes) {
-			requestAnimationFrame(() => this._firstSlotWidth = firstContainer.getBoundingClientRect().width);
+			// Firefox can have trouble rendering the cursor if the padding is a decimal
+			requestAnimationFrame(() => this._firstSlotWidth = Math.ceil(firstContainer.getBoundingClientRect().width));
```
GeorgePlukov, line 609: Any chance this could be happening across other input components?

svanherk: So I _think_ all the input components use this component as their base and will get the fix as well if they were having issues.  The visual diffs only updated for `input-percent` and `input-text`, but I believe that's because the other inputs either don't use the `slot` or the padding didn't change for those tests.  But I'll confirm we don't have any inputs with their own custom slots not using `input-text`

svanherk: Yeah looks like we're good.  Everything either uses `input-text` or something else that uses `input-text`, except `textarea` and `input-time` which don't offer the slotted sides.  `input-date` and `input-time` also make sure they themselves are an integer width using `Math.ceil` (https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-date.js#L196), so maybe they ran into similar issues at one point.

dbatiste (approved):

GeorgePlukov (approved):

dlockhart (approved):
Nice job sorting this out, crazy!

PR #2022, opened 2022-01-17
fix: Ignore dir in documentation

margaree:
Previously in the Daylight site `_dir` wasn't exposed because we filter out properties that start with underscores. The recent change to make `dir` public added this property to the properties tables for components that use `RtlMixin`.

This change just adds a description and values to `dir`, which does end up providing an easy to way to test RTL functionality in components that have it.

### DIFF
--- a/mixins/rtl-mixin.js
+++ b/mixins/rtl-mixin.js
@@ -4,6 +4,10 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 
 	static get properties() {
 		return {
+			/**
+			 * The directionality of the text of the document
+			 * @type {'ltr'|'rtl'}
+			 */
 			dir: { type: String, reflect: true }
 		};
 	}
--- a/mixins/rtl-mixin.js
+++ b/mixins/rtl-mixin.js
@@ -5,8 +5,7 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 	static get properties() {
 		return {
 			/**
-			 * The directionality of the text of the document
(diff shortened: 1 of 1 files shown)


mixins/rtl-mixin.js
```diff
@@ -4,6 +4,10 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
 
 	static get properties() {
 		return {
+			/**
+			 * The directionality of the text of the document
+			 * @type {'ltr'|'rtl'}
```
annabelsegalld2l, line 9: Dir will actually not be set if the direction is not RTL, so I think the only valid option is 'rtl' or 'undefined'.

margaree: Ah makes sense. Thoughts on if we should include that undefined would correspond to an ltr display then?

dbatiste: I think it would default to the document `dir`.  In practice, consumers should never be setting this explicitly except in test cases.  We would never want someone to set this to a value (overriding the document `dir`) without realizing the effect.  I'd _almost_ ignore this, or add to the comment as such.

annabelsegalld2l (approved):
Thanks for catching this Margaree!

dbatiste (approved):

mixins/rtl-mixin.js
dlockhart: I like where this ended up. Agreed that it should be ignored as Annabel said it's just a normal HTML property and our behaviour is basically the same as a native DOM node -- native nodes will have a value of empty string vs. ours will be undefined.

PR #2024, opened 2022-01-19
drop mention of ShadyCSS from README

omsmith:
Polyfill was removed in c3e26149b0dc4c974dcc1643aae7db56bea125c6


dlockhart (approved):
Good find, thanks Owen!

PR #2029, opened 2022-01-21
Adding options to d2l list

MitchelPaulin:
This PR adds a couple options to d2l-list in order to allow it to be usable in rubrics. Two options were added

- `padding=`: the slim padding still had some padding between list items, needed an option with no padding, replaced with a padding enum, should deprecate slim in favor of padding=slim going forward 
- `drag-target-handle-only`: the drag target for a list item was the entire cell, this did not work with rubrics because it made you unable to interact with the cells of the rubric, added an option to make the drag target the drag handle only 

All existing uses of d2l-lists should work exactly the same


components/list/list-item-mixin.js
```diff
@@ -55,9 +55,21 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
...
+			 * TODO: Remove in favor of padding="slim"
 			 * @type {boolean}
 			 */
 			slim: { type: Boolean },
+			/**
+			 * How much padding to render list items with
+			 * One of 'normal'|'slim'|'none', defaults to 'normal'
+			 * @type {string}
+			 */
+			padding: { type: String },
```
dbatiste, line 67: Let's call this property `paddingType` / attribute `padding-type` to distinguish it more from the CSS padding property.

dbatiste (commented):
I think we should make the same sort of change to `d2l-list-header` even if the padding doesn't need to change, at least to keep the API aligned.

I think we should update the existing visual-diff tests to use this new property, and also, include a new visual-diff test for the `padding-type="none"` case.

components/list/list-item-mixin.js
```diff
@@ -55,9 +55,21 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
...
+			 * TODO: Remove in favor of padding-type="slim"
 			 * @type {boolean}
 			 */
 			slim: { type: Boolean },
+			/**
+			 * How much padding to render list items with
+			 * One of 'normal'|'slim'|'none', defaults to 'normal'
+			 * @type {string}
+			 */
+			paddingType: { type: String, attribute: 'padding-type' },
```
dbatiste, line 67: Nit: we try to keep the public and private properties sorted alpha (separately), for both here and when they are assigned in the constructor.

components/list/README.md
```diff
@@ -363,7 +364,7 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
 | `selectable` | Boolean | Indicates an input should be rendered for selecting the item |
 | `selected` | Boolean | Whether the item is selected |
 | `skeleton` | Boolean | Renders the input as a skeleton loader |
-| `slim` | Boolean | Whether to render the list-item with reduced whitespace|
+| `padding-type` | String | List item whitespace (`normal` (default), `slim`, `none`)|
```
dbatiste, line 367: Nit: Same thing here... this should be moved up in the order.

dbatiste (approved):
Looks good.  Just a couple of small nit tweaks to be made.

PR #2030, opened 2022-01-21
DE46715 added ariaDescribeContent property for a11y

emsandrews:
Original Rally Story: 
[DE46715](https://rally1.rallydev.com/#/?detail=/defect/622118312995&fdp=true): [A11y] Update all FACE help dialogs focus to be on the close (x) button instead of the okay button
 
In FACE quiz accessibility reviews it was pointed out that when the user opens the help dialog with the screen reader, the focus goes directly to the footer making it difficult for non-sighted users to access the dialog content. 

The original suggestion was to have the focus land on the close button at the top so users could navigate to the content text more easily, but based on my conversation with Dave B this would require larger changes to the dialog component. 

This is an alternative solution where there is an optional property added to the dialog that allows the content to be read out automatically on opening the dialog by using the aria-describedby property in the dialog-mixin. 

If aria-describe-content="true" the dialog content will be read when the dialog is opened.

### DIFF
--- a/components/dialog/README.md
+++ b/components/dialog/README.md
@@ -83,6 +83,12 @@ The `d2l-dialog` element is a generic dialog that provides a slot for arbitrary 
 - `d2l-dialog-close`: dispatched with the action value when the dialog is closed for any reason
 <!-- docs: end hidden content -->
 
+### Accessibility Properties
+
+| Attribute | Description |
+|--|--|
+| `describe-content` | When set to 'true' reads the dialog content on open. |
+
 ### Methods
 
 - `resize`: resizes the dialog based on specified `width` and measured content height
--- a/components/dialog/README.md
+++ b/components/dialog/README.md
@@ -87,7 +87,7 @@ The `d2l-dialog` element is a generic dialog that provides a slot for arbitrary 
 
 | Attribute | Description |
 |--|--|
-| `describe-content` | When set to 'true' reads the dialog content on open. |
+| `describe-content` | When set, screen readers will announce the contents when opened. |
 
 ### Methods
 
--- a/components/dialog/dialog.js
+++ b/components/dialog/dialog.js
@@ -24,13 +24,14 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 	static get properties() {
 		return {
 			/**
-			 * Whether to read the contents of the dialog on open
-			 */
-			describeContent: { type: Boolean, attribute: 'describe-content' },
-			/**
 			 * Whether to render a loading-spinner and wait for state changes via AsyncContainerMixin
 			 */
 			async: { type: Boolean },
+
+			/**
+			 * Whether to read the contents of the dialog on open
+			 */
+			describeContent: { type: Boolean, attribute: 'describe-content' },
 
 			/**
 			 * The preferred width (unit-less) for the dialog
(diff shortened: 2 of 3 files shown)


components/dialog/dialog.js
```diff
@@ -151,9 +156,11 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 			'd2l-footer-no-content': !this._hasFooterContent
 		};
 
+		if (!this._textId) this._textId = getUniqueId();
+
 		const content = html`
 			${loading}
-			<div style=${styleMap(slotStyles)}><slot></slot></div>
+			<div id="${this._textId}" aria-describe-content="${this.ariaDescribeContent}" style=${styleMap(slotStyles)}><slot></slot></div>
```
mdulat, line 163: I don't think you need the `aria-describe-content="${this.ariaDescribeContent}"` here.

emsandrews: you are right! I will take that out :)

margaree (commented):
Overall I think this strategy makes sense and seems inline with example 3 in the w3 examples here https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

components/dialog/dialog.js
```diff
@@ -22,6 +22,10 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 
 	static get properties() {
 		return {
+			/**
+			 * Whether to read the contents of the dialog on open
+			 */
+			ariaDescribeContent: { type: Boolean, attribute: 'aria-describe-content' },
```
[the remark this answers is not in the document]
dlockhart: Agreed with everything Margaree mentioned here -- I like calling this `describe-content`.

Also it might be good to have some kind of guidance in the documentation for when and when not this should be used... like should we be defaulting it to `true` for all our prompt-style dialogs? And if so, maybe this doesn't need to be a customizable thing at all but just something that's built into the `<d2l-dialog-confirm>`?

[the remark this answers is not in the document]
mdulat: This is already built in to `d2l-dialog-confirm` but we need it in `d2l-dialog` as well. Not sure if it would make sense to build it in there as well unconditionally?

components/dialog/README.md
```diff
@@ -83,6 +83,12 @@ The `d2l-dialog` element is a generic dialog that provides a slot for arbitrary
 - `d2l-dialog-close`: dispatched with the action value when the dialog is closed for any reason
 <!-- docs: end hidden content -->
 
+### Accessibility Properties
+
+| Attribute | Description |
+|--|--|
+| `describe-content` | When set to 'true' reads the dialog content on open. |
```
dbatiste, line 90: I might just rephrase this so consumers don't attempt to assign "true".  Maybe `When set, screen readers will announce the contents when opened. Useful when the the dialog contains no other focusable elements.`

components/dialog/dialog.js
```diff
@@ -22,6 +23,10 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 
 	static get properties() {
 		return {
+			/**
+			 * Whether to read the contents of the dialog on open
+			 */
+			describeContent: { type: Boolean, attribute: 'describe-content' },
```
dbatiste, line 29: Nit: let's move this below `async` to keep the public properties sorta alpha.

components/dialog/dialog.js
```diff
@@ -103,6 +108,7 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 		this.width = 600;
 		this._handleResize = this._handleResize.bind(this);
 		this._handleResize();
+		this.describeContent = false;
```
dbatiste, line 111: Nit: same thing here... let's move this up below `async` so that the property assignments are sorta alpha.

dbatiste (commented):
Looks good... just a couple of super minor tweaks.

dbatiste (approved):
Looks great!

PR #2034, opened 2022-01-24
Non-breaking changes for Lit 2

annabelsegalld2l:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -332,7 +332,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 
 		requestAnimationFrame(async() => {
 
-			this.shadowRoot.querySelector('.d2l-dialog-content').scrollTop = 0;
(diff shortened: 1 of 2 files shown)


dlockhart (commented):
I don't think I'd be too worried about these two `shadowRoot` accesses since the one is inside a drag-start handler and the other is after an animation frame. Unless you noticed an issue that is.

annabelsegalld2l:
Closing - no changes required

PR #2035, opened 2022-01-24
Breaking changes for Lit 2

annabelsegalld2l:
This PR makes the breaking changes required to update to Lit 2 (`lit-html` 2 and `lit-element` 3). The Lit 2 upgrade will be a coordinated effort, so **this PR should not merge at this time**.

Breaking change checklist: 
- [x] Update `lit-html` -> `^2`, `lit-element` -> `^3`, `open-wc/testing` -> `^3` (if used)
- [x] `createRenderRoot` usages -> move from LitElement to ReactiveElement 
- [x] Remove Dependabot Lit rules
- [x] Bump any dependencies to their Lit 2 versions
- [x] Remove any manual lifecycle calls to Lit 1 reactive controllers
- [x] Backwards-compatible changes merged: (https://github.com/BrightspaceUI/core/pull/2034) (no changes required)

Renamed API's:
- [x] `UpdatingElement` -> `ReactiveElement`
- [x] `@internalProperty` -> `@state` (we don't use TypeScript decorators)
- [x] `static getStyles()` -> `static finalizeStyles(styles)`
- [x] `_getUpdateComplete()` -> `getUpdateComplete()`
- [x] `NodePart` -> `ChildPart`

Testing checklist: 
- [x] CI/ unit tests pass
- [x] Local testing on demo pages
- [x] Testing component in the LMS

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


controllers/subscriber/README.md
```diff
@@ -136,24 +136,6 @@ An example of what this could look like altogether:
 <younger-viewer for="rogers"></younger-viewer>
 ```
 
-NOTE: Until we are on Lit 2, the controller lifecycle events will need to be manually called:
-```js
-	connectedCallback() {
```
annabelsegalld2l, line 141: All these manual callbacks can be deleted 🎉

controllers/subscriber/subscriberControllers.js
```diff
@@ -118,9 +121,7 @@ export class IdSubscriberController {
 		});
 	}
 
-	hostUpdated(changedProperties) {
```
annabelsegalld2l, line 121: The hostUpdated lifecycle method has no parameters, so changedProperties will always be undefined. Removing the check allows the tests to pass properly. I'm not too sure on the effects this might have to the below code though - haven't done too much digging.

Context: https://lit.dev/docs/api/controllers/#ReactiveController.hostUpdated

controllers/subscriber/README.md
```diff
@@ -135,25 +135,16 @@ An example of what this could look like altogether:
 </cable-subscription>
 <younger-viewer for="rogers"></younger-viewer>
 ```
+NOTE: If wanting to use IdSubscriberController, you must call its `hostPropertiesChanged` function in the `updated` function, passing along the `changedProperties` parameter:
```
annabelsegalld2l, line 138: Added this note here for the IdSubscriberController to remember to call the hostPropertiesChanged function in the updated function.

controllers/subscriber/subscriberControllers.js
svanherk: Oh interesting!  This makes their `hostUpdated` function much less useful lol.  If we wanted to avoid the consumer having to know to call this, I guess we could keep track of the previous value and do our own comparison each time `updated` is called, but that feels like a waste when Lit already does exactly that comparison to create `changedProperties`... 🤔 

So I'm good with this strategy unless others feel strongly about hiding this from the consumer!

svanherk:
This all looks good to me!  Didn't approve just so it's harder to accidentally merge early lol, but let me know if you want me to

controllers/subscriber/subscriberControllers.js
annabelsegalld2l: I like that idea Stacey. I've pushed it here, but I can revert if we like the other approach better. We know that the property is a space-delimited string, so the comparison each time shouldn't be too performance-intensive.

svanherk: Right good call, I was still thinking the property was an array (it was initially) and would be complicated to compare.  Nice!

dlockhart (approved):
LGTM! 🎉

svanherk (approved):
🎉

PR #2041, opened 2022-01-25
fix: remove use of optional chaining

devpow112:
This was causing issues with downstream polymer builds (found while working in [Brightspace/brightspace-activity-details-ui](https://github.com/Brightspace/brightspace-activity-details-ui)).


annabelsegalld2l (approved):
Thanks for catching that Devon!

devpow112:
> Thanks for catching that Devon!

No problem 😄

PR #2057, opened 2022-01-31
Add ability to select all items across all pages.

dbatiste:
This PR updates the selection and list components to allow all items to be selected across all pages.  To summarize the changes:

* add `select-all-pages-allowed` to `d2l-list-header` to allow opt into this behaviour since consumers must code to interpret the `all-pages` selection state accordingly (i.e. client-side does not have _all_ the keys)
* add new `d2l-selection-select-all-pages` component that can be used to select all items across all pages
* update `d2l-list-header` to render `d2l-selection-select-all-pages` if allowed
* add `item-count` to the `SelectionMixin` so that `d2l-selection-select-all-pages` and `d2l-selection-summary` can display it accordingly. The consumer **must** provide this because all the items are not necessarily loaded on the client. We talked about more generic verbiage, but for now, this is a requirement to use this feature.
* update `d2l-selection-input` to automatically select itself if it's appended to a selection provider where selecting all pages is toggled on.
* update `d2l-selection-select-all` to have checked state when selection info state is either `all` or `all-pages`.

Example setup for `d2l-list`:
```html
<d2l-list item-count="50">
	<d2l-list-header slot="header" select-all-pages-allowed>
		...
	</d2l-list-header>
	<d2l-list-item selectable key="1" label="...">...</d2l-list-item>
	<d2l-list-item selectable key="2" label="...">...</d2l-list-item>
	...
</d2l-list>
```

Screenshot showing it enabled in `d2l-list`:
image


components/list/list-header.js
```diff
@@ -54,11 +60,11 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			}
 			d2l-selection-summary {
 				flex: none;
-				margin-left: 0.9rem;
+				margin-inline-start: 0.9rem;
```
dlockhart, line 63: Do we feel comfortable introducing CSS logical properties? I thought we were holding back a bit, although it's really just Safari 11 at this point that doesn't support it... and Safari 11 has all kinds of other issues.

components/selection/selection-mixin.js
```diff
@@ -38,6 +39,11 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 
 	static get properties() {
 		return {
+			/**
+			 * Total number of items. Required when selecting all pages is allowed.
+			 * @type {number}
+			 */
+			itemCount: { type: Number, attribute: 'item-count' },
```
dlockhart, line 46: Default to zero probably?

dlockhart (approved):

components/list/list-header.js
dbatiste: I switch it to be on the safe side.

PR #2059, opened 2022-01-31
Add d2l-selection-action-menu-item

dbatiste:
This PR adds a new menu-item implementation `d2l-selection-action-menu-item` for use in selection components similar to `d2l-selection-action`. It similarly provides a `requires-selection` property to control enabled/disabled bheaviour of the component based on whether any items are selected.

This PR builds on top of: https://github.com/BrightspaceUI/core/pull/2057 since it touches some of the same files.  My suggestion is to review this PR after it merges, but just posting this as draft until it does.


dlockhart:
If you'd like to isolate the change, you can have the base branch for this PR be the "all pages" stuff, and then when the other merges GitHub will automatically adjust it to `main`.

components/list/demo/list.html
```diff
@@ -67,8 +69,15 @@ <h2>List (with all the stuff)</h2>
...
-							<d2l-selection-action icon="tier1:share" text="Share" requires-selection></d2l-selection-action>
 							<d2l-selection-action icon="tier1:delete" text="Delete" requires-selection></d2l-selection-action>
+							<d2l-dropdown-button-subtle text="Actions">
+								<d2l-dropdown-menu>
+									<d2l-menu label="Actions">
+										<d2l-selection-action-menu-item text="Bookmark (requires selection)" requires-selection></d2l-selection-action-menu-item>
+										<d2l-selection-action-menu-item text="Advanced"></d2l-selection-action-menu-item>
+									</d2l-menu>
+								</d2l-dropdown-menu>
+							</d2l-dropdown-button-subtle>
```
dlockhart, line 80: If we wanted to clamp this down a bit in terms of flexibility (and make it simpler to use), wondering what something like this would look like:

```
<d2l-selection-action-menu text="Actions">
  <d2l-selection-action-menu-item text="Bookmark (requires selection)" requires-selection></d2l-selection-action-menu-item>
  <d2l-selection-action-menu-item text="Advanced" requires-selection></d2l-selection-action-menu-item>
</d2l-selection-action-menu>
```

Pros: similar to how we deal with `<d2l-selection-action>`, it puts us in control of what actually gets rendered (a subtle button), vs. technically they could put whatever they want here to open the menu. Under the covers, `<d2l-selection-action-menu>` would render the `<d2l-dropdown-button-subtle>`, the `<d2l-dropdown-menu>`, the `<d2l-menu>` and all the items.

Cons: might be tricky to map the slotted `<d2l-selection-action-menu-item>`s back up to the parent such that it can re-render with menu items. But I think the work @svanherk did for filter could be reused here.

dlockhart: Right, OK. So basically we're saying that this is just a special menu-item type that's aware of selection state vs. a menu version of `<d2l-selection-action>`.

dlockhart (approved):

PR #2064, opened 2022-02-02
Add support for native :focus-visible selectors in menu-item mixin

grant-cleary:
It turns out that when the focus-visible polyfill is imported into a menu item, it can explode into doing a lot of work on `visibilitychange` events relatively quickly. e.g. Each instance of the html-editor renders 210 menu items, so this balloons fast.

It looks like Safari is primed to add support for the native `:focus-visible` pseudo-selector in the near-to-mid future (maybe a bit optimistic), so after consulting with Gaudi, I think we're looking to remove the polyfill from this component specifically and make due with native `:focus-visible`, accepting that it might make the styling not quite so ideal for Safari users.


components/menu/menu-item-styles.js
```diff
@@ -19,6 +19,8 @@ export const menuItemStyles = css`
 	:host(:hover),
 	:host(.focus-visible),
 	:host(.focus-visible[first]),
+	:host(:focus-visible),
+	:host(:focus-visible[first]),
```
grant-cleary, line 23: I confirmed that the new selectors work as expected in Chrome. I think we probably do want to add them separately instead of replacing the class selectors, since these styles could be applied to other components and I don't particularly want to break them if they opt to use the polyfill.

dlockhart: I don't quite follow... why would other components put the `.focus-visible` CSS class on the menu item?

components/menu/menu.js
```diff
@@ -21,7 +20,7 @@ const keyCodes = {
  * @slot - Menu items
  * @fires d2l-menu-resize - Dispatched when size of menu changes (e.g., when nested menu of a different size is opened)
  */
-class Menu extends ThemeMixin(HierarchicalViewMixin(FocusVisiblePolyfillMixin(LitElement))) {
+class Menu extends ThemeMixin(HierarchicalViewMixin(LitElement)) {
```
grant-cleary, line 23: Comments on the original PR that added the polyfill here implied that this was needed solely to allow the polyfill to function on nested menu-items, so it seems like we also want it gone here.

components/menu/menu-item-styles.js
grant-cleary: Not a clue! If you don't think it's a risk we care about, I can remove it. I just opted for the least assumptions.

dlockhart: Seeing as George added it as part of that [original change](https://github.com/BrightspaceUI/core/pull/1888/), I think it's pretty safe to remove.

grant-cleary: Fair, I guess it's only been ~2 months, so it's extremely unlikely anybody is doing anything weird with those directly. Mostly I'm usually concerned with "better safe than sorry" with things I'm likely going to hotfix so I'm more conservative than I need to be even with safe changes like this. 🤷

dbatiste (approved):

PR #2070, opened 2022-02-02
fix: format using Intl API to avoid scientific notation

dlockhart:
This is related to DS12459, which is being investigated by @tylergee and Team Borg. I'm going to backport it to `20.22.2` and `20.22.1`.

Little known fact (by me at least):
> If number is longer than 21 digits, JavaScript uses exponential (aka scientific) notation. JavaScript also uses exponential notation if number starts with “0.” followed by more than five zeros.

It's that second case of a number starting with zero and then having more than five zeroes that comes up in the defect. It results in `0.0000005` being saved as `5e-7.0000000` which then causes our .NET parsing code to explode.

The solution is to leverage the `Intl.NumberFormat` API to do all the heavy lifting for us.

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -264,14 +264,13 @@ class InputNumber extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeC
 			countDecimalDigits(this._valueTrailingZeroes, false),
 			this.maxFractionDigits
 		);
-		let valueTrailingZeroes = this.value.toString();
-		const decimalDiff = (numDecimals - countDecimalDigits(valueTrailingZeroes, false));
-		if (decimalDiff > 0) {
-			if (decimalDiff === numDecimals) {
-				valueTrailingZeroes += '.';
+		const valueTrailingZeroes = new Intl.NumberFormat(
+			'en-US',
+			{
+				minimumFractionDigits: numDecimals,
+				useGrouping: false
 			}
-			valueTrailingZeroes = valueTrailingZeroes.padEnd(valueTrailingZeroes.length + decimalDiff, '0');
-		}
+		).format(this.value);
 		return valueTrailingZeroes;
 	}
 	set valueTrailingZeroes(val) {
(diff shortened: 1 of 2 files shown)


components/inputs/input-number.js
```diff
@@ -264,14 +264,13 @@ class InputNumber extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeC
 			countDecimalDigits(this._valueTrailingZeroes, false),
 			this.maxFractionDigits
 		);
-		let valueTrailingZeroes = this.value.toString();
-		const decimalDiff = (numDecimals - countDecimalDigits(valueTrailingZeroes, false));
-		if (decimalDiff > 0) {
-			if (decimalDiff === numDecimals) {
-				valueTrailingZeroes += '.';
```
dlockhart, line 271: The nice thing about this solution is that `Intl.NumberFormat` will pad out the zeroes for us, so we don't need to calculate the "decimal diff" and manually pad it out.

components/inputs/input-number.js
```diff
@@ -264,14 +264,13 @@ class InputNumber extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeC
 			countDecimalDigits(this._valueTrailingZeroes, false),
 			this.maxFractionDigits
 		);
-		let valueTrailingZeroes = this.value.toString();
```
dlockhart, line 267: The issue stems from `this.value` being `Number` and its string representation being `5e-7` at this point.

svanherk:
> If number is longer than 21 digits, JavaScript uses exponential (aka scientific) notation.

Does this fix protect us from that scenario as well?  Sounds fairly unlikely to actually happen but...

svanherk (approved):

dlockhart:
> Does this fix protect us from that scenario as well? Sounds fairly unlikely to actually happen but...

It does, although we JavaScript loses precision after 18 digits, so `1234567890123456789012` becomes `1234567890123456780000`.

I've noticed that our regular number inputs (not using the "trailing zeroes" functionality) also suffer from this, but that fix is going to need to go into the MVC framework. I'll work on that tomorrow.

PR #2078, opened 2022-02-03
Push MathJax font styles to document head to allow them to download

grant-cleary:
So... It turns out that MathJax fonts aren't loading properly within the html-block control when rendering is deferred; specifically because MathJax sprouts its font at-rules in a style element in whatever document context you give it. It turns out that within a component's shadowDOM, these fonts are just never downloaded.

My proposed solution is to mimic what MathJax does with its styles when rendering isn't deferred: sprout the font at-rules in the document head. This allows them to download as one would expect.

As usual, I'm entirely open to other ideas here. :)

### DIFF
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -1,4 +1,29 @@
 const mathjaxContextAttribute = 'data-mathjax-context';
+const mathjaxBaseUrl = 'https://s.brightspace.com/lib/mathjax/3.1.2';
+
+const mathjaxFontMappings = new Map([
+	['MJXTEX', 'MathJax_Main-Regular'],
+	['MJXTEX-B', 'MathJax_Main-Bold'],
+	['MJXTEX-I', 'MathJax_Math-Italic'],
+	['MJXTEX-MI', 'MathJax_Main-Italic'],
+	['MJXTEX-BI', 'MathJax_Math-BoldItalic'],
+	['MJXTEX-S1', 'MathJax_Size1-Regular'],
+	['MJXTEX-S2', 'MathJax_Size2-Regular'],
+	['MJXTEX-S3', 'MathJax_Size3-Regular'],
+	['MJXTEX-S4', 'MathJax_Size4-Regular'],
+	['MJXTEX-A', 'MathJax_AMS-Regular'],
+	['MJXTEX-C', 'MathJax_Calligraphic-Regular'],
+	['MJXTEX-CB', 'MathJax_Calligraphic-Bold'],
+	['MJXTEX-FR', 'MathJax_Fraktur-Regular'],
+	['MJXTEX-FRB', 'MathJax_Fraktur-Bold'],
+	['MJXTEX-SS', 'MathJax_SansSerif-Regular'],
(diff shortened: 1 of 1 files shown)


helpers/mathjax.js
```diff
@@ -158,6 +183,23 @@ export function loadMathJax(mathJaxConfig) {
 		}
 	};
 
+	if (mathJaxConfig && mathJaxConfig.deferTypeset && !document.head.querySelector('#d2l-mathjax-fonts') && !document.head.querySelector('#MJX-CHTML-styles')) {
```
grant-cleary, line 186: `MJX-CHTML-styles` is the `id` of the stylesheet that MathJax applies, which includes the font at-rules. If it exists in the document head already, we don't need to add this new style. Notably, we also don't want to use this `id` because MathJax does a lot more than just the at-rules with those styles, so we also don't want to clobber MathJax's styles with these.

dbatiste (commented):
I really wish MathJax would provide an out of the box solution that handles all this so we would not have to be duct-taping this stuff together.  Any thought on raising this with them in response to the thread that talks about how use MathJax in shadowDOM, or starting a new thread?

It would be much worthwhile to include a comment at the top of this file indicating the details of what needs to be checked when upgrading MathJax.

Aside from that, looks ok.

grant-cleary:
> I really wish MathJax would provide an out of the box solution that handles all this so we would not have to be duct-taping this stuff together. Any thought on raising this with them in response to the thread that talks about how use MathJax in shadowDOM, or starting a new thread?

Yep, I can do that for sure.

> It would be much worthwhile to include a comment at the top of this file indicating the details of what needs to be checked when upgrading MathJax.

Good point, yeah. I also find myself wondering if this would be a good candidate for automating deployment to the CDN, much like TinyMCE. That wouldn't change needing to add a comment here, but it's food for thought I think.

dbatiste (approved):

grant-cleary:
These same visual diff tests also failed across another PR. @dbatiste, any objections if I just merge in the new goldens?

grant-cleary:
Ack, accidentally deleted my previous comment... Any objections if I merge in the new goldens? These reports also failed across other PRs.

dbatiste:
> Ack, accidentally deleted my previous comment... Any objections if I merge in the new goldens? These reports also failed across other PRs.

Seems like the browser is doing something differently with shadows.  Is this happening consistently?

grant-cleary:
> Seems like the browser is doing something differently with shadows. Is this happening consistently?

Failures have been consistent across three runs, and I noticed the same set of failures here: https://github.com/BrightspaceUI/core/pull/2083
https://github.com/BrightspaceUI/core/pull/2086

dbatiste:
So yeah, if they are consistent I think it's fine.

grant-cleary:
> Hey Grant, you can merge your changes in now. After your PR is merged, I am going to remove some of the flaky tests that are causing us trouble so hopefully we won't need to re-run 5 times the next time we run into an issue like this 😄

Oh, haha, fair enough. Unfortunately I don't have permission to skip required steps in this repo so I think someone else would need to push it through for me. 😬

Issue #2076, opened 2022-02-03
d2l-scroll-wrapper does not work as expected with d2l-list


MitchelPaulin:
See this branch for how its being used https://github.com/Brightspace/d2l-rubric/tree/US135547-Add-Back-Scroll-Wrapper

The vertical portion will always attempt to scroll even when there is no need, in the screenshot you can see that the group is the correct size with no hidden elements causing the vertical scroll to kick in, so not sure why it does. I found a simple fix, if I just set `overflow-y: hidden` in the scroll wrapper it behaves like it should, though not sure if adding an API option to set the vertical scrolling behavior of the scroll wrapper would be an acceptable solution.

![Animation4](https://user-images.githubusercontent.com/30714020/152378498-f8163fb5-c45e-4bc2-8b45-984534250228.gif)
![Untitled](https://user-images.githubusercontent.com/30714020/152378516-05c8495d-1b5f-46e7-b078-f3ad31163a6b.png)

dbatiste:
This is fixed with the above merged change.

PR #2082, opened 2022-02-04
Remove the list-item drag image rotation since it cause problems with overflow and dimension calcs.

dbatiste:


components/list/list-item-mixin.js
```diff
@@ -281,10 +275,6 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 				width: 100%;
 				z-index: 5;
 			}
-			:host([draggable][selected]:not([disabled])) .d2l-list-item-active-border,
-			:host([draggable][selected]:not([disabled])) d2l-list-item-generic-layout.d2l-focusing + .d2l-list-item-active-border {
-				transform: rotate(1deg);
-			}
```
dlockhart, line 287: Would this enable us to remove the surrounding `<div class="d2l-list-item-drag-image">` if drag & drop isn't enabled? I had that question the last time I was in list and I think I came to the conclusion that we couldn't but it was because of this CSS.

Since list items get duplicated SO many times in large lists, the fewer DOM nodes -- especially for simple lists -- the better.

dbatiste: Good question.  That element currently also wraps the "<div class="d2l-list-item-active-border"></div>" element which would currently get included in the drag image.  I can play around with removing it.

dbatiste: So, the answer is, not without affecting the color of the bottom border in the drag image when a selected item is dragged.  The color difference is very subtle with the transparency but it is different, and if we ever changed the color it might be more noticeable.

dlockhart: > ...not without affecting the color of the bottom border in the drag image when a selected item is dragged

Sorry, I meant keep the `<div>` but only render it when `draggable` is set.

dbatiste: Yeah we should be able to do that, though I think Lit will generate two comment nodes to mark the part that results from conditional rendering.

dlockhart: Even if you do something like:

```javascript
const innerView = html`<inner list view>`;
return this.draggable ? html`<div class="d2l-list-item-active-border">${innerView}</div>` : innerView;
```

dlockhart (approved):
Nice!

annabelsegalld2l (approved):
Just tested this with BSI on the rubrics page, looks like it resolves the big whitespace problem 👍 
(https://github.com/BrightspaceUI/core/issues/2076)

I think the visual diff is a flake so no need to merge those changes.

dbatiste:
> Just tested this with BSI on the rubrics page, looks like it resolves the big whitespace problem 👍 (#2076)
> 
> I think the visual diff is a flake so no need to merge those changes.

Thanks Annabel.

Yeah I agree... I think that test is a flake.  I think I have seen it fail before and I suspect it is a timing issue.

Issue #2075, opened 2022-02-03
d2l-filter -- Scrollbar doesn't show up if using the loading state and have many dimension-set-values


Chebsi:
If you are using a dimension set with many values, and also use the loading state, the dropdown's scrollbar doesn't appear, so you can't see all the values. If you don't use loading, or if you close and reopen the filter, scrolling will work as expected.

Slack thread for context: https://d2l.slack.com/archives/C0PHG3QB0/p1643834555293599
Also interesting as noticed by Annabel it does not happen with the multi-dimension filter, just single dimension.

Here's one example based on the demo filters page.  I've added many items to the demo filter showing the loading state, and the scrollbar won't show up after the filter is done loading.
![filter-demo](https://user-images.githubusercontent.com/1237719/152362680-9421cba9-923a-491f-9fb7-f159324d5e41.gif)

annabelsegalld2l:
Hi Chebsi, this issue should be resolved in core version 1.231.1. Please re-open this issue you have any more trouble with the filter scrollbar 😄

PR #2083, opened 2022-02-04
DE47181: Add resizeObserver to dropdown content slot to re-render on filter resize

annabelsegalld2l:
https://github.com/BrightspaceUI/core/issues/2075 brought up the issue that our dropdowns do not render when the contents are changed. This PR adds a ResizeObserver to the filter content slot to catch any resizes and re-position the dropdown.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -687,6 +702,18 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		await adjustPosition();
 	}
 
+	__startObservingContent() {
+		if (this.__mutationObserver) this.__mutationObserver.disconnect();
+		this.__mutationObserver = new MutationObserver(this.__position);
+		const slottedNodes = this.__getContentSlottedNodes();
+		for (let i = 0; i < slottedNodes.length; ++i) {
+			this.__mutationObserver.observe(slottedNodes[i], {
```
annabelsegalld2l, line 710: I'm not sure about the performance impacts of this idea - it'll re-render whenever the contents change. For example, the filter adds the dimensions one-by-one when they are loaded asynchronously. This will cause us to run position every time.

dbatiste:
The dropdown appears to resize based on the screenshot in the referenced issue.  I think this must be due to updating the element with the `d2l-dropdown-content-container` class (i.e. by setting the `max-height`).  Do we know why the overflow/scroll isn't being updated at the same time?

components/dropdown/dropdown-content-mixin.js
```diff
@@ -374,6 +378,13 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this._showBackdrop = this._useMobileStyling && this.mobileTray;
 	}
 
+	renderOnResize(elem) {
```
annabelsegalld2l, line 381: TODO: add a comment here

components/dropdown/dropdown-content-mixin.js
```diff
@@ -403,6 +414,14 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		}
 	}
 
+	__disconnectResizeObserver() {
+		this.renderCount++;
+		if (this.el.offsetHeight !== 0 && this.__resizeObserver) {
```
annabelsegalld2l, line 419: We are just checking height here as the div has a height of zero but a non-zero width. Not sure if it's worth adding another option for consumers to also check width since it's less common to have a zero width and non-zero height.

components/filter/filter.js
```diff
@@ -256,7 +258,10 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		let dimensionHTML;
 		switch (dimension.type) {
 			case 'd2l-filter-dimension-set':
-				dimensionHTML = html`<div aria-live="polite">${this._createSetDimension(dimension)}</div>`;
+				dimensionHTML = html`
+				<div aria-live="polite" id="${ifDefined(singleDimension ? this._filterContainerId : undefined)}">
```
annabelsegalld2l, line 262: Only adding this id if the filter has a single dimension, otherwise it will be duplicated causing axe errors.

components/filter/filter.js
```diff
@@ -544,7 +549,16 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 		if (shouldRecount) this._setFilterCounts(dimension);
-		if (shouldUpdate) this.requestUpdate();
+		if (shouldUpdate)  {
+			this.requestUpdate();
+
+			const singleDimension = this._dimensions.length === 1;
+			if (singleDimension && !this._waitingForResize) {
+				this._waitingForResize = true;
+				const dropdown = this.shadowRoot.querySelector('d2l-dropdown-content');
+				dropdown.renderOnResize(this.shadowRoot.querySelector('#d2l-filter-container'));
```
annabelsegalld2l, line 559: We only need to call this once - when filters are initially loading into a single dimension filter.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -403,6 +414,14 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		}
 	}
 
+	__disconnectResizeObserver() {
+		this.renderCount++;
```
svanherk, line 418: Don't think we need this anymore

components/filter/filter.js
svanherk: Can we use a class instead?  And if any new dimensions are added that have the same problem, we can put the class on them as well.

components/filter/filter.js
svanherk: With HM is might get a bit more choppy with the loading, versus the demo page which is all loaded at the same time.  I'm thinking something like this would be good:
```
				if (dimension.searchValue) shouldSearch = true;
				shouldRecount = true;
				shouldResizeDropdown = true;
			} else if (prop === 'loading') {
				shouldResizeDropdown = true;
			}
		});

		if (shouldSearch) this._performDimensionSearch(dimension);
		if (shouldRecount) this._setFilterCounts(dimension);
		if (shouldUpdate) this.requestUpdate();
		if (shouldResizeDropdown) {
			if (this._dimensions.length === 1 && this.opened) {
				const dropdown = this.shadowRoot.querySelector('d2l-dropdown-content');
				const container= this.shadowRoot.querySelector('.d2l-filter-container');
				if (container) dropdown.renderOnResize(container);
		}
```
Then it only runs on relevant property updates (new values or changing loading states) and can run again if it goes back to loading state or if more options pop in.

components/filter/filter.js
```diff
@@ -509,7 +514,7 @@ class Filter extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		announce(this.localize('components.filter.clearAllAnnounce'));
 	}
 
-	_handleDimensionDataChange(e) {
+	async _handleDimensionDataChange(e) {
```
svanherk, line 517: Looks like this doesn't need to be async anymore

components/dropdown/dropdown-content-mixin.js
dbatiste: Is it possible to leverage the entries that are passed into the observer?  The purpose of those entries is to allow handlers to run logic using the provided dimensions without causing layout (which calling `offsetHeight` will).

components/dropdown/dropdown-content-mixin.js
```diff
@@ -403,6 +417,20 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
 	}
 
+	__disconnectResizeObserver(entries) {
+		for (let i = 0; i < entries.length; i++) {
+			const entry = entries[i];
+			if (this.__resizeObserver && entry.contentRect.height !== 0) {
+				this.__resizeObserver.disconnect();
+				// wrap in rAF for Firefox
+				requestAnimationFrame(() => {
+					this.__position();
```
svanherk, line 427: Should we confirm the dropdown is still open right before we call this?  In case the resize is because they're closing the dropdown and then this is unnecessary?

PR #2086, opened 2022-02-07
fix: switch to use dynamic mixin

dlockhart:
The use of `for await` here was causing [esbuild to complain](https://github.com/evanw/esbuild/issues/1930) (it's used by `@web/dev-server`) in Firefox.

Since the languages passed into this function are a static list, I don't actually think `for await` was required. Either way, switching this to use the newer `LocalizeDynamicMixin` also resolves this.

There are [a bunch of places](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22getLocalizeResources%22+%22for+await%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5&si=full&si=full) that copied this though, so they may also need to be updated.

### DIFF
--- a/lang/localize-core-element.js
+++ b/lang/localize-core-element.js
@@ -1,74 +1,11 @@
-import { LocalizeMixin } from '../mixins/localize-mixin.js';
+import { LocalizeDynamicMixin } from '../mixins/localize-dynamic-mixin.js';
 
-export const LocalizeCoreElement = superclass => class extends LocalizeMixin(superclass) {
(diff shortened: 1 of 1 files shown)


dlockhart:
Looking into the test failures.

dbatiste (approved):

dlockhart:
This caused a lot of timing issues in our tests in non-Chrome, so closing this in favour of a [BSI solution](https://github.com/Brightspace/brightspace-integration/pull/5697).

PR #2098, opened 2022-02-08
Add d2l-selection-action-dropdown.

dbatiste:
This PR adds a new `d2l-selection-action-dropdown` component, that like other selection components supports the `requires-selection` attribute to control the enabled/disabled behaviour based on the selection state.  It further enhances the `d2l-overflow-group` to respect this opener and its children.

It can be used in conjunction with any dropdown content in the same fashion as the existing dropdown openers.


components/overflow-group/overflow-group.js
```diff
@@ -67,11 +67,14 @@ function createMenuItemMenu(node) {
 		||  node.querySelector('d2l-dropdown-button-subtle');
 
 	const openerText = node.text || menuOpener.text;
+	const disabled = node.disabled;
```
dlockhart, line 70: For things that aren't inherently disable-able, this could be `undefined` I think. Maybe `!!node.disabled`?

components/selection/selection-action-dropdown.js
```diff
@@ -0,0 +1,55 @@
+import '../button/button-subtle.js';
+import { html, LitElement } from 'lit-element/lit-element.js';
+import { DropdownOpenerMixin } from '../dropdown/dropdown-opener-mixin.js';
+import { dropdownOpenerStyles } from '../dropdown/dropdown-opener-styles.js';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { LocalizeCoreElement } from '../../lang/localize-core-element.js';
+import { SelectionActionMixin } from './selection-action-mixin.js';
+
+/**
+ * A "d2l-button-subtle" opener for dropdown content.
```
dlockhart, line 10: Maybe add "for use with selection"?

dlockhart (approved):

dbatiste:
I noticed the `d2l-selection-action` and `d2l-selection-action-dropdown` components were not getting the expected spacing within the overflow group like other components - added that to the CSS.

PR #2108, opened 2022-02-11
fix: Daylight: Minor documentation tweaks to table, input-search, and selection-action-dropdown

margaree:
Changes:
- search: make h3 title consistent with other components
- selection-action-dropdown: the other selection components hide this event with the `Internal event` description so this change does the same
- table: the second properties table is shown in Daylight so this makes the format of the tables on this page consistent with the generated properties tables


dlockhart (approved):

margaree:
Sauce CI tests seem to have passed but are getting the error:
```
ERROR webdriver: Request failed with status 404 due to Error: ERROR The test with session id 49727bf4029948b7971c8c5865bfa02a has already finished, and can't receive further commands.
```

PR #2112, opened 2022-02-14
Add no-selection option to the d2l-list-header component.

dbatiste:
This PR adds a new property to the `d2l-list-header` component so that it can be used with lists that do no allowed for selection.  

I contemplated solutions that infer whether there are selectable items in the list by piggybacking on some of the existing selection-info logic that bubbles selected item info (which doesn't seem right to me), however it does not account for the empty list case.  All these considered, I decided to go with the simplest approach of adding the property.


components/list/list-header.js
```diff
@@ -15,12 +15,10 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 		return {
 			/**
-			 * @ignore
-			 * Whether to render a header with reduced whitespace
-			 * TODO: Remove
+			 * Whether to render select-all and selection summary
 			 * @type {boolean}
 			 */
-			slim: { reflect: true, type: Boolean },
+			noSelection: { type: Boolean, attribute: 'no-selection' },
```
dlockhart, line 21: Should default to `false` in the constructor.

components/list/README.md
```diff
@@ -311,6 +311,7 @@ The `d2l-list-header` component can be placed in the `d2l-list`'s `header` slot
 
 | Property | Type | Description |
 |---|---|---|
+| `no-selection` | Boolean | Whether to render select-all and selection summary |
```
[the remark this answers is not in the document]
dbatiste: Yeah, they can put actions in the header that don't require selection, and the header would still be providing some layout such as spacing and the overflow-group.

components/list/list-header.js
dbatiste: Ah yeah, thanks.

dlockhart (approved):

PR #2115, opened 2022-02-14
Add support for newline characters in d2l-dialog-confirm text.

dbatiste:
This PR adds support for newline characters in `d2l-dialog-confirm` `text`.  One paragraph will be rendered for each line of text provided.  Consumers will need to construct their text in HTML using `&#10;` or in JavaScript using `\n`.

In the future we may pull this out and combine it with other simple formatting if we ever support some type of SML in our components.


dlockhart (approved):
Nice! I like where this ended up.

dbatiste:
> Nice! I like where this ended up.

Thanks for the suggestion!

PR #2118, opened 2022-02-15
Add methods for highlighting and scrolling to list-items.

dbatiste:


components/list/list-item-mixin.js
```diff
@@ -344,6 +362,20 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 		if (node) node.focus();
 	}
 
+	async highlight() {
```
dlockhart, line 365: Do you want to short-circuit out of here if `this._highlight` or `this._highlighting` is already `true`? i.e. if it's already in the process of highlighting.

components/list/list-item-mixin.js
```diff
@@ -354,6 +386,19 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 		});
 	}
 
+	scrollTo() {
```
dlockhart, line 389: So `scrollTo()` is already a [built-in DOM method](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) on all elements. Probably don't want to override it?

dbatiste: Yeah, maybe `scrollToItem`?  I'll wait until I hear back from @alexli01 re the API in case there are other changes to make.

components/list/list-item-mixin.js
```diff
@@ -443,6 +488,8 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 		const classes = {
 			'd2l-visible-on-ancestor-target': true,
 			'd2l-list-item-content-extend-separators': this._extendSeparators,
+			'd2l-list-item-highlight': this._highlight,
+			'd2l-list-item-highlighting': this._highlighting,
```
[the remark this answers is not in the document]
dbatiste: I think that would be possible, but I think we've tried to limit sprouting publicly unless we need to target the host specifically.  That's probably not strictly true, but that's the rationale.  It does result in complex selectors, but they are already complex.

PR #2120, opened 2022-02-16
fix: Primary-secondary template: Remove -webkit-overflow-scrolling 

margaree:
It is causing issue with fixed position dialog

Momentum scrolling on iOS was enabled by default as of iOS 13 and so `-webkit-overflow-scrolling: touch` no longer as an effect. In addition, it is not supported by or recommended for any browsers ([caniuse](https://caniuse.com/?search=-webkit-overflow-scrolling) and [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling)).

I have added some changes to the demo to reproduce this issue. If someone would be able to give this page a quick test on Android Chrome that would be amazing.

I tested this on:
- Firefox, Chrome, Safari (mac)
- Edge (Windows)
- Safari, Chrome (iPhone SE iOS 15.1)


dlockhart (approved):
👍  Don't have an Android device tho

annabelsegalld2l (approved):
Tested on my Android and the bug appears to be fixed 🎉 The dialog pops in front of the template divider as expected. The buttons at the bottom of the fullscreen dialog are still getting cut off. It's fullscreen specific and the mobile regular dialog view has the buttons positioned correctly. I think you saw the same on the iphone? We can file a separate defect for that.

dbatiste (approved):

margaree:
Awesome, thanks @annabelsegalld2l ! Yup I did see the same on iPhone with the cut off buttons in the full screen dialog. I'll open a defect for that.

PR #2123, opened 2022-02-16
fix: Dialog: Height calculation could be slightly off because of rounding

margaree:
An issue was reported in #web-platform about a scrollbar appearing for the confirm dialog at large font sizes. Digging into the height calculation, the footer had a height of 110.39 but that was being rounded down, causing the height to be slightly off and the scrollbar to appear. Using `getBoundingClientRect().height` instead of `scrollHeight` allow for us to get the demo value.

~I'm going to run this through visual diff before opening it for reviews in order to check if this breaks anything.~ Tests passed

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -197,7 +197,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		let preferredHeight = 0;
 
 		const header = this.shadowRoot.querySelector('.d2l-dialog-header');
-		if (header) preferredHeight += header.getBoundingClientRect().height;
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Cool looks good! Could probably also use the ceiling instead of just letting the browser round.

dbatiste:
So much for "The scrollHeight value is equal to the minimum height the element would require in order to fit all the content in the viewport without using a vertical scrollbar.".

According to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight), `getBoundingClientRect` apparently returns a fractional value.  Using `ceiling` still might be a good idea.

PR #2154, opened 2022-03-01
fix: Menu Item: Safari: Focus and hover styles

margaree:
Safari focus and hover styles on our menu items seem to have not been working since the 20.22.1 release. This seems to be related to Safari not yet supporting focus-visible. 

I was originally on Safari version 15.0 when I noticed this. I upgraded to 15.3 and it was still a problem. It looks like Safari has started supporting :focus-visible in the 15.4 tech preview.

It seems like Safari just refuses to use any blocks of css that contain `:focus-visible`, and separating this out fixes it. I then re-added the original `:focus` styles so that Safari users still get focus.

Open to suggestions if this should be solved in a different way. I'll check other places in core and see if there is anything else potentially impacted by this.


dlockhart:
This breakage was somewhat intentional due to performance issues with the polyfill:
https://github.com/BrightspaceUI/core/pull/2064

Safari is close to shipping native support, so we figured a temporary loss of focus styles was worth it.

margaree:
Are we also okay with losing the hover styles in this time too? I'm fine to re-remove the `:focus` piece if that was intended.

dlockhart (approved):

PR #2160, opened 2022-03-02
fix: allow host component to be hidden

mdulat:


components/selection/selection-action-dropdown.js
```diff
@@ -24,7 +24,11 @@ class ActionDropdown extends LocalizeCoreElement(SelectionActionMixin(DropdownOp
 	}
 
 	static get styles() {
-		return dropdownOpenerStyles;
+		return [dropdownOpenerStyles, css`
+			:host([hidden]) {
+				display: none;
+			}
+		`];
```
dlockhart, line 31: I think this can actually go with `dropdownOpenerStyles`... that way all openers will get the fix.

dlockhart (approved):

PR #2164, opened 2022-03-02
Add sticky behaviour to d2l-list-header.

dbatiste:
This PR adds sticky positioning behaviour to the `d2l-list-header`.  It is turned on by default, but it can be turned off by specifying the `no-sticky` attribute on the header.

Screenshot of it in the `d2l-template-primary-secondary`:

image


components/list/list-header.js
```diff
@@ -35,14 +42,36 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+			}
+			:host([no-sticky]) {
+				background-color: transparent;
+				position: static;
+				z-index: auto;
+			}
+			.d2l-list-header-shadow {
+				transition: box-shadow 200ms ease-out;
+			}
+			:host([_scrolled]) .d2l-list-header-shadow {
```
dbatiste, line 67: This fancy header shadow is used for a couple reasons - to cover up whitespace that is reserved for the drop target indicator; and also to have a fancy drop-shadow that fades at the edges so that it looks good when the header does not go edge to edge (ex. when consumer container has padding).

components/list/list-header.js
```diff
@@ -60,6 +89,9 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			:host([padding-type="slim"]) .d2l-list-header-container {
 				min-height: 36px;
 			}
+			.d2l-list-header-extend-separator {
```
dbatiste, line 92: This is just adding support for the extend-separators case that was missed when `d2l-list-header` was first introduced.

components/list/list.js
```diff
@@ -87,11 +93,26 @@ class List extends SelectionMixin(LitElement) {
 			}, 0);
 
 		});
+
```
dbatiste, line 96: The `d2l-list-header` needs to know when the list is scrolled and the header is sticking so that it can show/hide the shadow.  The way this is accomplished is: the list renders a sentinel element, and we use `IntersectionObserver` to detect when it intersects the view-port. The list then sets a private property on the `d2l-list-header` indicating that state.

components/list/test/list.visual-diff.html
```diff
@@ -208,7 +236,7 @@
 		<div style="width: 400px;">
 			<div class="visual-diff" id="selectableHeader">
 				<d2l-list>
-					<d2l-list-header slot="header"></d2l-list-header>
+					<d2l-list-header slot="header" no-sticky></d2l-list-header>
```
dbatiste, line 239: Lots of tests get kinda messed up if the headers are allowed to sticky on these fixtures.  So, we opt out of sticking for these tests so they do not break other tests.

components/list/test/list.visual-diff.js
```diff
@@ -99,26 +108,28 @@ describe('d2l-list', () => {
...
-			{ name: 'some selected header', selector: '#selectableSomeSelectedHeader' },
-			{ name: 'all selected header', selector: '#selectableAllSelectedHeader' },
-			{ name: 'all selected header pages allowed', selector: '#selectableAllSelectedHeaderPages' },
 			{ name: 'slim', selector: '#selectableSlim' },
 			{ name: 'item-content', selector: '#selectableItemContent' },
 			{ name: 'item-content slim', selector: '#selectableItemContentSlim' },
 			{ name: 'skeleton', selector: '#selectableSkeleton' },
 			{ name: 'extended separators', selector: '#selectableSeparatorsExtended' }
 		] },
+		{ category: 'header', tests: [
```
dbatiste, line 124: Just moved the header specific tests from above into this category, and added the two new tests for the sticky states.

templates/primary-secondary/demo/integration.html
```diff
@@ -0,0 +1,355 @@
+<!DOCTYPE html>
```
dbatiste, line 1: This is a new demo page that contains the primary-secondary template along with a couple complicated components.  It is useful to have this page both for testing and demo purposes.  The dialog was removed from the standard demo page.

components/list/README.md
```diff
@@ -312,6 +312,7 @@ The `d2l-list-header` component can be placed in the `d2l-list`'s `header` slot
 | Property | Type | Description |
 |---|---|---|
 | `no-selection` | Boolean | Whether to render select-all and selection summary |
+| `no-sticky` | Boolean | Whether to enable sticky positioning for the header |
```
dlockhart, line 315: Nit: technically it's not whether to enable it, but whether to disable it

```suggestion
| `no-sticky` | Boolean | Disables sticky positioning for the header |
```

components/list/list-header.js
```diff
@@ -93,14 +125,28 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 		this.slim = false;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+
+		const parent = findComposedAncestor(this.parentNode, node => node && node.tagName === 'D2L-LIST');
+		if (!parent) return;
+
+		this._extendSeparator = parent.hasAttribute('extend-separators');
```
dlockhart, line 140: Suggest defaulting this to `false` in the constructor otherwise it'll be `undefined` if no parent is found.

components/list/list.js
dlockhart: Haha "sentinel" sounds so ominous, love it.

components/list/list-header.js
```diff
@@ -19,6 +21,11 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			 * @type {boolean}
 			 */
 			noSelection: { type: Boolean, attribute: 'no-selection' },
+			/**
+			 * Whether to enable sticky positioning for the header
```
dlockhart, line 25:
```suggestion
			 * Disables sticky positioning for the header
```

dlockhart (approved):
So sticky!

PR #2165, opened 2022-03-02
feat: switching to use dynamic localize mixin to support Hindi

dlockhart:
In order to support Hindi, we would have needed a new entry in our `localize-core-element.js` mixin. Instead, we can just switch to using the sweet `LocalizeDynamicMixin` which will "just work" when new languages are added. Less code too!

While I was in here, I removed our Serge overwrite rule which will now [fall back to the default](https://github.com/Brightspace/serge-localize/blob/master/static/fra.common.serge.inc#L20). This required renaming `zh.js` -> `zh-cn.js`.


dbatiste (approved):
🎉

components/form/form-element-mixin.js
```diff
@@ -195,6 +204,11 @@ export const FormElementMixin = superclass => class extends LocalizeCoreElement(
 		if (this.noValidate) {
 			return [];
 		}
+
+		// validation can be requested before first update, which can cause issues
+		// if validation logic depends on rendered DOM
+		await this._firstUpdatePromise;
```
dlockhart, line 210: I originally thought I could use `await this.updateComplete` but unfortunately:
> If `shouldUpdate()` returns true, which it does by default, then the update proceeds normally. If it returns false, the rest of the update cycle will not be called but the `updateComplete` Promise is still resolved.

We're using `shouldUpdate` in our localization mixins to prevent rendering until the resources have loaded. But this weird behaviour here means that `updateComplete` can still resolve before the translations, which I was seeing here -- hence using my own Promise to track it.

I had to copy this same logic below into `form-mixin`... if this ends up being "a thing" we may want to factor it out into its own mixin. I'm surprised Lit doesn't provide something like this TBH.

dlockhart (commented):
I don't want to jinx this, but I think this is finally working and ready for review! 🎉 

Comments are inline describing the fixes, but mostly this is about the slight change in timing to load the dynamic localize mixin causing some downstream test failures almost entirely in Safari.

dlockhart:
Ok I did jinx it as the visual diffs are doing unexpected things. I'll have to look into those next week. 😵‍💫

PR #2171, opened 2022-03-03
Making Switch demos larger to fit the tooltip

geurts:


components/switch/README.md
```diff
@@ -32,7 +32,7 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 ![Switch](./screenshots/switch.png?raw=true)
 <!-- docs: end hidden content -->
 
-<!-- docs: demo live name:d2l-switch -->
+<!-- docs: demo live name:d2l-switch autoSize:false size:medium align:baseline -->
```
margaree, line 35:
```suggestion
<!-- docs: demo live name:d2l-switch autoSize:false size:small -->
```
Same suggestion for the one below; the `size:small` would be large enough to fit the tooltip below - medium looks quite large for these demos. It's also tall enough that you don't need the `align:baseline` to fit the tooltip when the switch is centred.

geurts: Right, sorry @margaree, I only meant to go up 1 size (I thought the default was small). This is what I get for making edits this way and not testing it first I guess.

margaree (approved):

PR #2172, opened 2022-03-03
Update padding for list-item actions when extend-separators is turned on.

dbatiste:
This PR fixes the layout for list-item actions for the case when `extend-separators` is turned on.  Previously, the last action would be jammed up against the edge of the container.

Updated:
image

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -193,6 +193,13 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 			.d2l-list-item-actions-container {
 				padding: 0.55rem 0;
 			}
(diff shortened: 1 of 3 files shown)


dbatiste:
This had an unintended result seen in the filter visual-diff: in the case where there are no actions, there is extra space.  We only want the extra whitespace applied when there are children in the actions slot/container.  Fix incoming.

dlockhart (approved):

svanherk:
> This had an unintended result seen in the filter visual-diff: in the case where there are no actions, there is extra space. We only want the extra whitespace applied when there are children in the actions slot/container. Fix incoming.

🎉 for visual diffs!!!

PR #2176, opened 2022-03-03
feat: implemented FocusMixin

dlockhart:
The need for this arose from #2165, where if the component hadn't initially rendered when `focus()` was called, the focus wouldn't happen.

This mixin is pretty simple -- the consumer provides it with a CSS selector and it implements the `focus()` method. Inside there, if the component hasn't yet rendered, it remembers that and when it _does_ render, will the use the selector to focus.

### DIFF
--- a/components/button/test/button-mixin.test.js
+++ b/components/button/test/button-mixin.test.js
@@ -1,4 +1,4 @@
-import { defineCE, expect, fixture, oneEvent } from '@open-wc/testing';
+import { defineCE, expect, fixture } from '@open-wc/testing';
 import { html, LitElement } from 'lit-element/lit-element.js';
 import { ButtonMixin } from '../button-mixin.js';
 
@@ -46,17 +46,6 @@ describe('ButtonMixin', () => {
 		it('should stop propagation of click events if button is disabled with disabled-tooltip', async() => {
 			const el = await fixture(`<${tagName} disabled disabled-tooltip="tooltip text"></${tagName}`);
 			expect(() => el.click()).to.not.throw();
-		});
-
-	});
-
-	describe('focus management', () => {
-
-		it('should delegate focus to button', async() => {
-			const el = await fixture(`<${tagName}></${tagName}`);
-			const buttonEl = el.shadowRoot.querySelector('button');
-			setTimeout(() => el.focus());
-			await oneEvent(buttonEl, 'focus');
 		});
 
 	});
--- a/components/button/test/button-subtle.axe.js
+++ b/components/button/test/button-subtle.axe.js
@@ -20,7 +20,7 @@ describe('d2l-button-subtle', () => {
 
 	it('normal + focused', async() => {
 		const el = await fixture(html`<d2l-button-subtle text="Subtle Button"></d2l-button-subtle>`);
-		setTimeout(() => el.shadowRoot.querySelector('button').focus());
+		setTimeout(() => el.focus());
 		await oneEvent(el, 'focus');
 		await expect(el).to.be.accessible();
 	});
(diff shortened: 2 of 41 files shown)


components/button/test/button-mixin.test.js
```diff
@@ -50,15 +50,4 @@ describe('ButtonMixin', () => {
...
-	describe('focus management', () => {
-
-		it('should delegate focus to button', async() => {
-			const el = await fixture(`<${tagName}></${tagName}`);
-			const buttonEl = el.shadowRoot.querySelector('button');
-			setTimeout(() => el.focus());
-			await oneEvent(buttonEl, 'focus');
-		});
-
-	});
```
dlockhart, line 62: No point in testing what the `FocusMixin` is doing.

components/card/test/card.axe.js
```diff
@@ -20,7 +20,8 @@ describe('d2l-card', () => {
 
 	it('default link + focused', async() => {
 		const elem = await fixture(html`<d2l-card text="Link Text" href="https://d2l.com"><div slot="content">Content</div></d2l-card>`);
-		elem.focus();
+		setTimeout(() => elem.focus());
+		await oneEvent(elem, 'focus');
```
dlockhart, line 24: I've switched tests that weren't using the `setTimeout` approach to use it. Without this, it's possible that focus hadn't yet been applied when the accessibility validation happened.

components/inputs/input-checkbox.js
```diff
@@ -205,11 +208,6 @@ class InputCheckbox extends SkeletonMixin(RtlMixin(LitElement)) {
 		`;
 	}
 
-	focus() {
-		const elem = this.shadowRoot && this.shadowRoot.querySelector('input.d2l-input-checkbox');
-		if (elem) elem.focus();
-	}
```
dlockhart, line 211: Pretty much all of these look like this. There were a couple components doing complex focus logic (like `<d2l-calendar>`) and I just left them as they were.

components/inputs/input-number.js
```diff
@@ -364,16 +367,6 @@ class InputNumber extends LabelledMixin(SkeletonMixin(FormElementMixin(LocalizeC
 		});
 	}
 
-	async focus() {
```
dlockhart, line 367: I didn't love that we changed the native `focus()` method to be `async`, which is why I left the `FocusMixin` implementation non-async. If folks need to be notified when it happens, they can wait for the event.

components/button/test/button-subtle.axe.js
```diff
@@ -20,7 +20,7 @@ describe('d2l-button-subtle', () => {
 
 	it('normal + focused', async() => {
 		const el = await fixture(html`<d2l-button-subtle text="Subtle Button"></d2l-button-subtle>`);
-		setTimeout(() => el.shadowRoot.querySelector('button').focus());
```
dbatiste, line 23: Awesome that we can reduce these reference to internal elements.  :)

components/inputs/input-number.js
dbatiste: I think @grant-cleary had to do something a little more conditional [in the editor](https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor.js#L346).  

The [underlying focus method](https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor-mixin.js) is in fact `async` in his case as well.  @grant-cleary ... you might be able to use what Dave has here... but I know your case is a little more complex.  It's not just waiting for the shadowRoot to render, it has to wait for tinyMCE and all that stuff.  Not sure if that's something you want to consider here Dave.

grant-cleary: Haha yeah, I was looking at this and wondering if I could make use of it. I think we're definitely still a bit hosed by having to wait for TinyMCE though.

dlockhart: Yeah I did consider adding more complex options -- like a method instead of a selector or for Grant's case we could expose some kind of Promise... but at the end of the day decided against it to a) keep this simple and b) because it's pretty easy for complex cases to just implement their own `focus()`.

grant-cleary: Aye, what the editor does now works out well enough. Unfortunate that it needs to do such weird things, but such is life when you rely on third party packages.

dbatiste: Sounds good!

dbatiste (approved):

PR #2195, opened 2022-03-08
Fullscreen dialog should go fully fullscreen (no margins)  when htmleditor inside

annabelsegalld2l:
This issue was spotted by Martin in this thread: https://d2l.slack.com/archives/C01RUTZ16EM/p1646698643546799


components/dialog/dialog-fullscreen.js
```diff
@@ -84,6 +84,12 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 					width: auto;
 				}
 
+				dialog.d2l-dialog-outer.d2l-dialog-fullscreen-within,
+				div.d2l-dialog-outer.d2l-dialog-fullscreen-within {
+					/* no margins when there is a fullscreen element within */
+					margin: 0;
```
annabelsegalld2l, line 90: https://github.com/BrightspaceUI/core/pull/2195/files#diff-e9552afeeb4d9451d8ff1940942a2414163e00d1cb2572664f60cd6d5ddc7187R78 Defines a 1.5rem margin around the dialog for a border. If there is a fullscreen element within (like HTML editor), we don't want the margins. Note that this fix is only for the wide screen layout, as the mobile layout was using regular `d2l-dialog` styles that didn't have this issue.

components/dialog/test/dialog-fullscreen.visual-diff.js
```diff
@@ -88,6 +88,43 @@ describe('d2l-dialog-fullscreen', () => {
 					await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
 				});
 
+				it('fullscreen-within-on', async function() {
```
annabelsegalld2l, line 91: Same tests as in `d2l-dialog`, making sure toggling the fullscreen-within class functions as expected.

PR #2200, opened 2022-03-09
fix: Fullscreen Dialog: show footer on mobile

margaree:
~Running through the visual diffs to make sure I didn't miss something before opening PR officially.~ 4 visual diffs are just slightly different but seem fine.

In order for the fullscreen dialog height to be properly calculated on mobile, it seems that we need to use `window.innerHeight` rather than relying on `1vh`, which returns a taller result. I'm open to suggestions on if there is a better way to do this. This does seem to fix the problem.

So far I've tested this on:
- Chrome (Mac)
- Chrome (iOS - where the bug was most apparent)
- Safari (iOS)


margaree:
Yup, I can absolutely make that change! Thanks for testing that!

components/dialog/dialog-fullscreen.js
```diff
@@ -226,6 +220,9 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 				: 0;
 			const startTop = mediaQueryList.matches ? 42 : 0;
 			topOverride = iframeTop + startTop;
+		} else if (window.innerWidth <= 615 || (window.innerWidth <= 900 && window.innerHeight <= 420)) {
+			heightOverride.height = `${window.innerHeight - 42 - 2}px`;
```
annabelsegalld2l, line 224: Nit: a comment here like "render full window height - 42px padding - 2px border"  would be helpful for future us 😄

annabelsegalld2l (approved):

PR #2202, opened 2022-03-09
adding messageformat-validator

dlockhart:
Note: base branch is my `fix-localizations` branch since otherwise this would conflict with it. This will stay draft until that merges.

This integrates @bearfriend's excellent `messageformat-validator` into core. It caught a few problems with how we're using the `zero` and `one` options in the `plural` format, which I was able to resolve pretty easily. It also caught a few untranslated strings, which I'll follow up on with the localization team.

Unfortunately we can no longer have our `localize-core-element.js` inside the `lang` directory since the validator tries to validate it, so I've moved it out into `helpers`.


components/filter/test/filter.test.js
```diff
@@ -68,7 +68,7 @@ describe('d2l-filter', () => {
 			await elem.updateComplete;
 
 			const infoMessage = elem.shadowRoot.querySelector('.d2l-filter-dimension-info-message');
-			expect(infoMessage.textContent).to.include('0 search results');
+			expect(infoMessage.textContent).to.include('No search results');
 			expect(infoMessage.classList.contains('d2l-offscreen')).to.be.false;
```
dlockhart, line 72: @svanherk I think this is proof that the `zero` option was getting ignored in English, and now it works!

lang/da.js
```diff
@@ -18,10 +18,10 @@ export default {
 	"components.filter.clearDescription": "Ryd filtre for: {filterName}",
 	"components.filter.clearDescriptionSingle": "Ryd filtre",
 	"components.filter.loading": "Indlæser filtre",
-	"components.filter.filterCountDescription": "{number, plural, zero {Ingen filtre anvendt.} one {1 filter anvendt.} other {{number} filtre anvendt.}}",
+	"components.filter.filterCountDescription": "{number, plural, =0 {Ingen filtre anvendt.} one {1 filter anvendt.} other {{number} filtre anvendt.}}",
```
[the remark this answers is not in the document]
svanherk: Huh... weird

components/filter/test/filter.test.js
svanherk: Nice, surprised I never noticed that when I made them 🤦‍♀️

svanherk (approved):
Looks good!  Should we get this repo into `BrightspaceUI`?  Or maybe we'd talked about this fully living in `actions`, I forget

dlockhart:
> Should we get this repo into BrightspaceUI? Or maybe we'd talked about this fully living in actions, I forget

Yeah that's a good question -- I'll start a thread to discuss.

PR #2203, opened 2022-03-09
feat: adding id-list attribute manipulation helpers

dlockhart:
Before we can change our button-like components to leverage `aria-describedby` for their `description` attributes, we need to fix `<d2l-tooltip>` (which also uses `aria-labelledby` and `aria-describedby`) so that it doesn't clobber any existing values in those attributes.

To solve that, I created two DOM helpers which can be used to manipulate space-separated ID-list attributes. I switched tooltip to use them.

### DIFF
--- a/helpers/dom.js
+++ b/helpers/dom.js
@@ -7,6 +7,53 @@ export function cssEscape(val) {
 	return val;
 }
 
+export function elemIdListAdd(elem, attrName, value) {
+
+	if (elem === undefined || elem === null || !elem.getAttribute || !elem.setAttribute) {
+		throw new TypeError('elemIdListAdd: "elem" must be a valid DOM Element');
+	}
+	if (typeof(attrName) !== 'string') {
+		throw new TypeError('elemIdListAdd: "attrName" must be a valid string');
+	}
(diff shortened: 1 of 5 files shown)


helpers/dom.js
```diff
@@ -7,6 +7,53 @@ export function cssEscape(val) {
...
+		throw new TypeError('elemIdListRemove: "value" must be a valid ID string');
+	}
+
+	const existingValue = elem.getAttribute(attrName) || '';
+
+	const parts = existingValue.split(' ');
+	const index = parts.indexOf(value);
+	if (index === -1) return;
+
+	parts.splice(index, 1);
```
dbatiste, line 48: Nit: it looks like `splice` could be avoided if `parts.length === 1`...?

```javascript
if (parts.length === 1) {
    elem.removeAttribute(attrName);
} else {
    parts.splice(index, 1);
    elem.setAttribute(attrName, parts.join(' '));
}
```

dlockhart: Agreed, thanks -- updated.

dbatiste (approved):

PR #2205, opened 2022-03-09
Attempt 2 for fixing stacking issue with fullscreen editor in list items.

dbatiste:
This PR is the 2nd attempt to work around some buggy behaviour in Safari.  The stacking context is messed up - the fullscreen editor should be on top.  This scenario is more complex (fullscreen editor in list-item in scroll-wrapper in fullscreen dialog) than the previous scenario, though it is not clear why Safari is messing up.

Before the fix:
image

After the fix:
image


dlockhart (approved):
😵‍💫

PR #2206, opened 2022-03-09
feat: button descriptions use aria-describedby

dlockhart:
As requested by @aid848, this switches our button-like components to use `aria-describedby` for their descriptions instead of `aria-label`. We feel as though having `description` be a supplementary label and not a primary label is both more to the spirit of the attribute name and more useful overall.

I've [audited the places](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F628716941319) using `description` (there were 8) and none of them were of concern or [have been addressed](https://github.com/Brightspace/lms/pull/20344).

### DIFF
--- a/components/button/README.md
+++ b/components/button/README.md
@@ -51,7 +51,7 @@ The `d2l-button` element can be used just like the native button element, but al
 
 | Property | Type | Description |
 |--|--|--|
-| `description` | String | A description to be added to the `button` for accessibility |
+| `description` | String | A description to be added to the `button` for accessibility for additional context |
 | `disabled` | Boolean | Disables the button |
 | `disabledTooltip` | String | Tooltip text when disabled |
 | `primary` | Boolean | Styles the button as a primary button |
@@ -130,6 +130,7 @@ The `d2l-button-icon` element can be used just like the native `button`, for ins
 
 | Property | Type | Description |
 |--|--|--|
+| `description` | String | A description to be added to the `button` for accessibility for additional context |
 | `icon` | String, required | [Preset icon key](../../components/icons#preset-icons) (e.g. `tier1:gear`) |
 | `text` | String, required | Accessible text for the buton |
 | `disabled` | Boolean | Disables the button |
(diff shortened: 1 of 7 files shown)


components/button/README.md
```diff
@@ -51,7 +51,7 @@ The `d2l-button` element can be used just like the native button element, but al
 
 | Property | Type | Description |
 |--|--|--|
-| `description` | String | A description to be added to the `button` for accessibility |
+| `description` | String | A description to be added to the `button` for accessibility for additional context |
```
dlockhart, line 54: Just alignment with the others.

components/button/button-icon.js
```diff
@@ -17,6 +17,12 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 
 	static get properties() {
 		return {
+			/**
+			 * A description to be added to the button for accessibility when text on button does not provide enough context
+			 * @type {string}
+			 */
+			description: { type: String },
```
dlockhart, line 24: `<d2l-button-icon>` didn't have a `description`, so I've added it here for consistency since technically it could be useful.

components/button/button-subtle.js
```diff
@@ -182,6 +185,7 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 				<span class="d2l-button-subtle-content">${this.text}</span>
 				<slot></slot>
 			</button>
+			${this.description ? html`<span id="${this._describedById}" hidden>${this.description}</span>` : null}
```
dlockhart, line 188: I'm using `hidden` here instead of offscreen as @dbatiste noticed that hidden elements can act in this way, which is pretty sweet. Tested this in NVDA and Voiceover.

dbatiste (approved):
I think we have access to a couple of Jaws licenses.  I'm not sure how we go about using them though - do we need to remote into a machine, or do we need someone with a Windows machine to give it a try?

dlockhart:
From @annabelsegalld2l:
> Tested JAWS, NVDA on Windows FF, Edge, Chrome

Thanks Annabel, you rock. Going to merge & release this!

PR #2208, opened 2022-03-10
LIT2 MIGRATION DO NOT MERGE

Vitalii-Misechko:


annabelsegalld2l:
From my understanding, this is a test PR to get a branch of core to test with in the other repos, correct? If so, I think that we can close it and not delete the branch, and your semver commands will still work in the other repos `package.json` files. We will not be merging this so there is no review needed from our side!

Issue #2096, opened 2022-02-08
`d2l-list` and `d2l-scroll-wrapper`: Dragging to the last spot adds padding causing the scroll wrapper to start vertical scrolling


MitchelPaulin:
This is likely only a problem with the `padding-type="none"` option. Dragging to the last spot causes the size of the last element to increase slightly and the scroll wrapper to start vertical scrolling creating a stuttering effect. 

#### Known Workaround 
Adding 6px of padding to the last list item can mitigate the issue. 

![Animation5](https://user-images.githubusercontent.com/30714020/153027352-9e112a1f-b291-424f-ba2a-b2b3c2b6707f.gif)

dbatiste:
We've had a bit of conversation on this.  Currently the marker end-caps are 12px high and it requires some whitespace below the last item to display correctly.  We never really expected the list component to be used in this way.  Nevertheless, we are considering some alternate solutions for the end caps and probably moving the marker up further.  For the moment though, I think the app needs to reserve some space for it until we can get a few other list changes done.

PR #2211, opened 2022-03-10
Make vertical scrolling behavior on scroll wrapper configurable

MitchelPaulin:
Being able to set `overflow-y: hidden` mostly resolves this issue https://github.com/BrightspaceUI/core/issues/2096

### DIFF
--- a/components/scroll-wrapper/scroll-wrapper.js
+++ b/components/scroll-wrapper/scroll-wrapper.js
@@ -57,7 +57,7 @@ class ScrollWrapper extends FocusVisiblePolyfillMixin(RtlMixin(LitElement)) {
 				box-sizing: border-box;
 				outline: none;
 				overflow-x: auto;
-				overflow-y: var(--d2l-scroll-wrapper-overflow-y, auto);
+				overflow-y: var(--d2l-scroll-wrapper-overflow-y, visible);
 			}
 			.d2l-scroll-wrapper-container.focus-visible {
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine), 0 2px 12px 0 rgba(0, 0, 0, 0.15);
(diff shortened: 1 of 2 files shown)


components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -59,6 +67,18 @@ class ScrollWrapper extends FocusVisiblePolyfillMixin(RtlMixin(LitElement)) {
...
+			}
+			:host([overflow-y="clip"]) .d2l-scroll-wrapper-container {
+				overflow-y: clip;
+			}
+			:host([overflow-y="scroll"]) .d2l-scroll-wrapper-container {
+				overflow-y: scroll;
+			}
+			:host([overflow-y="auto"]) .d2l-scroll-wrapper-container {
+				overflow-y: auto;
+			}
```
MitchelPaulin, line 81: Doing it this way avoids having to use `unsafeCSS`

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -53,11 +53,11 @@ class ScrollWrapper extends FocusVisiblePolyfillMixin(RtlMixin(LitElement)) {
 			:host([hidden]) {
 				display: none;
 			}
-
 			.d2l-scroll-wrapper-container {
 				box-sizing: border-box;
 				outline: none;
 				overflow-x: auto;
+				overflow-y: var(--d2l-scroll-wrapper-overflow-y, visible);
```
dbatiste, line 60: I think you tried `auto` originally and then switched to `visible`.  Did you run into an issue with `auto`?

MitchelPaulin: Yes, auto failed the vdiff tests, it looked related as well since it was a failure associated with the wrapper. These failures now are associated with some unrelated component

dbatiste (approved):

PR #2212, opened 2022-03-10
Visual-diff: Input-time: Use timeout to fix flakiness

margaree:
Ran several times locally and with this pattern I didn't see flake. I'll try several times here too and we'll see.


margaree:
This has passed visual diff several times in a row so I think it's stable. The change to the helper didn't end up helping but cleans things up a bit.

I also tried listening for `d2l-input-time-dropdown-toggle` from `input-time` and awaiting an updatecomplete but that was still flaky. The timeout seems to be most reliable. I can try reducing it as well - 100 was just a starting point.

dlockhart (approved):

PR #2220, opened 2022-03-14
US137000 - first attempt at number fix

annabelsegalld2l:


components/inputs/test/input-number.test.js
```diff
@@ -366,6 +367,20 @@ describe('d2l-input-number', () => {
 
 			expect(fired).to.be.false;
 		});
+
+		it('should set value to undefined', async() => {
```
annabelsegalld2l, line 371: Can't get the tests to line up correctly with the fix yet. Testing in engagement-dashboard shows me the correct behaviour now (blank instead of zero), but I can't get the tests set up to match that behaviour 😢

dlockhart: Not sure if you've tried this: reach into the inner `<input type="text">` and blank out the value, then trigger a `change` event originating from it -- essentially trying to simulate the user clearing it.

dlockhart (approved):
LGTM -- good job tracking this down... at least it was something you'd seen before!

dlockhart:
Looks like the lit-analyzer isn't a happy camper:
> 'numberConverter' is not a valid type for the default converter. Have you considered '{attribute: false}' instead?
    170:  value: { type: numberConverter },
    no-incompatible-property-type

components/inputs/test/input-number.test.js
```diff
@@ -368,6 +390,33 @@ describe('d2l-input-number', () => {
...
+				const elem = await fixture(inputWrapper);
+				elem.number = number;
+				await elem.updateComplete;
+				const inputVal = elem.shadowRoot.querySelector('d2l-input-number').value;
+
+				expect(inputVal).to.equal(number);
+			});
+		});
+
+		[undefined, null].forEach((number) => {
```
annabelsegalld2l, line 407: Summary of test results:

Lit 1:
With default number converter: Passes
With custom converter: Passes

Lit 2: 
With default number converter: Fails due to Lit internal attribute conversion changes (sets to 0 instead of undefined)
With custom converter: Passes

components/inputs/input-number.js
```diff
@@ -17,6 +17,12 @@ const HINT_TYPES = {
 	INTEGER: 4
 };
 
+// US137000 - prevent Lit default converter from converting undefined to 0
+const numberConverter = {
```
annabelsegalld2l, line 21: Reasoning behind these changes, for future reference: 

This has to do with the changes in how attributes are converted to properties in Lit 2.
In Lit 1: https://lit.dev/docs/v1/components/properties/#conversion For Numbers, when the attribute is defined, set the property to `Number(attributeValue)`.

In Lit 2: https://lit.dev/docs/components/properties/#conversion-type If the element has the corresponding attribute, set the property to `Number(attributeValue)`.

In Lit 1 passing in `undefined` does not affect the property (since it's undefined).
In Lit 2, the undefined gets automatically converted to 0.

To emulate the Lit 1 behaviour for compatibility reasons, we need to create this custom converter that converts null/undefined to undefined, and parses numbers as expected.

dlockhart (approved):

Issue #2095, opened 2022-02-08
`d2l-list` dragging between two different instances of `d2l-list` shows a blue line even though it should not be possible


MitchelPaulin:
If you have two instances of `d2l-list` on the same page and try to drag one element from one list to another list the blue line shows up implying it would be possible to do even though it should not be (note: dragging the item to another list is properly rejected, its only a visual bug).

![d2l-list](https://user-images.githubusercontent.com/30714020/153025977-053f2bc4-6320-476f-9a1e-8e0f55aa6394.gif)

dbatiste:
The feature was actually implemented with this in mind - that it would be possible to drag & drop items between any lists on the page, although preventing this behaviour was not because it was thought it would be unlikely that would have multiple lists on the page with drag & drop enabled.

dbatiste:
We can possibly make some changes so that d&d is only valid in the same list hierarchy for now, and figure out the API for x-root-list d&d when the time comes.

PR #2225, opened 2022-03-14
Added a drag group id to disable dragging if the groups have different ids

MitchelPaulin:
This addresses https://github.com/BrightspaceUI/core/issues/2095

Adds a single property `drag-group-id` to specify a drag group for inter list dragging. 

This approach allows for a lot of customization. For example if you have four lists on a page and you want to allow dragging between 1 and 2 and 3 and 4 but not with each other you would be able to accomplish it by giving 1,2 the same `drag-group-id` and 3,4 the same `drag-group-id`.

This fixes the issue with rubric issue while leaving a multi list drag possible to implement in the future. 

temp4

### DIFF
--- a/components/list/demo/list-drag-and-drop-position.js
+++ b/components/list/demo/list-drag-and-drop-position.js
@@ -13,7 +13,8 @@ class ListDemoDragAndDropPosition extends LitElement {
 			// below are for demonstration only
 			grid: { type: Boolean },
 			hrefs: { type: Boolean },
-			selectable: { type: Boolean }
+			selectable: { type: Boolean },
+			dragGroupId: { type: String, attribute: 'drag-group-id' }
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


dbatiste:
Hey Mitchel, I think I would rather hold off on adding this feature to the API and just enhance the validation that's currently being done. 

So for now, I think we should just add to the validation in [_onHostDragEnter](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-drag-drop-mixin.js#L674) to check if the source and target root lists are the same.  We can get a reference to the source root list from the first drag item (no need to check them all because they all must be from the same list hierarchy).  There is a [_getRootList](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-mixin.js#L467) method we can use. 

I can put together a PR to do this, or if you are inclined, that's ok too.

PR #2226, opened 2022-03-14
Disable drag events that cross list instances

MitchelPaulin:


components/list/list-item-drag-drop-mixin.js
```diff
@@ -679,8 +679,12 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 		const invalidDropTarget = dragState.dragTargets.find(dragTarget => {
 			return isComposedAncestor(dragTarget, this);
 		});
+
 		if (invalidDropTarget) return;
 
+		// assert that both the source and target are from the same list - may allow this in the future
+		if (this._getRootList() !== dragState.dragTargets[0]?._getRootList()) return;
```
dbatiste, line 686: I think we should be careful using optional chaining.  We still support a couple browser versions older than [when it was introduced](https://caniuse.com/mdn-javascript_operators_optional_chaining).

dlockhart: So Rollup + Babel will transpile this for us in BSI. The larger concern about this was that `polymer-cli` explodes when we use it, which blocks `polymer test` from working. I broke that with another change a couple weeks ago and am going to revert that today.

dbatiste (approved):

PR #2228, opened 2022-03-14
fix: going back to a polymer-cli friendly static getter

dlockhart:
Like optional chaining, `polymer-cli` hates the new-style static getters. While this is a great opportunity to force repos to switch to `@web/test-runner` and remove `polymer-cli` entirely, it could slow down the Lit 2 upgrade. So for now, I'm putting this back to use the old syntax.


annabelsegalld2l (approved):
Thanks for reverting this change Dave 😄 Happy to put it back in a few weeks when the Lit migration is over.

PR #2230, opened 2022-03-15
DE47655 Remove word "group" from start/end date widget in screen readers

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F628618321793

## Description

> Problem - Actual Behaviour:
>
> Navigating to the Due/Start/End date fields using a screenreader announces the label + selector group. Audio is "Group Start Date" or "Group Due Date" which is confusing as a group is an actual concept within Brightspace, but does not apply here. 
>
>
> Expected Behaviour:
>
> The component group should not be announced through the screenreader and only read out the actual label for Start Date and then the date selector.

### Problem Location

[The "group" keyword is associated with the fieldset tag in JAWS](https://freedomscientific.github.io/VFO-standards-support/html.html#toc-f).

The fieldset in question is found in [`d2l-input-fieldset`](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-fieldset.js), which is a core UI component within another core UI component, [`d2l-input-date-time`](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-date-time.js).

The problem occurs twice in [`d2l-activity-availability-dates-editor`](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-availability-dates-editor.js), once for the start date and end date each, and once in the [`d2l-activity-due-date-editor`](https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-editor/d2l-activity-due-date-editor.js).

As the `d2l-input-date-time` component may be used elsewhere, the fix should be toggled off by default to maintain behaviour for other uses.

### Fix

The fix involves adding `role="presentation"` to the fieldset in `<d2l-input-fieldset>`, as well as making the "date/time" label in `<d2l-input-date>` and `<d2l-input-time>` more verbose.  The presentation role tells screen readers that the fieldset should not be read aloud, while allowing labels to be read for contained elements (unlike `aria-hidden="true"`).  These changes will only take place if the **suppress-group** property is true.

In JAWS, this removes the word "group".  In NVDA, this removes the word "grouping".
The text formerly read out by the fieldset's legend is now the label on the input date and time components themselves, preserving the information served while avoiding the term "group".

When **suppress-group** is not present, the components behave as they did without the changes.

#### JAWS

Before change/with fix toggled off:

> Edit-activity dash course dash google chrome, complimentary region, **start date group**, date edit has pop-up, m slash d slash yyyy, type and text, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

After change:

> Edit-activity dash course dash google chrome, complimentary region, **start date**, date edit has pop-up, m slash d slash yyyy, type and text, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

#### NVDA

Before change/with fix toggled off:

> Complimentary landmark, **start date grouping**, date edit submenu m slash d slash yyyy blank, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

After change:

> Complimentary landmark, **start date**, date edit submenu m slash d slash yyyy blank, use date format m slash d slash yyyy, Arrow down or press enter to access mini-calendar

## Video/Screenshots

Testing revealed the "group" keyword was associated with the fieldset tag, seen here within the `d2l-input-label-fieldset` class:

A11y Pre-Fix 3

With the changes described above:

A11y Fix 3

Note that when the optional fix is toggled, NVDA's element list becomes more navigable with the full labels.

Without `suppress-group`:

NVDA Pre-Fix

With `suppress-group`:

NVDA Fix 2

## Related PRs

- Activities: https://github.com/BrightspaceHypermediaComponents/activities/pull/2466

## Remaining Work

- [ ] Dev Testing: another developer will test this code on their machine

### DIFF
--- a/components/inputs/input-date-time.js
+++ b/components/inputs/input-date-time.js
@@ -236,6 +236,8 @@ class InputDateTime extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMi
 		const dateOpened = this.opened && !this._timeOpened && !this.disabled && !this.skeleton;
 		const parsedValue = this.value ? (this.localized ? this.value : getLocalDateTimeFromUTCDateTime(this.value)) : '';
 		const tooltip = (this.validationError && !this.opened && this.childErrors.size === 0) ? html`<d2l-tooltip align="start" announced for="${this._inputId}" state="error">${this.validationError}</d2l-tooltip>` : null;
+		const dateLabel = this.localize('components.input-date-time.date');
+		const timeLabel = this.localize('components.input-date-time.time');
 		const inputTime = !timeHidden ? html`<d2l-input-time
 				?novalidate="${this.noValidate}"
 				@blur="${this._handleInputTimeBlur}"
@@ -247,7 +249,7 @@ class InputDateTime extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMi
 				.forceInvalid=${this.invalid}
 				@mouseout="${this._handleInputTimeBlur}"
 				@mouseover="${this._handleInputTimeFocus}"
-				label="${this.localize('components.input-date-time.time')}"
+				label="${ifDefined(this.suppressGroup && this.label ? `${this.label}, ${timeLabel}` : timeLabel)}"
 				label-hidden
 				.labelRequired="${false}"
 				max-height="430"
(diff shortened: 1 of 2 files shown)


ChrisKraljevicD2L (approved):

dlockhart:
There are still email conversations happening about this change, so please don't merge anything yet -- I'm still not convinced that this is a change we want to make.

Kiefer-Co-D2L:
> There are still email conversations happening about this change, so please don't merge anything yet -- I'm still not convinced that this is a change we want to make.

No worries, we won't be merging this without @BrightspaceUI/gaudi-dev approval!  We're still waiting to hear back from other correspondents about what exact screen reader text would be ideal.

Kiefer-Co-D2L:
Based on discussion in Slack:

> it looks like based on Carin's screenreader expertise we should be fine keeping the current "Start Date Group" and "Start Date Grouping" behaviour. So we are going to just end up closing this for now.

PR #2233, opened 2022-03-17
Adjust dialog resizing timing to ensure contents have time to render first

grant-cleary:
There are a couple issues at play here, which I'll comment on a bit more inline where they're relevant. The gist of these issues, however, is that Lit elements rendered inside dialogs don't always have time to render fully before our resizing kicks in, leading to a situation where the dialog height can end up rendered improperly on first resize.

This is most noticeable in Firefox, where it happened basically everywhere for all dialogs. If you render a Lit element and a native element inside a dialog in Firefox, its initial resize will only account for the native select element, as the Lit elements haven't rendered yet. Any kind of debugging attached prior to the first height calculation gives the elements sufficient time to render and the problem vanishes.


components/dialog/dialog-mixin.js
```diff
@@ -354,7 +359,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			this.dispatchEvent(new CustomEvent(
 				'd2l-dialog-open', { bubbles: true, composed: true }
 			));
-		});
+		}, 0);
```
grant-cleary, line 362: rAF doesn't give the contents time to render in Firefox specifically. This isn't the first time I've encountered situations where FF schedules microtasks differently from Chrome in a way that messes up timings. `setTimeout(, 0)` does do the trick though.

components/dialog/dialog-styles.js
```diff
@@ -11,6 +11,11 @@ export const dialogStyles = css`
 		display: block;
 	}
 
+	:host([opened][_state="showing"]),
+	:host([opened][_state="hiding"]) {
+		visibility: visible
+	}
```
grant-cleary, line 17: Shoot, I accidentally undid part of this commit, so this is incomplete.

components/dialog/dialog-styles.js
```diff
@@ -11,6 +11,15 @@ export const dialogStyles = css`
...
 	}
 
+	:host([opened]:not([_state="showing"])) {
+		visibility: hidden;
+	}
+
+	:host([opened][_state="showing"]),
+	:host([opened][_state="hiding"]) {
+		visibility: visible;
+	}
```
grant-cleary, line 21: This is necessary because of the timing change from the switch from rAF -> setTimeout. Before the dialog finishes opening (i.e. before the state transition starts), there's a brief moment where the dialog pops in at max width before the transition starts and the backdrop gets applied. This additional CSS ensures that the dialog only becomes visible once we start the state transition.

components/dialog/dialog-mixin.js
```diff
@@ -124,7 +124,12 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 	}
 
 	resize() {
-		return this._updateSize();
+		return new Promise(resolve => {
+			setTimeout(async() => {
+				await this._updateSize();
+				resolve();
+			});
+		});
```
grant-cleary, line 132: This resolves a separate issue I found both in Chrome and Firefox without my changes. The list item in the demo page doesn't finish rendering by the time the async dialog calls this resize method, hence the dialog sizes to zero height. It's a similar sort of problem and using rAF does resolve it in Chrome, but much like the other issue, we need `setTimeout(, 0)` to resolve this in Firefox.

components/dialog/dialog-styles.js
grant-cleary: Ah, sorry, wasn't clear. It sets `width: auto`, which in most cases fills the viewport.

grant-cleary:
> Visual diff failures are a bit odd... I wonder if focus is weird now?

Yep, just checked in the demo and focus is now not being applied properly; presumably because it's now happening too early.

grant-cleary:
Yeah okay, the focus issues come about because of the first `_focusInitial` call. Now that the dialog isn't visible until after, that first call focuses the document body.

grant-cleary:
- [ ] Don't merge me until `core` is bumped!

grant-cleary:
I'll investigate those fullscreen tests explicitly. Unsure why they would be behaving differently than the non-fullscreen ones, but I'll figure it out.

grant-cleary:
Ok, so for the fullscreen stuff... It looks like before my change, focus was being set on the close button, whereas now it's being set on the tabs (inside the dialog). This... Seems good? But it's a change in behaviour so I'm less sure.

dbatiste:
> Ok, so for the fullscreen stuff... It looks like before my change, focus was being set on the close button, whereas now it's being set on the tabs (inside the dialog). This... Seems good? But it's a change in behaviour so I'm less sure.

Yeah I am surprised that the focus was being places on the abort button.  It's supposed to go on the first focusable in the content.  Are you sure it's going to the tab though... if it is, the focus style would look like this:
image

grant-cleary:
> Yeah I am surprised that the focus was being placed on the abort button. It's supposed to go on the first focusable in the content. Are you sure it's going to the tab though... if it is, the focus style would look like this: image

`d2l-tabs` becomes the active element for sure when I check on the demo page, but the focused tab definitely isn't getting `focus-visible` applied.

grant-cleary:
> `d2l-tabs` becomes the active element for sure when I check on the demo page, but the first focused tab definitely isn't getting `focus-visible` applied. The latter ones definitely do when I arrow key around, and if I arrow key to the right, I end up on the second tab as I would expect.

I can also confirm that `focus()` is being called on the tab appropriately. I breakpointed into the `d2l-tabs` focus method and can confirm that it's selecting the right tab and focusing it. For some reason that's just not applying the focus-visible class.

dbatiste:
> > Yeah I am surprised that the focus was being placed on the abort button. It's supposed to go on the first focusable in the content. Are you sure it's going to the tab though... if it is, the focus style would look like this: image
> 
> `d2l-tabs` becomes the active element for sure when I check on the demo page, but the first focused tab definitely isn't getting `focus-visible` applied. The latter ones definitely do when I arrow key around, and if I arrow key to the right, I end up on the second tab as I would expect.

Ok great.  Yep I forgot about the focus-visible.

grant-cleary:
To re-open the loop (and then close it again), fixing the focus behaviour here is going to be a bit more work than initially expected, so I'm going to handle that in a separate PR.
- [ ] Wait for tabs focus fix

dbatiste (approved):

PR #2236, opened 2022-03-18
feat: Add d2l-tag-list, d2l-tag-list-item, and TagListItemMixin

svanherk:
This PR adds the basic tag-list components and simple keyboard navigation.  Additional functionality coming in future PRs.

### DIFF
--- a/components/tag-list/README.md
+++ b/components/tag-list/README.md
@@ -1,0 +1,52 @@
+
+# Tag List
+*This component is in progress and the api is not yet stable.*
+
+Tag lists are used to present a list of compact, discrete pieces of information.
+
+<!-- docs: demo display:block -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/tag-list/tag-list.js';
+  import '@brightspace-ui/core/components/tag-list/tag-list-item.js';
+</script>
+
+<d2l-tag-list description="Example Tags">
+  <d2l-tag-list-item text="Tag"></d2l-tag-list-item>
+  <d2l-tag-list-item text="Another Tag"></d2l-tag-list-item>
+  <d2l-tag-list-item text="A Third Very Very Very Long Tag"></d2l-tag-list-item>
+</d2l-tag-list>
+```
+
+## Tag List [d2l-tag-list]
+
+The `d2l-tag-list` element can take a combination of any type of `d2l-tag-list-item` and adds the appropriate keyboard navigation, list semantics, clearing behaviour and responsive behaviour.
+
+<!-- docs: demo live name:d2l-tag-list display:block -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/tag-list/tag-list.js';
+  import '@brightspace-ui/core/components/tag-list/tag-list-item.js';
+</script>
+
+<d2l-tag-list description="Example Tags">
+  <d2l-tag-list-item text="Tag"></d2l-tag-list-item>
+  <d2l-tag-list-item text="Another Tag"></d2l-tag-list-item>
+  <d2l-tag-list-item text="A Third Very Very Very Long Tag"></d2l-tag-list-item>
+</d2l-tag-list>
+```
+
+## Tag List Item [d2l-tag-list-item]
+The `d2l-tag-list-item` provides the appropriate `listitem` semantics and styling for children within a tag list. Tag List items do not work outside of a Tag List and should not be used on their own.
+
+<!-- docs: demo live name:d2l-tag-list-item display:block -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/tag-list/tag-list.js';
+  import '@brightspace-ui/core/components/tag-list/tag-list-item.js';
+</script>
+
+<d2l-tag-list description="Example Tags">
+  <d2l-tag-list-item text="Tag"></d2l-tag-list-item>
+</d2l-tag-list>
+```
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -38,7 +38,6 @@ export const TagListItemMixin = superclass => class extends superclass {
 				color: var(--d2l-color-ferrite);
 				cursor: pointer;
 				display: flex;
-				min-height: 30px;
 				min-width: 0;
 				padding: 0.25rem 0.6rem;
 				transition: background-color 0.2s ease-out, box-shadow 0.2s ease-out;
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -66,8 +66,6 @@ export const TagListItemMixin = superclass => class extends superclass {
 		super();
 		/** @ignore */
 		this.role = 'listitem';
-		/** @ignore */
-		this.tabIndex = -1;
 	}
 
 	_renderTag(tagContent) {
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -9,11 +9,7 @@ export const TagListItemMixin = superclass => class extends superclass {
 			/**
 			 * @ignore
 			 */
-			role: { type: String, reflect: true },
-			/**
-			 * @ignore
-			 */
-			tabindex: { type: Number, reflect: true }
+			role: { type: String, reflect: true }
 		};
 	}
 
@@ -45,8 +41,7 @@ export const TagListItemMixin = superclass => class extends superclass {
 				min-height: 30px;
 				min-width: 0;
 				padding: 0.25rem 0.6rem;
-				transition-duration: 0.2s;
-				transition-timing-function: ease-out;
+				transition: background-color 0.2s ease-out, box-shadow 0.2s ease-out;
 			}
 			:host(:hover) .tag-list-item-container,
 			:host(:focus) .tag-list-item-container {
@@ -72,7 +67,7 @@ export const TagListItemMixin = superclass => class extends superclass {
 		/** @ignore */
 		this.role = 'listitem';
 		/** @ignore */
-		this.tabindex = -1;
+		this.tabIndex = -1;
 	}
 
 	_renderTag(tagContent) {
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -1,0 +1,86 @@
+import '../colors/colors.js';
+import { css, html } from 'lit-element/lit-element.js';
+import { labelStyles } from '../typography/styles.js';
+
+export const TagListItemMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			/**
+			 * @ignore
+			 */
+			role: { type: String, reflect: true },
+			/**
+			 * @ignore
+			 */
+			tabindex: { type: Number, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [labelStyles, css`
+			:host {
+				display: grid;
+				outline: none;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.tag-list-item-content {
+				height: 1rem;
+				margin: auto;
+				min-width: 0;
+				overflow: hidden;
+				text-overflow: ellipsis;
+				white-space: nowrap;
+			}
+			.tag-list-item-container {
+				background-color: var(--d2l-color-regolith);
+				border-radius: 6px;
+				box-shadow: inset 0 0 0 1px var(--d2l-color-gypsum), 0 2px 4px rgba(0, 0, 0, 0.03);
+				box-sizing: border-box;
+				color: var(--d2l-color-ferrite);
+				cursor: pointer;
+				display: flex;
+				min-height: 30px;
+				min-width: 0;
+				padding: 0.25rem 0.6rem;
+				transition-duration: 0.2s;
+				transition-timing-function: ease-out;
+			}
+			:host(:hover) .tag-list-item-container,
+			:host(:focus) .tag-list-item-container {
+				background-color: var(--d2l-color-sylvite);
+			}
+			:host(:hover) .tag-list-item-container {
+				box-shadow: inset 0 0 0 1px var(--d2l-color-mica), 0 2px 4px rgba(0, 0, 0, 0.03);
+			}
+			:host(:focus) .tag-list-item-container {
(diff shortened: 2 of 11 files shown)


components/tag-list/test/tag-list-item-mixin-consumer.js
```diff
@@ -0,0 +1,73 @@
+import '../../dropdown/dropdown.js';
```
svanherk, line 1: This component is basically to help with development and is only used for the test and demo pages.  It's a stripped down version of what the `d2l-tag-list-item-user` will be, which will live in another repo.

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,97 @@
...
+		super.firstUpdated(changedProperties);
+
+		requestAnimationFrame(() => {
+			this._items = this._getTagListItems();
+			if (this._items.length > 0) this._items[0].setAttribute('tabindex', 0);
+		});
+	}
+
+	render() {
+		const description = this.description ? html`<div id="d2l-tag-list-description" class="d2l-offscreen">${this.description}</div>` : null;
```
svanherk, line 64: Looking to move this to `hidden` instead of offscreen

components/tag-list/tag-list-item.js
```diff
@@ -0,0 +1,25 @@
...
+			/**
+			 * REQUIRED: Text to display
+			 * @type {string}
+			 */
+			text: { type: String }
+		};
+	}
+
+	render() {
+		const tag = this._renderTag(this.text);
```
annabelsegalld2l, line 17: I think you can just return `this._renderTag(this.text)`, no need to wrap the html in another html.

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,96 @@
...
+			${description}
+		`;
+	}
+
+	async arrowKeysFocusablesProvider() {
+		return this._items;
+	}
+
+	focus() {
+		if (this._items[0]) this._items[0].focus();
```
dlockhart, line 80:
```suggestion
		if (this._items.length > 0) this._items[0].focus();
```

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,96 @@
...
+		super.firstUpdated(changedProperties);
+
+		requestAnimationFrame(() => {
+			this._items = this._getTagListItems();
+			if (this._items.length > 0) this._items[0].setAttribute('tabindex', 0);
+		});
+	}
+
+	render() {
+		const description = this.description ? html`<div id="d2l-tag-list-description" hidden>${this.description}</div>` : null;
```
dlockhart, line 63: Since `description` is required, you could simplify this to just always render it.

components/tag-list/README.md
```diff
@@ -0,0 +1,52 @@
+
```
dlockhart, line 1: You'll also want a link to this README from the root README at some point.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -0,0 +1,86 @@
...
+			}
+			.tag-list-item-container {
+				background-color: var(--d2l-color-regolith);
+				border-radius: 6px;
+				box-shadow: inset 0 0 0 1px var(--d2l-color-gypsum), 0 2px 4px rgba(0, 0, 0, 0.03);
+				box-sizing: border-box;
+				color: var(--d2l-color-ferrite);
+				cursor: pointer;
+				display: flex;
+				min-height: 30px;
```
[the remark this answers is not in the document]
svanherk: So the problem is the `user-profile-image` which is in `px`.  The new size `x-small` size will be `24px` with `3px` of padding top and bottom, so without this small `font-size`s result in tags of different sizes.  I brought up switching the `user-profile-image`s to `rem` as well with Jeff and Noah, I could raise that again if we think that's the better solution.

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,96 @@
...
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.tag-list-container {
+				display: flex;
+				flex-wrap: wrap;
+				margin: -${ITEM_SPACING}px -${ITEM_SPACING}px 0 0;
```
dbatiste, line 34: Did this work as expected?  I don't think using non `css` expressions is recommended here.  Perhaps take a peek at [this section of the Lit doc](https://lit.dev/docs/components/styles/#expressions) - perhaps use a CSS variable instead.

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
dlockhart: Hmm, so is the behaviour you want that all items are as tall as the tallest item? Because if so, you could use `align-items: stretch` on the outer list that's applying flexbox to its items.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -0,0 +1,86 @@
...
+				min-width: 0;
+				padding: 0.25rem 0.6rem;
+				transition-duration: 0.2s;
+				transition-timing-function: ease-out;
```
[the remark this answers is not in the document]
svanherk: This is to get a slight fade effect for the hover and focus effects (basically background colour and border colour).  If I add this just to `:focus` and `:hover` it doesn't do the transition when you stop hovering/focusing, only when first applied.  [The spec](https://xd.adobe.com/view/683962a6-87de-4658-b151-31b78a395030-3d6a/) actually wanted different timing for hover versus focus and I couldn't find a simple way to do that with pure css, but maybe there's a way I don't know about?  Noah was fine with making them the same for simplicity

components/tag-list/tag-list.js
svanherk: My worry was people might still forget to set it and then I could degrade gracefully, but I guess if we saw that happening we could decide to start throwing an error or something similar.  Will update!

components/tag-list/tag-list.js
svanherk: Odd, so this does work totally fine and Lit lets me do it.  But if I change it to string instead of a number, it gets mad.  I can wrap it in a css literal to be safe though.  I didn't use a variable because I didn't want people to be able to override it, this is just for ease of a developer updating it later (if needed).

components/tag-list/tag-list-item-mixin.js
```diff
@@ -0,0 +1,86 @@
...
+			 * @ignore
+			 */
+			tabindex: { type: Number, reflect: true }
+		};
+	}
+
+	static get styles() {
+		return [labelStyles, css`
+			:host {
+				display: grid;
```
dbatiste, line 23: How is the `grid` layout being used?  I don't see any other grid definition that I _think_ would be needed.

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
dlockhart: Ah OK, I'm used to seeing `transition: property-name duration timing-function`. I think this will essentially transition ALL properties potentially.

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
svanherk: It seems like that only takes affect for tags on the same line, like this:
image

Which doesn't look awful, but I was trying to avoid measuring the height of a tag on each line to figure out how many lines we have.  And in actuality, will we have different types of tags mixed together? Probably not.  But playing around it looks like I may have accidentally fixed this later on by adding a height of `1rem` to the inner content, which hides overflow.  Means that the profile image would overflow on small font sizes though, like this:
image
(this is root size of `14px`)

components/tag-list/tag-list.js
dlockhart: Yeah or throw an exception after a second or so if it's still unset like we do with input labels.

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
svanherk: I can switch it to that, that would make it more readable and it's only two properties anyways

components/tag-list/tag-list-item-mixin.js
```diff
@@ -0,0 +1,86 @@
...
+	static get properties() {
+		return {
+			/**
+			 * @ignore
+			 */
+			role: { type: String, reflect: true },
+			/**
+			 * @ignore
+			 */
+			tabindex: { type: Number, reflect: true }
```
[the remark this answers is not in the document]
svanherk: I find if I remove this, this doesn't get set as an attribute on the custom element and then I can't set the focus through JS.  But maybe I'm missing something?

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,96 @@
...
+			::slotted(*) {
+				margin: ${ITEM_SPACING}px ${ITEM_SPACING}px 0 0;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		/** @ignore */
+		this.role = 'application';
```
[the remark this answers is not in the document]
svanherk: Will switch to that!  I wasn't sure what was the greater of two evils... this or extra dom nodes lol

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
dlockhart: Hmm, maybe a casing thing. The JS property is `tabIndex` -- I'd set that in the constructor and later when you want to change it instead of the attribute.

components/tag-list/tag-list.js
```diff
@@ -0,0 +1,96 @@
...
+		this.arrowKeysDirection = 'leftrightupdown';
+		this._items = [];
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		requestAnimationFrame(() => {
+			this._items = this._getTagListItems();
+			if (this._items.length > 0) this._items[0].setAttribute('tabindex', 0);
```
dbatiste, line 58: Not that you necessarily need to support it, but technically the first tag could be `hidden`, in such case focus could never get into the tags list and arrow keys wouldn't work.  Just mentioning in case this was something that might be expected.

components/tag-list/tag-list-item-mixin.js
svanherk: So this was the only `display` value that would make the ellipsis truncating work properly with `flex`.  I unfortunately don't remember the details super well since I got all that working about a month and a half ago, but if I remove this the functionality completely breaks and the `nowrap` tags refuse to shrink and instead just overflow the container.  I remember trying a lot of combinations and this was the only one that worked, but I can try others if you have an idea!

components/tag-list/tag-list.js
svanherk: For sure, I'll definitely need to handle hidden tags anyways once we start truncating after some number of lines.  Haven't decided yet if that will be a separate `_visibleItems` array that I could use here instead, or if I want a `truncatingIndex` and I'll need to check whether anything before that is also hidden.

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
svanherk: Yeah menu-items confused me because they both sprout it, and menu adds it.  Looks like it was a casing issue, but I could also remove it from the items altogether and have the list add it to every single one.

components/tag-list/tag-list.js
```diff
@@ -1,8 +1,8 @@
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { ArrowKeysMixin } from '../../mixins/arrow-keys-mixin.js';
-import { ifDefined } from 'lit-html/directives/if-defined.js';
 
-const ITEM_SPACING = 6;
+//const ITEM_SPACING = '6';
+const CSS_ITEM_SPACING = css`6`;
```
dlockhart, line 5: Nit: 
```suggestion
const CSS_ITEM_SPACING = css`6`;
```

components/tag-list/tag-list-item-mixin.js
```diff
@@ -72,7 +67,7 @@ export const TagListItemMixin = superclass => class extends superclass {
 		/** @ignore */
 		this.role = 'listitem';
 		/** @ignore */
-		this.tabindex = -1;
```
dlockhart: Nit: I'm not totally sure, but probably don't still need the `@ignore` on this as I don't _believe_ the docs care about non-properties? Maybe I'm wrong though.

dlockhart (approved):

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
dbatiste: Oh wow... `MenuItemMixin` also uses lower case.  We should probably correct that.  Probably isn't an issue since `d2l-menu` sets it anyway.  Looks like `d2l-tab-internal` uses the correct case.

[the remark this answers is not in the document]
svanherk: So it looks like our constructor tests hate this, which is maybe why `MenuItemMixin` uses lowercase?  And `d2l-tab-internal` [doesn't have a constructor test](https://github.com/BrightspaceUI/core/blob/main/components/tabs/test/tabs.test.js#L18)...

components/tag-list/tag-list.js
svanherk: Even though [Lit docs](https://lit.dev/docs/components/styles/#expressions) say to do this, [our linter hates it lol](https://github.com/BrightspaceUI/core/runs/5606443294?check_suite_focus=true#step:7:10).  Starting to think this isn't worth it lol, and if we ever want to change the 6 we can change the variable and the 4 places in css 😒

components/tag-list/tag-list-item-mixin.js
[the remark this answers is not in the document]
dbatiste: So it would probably 💥 if it were constructed with `createElement`...
image
`MenuItemMixin` using lower case seems like a mistake, since it wouldn't really do anything.  We just never noticed because `d2l-menu` assigns the attribute anyway.

components/tag-list/tag-list.js
svanherk: Just going to remove having this as a variable.  Our style linter doesn't allow for what the Lit docs recommends - it looks at all `css` tags and checks for valid css (meaning something like `whatever { color: red; }`, not `red`).

PR #2237, opened 2022-03-18
Fix some typos, add some more detail to ArrowKeysMixin README

svanherk:
Noticed this when using the mixin for tags - I think the current README example just wouldn't work?  Or am I missing something?

Also fixed a typo in the function name.

### DIFF
--- a/mixins/arrow-keys-mixin.md
+++ b/mixins/arrow-keys-mixin.md
@@ -12,14 +12,14 @@ If the elements are known up-front and are in the element's local DOM scope, sim
 import { ArrowKeysMixin } from '@brightspace-ui/core/mixins/arrow-keys-mixin.js';
 class MyElement extends ArrowKeysMixin(LitElement) {
   render() {
-    return html`
+    return this.arrowKeysContainer(html`
       <div>
         ...
         <a href="..." class="d2l-arrowkeys-focusable">link 1</a>
         <a href="..." class="d2l-arrowkeys-focusable">link 2</a>
         ...
       </div>
-    `;
+    `);
   }
 }
 customElements.define('my-element', MyElement);
@@ -31,14 +31,14 @@ If the elements are not known up front, or the elements cannot be simply queried
 import { ArrowKeysMixin } from '@brightspace-ui/core/mixins/arrow-keys-mixin.js';
 class MyElement extends ArrowKeysMixin(LitElement) {
   render() {
-    return html`
+    return this.arrowKeysContainer(html`
       <div>
         ...
         <a href="...">link 1</a>
         <a href="...">link 2</a>
         ...
       </div>
-    `;
+    `);
   }
   async arrowKeysFocusablesProvider() {
     return [ /* array containing focusable elements */]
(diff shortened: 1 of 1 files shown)


mixins/arrow-keys-mixin.md
```diff
@@ -12,14 +12,14 @@ If the elements are known up-front and are in the element's local DOM scope, sim
 import { ArrowKeysMixin } from '@brightspace-ui/core/mixins/arrow-keys-mixin.js';
 class MyElement extends ArrowKeysMixin(LitElement) {
   render() {
-    return html`
+    return this.arrowKeysContainer(html`
```
dbatiste, line 15: Yeah, I agree this needs to be added.

The alternative is that the consumer wires up the handler themselves, [which is done in at least one place](https://github.com/Brightspace/folio-app/blob/main/src/web-components/radio-button/radio-button-group.js#L72).  In this case, the `host` is the container, but generally, I'd rather see them use `arrowKeysContainer()` so we can better contain the implementation details.

mixins/arrow-keys-mixin.md
```diff
@@ -51,4 +51,8 @@ customElements.define('my-element', MyElement);
 
 - `arrowKeysDirection` (String): Indicates which arrow keys are allowed (default is leftright)
 - `arrowKeysNoWrap` (Boolean): Whether focus should wrap from end-to-start and start-to-end
-- `arrowKeysBeforeFocus` (Function): Async callback invoked before focus us applied
+
+**Functions:**
+- `arrowKeysContainer`: Hooks up the event listener, to be used in your `render` function to wrap the focusable elements
```
dbatiste, line 56: I think we usually include the argument when we list the methods. Ex. [lists](https://github.com/BrightspaceUI/core/tree/main/components/list#methods).  Not sure if we're entirely consistent though. So I think I would write this as "`arrowKeysContainer(TemplateResult)`: ...." or "`arrowKeysContainer(htmlTemplate)`: ....", or something along those lines.

mixins/arrow-keys-mixin.md
```diff
@@ -51,4 +51,8 @@ customElements.define('my-element', MyElement);
...
+- `arrowKeysFocusablesProvider`: Override to provide the focusable elements if necessary (see above)
+- `arrowKeysOnBeforeFocus`: Override to provide an async callback invoked before focus is applied
```
dbatiste, line 58: I might mention this is optional.  Background - I think tabs exclusively uses this to update some visuals (scroll position, expansion of collapsed tabs) before trying to apply focus, otherwise the browser will do it's own thing to scroll overflowing elements into view which messes up the designed behaviour.

dbatiste (approved):
Just a couple minor suggestions.

mixins/arrow-keys-mixin.md
svanherk: Took another stab at this section!  Yeah we don't seem to be super consistent, tried to copy the general style I saw on a few pages.  Feel free to suggest edits, not tied to any of them lol

dbatiste:
Looks good to me!

PR #2238, opened 2022-03-18
Use roving tabindex for focus management to ensure forceFocusVisible behaves as expected

grant-cleary:
When `forceFocusVisible` is applied to a `d2l-tabs` component (the prime example is when it's the first focusable child of a dialog), we end up applying focus-visible to the focus-start span instead of the actual tab and we don't get the focus styles we expect. This PR aims to fix that by switching to use a roving tabindex for focus management within the tabs container.

In general, the idea is as follows:
- The selected tab starts with `tabindex="0"` by default; all others start with `tabindex="-1"`
- Arrowing between tabs switches tabindex to `0` for the new focused tab and sets the previous tab back to `-1`
- Tabbing out of the container (including to the "see more" arrow buttons if there are enough tabs) resets tabindex back to its initial state to ensure focus always goes back to the selected tab
- Changing the selected tab also sets tabindex to `0` on the newly selected tab, and resets all others to tabindex `-1`


components/tabs/tab-internal.js
```diff
@@ -95,7 +95,6 @@ class Tab extends RtlMixin(LitElement) {
 		super();
 		this.ariaSelected = 'false';
 		this.role = 'tab';
-		this.tabIndex = -1;
```
grant-cleary, line 98: I opted to handle tabindex management in the `tabs` component itself, including tracking its status in the `tabInfo` objects. I'm not tied to that idea; it just seemed cleaner. I'm happy to manage the tabindex within `tab-internal` if that's more desirable.

components/tabs/tabs.js
```diff
@@ -552,22 +560,9 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
...
-
-	_handleFocusStart(e) {
-		if (e.relatedTarget && e.relatedTarget.role === 'tab') {
-			const previousFocusable = getPreviousFocusable(e.target, false);
-			if (previousFocusable) previousFocusable.focus();
-		} else {
-			this._focusSelected();
-		}
+	_handleFocusOut(e) {
+		if (!e.relatedTarget || e.relatedTarget.role === 'tab') return;
```
grant-cleary, line 564: If the `focusout` event is being generated because we're moving between tabs, don't reset focusables.

components/tabs/tabs.js
```diff
@@ -603,7 +598,10 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 				selected: panel.selected,
 				state: state
 			};
-			if (tabInfo.selected) selectedTabInfo = tabInfo;
+			if (tabInfo.selected) {
+				selectedTabInfo = tabInfo;
+				tabInfo.activeFocusable = true;
```
grant-cleary, line 603: I'm a bit less sure about this piece, as I'm not 100% sure how one adds a new selected panel. Do we need to clear `activeFocusable` from all of the other tabs when this happens?

grant-cleary:
Visual diff failures are as expected - the focus ring is now being applied appropriately.

components/tabs/tab-internal.js
dbatiste: It is possible that tabs can be added and removed afterwards and we should account for those scenarios.  For that reason, it still may be worthwhile initializing the `tabindex="-1"` here.  Also, if the selected tab is removed, a new tab must be selected... and we'd expect its `tabindex` to then get set to 0.

components/tabs/tabs.js
```diff
@@ -279,7 +278,14 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 		};
 
 		this.arrowKeysOnBeforeFocus = async(tab) => {
+			const prevFocusable = this._tabInfos.find(info => info.activeFocusable);
+			if (prevFocusable) prevFocusable.activeFocusable = false;
+
 			const tabInfo = this._getTabInfo(tab.controlsPanel);
+			tabInfo.activeFocusable = true;
+
+			this.requestUpdate();
```
dbatiste, line 287: Are you sure we don't need to `await this.updateComplete`?  Otherwise the active focusable tabindex values might not be updated by the time the arrow keys mixin calls `focus`...?  Or are we effectively relying on the code below to provide that delay?  If it doesn't need to be scrolled, and it doesn't need to be expanded, then there might not be any other delay.  This is something to check if you haven't already.

components/tabs/tabs.js
```diff
@@ -340,22 +348,22 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 							<d2l-icon icon="tier1:chevron-left"></d2l-icon>
 						</button>
 					</div>
-					<span class="d2l-tabs-focus-start" @focus="${this._handleFocusStart}" tabindex="0"></span>
```
dbatiste, line 343: I think there is some CSS for these focus-trap classes that can also be removed.

components/tabs/tabs.js
```diff
@@ -301,6 +307,8 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 			}
 		};
 
+		if (this._tabInfos.length > 0) this._tabInfos[0].activeFocusable = true;
```
dbatiste, line 310: I think this is already being handled in the slot change handler below, ya?

components/tabs/tabs.js
grant-cleary: That said, I think this check is wrong. I believe it should be:
```e.relatedTarget && e.relatedTarget.role === 'tab'```

If the relatedTarget is `null`, that means we're potentially removing focus from the document entirely, but we still want to reset focusables, I imagine.

File: components/tabs/tabs.js (hunk 3 above)
Comment (dbatiste) [reply components/tabs/tabs.js:603]: If they try to add a second selected tab then I think it's an invalid state - what happens in that case now... do we do anything at all?

File: components/tabs/tabs.js (hunk 6 above)
Comment (grant-cleary) [reply components/tabs/tabs.js:310]: Yep, you're right. I forgot that we'll always get a `slotchange` event firing on first update.

File: components/tabs/tabs.js (hunk 4 above)
Comment (grant-cleary) [reply components/tabs/tabs.js:287]: I think you're probably right. In practice I never ran into this scenario and I sort of just mimicked what the code does when a new tab is selected, but I imagine there's some scenario where the timing wouldn't work as expected.

File: components/tabs/tabs.js (hunk 3 above)
Comment (grant-cleary) [reply components/tabs/tabs.js:603]: It doesn't look like it, but I didn't come up with a scenario that tested this particular scenario. I can contrive something on main to see what happens here and try to reproduce that in the new logic.

File: components/tabs/tabs.js [hunk 7]
```diff
@@ -552,22 +560,9 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
...
-			if (previousFocusable) previousFocusable.focus();
-		} else {
-			this._focusSelected();
-		}
+	_handleFocusOut(e) {
+		if (!e.relatedTarget || e.relatedTarget.role === 'tab') return;
+		this._resetFocusables();
 	}
 
 	_handlePanelSelected(e) {
```
grant-cleary, line 568: It occurs to me that I missed this case. If the panel is selected, we also set its associated tab as selected. If that's the case, we also want to change which tab is focusable as well.

components/tabs/demo/tabs.html
```diff
@@ -73,30 +79,32 @@ <h3>Tabs (responsive)</h3>
...
-						<d2l-tab-panel text="F13">Tab content for F13</d2l-tab-panel>
-						<d2l-tab-panel text="S13">Tab content for S13</d2l-tab-panel>
-						<d2l-tab-panel text="W13">Tab content for W13</d2l-tab-panel>
-						<d2l-tab-panel text="F12">Tab content for F12</d2l-tab-panel>
-						<d2l-tab-panel text="S12">Tab content for S12</d2l-tab-panel>
-						<d2l-tab-panel text="W12">Tab content for W12</d2l-tab-panel>
-						<d2l-tab-panel text="F11">Tab content for F11</d2l-tab-panel>
-						<d2l-tab-panel text="S11">Tab content for S11</d2l-tab-panel>
-					</d2l-tabs>
+					<div style="max-width: 500px;">
```
grant-cleary, line 82: This demo was broken before, as it wasn't actually responsive. The tabs container just ended up overflowing the demo snippet container.

components/tabs/tabs.js
```diff
@@ -780,6 +777,19 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
...
+		if (selectedTab) this._setFocusable(selectedTab);
+		this.requestUpdate();
+	}
+
+	_setFocusable(tabInfo) {
+		const currentFocusable = this._tabInfos.find(ti => ti.activeFocusable);
+		if (currentFocusable) currentFocusable.activeFocusable = false;
+
+		tabInfo.activeFocusable = true;
+	}
```
grant-cleary, line 791: I broke this out into a helper function because it was being used all over the place.

grant-cleary:
Bah, I did a rebase and it removed some linting changes I made.

dbatiste (approved):

PR #2244, opened 2022-03-21
fix: handling text change in updated

dlockhart:
This was noticed [in my-courses](https://github.com/Brightspace/d2l-my-courses-ui/pull/922/). Tab panels were relying on the `attributeChangedCallback` to update the `aria-label` and fire the `d2l-tab-panel-text-changed` event -- which `<d2l-tabs>` uses to keep the actual tab label updated.

But `attributeChangedCallback` isn't called when properties are set directly. This switches to using `updated()`. I've tested this in my-courses and it addresses the issue.

### DIFF
--- a/components/tabs/tab-panel-mixin.js
+++ b/components/tabs/tab-panel-mixin.js
@@ -51,17 +51,6 @@ export const TabPanelMixin = superclass => class extends superclass {
 		this.selected = false;
 	}
 
-	async attributeChangedCallback(name, oldval, newval) {
-		super.attributeChangedCallback(name, oldval, newval);
-		if (name === 'text') {
-			this.setAttribute('aria-label', this.text);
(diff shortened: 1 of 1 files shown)


dbatiste (approved):

svanherk (approved):
This functionality was originally added for filter, if you still have a site up might be good to do a quick check in the My Courses filter and the Quick Eval filter (I think the filter tabs load even if there's no entries in the table)

svanherk:
Actually, I think it was to add the selected count the tab name.  So testing in my courses by applying some filters is probably enough (make sure you're not showing semester/department tabs, and you'll be able to see the semester and department filters)

dlockhart:
> So testing in my courses by applying some filters is probably enough 

Seems to work!

PR #2246, opened 2022-03-22
Remove SauceLabs

dlockhart:
We've been running SauceLabs and Playwright in parallel for quite some time now, and I think we can conclude that the Sauce tests are slower, flakier, more expensive ($$) and aren't finding any defects beyond what Playwright does.


svanherk (approved):
Do we want to update `create` to use Playwright instead of Sauce too?

PR #2250, opened 2022-03-23
fix: Daylight: tabs: re-add d2l-tab-panel-text-changed event description

margaree:


svanherk (approved):

svanherk:
I got the same failed visual diff test on my branch

PR #2254, opened 2022-03-23
fix: Add info about lit repeat to filter readme

svanherk:
Not attached to the title "Troubleshooting" lol

### DIFF
--- a/components/filter/README.md
+++ b/components/filter/README.md
@@ -135,7 +135,7 @@ Filter with a single dimension:
 
 Lit tries to reuse DOM nodes when it can to help with performance, but in this case we don't want unique dimensions and values to be reused - otherwise we can't detect additions/removals properly.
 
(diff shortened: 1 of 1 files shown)


components/filter/README.md
```diff
@@ -149,6 +149,24 @@ The filter will announce changes to filter selections, search results, and when
 
 <!-- docs: end hidden content -->
 
+### Troubleshooting
+
+If your use-case requires that dimensions or dimension values be added or removed after initial load, you may need to use the [Lit `repeat` directive with a `KeyFnc` set](https://lit.dev/docs/templates/directives/#repeat) to tell Lit not to reuse a DOM node if the `key` has changed:
+```html
```
margaree, line 155:
```suggestion
```js
```

margaree (approved):

components/filter/README.md
```diff
@@ -149,6 +149,24 @@ The filter will announce changes to filter selections, search results, and when
 
 <!-- docs: end hidden content -->
 
+### Troubleshooting
```
dlockhart, line 152: Should we straight-up recommend using `repeat` any time they're iterating over dimensions or values using a `forEach`? Then you could call this section "Iterating with forEach" or something like that.

svanherk:
Round two - this sound better?  Was still gonna add a small snippet to the Daylight site

dlockhart (approved):

PR #2258, opened 2022-03-25
feat: US136317- Add slim attribute to d2l-button-subtle

svanherk:
Add `slim` attribute to `d2l-button-subtle` and update demo pages and visual-diff tests.

Design details:
- Height should be `1.5rem` (`30px`)
- Side padding should reduce to `0.5rem` (`10px`)
- When there is a button, that side's padding should reduce to `0.4rem` (`8px`)

image

### DIFF
--- a/components/button/button-subtle.js
+++ b/components/button/button-subtle.js
@@ -41,6 +41,12 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 			iconRight: { type: Boolean, reflect: true, attribute: 'icon-right' },
 
 			/**
+			 * Whether to render the slimmer version of the button
+			 * @type {boolean}
+			 */
+			slim: { type: Boolean, reflect: true },
+
+			/**
 			 * REQUIRED: Text for the button
 			 * @type {string}
 			 */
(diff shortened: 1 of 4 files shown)


components/button/button-subtle.js
```diff
@@ -59,19 +65,41 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 				}
 
 				button {
+					--d2l-button-subtle-padding-left: 0.6rem;
+					--d2l-button-subtle-padding-right: 0.6rem;
 					background-color: transparent;
 					border-color: transparent;
 					font-family: inherit;
-					padding: 0.55rem 0.6rem;
+					padding: 0 var(--d2l-button-subtle-padding-right) 0 var(--d2l-button-subtle-padding-left);
```
svanherk, line 73: Removing the `0.55rem` top and bottom padding is what fixes the problem with the regular subtle buttons being slightly larger than `42px`.  The `min-width` of `calc(2rem + 2px)` on all buttons is enough to determine the size, and the padding here made it a tad too big.  I can't find a case where the `padding` was needed... maybe old browsers?  Running visual-diff to see what it finds, also haven't tested Safari yet

components/button/button-subtle.js
```diff
@@ -59,19 +65,41 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 				}
 
 				button {
+					--d2l-button-subtle-padding-left: 0.6rem;
+					--d2l-button-subtle-padding-right: 0.6rem;
```
dlockhart, line 69: Suggestion: call these `--d2l-button-subtle-padding-inline-start` and `-inline-end`. This will align the terminology with the upcoming [CSS Logical Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) that we'll be switching to as soon as the browser support is there. It will let us remove the RTL-specific CSS.

components/button/demo/button-subtle.html
```diff
@@ -18,6 +18,7 @@ <h2>Subtle Button with Text Only</h2>
 			<d2l-demo-snippet>
 				<template>
 					<d2l-button-subtle id="normal" text="Subtle Button"></d2l-button-subtle>
+					<d2l-button-subtle slim id="normal" text="Slim Subtle Button"></d2l-button-subtle>
```
dlockhart, line 21: Not sure why these had `id`s in the first place, but we don't want elements with the same `id` in the same shadow root. I think they can just be removed everywhere in this file.

components/button/test/button-subtle.visual-diff.html
```diff
@@ -14,22 +14,53 @@
 </head>
 <body class="d2l-typography">
 	<div class="visual-diff">
-		<d2l-button-subtle id="normal" text="Subtle Button"></d2l-button-subtle>
```
dlockhart, line 17: I'd suggest renaming the golden filenames to match what they'll be after this change. That way the visual diff changes will be more realistic.

components/button/demo/button-subtle.html
svanherk: Good catch, I'm guessing they were copied from the visual-diff file originally.  Agreed that they aren't needed

dlockhart (approved):

PR #2266, opened 2022-03-29
Layout updates for list-items.

dbatiste:
This PR makes some layout changes to list-items.  Hover, focus, and selected states now render a border around the outside of the item.  Separators no longer necessarily span the entire with of the item, but rather exclude the drag-handle.  Nested item states will no longer be contained by parent item states. There are many permutations of options, but these screenshots highlight some key differences.

Selected state
image

Focus/Hover state (only diff is only the hover state has box-shadow)
image

Nested selected state (nested items are not contained within parent border even though they are children)
image

Draggable items (separator is not rendered underneath the drag handle)
image


components/list/list-item-generic-layout.js
```diff
@@ -37,6 +37,11 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 
 	static get properties() {
 		return {
+			/**
+			 * How to align content in the nested slot
+			 * @type {'content'|'control'}
+			 */
+			alignNested: { type: String, reflect: true, attribute: 'align-nested' },
```
dlockhart, line 44: Needs a default set in the constructor.

dlockhart (approved):

PR #2268, opened 2022-03-29
US136319 - Add ability for tooltip to only show when an element is truncating

svanherk:
Just putting up what I have for early feedback and handover, but I don't love the way the attribute works.  It's basically a combination of a boolean and a string.  If the attribute is not set at all, it'll be `undefined` and none of the new code runs.  If it's set _like_ a boolean (meaning the value is an empty string), we use the tooltip's target element.  This is how many use cases will use it, and how I originally had it always working.  But through testing I found more complicated use cases where the tooltip is not pointing to the element that's truncating (like in the `list` visual-diff example).

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -149,6 +149,12 @@ class Tooltip extends RtlMixin(LitElement) {
 			 */
 			offset: { type: Number }, /* tooltipOffset */
 			/**
+			 * ADVANCED: Only show the tooltip if we detect the element of the id passed is truncating
+			 * If no id is set, the tooltip target element is used
+			 * @type {string}
+			 */
+			onlyShowIfTruncating: { type: String, attribute: 'only-show-if-truncating' },
+			/**
 			 * ADVANCED: Force the tooltip to open in a certain direction. If no position is provided, the tooltip will open in the first position that has enough space for it in the order: bottom, top, right, left.
 			 * @type {'top'|'bottom'|'left'|'right'}
 			 */
@@ -650,6 +656,28 @@ class Tooltip extends RtlMixin(LitElement) {
 		return spaceAround;
 	}
 
+	async _determineIfTruncating() {
+		let target = this._target;
+		if (this.onlyShowIfTruncating) {
+			target = target.querySelector(`#${cssEscape(this.onlyShowIfTruncating)}`);
+		}
+		const clone = target.cloneNode(true);
+		clone.removeAttribute('id');
+		clone.style.position = 'absolute';
+		if (this.getAttribute('dir') === 'rtl') {
+			clone.style.right = '-10000px';
+		} else {
+			clone.style.left = '-10000px';
+		}
+
+		target.appendChild(clone);
+		await this.updateComplete;
+		const truncating = clone.offsetWidth > target.offsetWidth;
+		target.removeChild(clone);
+
+		return truncating;
+	}
+
 	_findTarget() {
 		const ownerRoot = this.getRootNode();
 
@@ -749,7 +777,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		}
 	}
 
-	_onTargetFocus() {
+	async _onTargetFocus() {
+		if (this.onlyShowIfTruncating !== undefined) {
+			const truncating = await this._determineIfTruncating();
+			if (!truncating) return;
+		}
 		if (this.disableFocusLock) {
 			this.showing = true;
 		} else {
@@ -759,7 +791,11 @@ class Tooltip extends RtlMixin(LitElement) {
 	}
 
 	_onTargetMouseEnter() {
-		this._hoverTimeout = setTimeout(() => {
+		this._hoverTimeout = setTimeout(async() => {
+			if (this.onlyShowIfTruncating !== undefined) {
+				const truncating = await this._determineIfTruncating();
+				if (!truncating) return;
+			}
 			resetDelayTimeout();
 			this._isHovering = true;
 			this._updateShowing();
--- a/components/tooltip/demo/tooltip.html
+++ b/components/tooltip/demo/tooltip.html
@@ -32,6 +32,26 @@
 			padding: 20px;
 			width: 200px;
 		}
+
+		.truncated {
+			border: 1px solid #cdd5dc;
+			border-radius: 6px;
+			height: 200px;
+			padding: 20px;
+			width: 200px;
+		}
+
+		.truncated d2l-button {
+			margin-bottom: 20px;
+			max-width: 100%;
+		}
+
+		.truncated d2l-button span {
+			display: block;
+			overflow: hidden;
+			text-overflow: ellipsis;
+			white-space: nowrap;
+		}
 	</style>
 </head>
 
@@ -109,6 +129,22 @@
 					<d2l-button id="tooltip-bounded">Bounded Tooltip</d2l-button>
 					<d2l-tooltip for="tooltip-bounded">
 						A bounded tooltip message that will wrap to more than one line.
+					</d2l-tooltip>
+				</div>
+			</template>
+		</d2l-demo-snippet>
+
+		<h2>Tooltip (only show if truncating)</h2>
+		<d2l-demo-snippet>
+			<template>
+				<div class="truncated">
+					<d2l-button id="tooltip-short"><span>Short Tooltip</span></d2l-button>
+					<d2l-tooltip for="tooltip-short" only-show-if-truncating>
+						This tooltip will not show.
+					</d2l-tooltip>
+					<d2l-button id="tooltip-long"><span>Very Very Very Very Long Tooltip</span></d2l-button>
+					<d2l-tooltip for="tooltip-long" only-show-if-truncating>
+						Very Very Very Very Long Tooltip - this tooltip will show because the text is truncating.
 					</d2l-tooltip>
 				</div>
 			</template>
--- a/components/tooltip/test/tooltip-helper.js
+++ b/components/tooltip/test/tooltip-helper.js
@@ -11,3 +11,30 @@ export async function show(page, selector) {
 	page.$eval(`${selector} d2l-tooltip`, tooltip => tooltip.show());
 	return openEvent;
 }
+
+export function getRect(page, selector) {
+	return page.$eval(selector, elem => {
+		let x, y, width, height;
+		const openerRect = elem.getBoundingClientRect();
+		const content = elem.querySelector('d2l-tooltip');
+		if (!content.showing) {
+			x = openerRect.x;
+			y = openerRect.y;
+			width = openerRect.right - x;
+			height = openerRect.bottom - y;
+		} else {
+			const contentWidth = content.shadowRoot.querySelector('.d2l-tooltip-content');
+			const contentRect = contentWidth.getBoundingClientRect();
+			x = Math.min(openerRect.x, contentRect.x);
+			y = Math.min(openerRect.y, contentRect.y);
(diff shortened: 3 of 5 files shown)


components/tooltip/tooltip.js
```diff
@@ -148,6 +148,12 @@ class Tooltip extends RtlMixin(LitElement) {
 			 * @type {number}
 			 */
 			offset: { type: Number }, /* tooltipOffset */
+			/**
+			 * ADVANCED: Only show the tooltip if we detect the element of the id passed is truncating
+			 * If no id is set, the tooltip target element is used
+			 * @type {string}
+			 */
+			onlyShowIfTruncating: { type: String, attribute: 'only-show-if-truncating' },
```
margaree, line 156: Are cases like the list item one that mean this can't just be a bool actually likely/planned to happen?

components/tooltip/tooltip.js
```diff
@@ -749,7 +777,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		}
 	}
 
-	_onTargetFocus() {
+	async _onTargetFocus() {
+		if (this.onlyShowIfTruncating !== undefined) {
```
margaree, line 781: I _think_ this could just be done in the resize observer rather than on each focus/hover, since that seems to run even if the tooltip isn't visible

svanherk: Hmm, I wonder if a combination of both would work well.  We didn't really want to do the measuring/calculating on resize for all these tooltips in case they never actually open the tooltips, but if the page hasn't changed size there's no reason to recalculate again.  I wonder if we measure on hover/focus, but then don't remeasure if the resizeObserver hasn't fired since?

components/tooltip/tooltip.js
svanherk: Maybe not, I had it as a `boolean` when working with tags, and didn't realize it wouldn't work for list items until I was writing the visual-diffs. The main use cases I can think of (tags, buttons, and just general text in the immersive nav or wherever) should work fine with the boolean.  So we could always defer that, and add an additional `truncating-target` attribute if needed.

components/tooltip/tooltip.js
margaree: ah that's a good point and good idea - I'll test that out

components/tooltip/tooltip.js
```diff
@@ -650,6 +657,27 @@ class Tooltip extends RtlMixin(LitElement) {
 		return spaceAround;
 	}
 
+	async _determineIfTruncating() {
+		// if no resize has happened since truncation was previously calculated the result will not have changed
+		if (!this._resizeRunSinceTruncationCheck || !this.onlyShowIfTruncating) return;
+
+		const target = this._target;
+		const clone = target.cloneNode(true);
```
dbatiste, line 665: I am curious what the behaviour would be for a HM component, or a component that does a lot of work when it is constructed or connected.  Presumably the clone automatically has all the internal DOM, but I am guessing it might do some HM request, but I am not sure.

components/tooltip/tooltip.js
```diff
@@ -650,6 +657,27 @@ class Tooltip extends RtlMixin(LitElement) {
 		return spaceAround;
 	}
 
+	async _determineIfTruncating() {
```
dbatiste, line 660: Nit: I might be inclined to name this method `_updateTruncating` to better reflect what it's doing.

components/tooltip/tooltip.js
```diff
@@ -650,6 +657,27 @@ class Tooltip extends RtlMixin(LitElement) {
...
+		// if no resize has happened since truncation was previously calculated the result will not have changed
+		if (!this._resizeRunSinceTruncationCheck || !this.onlyShowIfTruncating) return;
+
+		const target = this._target;
+		const clone = target.cloneNode(true);
+		clone.removeAttribute('id');
```
dbatiste, line 666: Curious if you ran into a problem with the id being present on the clone, or is this just a careful measure?

margaree: @svanherk not sure if you had thoughts on this but I believe it's just a careful measure - in practice the element wouldn't be around long enough to cause problems

components/tooltip/tooltip.js
margaree: That could potentially be complicated - is there a case you have in mind where HM component content would potentially be truncated and show up in an overflow tooltip?

dbatiste: `d2l-activity-name` and `d2l-hc-name` come to mind.  Maybe they would just want to bake the tooltip into them.  But then they would need to be focusable, ya?  Making things focusable that would not normally be focusable worries me though.

components/tooltip/tooltip.js
```diff
@@ -650,6 +657,27 @@ class Tooltip extends RtlMixin(LitElement) {
...
+		clone.style.position = 'absolute';
+		if (this.getAttribute('dir') === 'rtl') {
+			clone.style.right = '-10000px';
```
dbatiste, line 669: Although unlikely, content this size (at 19px font-size) would be possible to spill into the viewport.  I suspect this doesn't align with the driving use-case, but I have a sneaking feeling this is going to get used in all sorts of contexts, possibly even with user-authored text.
image

The safest way to address this is probably to place the element in a containing element offscreen that doesn't show overflow or allow wrapping (like with `.d2l-offscreen`), and use `scrollWidth` instead of `offsetWidth`.

Another way might be to apply `visibility: hidden;` but I think it could result in horizontal scrollbar if the invisible text stretch beyond the opposite side of the viewport.

components/tooltip/tooltip.js
margaree: If they did just baked it in, then the potentially truncated element would need to be focusable if they wanted the tooltip to show up on focus. If they just wanted to show tooltip on hover then I believe it should work out of the box.

components/tooltip/tooltip.js
```diff
@@ -863,5 +881,31 @@ class Tooltip extends RtlMixin(LitElement) {
...
+
+		if (this.getAttribute('dir') === 'rtl') {
+			clone.style.right = '-10000px';
+		} else {
+			clone.style.left = '-10000px';
+		}
+
+		document.body.appendChild(clone);
+		await this.updateComplete;
+		this._truncating = clone.scrollWidth > target.offsetWidth;
```
dbatiste, line 906: This calc assumes that the typography of the clone and the target will be the same.  However, consider...  

```html
<div class="d2l-body-small">
   <span id="the-id">...</span>
   <d2l-tooltip for="the-id" only-show-if-truncating>...</d2l-tooltip>
</div>
```
By appending the clone to the body, the clone measurement would be larger than if measured in the context of the target.  `font-weight`, `font-size`, and `letter-spacing` are the cascading properties I mostly worry about.  The heading classes would be cases where the target measurement would be bigger than the clone.

Previously the clone was being appended to the target, which I _think_ would have avoided this potential issue.  Did you run into troubles with it?  It's certainly possible to use `getComputedStyle` to mitigate this scenario, although the nice thing about appending to the target is that it factors in all the possible properties that might affect the width.  If the target happens to be a web component that only has named slots, or has some special behaviour associated slot changes, appending the the target may indeed be undesirable.  

I'm not sure if it will come up... it really depends on what the consumers do.

svanherk: > Previously the clone was being appended to the target, which I think would have avoided this potential issue. Did you run into troubles with it?

Yeah I ran into this issue early on with some examples I was using, which was why I had it appending to `target` instead - so that it would inherit the same font sizes and font styles.  I think we need to account for that issue, since I expect most of the time this will be used with small text sizes?

dbatiste: I think when I played around with it, I had two elements...
```html
<div style="position: absolute; left: -10000px; width: 1px; overflow: hidden; whitespace: nowrap;">
   <the-clone></the-clone>
</div>
```
... and I measured the `scrollWidth` of `the-clone`.  Not sure if that's what you tried.

margaree: I've now tried both ways and am getting the same `scrollWidth` of 60 still for both even though it is definitely not 60 - must be doing something different🤔

dbatiste: I think it is getting this CSS from the demo page styles, which is preventing the button inside from growing.  I'm not entirely sure how it arrives at `60px` though.  If I set the clone's `max-width` to `none`, it gets the proper size. 
```css
.truncated d2l-button {
	margin-bottom: 20px;
	max-width: 100%;
}
```
In case you are interested... this is the code I tried...
```javascript
const cloneContainer = document.createElement('div');
cloneContainer.style.left = '-10000px';
cloneContainer.style.overflow = 'hidden';
cloneContainer.style.position = 'absolute';
cloneContainer.style.width = '1px';
cloneContainer.style.whiteSpace = 'nowrap';

const clone = target.cloneNode(true);
clone.removeAttribute('id');
clone.style.maxWidth = 'none';
cloneContainer.appendChild(clone);

if (this.getAttribute('dir') === 'rtl') {
	cloneContainer.style.right = '-10000px';
} else {
	cloneContainer.style.left = '-10000px';
}

target.appendChild(cloneContainer);
await this.updateComplete;
this._truncating = clone.scrollWidth > target.offsetWidth;
this._resizeRunSinceTruncationCheck = false;
target.removeChild(cloneContainer);
```
There's still the potential issue of dropping an element into a component's slot that it doesn't expect though.

margaree: Ah that works for me as well! Thanks for digging into that.

That is a good point about the potential issue with putting an element into a component's slot - it could in theory result in some weird behaviour. In that case is do you think the safer option to put the clone on document and hope that `getComputedStyle` is enough to get an accurate measurement on the clone width, or hope that consumers have filters on any `slotchange` events/that this doesn't trigger any strange behaviours with slots? I tested with the subtle button slot and that seemed fine but I realize that's not a complex case with any `slotchange` listeners doing calculations.

dbatiste: That's a good question - trade-offs either way.  

If the target is a web component that doesn't have a default (unnamed) slot, then the clone would not get rendered at all, but that scenario probably doesn't exist in our components today.  

What do we envision this being used for - buttons, dropdown openers maybe? bits of text, including headings? perhaps variants of tabs? Input elements cannot have children, but that case probably doesn't make sense anyway.  

I think I'd be inclined to append to the target, and keep the other sol'n in our back pocket - we can always pivot if we need to, and regressions are not a current risk since this is a new feature.  I'd be ok with either approach though.

margaree: Yup I think those are along the lines for the potential use cases. Ok yeah I agree about appending to the target for now. I'll update that and add a comment in the code about the document alternative in case there end up being issues.

components/tooltip/tooltip.js
```diff
@@ -148,6 +148,11 @@ class Tooltip extends RtlMixin(LitElement) {
 			 * @type {number}
 			 */
 			offset: { type: Number }, /* tooltipOffset */
+			/**
+			 * ADVANCED: Only show the tooltip if we detect the target element is truncated
+			 * @type {boolean}
+			 */
+			onlyShowIfTruncating: { type: Boolean, attribute: 'only-show-if-truncating' },
```
dbatiste, line 155: Nit: I might consider renaming this `showTruncatedOnly`.  I'd slightly less wordy, and there might be a tiny bit of precedent for it.  None of the names I can think of are amazing tbh.  🤷‍♂️

dbatiste (approved):
Just the property name to consider.  I'm happy with whatever you decide though.

PR #2274, opened 2022-03-30
Tag List: Line clamping

margaree:
This adds line clamping such that:
- Large widths: max 1 line of tags
- Medium widths: max 2 lines of tags
- Small widths: max 3 lines of tags

This follows the overflow-group resizeObserver method with modifications to have multiple lines. I could also try the `requestIdleCallback` idea, this just seemed like a straightforward place to start.

To dos:
- [x] handle focus after selected the "+ X more" with enter
- [x] add lang term for the button text
- [x] if the page is resized super quickly, the max-height calculation doesn't happen - might be limitation/unlikely, look into a bit more (update: can no longer reproduce this)


components/tag-list/tag-list.js
```diff
@@ -39,41 +63,157 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+				<d2l-button-subtle slim text="Show Less" @click="${this._toggleHiddenTagVisibility}" class="d2l-tag-list-button"></d2l-button-subtle>
+			` : html`
+				<d2l-button-subtle slim text="+ ${hiddenCount} more" @click="${this._toggleHiddenTagVisibility}" class="d2l-tag-list-button"></d2l-button-subtle>
+			`;
+		}
+
 		const list = html`
 			<div role="list" class="tag-list-container" aria-describedby="d2l-tag-list-description">
-				<slot></slot>
+				<slot @slotchange="${this._handleSlotChange}"></slot>
```
margaree, line 107: I added a `slotchange` listener since it sounded like based on today's meeting that would be needed going forwards anyways, and other functionality was being added to that initial piece anyways that it made a bit of sense to put it in its own function rather than crowd `firstUpdated`

components/tag-list/tag-list.js
```diff
@@ -39,41 +63,157 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+				${button}
 			</div>
 		`;
+
+		const outerContainerStyles = {
+			maxHeight: this._showHiddenTags ? 'unset' : `${(this._itemHeight + MARGIN_TOP_HEIGHT) * this._lines}px`
```
margaree, line 113: When we click the button to show the hidden tags, we want it to take up all the room, else we want it limited based on number of lines and line height

components/tag-list/tag-list.js
```diff
@@ -39,41 +63,157 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+		};
+
 		return html`
-			<div role="application">
+			<div role="application" class="tag-list-outer-container" style="${styleMap(outerContainerStyles)}">
+				<d2l-button-subtle aria-hidden="true" slim text="+ ## more" class="d2l-tag-list-hidden-button"></d2l-button-subtle>
```
margaree, line 118: Used this in order to have a button width to use in the calculations in `_chomp`.

Will update to use lang term (as well as in the couple other places that need lang terms)

components/tag-list/tag-list.js
```diff
@@ -39,41 +63,157 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+			showing.count -= 1;
+		}
+		this._chompIndex = showing.count;
+	}
+
+	_getItemLayouts(filteredNodes) {
+		const items = filteredNodes.map((node) => {
+			const computedStyles = window.getComputedStyle(node);
+
+			return {
```
margaree, line 198: This can probably get simplified a bit based on the pieces we'd actually need

components/tag-list/tag-list.js
```diff
@@ -84,6 +224,30 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+		this._availableWidth = Math.ceil(entries[0].contentRect.width);
+		if (this._availableWidth >= PAGE_SIZE.large) this._lines = PAGE_SIZE_LINES.large;
+		else if (this._availableWidth < PAGE_SIZE.large && this._availableWidth >= PAGE_SIZE.medium) this._lines = PAGE_SIZE_LINES.medium;
+		else this._lines = PAGE_SIZE_LINES.small;
+		this._chomp();
+	}
+
+	_handleSlotChange() {
+		requestAnimationFrame(() => {
+			this._getItems();
```
margaree, line 237: Probably don't need the separate function and can move the 2 lines into here

components/tag-list/tag-list.js
```diff
@@ -24,12 +40,20 @@ class TagList extends ArrowKeysMixin(LitElement) {
 			.tag-list-container {
 				display: flex;
 				flex-wrap: wrap;
-				margin: -6px -6px 0 0;
+				margin: -${MARGIN_TOP_HEIGHT}px -6px 0 0;
```
svanherk, line 43: Haha I had this before too!  See thread here: https://github.com/BrightspaceUI/core/pull/2236#discussion_r830198507
It seems to work because it's an integer, but maybe we're good with that and if it ever breaks visual-diff _should_ catch it

margaree: Haha interesting! I can change back to hard coded then or we can keep an eye on it

components/tag-list/tag-list.js
```diff
@@ -39,41 +64,159 @@ class TagList extends ArrowKeysMixin(LitElement) {
 		/** @ignore */
 		this.arrowKeysDirection = 'leftrightupdown';
 		this._items = [];
+		this._resizeObserver = null;
+		this._showHiddenTags = false;
```
dlockhart, line 68: May want to default `_chompIndex` to something large?

components/tag-list/tag-list.js
```diff
@@ -39,41 +64,159 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+			}
+		});
+
+		let button = null;
+		if (hasHiddenTags) {
+			button = this._showHiddenTags ? html`
+				<d2l-button-subtle
+					class="d2l-tag-list-button"
+					@click="${this._toggleHiddenTagVisibility}"
+					@keydown="${this._handleButtonKeydown}"
```
dlockhart, line 103: I think `@click` should be handling the ENTER case already just as part of a normal `<button>`'s API? Ideally folks (ourselves included) don't need special handlers for keyboard events on buttons.

I see we're also moving focus as part of the keyboard handling... hopefully we can either also do that as part of the `click` handling?

components/tag-list/tag-list.js
```diff
@@ -39,41 +64,159 @@ class TagList extends ArrowKeysMixin(LitElement) {
...
+		};
+
+		/**
+		 * _lines is determined by page width in _handleResize function
+		 * For each line we calculate the max items that can fit in that width, then go to the next line
+		 * If on the last line there is/are item(s) that won't fit in the width, we mark them as soft-hide and set isOverflowing
+		 */
+		let isOverflowing = false;
+		let overflowingIndex = 0;
+		for (let k = 1; k <= this._lines; k++) {
```
dlockhart, line 164: Could `this._lines` not be initialized at this point?

components/tag-list/test/tag-list.test.js
```diff
@@ -47,7 +47,7 @@ describe('d2l-tag-list', () => {
 		].forEach(testcase => {
 			it(testcase.name, async() => {
 				const list = await fixture(basicFixture);
-				await new Promise(resolve => requestAnimationFrame(resolve));
+				await aTimeout(100);
```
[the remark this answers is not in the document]
dlockhart: In similar scenarios I've had to await `updateComplete` on the individual items -- the fixture will handle the root node but there's no guarantee child elements have rendered unfortunately.

components/tag-list/tag-list.js
margaree: Yup `@click` does handle enter. I'd thought there wasn't a way to check if it was a keydown vs click event from the event so I separated them but I was definitely wrong🤦‍♀️- will change.

We wouldn't want to move focus on click, correct? It seemed to make sense to me to do so on enter since otherwise the keyboard focus goes into a bit of a no-mans-land and this way it becomes easy to toggle.

dlockhart: I think yeah ideally there isn't a visible focus change, but if we can move focus and the browser interprets it as part of the click and therefore triggers the invisible `:focus-visible` stuff, then that would be sweet.

dlockhart: But also I think we do visibly move focus for similar scenarios when the user clicks -- e.g. closing a dialog and moving focus back to the opener. It's not quite the same but if it involved specialized code and wasn't too ugly it might be fine.

dlockhart (approved):

dbatiste:
I am sort of late coming back to this PR.  What are people's thoughts on shorter naming... `d2l-tags` and `d2l-tag` as opposed to baking the list semantics into the name?

svanherk:
> I am sort of late coming back to this PR. What are people's thoughts on shorter naming... d2l-tags and d2l-tag as opposed to baking the list semantics into the name? (not against what it currently is... but thought I would toss that out in case it wasn't previously considered)

Yeah that was discussed, ultimately (if I'm remembering correctly) we went with `list` and `list-item` not to expose the semantics, but to make it more obvious that the individual tags _need_ to be in the tag list component and shouldn't be used on their own.  We felt people might expect to be able to use just a `d2l-tag`, but not ` d2l-tag-list-item`.

dlockhart:
> We felt people might expect to be able to use just a d2l-tag, but not d2l-tag-list-item.

That was the main motivation I think. And we chose list not so much because of the semantics (it could have been `<d2l-tag-collection>` and `<d2l-tag-collection-item>`) but more because it was short. :)

PR #2277, opened 2022-03-31
Fix input-text to include the unit in the aria-label.

dbatiste:
This PR updates the `aria-label` to include the `unit`.  Previously, when the label was hidden, an accessible `unit` would not be rendered or announced because it is rendered as part of the `label`.  I also played around with rendering the unit as part of the aria description, but appending to the label provided the best announcement, and also keeps it consistent with how it's announced when the label is visible.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -469,11 +469,12 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 	}
 
 	_getAriaLabel() {
-		if (this.label && (this.labelHidden || this.labelledBy)) {
-			return this.label;
-		}
-		if (this.hasAttribute('aria-label')) {
-			return this.getAttribute('aria-label');
+		let label;
+		if (this.label && (this.labelHidden || this.labelledBy)) label = this.label;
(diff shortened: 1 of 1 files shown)


components/inputs/input-text.js
```diff
@@ -469,11 +469,12 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
...
 
 	_getAriaLabel() {
-		if (this.label && (this.labelHidden || this.labelledBy)) {
-			return this.label;
-		}
-		if (this.hasAttribute('aria-label')) {
-			return this.getAttribute('aria-label');
+		let label;
+		if (this.label && (this.labelHidden || this.labelledBy)) label = this.label;
+		if (this.hasAttribute('aria-label')) label = this.getAttribute('aria-label');
```
dlockhart, line 474: Previously, if `aria-label` AND `label` were set, `label` would take precedence. Now `aria-label` will overwrite it.

```suggestion
		let label;
		if (this.label && (this.labelHidden || this.labelledBy)) {
			label = this.label;
		} else if (this.hasAttribute('aria-label')) {
			label = this.getAttribute('aria-label');
		}
```

PR #2279, opened 2022-03-31
Add support for dropdowns to have opener in a slot

annabelsegalld2l:
Currently, we only support direct children of `d2l-dropdown` to act as openers. Adding a `slotted-opener` attribute allows us to optionally check a `name="opener"` slot to see if any openers exist. This will catch both the default slot AND any slot overrides. 

This will be used by the user profile card. I would like to add tests, but since this is a no-op for all other dropdowns, I'll add more tests in a separate PR.

### DIFF
--- a/components/dropdown/dropdown.js
+++ b/components/dropdown/dropdown.js
@@ -22,8 +22,20 @@ class Dropdown extends DropdownOpenerMixin(LitElement) {
 	 */
 	getOpenerElement() {
 		if (!this.shadowRoot) return undefined;
+		if (this.slottedOpener) return this.getSlottedOpener();
 		return this.shadowRoot.querySelector('slot')
 			.assignedNodes()
+			.filter(node => node.classList && node.classList.contains('d2l-dropdown-opener'))[0];
+	}
+
+	/**
+	 * Gets the slotted opener element with class "d2l-dropdown-opener" (required by dropdown-opener-mixin).
+	 * @return {HTMLElement}
+	 */
+	getSlottedOpener() {
+		if (!this.shadowRoot) return undefined;
+		return this.querySelector('slot[name="opener"]')
+			.assignedNodes({ flatten: true })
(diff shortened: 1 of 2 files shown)


components/dropdown/dropdown.js
```diff
@@ -22,6 +22,11 @@ class Dropdown extends DropdownOpenerMixin(LitElement) {
 	 */
 	getOpenerElement() {
 		if (!this.shadowRoot) return undefined;
+		if (this.slottedOpener)  {
```
dlockhart, line 25: Instead of introducing an attribute, we could also first check the default slot and if it returns nothing, look in the `opener` slot? Shouldn't affect performance too much. 🤷

components/dropdown/dropdown.js
```diff
@@ -22,6 +22,11 @@ class Dropdown extends DropdownOpenerMixin(LitElement) {
...
+			return this.querySelector('slot[name="opener"]')
```
dbatiste, line 26: Is this the type of scenario you are wanting to support?
```html
my-component
   --shadowDOM
      <d2l-dropdown>
         <slot></slot>
         <d2l-dropdown-content max-width="400">
            ...
         </d2l-dropdown-content>
      </d2l-dropdown>
```

```html
<my-component>
   <d2l-button class="d2l-dropdown-opener">Open it!</d2l-button>
</my-component>
```

If so, can we just add the `{ flatten: true }` to the original code?  The `flatten` option will resolve the slots.

annabelsegalld2l: Hmm.. that doesn't seem to be working for me. It can't find the default slot in that case.

components/dropdown/dropdown.js
annabelsegalld2l: The default slot will always have content (ie. the dropdown-content). Did you mean to check the default slot and do the filtering for the opener as well?

PR #2285, opened 2022-04-01
DE47878 - Switch list to acting on keydown events instead of keyup

svanherk:
`DE47878` refers to the the first list option being selected when the filter is opened using the enter key.  This is an issue with `d2l-list` listening to `keyup` events instead of `keydown`, mixed with moving `focus` into the dropdown on open.

This PR switches everything that happens on `keyup` to instead happen on `keydown`, and removes the code that was previously suppressing all the corresponding `keydown` events.

### DIFF
--- a/components/list/demo/list-drag-and-drop.js
+++ b/components/list/demo/list-drag-and-drop.js
@@ -167,7 +167,7 @@ class ListDemoDragAndDrop extends LitElement {
 		await this.updateComplete;
 
 		if (e.detail.keyboardActive) {
-			requestAnimationFrame(() => {
+			setTimeout(() => {
 				if (!this.shadowRoot) return;
 				const newItem = this.shadowRoot.querySelector('d2l-list').getListItemByKey(sourceListItems[0].key);
 				newItem.activateDragHandle();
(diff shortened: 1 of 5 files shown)


components/list/demo/list-drag-and-drop.js
```diff
@@ -167,7 +167,7 @@ class ListDemoDragAndDrop extends LitElement {
 		await this.updateComplete;
 
 		if (e.detail.keyboardActive) {
-			requestAnimationFrame(() => {
+			setTimeout(() => {
```
dlockhart, line 170: Animation frame wasn't consistently waiting long enough?

dlockhart (approved):
Seems relatively straightforward!

components/list/demo/list-drag-and-drop.js
svanherk: Right, meant to put a comment on this - this is unrelated to the change I made and this demo is broken currently in `main`.  Was talking to Dave B about it and this was the fix he had locally and just hadn't committed it yet, so I offered to do it as part of my PR

dbatiste (approved):
Looks good!

PR #2288, opened 2022-04-04
Add mathjax test context

devpow112:
We are currently using this is rubrics vdiff, tests and demo pages to auto setup the mathjax context. I'd like to include it with the core (similar to https://github.com/BrightspaceUI/core/blob/main/tools/resize-observer-test-error-handler.js) so that we can use it from this component across multiple repos in testing scenarios (similar to https://github.com/BrightspaceUI/core/blob/main/web-test-runner.config.js#L31-L37).

### DIFF
--- a/tools/mathjax-test-context.js
+++ b/tools/mathjax-test-context.js
@@ -1,0 +1,4 @@
+document.documentElement.dataset.mathjaxContext = JSON.stringify({
+	outputScale: 1.1,
+	renderLatex: true
+});


devpow112:
Looks like references to `us125413-mathjax-render-latex` in the documentation can be removed as well (just confirming)
https://github.com/BrightspaceUI/core/tree/main/components/html-block (last section)

dbatiste (approved):

PR #2289, opened 2022-04-04
Use mathjax context in demos

devpow112:
Also update documentation. Should have done this with my other PR but didn't. This expands on #2288 and switches `core` demos to use the new file. Required a bit of re-work since didn't realize we had a demo page that disabled it. This should work, you need to explicitly set `latex=false` on a URL for a demo page or test to disable the test context render latex property, otherwise it's enabled.

### DIFF
--- a/components/html-block/README.md
+++ b/components/html-block/README.md
@@ -100,7 +100,7 @@ Examples are provided to display how user-authored math can be embedded within y
 </d2l-html-block>
 ```
 
-**LaTeX:** Rendering LaTeX requires the `mathjax` context to be set correctly. For testing and or demo pages you can do the following.
+**LaTeX:** Rendering LaTeX requires the `mathjax` context to be set correctly. For testing and/or demo pages you can do the following.
 
 <!-- docs: demo code -->
 ```html
@@ -117,7 +117,7 @@ Examples are provided to display how user-authored math can be embedded within y
 
 ### Add Context Automatically to Demos and Tests
 
-You can automatically set-up the `mathjax` context for demo pages and unit tests when using `@web/dev-server` and `@web/test-runner` by adding the following plugin to your configuration
+You can automatically set-up the `mathjax` context for demo pages and unit tests when using `@web/dev-server` and `@web/test-runner` by adding the following plugin to your configuration.
 
 ```javascript
 export default {
--- a/components/html-block/README.md
+++ b/components/html-block/README.md
@@ -67,6 +67,7 @@ To use `d2l-html-block` within another Lit component, use the [unsafeHTML](https
 ```
 
 ### Rendering MathML and LaTeX
+
 Examples are provided to display how user-authored math can be embedded within your webpage.
 
 **MathML:**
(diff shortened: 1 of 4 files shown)


tools/mathjax-test-context.js
```diff
@@ -1,4 +1,6 @@
-document.documentElement.dataset.mathjaxContext = JSON.stringify({
+const disabled = window.location.search.indexOf('latex=false') !== -1;
+
+document.getElementsByTagName('html')[0].dataset.mathjaxContext = JSON.stringify({
 	outputScale: 1.1,
-	renderLatex: true
+	renderLatex: !disabled
```
devpow112, line 5: Will be `true` unless explicitly disabled. This seems like a better state for most tests and demo pages using HTML block.

components/html-block/README.md
```diff
@@ -99,29 +100,41 @@ Examples are provided to display how user-authored math can be embedded within y
...
 <script type="module">
   import '@brightspace-ui/core/components/html-block/html-block.js';
-
-  window.D2L = {};
-  D2L.LP = {};
-  D2L.LP.Web = {};
-  D2L.LP.Web.UI = {};
-  D2L.LP.Web.UI.Flags = {
-    Flag: (feature, defaultValue) => {
-      if (feature === 'us125413-mathjax-render-latex') return true;
```
devpow112, line 115: The `us125413-mathjax-render-latex` flag isn't used anywhere and isn't required for LaTeX to work from what I can tell.

devpow112: Was marked as removed as part of https://github.com/Brightspace/lms-launch-darkly-flags/pull/2925 and later deleted

devpow112:
> > This looks ok to me.
> > @grant-cleary : do you have any concerns with this?
> 
> No real concerns here. I'm hoping folks with components integrated into the LMS don't start overriding these context objects of their own accord, but... Well there's nothing we can really do to stop that, so 🤷.

Yeah this is definitely not intended for that, just for testing/demo help

grant-cleary:
> Yeah this is definitely not intended for that, just for testing/demo help

Oh yeah, sorry, not specifically your change I'm thinking about here. I'm just not sure I like how we decided to talk about the MathJax context in the readme. The wording almost makes it sound like this is a thing we expect consumers to do manually, but... Well we kind of do for off-stack services, but not for stuff integrated in the LMS.

... Anyways, just idle musings. I'm happy with this as is. :P

devpow112:
> > No real concerns here. I'm hoping folks with components integrated into the LMS don't start overriding these context objects of their own accord, but... Well there's nothing we can really do to stop that, so 🤷.
> 
> Maybe we should include a console.warn/error in the `mathjax-test-context.js` file?

I could add something like `console.warn('Using mathjax test context, this is intended for demo pages and tests only')` if that woks for you @dbatiste @grant-cleary.

grant-cleary:
> I could add something like `console.warn('Using mathjax test context, this is intended for demo pages and tests only')` if that woks for you @dbatiste @grant-cleary.

Seems good to me.

PR #2298, opened 2022-04-07
feat: tag-list: truncate tag-list-items at 320px and show tooltip for truncated items

margaree:


components/tag-list/tag-list-item-mixin.js
```diff
@@ -65,12 +63,27 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+		const tooltip = hasTruncationTooltip ? html`
+				<d2l-tooltip for="${this._id}" show-truncated-only>
+					${tagContent}
+				</d2l-tooltip>
+			` : null;
 		return html`
-			<div class="tag-list-item-container d2l-label-text">
-				<div class="tag-list-item-content">${tagContent}</div>
+			${tooltip}
+			<div class="tag-list-item-container d2l-label-text" id="${this._id}" tabindex="0">
```
annabelsegalld2l, line 85: Do we need the container itself to be tabbable? Previously we had the first tag have a tabindex of 0 and the rest with -1.

components/tag-list/test/tag-list.test.js
```diff
@@ -47,17 +47,14 @@ describe('d2l-tag-list', () => {
...
 
 				const startItem = list._items[testcase.start];
-				await startItem.updateComplete;
-
 				startItem.focus();
 				dispatchKeydownEvent(startItem, testcase.key);
 
-				await new Promise(resolve => requestAnimationFrame(resolve));
-				expect(getComposedActiveElement()).to.equal(list._items[testcase.result]);
+				await nextFrame();
```
margaree, line 56: https://open-wc.org/docs/testing/helpers/#nextframe handles the requestAnimationFrame waiting for us

components/tag-list/test/tag-list.test.js
```diff
@@ -47,17 +47,14 @@ describe('d2l-tag-list', () => {
...
+				expect(getComposedActiveElement().getRootNode().host).to.equal(list._items[testcase.result]);
```
margaree, line 57: This is because of the focus going to the inner div rather than the outer tag-list-item component now

components/tag-list/tag-list-item-mixin.js
annabelsegalld2l: Could we use tabindex="-1" instead? That means it is focusable via script (so our arrowkeys mixin), but not via tab.

margaree: Yup that works!

annabelsegalld2l (approved):
This LGTM but I'd like to get another opinion from Stacey when she is back 😄

components/tag-list/tag-list-item-mixin.js
```diff
@@ -65,12 +63,27 @@ export const TagListItemMixin = superclass => class extends superclass {
...
 		/** @ignore */
 		this.role = 'listitem';
+		this._id = getUniqueId();
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		const container = this.shadowRoot.querySelector('.tag-list-item-container');
+		this.addEventListener('focus', () => container.focus());
```
dbatiste, line 73: By redirecting the focus from the host, the focus styles above (`:host(:focus) .tag-list-item-container`) doesn't appear to work as expected.  What's the reason for moving the focus inside the item from the host?

components/tag-list/tag-list-item-mixin.js
```diff
@@ -17,30 +19,26 @@ export const TagListItemMixin = superclass => class extends superclass {
...
 				min-width: 0;
+				outline: none;
+				overflow: hidden;
 				padding: 0.25rem 0.6rem;
+				text-overflow: ellipsis;
 				transition: background-color 0.2s ease-out, box-shadow 0.2s ease-out;
+				white-space: nowrap;
 			}
 			:host(:hover) .tag-list-item-container,
 			:host(:focus) .tag-list-item-container {
```
dbatiste, line 44: I'm not sure what the focus styles are supposed to look like... currently it looks like:
image
... which doesn't seem to match the styling we've used in several other places.
image
I imagine we want tags to look a little different from buttons though...?  Still seems a little off to me... not sure if Design is still iterating. 🤷‍♂️

components/tag-list/tag-list-item-mixin.js
margaree: The tooltip wouldn't show up on arrow key "focus" unless the focus was placed on the `.tag-list-item-container`. ([related](https://github.com/BrightspaceUI/core/pull/2298#discussion_r845422966))

The other option I can think of here that would not involve having to move the focus within the tag-list-item is the `tag-list` adds sibling tooltips to each list tag list item on slotupdate, but that doesn't seem like a great option to me vs. just having the tooltip live in the component (though the way this has to be modified for the tooltip to appear on focus is also not great).

components/tag-list/tag-list-item-mixin.js
margaree: Oh interesting - I actually am not seeing the extra outline when I run it so I'll dig into why that might be happening.

components/tag-list/tag-list-item-mixin.js
dbatiste: Cool yeah.  So looks like the CSS selector just needs to be updated.  😄

components/tag-list/tag-list-item-mixin.js
```diff
@@ -17,41 +19,38 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+				max-width: 320px;
 				min-width: 0;
+				outline: none;
+				overflow: hidden;
 				padding: 0.25rem 0.6rem;
+				text-overflow: ellipsis;
 				transition: background-color 0.2s ease-out, box-shadow 0.2s ease-out;
+				white-space: nowrap;
+			}
+			.tag-list-item-container:focus {
```
dbatiste, line 44: If a tag is focused... when hovered, the outline disappears.  Is that expected?

margaree: Nope! Will fix, thanks for catching

components/tag-list/tag-list.js
```diff
@@ -140,7 +142,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 	}
 
 	focus() {
-		if (this._items.length > 0) this._items[0].focus();
+		if (this._items && this._items.length > 0) this._items[0].focus();
 	}
 
 	_chomp() {
```
dbatiste, line 148: I wish GitHub would let me comment on lines outside your changes.  😦 

Is the purpose of this button simply for measurement?  If so, does the measurement ever change based on the number of items being chomped, or does it simply assume a width with the `##` value?  If it never changes, then perhaps we can avoid measuring every time we chomp.  It does seem like assuming a fixed with button is much easier to handle.

components/tag-list/tag-list.js
```diff
@@ -166,9 +168,10 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 
 			for (let i = overflowingIndex; i < this._itemLayouts.length; i++) {
 				const itemLayout = this._itemLayouts[i];
+				const itemWidth = Math.min(itemLayout.width, this._availableWidth);
 
-				if (!isOverflowing && showing.width + itemLayout.width < this._availableWidth) {
-					showing.width += itemLayout.width;
+				if (!isOverflowing && ((showing.width + itemWidth) <= (this._availableWidth + MARGIN_TOP_RIGHT))) {
```
dbatiste, line 173: What's the purpose of the `MARGIN_TOP_RIGHT`?

margaree: It corresponds to the 6px of top and right margin on tag items. I'd also had it used in the css but that was potentially an issue so I left it as a hard coded value there.

components/tag-list/tag-list.js
```diff
@@ -228,19 +230,34 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
 	_handleSlotChange() {
-		requestAnimationFrame(() => {
+		if (!this._hasResized) return;
+
+		requestAnimationFrame(async() => {
 			this._items = this._getTagListItems();
-			this._itemLayouts = this._getItemLayouts(this._items);
+			if (!this._items || this._items.length === 0) return;
+
+			const numItems = this._items.length;
```
dbatiste, line 251: Note: [forEach expects a synchronous function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).  It does not wait for promises.  It might seem like it's working, but I don't think it's doing what we think it is. Because it's not actually waiting, there is no guarantee that the previous items are actually done updating when the last item update completes.

To wait in sequence, I think it is possible to use `for...of` loop, or _maybe_ it's `for await...of`.  However a better approach is to wait in parallel by mapping to an array of promises.  This example uses `map`, but they could be manually collected in an array. 

```javascript
await Promise.all(this._items.map(item => item.updateComplete));
```

components/tag-list/tag-list.js
```diff
@@ -228,19 +230,34 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
+			const updateItems = new Promise((resolve) => {
+				this._items.forEach(async(item, index) => {
+					await item.updateComplete;
+					if (index === numItems - 1) resolve();
+				});
+			});
+			await updateItems;
 
-			if (this._items.length === 0) return;
+			this._itemLayouts = this._getItemLayouts(this._items);
```
dbatiste, line 260: I added an add/remove button to the demo page to test that behaviour.  The chomping seems to not work quite right when adding items ... it seems like once the items overflow, the `_getItemLayouts` doesn't always return updated results.  The remove doesn't appear to work as expected either.

margaree: Okay interesting - I'll dig more into this

margaree: Ah so it is returning updated results but because some of them are hidden after the third addition (with the screen size I'm testing) they're getting filtered out (https://github.com/BrightspaceUI/core/blob/2a071093a783cc93dc0f17f2ef48d4aff3f7ac35/components/tag-list/tag-list.js#L212). I can probably check for the `data-is-chomped` and not filter them in that case even if hidden

components/tag-list/tag-list.js
```diff
@@ -65,19 +64,26 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		if (this._resizeObserver) this._resizeObserver.disconnect();
+		if (this._subtleButtonResizeObserver) this._subtleButtonResizeObserver.disconnect();
 	}
 
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
+		const subtleButton  = this.shadowRoot.querySelector('.d2l-tag-list-hidden-button');
```
dbatiste, line 81: This seems to be working ok, measuring just the once when it's first rendered.  

If anyone in the future decides to actually update the hidden button with a count, they might get themselves into a 🐔 & 🥚  situation since the chomping depends on this, and this would depend on the chomping.

dbatiste (approved):
This seems to be working pretty well.

I find it a little strange resizing past 3 hidden tags.  At first it says "+3 more", but then I make it a tiny bit smaller and the second row is revealed, revealing 1 on the first row and 2 on the second row. This is just due to the "breakpoints".  I'm not sure the rationale for having the breakpoints, but they seem to be working as intended.

svanherk:
So I _think_ the reason you maybe ran into a bunch of focus issues with tooltip that I didn't originally is because I'd put the `tooltip` in the `d2l-tag-list-item` component, not the `TagListItemMixin`.  I did this because the other usage we know of (user  tags) wouldn't need a tooltip, since it will open the whole profile card (which will show the user's full name anyways).  Does having the tooltip built into the mixin clash with the profile card?

annabelsegalld2l:
> So I _think_ the reason you maybe ran into a bunch of focus issues with tooltip that I didn't originally is because I'd put the `tooltip` in the `d2l-tag-list-item` component, not the `TagListItemMixin`. I did this because the other usage we know of (user tags) wouldn't need a tooltip, since it will open the whole profile card (which will show the user's full name anyways). Does having the tooltip built into the mixin clash with the profile card?

Just pulled in Margaree's changes. The tooltip is optional, so having no tooltip with the card is OK and works fine. The opened card will render the entire name. I tried to add the tooltip in to see what happens. Looks like it doesn't close when the card opens, so the tooltip renders on top, and the profile image is also inside the tooltip which is wrong. For now we are safe to ignore since the tooltip isn't required for the card, but good info to know 👍

PR #2301, opened 2022-04-07
List-item layout fixes.

dbatiste:
This PR makes a few changes to update the layout of list-items.  Key things are:

* Update `d2l-list-item-content` margins so they are only rendered for secondary and supporting content when content is assigned to the slots.
* Update the size of the control containers (drag handle & selection checkbox) to be 44px wide for a11y.
* Update the separator width to be 100% - previously this was reduced to avoid funky rendering along with the border radius, which is no longer an issue because the outside container is being extended.
* Update to not show left/right borders when extending separators, since it is expected that the list is being placed in a container that provides a border (we don't want double borders at edges)
* Update margins so that checkbox is aligned with top of text.

### DIFF
--- a/components/list/list-item-content.js
+++ b/components/list/list-item-content.js
@@ -25,16 +25,20 @@ class ListItemContent extends LitElement {
 			.d2l-list-item-content-text-secondary {
 				color: var(--d2l-list-item-content-text-secondary-color, var(--d2l-color-tungsten));
 				margin: 0;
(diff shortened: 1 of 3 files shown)


annabelsegalld2l (approved):
These changes and the visual diffs look good to me. We will need to send an update in web-platform to let folks know to expect some list visual diffs.

PR #2315, opened 2022-04-11
Fix so that NVDA doesn't announce list items 3 times when moving them up with keyboard.

dbatiste:
This PR fixes an issue where NVDA announces the position and instruction for keyboard "move" button three times when moving an item up. (it would announce once as expected when moving down). Removing the `aria-live` fixes the issue, and the position & instruction are announced once as expected, moving up or down. Checked behaviour with both NVDA and VoiceOver.


dlockhart (approved):
I wish all fixes were this easy!

PR #2319, opened 2022-04-13
feat: tag-list and tag-list-item: clearable

margaree:
This follows similar logic to how multi select (https://github.com/BrightspaceUILabs/multi-select) is handling deleting items.

`autoremove` on `multi-select-list` was never used so that is not implemented here. This fires an event on a tag list item being deleted and then relies on the consumer to handle that.


components/tag-list/tag-list-item-mixin.js
```diff
@@ -71,24 +108,71 @@ export const TagListItemMixin = superclass => class extends superclass {
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		const container = this.shadowRoot.querySelector('.tag-list-item-container');
+		const container = this.shadowRoot.querySelector('.tag-list-item-content');
 		this.addEventListener('focus', (e) => {
 			// ignore focus events coming from inside the tag content
-			if (e.composedPath()[0] === this) container.focus();
+			if (e.composedPath()[0] !== this) return;
+			const event = new FocusEvent('focus', { bubbles: true, cancelable: true });
```
margaree, line 115: In order to hopefully improve the screenreader experience (seems to return it to how it was before the tooltip truncation changes + adds in the ability to use the button), this keeps the actual focus on the `d2l-tag-list-item` (or whatever is using the mixin) but then triggers the focus event on the div that's connected to the tooltip, showing the tooltip if truncated.

Let me know any concerns I might be missing around doing things this way.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -71,24 +108,71 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+		});
+		this.addEventListener('blur', () => {
+			const event = new FocusEvent('blur', { bubbles: true, cancelable: true });
+			container.dispatchEvent(event);
 		});
-		this.addEventListener('blur', () => container.blur());
+		this.addEventListener('keydown', this._handleKeydown);
+	}
+
+	deleteItem() {
```
margaree, line 125: This was in multi select and is useful for the demo. If this seems like something that should be kept, I'll document it in the README as something consumers can use.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -71,24 +108,71 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+		this.parentNode.removeChild(this);
+	}
+
+	handleClearItem(e) {
+		if (!this.clearable) return;
+
+		let handleFocus = false;
+		if (e) {
+			const listItemParent = findComposedAncestor(e.composedPath()[0], (node) => node.role === 'listitem');
```
margaree, line 134: Basically this is so that when "Clear All" is clicked, the focus handling isn't triggered. Something that could be an issue is that it sounds like we want to use tags in `d2l-list-item` so it might be best to do this a different way.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -71,24 +108,71 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+			handleFocus = listItemParent ? true : false;
+		}
+		/** Dispatched when a user selects to delete a tag list item. The consumer must handle the actual element deletion. */
+		this.dispatchEvent(new CustomEvent(
+			'd2l-tag-list-item-cleared',
+			{ bubbles: true, composed: true, detail: { value: this.text, handleFocus } }
```
margaree, line 140: Note that value will be undefined for any sort of custom list items.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -71,24 +108,71 @@ export const TagListItemMixin = superclass => class extends superclass {
...
+		this.handleClearItem(e);
 	}
 
 	_renderTag(tagContent, hasTruncationTooltip) {
+		const buttonText = typeof tagContent === 'object'
+			? this.localize('components.tag-list.clear', { value: '' })
+			: this.localize('components.tag-list.clear', { value: tagContent });
 		const tooltip = hasTruncationTooltip ? html`
-				<d2l-tooltip for="${this._id}" show-truncated-only>
+				<d2l-tooltip for="${this._id}" offset="20" show-truncated-only>
```
margaree, line 156: This offset is because the div is now within the container so it needs some cushion. This isn't ideal if the tooltip were to open to the right when clearable. It seems unlikely but is possible depending on layout.

components/tag-list/tag-list.js
```diff
@@ -143,6 +160,23 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 		`;
 	}
 
+	updated(changedProperties) {
```
margaree, line 163: I put this in updated rather than just in first update as `clearable` has the potential to change

components/tag-list/tag-list.js
```diff
@@ -222,15 +267,55 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
 
-	_getTagListItems() {
+	async _getTagListItems() {
 		const slot = this.shadowRoot && this.shadowRoot.querySelector('slot');
 		if (!slot) return;
-		return slot.assignedNodes({ flatten: true }).filter((node) => {
+
+		const results = await this._filterAsync(slot.assignedNodes({ flatten: true }), async node => {
 			if (node.nodeType !== Node.ELEMENT_NODE) return false;
+			await node.updateComplete;
```
margaree, line 276: I was finding that it was necessary to await an updateComplete from the slot items, otherwise on clicking the "add tag" button in the demo, it wasn't getting the newest item and wasn't behaving properly (didn't have the role of `listitem` yet).

components/tag-list/tag-list.js
```diff
@@ -249,11 +334,9 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 		if (!this._hasResized) return;
 
 		requestAnimationFrame(async() => {
-			this._items = this._getTagListItems();
+			this._items = await this._getTagListItems();
 			if (!this._items || this._items.length === 0) return;
 
-			await Promise.all(this._items.map(item => item.updateComplete));
```
margaree, line 255: This is no longer needed after moving the updateComplete into `_getTagListItems`

lang/en.js
```diff
@@ -95,6 +95,8 @@ export default {
 	"components.switch.visibility": "Visibility",
 	"components.tabs.next": "Scroll Forward",
 	"components.tabs.previous": "Scroll Backward",
+	"components.tag-list.clear": "Click, press backspace, or press delete key to remove item {value}",
```
margaree, line 98: This feels a bit clunky - any ideas on alternative phrasing?

components/tag-list/demo/tag-list.html
```diff
@@ -9,6 +9,18 @@
 			import '../tag-list.js';
 			import '../tag-list-item.js';
 			import '../test/tag-list-item-mixin-consumer.js';
+
+			window.addEventListener('load', () => {
```
dbatiste, line 13: I _think_ this could be simplified:
```javascript
document.addEventListener('d2l-tag-list-item-cleared', e => {
   e.target.deleteItem();
   ...
});
```

If the script needs to be specific to a snippet, it could also be embedded in fixture instead of outside.

dbatiste: Usually we would place such script at the very bottom of the page, just before the page `body`, to avoid the script blocking the page rendering, but in this case it's not a huge deal.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -58,11 +79,27 @@ export const TagListItemMixin = superclass => class extends superclass {
 					transition: none;
 				}
 			}
+			.tag-list-item-content + d2l-button-icon {
```
dbatiste, line 82: This appears to be targeting the clear button.  Is it possible to apply a specific class to the button so we can avoid the sibling selector?  I am guessing we cannot simply target `d2l-button-icon` since their tag content might contain such an element?

components/tag-list/tag-list-item-mixin.js
dbatiste: So I guess the expectation is that the consuming component would have some array of data, they would remove the item from their array, and re-render?  It seems like consumers wouldn't really have a need to call this, if the tag-list is always a product of it's state.  If we wanted to exclude this method, we could set up a demo component in the demo.  🤷‍♂️

components/tag-list/demo/tag-list.html
```diff
@@ -52,6 +52,27 @@ <h2>Basic Tag List</h2>
...
+					<d2l-tag-list-item-mixin-consumer name="Custom Tag List Item"></d2l-tag-list-item-mixin-consumer>
+					<d2l-tag-list-item text="Example Tag 5"></d2l-tag-list-item>
+					<d2l-tag-list-item text="Example Tag 6"></d2l-tag-list-item>
+					<d2l-tag-list-item text="Example Tag 7"></d2l-tag-list-item>
+				</d2l-tag-list>
+			</d2l-demo-snippet>
+
 		</d2l-demo-page>
 	</body>
+	<script type="module">
```
dbatiste, line 70: Did you run into trouble without the rAF and the `type="module"`?

This seems to work ok for me...
```javascript
<script>
	document.addEventListener('d2l-tag-list-item-cleared', e => {
		e.target.deleteItem();
		console.log(`d2l-tag-list-item-cleared event dispatched. Value: ${e.detail.value}, handleFocus: ${e.detail.handleFocus}`);
	});
</script>
```
I think `<script>...</script>` should be just inside the `</body>`.  I _think_ `head` and `body` are the only valid elements inside of `html`.

margaree: Ah yup that works for me as well! I'd needed more time before when adding the listener to individual items but just adding the listener to the document it's fine.

Ah, I'll move that then. Thanks!

components/tag-list/tag-list.js
```diff
@@ -222,17 +266,57 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
+		if (!this.shadowRoot) {
+			return [];
+		}
+
+		const showMoreButton = this.shadowRoot.querySelector('.d2l-tag-list-button') || [];
+		const clearButton = this.shadowRoot.querySelector('.d2l-tag-list-clear-button') || [];
+		return this._items.slice(0, this._chompIndex).concat(showMoreButton).concat(clearButton);
+	}
+
+	_handleClearAll(e) {
```
dbatiste, line 299: Where does focus go after doing clear-all?  Where _should_ it go?  🤔

components/tag-list/tag-list.js
```diff
@@ -222,15 +267,55 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
+
+	_handleItemDeleted(e) {
+		if (!this.clearable) return;
+		if (!e || !e.detail || !e.detail.handleFocus) return;
+
+		const rootTarget = event.composedPath()[0];
+		const children = this._getVisibleEffectiveChildren();
+		const itemIndex = children.indexOf(rootTarget);
+		if (children.length > 1) {
+			if (children[itemIndex - 1]) children[itemIndex - 1].focus();
```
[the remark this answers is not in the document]
dbatiste: I guess to my Q above... what happens when `children.length === 0`?

components/tag-list/tag-list.js
```diff
@@ -222,17 +266,57 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
+
+		const results = await filterAsync(slot.assignedNodes({ flatten: true }), async node => {
 			if (node.nodeType !== Node.ELEMENT_NODE) return false;
+			await node.updateComplete;
+
 			const role = node.getAttribute('role');
+			if (role !== 'listitem') return false;
+
+			if (this.clearable) node.setAttribute('clearable', 'clearable');
+			node.addEventListener('d2l-tag-list-item-cleared', this._handleItemDeleted.bind(this));
```
dbatiste, line 281: Do we need to wire up the listener to the tag element?  Can we wire up the handler once above in `render` to the tag list container?  It would ensure that we never wire up more than one handler for removing the tag.  It also ensures that the handler runs in the correct scope.  

I know this is a contrived example because we have no intentions of supporting drag & drop of tags, but if a tag were drag & dropped from one tag list to another and then cleared, it looks like a second handler would get wired up but the original one would still run against the original scope. I think it just highlights some of the pitfalls wiring up things across scopes.

margaree: Yup that works as well and does sound like a safer option.

components/tag-list/tag-list.js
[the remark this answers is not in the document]
margaree: The focus seems to stay on `d2l-tag-list` but there's nothing visible so it is a bit weird. Perhaps it should go to the next focusable element in the document? I'll check with Noah on this.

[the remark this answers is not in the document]
dbatiste: If there was a way to add tags, I might expect the focus to get moved to that, but I am not sure.  Or maybe the consumer listens for a clear event and manages the focus themselves.  Not sure.  🤔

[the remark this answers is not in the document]
margaree: There is a future story to add the tag input so agreed that if that existed, then it would make sense to send focus there. I think it would likely make the most sense for the consumer to handle focus in that case. We could leave handling the focus entirely to the consumer, or just note that they would need to handle it for when there are no remaining tag list items.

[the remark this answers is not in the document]
dbatiste: It also seems like a weird configuration where a user would be able to remove tags, but not add them back.

dbatiste (approved):

PR #2322, opened 2022-04-14
Add support for a more compressed HTML block with no top/bottom margins on first/last rendered element respectively

grant-cleary:
One feature we've been asked for over in editor-land is to be able to display what effectively looks like "inline" content (i.e. could be a literal single-line text box). This is challenging right now because the default content styles on the `html-block` include vertical margins that force the block to be taller. This PR tries to address that by providing a configuration option to strip those out on the first/last children of the rendered div.

I'm not really tied to any particular part of this solution, so if there's doubt around whether this is desirable (or if it is, whether this is the right way to approach it), I'm happy to rethink things.


components/html-block/html-block.js
```diff
@@ -11,6 +11,12 @@ export const htmlBlockContentStyles = css`
 		font-weight: 400;
 		line-height: 1.2rem;
 	}
+	.d2l-html-block-no-vertical-margins > *:first-child {
```
grant-cleary, line 14: For a narrower net, I can restrict this to `h1-6`, `p`, `b`, `strong`, etc., though I'm not sure if that's necessary. Any inline styles on the rendered content will already take precedence here.

grant-cleary: This also doesn't work properly if the first (or last) node is a text node (or anything that isn't an element), because the child selectors ignore them. I'm not sure if there's a reasonable work-around for that, unless we apply this CSS dynamically based on what's in the rendered content, which... Well it seems sort of gross to me.

components/html-block/html-block.js
```diff
@@ -194,7 +206,12 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		super.firstUpdated(changedProperties);
 
 		if (this._renderContainer) return;
-		this.shadowRoot.innerHTML += `<div class="d2l-html-block-rendered${this.compact ? ' d2l-html-block-compact' : ''}"></div><slot ${!this.noDeferredRendering ? 'style="display: none"' : ''}></slot>`;
+		this.shadowRoot.innerHTML += '<div class="d2l-html-block-rendered'
+			+ `${this.compact ? ' d2l-html-block-compact' : ''}`
+			+ `${this.noVerticalMargins ? ' d2l-html-block-no-vertical-margins' : ''}`
+			+ '"></div><slot'
+			+ `${!this.noDeferredRendering ? ' style="display: none"' : ''}`
+			+ '></slot>';
```
grant-cleary, line 214: Formatting because the one line was starting to get ridiculously long. I still want to spend time investigating whether we can get Lit rendering to play nicely with MathJax, but... That's for another time.

grant-cleary:
Visual diff changes look as expected for what I'm doing here, though I made some small tweaks to the fixture for the new param to make the last-child CSS more obvious. That said, I'll hold off on merging the new goldens into this PR until we're sure this is what we want.

components/html-block/html-block.js
```diff
@@ -9,7 +9,13 @@ export const htmlBlockContentStyles = css`
 	.d2l-html-block-compact {
 		font-size: 0.8rem;
 		font-weight: 400;
-		line-height: 1.2rem;
+		line-height: 1.2;
```
grant-cleary, line 12: I noticed this while I was in here. Using a fixed default line-height here is extremely problematic because font sizes can be significantly larger than this. I'm unsure whether moving to a unitless ratio is actually ideal though. Would certainly entertain any thoughts on this.

dlockhart: Isn't this now going to be 1.2X the size of the current font size in the context of where the blog is being rendered? For example, with `<div style="font-size:1000px"><d2l-html-block></d2l-html-block></div>` won't that now have a line height of `1200px`? vs. before it was based on the root font size.

grant-cleary: Yeah, that's why I'm unsure what the right answer is here. The previous value of `1.2rem` is problematic because in the case that we have that div, the line height gets pinned to `1.2 * rootFontSize` but still tries to fit a 1000px font in that line. So the lines overlap and things get messy. Basically I'm not sure fixing a line height makes a ton of sense when we're rendering user-authored HTML.

grant-cleary: e.g. At a 42px font size this is what it looks like now in main:
HtmlBlockFontSize

dlockhart: Ah I see, so the flexible line height is more about making things look right in general and less about this particular "get rid of margins" feature. I'm curious... if you remove the `line-height` altogether, what's the inherited value from the outer page?

grant-cleary: Oh, sorry, yes. I conflated these things in my head because on the editor-side I had to make an equivalent change to avoid pop-in when the editor loads. But yeah, this is a separate issue for sure. I'll check what the default is.

grant-cleary: Interesting. On the demo page, nothing is explicitly inherited, which should default it to the initial value of `normal`... But that doesn't seem to happen and everything still overlaps, which isn't what I expected. I'll check the LMS.

grant-cleary: The LMS behaves similarly. Setting `line-height: inherit` explicitly has the same behaviour (different from `line-height: normal`), so despite Chrome being absolutely convinced that we're not setting a line-height anywhere, it doesn't seem to be defaulting properly. 😕

dlockhart: Interesting. And I think in some places in the LMS we're including a reset stylesheet which may mess with the defaults even further.

components/html-block/html-block.js
dbatiste: What's the reason for including the universal selector, `*`?  I think this can be omitted.

grant-cleary: I originally had a more specific selector in there before I realized we would need it for every element, so I just replaced it with `*` to confirm. You're right that it's redundant though, so I'll remove it.

dbatiste:
Just for context - what's the use case driving this?  Is it quizzing... question or multiple choice answers?  Any chance we have a screenshot of what they are trying to accomplish?  That might be helpful for future us, if we have it. No big deal if we don't.

components/html-block/html-block.js
grant-cleary: Yeah, this has made me extra confused about what behaviour is desired/expected in this case. A unitless line-height made sense to my frazzled end-of-week brain because it would maintain existing behaviour if `fontSize === rootFontSize` while also maintaining a consistent ratio of whitespace to font size... But this behaviour now makes me uncertain.

grant-cleary:
> Just for context - what's the use case driving this? Is it quizzing... question or multiple choice answers? Any chance we have a screenshot of what they are trying to accomplish? That might be helpful for future us, if we have it. No big deal if we don't.

Yep, this is for quizzing. Also, I do have such a screenshot! It's in the context of the HTML editor, but it's applicable more broadly as well.

Editor readonly mode (applicable to the html-block changes specifically):

EditorNoVerticalMargins_ReadOnly

Editor write mode (meant to look exactly like the html-block, but separate CSS):

EditorNoVerticalMargins

grant-cleary:
> Yep, this is for quizzing. ...

Note that we hit some weird edge cases pretty quickly as well. The desire for something that looks like a single-line field gets strange if an end user plays with the font size or adds any kind of HTML markup more complicated than plain text... So there are very stringent limitations to this use case and I don't imagine it will see much (if any) use outside of the inline HTML editor. As a sidenote, this will now be two properties that aren't likely to be used outside of the inline editor and that does make me sort of uncomfortable.

components/html-block/html-block.js
```diff
@@ -9,7 +9,13 @@ export const htmlBlockContentStyles = css`
...
+	}
+	.d2l-html-block-rendered > :first-child {
```
grant-cleary, line 14: This will have no impact if `no-deferred-rendering` is set, but in that case none of these styles apply anyways since we're not actually rendering anything into the component's shadowDOM.

components/html-block/html-block.js
```diff
@@ -135,7 +141,12 @@ class HtmlBlock extends RtlMixin(LitElement) {
 			 * unless your HTML relies on script executions that may break upon stamping.
 			 * @type {Boolean}
 			 */
-			noDeferredRendering: { type: Boolean, attribute: 'no-deferred-rendering' }
+			noDeferredRendering: { type: Boolean, attribute: 'no-deferred-rendering' },
+			/**
+			 *  Whether to remove top/bottom margins from first and last rendered elements, respectively.
+			 * @type {Boolean}
+			 */
+			noVerticalMargins: { type: Boolean, attribute: 'no-vertical-margins' }
```
grant-cleary, line 149: Shoot, forgot to remove this.

components/html-block/html-block.js
grant-cleary: As an additional note, these selectors ignore text (or other non-element) nodes, so as can be seen in the demo, things can get weird if the first node isn't an element, because this CSS will happily strip the top margin from the first element node, even if some text node precedes it.

components/html-block/html-block.js
dbatiste: Grant, how is that working above in the screenshot with `42px` font-size and `1.2rem` line-height?  The line-height would equate to `24px`... I would expect things to be overlapping a lot more.  🤔

dbatiste:
The margin change looks good based on the strategy we discussed offline.  The line-height bit I am not sure of.

components/html-block/html-block.js
grant-cleary: Oh yeah, it's much worse in the compact case. The screenshot above isn't from the compact case.

grant-cleary:
> The margin change looks good based on the strategy we discussed offline. The line-height bit I am not sure of.

Yeah, I think now that these both look like more substantial changes, I'm going to split the line-height stuff into its own PR to keep things more sensible.

dbatiste (approved):

PR #2324, opened 2022-04-14
Clarify some focus helper descriptions

bearfriend:


components/tooltip/test/tooltip-truncating.visual-diff.js
```diff
@@ -30,9 +30,11 @@ describe('d2l-tooltip truncating', () => {
 
 		it(testCase.name, async function() {
 			const selector = `#${testCase.name}`;
-			await page.$eval(`${selector}${testCase.focus}`, (content) => {
+			const delay = await page.$eval(`${selector}${testCase.focus}`, (content) => {
 				content.focus();
+				return content.nextElementSibling.delay;
```
bearfriend, line 35: This is probably better.
```suggestion
				return content.parentNode.querySelector('d2l-tooltip').delay;
```

components/tooltip/test/tooltip-truncating.visual-diff.js
```diff
@@ -30,9 +30,11 @@ describe('d2l-tooltip truncating', () => {
...
 			});
+			page.waitForTimeout(delay);
```
bearfriend, line 37: Alternatively:
```suggestion
			await page.waitForSelector(`${selector} d2l-tooltip`, {
				state: 'visible',
				timeout: delay
  			});
```

bearfriend: Whoops.
```suggestion
			await page.waitForTimeout(delay);
```

components/tooltip/test/tooltip-truncating.visual-diff.js
dbatiste: Yeah.  Certainly clearer too!  😄

components/tooltip/test/tooltip-truncating.visual-diff.js
dbatiste: Are the `d2l-tooltip-show` and `d2l-tooltip-hide` events not useful in this case?

bearfriend: Hmm. I guess you'd have to add a listener in an eval and then return and await a promise? I'm not super familiar with playwright.

At that point is it just easier to wait for `d2l-tooltip[showing]`?

dbatiste: If there's any animation associated with showing, we'd want to make sure we wait for that to complete also.

Tooltip actually has some helpers to make things slightly easier (as do a few other components): https://github.com/BrightspaceUI/core/blob/main/components/tooltip/test/tooltip-helper.js

The other visual-diff file makes use of its `show()` helper.

dbatiste: It looks like tooltip doesn't actually wait for the animation to complete before dispatching the event.  The helper probably still works because we emulate `prefers-reduced-motion: reduce` for the duration of the test, and tooltip has CSS to honor it.  Nevertheless, I would use the helper.

dbatiste: Or if [oneEvent](https://github.com/BrightspaceUI/visual-diff/blob/main/helpers/oneEvent.js#L1) is better suited for this particular test, it could be used on its own.  That's what the above helpers use.  It will return a promise.

Strangely, the helpers mentioned above are `async` when they don't need to be.  A promise is already returned from them, and `await` is not being used within.

bearfriend: That `oneEvent` is exactly what I was imagining doing.

svanherk: I fixed the flake for this test recently in a different PR, you _should_ be able to just go back to the readme updates and visual-diff should be good 🤞

dbatiste (approved):

PR #2328, opened 2022-04-18
Added interception of dialog-mixin close event

alexvlsem:
https://d2l.slack.com/archives/C0PHG3QB0/p1649786796542179

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -21,7 +21,6 @@ if (window.D2L.DialogMixin.preferNative === undefined) {
 
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 const abortAction = 'abort';
-const forceCloseAction = 'force-close';
 const defaultMargin = { top: 100, right: 30, bottom: 30, left: 30 };
 
 export const DialogMixin = superclass => class extends RtlMixin(superclass) {
@@ -140,7 +139,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 
 	_close(action) {
 		if (!this._state) return;
-		this._action = action ?? abortAction;
+		this._action = action;
 
 		clearDismissible(this._dismissibleId);
 		this._dismissibleId = null;
@@ -244,7 +243,6 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		// handle "dialog-action" for backwards-compatibility
 		if (!e.target.hasAttribute('data-dialog-action') && !e.target.hasAttribute('dialog-action')) return;
 		const action = e.target.getAttribute('data-dialog-action') || e.target.getAttribute('dialog-action');
-		this._interceptDialogClosing = e.target.hasAttribute('data-intercept-dialog-closing');
 		e.stopPropagation();
 		this._close(action);
 	}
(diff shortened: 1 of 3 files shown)


components/dialog/dialog-mixin.js
```diff
@@ -139,11 +140,16 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
 	_close(action) {
 		if (!this._state) return;
-		this._action = action;
+		this._action = action ?? abortAction;
 
 		clearDismissible(this._dismissibleId);
 		this._dismissibleId = null;
 
+		if (this._isCloseAborted()) {
+			this._dismissibleId = setDismissible(() => this._close(abortAction));
```
jorge-ramirez-arredondo, line 149: What's the reasoning here behind clearing and setting a new dismissible instead of just preventing the original from being dismissed?

alexvlsem: As I understand, this is how init() method triggered by setDismissible() method works. 
https://github.com/BrightspaceUI/core/blob/bc58892cb5d4c32f82a4220c7ff638f9de0bbdd5/helpers/dismissible.js#L19
Each time when Esc is pressed the callback function is deleted from the stack.

alexvlsem: When the dialog is triggered to close by pressing Esc the callback function (_close) is deleted from the stack but if the close is aborted the dialog cannot be close by Esc the next time.

components/dialog/dialog-mixin.js
```diff
@@ -238,6 +244,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		// handle "dialog-action" for backwards-compatibility
 		if (!e.target.hasAttribute('data-dialog-action') && !e.target.hasAttribute('dialog-action')) return;
 		const action = e.target.getAttribute('data-dialog-action') || e.target.getAttribute('dialog-action');
+		this._interceptDialogClosing = e.target.hasAttribute('data-intercept-dialog-closing');
```
dlockhart, line 247: What's the reason for using this `data-` attribute + flag vs. just always firing the `d2l-dialog-abort-close` event when aborting (which isn't composed and doesn't bubble anyway) and then bypassing the abort if the event has `defaultPrevented` (like you're already doing).

Just seems like an extra unnecessary thing folks have to do to opt into this.

components/dialog/dialog-mixin.js
```diff
@@ -298,6 +305,28 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		this._useNative = false;
 	}
 
+	_isCloseAborted() {
+
+		if (this._action === forceCloseAction) {
+			return false;
+		}
+		else if (this._action === abortAction || this._interceptDialogClosing) {
```
dlockhart, line 313: Nit:
```suggestion
		} else if (this._action === abortAction || this._interceptDialogClosing) {
```

dlockhart: Alternately, you could also invert this and add it to the first check which returns early:

```javascript
if (this._action === forceCloseAction || this._action !== abortAction) {
    return false;
}

const abortEvent = new CustomEvent('d2l-dialog-abort-close', {
	cancelable: true,
	detail: {
		forceCloseDialog: this._close.bind(this, forceCloseAction),
		action: this._action
	}
});
this.dispatchEvent(abortEvent);

return abortEvent.defaultPrevented;
```

components/dialog/dialog-mixin.js
alexvlsem: Since the dialog can be closed not only by 'X' and 'Esc' buttons (they assign 'abort' action)  I suggest to introduce 'data-intercept-dialog-closing' attribute for other buttons that are able to trigger dialog closing.

dlockhart: After chatting with @dbatiste, here's a suggestion:

What if instead we called the event something like `dialog-dialog-before-close` and ALWAYS fired it whenever the browser is about to close. So then you don't need the `data-` attribute, and you don't need `this._interceptDialogClosing` and I don't think you need `this._action = action ?? abortAction`.

The event would always fire, and consumers could call `preventDefault()` on it if they wanted to prevent the actual close from occurring, just like you have now. The special `force-close` action shouldn't be required either since the consumer could just _not_ call `preventDefault()` if that's what they wanted.

alexvlsem: Updated according to your suggestion.

dlockhart (approved):
This is looking good to me, thanks for bearing with us as we explored different options! I'll release this right now.

PR #2341, opened 2022-04-18
fix: tag-list: update README to better demo clearable, update development note

margaree:


components/tag-list/README.md
```diff
@@ -31,8 +31,12 @@ The `d2l-tag-list` element can take a combination of any type of `d2l-tag-list-i
 <script type="module">
   import '@brightspace-ui/core/components/tag-list/tag-list.js';
   import '@brightspace-ui/core/components/tag-list/tag-list-item.js';
-</script>
 
+  document.addEventListener('d2l-tag-list-item-cleared', (e) => {
```
dbatiste, line 35: Something I just realized - I think the events should be present tense.  I think this is the practice that native browser events follow and we've tried to as well... although I think a couple have slipped through.

margaree: No one would be using this yet so I'll change that now

dbatiste (approved):

PR #2351, opened 2022-04-20
fix: TagListItemMixin: Use focus-visible instead of focus

margaree:
Related to https://github.com/BrightspaceUILabs/user-profile-card/pull/134

~This is being added to give an option for custom tag list items to have control over when the item looks to be focused.~ In the example PR above, even when the card is opened and the focus goes to the user name, the tag list item "opener" also still appears focused (this can be seen in the failed visual diffs for that PR). ~The consumer of `TagListItemMixin` would then control `hide-focus-style` so that focus styles only show when desired (e.g., when something in the card hasn't taken the focus).~

~I am definitely open to other suggestions on this. I found that using `:focus-visible` instead of `:focus` here is almost a solution but isn't usable for Safari, and the `.focus-visible` class is just always on the custom tag list item so that has the same problem as before.~ Going with the `:focus-visible`


dlockhart:
> I found that using :focus-visible instead of :focus here is almost a solution but isn't usable for Safari...

I think Safari shipped this recently, but likely only in version 15 (in Monterrey). If that's working, then I'd consider that "good enough" since I'm hoping we can soon remove the polyfill anyway.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -63,12 +64,12 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 				overflow: hidden;
 				text-overflow: ellipsis;
 			}
-			:host(:focus) .tag-list-item-container,
-			:host(:focus:hover) .tag-list-item-container {
+			:host(:focus-visible) .tag-list-item-container,
```
dbatiste, line 67: Hmmm, this might require us to use `forceFocusVisible` when moving focus from the "interactive" toggle.

margaree: Ah - does that mean we shouldn't go with `focus-visible` here? Or is that just something that will need to be kept in mind to be tested?

dbatiste: I think we should test this out in conjunction with my "interactive" changes.  I add a demo to the tag-list in my PR for that case.

margaree: Sounds good, I'm good to wait until yours merges

margaree: After merging your changes in this appears to still work as expected both in non-interactive and interactive mode

PR #2354, opened 2022-04-20
Make sure to set MathJax context in html-block demo

grant-cleary:
The README was left slightly inaccurate after some of my recent work. It's required that the `mathjax` context be set any time math is to be rendered.

### DIFF
--- a/components/html-block/README.md
+++ b/components/html-block/README.md
@@ -68,7 +68,7 @@ To use `d2l-html-block` within another Lit component, use the [unsafeHTML](https
 
 ### Rendering MathML and LaTeX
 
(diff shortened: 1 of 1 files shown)


components/html-block/README.md
```diff
@@ -68,14 +68,15 @@ To use `d2l-html-block` within another Lit component, use the [unsafeHTML](https
 
 ### Rendering MathML and LaTeX
 
-Examples are provided to display how user-authored math can be embedded within your webpage.
+Examples are provided to display how user-authored math can be embedded within your webpage. Note that rendering math requires the `mathjax` context to be set correctly. For testing and/or demo pages, you can import `@brightspace-ui/core/tools/mathjax-test-context.js` to set this context for you.
```
dbatiste, line 71: I might be extra explicit and say "For testing and/or demo pages ONLY, ...".  I am worried that even with the console warning, people are going to import this in their non-demo code.

dbatiste (approved):

components/html-block/README.md
grant-cleary: Good point. I'll make that change.

PR #2359, opened 2022-04-21
tag-list - use gap instead of margins for separation

dbatiste:
This PR updates `d2l-tag-list` to use the `gap` CSS property (applies to both flex and grid layouts) instead of margins.  This avoids having an extra margin around any of the outside boundaries of the tag-list, which is important for "interactive" mode since the whole tag list will be surrounded by a focusable container.

To give an idea of what this is... with margins:
image

With gap:
image


components/tag-list/tag-list.js
```diff
@@ -349,7 +343,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 			});
 			this._chomp();
 			this.requestUpdate();
-		});
+		}, 40);
```
dbatiste, line 346: I was getting incorrect layout/dimensions without this delay.  Unfortunately the longer this delay, the more noticeable the FOUC in Safari, though it was still noticeable before these changes.

components/tag-list/tag-list.js
```diff
@@ -214,7 +210,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 
 			for (let i = overflowingIndex; i < this._itemLayouts.length; i++) {
 				const itemLayout = this._itemLayouts[i];
-				const itemWidth = Math.min(itemLayout.width, this._availableWidth);
+				const itemWidth = Math.min(itemLayout.width + MARGIN_TOP_RIGHT, this._availableWidth);
```
dbatiste, line 213: I am going to rename this const to `GAP` - just including it with my interactive changes to avoid conflicts.

margaree (approved):

dbatiste:
Lots of visual-diff changes - all expected due to no extra margin.

PR #2362, opened 2022-04-21
fix: tag-list: accessibility improvements

margaree:
Changes to the following:
- Fix the description being read on VoiceOver
- Improve keyboard navigation experience for screen reader users

The keyboard changes are a bit "bold" so let me know if they disobey some sort of recommended pattern.

### DIFF
--- a/components/tag-list/tag-list.js
+++ b/components/tag-list/tag-list.js
@@ -145,8 +145,9 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 				</d2l-button-subtle>
 			` : html`
 				<d2l-button-subtle
-					class="d2l-tag-list-button"
+					class="d2l-tag-list-button d2l-tag-list-button-show-more"
 					@click="${this._toggleHiddenTagVisibility}"
+					description="${this.localize('components.tag-list.show-more-description')}"
 					slim
 					text="${this.localize('components.tag-list.num-hidden', { count: hiddenCount })}">
 				</d2l-button-subtle>
@@ -158,7 +159,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 		};
 
 		const list = html`
-			<div role="list" class="tag-list-container" aria-describedby="d2l-tag-list-description" @d2l-tag-list-item-clear="${this._handleItemDeleted}">
+			<div role="list" class="tag-list-container" aria-label="${this.description}" @d2l-tag-list-item-clear="${this._handleItemDeleted}">
 				<slot @slotchange="${this._handleSlotChange}"></slot>
 				${overflowButton}
 				<d2l-button-subtle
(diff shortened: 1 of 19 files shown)


components/tag-list/tag-list.js
```diff
@@ -158,7 +159,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 		};
 
 		const list = html`
-			<div role="list" class="tag-list-container" aria-describedby="d2l-tag-list-description" @d2l-tag-list-item-clear="${this._handleItemDeleted}">
+			<div role="list" class="tag-list-container" aria-label="${this.description}" @d2l-tag-list-item-clear="${this._handleItemDeleted}">
```
margaree, line 162: This uses `aria-label` for communicating the description in order to fix the issue where it wasn't being read in VoiceOver

components/tag-list/tag-list.js
```diff
@@ -179,13 +180,12 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
 				<d2l-button-subtle aria-hidden="true" slim text="${this.localize('components.tag-list.num-hidden', { count: '##' })}" class="d2l-tag-list-hidden-button"></d2l-button-subtle>
 				${this.arrowKeysContainer(list)}
-				<div id="d2l-tag-list-description" hidden>${this.description}</div>
 			</div>
 		`;
 	}
 
 	async arrowKeysFocusablesProvider() {
-		return this._showHiddenTags ? this._items : this._items.slice(0, this._chompIndex);
+		return this._getVisibleEffectiveChildren();
```
margaree, line 188: I changed the keyboard arrow key work flow to include the clear and show more/less buttons. In my opinion, when using a screenreader this made the workflow more intuitive (e.g., arrow key onto the "+ X more" button, the user immediately knows there are more hidden tags rather than not knowing that until they tab onto that button)

components/tag-list/tag-list.js
```diff
@@ -352,14 +353,17 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
-	async _toggleHiddenTagVisibility() {
+	async _toggleHiddenTagVisibility(e) {
 		this._showHiddenTags = !this._showHiddenTags;
 
 		if (!this.shadowRoot) return;
 
 		await this.updateComplete;
-		const button = this.shadowRoot.querySelector('.d2l-tag-list-button');
-		if (button) button.focus();
+		if (e.target.classList.contains('d2l-tag-list-button-show-more')) this._items[this._chompIndex].focus();
```
margaree, line 362: This switches the focus behaviour such that when the user selects the "+ X more" button, instead of their focus being taken to the "Show less" button, it is taken to the next list item that they wouldn't have seen/heard yet.

dbatiste (approved):

PR #2366, opened 2022-04-21
fix: tag-list: hide content until calculations completed on slot change

margaree:
On Safari there would be a brief FOUC on page load that would show all the slot items and then soon after chomp the ones necessary. This change hides the tag list items until calculations are made. It does this on each slot change rather than just the first because I noticed on chrome that when items were being added to the slot, the FOUC could be seen there. This change does lead to a brief flicker on slot change on chrome but I think it's an improvement over the visible-then-hidden items.


components/tag-list/tag-list.js
```diff
@@ -68,11 +69,16 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 				visibility: hidden;
 			}
 			.d2l-tag-list-clear-button {
+				position: absolute;
```
margaree, line 72: This was added because of the clear button taking up space invisibly in list (I can remove if this change is already being made elsewhere - just thought I'd throw it in while I was here)

components/tag-list/tag-list.js
```diff
@@ -314,37 +327,37 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
-			});
-			this._chomp();
-			this.requestUpdate();
-		}, 40);
+		this._itemLayouts = this._getItemLayouts(this._items);
+		this._itemHeight = this._items[0].offsetHeight;
+		this._items.forEach((item, index) => {
+			item.setAttribute('tabIndex', index === 0 ? 0 : -1);
+		});
+		this._chomp();
```
margaree, line 359: This is just indentation changes

components/tag-list/tag-list.js
```diff
@@ -314,37 +327,37 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
...
-				this._chompIndex = 10000;
-				return;
-			}
+		this._items = await this._getTagListItems();
+		if (!this._items || this._items.length === 0) {
+			this._chompIndex = 10000;
+			return;
+		}
 
-			this._itemLayouts = this._getItemLayouts(this._items);
```
dbatiste, line 340: I think without the `setTimeout`, `_getItemLayouts()` seems to be getting incorrect widths.  I log the `itemWidth` in the `_chomp()` method and compare that to what the browser console says.  The browser seems to need some time to do layout... not sure why...?

components/tag-list/tag-list.js
dbatiste: Looks great!  My PR will have some ugly conflicts - I will just get your changes and reapply mine on top to use the mixin.

components/tag-list/tag-list.js
margaree: Interesting - I'm getting the correct widths in `_chomp`. Is this just on Safari or also on Chrome?

dbatiste: Hmmm. I am using Chrome right now.

dbatiste: I updated my Chrome and now I get proper dimensions.  Interestingly, Safari is also getting proper values now.  So strange.  🤯

dbatiste (approved):

PR #2368, opened 2022-04-21
fix: Dialog: Move location of documentation of d2l-dialog-before-close

margaree:
The comment documentation of `d2l-dialog-before-close` doesn't work as expected (description ends up empty) because it needs to be above the `dispatchEvent`, which also needs to contain the event name (e.g., the actual contents of `abortEvent` would need to be within `this.dispatchEvent`). This is a bit complicated in this case because of the function returning `abortEvent.defaultPrevented` and so in this case it makes sense to document in each case that is using the `DialogMixin`.


dlockhart (approved):
Oops, thanks!

PR #2370, opened 2022-04-21
Use unitless line-heights for html-block contents

grant-cleary:
As promised, here's a draft of what unitless line height changes would need to look like if we wanted to stay consistent with existing styling for our default fonts. Comparison images below (before on the left, after on the right):

Normal sized text (0.95rem for `p` tag):

HtmlBlockLineHeights_NormalText_Comparison

Large sized text (everything wrapped in a 58px font div to emphasize the overlap):

HtmlBlockLineHeights_LargeText_Comparison


components/html-block/html-block.js
```diff
@@ -9,7 +9,10 @@ export const htmlBlockContentStyles = css`
 	.d2l-html-block-compact {
 		font-size: 0.8rem;
 		font-weight: 400;
-		line-height: 1.2rem;
+		line-height: 1.5;
```
grant-cleary, line 12: 1.5 = 1.2rem / 0.8rem

components/html-block/html-block.js
```diff
@@ -9,7 +9,10 @@ export const htmlBlockContentStyles = css`
...
+	}
+	.d2l-html-block-rendered {
+		line-height: 1.47;
```
grant-cleary, line 15: 1.47 ~= 1.4rem / 0.95rem

components/html-block/html-block.js
```diff
@@ -22,32 +25,32 @@ export const htmlBlockContentStyles = css`
...
 	h5 {
 		font-size: 0.83em;
-		line-height: 16px;
+		line-height: 1.01;
 		margin: 22.13px 0;
 	}
 	h6 {
 		font-size: 0.67em;
-		line-height: 13px;
+		line-height: 1.02;
```
grant-cleary, line 53: These numbers are gross.

components/html-block/html-block.js
dbatiste: In this particular case, there would be no difference under any circumstance, given that the font-size is based on `rem`, correct?

components/html-block/html-block.js
```diff
@@ -22,32 +25,32 @@ export const htmlBlockContentStyles = css`
 	}
 	h1 {
 		font-size: 2em;
-		line-height: 37px;
+		line-height: 0.97;
 		margin: 21.43px 0;
```
dbatiste, line 29: Aside from your change... these fractional values are pretty useless imo.  I am sure they were calculated from somewhere at some point.

components/html-block/html-block.js
grant-cleary: In the case that the default font size is used, yes. Because `line-height` is based on the element's font, this does end up needing to be unitless to cover the case where someone changes the font-size on an element-level.

components/html-block/html-block.js
grant-cleary: Haha probably a similar thing to what I did. To keep things as close as possible to existing behaviour I just did the divisions and rounded to two decimal places. I do wonder how useful they are in the line-height case because they're all so very close to 1.

components/html-block/html-block.js
```diff
@@ -22,32 +25,32 @@ export const htmlBlockContentStyles = css`
...
 	h3 {
 		font-size: 1.2em;
-		line-height: 23px;
+		line-height: 1;
 		margin: 18.72px 0;
 	}
 	h4 {
 		font-size: 1em;
-		line-height: 20px;
+		line-height: 1.05;
```
dbatiste, line 43: At `20px` font-size, `0.05` translates to `1px` - maybe worthwhile.  The extra `0.01` for `h5` doesn't appear to accomplish much.  Just contemplating whether we need/want to use such values.  🤷

grant-cleary: Yeah, that's likely true. This one is definitely an outlier in how far away it is from 1 (comparatively). The others are so close that they do really come down to decimal points for any "reasonable" font size.

components/html-block/html-block.js
dlockhart: Might be worth putting these calculations in a comment so we can remember how they're derived and in case we ever need to re-calculate them based on changes to our root sizes.

components/html-block/html-block.js
dlockhart: Yeah agreed... seems like we could round these all down to `1` and nobody would notice.

dbatiste (approved):
I am a little surprised that some of the lines in the new visual-diff wrapped.  Otherwise looks good.

grant-cleary:
> I am a little surprised that some of the lines in the new visual-diff wrapped. Otherwise looks good.

Ah... https://github.com/BrightspaceUI/core/blob/main/components/html-block/html-block.js#L152

So that specific visual diff test would just need to use a wider viewport to prevent the breaks.

dlockhart (approved):

PR #2375, opened 2022-04-22
Initial InteractiveMixin for Tag List within Interactive Elements

dbatiste:
This PR adds a new mixin, `InteractiveMixin` for use by components that use arrow-keys, as descendants of other components that also use the arrow keys.  In effect, if the "interactive" component is a descendant of an element that is a `role="grid"` or a `d2l-list`, then this behaviour will be enabled. This enables using such components together without conflicting arrow-key behaviour (one such problem being the inner component trapping the focus forever) or confusion for keyboard/screen reader users as they move focus between contexts with arrow keys.  

When the component is in such a context, keyboard users will land on a button to toggle the component into interactive mode via `Enter` key.  Once in interactive mode, keyboard users can use the `Escape` key to exit.  When the component is not in interactive mode, focus will skip past the focusable elements within the component. Mouse users can interact normally with the component, and moving the focus in/out of the interactive contents using the mouse automatically toggles the interactive mode in case users switch from mouse to keyboard, or vice versa.

The strategy is to wrap the interactive component, and manage the focus by intercepting focus entering or leaving the interactive contents.

Non-interactive mode (focus on toggle):
image

Interactive mode (focus on contents):
image


components/tag-list/tag-list.js
```diff
@@ -48,7 +49,7 @@ class TagList extends LocalizeCoreElement(ArrowKeysMixin(LitElement)) {
 	}
 
 	static get styles() {
-		return css`
+		return [super.styles, css`
```
dlockhart, line 52: Not sure which is better/clearer, this or `InteractiveMixin.styles`?

mixins/interactive-mixin.js
```diff
@@ -0,0 +1,115 @@
...
+
+	async _handleInteractiveEndFocus() {
+		// focus moved to trap-end either forwards from contents or backwards from outside - focus interactive toggle
+		this._interactive = false;
+		await this.updateComplete;
+		this.shadowRoot.querySelector('.interactive-container').focus();
+	}
+
+	async _handleInteractiveKeyDown(e) {
+		if (!this._hasInteractiveAncestor) return;
```
dlockhart, line 71: I don't think the event handler would have even been wired up if `_hasInteractiveAncestor` were `false`?

mixins/interactive-mixin.js
```diff
@@ -0,0 +1,115 @@
...
+		const parentGrid = findComposedAncestor(this.parentNode, node => {
+			return ((node.tagName === 'D2L-LIST' && node.grid) || (node.nodeType === Node.ELEMENT_NODE && node.getAttribute('role') === 'grid'));
+		});
+		this._hasInteractiveAncestor = (parentGrid !== null);
+	}
+
+	focus() {
+		if (!this.shadowRoot) return;
+		if (this._hasInteractiveAncestor && !this._interactive) forceFocusVisible(this.shadowRoot.querySelector('.interactive-container'));
+		else this._focusInteractiveContent();
```
dlockhart, line 50: Having them implement this method makes the contract a little weird... how awkward would it be if they passed this in alongside the label as a delegate function to call?

mixins/interactive-mixin.js
```diff
@@ -0,0 +1,115 @@
...
+			if (nextFocusable) nextFocusable.focus();
+		} else {
+			// focus moved to trap-start backwards from within contents - toggle to non-interactive and apply focus
+			this._interactive = false;
+			await this.updateComplete;
+			this.shadowRoot.querySelector('.interactive-container').focus();
+		}
+	}
+
+	_renderInteractiveContainer(inner, label) {
```
dlockhart, line 96: Should this method be "private" since it's the main entrypoint to the mixin? I know you're going of treating it more like a protected method... but can/should mixins have public methods if they're not going to collide with each other?

mixins/interactive-mixin.md
```diff
@@ -0,0 +1,23 @@
...
+
+class MyComponent extends InteractiveMixin(LitElement) {
+
+  render() {
+    return this._renderInteractiveContainer(
+      html`
+        <div>interactive content</div>
+      `, 'My Label'
+    );
+  }
```
dlockhart, line 20: You may want to add a note about _not_ implementing `focus()` in the component and show how to provide an alternate.

mixins/interactive-mixin.js
dbatiste: That's a good point.  I think I might have written this before the conditional rendering was in place.  I think this can be removed.

components/tag-list/tag-list.js
dbatiste: Hmmm, I get JavaScript error when trying `InteractiveMixin.styles`.

mixins/interactive-mixin.js
dbatiste: Yep, that would be fine too.  👍

mixins/interactive-mixin.js
dbatiste: Yeah, it probably should be public.  I don't think we've been entirely consistent with this.  In some cases I think they been marked at private because we don't want them to be used publicly, but rather only within our stuff.  In this case though, I don't think that's the case. An argument could be made they should still be public though.

components/tag-list/tag-list.js
dlockhart: Ah ok cool, just figured since it was `static` you could do that. No worries.

dlockhart (approved):
Cool!

PR #2377, opened 2022-04-22
Correct order of compact CSS styles

grant-cleary:
Just messed up some CSS ordering with my last PR. :(


components/html-block/html-block.js
```diff
@@ -20,6 +15,11 @@ export const htmlBlockContentStyles = css`
 	.d2l-html-block-rendered > :last-child {
 		margin-bottom: 0;
 	}
+	.d2l-html-block-compact {
+		font-size: 0.8rem;
+		font-weight: 400;
+		line-height: 1.5; /* 1.2rem / 0.8rem */
+	}
```
grant-cleary, line 22: Both this and the `.d2l-html-block-rendered` selectors have the same specificity when rendered in the HTML block proper, so the compact styles were always being overridden by the new 1.47 standard line-height.

dlockhart (approved):

PR #2385, opened 2022-04-26
Add comment indicating importance of d2l-html-block-rendered class

grant-cleary:
The `d2l-html-block-rendered` class will soon be used to target CSS on things outside of the html-block component, so we should be sure it doesn't get changed lightly.

Alternatively, if we want to switch up the CSS to target and use a different class, that's also fine... Though the same caveat applies in that case.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -203,6 +203,8 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		super.firstUpdated(changedProperties);
 
 		if (this._renderContainer) return;
+
+		// The d2l-html-block-rendered class is used to apply CSS outside of the html-block component. Do not change lightly.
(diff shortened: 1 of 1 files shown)


components/html-block/html-block.js
```diff
@@ -203,6 +203,8 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		super.firstUpdated(changedProperties);
 
 		if (this._renderContainer) return;
+
+		// The d2l-html-block-rendered class is used to apply CSS outside of the html-block component. Do not change lightly.
```
grant-cleary, line 207: There's already a test that will implicitly fail if we change this class. It's relying on querying `.d2l-html-block-rendered` and returning something with valid inner HTML.

dlockhart (approved):

PR #2386, opened 2022-04-26
InteractiveMixin A11y Fixes

dbatiste:
This PR addresses two a11y issues with the `InteractiveMixin`:

1. VoiceOver + Safari: reads the label, not the description/instructions when tabbing to the interactive wrapper. Pressing `Enter` does move focus into the tag list but nothing is announced about the tags
2. NVDA + Firefox: reads the label and the description/instructions but it prevents the user from toggling when pressing `Enter` (without NVDA on the keyboard interactions work fine)

The solution to the above issues is to change the rendering so that the toggle is an offscreen `button` (keyboard access only) that is sibling to the interactive contents, instead of nesting the contents within a toggle `div`.  Seemingly, screen readers do not like it when focusables are nested within other focusables.

### DIFF
--- a/components/tag-list/demo/tag-list.html
+++ b/components/tag-list/demo/tag-list.html
@@ -73,7 +73,7 @@
 
 			<h2>Tag List with Interactive</h2>
 			<d2l-demo-snippet fullscreen>
-				<div role="grid">
+				<div grid>
 					<d2l-tag-list description="A bunch of example tags" clearable>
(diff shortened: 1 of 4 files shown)


margaree (approved):
Re-tested with NVDA and VoiceOver and this works great!

dbatiste:
> Re-tested with NVDA and VoiceOver and this works great!

Thanks!

PR #2390, opened 2022-04-27
Fix list-header height so buttons don't get clipped when sticking.

dbatiste:
This PR updates the min-height to make room for buttons while sticking.  This was previously larger but [was changed when list-item layouts where updated](https://github.com/BrightspaceUI/core/commit/b4f804bacf5337314eee0da3cc1b16eedf4ef2dc).

Before:
image

After:
image


dbatiste:
I think I am going to try and render a slimmer layout when there are no actions in the slot and `select-all-pages-allowed` is turned off (because it renders a subtle button that also requires the larger min-height).  That should address the filter case better.

margaree (approved):

PR #2396, opened 2022-04-28

dlockhart:
Safari shipped native `<dialog>`s in version 15.4 and our fullscreen dialog in mobile viewports is very broken. I'm also seeing focus-trap issues when the native `<dialog>` is used.

Since this is affecting production and we need to cert & hotfix it quickly, this is a temporary fix that will just cause Safari to fall back to the non-native dialog solution that it's been using previously. That'll buy us time to address this properly.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -14,7 +14,8 @@ import { tryGetIfrauBackdropService } from '../../helpers/ifrauBackdropService.j
 window.D2L = window.D2L || {};
 window.D2L.DialogMixin = window.D2L.DialogMixin || {};
 
-window.D2L.DialogMixin.hasNative = (window.HTMLDialogElement !== undefined);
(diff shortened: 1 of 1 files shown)


components/dialog/dialog-mixin.js
```diff
@@ -14,7 +14,8 @@ import { tryGetIfrauBackdropService } from '../../helpers/ifrauBackdropService.j
 window.D2L = window.D2L || {};
 window.D2L.DialogMixin = window.D2L.DialogMixin || {};
 
-window.D2L.DialogMixin.hasNative = (window.HTMLDialogElement !== undefined);
+window.D2L.DialogMixin.hasNative = (window.HTMLDialogElement !== undefined)
+	&& (navigator.vendor && navigator.vendor.toLowerCase().indexOf('apple') === -1);
```
dlockhart, line 18: I tested this and the vendor is still "Apple Computer Inc." even in Chrome/Firefox on iOS, which is what we want.

dbatiste (approved):

PR #2398, opened 2022-04-28
Visual Diff Flake

svanherk:
Try to address the visual-diff flake with the mobile dropdown in nested dialog tests


svanherk:
Ran this 5 times and the mobile dropdown dialog tests didn't fail, but it wouldn't be the first time they've tricked me like that.  Hopefully this helps!

dlockhart (approved):

PR #2402, opened 2022-04-28

dlockhart:
April hotfix version of #2396.


dbatiste (approved):

jameswklassen (commented):
This is great. Thanks for the quick action on this.

PR #2405, opened 2022-04-28

dlockhart:
I needed to redo this since when I backported some of the colour stuff I messed up and the existing `1.242.x` branch wasn't pointing at the right place. I've fixed that and we should be good now.


jameswklassen (approved):

dlockhart:
Background on what actually happened with that backport was:
- In the `1.x` branch, I did a `fix: update package-lock` just to get things updated and not pollute the visual diff. But then when that merged it tagged & released as `1.242.4`.
- Then the actual colour changes went into `1.243.0`.
- Then today when I went back to `1.242.x` and did a `fix:` it tried to release `1.242.4` again since `1.242.3` was the last tag on that branch and the release blew up.

The "fix" here to blow away the `1.242.x` branch and repoint it at the `1.242.4` tag isn't technically correct since it does pick up those unwanted `package-lock` changes. But going through them... they're all dev tooling changes and I don't think will have any impact on how BSI is going to build core, so we should be good.

dbatiste (approved):

dbatiste (approved):

PR #2408, opened 2022-04-29
Dlockhart/safari div dialog 1.x

dlockhart:
Hopefully the last one! Portfolio is a FRA pointing at core `1.x` so we actually need to put this fix into the `1.x` branch as well. This should release as `1.243.1` on top of the ferrite colour changes.


dbatiste (approved):

jameswklassen (approved):
Thanks 😄 

🤞 this is the last one

PR #2413, opened 2022-04-29
Update list-items to support interactive element in list-item content.

dbatiste:
This PR updates the list-item components to support interactive elements inside their content.  The consumer must specify `no-primary-action` to ensure the interactive elements are not blocked by action layers for mouse users.

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -43,7 +43,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			 * Whether to constrain actions so they do not fill the item. Required if slotted content is interactive.
 			 * @type {boolean}
 			 */
-			noPrimaryAction: { type: Boolean, attribute: 'no-primary-action' },
+			noPrimaryAction: { type: Boolean, attribute: 'no-primary-action', reflect: true },
 			/**
 			 * @ignore
 			 */
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -198,7 +198,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			<slot name="control-action" class="d2l-cell" data-cell-num="3"></slot>
 			<slot name="control" class="d2l-cell" data-cell-num="4"></slot>
 			<slot name="actions" class="d2l-cell" data-cell-num="6"></slot>
-			<slot name="content" class="d2l-cell" data-cell-num="7" @focus="${this._preventFocus}"></slot>
+			<slot name="content" class="d2l-cell" data-cell-num="7" @focus="${!this.noPrimaryAction ? this._preventFocus : null}"></slot>
 			<slot name="nested"></slot>
 		`;
 	}
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -1,10 +1,7 @@
 import { css, html, LitElement } from 'lit';
 import { findComposedAncestor, getNextAncestorSibling, getPreviousAncestorSibling, isComposedAncestor } from '../../helpers/dom.js';
-import {
-	getComposedActiveElement,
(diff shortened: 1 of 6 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -186,7 +198,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			<slot name="control-action" class="d2l-cell" data-cell-num="3"></slot>
 			<slot name="control" class="d2l-cell" data-cell-num="4"></slot>
 			<slot name="actions" class="d2l-cell" data-cell-num="6"></slot>
-			<slot name="content" @focus="${this._preventFocus}"></slot>
+			<slot name="content" class="d2l-cell" data-cell-num="7" @focus="${!this.noPrimaryAction ? this._preventFocus : null}"></slot>
```
dbatiste, line 201: Adding `d2l-cell` class and `data-cell-num` includes this element in the search for focusables when using arrow keys to move among the cells/focusables within cells.

components/list/list-item-generic-layout.js
```diff
@@ -276,14 +288,18 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
 		let focusable = null;
 		let siblingNum = 1;
 		while (!focusable || siblingNum < num) {
 			node = this._getNextSiblingInCell(node);
+
 			if (!node) break;
 			++siblingNum;
+
 			focusable = isFocusable(node, true) ? node : getFirstFocusableDescendant(node);
+			if (isInteractiveDescendant(focusable)) focusable = null;
```
dbatiste, line 302: We need to prevent the generic layout's logic from trying to focus on focusable elements within an interactive widget such as tag-list.  Without this check, the logic would find and focus on them because such widgets in non-interactive mode are not actually inert (because they are still interactive with the mouse), even though it should skip past them.

components/list/list-item-generic-layout.js
```diff
@@ -42,6 +39,11 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			 * @type {'content'|'control'}
 			 */
 			alignNested: { type: String, reflect: true, attribute: 'align-nested' },
+			/**
+			 * Whether to constrain actions so they do not fill the item. Required if slotted content is interactive.
+			 * @type {boolean}
+			 */
+			noPrimaryAction: { type: Boolean, attribute: 'no-primary-action' },
```
dlockhart, line 46: I think you want `reflect: true` since you're using CSS attribute selectors.

mixins/arrow-keys-mixin.js
```diff
@@ -96,6 +96,7 @@ export const ArrowKeysMixin = superclass => class extends superclass {
 		} else {
 			return;
 		}
+		e.stopPropagation();
```
dbatiste, line 99: Without this, the event would bubble up and the list-item's generic layout code would take control of focus when we don't want it to.  We want focus to be managed by the interactive component until they exist interactive mode.

mixins/interactive-mixin.js
```diff
@@ -70,17 +77,19 @@ export const InteractiveMixin = superclass => class extends LocalizeCoreElement(
...
-					</button>
+				<button
+					class="interactive-toggle d2l-offscreen"
+					@blur="${this._handleInteractiveToggleBlur}"
+					@click="${this._handleInteractiveToggleClick}"
+					@focus="${this._handleInteractiveToggleFocus}"
+					tabindex="${ifDefined(this._hasInteractiveAncestor && !this._interactive ? '0' : '-1')}">
+					${`${label}, ${this.localize('components.interactive.instructions')}`}
+				</button>
+				<div class="interactive-trap">
```
dbatiste, line 88: This extra `div` just makes it easier for the above helper method to detect when focus has moved into the content (excluding the `button`).

dlockhart (approved):

PR #2414, opened 2022-04-29
Work around lack of newline support in MathJax 3

grant-cleary:
So... Yeah... Not super keen on this, but I ran through some of the options I found more appealing and came up somewhat short. The root of the problem here is that MathJax 3 claims not to support automated linebreaking. This is true, but also misleading, as this evidently extends to **all** linebreaking. Unfortunately MathJax doesn't even try to render newlines and once it's done typesetting, any linebreaks look exactly like spaces. e.g.:
```
<mspace></mspace>                     -> <mjx-mspace></mjx-mspace>
<mspace linebreak="newline"></mspace> -> <mjx-mspace></mjx-mspace>
```
As such, we can't actually identify the difference between a newline and a space post-typesetting. Similarly, MathJax explodes if you try to replace the above with `<p></p>` or `<br>` (which is also weird, because the MathJax 3 docs claim that they convert `<br>` to newlines... This appears to be false), so we can't really do a replacement pre-typesetting. The only thing MathJax persists that we can meaningfully set is the class and style on the element.

Sprouting a style seems marginally cleaner, since the code can live in `mathjax.js` in its entirety. Sprouting a class also works, but it does mean we would need changes in `html-block` and in the CSS injected into HTML files to apply the appropriate styling.


helpers/mathjax.js
```diff
@@ -64,14 +64,19 @@ export class HtmlBlockMathRenderer {
 			return elem;
 		}
 
+		// MathJax 3 does not support newlines, but it does persist styles, so add custom styles to mimic a linebreak
+		// This work-around should be removed when linebreaks are natively supported.
+		const inner = elem.innerHTML.replaceAll('<mspace linebreak="newline">', '<mspace linebreak="newline" style="display: block; height: 1rem;">');
```
grant-cleary, line 69: I hate everything about this.

dbatiste:
So the idea is that once MathJax supports the expected  line-breaking behaviour, we remove this?

I agree with the inline style.  Moving the styles to a class seems to complicate things.

Is there a MathJax GitHub issue where this is being discussed?  If so, it would not be a bad idea to link it here and add a comment to the code as well so we know what to check in terms of support.

grant-cleary:
> So the idea is that once MathJax supports the expected line-breaking behaviour, we remove this?

That's right, yeah.

> Is there a MathJax GitHub issue where this is being discussed? If so, it would not be a bad idea to link it here and add a comment to the code as well so we know what to check in terms of support.

There is. I can do that for sure.

dbatiste (approved):
I hesitate to begin adding visual-diff tests for individual MathML features that MathJax may or may not support.  There might be value in this case since we have this specific work-around in place, and could be sensitive to MathJax version changes.

grant-cleary:
> I hesitate to begin adding visual-diff tests for individual MathML features that MathJax may or may not support. There might be value in this case since we have this specific work-around in place, and could be sensitive to MathJax version changes.

That's a good point, yeah. I think adding some multi-line math to the existing math visual diff test would be relatively uncontroversial.

grant-cleary:
Haha as expected, the MathML visual diff tests don't quite work as expected. I notice the visual-diff HTML page also has this issue, so I'll dig in a bit.

grant-cleary:
I think some of the previous html-block/MathJax work must have fixed the issues with the visual diff tests. Either that or we're about to get a string of flaky failures. 😬

PR #2418, opened 2022-05-02
fix: content not available early in mobile rendering pipeline

dlockhart:
Noticed this JS error gets thrown during dialog initialization in `_updateOverflow` if the viewport is initially small enough to fall into the "mobile" category.

The reason is that `_updateSize` is called from `resize()` which is called from `_handleResize()` which is called [from Dialog's constructor](https://github.com/BrightspaceUI/core/blob/main/components/dialog/dialog.js#L109). In Lit 2, the constructor runs before there's a shadow root.

In mobile mode, the code skips past all [this autoSize logic](https://github.com/BrightspaceUI/core/blob/main/components/dialog/dialog-mixin.js#L461) which has a bunch of `await this.updateComplete`s in it and therefore defers the run long enough for the component to have rendered.

### DIFF
--- a/components/dialog/dialog.js
+++ b/components/dialog/dialog.js
@@ -106,7 +106,6 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 		this.describeContent = false;
 		this.width = 600;
 		this._handleResize = this._handleResize.bind(this);
-		this._handleResize();
 	}
 
 	get asyncContainerCustom() {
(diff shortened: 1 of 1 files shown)


dlockhart:
Agreed yeah. It was added here:
https://github.com/BrightspaceUI/core/commit/2e09c4e1af8ee637ea6d4b9ff9d8d3381e436567

I could move that call into the `connectedCallback` as well if you think that might be safer.

dbatiste:
> Agreed yeah. It was added here: [2e09c4e](https://github.com/BrightspaceUI/core/commit/2e09c4e1af8ee637ea6d4b9ff9d8d3381e436567)
> 
> I could move that call into the `connectedCallback` as well if you think that might be safer.

I think there's a `setTimeout(,0)`, and the various `await this.updateComplete` calls, so the timing probably "just works" for `constructor` or `connectedCallback` most of the time, even though most of the code relies on the local DOM having been constructed.

dlockhart:
After a bit more discussion, I've change this so that the call from dialog happens in `firstUpdated`, which will guarantee that the local DOM has been constructed. Fullscreen dialogs are different in that they don't make the call down to the mixin's `resize()`, so I've left them as-is for now.

dbatiste (approved):

PR #2424, opened 2022-05-02
fix: exclude d2l-focus-trap for native dialogs

dlockhart:
Native `<dialog>` elements essentially implement the `inert` behaviour in that they come with their own focus trap logic, so we don't need `<d2l-focus-trap>` when native dialogs are used.

The only difference between theirs and ours is that focus will cycle up to the browser chrome and then back into the window, just like it normally does -- which I think is better anyway.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -14,6 +14,8 @@ import { tryGetIfrauBackdropService } from '../../helpers/ifrauBackdropService.j
 window.D2L = window.D2L || {};
 window.D2L.DialogMixin = window.D2L.DialogMixin || {};
 
+// while implemented in Webkit, native <dialog> focus mangement across slotted content is buggy
+// https://bugs.webkit.org/show_bug.cgi?id=233320
 window.D2L.DialogMixin.hasNative = (window.HTMLDialogElement !== undefined)
 	&& (navigator.vendor && navigator.vendor.toLowerCase().indexOf('apple') === -1);
 if (window.D2L.DialogMixin.preferNative === undefined) {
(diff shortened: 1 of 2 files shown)


components/dialog/dialog-mixin.js
```diff
@@ -14,6 +14,8 @@ import { tryGetIfrauBackdropService } from '../../helpers/ifrauBackdropService.j
 window.D2L = window.D2L || {};
 window.D2L.DialogMixin = window.D2L.DialogMixin || {};
 
+// while implemented in Webkit, native <dialog> focus mangement across slotted content is buggy
+// https://bugs.webkit.org/show_bug.cgi?id=233320
```
dlockhart, line 18: I was unable to get native dialogs working in Safari because of this issue ^. I applied some fixes (coming in a separate PR) to the styling so that the original problem was resolved, but focus management is still buggy. So we'll need to live with this opt-out for a bit.

components/tag-list/test/tag-list.visual-diff.js
```diff
@@ -21,28 +21,28 @@ describe('d2l-tag-list', () => {
 
 	it('is correct at 1400px page width', async function() {
 		const rect = await visualDiff.getRect(page, '#default');
-		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false, clip: rect });
```
dlockhart, line 24: Trying to address some flake here:
https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/tag-list/2022-05-02-20.34.45.71-d27e74a00cf834822f1857e5a2ecf216b200f8ca-report.html

dbatiste (approved):
Looks good!

PR #2427, opened 2022-05-03
fix: aligning how fullscreen and normal dialogs are styled in mobile

dlockhart:
This came up as part of my investigation into the native `<dialog>` issue in Safari.

Despite looking and behaving the same, there's some variation in how normal and fullscreen dialogs are sized and styled when they switch into mobile mode. Sizing of normal dialogs is done by simply positioning them `42px` from the top of the viewport and setting the height to the viewport height minus that. Flexbox does the rest. Fullscreen dialogs on the other hand were relying on manually setting the height of their contents each time the viewport resized -- this variation was the source of the original Safari/dialog issue.

Additionally, there was a fair amount of CSS that was copy-pasted between normal and fullscreen dialogs. I've moved that into the shared `dialog-styles.js` and introduced a CSS class to trigger it. Confirm dialog -- which doesn't opt into the special mobile styles -- is the reason we can't apply it everywhere.


components/dialog/dialog-fullscreen.js
```diff
@@ -226,9 +210,6 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 				: 0;
 			const startTop = mediaQueryList.matches ? 42 : 0;
 			topOverride = iframeTop + startTop;
-		} else if (window.innerWidth <= 615 || (window.innerWidth <= 900 && window.innerHeight <= 420)) {
-			heightOverride.height = `${window.innerHeight - 42 - 2}px`; // render full window height - 42px top padding - 2px border
```
dlockhart, line 230: This change here is going to result in a `2px` increase in height for all fullscreen dialogs. As is written in the comment, this takes the window height and subtracts `44px`, whereas the CSS solution is just the viewport height minus `42px`. I think this is an OK outcome to align things.

dbatiste (approved):

components/dialog/dialog-fullscreen.js
svanherk: Looks like this was recently added to fix [an issue on mobile](https://github.com/BrightspaceUI/core/pull/2200).  I don't know much about that change but just a heads up!

dlockhart: You're right, this does reintroduce that problem... so weird that normal dialogs don't exhibit it now that they're essentially the same. Back to the drawing board I guess!

components/dialog/dialog-mixin.js
```diff
@@ -1,4 +1,5 @@
 import '../focus-trap/focus-trap.js';
+import '../../helpers/viewport-size.js';
```
dlockhart, line 2: I noticed in fullscreen dialogs that `--d2l-vh` wasn't defined vs. in regular dialogs it was, which was what was causing #2200. All we need to do is include this in both places (i.e. move it to the mixin) and bingo!

components/dialog/dialog-fullscreen.js
dlockhart: Ok fixed! We had a viewport-size helper that defined `--d2l-vh` but it needed to actually be included on the page in the case of fullscreen dialogs to work. :)

Thanks for catching that Stacey.

components/dialog/test/dialog-fullscreen.visual-diff.js
```diff
@@ -85,7 +85,7 @@ describe('d2l-dialog-fullscreen', () => {
 					await page.$eval('#dialogLong #bottom', (bottom) => {
 						bottom.scrollIntoView();
 					});
-					await visualDiff.screenshotAndCompare(page, this.test.fullTitle());
+					await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false });
```
dlockhart, line 88: @svanherk is a genius -- thanks! I'll look into un-skipping the other tests with this approach too.

PR #2430, opened 2022-05-03
Make sure we're also fixing newlines in noDeferredRendering case

grant-cleary:
Occurs to me that I missed the `noDeferredRendering` case. Not sure we need to bother backporting this because it's already a bit of an edge case, but we should still fix it.

### DIFF
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -56,19 +56,24 @@ export class HtmlBlockMathRenderer {
 
 		await loadMathJax(mathJaxConfig);
 
+		// MathJax 3 does not support newlines, but it does persist styles, so add custom styles to mimic a linebreak
+		// This work-around should be removed when linebreaks are natively supported.
+		// MathJax issue: https://github.com/mathjax/MathJax/issues/2312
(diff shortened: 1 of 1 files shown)


helpers/mathjax.js
```diff
@@ -56,19 +56,24 @@ export class HtmlBlockMathRenderer {
...
+		// This work-around should be removed when linebreaks are natively supported.
+		// MathJax issue: https://github.com/mathjax/MathJax/issues/2312
+		// A duplicate that explains our exact issue: https://github.com/mathjax/MathJax/issues/2495
+		const lineBreakStyle = 'display: block; height: 0.5rem;'
+
 		// If we're opting out of deferred rendering, we need to rely
 		// on the global MathJax install for rendering.
 		if (options.noDeferredRendering) {
+			elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
+				elm.setAttribute('style', lineBreakStyle);
```
grant-cleary, line 69: So `mspace` is indeed an `HTMLElement`, but for some reason (in Chrome at least), `mspace.style` is undefined. I'm wondering if it doesn't fully implement the spec.

dbatiste (approved):

Issue #2150, opened 2022-02-28
Feature request: allow a multi-dimensional filter button to have custom text (`d2l-filter`)


Chebsi:
Currently with multi-dimensional filters, the filter button text is "Filters" by default, and this can't be changed. The request is to allow custom text for multi-dimensional filters.

My specific use case is in our Nova project, we have many different filters but we want to bring attention to the "Skills" filter, since it's more important. So "Skills" stands out and has its own single-dimension filter. 

Beside it, we have "Other filters", which is a multi-dimensional filter, containing a bunch of other filters you can apply to the collection. Right now it has to be called "Filters", but we want it to say "Other filters". This is a quick little mockup of what's desired (the "Other filters" text)

![image](https://user-images.githubusercontent.com/1237719/156056378-8cbff74b-2efa-421d-a0d9-d29ed1000f18.png)

geurts:
Thanks Moe, this seems reasonable to me! Are you able to wait until next week to hear back on this? Stacy will be back and we'll review this together and circle back with you.

Relatedly: Are you switching from individual filters to a multi-dimensional filter at a breakpoint? Asking because we have plans to implement overflow handling similar to what we did for d2l-overflow-group (which is for buttons). Looks like this will be tackled in early/mid April.

Chebsi:
> Are you able to wait until next week to hear back on this?

For sure, I can certainly wait to hear back after your review.

> Are you switching from individual filters to a multi-dimensional filter at a breakpoint?

Yes! In larger viewports we have each of the "Other filters" as their own single-dimension filter. Good to know there are already plans for this.

PR #2431, opened 2022-05-03
US123519 - Add ability to override the text used for filters with multiple dimensions

svanherk:
This addresses https://github.com/BrightspaceUI/core/issues/2150.  Went back and forth with Jeff on whether to provide a locked down option (like only "Filters" or "Other filters"), but we decided to just let the consumer choose whatever they needed based on their use case.


components/filter/filter.js
```diff
@@ -52,6 +52,11 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			 * @type {boolean}
 			 */
 			opened: { type: Boolean, reflect: true },
+			/**
+			 * Optional override for the button text used for a multi-dimensional filter
+			 * @type {string}
+			 */
+			text: { type: String },
```
dlockhart, line 59: I wish we'd consider removing these in favour of the Daylight Site, but until we make that call this needs to be duplicated here:
https://github.com/BrightspaceUI/core/tree/main/components/filter#properties

dlockhart (approved):

components/filter/filter.js
svanherk: Ah right. If anyone wants to take this to a vote, my vote would be "do not maintain this in the readme as well".  I don't see a public Daylight site as a blocker for that... the way we make breaking changes constantly already means we aren't ready to actually support this externally

dlockhart: I agree -- I'm in that camp as well.

PR #2446, opened 2022-05-06
fix: localize dynamic mixin: fix issue where nested mixin lang terms are not being used

margaree:
When `LocalizeDynamicMixin` is used multiple times in an inheritance chain, the top-most `localizeConfig` always takes precedence.

For example, if you have:

```javascript
class MyElement extends LocalizeDynamicMixin(LocalizeCoreElement(LitElement)) {
    static localizeConfig() { ... }
}
```

Then the `localizeConfig` provided by `MyElement` will be used everywhere, even inside `LocalizeCoreElement` -- which defines its own `localizeConfig`.

The fix: as `_getAllLocalizeResources()` is called up the inheritance chain, instead of just passing the original top-level `localizeConfig`, we instead check if a new one is defined at each level and if so use that instead.


mixins/localize-mixin.js
```diff
@@ -172,7 +172,9 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
 		}
 		// eslint-disable-next-line no-prototype-builtins
 		if (this.hasOwnProperty('getLocalizeResources') || this.hasOwnProperty('resources')) {
-			const res = this.getLocalizeResources([...possibleLanguages], config);
+			let superConfig = config;
+			if (superCtor.localizeConfig && superCtor.localizeConfig.importFunc) superConfig = superCtor.localizeConfig;
```
dlockhart, line 176: The issue I'm seeing with this is that if you reverse the order to be `class extends test2LocalizeDynamicMixn(test`LocalizeDynamicMixn((LitElement)))` then it starts failing.

margaree: Interesting - I don't think I'm seeing that locally but will dig some more shortly

dlockhart: The setup I have is slightly different -- I modified the existing `LocalizeTest` so that it does:

```
// this fails
class LocalizeTest extends LocalizeCoreElement(LocalizeDynamicMixin(LitElement)) {

// this is OK
class LocalizeTest extends LocalizeDynamicMixin(LocalizeCoreElement(LitElement)) {
```

margaree: So far my finding is that if whatever is using `LocalizeDynamicMixin` is in a separate mixin then it passes in both orders, but when it's in the same file (which is a totally possible usecase) then it doesn't pass when it's in the first order🤔

dlockhart: Ok, so after really digging into this and a lot of `console.log`-ing, I think your original solution was correct.

The order of the inheritance chain matters -- it needs to go `LocalizeTest` -> `LocalizeDynamicMixin` (so dynamic mixin can access localize-test's `localizeConfig) -> `LocalizeCoreElement`. The reverse order just isn't going to work, and I think that's OK.

bearfriend:
Is there anything that should be added to the README about proper order of inheritance?

PR #2450, opened 2022-05-09
Remove relative positioning; consumers should decide this

grant-cleary:
I believe we only want this specified for untrusted HTML. Otherwise if we were to try to render a profile card or something of the like inside an html-block and it didn't fit the bounds of the component, we wouldn't have a means to position it properly outside of fixed positioning shenanigans.


dbatiste (approved):
Makes sense.  I am surprised this is here.  There must have been to a reason to have it.

PR #2451, opened 2022-05-10
fix: US138545 make sure blur event gets to d2l-input-text-container div

mdulat:
https://rally1.rallydev.com/#/?detail=/userstory/632387188999&fdp=true

As part of updating Activity Feed (FRA) to Lit 2, I noticed that the `d2l-input-text` sometimes validated on blur, but most of the time it did not. It's because the `d2l-input-text-container` div usually doesn't get the blur event. This div handles the blur event to do the validation. Adding `tabindex=-1` to this div forces it to get the event.

I don't understand why without the tabindex it gets the event _sometimes_, however with the tabindex it seems to get it _always_.

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -402,7 +402,7 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 
 		const input = html`
 			<div class="d2l-input-container">
-				<div class="d2l-input-text-container d2l-skeletize" style="${styleMap(inputContainerStyles)}">
+				<div class="d2l-input-text-container d2l-skeletize" tabindex="-1" style="${styleMap(inputContainerStyles)}">
 					<input aria-atomic="${ifDefined(this.atomic)}"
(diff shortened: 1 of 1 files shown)


dbatiste:
Hey Margaret, what are the steps to reproduce the issue in the core demo page?  I would like to better understand what's happening here.  The `blur` event does not bubble, so maybe there is something happening there.  An alternative might be to use `focusout`, but I think I need to see what's going on.

mdulat:
> Hey Margaret, what are the steps to reproduce the issue in the core demo page? 

I can't actually reproduce this on the demo page, it only happens on my local Activity Feed that uses core's `d2l-input-text`, which hasn't been deployed yet. The currently deployed AF uses a styled `input` so this can't be seen on a quad site yet.

Also, today my fix isn't working locally either, so adding the `tabindex` wasn't actually a fix, it still only works occasionally/randomly. 

I think I can just call `requestValidate` when the user clicks Save instead of relying on `blur` doing the validation. But if you'd like to investigate, we can have a zoom so I can show you my local, or you can wait until I deploy this so you can see it on a quad site. For now I'll close this PR though.

PR #2455, opened 2022-05-10
Make sure autoloaded MathJax extensions are actually loaded before typesetting

grant-cleary:
The MathJax output processor we're using loads a bunch of TeX extensions automatically, but this loading is lazy and only happens when math using it is detected on the page. This means, however, that we also need to explicitly wait for these extensions to load before we can typeset anything.

Context: The symptom here was that the LaTeX `\color` command wasn't always working properly. A forced re-render of the math seemed to kick it into functioning properly, hence my realization that the extension wasn't always loading when first typesetting.

### DIFF
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -383,6 +383,7 @@
 									<mo>)</mo>
 								</mrow>
 							</math>
+							$$ {\color{red}x} + {\color{blue}y} = {\color{green}z} $$
 							<p>The wizard (<span data-replace-me-id="0">Elmer Fudd</span>) quickly jinxed the gnomes before they vaporized.</p>
 						</div>
 					</d2l-html-block>
@@ -495,6 +496,7 @@
 				<template>
 					<d2l-html-block>
 						<div>$$ f(x) = \int \mathrm{e}^{-x}\,\mathrm{d}x $$ $$ x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a} $$</div>
+						<div>$$ {\color{red}x} + {\color{blue}y} = {\color{green}z} $$</div>
 					</d2l-html-block>
 				</template>
 			</d2l-demo-snippet>
@@ -507,7 +509,7 @@
 						<div>
 							An equation rendered using LaTeX...
 							\( f(x) = \int \mathrm{e}^{-x}\,\mathrm{d}x \)
-							... and some text!
+							... and some text ...
 						</div>
 					</d2l-html-block>
 				</template>
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -69,7 +69,7 @@ export class HtmlBlockMathRenderer {
 				elm.setAttribute('style', lineBreakStyle);
 			});
 			await window.MathJax.startup.promise;
-			window.MathJax.typeset([elem]);
+			await window.MathJax.typesetPromise([elem]);
 			return elem;
 		}
 
@@ -82,7 +82,7 @@ export class HtmlBlockMathRenderer {
 
 		elem.appendChild(temp);
 		await window.MathJax.startup.promise;
-		window.MathJax.typesetShadow(temp.shadowRoot);
+		await window.MathJax.typesetShadow(temp.shadowRoot);
 
 		return temp.shadowRoot.firstChild;
 	}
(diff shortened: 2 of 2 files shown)


helpers/mathjax.js
```diff
@@ -69,7 +69,7 @@ export class HtmlBlockMathRenderer {
 				elm.setAttribute('style', lineBreakStyle);
 			});
 			await window.MathJax.startup.promise;
-			window.MathJax.typeset([elem]);
+			await window.MathJax.typesetPromise([elem]);
```
grant-cleary, line 72: `typesetPromise` basically wraps the usual MathJax rendering functions in the same `handleRetriesFor` promise we're manually using before.

helpers/mathjax.js
```diff
@@ -182,12 +182,15 @@ export function loadMathJax(mathJaxConfig) {
...
 				//  you need to rerender the shadowRoot later.
 				//
-				window.MathJax.typesetShadow = function(root) {
-					const InputJax = startup.getInputJax();
-					const OutputJax = startup.getOutputJax();
-					const html = mathjax.document(root, { InputJax, OutputJax });
-					html.render().typeset();
-					return html;
+				window.MathJax.typesetShadow = async function(root) {
+					return await mathjax.handleRetriesFor(() => {
```
grant-cleary, line 186: This is basically how MathJax handles missing extensions. It tries to render and if it hits a missing extension, it loads it and triggers a retry. Because we're not using MathJax's built-in adaptor, we have to do this ourselves.

grant-cleary:
There's a piece of this I haven't figured out yet... When non-deferred rendering and normal deferred rendering are mixed on the same page, whichever case loads last doesn't seem to render properly. I haven't yet managed to get to the bottom of why this happens, but I'm also not super concerned because the only place that happens in the LMS is on the homepage (custom widgets).

grant-cleary:
Another option would be to load all of the additional extensions all the time, though this would increase the download size of the output processor.

grant-cleary:
Ugh, it's somewhat worse than that. It looks like it's only loading properly in the first rendered HTML block. Going to dig in and see what needs to be reset to make that stop being a thing.

helpers/mathjax.js
```diff
@@ -186,6 +191,17 @@ export function loadMathJax(mathJaxConfig) {
 					const InputJax = startup.getInputJax();
 					const OutputJax = startup.getOutputJax();
 					const html = mathjax.document(root, { InputJax, OutputJax });
+
```
dbatiste, line 194: Any chance you can remind me what the difference is between these two calls?

grant-cleary: Yeah, so the second function is one I had to newly add to capture the input/output Jax settings properly. MathJax's `typesetPromise` API seems to use their default document for reasons. For non-deferred rendering cases we're not actually in a shadowRoot (or we can't really determine if we are), so instead we just pass in the elements we care about rendering and take the document itself as the parent.

grant-cleary: A perhaps more elegant solution may be to check if the element is in a shadowRoot, and if it is, pass that shadowRoot into the first element, add the `elem` to typeset as an optional argument, and then just combine it into one call.

dbatiste: Just to clarify, in what cases would we not be able to determine if we're in a shadow DOM?  Assuming the element is connected, `node.getRootNode()` could always be compared to the `document` element.  If it's different, it's a shadow DOM.

I am mildly concerned about the case where an `d2l-html-block` is inside the shadowDOM of an app component.

When we use `html.options.elements = [elem];`, is that effectively telling MathJax... typeset the elements in this array?

grant-cleary: Yes, the `no-deferred-rendering` case has never worked inside the shadowDOM of an app component. I believe we didn't care when first implementing this because we didn't actually want anybody to use it. It's not particularly difficult to make it work, it's just something we explicitly didn't care for supporting before.

Yes, `html.options.elements = [elem];` limits which elements we want MathJax to typeset.

helpers/mathjax.js
```diff
@@ -68,8 +68,9 @@ export class HtmlBlockMathRenderer {
 			elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
 				elm.setAttribute('style', lineBreakStyle);
 			});
+
 			await window.MathJax.startup.promise;
-			window.MathJax.typeset([elem]);
+			window.MathJax.typesetShadow(elem.getRootNode(), elem);
```
grant-cleary, line 73: Note that this works either way. The `ShadowAdaptor` we have can handle being passed both a shadowRoot and a `document`, so there's no need to do any kind of checking here. Whatever the root node is, we can just pass it along.

components/html-block/demo/html-block.html
```diff
@@ -507,7 +509,7 @@ <h2>HTML Block (LaTeX inline math)</h2>
 						<div>
 							An equation rendered using LaTeX...
 							\( f(x) = \int \mathrm{e}^{-x}\,\mathrm{d}x \)
-							... and some text!
+							... and some text ...
```
grant-cleary, line 512: Oops. Copy/paste error. I'm not going to bother changing it back though; hardly seems worth the CI run-time.

helpers/mathjax.js
dbatiste: Thanks, I remember that conversation now.  😄

grant-cleary: Yeah, given that we can't use MathJax's API anyways, I did just push a commit that added this support. It's almost more work not to do it at this point, haha.

dbatiste (approved):

PR #2462, opened 2022-05-13
Use safer replace API to support older browsers

grant-cleary:
Backport of this PR: https://github.com/BrightspaceUI/core/pull/2458


dlockhart (approved):

grant-cleary:
@dlockhart I wouldn't normally ask something like this, but the visual diff failures are just constantly alternating between focus timing issues for these dialogs... Would you be horribly opposed if I just merged this without the visual diff tests passing? I've lost count of the number of re-runs I've done and the focus just keeps alternating. 😞

dlockhart:
Yeah that's totally fine! A bunch of work has gone into the visual diffs in newer versions anyway.

svanherk:
If we think we'll need to backport more stuff to `core` `v1`, we could backport the flake fixes we've made to those tests?  Should be really quick and maybe cause less pain next time

grant-cleary:
> If we think we'll need to backport more stuff to `core` `v1`, we could backport the flake fixes we've made to those tests? Should be really quick and maybe cause less pain next time

I have a feeling I may need to do this, so yeah, that's a good thought and probably something I should have done from the get-go rather than re-running the tests for 2 hours 😅. In the interest of time I'll merge this one as is, but if/when I need to do this again, I can backport some of those stability fixes.

PR #2472, opened 2022-05-17
Update dropdown position logic to handle wrapping case

dbatiste:
This PR fixes the dropdown content positioning logic to handle the target wrapping case.  The problem would only arise when the dropdown was previously shown with the target not wrapping, and then subsequently shown with the target wrapping.

For example, before fix, not wrapping:
image

Then resize to cause wrapping, before fix, wrapping:
image

After fix, wrapping:
image

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -683,7 +683,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 			const centerDelta = contentRect.width - targetRect.width;
 			const position = this._getPosition(spaceAround, centerDelta);
-			if (position) {
+			if (position !== null) {
 				this._position = position;
 			}
 


components/dropdown/dropdown-content-mixin.js
```diff
@@ -683,7 +683,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 			const centerDelta = contentRect.width - targetRect.width;
 			const position = this._getPosition(spaceAround, centerDelta);
-			if (position) {
+			if (position !== null) {
```
dbatiste, line 686: The implicit check is the root cause.  When the target wraps, the new position (left offset of content from the target's left edge) value is `0` or `-0`, and so this position assignment gets skipped, leaving the old rendered position unchanged in the DOM.

dlockhart (approved):

PR #2478, opened 2022-05-17
Add focus-visible to link.

dbatiste:


components/link/link.js
```diff
@@ -42,7 +44,7 @@ export const linkStyles = css`
  * This component can be used just like the native anchor tag.
  * @slot - The content (e.g., text) that when selected causes navigation
  */
-class Link extends FocusMixin(LitElement) {
+class Link extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
```
dlockhart, line 47: So unfortunately this won't work for folks who just import the styles and apply them to a native `<a>` element -- a technique which I personally prefer over introducing unneeded custom elements. I guess that's just a known issue until we can ditch the polyfill.

Unless the polyfill is handling native elements in a magical way!

dlockhart (approved):

components/link/link.js
dbatiste: Yeah, the polyfill appears to do some magical global stuff.  It seems to work in my profile-card case where the polyfill happens to be loaded as a result of including other elements that use it, such as buttons. This means I should add an explicit import for it in that case, in case those elements are ever removed from the card.

PR #2481, opened 2022-05-17
Add d2l-dropdown-before-open event to DropdownContentMixin.

dbatiste:
This PR adds a new event `d2l-dropdown-before-open` to `DropdownContentMixin` that enables consumers to make updates before the dropdown calculates dimensions and position. In order for the consumer to hook in, they append a Promise to the event detail like so:

``` javascript
_onDropdownBeforeOpen(e) {
    this._someProp = 'new value';
    e.detail.ready.push(this.updateComplete);
}
```

Or, more generally:
``` javascript
_onDropdownBeforeOpen(e) {
    e.detail.ready.push(new Promise(resolve => {
        // do something
        ...
        resolve();
    }));
}
```

This seems to work pretty well in `d2l-labs-user-profile-card`, where we want to update the size of the card based on the dimensions of its bounding container.

We _could_ provide a different event API if we wanted.


components/dropdown/dropdown-content-mixin.js
```diff
@@ -575,11 +575,25 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 		if (newValue) {
 
-			if (ifrauBackdropService && this.mobileTray && this._useMobileStyling) {
-				this._ifrauContextInfo = await ifrauBackdropService.showBackdrop();
-			}
+			const evt = new CustomEvent('d2l-dropdown-before-open', {
```
dbatiste, line 578: @margaree : I'm not sure where this event should be documented, given the way it is dispatched.  Does it need to be documented on the class headers (-content, -menu, -tabs), or can it be documented inline here?  I see a mix for the other events.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -575,11 +575,25 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+				bubbles: true,
+				composed: true,
+				detail: { ready: [] }
```
dbatiste, line 581: Could name this `clientReady` instead.  Open to other ideas.

dbatiste:
I've been experimenting more with `d2l-labs-user-profile-card` and I think I have a simpler solution specific to it, which avoids this somewhat strange promise carrying stuff.  We might still want this, but I'm not ready to merge it.

components/dropdown/dropdown-content-mixin.js
margaree: Yup you're right - because of how it's dispatched, in a variable + separate `dispatchEvent`, it would need to be on the class headers for this mixin's consumers (e.g., https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-input.js#L15)

components/dropdown/dropdown-content-mixin.js
```diff
@@ -575,6 +575,15 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
 		if (newValue) {
 
+			const evt = new CustomEvent('d2l-dropdown-before-open', {
+				bubbles: true,
+				composed: true,
+				detail: { ready: [] }
+			});
+			this.dispatchEvent(evt);
+
+			if (evt.detail.ready.length > 0) await Promise.all(evt.detail.ready);
```
dlockhart, line 585: The `if` shouldn't be necessary:
> This returned promise will resolve when all of the input's promises have resolved, or if the input iterable contains no promises.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -575,6 +575,15 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 
 		if (newValue) {
 
+			const evt = new CustomEvent('d2l-dropdown-before-open', {
+				bubbles: true,
+				composed: true,
```
dlockhart, line 580: I'm guessing this needs to bubble and be composed because the event is originating from the dropdown content by the typical wire-up will be on dropdown?

dlockhart (approved):

PR #2485, opened 2022-05-18
Visibility Switch > Availability Conditions: remove existing switch tooltip functionality

dcostiuc:
Story: https://rally1.rallydev.com/#/15545167705d/dashboard?detail=%2Fuserstory%2F637810368677

We are removing the tooltip property/functionality for the switch component as we have decided that it is not really used and not worth offering as a feature to consumers.

Searched for `tooltip` and `tool tip` in `components/switch/*` and removed anything related to that.

Also removed the 2 switch tooltip `visual-diff` screenshots.

Note that we still need to go through the LMS and remove any code referencing the switch's tooltip property.
(There will be another PR made, related to this one, though neither is a blocker to each other for merging)

Edit: This PR is [here](https://github.com/Brightspace/lms/pull/23434) for [this story](https://rally1.rallydev.com/#/15545167705d/dashboard?detail=%2Fuserstory%2F638185363003)


dcostiuc:
Regarding the visual-diff failure, I removed a switch visual-diff tests from `switch.visual-diff.html` [here](https://github.com/BrightspaceUI/core/pull/2485/files#diff-366cfdd35e4d043b5c40626e5b3e60492427c121acb6429c70d95c4e881da7b2L48-L51), which may have caused slight shifting of the page and thus the 1-pixel difference. 

Otherwise I'm not sure why that 1 visual-diff test would fail, as I'm not updating the puppeteer version here, nor did I change any styling of the switch element. 

Going to merge the GH Actions bot PR in, but feel free to let me know if there are any concerns and I can always revert the commit.

components/switch/switch-mixin.js
```diff
@@ -192,7 +186,6 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 		const text = html`<span id="${this._textId}" class="d2l-switch-text">${this.text}</span>`;
 		const textPosition = (this.textPosition === 'start' || this.textPosition === 'hidden'
 			? this.textPosition : 'end');
-		const tooltip = (this.tooltip ? html`<d2l-tooltip for="${this._switchId}">${this.tooltip}</d2l-tooltip>` : '');
```
dbatiste, line 195: Note that `this._switchId` is no longer needed with this removed.  It is assigned below on the `div`, and initialized in the `constructor`.  I don't _think_ it will be needed for the other stories.  If you do, it's fine to keep it around since those changes are coming shortly, otherwise I would remove it.

dcostiuc: Yup, I agree. Thanks for pointing that out!

dbatiste (approved):

PR #2490, opened 2022-05-19
US138697 - Update localize event

svanherk:
This PR does a few things:
- Update the name of the localize event we fire to be more accurate.  I audited all outside usages and they're all using the Polymer behaviour, which does still fire the old event.
- Make that event no longer `bubble` or `composed`, because it doesn't need to be.
- Fire the event every time the resources change, not just after the first time.  This doesn't change how anything works here in `core` (because we always listen in `firstUpdated`, meaning we miss the first event and will only get language updates) but this is helpful for HM components using OSLO.  They may get their data faster then the langterms, and they need to know when to update any stored terms.  Ideally they don't store these terms, but we have some components that send these terms in an event, and they need to know when to resend the event (rather than resend every `render`).

Not tied to the event name, if anyone has a different preference.

### DIFF
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -422,7 +422,7 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 
 		this.addEventListener('blur', () => this._isInitialFocusDate = true);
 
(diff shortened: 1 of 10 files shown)


dlockhart (approved):
Good with the name 👍

joshhoey (approved):

PR #2497, opened 2022-05-24
Set display properly on inline/no-deferred-rendering html-block

grant-cleary:
The precedence of the `html-block`'s CSS rules meant that `display: none` wasn't being appropriately set on the render container when both `inline` and `no-deferred-rendering` were set. This technically doesn't make a visual difference, since the render container is rendered with 0 width and 0 height anyways in the `no-deferred-rendering` case. Still, I think it's better to set the CSS rules properly.


components/html-block/html-block.js
```diff
@@ -163,7 +163,7 @@ class HtmlBlock extends RtlMixin(LitElement) {
 				display: none;
 			}
 			:host([inline]),
-			:host([inline]) div.d2l-html-block-rendered {
+			:host([inline]:not([no-deferred-rendering])) div.d2l-html-block-rendered {
```
grant-cleary, line 166: Swapping this inline CSS block with the hidden case also solves this problem, but I personally like the more explicit rules.

grant-cleary: Hmm, but we probably also need another exception on the host, so perhaps it would indeed just be cleaner to rely on implicit ordering. Certainly open to other thoughts on that.

svanherk:
@grant-cleary Is this still a change we want in?  Not great timing to ask because Dave B's away this week anyways, but I was just reminded about it because we get a message saying we haven't reviewed it every day 😛

grant-cleary:
> @grant-cleary Is this still a change we want in? Not great timing to ask because Dave B's away this week anyways, but I was just reminded about it because we get a message saying we haven't reviewed it every day 😛

Oh, right... Haha ideally, yes.

components/html-block/html-block.js
dbatiste: I try order the selectors to avoid `:not(...)` since I find it _usually_ results in simpler CSS. I use it, but sparingly when necessary or when the alternative is difficult for the reader to understand. The `:not(...)` represents an extra selector check the browser needs to do. Whether the browser has optimized the matching to include negating selectors I am unsure.

grant-cleary: Yeah, I have it restructured to rely on the CSS order in my simplification work, so I'll align that here.

dbatiste (approved):

PR #2500, opened 2022-05-25
Update list-item grid focus logic for next item

dbatiste:
This PR updates the focus logic to better handle the case where the next/previous item does not contain focusable items in the target cell.  Previous to this change, focus would not move.  With this change, focus moves to the first focusable in the item if a corresponding focusable is not found for the given cell.

Solid line - where we try to place focus
Dashed line (**new**) - where we fallback to moving focus

**Example 1 - No Interactive content in cell:**
image

**Example 2 - Disabled selection:**
image


dlockhart (approved):
LG!

PR #2505, opened 2022-05-26
Fix instances of one category being used incorrectly

bearfriend:
`one` !== `1`. In French and Hindi, `one` is used for `0` as well. In other languages, it's used for [all kinds](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html) of different numbers. We can't hard code numbers in flexible categories. The only time that's appropriate (but still not necessary) is when using explicit numbers like `=0`, `=1`, `=462` etc.

Examples of incorrect renderings:

Because of the hardcoded `1`, these all render saying there is 1 error, instead of the actual number

Gaelic - 11 errors
Screen Shot 2022-05-26 at 12 46 05 PM

Slovenian - 101 errors
Screen Shot 2022-05-26 at 12 47 32 PM

Ukrainian - 21 erorrs
Screen Shot 2022-05-26 at 12 49 49 PM

Cebuano - 0 - 60 errors (gif)
cebuano-one


bearfriend:
@margaree This is what I meant about the `one` plural category.

margaree (approved):
Very interesting - thanks for the further explanation and fixing this!

PR #2518, opened 2022-06-01
fix: simplify calendar dropdown height calculation

dlockhart:
This functionality was original added in #1460 to attempt to calculate ahead of time how tall the tallest possible calendar (one with 6 rows) would be. That way the dropdown would open in the correct direction, even if the user changed the month and it got taller.

In reality, this was no longer doing anything -- I always see a height of `10px`, probably due to the timing of the calculation. And because it was `visibility: hidden`, that content was actually getting rendered and caused the viewport to grow when date inputs were placed next to the edge:

Screen Shot 2022-06-01 at 4 02 12 PM

I think it's simpler and easier to just use the built-in default of `415px`.


margaree (approved):

components/inputs/input-date.js
```diff
@@ -252,7 +237,7 @@ class InputDate extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixin(
 				@d2l-dropdown-open="${this._handleDropdownOpen}"
 				@d2l-dropdown-focus-enter="${this._handleFocusTrapEnter}"
 				max-width="335"
-				min-height="${this._hiddenCalendarHeight}"
+				min-height="415"
```
dbatiste, line 240: It looks (based on the comment above) that this assumes the root font-size is 20px; What happens when it's larger?  Is the experience any worse than before?

dbatiste: Was just thinking, if needed, we could get the computed root `font-size` from the `html` element, and select from hard-coded heights.  18, 20, 22, and 24px being our discrete sizes.

PR #2522, opened 2022-06-02
feat: filter-tags: only show clear all button

margaree:
This hides the Clear All button when there are less then 4 selected filters in `filter-tags`. ~It's possible we will want this functionality by default for `tag-list` which would be a quick switch if needed, but for now just adding it into `filter-tags`.~ I also added some `rtl` visual diffs to filter-tags


svanherk (approved):
Looks good to me!  Visual diff tests look good too, other than the dialog ones we're looking into

margaree:
FYI @svanherk after talking with Noah this has been updated to just hide clear all for all tag lists when < 4 tags. 

I'm not expecting the new visual diffs to change. The interactive one should but I'm going to add another tag to that one so that clear all does get displayed.

svanherk (approved):
Looks even better! 😄

PR #2524, opened 2022-06-02
fix: completely disable native dialogs due to Chrome issue

dlockhart:
Starting in Chrome 102, when a modal dialog is opened (via `showModal()` like we are) and that dialog is inside a shadow root (which is always true for us), any elements inside that dialog have a `null` value for `offsetParent`.

This is important because our `isFocusable` helper uses `offsetParent` to determine if an element is visible or not. So as a result of the change, focusing on the first elements inside dialogs broke.

I've raised this issue with Chromium:
https://bugs.chromium.org/p/chromium/issues/detail?id=1331803

I also noticed the same behaviour in Safari, although we disabled native dialogs in Safari a while back. Firefox has a similar behaviour but it returns the `<body>` instead of `null` so it technically "works".


components/dialog/dialog-mixin.js
```diff
@@ -17,7 +17,9 @@ window.D2L.DialogMixin = window.D2L.DialogMixin || {};
 
 // while implemented in Webkit, native <dialog> focus mangement across slotted content is buggy
 // https://bugs.webkit.org/show_bug.cgi?id=233320
-window.D2L.DialogMixin.hasNative = (window.HTMLDialogElement !== undefined)
+// starting in Chrome 102, all elements inside <dialog>s that are inside shadow roots have null offsetParent
+// https://bugs.chromium.org/p/chromium/issues/detail?id=1331803
+window.D2L.DialogMixin.hasNative = false && (window.HTMLDialogElement !== undefined)
```
dlockhart, line 22: As discussed in Slack, I've left all the testing infrastructure for native dialogs in place to make it easier for us to re-enable these later. I'm not sure if visual-diffs will occur though... we'll see.

dbatiste (approved):

dbatiste:
I had to update the "input-date-time functionality change date" test because the tests shared the same fixture and the "change-time" test was skipped due to this craziness: https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/input-date-time/2022-06-03-15.40.12.967-02465e23ae2a7cc0e30301745d95d69cbca5d011-report.html

PR #2526, opened 2022-06-02
Tooltip Help Component: add new text like button help opener

dcostiuc:
This PR introduces a new reusable `d2l-tooltip-help` component that renders as a `"button"` linked to a `tooltip`, but looks like `text` with a formatted underline.

~~Note that **styling details** (specific values of properties) will change and will be updated in a future PR (or as part of this PR, if we want to keep it open until then).~~
Styling details have been finalized.

Note that the new changes to the `Visibility Switch` component [add link to PR once ready] depend on this component.

**For testing, various combinations of the following were experimented with:**
* The latest version of different browsers (Chrome, Firefox, Edge, Safari)
* Various screen reader tools on different browsers (NVDA, JAWS, VoiceOver, Voice Assistant), and OSs (Windows, Mac, Android)
* Various edge/test cases for the opener text length, the tooltip content length and type (regular text vs formatted HTML with `ul`/`ol` elements), various kinds of languages (English, Russian, Chinese), RTL support, keyboard vs mouse interaction (focus, hover and click states)
* stand-alone Help Tooltip, Help Tooltip in a sentence, Help Tooltip in a paragraph
* Various levels of zoom and viewport size

**Existing/Known Issues:**
* On Safari, the `border-radius` styling is not rendered for the focus outline


dcostiuc:
Not sure if we should merge this directly into `main`, or just to [feat/visibility-switch/indication-of-availability-conditions](https://github.com/BrightspaceUI/core/tree/feat/visibility-switch/indication-of-availability-conditions).

Technically the changes here are standalone, and if merged into main, we could just merge it back into the visibility switch feature branch.

But, at the same time, the styling details aren't final and we're planning to finalize them with Jeff in a live pairing session next week.

dcostiuc:
For reference, this is more or less how the Daylight page for the tooltip should look like with the changes to the tooltip component README (well, before any additional changes from reviewing this PR are made).

The Best Practices here were based on what Jeff wrote in the [Design Spec](https://www.figma.com/proto/A8jk5soJf1bz5z560Uld32/Help-Tooltips?page-id=0%3A1&node-id=2%3A100&viewport=469%2C216%2C0.32&scaling=min-zoom&starting-point-node-id=2%3A98)

image

components/switch/README.md
```diff
@@ -49,7 +49,6 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 | `disabled` | Boolean | Disables the switch |
 | `on` | Boolean | Whether the switch is "on" or "off" |
 | `text-position` | String | Valid values are: `start`, `end` (default), and `hidden` |
-| `tooltip` | String | Text to display in a tooltip for the switch |
```
dcostiuc, line 52: Note that the changes in this file and several others below where there is just stuff being deleted are actually from [this PR](https://github.com/BrightspaceUI/core/pull/2485), and it got added here probably because I switched between the `feature branch` and `main` as the base branch of the PR.

components/tooltip/README.md
```diff
@@ -151,3 +151,44 @@ In the following example to constrain the tooltip to the dashed boundary we can
...
+* The contents of the tooltip should elaborate on the label
+* Keep help text short and concise (full sentences are not necessary)
+* Use a help tooltip when there are space limitations, such as in a table, list, or narrow sidebar
+<!-- docs: end dos -->
+
+<!-- docs: start donts -->
+* Don’t overuse help tooltips (users end up hunting for information and even experts feel obligated to check their contents)
+* Don't use help tooltips when you are able to use inline help text instead
+<!-- docs: end donts -->
+<!-- docs: end best practices -->
```
dcostiuc, line 194: Does it make sense to have another Best Practices section for the `Help Tooltip` when we already have one at the top for tooltips in general? (are we strict with having 1 "Best Practices" per component page?)

Do we always need to start off the Dos with `"Use..."` and Don'ts with `"Don't"`? (not currently being done above)

In addition to any other changes to this PR, I'm especially open to suggestions for revising the dos and donts content as well (perhaps removing what's in the brackets, or perhaps rewording some things)

Let me know how good or bad it currently sounds! 🙂

Thank you @geurts for providing [a lot of the content](https://www.figma.com/proto/A8jk5soJf1bz5z560Uld32/Help-Tooltips?page-id=0%3A1&node-id=2%3A100&viewport=469%2C216%2C0.32&scaling=min-zoom&starting-point-node-id=2%3A98) here.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,92 @@
+import '../colors/colors.js';
+import '../tooltip/tooltip.js';
+import { css, html, LitElement } from 'lit';
+import { bodySmallStyles } from '../typography/styles.js';
+import { FocusMixin } from '../../mixins/focus-mixin.js';
+import { FocusVisiblePolyfillMixin } from '../../mixins/focus-visible-polyfill-mixin.js';
+
+/**
+ * A component used to display additional information when users focus or hover on certain text.
```
dcostiuc, line 9: Perhaps I'm being too picky here, but do we need this to match 100% to what we say in the [readme/Daylight page](https://github.com/BrightspaceUI/core/pull/2526/files#diff-1c38718d67e5f3d531c46a45d0a1ebae23b035384623ebdc2eeca8f8e87a99b1R157)?

components/tooltip/README.md
```diff
@@ -151,3 +151,44 @@ In the following example to constrain the tooltip to the dashed boundary we can
...
+
+<!-- docs: start hidden content -->
+### Properties
+
+| Property | Type | Description |
+|--|--|--|
+| `text` | String, required | Provide the `text` that will render as the Help Tooltip opener.|
+
+### Events
+* `d2l-tooltip-show`: dispatched when the opener text is hovered over, focused on, or clicked
```
margaree, line 176: Does this component actually dispatch these events or are those more the `d2l-tooltip`'s events?

margaree (commented):
Since this is just adding a new component and not really impacting anything existing (other than the tooltip change if we want to do that at the tooltip level) this would be safe to just branch off of main and also reduce the complication of the issues around having the switch changes in there

components/tooltip/README.md
dlockhart: I wouldn't expect this component to dispatch any events unless a use case comes up where they're needed.

components/tooltip/test/tooltip-help.axe.js
```diff
@@ -0,0 +1,34 @@
...
+
+		await expect(helpTooltip).to.be.accessible();
+	});
+
+	it('should pass all aXe tests for state info (show)', async() => {
+		const helpTooltip = await fixture(tooltipFixture);
+		const actualTooltip = helpTooltip.shadowRoot.querySelector('d2l-tooltip');
+
+		actualTooltip.setAttribute('state', 'info');
+		actualTooltip.setAttribute('showing', 'showing');
```
dlockhart, line 20: You may need an `await actualTooltip.updateComplete;` here... otherwise I'm not sure it's actually testing what you think it is.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,92 @@
...
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		if (!this.text || this.text.length === 0) {
+			console.warn('Help Tooltip component requires text.');
+		}
+	}
+
+	render() {
+		return html`
+			<span
```
dlockhart, line 66: Could we just use a `<button>` here instead and style it to look how we want? Then we wouldn't need to listen for key down or put a `tabindex` on it.

components/tooltip/test/tooltip-help.visual-diff.js
```diff
@@ -0,0 +1,98 @@
...
+		{ case: 'help tooltip hovered and focused', hover: true, focus: true },
+		{ case: 'help tooltip hovered', hover: true },
+		{ case: 'help tooltip focused', focus: true },
+		{ case: 'help tooltip clicked', click: true }
+	].forEach((testCase) => {
+
+		it(testCase.case, async function() {
+			const openEvent = page.$eval(tooltipHelpSelector, async(elem) => {
+				return new Promise((resolve) => {
+					const tooltip = elem.shadowRoot.querySelector('d2l-tooltip');
```
dbatiste, line 37: It looks like the `d2l-tooltip-show` event is composed, so it should be possible to avoid querying the internal shadowDOM to locate the tooltip.  Ideally tests should treat the `d2l-tooltip-help` as a "black box" from a testing perspective.  So I _think_ it should be possible to just have `elem.addEventListener('d2l-tooltip-show', resolve)`;

components/tooltip/test/tooltip-help.visual-diff.js
```diff
@@ -0,0 +1,98 @@
...
+	[
+		'help-tooltip-in-sentence',
+		'help-tooltip-in-paragraph',
+	].forEach((testName) => {
+
+		it(testName, async function() {
+			const selector = `#${testName}`;
+			const openEvent = page.$eval(`${selector} d2l-tooltip-help`, async(elem) => {
+				return new Promise((resolve) => {
+					const tooltip = elem.shadowRoot.querySelector('d2l-tooltip');
```
dbatiste, line 85: Same thing here... I think we should be able to avoid querying the internals.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,92 @@
...
+				${this.text}
+			</span>
+			<d2l-tooltip class="help-tooltip" for="d2l-tooltip-help-text" delay=0 offset=13>
+				<slot></slot>
+			</d2l-tooltip>
+		`;
+	}
+
+	_handleKeyDown(event) {
+		if (event.defaultPrevented) {
```
dbatiste, line 81: This check doesn't seem to be necessary.

components/tooltip/tooltip.js
```diff
@@ -340,6 +340,17 @@ class Tooltip extends RtlMixin(LitElement) {
 				animation: d2l-tooltip-right-animation 200ms ease;
 			}
 
+			::slotted(ul),
```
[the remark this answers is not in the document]
dcostiuc: @dbatiste @geurts Could you confirm that we do want this change for the tooltip in general? 

That is a good point that perhaps somewhere in D2L, some team has existing cases where they use `lists within a tooltip`, and maybe they have vdiff tests for that, at which point after we get this change into core, those tests will start to fail (though I suppose at that point, any such team would eventually reach out to us).

I think we do want to make this change a general one though. The idea is that, without that styling, the bullet points in a tooltip [are awkwardly and unnecessarily indented](https://d2l.slack.com/archives/C03FC0GDVB7/p1652987800035169?thread_ts=1652985508.177829&cid=C03FC0GDVB7), and we probably only happened to notice this due an example of the help tooltip/visibility switch in the design spec.

components/tooltip/tooltip-help.js
dcostiuc: Yeah, we could. I think this is something we kept going back and forth on with @dbatiste and @geurts.

There have been several comparisons done for things like how the screen reader would read the component depending on whether it's a `span` or a `button`.

We last settled on [going with a `span`](https://d2l.slack.com/archives/C03H8KH2FKN/p1653919910721479?thread_ts=1653676341.277049&cid=C03H8KH2FKN), @geurts do you remember why we preferred that over just using a `button` tag?

components/tooltip/test/tooltip-help.visual-diff.html
```diff
@@ -0,0 +1,74 @@
...
+		}
+		.visual-diff > d2l-tooltip-help {
+			display: block;
+			margin-bottom: 10rem;
+			padding: 20px;
+		}
+	</style>
+</head>
+<body class="d2l-typography">
+	<div class="visual-diff">
```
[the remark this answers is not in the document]
dcostiuc: Ah ok, I was initially following the `tooltip` visual diff as an exemplar, but looks like that's another case where just wrapping all the test cases in one `visual-diff div` is the exception, rather than the convention.

components/tooltip/demo/tooltip.html
```diff
@@ -150,6 +151,13 @@ <h2>Tooltip (only show if truncating)</h2>
 			</template>
 		</d2l-demo-snippet>
 
+		<h2>Help Tooltip</h2>
+		<d2l-demo-snippet>
+			<template>
+				<d2l-tooltip-help text="Helpful label">Contents should elaborate on the label (be short and concise)</d2l-tooltip-help>
```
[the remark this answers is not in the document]
dcostiuc: Noted

components/tooltip/test/tooltip-help.visual-diff.js
dcostiuc: Yup, you're right, that works and makes sense that it would work. Previously I had that but my tests were failing so I think as part of my debugging and experimenting, I ended up just directly setting it and left it at that once I had those tests passing.

dcostiuc: Brought it back though, since I agree that it feels weird to dig into the internals of the component that we're testing like that.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,92 @@
...
+		super.firstUpdated(changedProperties);
+		if (!this.text || this.text.length === 0) {
+			console.warn('Help Tooltip component requires text.');
+		}
+	}
+
+	render() {
+		return html`
+			<span 
+				id="d2l-tooltip-help-text"
```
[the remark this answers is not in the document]
dcostiuc: Is there a reason to use `getUniqueId` instead of specifying an id?

What do you mean by `space`?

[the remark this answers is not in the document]
margaree: So in general you wouldn't want duplicate ids on a page since a selector for that id would just grab the first occurrence. In this case, functionally I believe it would work either way since it would be getting used in the shadow root. To me personally it still seems like good practice to use unique ids when using ids (e.g., someone could copy/paste this code but actually have potential duplicate id issues), but others may differ on this.

By space I meant `span`🤦‍♀️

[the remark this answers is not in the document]
dlockhart: I wouldn't worry about using `getUniqueId` here -- we used to avoid them when we needed to handle browsers that didn't support shadow DOM, but we're thankfully past that now.

components/tooltip/tooltip-help.js
dlockhart: Yeah I'd be curious about the reasoning. In theory, `<span role="button">` and `<button>` should behave exactly the same, which is why the general advice is to just always use the correct native element for the job instead of adapting another one using ARIA. Plus `<button>` gives you built-in goodness like handling click/keyboard and tab stops.

components/tooltip/tooltip-help.js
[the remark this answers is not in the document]
dcostiuc: I'll keep the Id as is then

components/tooltip/tooltip.js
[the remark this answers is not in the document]
dcostiuc: Got [confirmation from Jeff](https://d2l.slack.com/archives/C03H8KH2FKN/p1654612810042419?thread_ts=1654611884.788179&cid=C03H8KH2FKN) that "reduced bullet/list indentation is desirable in any tooltip :+1:"

components/tooltip/tooltip-help.js
dcostiuc: After [discussing with Jeff and Dave B](https://d2l.slack.com/archives/C03H8KH2FKN/p1654633432212059?thread_ts=1654613644.917059&cid=C03H8KH2FKN), we agree that using a `<button>` works better here.

components/tooltip/README.md
dcostiuc: I'll remove any documentation for events for this component, since we don't want consumers to rely on that fact, and as mentioned, there shouldn't currently be a use case for those events.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,83 @@
+import '../colors/colors.js';
+import '../tooltip/tooltip.js';
+import { css, html, LitElement } from 'lit';
+import { bodySmallStyles } from '../typography/styles.js';
```
dbatiste, line 4: I think this can be removed now since it's no longer being used.  Can also be removed from the static styles assignment.

components/tooltip/demo/tooltip.html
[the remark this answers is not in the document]
dcostiuc: Help-tooltip-in-sentence example has been added to the demo, though there is currently a visible discrepancy between `d2l-body-small`-styled text and the `opener text` styling.

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,87 @@
...
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		if (!this.text || this.text.length === 0) {
+			console.warn('Help Tooltip component requires text.');
+		}
+	}
+
+	render() {
+		return html`
+			<button id="d2l-tooltip-help-text" class="d2l-body-small">
```
margaree, line 77: This class can also be removed since you're no longer importing `bodySmallStyles`

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,87 @@
...
+				${this.text}
+			</button>
+			<d2l-tooltip class="help-tooltip" for="d2l-tooltip-help-text" delay=0 offset=13>
```
margaree, line 80:
```suggestion
			<d2l-tooltip for="d2l-tooltip-help-text" delay=0 offset=13>
```
this class doesn't seem to be used

components/tooltip/tooltip-help.js
```diff
@@ -0,0 +1,87 @@
...
+			#d2l-tooltip-help-text {
+				background: none;
+				border: none;
+				color: inherit;
+				font-family: inherit;
+				font-size: inherit;
+				font-weight: inherit;
+				line-height: inherit;
+				padding: 0;
+				text-decoration-color: inherit;
```
margaree, line 46: super nit: based on https://www.w3schools.com/cssref/pr_text_text-decoration.asp it looks like these `text-decoration` properties can all be combined into just `text-decoration` rather than be split over multiple properties.

dbatiste: IIRC, I think I discovered that for some strange reason, Safari didn't like the short-hand version. I found that it worked by specifying the properties individually.  I don't mind double checking that though.

margaree: Ah interesting - thanks for looking into that

dcostiuc: Yep initially we used the shorthand but after Dave testing the component out on safari, we that that property was just ignored and resorted to styling it all out separately as a work around.

dbatiste (approved):
Looks great!

geurts (dismissed):
Here are some language changes/suggestions.

components/tooltip/README.md
```diff
@@ -151,3 +151,53 @@ In the following example to constrain the tooltip to the dashed boundary we can
...
 ```
+
+## Help Tooltip [d2l-tooltip-help]
+
+The `d2l-tooltip-help` component is used to display additional information when users focus or hover over some text.
+
+### Best Practices
+
+<!-- docs: start dos -->
```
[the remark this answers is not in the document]
dcostiuc: @margaree In terms of our conventions for the documentation site, is it fine if we have a paragraph for the Best Practices before the Dos and Donts, or should the information above be converted to more Dos and Donts points?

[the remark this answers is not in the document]
margaree: Usually it's converted to Dos and Donts bullet points

PR #2532, opened 2022-06-06
fix: check for null parent

dlockhart:
I noticed this in a test fixture in another repo where it opened a dialog but then didn't close the dialog before moving on to the next fixture. The backdrop would then try to hide itself but it had already been detached from the DOM, so `parent` became `null`.

### DIFF
--- a/components/backdrop/backdrop.js
+++ b/components/backdrop/backdrop.js
@@ -176,8 +176,8 @@ function hideAccessible(target) {
 	};
 
 	let parent = getComposedParent(target);
-	while (parent !== document.documentElement) {
-		if (parent.nodeType === Node.ELEMENT_NODE && parent !== null) {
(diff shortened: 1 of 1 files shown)


components/backdrop/backdrop.js
```diff
@@ -177,7 +177,7 @@ function hideAccessible(target) {
 
 	let parent = getComposedParent(target);
 	while (parent !== document.documentElement) {
-		if (parent.nodeType === Node.ELEMENT_NODE) {
+		if (parent.nodeType === Node.ELEMENT_NODE && parent !== null) {
```
svanherk, line 180:
```suggestion
		if (parent !== null && parent.nodeType === Node.ELEMENT_NODE) {
```
Should this be switched?

svanherk (approved):
I think I've seen this before in other tests too, maybe my courses

PR #2536, opened 2022-06-09
automatically do a minor release each time the LMS releases

dlockhart:
Switching core to use the new `MINOR_RELEASE_WITH_LMS` functionality.


dlockhart:
TODO: I need to make the `RALLY_API_KEY` secret available to this org

svanherk:
This is technically an issue with every repo that uses `semantic-release`, right?  And moving to CodeArtifact and NPM removes a previous workaround where you could point to a specific tag using GitHub refs directly...
I'm wondering if this should just be added to the semantic-release action, and you "opt-in" by providing a location to your `.lmsrelease` file 🤔

.github/workflows/release.yml
```diff
@@ -27,6 +27,26 @@ jobs:
...
+      - name: Update Repo's Last LMS Release
+        if: steps.repo-lms-version.outputs.VALUE != steps.get-lms-version.outputs.LMS_VERSION
+        run: |
+          echo "Updating last LMS release..."
+          echo "${{ steps.get-lms-version.outputs.LMS_VERSION }}" > .lmsrelease
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+          git add .lmsrelease
+          git commit -m "feat: updating tracked LMS release version [skip ci]"
+          git push --follow-tags --repo=https://${{ inputs.GITHUB_TOKEN }}@github.com/${{ github.repository }}
```
svanherk, line 47: Should this be `${{ secrets.D2L_GITHUB_TOKEN }}` or `$GITHUB_TOKEN`?

dlockhart: Yeah probably one or the other -- they should both be equivalent in this context. Not sure how this worked in my version of this that I tested with.

svanherk (approved):
So in theory, you could try releasing this with no `fix:` or `feat:` and it should do a minor bump...

PR #2537, opened 2022-06-09
DE48739 - Select-all button no longer checks off disabled selection-items

oelhanafey:
https://rally1.rallydev.com/#/?detail=/defect/638775912441&fdp=true

Changed the Selection-Mixin, selecting all items in the list will only check off elements that are enabled.

Would it be worth adding a disabled element to the select-all example on the Daylight site?


dbatiste (approved):
Looks good!

Re your question about adding an example of this scenario to the Daylight site - I might run it by Jeff. We have the tests for the behaviour, and I don't think it's necessary to call attention to this scenario in the docs. If he does want it, I would advocate for just adding a disabled item to the "Select All" example. I wouldn't create a separate example just for this.

PR #2544, opened 2022-06-15
feat: Move to node 16

svanherk:
Many of our repos (and many more to come shortly) are moving to node 16.  We should too so we don't need to keep switching back and forth


dlockhart (approved):
Do we want to wait until BSI and the LMS can go to Node 16, so that devs don't need to switch back and forth?

dlockhart:
It's at least 14, and the lock file is the new format so yeah I guess it's not blocking anything. Just up to Gaudi I guess around whether we want to wait for BSI. I'm good with doing this now.

PR #2546, opened 2022-06-16
Selection component bug fixes

oelhanafey:
https://rally1.rallydev.com/#/?detail=/defect/638775260887&fdp=true
https://rally1.rallydev.com/#/?detail=/defect/638775489995&fdp=true

Fixed a few defects with the selection component:
- Disabled single selection (radio) elements can no longer be clicked on
- Disabled single selection elements no longer have a hover state. Since the radio elements are defined using divs instead of inputs and divs do not have a disabled property, a class needed to be used instead.
- Navigating single selection elements with arrow keys no longer selects disabled elements

There were no tests written for SelectionMixin's `_handleRadioKeyUp` arrow key navigation so I was attempting to write some but I had no luck, specifically because of this line:
```
if (!e.composedPath()[0].classList.contains('d2l-selection-input-radio')) return;
```
I was unable to write a test or construct an event with a defined composed path. If anybody has any ideas here, please let me know.


components/inputs/input-radio-styles.js
```diff
@@ -26,7 +26,7 @@ export const radioStyles = css`
 		background-image: url("data:image/svg+xml,%3Csvg%20width%3D%2210%22%20height%3D%2210%22%20viewBox%3D%220%200%2010%2010%22%20xmlns%3D%22http%3A//www.w3.org/2000/svg%22%3E%0A%09%3Ccircle%20cx%3D%225%22%20cy%3D%225%22%20r%3D%225%22%20fill%3D%22%23494c4e%22%3E%3C/circle%3E%0A%3C/svg%3E");
 	}
 	.d2l-input-radio,
-	.d2l-input-radio:hover:disabled,
```
dbatiste, line 29: Are you sure that you want to remove the original selector? I think we would still expect `d2l-input-radio` to be applied as `<input type="radio" class="d2l-input-radio">` in almost all usages. The `d2l-selection-input` is sort of an exceptional case.

dbatiste:
I think the changes here are valid, but I also think that disabled selection-inputs that are part of a group are a little weird. Take for example the following example:

``` html
<d2l-demo-selection selection-single>
	<ul>
		<li><d2l-selection-input key="geo" label="Geography"></d2l-selection-input>Geography</li>
		<li><d2l-selection-input selected disabled key="sci" label="Science"></d2l-selection-input>Science</li>
		<li><d2l-selection-input key="mth" label="Math"></d2l-selection-input>Math</li>
	</ul>
</d2l-demo-selection>
```
The second item is disabled but selected. By selecting a different item in the group, the disabled item becomes unselected, however it can never be re-selected. I think it is for this sorta reason that the disabled cases where not fully flushed out when this was originally introduced.

components/inputs/input-radio-styles.js
oelhanafey: That is a good point. The new selector should be added in addition to the original one, not replace it.

dbatiste (approved):

PR #2548, opened 2022-06-17
Add quote icon

devpow112:
This icon is used in a few places I can think off.
https://github.com/Brightspace/d2l-outcomes-overall-achievement/blob/main/src/custom-icons/quote.js#L24-L34
https://github.com/Brightspace/d2l-rubric/blob/main/src/components/icons/rubric-quotation-icon.js#L19-L29
https://github.com/Brightspace/d2l-activity-details-ui/blob/main/src/resources/quotation-mark-icon.svg
Would be nice to just add it to core so I can use it instead of having to include the SVG in multiple locations.


dbatiste:
Curious about the context - is there ever an ending quote?

devpow112:
> Curious about the context - is there ever an ending quote?

Not in the context's I've seen. It's always just the start icon. Make a better name would be `quote-start`?

devpow112:
Looks like in all the cases I noted it's used when feedback is displayed. This is the outcomes overall stuff
image

dbatiste:
> > Curious about the context - is there ever an ending quote?
> 
> @dbatiste not in the context's I've seen. It's always just the start quote. Maybe a better name would be `quote-start`?

Yeah I think I would. An alternate would be `quote-open`. Seems odd that there wouldn't be a closing quote in the example you attached. Technically the consumer might be able to to rotate it 180deg if they wanted a closing quote, but I think it would be better to not leave that on the consumer to figure out.

dbatiste:
Also the fill color is different.  Are you planning to override it, or assume the new fill color?

devpow112:
> Also the fill color is different. Are you planning to override it, or assume the new fill color?

@dbatiste probably override, I'll have to see how the new one looks. It's probably better to have the darker color for accessibility but I'll have to check with PM and design.

dlockhart:
This is a little bit different than our normal icons... since as Dave pointed out the fill colour is different, and we would likely never see it used in our "typical" icon use cases like inside a `<d2l-button-icon>` or menu opener.

Wondering if maybe this could be a one-off and use the `<d2l-icon-custom>` until we see a few more examples of it? And even then, maybe we could put it into some shared `blockquoteStyles` or something as a background image...

devpow112:
> This is a little bit different than our normal icons... since as Dave pointed out the fill colour is different, and we would likely never see it used in our "typical" icon use cases like inside a `<d2l-button-icon>` or menu opener.
> 
> Wondering if maybe this could be a one-off and use the `<d2l-icon-custom>` until we see a few more examples of it? And even then, maybe we could put it into some shared `blockquoteStyles` or something as a background image...

@dlockhart yeah right now we are using the `d2l-custom-icon` stuff just wanted to eliminate having the same SVG data in multiple repos.

dlockhart:
Gotcha. Yeah I don't feel strongly about it, it's just a bit weird as an icon but I can live with it!

devpow112:
> Gotcha. Yeah I don't feel strongly about it, it's just a bit weird as an icon but I can live with it!

Let me see if I can add a block quite styling for it. Would typograph be a good place for that to live?

dbatiste:
That kinda makes me wonder if the [blockquote](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) element should be used. I'm not sure it's ever been used in the LMS, but it's the most semantically correct element for a block quotation.

devpow112:
> 



> That kinda makes me wonder if the [blockquote](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) element should be used. I'm not sure it's ever been used in the LMS, but it's the most semantically correct element for a block quotation.

Hmmm, that might be a good idea.

dlockhart:
Yeah like I was imagining some shared styles in typography where you could import them into your component and go:

```html
<blockquote class="d2l-blockquote">stuff</blockquote>
```

dlockhart:
The big advantage there is the colour and more importantly the positioning of the quote vs. the quoted text could be consistent -- plus we'd be using the correct semantic element for this purpose.

devpow112:
Ok, I'll take a look and see what I can get working. I'll close this off for now as I'm good with the idea of the block quote stuff.

PR #2551, opened 2022-06-17
feat: add blockquote styling

devpow112:
This adds styling for use with `<blockquote>`. I also added the typography demos to the demo page index.

test

Currently For RTL I'm not flipping/changing the quote image just relocating it. This is the current behaviour I observed in the places I know this is used. Previous PR discussing this work https://github.com/BrightspaceUI/core/pull/2548.


components/typography/test/typography.visual-diff.js
```diff
@@ -61,6 +61,11 @@ describe('d2l-typography', function() {
 			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 		});
 
+		it('blockquote', async function() {
```
devpow112, line 64: I should probably add an RTL for this.

devpow112:
Is there a reason all our typography vdiffs are `max-width: 320px`. WE have narrow and wide diffs but they are the same.

components/typography/demo/typography.html
```diff
@@ -6,6 +6,7 @@
 		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
 		<script type="module">
 			import '../../demo/demo-page.js';
+			import '../typography.js';
```
dbatiste, line 9: Did you run into a problem that required this import.  It should already be imported by demo-page [here](https://github.com/BrightspaceUI/core/blob/main/components/demo/demo-page.js#L4).

components/typography/styles.js
```diff
@@ -1,3 +1,4 @@
+import '../colors/colors.js';
```
dbatiste, line 1: Looks like this was a missing import eh? `d2l-body-small` requires it as well.

components/typography/test/typography.visual-diff.js
dbatiste: Yeah I think it would be worthwhile. In this case I think the fixture can just be set up with `<blockquote dir="rtl" ...>`.

components/typography/typography.js
```diff
@@ -141,6 +141,28 @@ if (!document.head.querySelector('#d2l-typography-font-face')) {
...
+			font-size: 0.8rem;
+			font-weight: 400;
+			line-height: 1.2rem;
+			margin-left: 0;
+			margin-right: 1.2rem;
+			padding-left: 1.2rem;
+			padding-top: 0.2rem;
+		}
+
+		.d2l-typography .d2l-blockquote[dir="rtl"] {
```
dbatiste, line 159: These styles will be in the global scope so the `blockquote` element wouldn't have a `dir="rtl"` attribute on it. So I think this needs to be `[dir="rtl"] .d2l-typography .d2l-blockquote`, given that we render `dir` on the `html` element.


Aside, in the web component case we sprout `dir="rtl"` on the host using the `RtlMixin`, and we do that because selectors inside of shadow roots can't be configured to select based on attributes in the parent DOM scope(s).

dbatiste:
> Is there a reason all our typography vdiffs are `max-width: 320px`. There are narrow and wide diffs but they are the same.

I think the `320px` is being used to limit the size of the fixture so that we don't have huge screenshots. The bigger the fixture is has to compare, the slower it is. Also, I vaguely recall setting that value (or it might have been someone else 🤔 ) to also force wrapping of text in a couple of the tests to better test line-height.

The `typography.js` has a media-query baked in it, which reduces the font-size slightly when the view-port is smaller. The wide & narrow visual-diff tests are configured to adjust the view-port size to capture this behaviour.

components/typography/typography.js
dbatiste: The reason why the current selector works in the demo-snippet is because the demo-snippet has some code to set `dir="rtl"` directly on the element to simulate what `RtlMixin` does. While `[dir="rtl"] .d2l-typography .d2l-blockquote` is what we need for our app, it won't work for the demo-snippet RTL switch. To make it work in both scenarios we'd need:

``` css
[dir="rtl"] .d2l-typography .d2l-blockquote,
.d2l-typography .d2l-blockquote[dir="rtl"] {
    ...
}
```

components/typography/demo/typography.html
devpow112: Nope added when I was trying to figure something out and forgot about it. Will remove

components/typography/typography.js
devpow112: Ok will adjust

devpow112:
I know I stated

> Currently For RTL I'm not flipping/changing the quote image just relocating it. This is the current behaviour I observed in the places I know this is used. 

but wondering if I should. I can add a pseudo element for the background so I can just flip the image in CSS. Thoughts?

index.html
```diff
@@ -21,6 +21,7 @@ <h1 class="d2l-heading-2">Demos</h1>
 
 	<h2 class="d2l-heading-3">Components</h2>
 	<ul>
+		<li><a href="components/typography/demo/typography.html">d2l-typography</a></li>
```
dbatiste, line 24: We should move this down so it's  in alpha order.

dbatiste:
> I can add a pseudo element for the background so I can just flip the image in CSS. Thoughts?

I was thinking about this too, except for a different reason. If we ever wanted this to display both an opening and closing quote, the single background-image would be insufficient, and so I was ponding whether we could use `::before` / `::after` for them. But yeah, I would expect the image to get flipped in RTL so I think this is worth exploring.

devpow112:
I'll take a look at flipping the quote Monday

index.html
devpow112: Will do

components/typography/typography.scss
```diff
@@ -137,3 +137,26 @@
 	font-weight: 700;
 	letter-spacing: 0.2px;
 }
+
+
+@mixin d2l-blockquote {
```
dlockhart, line 142: Debatable whether we need to keep supporting Sass with new stuff I think? I'd love to be able to phase it out one day when everything is Lit, so the more we add here the harder that'll be.

devpow112: If you want I can exclude the blockquote stuff. I just added it in all files.

dbatiste: Yeah I would exclude it.

components/typography/test/typography.visual-diff.js
```diff
@@ -61,6 +61,16 @@ describe('d2l-typography', function() {
...
+			const rect = await visualDiff.getRect(page, '#blockquote', 0);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+
+		it('blockquote-rtl', async function() {
+			const rect = await visualDiff.getRect(page, '#blockquote-rtl', 0);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
```
dlockhart, line 72: So technically these visual diffs are only testing half the code -- the part in `typography.js`, but not the shared styles in `styles.js`. That's true for everything in this visual-diff file though, so I guess we can just follow along. But probably something for us (Gaudi) to resolve.

dlockhart (approved):

components/typography/test/typography.visual-diff.js
devpow112: Yeah was thinking the same thing. Might need a separate vdiff page for testing both scenarios.

dbatiste (approved):
Looks good to me.

PR #2556, opened 2022-06-20
Changed menu-item focus to focus-visible to prevent showing focus when navigating with mouse

oelhanafey:
https://rally1.rallydev.com/#/?detail=/userstory/451292357272&fdp=true

Focus is no longer shown when navigating menu components with the mouse to avoid confusing situations where 1 item is focused and another item is hovered.

Also fixed problem where disabled elements were not showing correct styles in Safari because it could not parse `focus-visible`, so I separated the `focus-visible` selector from others.

The Menu component must also extend the `FocusVisiblePolyfillMixin` in order to pass the `focus-visible` state down to nested menu-item elements in the Shadow DOM even though the menu component itself does not use focus-visible.


eKoopmans:
LGTM, but I'll wait for another set of eyes before approving 👍.

components/menu/menu-item-styles.js
```diff
@@ -37,7 +37,13 @@ export const menuItemStyles = css`
...
 	}
 
-	:host([disabled]), :host([disabled]:hover), :host([disabled]:focus-visible) {
+	:host([disabled]), :host([disabled]:hover), :host([disabled].focus-visible) {
+		cursor: default;
+		opacity: 0.75;
+	}
+
+	/** separated because safari is having trouble parsing these */
+	:host([disabled]:focus-visible) {
```
dlockhart, line 46: I'm confused as to why this needed the `:focus-visible` pseudo-class... shouldn't it have been handled by `.focus-visible` polyfill class in the ruleset above?

oelhanafey: You are right, it is redundant. I think I was just hesitant to remove it because I felt like the pseudo-class would be more stable through browser updates. I will remove them 👍

eKoopmans (approved):

components/menu/menu-item-styles.js
dlockhart: Cool! Yeah I think once we have decent across-the-board browser support for native `:focus-visible`, we'll go through all these and switch them over. Need to wait probably another year and a half though for more people to have the latest Safari.

PR #2558, opened 2022-06-20
chore: Remove screenshots

eKoopmans:
Removing screenshots from core because they are out of date, and superseded by Daylight.

**Note:** The second commit, `Remove all remaining screenshots`, involves readmes that are not directly displayed in Daylight. If there are any docs there that should keep their screenshots, please flag them, thanks!

Rally: https://rally1.rallydev.com/#/?detail=/task/640674899109&fdp=true


svanherk (approved):
Thanks! 🎉

margaree (approved):

PR #2559, opened 2022-06-20
Add prism helper.

dbatiste:
This PR adds a prism helper for dynamically loading prism. There are a few things included in this PR, but it all goes together and it's all fairly straight-forward. Specifically it includes:

* Light & dark code sample styles (subject to change), including generating CSS color variables
* Exported `getCodeColors(mode)` method the editor can use to get the token colors it needs to match CodeMirror styles
* Exported `codeStyles` that `d2l-html-block` can use to target elements its shadowDOM scope
* Exported `formatCodeElement(elem)` method for `d2l-htmleditor` and `d2l-html-block` to use
* Lazy-loading Prism, plugins (ex. line-numbers), and languages/grammar from the CDN
* A `HtmlBlockCodeRenderer` plugin for `d2l-html-block` rendering

image

Debating best way to add Visual-Diff for this. Maybe one per token, light & dark, with line numbers & without.


helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+	code[class*="language-"].d2l-code {
+		${unsafeCSS(generateColorVariables(colorModes.LIGHT))}
+	}
+
+	pre[class*="language-"].d2l-code-dark,
+	code[class*="language-"].d2l-code-dark {
+		${unsafeCSS(generateColorVariables(colorModes.DARK))}
+	}
+
+	pre[class*="language-"].d2l-code,
```
dbatiste, line 157: Note: many of these styles are "standard" across Prism themes. We need to be careful to preserve specificity in cases where another tool (ex. tinyMCE) happens to include Prism styles, or a content author has referenced their own version of Prism.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+codeLanguages.set('racket', 'Racket');
+codeLanguages.set('sql', 'SQL');
+codeLanguages.set('wolfram', 'Wolfram');
+
+export const colorModes = {
+	LIGHT: 'light',
+	DARK: 'dark'
+};
+
+const darkColors = Object.freeze({
```
dbatiste, line 39: Light & dark colors are subject to change as we get more input from designers.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+const getLanguageInfo = elem => {
+	const classes = elem.classList;
+	for (let i = 0; i < classes.length; i++) {
+		if (classes[i].startsWith('language-')) {
+			const key = classes[i].substring(9);
+			const desc = codeLanguages.get(key);
+			if (desc) return { key: key, desc: desc };
+		}
+	}
+	return { key: 'plain', desc: codeLanguages.get('plain') };
```
dbatiste, line 309: If the specified language is not recognized, we fall back to the `plain` language component which will be formatted without syntax highlighting, but will still include line numbers if requested.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+};
+
+const languageDependencies = {
+	arduino: [ 'cpp' ],
+	cpp: [ 'c' ],
+	racket: [ 'scheme' ]
+};
+
+const languagesLoaded = {
```
dbatiste, line 318: These language components are automatically loaded with the main `prism.js` download, so they are initialize with resolved promises.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+		if (classes[i].startsWith('language-')) {
+			const key = classes[i].substring(9);
+			const desc = codeLanguages.get(key);
+			if (desc) return { key: key, desc: desc };
+		}
+	}
+	return { key: 'plain', desc: codeLanguages.get('plain') };
+};
+
+const languageDependencies = {
```
dbatiste, line 312: These languages depend on other language components being loaded (in order).

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+	plain: Promise.resolve()
+};
+
+const loadLanguage = async key => {
+	if (languagesLoaded[key]) return languagesLoaded[key];
+
+	// Prism languages can extend other anguages and must be loaded in order
+
+	// eslint-disable-next-line no-async-promise-executor
+	languagesLoaded[key] = new Promise(async resolve => {
```
dbatiste, line 332: We need to be very careful of race conditions if multiple code elements are being formatted at the same time. This is why `languagesLoaded[key]` is assigned a Promise immediately instead of waiting for language dependencies to load before assigning the Promise. Hence the eslint override.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+
+	return pluginsLoaded[plugin];
+};
+
+let prismLoaded;
+
+const loadPrism = () => {
+	if (prismLoaded) return prismLoaded;
+
+	prismLoaded = Promise.all([
```
dbatiste, line 368: We use a `Promise.all` instead of sequential `await` calls so we can load the JS and CSS in parallel.

helpers/prism.js
```diff
@@ -0,0 +1,432 @@
...
+}
+
+export class HtmlBlockCodeRenderer {
+
+	get canRenderInline() {
+		return true;
+	}
+
+	async render(elem) {
+		const codeElements = [...elem.querySelectorAll('.d2l-code')];
```
dbatiste, line 421: We specifically only want to format elements that have `d2l-code` class.  Authors could have used `pre` or `code` elements for other reasons and we don't want to format those elements.

helpers/prism.js
grant-cleary: Efficiency!

.gitignore
```diff
@@ -2,6 +2,8 @@ generated/
 node_modules/
 components/*/test/screenshots/current/
 components/*/test/screenshots/golden/
+helpers/test/screenshots/current/
+helpers/test/screenshots/golden/
```
dlockhart, line 6: I'd love to re-organize everything so that tests & demos have their own top-level directory instead of nested inside each component. Would make `eslint` config easier too.

grant-cleary (approved):

dlockhart (approved):
Very cool!

helpers/prism.js
```diff
@@ -0,0 +1,430 @@
...
+codeLanguages.set('kotlin', 'Kotlin');
+codeLanguages.set('latex', 'LaTeX');
+codeLanguages.set('markup', 'Markup');
+codeLanguages.set('matlab', 'MATLAB');
+codeLanguages.set('plain', 'Plain Text');
+codeLanguages.set('python', 'Python');
+codeLanguages.set('r', 'R');
+codeLanguages.set('racket', 'Racket');
+codeLanguages.set('sql', 'SQL');
+codeLanguages.set('wolfram', 'Wolfram');
```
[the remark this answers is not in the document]
dbatiste: I was wondering about this when I first wrote it, but admittedly I wasn't sure so I avoided the array initialization approach because I was averse to creating arrays that would be temporary.  So... I decided to do up the 3 methods of initializing and run a test - [benchmark can be run here](https://jsbench.me/wrl4ohuosv/1).

This is the result: 
Initialize with `set`: _fastest_
Initialize with `array`: 39.21% slower
Initialize with `object`: 63.87% slower

However, I'm not sure how much trust to put it the tool, and the result may vary depending on the size of the initial dataset. I think it's fast regardless, so putting the perf aside, I am happy to switch it if we prefer, otherwise I'll just leave it as-is.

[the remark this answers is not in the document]
grant-cleary: Interesting. I'm more than a bit surprised that the object/array initialization methods are slower. You would think the engine would treat them equivalently. Good to know I suppose.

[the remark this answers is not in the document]
dbatiste: I suppose that in the slower cases, we are effectively creating an intermediate data structure, and the Map constructor then processes that and populates it's own the same way that `set` would. 🤷‍♂️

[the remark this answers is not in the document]
grant-cleary: Yeah, that would make sense. I suppose the constructor exists more for converting existing iterables into maps than for actual initialization.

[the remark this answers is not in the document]
dlockhart: Crazy! Yeah I'm good with whatever. I personally find the object initialization the easiest to read and it'll end up making our code smaller without N `codeLanguages.set(` calls, but 🤷 .

PR #2564, opened 2022-06-21
fix: minor adjustments to blockquote line height

devpow112:
Realized when integrating into app location that the line height was slightly different then the use cases.


components/typography/typography.js
```diff
@@ -112,41 +112,11 @@ if (!document.head.querySelector('#d2l-typography-font-face')) {
 			margin: 1.5rem 0 1.5rem 0;
 		}
 
-		@media (max-width: 615px) {
```
devpow112, line 115: This was just re-ordering to be more consistent with what was there before.

dlockhart (approved):

PR #2566, opened 2022-06-21
fix: remove blockquote color

devpow112:
The color varies in 1 of the 3 use cases I've tried integrating so not setting a default font color seems like a better idea. This means I'll only have to make style adjustments in 1 place instead of 2. This should be the last change related to this stuff, promise 😄.


dbatiste:
Are the color differences legit, or should they arguably be using the same color?

devpow112:
> Are the color differences legit, or should they arguably be using the same color?

Yes. 2 of the 3 use cases are for an overall feedback type of thing where the feedback text is the main thing the user is meant to see (these are darker using our standard font color). The last is for feedback on an outcome where the achieved level is the main thing the user is meant to see and the text is secondary (text is lighter in this case). By removing the color here it will mean I only have to make style adjustments for 1 of the 3 cases instead of 2.

dbatiste (approved):

PR #2569, opened 2022-06-22
Filter scrollbar disappearing defect

oelhanafey:
https://rally1.rallydev.com/#/?detail=/defect/639773755745&fdp=true

Solved a bug where scrollbars were disappearing when unfiltering search results in the filter component. (See rally ticket)

The reason this bug was occurring was that the filter component was not resizing the dropdown after each search, so the size of the dropdown would be set when it was first opened. If the dropdown was opened already filtered to a small number of elements then the size of the dropdown would be set to smaller than the total number of unfiltered elements and there would be no scrollbar after unfiltering the search results.

Now it calls the resizeDropdown function after every search, fixing the problem.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -550,7 +550,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		if (shouldRecount) this._setFilterCounts(dimension);
 		if (shouldUpdate)  this.requestUpdate();
 		if (shouldResizeDropdown) {
-			this._resizeDropdown();
+			this._requestDropdownResize();
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


oelhanafey:
Adding some additional context after discussing with Stacey:

**Why do we only call resize on components with a single dimension?**
The multi-dimension `filter` component is able to resize the dropdown without an explicit function call. This is likely because the multi-dimension `filter` component is built off of the `hierarchical-view` component which has its own observers that are able to tell if the dropdown size changes and resize it.

**Do we need to call the resize function every time the search is run or is there a way to tell when the issue will occur?** 
I think that we do. The resize issue doesn’t only happen in the specific case mentioned in the defect, it happens every time. When you run a search from unfiltered and filter to a single element (see the attached picture), there is still a scrollbar and a weird margin below the element that goes away after you request a resize. It only becomes disruptive/unusable when filtering from less elements to more elements with a small screen size but the problem is there regardless.

screen_shot_2022-06-22_at_2 10 21_pm

eKoopmans (approved):
LGTM!

PR #2573, opened 2022-06-24
Fix the readme example for d2l-validator-custom.

dbatiste:
This PR corrects the example in the readme for `d2l-validator-custom`. The event is dispatched on the validator, not the element being validated. The tests are working as expected.


oelhanafey (approved):
Yep the validator is the one that is triggering the event. LGTM

PR #2580, opened 2022-06-27
fix: Daylight: Tooltip: update tooltip best practices heading level

margaree:


dlockhart (approved):

eKoopmans (approved):

dcostiuc:
Good catch Margaree! 

For what it's worth , I was so used to seeing that `Best Practices` section formatted that way that I thought we just preferred it like that, even if we're doing things a bit differently for the `Tooltip` page.

Point being, I probably wouldn't have caught this discrepancy or noted it myself, so thanks for taking care of it!

PR #2589, opened 2022-06-28
Empty State Component

oelhanafey:
https://rally1.rallydev.com/#/?detail=/userstory/640869004423&fdp=true

Design docs: [here](https://www.figma.com/proto/sAfbvafY6gRz8ZpHrG1Qmu/Empty-States?page-id=0%3A1&node-id=27%3A1236&viewport=-1088%2C734%2C0.34&scaling=min-zoom&starting-point-node-id=2%3A4)

Current work for empty state component. Added the `empty-state-text` component which is meant to be placed when there is no data available to be displayed. It takes a description and action-text to be placed in the button. In lieu of the action-text, consumers can also place a `d2l-link` or `d2l-button-subtle` in the default slot to render your own custom action.


components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+				padding-right: 0;
+			}
+			::slotted(*) {
+				display: none;
+			}
+			::slotted(d2l-button-subtle:first-child),
+			::slotted(d2l-link:first-child) {
+				display: inline;
+			}
+			span {
```
oelhanafey, line 50: `d2l-button-subtle` uses the typography `labelStyles` which has a static font-size of `0.7rem`. In order for the description text to look good in-line next to the button, I have also set the font-size of the description to a static `0.7rem`. The downside to this is that font-size will not change at different breakpoints.

components/empty-state/test/empty-state-text.visual-diff.html
```diff
@@ -0,0 +1,57 @@
...
+    <script type="module">
+        import '../..//typography/typography.js';
+        import '../empty-state-text.js';
+        import '../../button/button-subtle.js';
+        import '../../link/link.js';
+    </script>
+</head>
+<body class="d2l-typography">
+
+    <div class="visual-diff">
```
oelhanafey, line 16: Need to add margins to separate different visual diff tests

components/empty-state/demo/empty-state.html
```diff
@@ -0,0 +1,32 @@
...
+		<meta charset="UTF-8">
+		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+		<script type="module">
+			import '../../demo/demo-page.js';
+			import '../../demo/demo-snippet.js';
+			import '../empty-state-text.js';
+		</script>
+		<style>
+			.d2l-typography p:first-child { margin-top: 0; }
+			.d2l-typography p:last-child { margin-bottom: 0; }
```
dlockhart, line 14: Was this CSS still needed? Don't see any `<p>` tags in the document.

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+				padding-right: 0.5rem;
+			}
+		`;
+	}
+
+	render() {
+		if (this.actionText !== undefined) {
+			return html`
+				<div>
+					<span>${this.description}</span>
```
dlockhart, line 63: I think maybe a `<p>` element with its margins squished and `display: inline` might be more appropriate semantically?

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+					<d2l-button-subtle dir=${ifDefined(this.dir === 'rtl' ? 'rtl' : undefined)} @click=${this._handleActionClick} text=${this.actionText}  h-align="text"></d2l-button-subtle>
```
dlockhart, line 64: I'm surprised you needed to pass `dir` down... normally that should just be handled automatically since subtle buttons already use the `RtlMixin`.

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+				</div>
+			`;
+		}
+	}
+
+	_handleActionClick(e) {
+		e.stopPropagation();
+		this.dispatchEvent(new CustomEvent(
+			'd2l-empty-state-action',
+			{ bubbles: true })
```
dlockhart, line 82: We try to use `bubbles: true` very sparingly since 99% of the time folks just wire up to the element directly and don't need to listen for events further up the DOM tree. So unless this is needed for a known use-case, I'd go with no bubbling.

components/empty-state/test/empty-state-text.axe.js
```diff
@@ -0,0 +1,23 @@
...
+		const el = await fixture(html`
+            <d2l-empty-state-text description='There are currently no courses here.' action-text='Create a Course'></d2l-empty-state-text>
+        `);
+		await expect(el).to.be.accessible();
+	});
+
+	it('custom link', async() => {
+		const el = await fixture(html`
+            <d2l-empty-state-text description='There are currently no courses here.'>
+				<d2l-link  href="https://www.d2l.com/">Create a Course</d2l-link>
```
dlockhart, line 17: I wasn't in the planning sessions for this, but I'm curious about how we landed on using a slot for the link use-case? It seems like for the majority of uses of this, the "Create a X" action will perform a navigation, and therefore this should almost always be a link (`<a>` tag) vs. performing an action (like "Clear Search" in the design) where it should be a button.

I think I'm just making the argument that the link-ness or button-ness should just be built into the component's API first-class using attributes, instead of relying on a slot where they can throw anything in there.

A proposal -- this would render a link:
```html
<d2l-empty-state-text description="blah" action-text="Create" action-href="/create/thing">
```

And this would render a button:
```html
<d2l-empty-state-text description="blah" action-text="Create" @empty-state-action="${this.handleClick}">
```

components/empty-state/test/empty-state-text.test.js
```diff
@@ -0,0 +1,61 @@
...
+		const actionText = 'Create a Course';
+		const el = await fixture(html`
+			<d2l-empty-state-text description='There are currently no courses.' action-text=${actionText}>
+				<d2l-link  href="https://www.d2l.com/">Should Not Exist</d2l-link>
+			</d2l-empty-state-text>
+		`);
+		expect(el.shadowRoot.querySelector('d2l-link')).to.not.exist;
+		expect(el.shadowRoot.querySelector('d2l-button-subtle')).to.exist;
+		expect(el.shadowRoot.querySelector('d2l-button-subtle').text).to.be.equal(actionText);
+	});
```
dlockhart, line 53: Except for the test for the event firing below, I might ditch the rest of these and just rely on visual-diff to test those cases. They're going to be pretty fragile when the internals of the component change, and aren't really adding much value when visual-diff exists to check the same things.

index.html
```diff
@@ -73,6 +73,7 @@ <h2 class="d2l-heading-3">Components</h2>
 			</ul>
 		</li>
 		<li><a href="components/expand-collapse/demo/expand-collapse-content.html">d2l-expand-collapse-content</a></li>
+		<li><a href="components/empty-state/demo/empty-state.html">d2l-empty-state</a></li>
```
dlockhart, line 76: I'd flip this with expand/collapse to be alphabetical.

oelhanafey: Oops, thanks!

components/empty-state/test/empty-state-text.test.js
oelhanafey: That's a good point, I didn't consider that the visual diffs overlap with many of these. I suppose it is better to just test that the component is working functionally and leave these sorts of tests to the visual diffs. 👍

dlockhart: Yeah exactly. Existing examples don't always follow that reasoning (we didn't have visual-diff out of the gate either), but generally yeah if you're testing that setting a certain attribute results in something appearing a certain way or in a certain place then visual-diff is great for that without worrying too much about _how_ that happened DOM & CSS wise.

components/empty-state/test/empty-state-text.axe.js
oelhanafey: Admittedly, it was not one of the things we expanded upon during the planning session. I was actually thinking similarly about this and was going to propose that we change the default action in the component to be a link rather than a button but I really like your suggestion to alter the API and remove the need to use a slot at all.

There was some discussion on whether we would want to support other actions like a primary button on a page with no other actions, but I think that would be more for the illustration empty-state component.

I will bring this idea to the team. Thanks!

components/empty-state/empty-state-text.js
oelhanafey: I didn't realize that was the expected behaviour. This is what happens when you don't pass the `dir` down. The button does not use the `rtl` style for `h-align` and aligns the button to the left side. I could further look into why this is happening.

Screen Shot 2022-06-29 at 11 18 04 AM

components/empty-state/empty-state-text.js
oelhanafey: I agree, thanks!

components/empty-state/test/empty-state-text.axe.js
dlockhart: Cool, sounds good! There are other options as well... like you could have separate components `<d2l-empty-state-text-link>`, `<d2l-empty-state-text-button>`, or could leverage a `<d2l-empty-state-text action-type="link|button|link-primary">`. The `action-type` approach can get weird because they can choose `link` without passing an `href` or `button` AND pass an `href`, so we typically try to avoid that kind of API.

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+				line-height: 1rem;
+				padding-right: 0.5rem;
+			}
+		`;
+	}
+
+	render() {
+		if (this.actionText !== undefined) {
+			return html`
+				<div>
```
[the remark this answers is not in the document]
oelhanafey: Yep, that would be cleaner!

components/empty-state/empty-state-text.js
dlockhart: We've just been bitten before with things like dialog where multiple dialogs can actually be nested and we had event bubbling incorrectly turned on so consumers could get events from dialogs they weren't even aware of. More of an issue when `composed: true` is also set (which allows events to cross shadow boundaries), but just generally I'm really conservative with event bubbling haha.

components/empty-state/empty-state-text.js
dlockhart: It may just be a problem with the demo-snippet as it has limited capabilities to pass that down through shadow roots. Try putting `dir="rtl"` on the `<html>` element of the page and I'm guessing that'll work.

oelhanafey: Oh I was using it incorrectly. I was putting `dir="rtl"` on the `empty-state-text` component directly. Putting it on the `<html>` element fixed the problem!

components/empty-state/empty-state-text.js
dbatiste: Normally when we render the `0.7rem` font-size it is by using [bodySmallStyles](https://github.com/BrightspaceUI/core/blob/main/components/typography/styles.js#L65), but it renders with a tungsten color, so we'd be deviating from our typography here.  I think we should review the issue with using the 0.8rem size with Neroli & Jeff, and call attention to the difference from our typography.

components/empty-state/empty-state-text.js
oelhanafey: Haha okay sounds good. It doesn't seem like it is necessary to bubble in this situation.

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+			::slotted(*) {
+				display: none;
+			}
+			::slotted(d2l-button-subtle:first-child),
+			::slotted(d2l-link:first-child) {
+				display: inline;
+			}
+			span {
+				font-size: 0.7rem;
+				font-weight: 400;
```
nathankoop, line 52: do we need to set font-weight here? I'm pretty sure that 400 is the default, or is it overridden elsewhere?

components/empty-state/empty-state-text.js
```diff
@@ -0,0 +1,88 @@
...
+			return html`
+				<div>
+					<span>${this.description}</span>
+					<d2l-button-subtle dir=${ifDefined(this.dir === 'rtl' ? 'rtl' : undefined)} @click=${this._handleActionClick} text=${this.actionText}  h-align="text"></d2l-button-subtle>
+				</div>
+			`;
+		}
+		else {
+			return html`
+				<div>
```
dbatiste, line 70: Do we need to conditionally render the whole subtree?  Seems like we should be able to use:

``` javascript
<p>${this.description}</p>
${ this.actionText ? html`<d2l-button-subtle ...>...</d2l-button-subtle>` : html`<slot></slot>` }
```

components/empty-state/test/empty-state-text.axe.js
dbatiste: The reason the slot was suggested is that the subtle button is the primary use-case we expect, but we would like the ability to override, sometimes with a primary button, but possibly even with a button-menu. The slot would enable them to do that without surfacing those semantics on the empty-state component itself. Honestly, I feel conflicted about even baking the subtle button in, but Jeff & Neroli really want consumers to follow that pattern.  If we go the slot route, we can target and hide slotted elements that we don't want them to put in there. 

If we go with different components, we can lock things down better, but we will have the two variants (due to very different layouts) multiplied by the different types of actions. 🤷 Lots of components.

Completely agree on overloading a single component - it can get a little messy.

Anyway, that was the original thinking, and we can certainly pivot from here.

components/empty-state/test/empty-state-text.visual-diff.html
dbatiste: Most other visual-diff test files include a reference to :

``` html
<link rel="stylesheet" href="../../../test/styles.css" type="text/css">
```

It include some useful styles you'll probably want.
https://github.com/BrightspaceUI/core/blob/main/test/styles.css

components/empty-state/test/empty-state-text.axe.js
dlockhart: > ...the subtle button is the primary use-case we expect

I guess I'm questioning that? Only based on the examples in the design though, so if there's better data out there let's use that. But from the design, "Create Learning Paths", "View All Activities", "Create an Assignment" should be all be links because they navigate and only "Clear Search" should be a button since it's an action. I think this is why designers keep asking for a `<d2l-link-subtle>`! :)

components/empty-state/empty-state-text.js
oelhanafey: Yep, that should be doable

components/empty-state/test/empty-state-text.visual-diff.html
oelhanafey: Thanks!

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -44,11 +44,9 @@ class EmptyStateIllustratedButton extends RtlMixin(LitElement) {
 
 	render() {
 		return html`
-		<div>
-                ${this.illustration ? html`${runAsync(this.illustration, () => getIllustration(this.illustration), {
+			${this.illustration ? html`${runAsync(this.illustration, () => getIllustration(this.illustration), {
```
dbatiste, line 47: Nit: I might be inclined to name this `illustrationName`/`illustration-name` for clarity... so consumers don't get the impression that they can stuff an `svg` in there.

components/empty-state/demo/empty-state.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+		<link rel="stylesheet" href="../../demo/styles.css" type="text/css">
+		<script type="module">
+			import '../../demo/demo-page.js';
+			import '../../demo/demo-snippet.js';
```
dbatiste, line 9: We don't often do this, but the demo-page automatically imports the demo-snippet so that demo pages don't have to. It does it purely as a convenience.

components/empty-state/getIllustration.js
```diff
@@ -0,0 +1,25 @@
+import { html } from 'lit';
+import { loadSvg } from '../../generated/empty-state/presetIllustrationLoader.js';
+import { unsafeSVG } from 'lit/directives/unsafe-svg.js';
+
+function fixSvg(svgStr) {
```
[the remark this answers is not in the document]
dbatiste: No, that's not necessary. It's already private to this module since it's not being exported. In components we prefix private members with `_` because the component is effectively public.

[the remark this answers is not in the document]
dbatiste: Is this `fixSvg` function necessary? For icons, it does a few things that aren't necessary here.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -0,0 +1,63 @@
...
+ * The `d2l-empty-state-illustrated-button` component is an empty state component that displays an illustration and action button. The illustration property can be set to use one of the preset illustrations or a custom SVG illustration can be added in the default slot.
+ * @fires d2l-empty-state-action - Dispatched when the action button is clicked
+ * @slot - Custom SVG content if illustration property is not set
+ */
+class EmptyStateIllustratedButton extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			/**
+			 * Optional: The action text to be used in the subtle button
```
dbatiste, line 19: Is the idea with making this "Optional", that the consumer could omit it, and they'd get an empty-state component without an action? It would be a little strange to use a button-specific component, only to exclude the button. I'm not sure if there is a use-case for not having an action. Worth discussing with Design.

I think this goes for the link variants as well, and confirm wether this case needs to be handled. The link variants also have properties that should seemingly be required, in the same way that `d2l-link` requires an `href`.

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+class EmptyStateIllustratedLink extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			/**
+			  * Optional: The action URL or URL fragment of the link
+			  * @type {String}
+			  */
+			actionHref: { type: String, attribute: 'action-href' },
```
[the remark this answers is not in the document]
dbatiste: I think this relates to my comment above about whether or not these properties need to be required/optional.

components/empty-state/empty-state-illustrated-button.js
oelhanafey: That was my thought process, yes. But I agree, this is something that I had noted down to discuss with Design.

components/empty-state/empty-state-illustrated-button.js
oelhanafey: Yep, good call

components/empty-state/getIllustration.js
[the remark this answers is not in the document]
oelhanafey: Nope, it doesn't seem necessary, thanks.

components/empty-state/empty-state-illustrated-link.js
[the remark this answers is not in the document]
dlockhart: I actually went the opposite way recently with `d2l-navigation-link-icon` where it throws if an `href` isn't provided:
https://github.com/BrightspaceUILabs/navigation/blob/main/d2l-navigation-link-icon.js#L89

That way we're ensuring that people aren't using the link one just for its styling when they should be using the button one.

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -0,0 +1,62 @@
...
+			/**
+			 * Optional: The name of the preset image you would like to display in the component
+			 * @type {String}
+			 */
+			illustration: { type: String },
+			/**
+			 * REQUIRED: A title for the empty state
+			 * @type {String}
+			 */
+			title: { type: String },
```
dlockhart, line 41: So `title` is actually a reserved HTML attribute name for showing a native browser tooltip on-hover -- I'm guessing if you hover over the component you'll see this as a tooltip, which isn't what you intended. :)

I'd recommend `title-text` as an alternate name, which we've used elsewhere.

components/empty-state/empty-state-illustrated-link.js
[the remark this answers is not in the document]
oelhanafey: That makes sense and would probably fit better with our decision to go with a more restricted API.

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -0,0 +1,62 @@
...
+
+	static get styles() {
+		return [emptyStateStyles, emptyStateIllustratedStyles, bodyCompactStyles, heading2Styles];
+	}
+
+	render() {
+		return html`
+			${this.illustration ? html`${runAsync(this.illustration, () => getIllustration(this.illustration), {
+		success: (illustration) => illustration
+	}, { pendingState: false })}` : html`<slot></slot>`}
```
dlockhart, line 53: Are all the illustrations the same size? If so, it might be possible to render a wrapper with a fixed size that would help to avoid jitter when the illustration SVG loads in? Might be easy to test out if you faked a delay in `getIllustration` of like 1-2 seconds and see what happens.

components/empty-state/empty-state-styles.js
```diff
@@ -0,0 +1,61 @@
+import '../colors/colors.js';
+import { css } from 'lit';
+
+export const emptyStateStyles = css`
+
+	:host {
+		border: 1px solid var(--d2l-color-mica);
+		border-radius: 6px;
```
dlockhart, line 8: If the design intent here was to match the radius of our buttons, we actually use `0.3rem` -- which is equivalent to `6px` except that if the user is using a larger font size, it'll be a larger radius.

components/empty-state/getIllustration.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export async function getIllustration(illustration) {
+
+	if (illustration) {
+		const svg = await loadSvg(illustration);
+		return fixSvg(svg ? svg.val : undefined);
+	}
+
+}
```
dlockhart, line 25: Nit on the illustration filenames (I can't comment on the SVGs!): I'd go with `checklist.svg` and `desert-road.svg` since "illustration" is redundant and we typically go with hyphens not underscores to separate words in filenames.

components/empty-state/empty-state-illustrated-link.js
oelhanafey: Based on my 2 illustration sample size, I suspect they might not be but I will note it down to bring up with the team.

components/empty-state/getIllustration.js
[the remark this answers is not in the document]
dbatiste: Cool... that might eliminate the need for this module altogether. You may want to check what happens when the consumer provides an invalid illustration name.

components/empty-state/getIllustration.js
oelhanafey: Yep, these were more of a placeholder for my testing. We still need to have a chat about the illustrations and illustration names we will be using.

components/empty-state/getIllustration.js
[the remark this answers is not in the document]
oelhanafey: Yeah, I was trying to avoid repeated code but I suppose for a small function like `getIllustration` it isn't that bad. 

An incorrect illustration name will just result in no svg rendered in the DOM. There are no errors.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -0,0 +1,127 @@
...
+
+	static get styles() {
+		return [bodyCompactStyles, emptyStateStyles, emptyStateIllustratedStyles];
+	}
+
+	constructor() {
+		super();
+
+		this._contentWidth = 0;
+		this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
```
oelhanafey, line 60: Instead of using the viewport typography breakpoints, we decided to resize (titleText sizing and illustrations margins) based on the size of the container.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -0,0 +1,127 @@
...
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._resizeObserver.disconnect();
+	}
+
+	render() {
+		const illustrationContainerStyle = {
+			height: `${Math.min(this._contentWidth, 500) / 1.5}px`,
+		};
```
oelhanafey, line 88: All preset illustrations have an aspect ratio of `1.5:1`. Since the illustrations are loaded in asynchronously we should set the height of the container ahead of time to prevent jitter when loading the page.

In the case that someone uses a custom SVG in the slot, I did not set the height of the container ahead of time since we cannot guarantee that the consumer will use an SVG with a similar aspect ratio.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -0,0 +1,127 @@
...
+
+		const titleSmall = this._contentWidth <= 615;
+		const titleClass = {
+			'd2l-empty-state-title': true,
+			'd2l-empty-state-title-small': titleSmall,
+			'd2l-empty-state-title-large': !titleSmall,
+		};
```
oelhanafey, line 95: Instead of using the viewport typography breakpoints, we decided to resize (titleText sizing and margins) based on the size of the container using `ResizeObserver`.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -1,63 +1,139 @@
...
+			 */
+			primary: { type: Boolean },
 			/**
 			 * REQUIRED: A title for the empty state
-			 * @type {String}
+			 * @type {string}
 			 */
-			title: { type: String }
+			titleText: { type: String, attribute: 'title-text' },
+			_contentWidth: { type: Number }
```
dlockhart, line 48: For internal state that you don't intend for consumers to set/be aware of, you can use Lit 2's new state stuff:

```suggestion
			_contentWidth: { state: true }
```

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -1,63 +1,139 @@
...
+		this._resizeObserver.disconnect();
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		if (!this.titleText || this.titleText.length === 0) {
+			console.warn('d2l-empty-state-illustrated-button component requires titleText.');
+		}
+
```
dlockhart, line 79: I don't think this is going to work in practice -- since often these are going to be lang terms that may not be set immediately on first-render.

Currently the only way we've successfully handled required attributes like this is to wait a few seconds after render (or update) and then check. Here's [a recent example in navigation](https://github.com/BrightspaceUILabs/navigation/blob/main/d2l-navigation-link-icon.js#L98). Longer term, we may want to look into some kind of mixin/helper/controller that could help make this kind of thing easier.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -1,63 +1,139 @@
...
+			console.warn('d2l-empty-state-illustrated-button component requires a description.');
+		}
 	}
 
 	render() {
+		const illustrationContainerStyle = {
+			height: `${Math.min(this._contentWidth, 500) / 1.5}px`,
+		};
+
+		const titleSmall = this._contentWidth <= 615;
```
dlockhart, line 90: The way this is written, every time `_contentWidth` changes it's going to cause a re-render. But in reality, only when the value drops below `615` or `500` do we really need to re-render.

I think you can optimize it a bit by creating `_titleSmall` boolean and `_illustrationHeight` state properties that get set from the resize handler. Then only when those values change will Lit actually re-render.

components/empty-state/empty-state-illustrated-button.js
```diff
@@ -1,63 +1,139 @@
...
+			'd2l-empty-state-title-large': !titleSmall,
+		};
+
+		let actionButton = nothing;
+		if (this.actionText && this.primary) {
+			actionButton = this.actionText && html`<d2l-button class='d2l-empty-state-action' @click=${this._handleActionClick} primary>${this.actionText}</d2l-button>`;
+		}
+		else if (this.actionText && !this.primary) {
+			actionButton = this.actionText && html`<d2l-button-subtle class='d2l-empty-state-action' @click=${this._handleActionClick} text=${this.actionText}></d2l-button-subtle>`;
+		}
```
dlockhart, line 103: Nit: could be simplified a bit I think:

```suggestion
		let actionButton = nothing;
		if (this.actionText) {
			actionButton = this.primary ? html`<d2l-button class='d2l-empty-state-action' @click=${this._handleActionClick} primary>${this.actionText}</d2l-button>` :
				html`<d2l-button-subtle class='d2l-empty-state-action' @click=${this._handleActionClick} text=${this.actionText}></d2l-button-subtle>`;
		}
```

components/empty-state/empty-state-illustrated-button.js
dlockhart: Also I think you only care about `_illustrationHeight` (and `illustrationContainerStyle`) if `illustration-name` is actually set -- so again some re-renders could be avoided I think when they're using a custom SVG.

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -1,62 +1,127 @@
...
 	static get styles() {
-		return [emptyStateStyles, emptyStateIllustratedStyles, bodyCompactStyles, heading2Styles];
+		return [bodyCompactStyles, emptyStateStyles, emptyStateIllustratedStyles];
+	}
+
+	constructor() {
+		super();
+
+		this._contentWidth = 700;
+		this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
```
dlockhart, line 57: These are now sharing a lot more logic than they originally were... wondering if it might make sense to move that duplicated stuff into a base class, similar to what @eKoopmans did with the [object property items](https://github.com/BrightspaceUI/core/blob/main/components/object-property-list/object-property-list-item-link.js#L10).

components/empty-state/empty-state-styles.js
```diff
@@ -36,18 +36,35 @@ export const emptyStateIllustratedStyles = css`
 		text-align: center;
 	}
 
-	#d2l-empty-state-description {
-		margin-bottom: 0.3rem;
+	:host([illustration-name="checklist"]) .d2l-empty-state-title {
+		margin-top: 0;
```
dlockhart, line 40: Seems a little weird to have custom styles just for one of the illustration types... could the SVG itself by modified a bit instead?

components/empty-state/README.md
```diff
@@ -0,0 +1,104 @@
...
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/data-tracking.svg?sanitize=true) | data-tracking |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/desert-road.svg?sanitize=true) | desert-road |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/fish-hook.svg?sanitize=true) | fish-hook |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/oven.svg?sanitize=true) | oven |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/pipeline.svg?sanitize=true) | pipeline |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/race.svg?sanitize=true) | race |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/rockets.svg?sanitize=true) | rockets |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/empty-state/images/tumbleweed.svg?sanitize=true) | tumbleweed |
```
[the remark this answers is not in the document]
oelhanafey: The intention was for the illustration catalogue to be directly on the empty state documentation page. I think I might be able to augment the catalogue generator code to directly edit the catalogue in the README.

components/empty-state/empty-state-illustrated-button.js
oelhanafey: I did originally do it that way with a `_titleSmall` attribute. The reason I changed it was so that I could preset the height of the container before the illustration loads in to prevent the initial jitter. I could optimize it somewhat to not update the height when width is `> 500` since height no longer changes but if it is less than that, then it would need to have the exact width measurements in order to get the correct height. Is that what you are proposing?

Agree with you on the second comment, it can be made more efficient there. Thanks

components/empty-state/empty-state-illustrated-link.js
oelhanafey: Yep, they are very similar apart from the specific actions that they have. I think it would make sense

components/empty-state/empty-state-illustrated-button.js
dlockhart: I'm guessing there are cases where they're giving false negatives, yeah.

components/empty-state/empty-state-illustrated-button.js
dlockhart: Yeah exactly. I was proposing two separate state properties -- one to make the title small and one for the illustration height (that would only get changed when width drops below 500), but you're correct that you could accomplish the same thing by just not updating `_contentWidth` until it drops below 500 (or goes back above 500).

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -22,74 +18,17 @@ class EmptyStateIllustratedLink extends RtlMixin(LitElement) {
...
-		const titleClass = {
-			'd2l-empty-state-title': true,
-			'd2l-empty-state-title-small': titleSmall,
-			'd2l-empty-state-title-large': !titleSmall,
-		};
+		const illustrationContainerStyle = this._getIllustrationContainerStyle();
+		const titleClass = this._getTitleClass();
 
 		const actionLink = this.actionText && this.actionHref
 			? html`<d2l-link class='d2l-body-compact d2l-empty-state-action' href=${this.actionHref}>${this.actionText}</d2l-link>`
```
dlockhart, line 34: A good general practice we've been using to try to cut down on the number of web components on a page (since they each cost a little performance) is to use a regular `<a>` whenever we can inside our Lit components.

So you could import `linkStyles` from `../link/link.js`, add those to the styles array and add `d2l-link` to the CSS classes on this and switch it to a `<a>`.

dlockhart: Oh also, double quotes around the attributes for consistency with other stuff.

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -0,0 +1,57 @@
...
+		};
+	}
+
+	static get styles() {
+		return [super.styles, linkStyles];
+	}
+
+	constructor() {
+		super();
+		this._illustratedComponentType = 'link';
```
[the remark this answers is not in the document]
oelhanafey: Hmm it is possible. To do that, I imagine we would need to add a unique ID to each component that we could reference. There doesn't seem to be any precedence for this in our other components which just reference the component type but it could be valuable for us to start doing this.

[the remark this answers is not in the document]
margaree: Ah sorry I was more thinking we could just go the other way and simplify the error messages to not include type of link or button

components/empty-state/empty-state-illustrated-link.js
```diff
@@ -0,0 +1,57 @@
...
+
+		return html`
+			${this.illustrationName
+		? html`
+			<div style="${styleMap(illustrationContainerStyle)}">
+				${runAsync(this.illustrationName, () => this._getIllustration(this.illustrationName), { success: (illustration) => illustration }, { pendingState: false })}
+			</div>`
+		: html`<slot></slot>`}
+
+			<p class="${classMap(titleClass)}" id="d2l-empty-state-title">${this.titleText}</p>
```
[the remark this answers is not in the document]
oelhanafey: Nope, I missed this one when changing to using the classMap. Thanks!

components/empty-state/test/empty-state-illustrated-button.test.js
```diff
@@ -0,0 +1,27 @@
+import '../empty-state-illustrated-button.js';
+import { fixture, html, oneEvent } from '@open-wc/testing';
+import { runConstructor } from '../../../tools/constructor-test-helper.js';
+
+describe('d2l-empty-state-illustrated-button',  () => {
```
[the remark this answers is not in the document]
oelhanafey: My concern with this is that it could slow down our unit tests a lot. After a discussion with Dave L that attributes may not be set instantly, I am now checking for missing attributes after a 3 second delay. Checking all the error messages for my components would add an additional 18+ seconds.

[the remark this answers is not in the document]
margaree: Ah yup that sounds like a valid concern

components/empty-state/empty-state-illustrated-link.js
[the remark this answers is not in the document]
oelhanafey: May not be necessary, but I figured that this was not a lot of overhead and would allow more specific error messages. It allows the error messages to denote the exact component name which could make it easier to find when searching for it rather than something like `Illustrated empty state component is missing x` and then requiring the consumer to search and find out what the name of the component is.

components/empty-state/empty-state-simple-button.js
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this._validateDescription();
+	}
+
+	render() {
+		const actionButton = this.actionText
+			? html`<d2l-button-subtle class="d2l-empty-state-action" @click=${this._handleActionClick}  h-align="text" text=${this.actionText} slim></d2l-button-subtle>`
```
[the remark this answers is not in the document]
oelhanafey: Thanks, I made this change. Similarly, I also realized that most of my empty state component calls suffered from this same issue and would often span 150-200 characters for a line so I tried to split most of those calls up on to multiple lines for readability.

https://github.com/BrightspaceUI/core/blob/f28cd191df2a8b999d86195b7761b21b50ad0e43/components/empty-state/README.md?plain=1#L101

I thought this would also make the documentation more readable but I am unsure if this would appear as expected in the Daylight code blocks. It didn't seem to render properly on the demo page which is why I did not make those changes there.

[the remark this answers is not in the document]
oelhanafey: Thanks for checking!

margaree (approved):

PR #2592, opened 2022-06-29
[ignore for now] fix: tag-list: initial work to fix flicker

margaree:
The problem I am seeing is that when `d2l-filter-tags` is used in Nova, which wraps `filter-tags` in a flex box in order to right align, we need to remove the `width: 100%` in order for the flex to actually work, but then there is flicker caused by continuous re-triggering of the ResizeObserver when a certain number of filters are hit which could be grouped into the +X more hidden group.

This updates `tag-list` to use the `parentNode` for the `ResizeObserver` or the wrapping component's `parentNode`. The issue seems to be basically when the parent is a non-block element that doesn't set the width at 100%.

I modified the filter-tags demo in order to show the problem.

To do:
- [x] ensure visual diffs pass or change in an expected way
- [ ] test with user-profile-card-tag-list-item (including consistent-eval usage)
- [ ] test Daylight tag list and filter tags examples

### DIFF
--- a/components/filter/demo/filter-tags.html
+++ b/components/filter/demo/filter-tags.html
@@ -11,6 +11,11 @@
 			import '../filter-dimension-set-value.js';
 			import '../filter-tags.js';
 		</script>
+		<style>
+			d2l-demo-snippet {
+				max-width: 1400px;
+			}
+		</style>
 	</head>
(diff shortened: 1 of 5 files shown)


margaree:
Closing in favour of alternative PR

PR #2596, opened 2022-06-29
fix: Have menu wait before checking the role of its menu items

svanherk:
Using both `LocalizeDynamicMixin` and `MenuItemMixin` in a component is resulting in the `menu` trying to get its menu-items before their `role` attribute has been set (because of the delay in updating caused by `LocalizeDynamicMixin`).  There's a number of ways we could try to solve this (changes to how localizing works, using a MutationObserver in `menu`, adding a way for menu items to re-trigger the menu initializing code, etc.) This PR first checks that all the potential menu items have updated before checking their `role` attribute.

### DIFF
--- a/components/menu/menu.js
+++ b/components/menu/menu.js
@@ -234,9 +234,7 @@ class Menu extends FocusVisiblePolyfillMixin(ThemeMixin(HierarchicalViewMixin(Li
 			items.unshift(returnItem);
 		}
 		// Wait for menu items to have their role attribute set
-		await Promise.all(items.map(async(item) => {
-			await item.updateComplete;
-		}));
+		await Promise.all(items.map(item => item.updateComplete));
 		return items.filter((item) => {
 			const role = item.getAttribute('role');
 			return (role === 'menuitem' || role === 'menuitemcheckbox' || role === 'menuitemradio' || item.tagName === 'D2L-MENU-ITEM-RETURN');
--- a/components/menu/menu.js
+++ b/components/menu/menu.js
@@ -224,7 +224,7 @@ class Menu extends FocusVisiblePolyfillMixin(ThemeMixin(HierarchicalViewMixin(Li
 		return this.shadowRoot && this.shadowRoot.querySelector('d2l-menu-item-return');
 	}
 
-	_getMenuItems() {
+	async _getMenuItems() {
 		const slot = this.shadowRoot && this.shadowRoot.querySelector('slot');
 		if (!slot) return;
 		const items = slot.assignedNodes({ flatten: true }).filter((node) => node.nodeType === Node.ELEMENT_NODE);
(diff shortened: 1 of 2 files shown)


components/menu/menu.js
```diff
@@ -233,6 +233,10 @@ class Menu extends FocusVisiblePolyfillMixin(ThemeMixin(HierarchicalViewMixin(Li
 		if (returnItem) {
 			items.unshift(returnItem);
 		}
+		// Wait for menu items to have their role attribute set
+		await Promise.all(items.map(async(item) => {
+			await item.updateComplete;
```
bearfriend, line 238: Is this inner await necessary?

components/menu/menu.js
```diff
@@ -233,6 +233,10 @@ class Menu extends FocusVisiblePolyfillMixin(ThemeMixin(HierarchicalViewMixin(Li
...
+		}));
```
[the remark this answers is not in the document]
svanherk: Ah nice, much cleaner

eKoopmans (approved):
LGTM, thanks for digging into this @svanherk 😄.

dbatiste (approved):
I think this should be ok.

Maybe potential for a race condition if something is expecting `this._items` to be initialized earlier...?

bearfriend (approved):

svanherk:
> The only reference I was seeing that might be suspect is getMenuType(), which d2l-dropdown-menu needs, but doesn't need until it's opened... I think.

Yeah that _seems_ correct... do you know why that function checks the `role` property, but `_getMenuItems` checks the `role` attribute?

components/menu/test/menu.test.js
```diff
@@ -55,13 +91,16 @@ describe('d2l-menu', () => {
 			elem = await fixture(html`
 				<d2l-menu>
 					<d2l-menu-item id="hidden_a" text="a" hidden></d2l-menu-item>
-					<d2l-menu-item id="a1" text="a"></d2l-menu-item>
+					${delayedMenuItem}
```
svanherk, line 94: Swapping this out makes about ten tests fail if we don't have the fix.

components/menu/test/menu.test.js
```diff
@@ -55,13 +91,16 @@ describe('d2l-menu', () => {
...
 					<d2l-menu-item id="b1" text="b" disabled></d2l-menu-item>
 					<d2l-menu-item id="a2" text="a"></d2l-menu-item>
 					<d2l-menu-item id="c1" text="C"></d2l-menu-item>
 					<d2l-menu-item id="d1" text="d"></d2l-menu-item>
 				</d2l-menu>
 			`);
+			const delayed = elem.querySelector('#a1');
+			await delayed.updateComplete;
+			await nextFrame();
 			await nextFrame();
```
svanherk, line 104: This is not great, but seems stable locally.  We'll see what CI thinks.  One `nextFrame` makes the tests flaky - maybe a red flag about this change or about how the localize mixin works in general

PR #2599, opened 2022-06-30
test specifically for slow items

dlockhart:
I was thinking more of a test that specifically targeted this one case, vs. adding it to the focus tests. I think it would let you simplify things a bit... was also able to remove a lot of the `nextFrame()` stuff (I think, unless this is flaky in CI).


svanherk:
Also good with this!  Definitely cleaner.  Weirdly, I couldn't get the flow to work properly and actually cause a delay unless I used the `unsafeStatic` wrapper to make the `_ready` property actually do something

svanherk (approved):

PR #2602, opened 2022-06-30
(Help) Tooltip > Various tweaks and changes to the Tooltip documentation page

dcostiuc:
Various tweaks and changes to the Tooltip documentation page, based on discussions with Jeff and Margaree.

### DIFF
--- a/components/tooltip/README.md
+++ b/components/tooltip/README.md
@@ -19,6 +19,7 @@ Tooltips display additional information when users focus or hover on a point of 
     }, 20);
   });
 </script>
+
 <d2l-button id="tooltip-button">Hover here</d2l-button>
 <d2l-tooltip id="tooltip" for="tooltip-button">
   Tooltip message example
@@ -77,16 +78,18 @@ The `d2l-tooltip` component is used to display additional information when users
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/inputs/input-text.js';
-  import '@brightspace-ui/core/components/tooltip/tooltip.js';  <!-- docs: start hidden content --> 
+  import '@brightspace-ui/core/components/tooltip/tooltip.js';
+</script>
+<!-- docs: start hidden content --> 
+<script>
   window.addEventListener('load', function () {
     setTimeout(function() {
       const tooltip = document.querySelector('d2l-tooltip[for="tooltip-error"]');
       tooltip.showing = true;
     }, 20);
   });
-<!-- docs: end hidden content -->
 </script>
-
+<!-- docs: end hidden content -->
 <d2l-input-text placeholder="Hover for Error" id="tooltip-error" aria-invalid="true" label="Text Input" label-hidden style="max-width:250px;"></d2l-input-text>
 <d2l-tooltip for="tooltip-error" state="error">
 	Your error message will display here
@@ -136,16 +139,18 @@ In the following example to constrain the tooltip to the dashed boundary we can 
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/button/button.js';
-  import '@brightspace-ui/core/components/tooltip/tooltip.js'; <!-- docs: start hidden content --> 
+  import '@brightspace-ui/core/components/tooltip/tooltip.js';
+</script>
+<!-- docs: start hidden content --> 
+<script>
   window.addEventListener('load', function () {
     setTimeout(function() {
       const tooltip = document.querySelector('d2l-tooltip[for="tooltip-boundary"]');
       tooltip.showing = true;
     }, 20);
   });
-<!-- docs: end hidden content -->
 </script>
-
+<!-- docs: end hidden content -->
 <style>
   .boundary {
     border: 1px dashed #cdd5dc;
@@ -191,16 +196,18 @@ The `d2l-tooltip-help` component is used to display additional information when 
 <!-- docs: demo live name:d2l-tooltip-help autoSize:false size:medium -->
 ```html
 <script type="module">
-  import '@brightspace-ui/core/components/tooltip/tooltip-help.js'; <!-- docs: start hidden content --> 
+  import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
+</script>
+<!-- docs: start hidden content --> 
+<script>
   window.addEventListener('load', function () {
     setTimeout(function() {
       const tooltip = document.querySelector('d2l-tooltip-help[text="Helpful label"]:not(#help-tooltip)').shadowRoot.querySelector('d2l-tooltip');
       tooltip.showing = true;
     }, 20);
   });
-<!-- docs: end hidden content -->
 </script>
-
+<!-- docs: end hidden content -->
 <p class="d2l-body-compact">
   This is some sample text.
   <d2l-tooltip-help text="Helpful label">Contents should elaborate on the label (be short and concise)</d2l-tooltip-help>
--- a/components/tooltip/README.md
+++ b/components/tooltip/README.md
@@ -24,15 +24,12 @@ Tooltips display additional information when users focus or hover on a point of 
   Tooltip message example
 </d2l-tooltip>
 
-<d2l-tooltip-help id="help-tooltip" text="Helpful label">
-  Contents should elaborate on the label (be short and concise)
+<d2l-tooltip-help id="help-tooltip" text="Hover here">
+  Tooltip message example
 </d2l-tooltip-help>
 ```
-<!-- docs: start hidden content -->
-TODO: Look more into this bug and eventually fix it
-<!-- docs: end hidden content -->
-Note: there is a known defect with tooltips where they are sometimes rendered offset by some amount unexpectedly, the first time that they are shown.
 
+Note: there is a [known defect](https://rally1.rallydev.com/#/?detail=/defect/641656338755&fdp=true) within the Daylight site for tooltips, where they are sometimes rendered offset by some amount unexpectedly.
 
 ## Accessibility
 
--- a/components/tooltip/README.md
+++ b/components/tooltip/README.md
@@ -7,11 +7,15 @@ Tooltips display additional information when users focus or hover on a point of 
 <script type="module">
   import '@brightspace-ui/core/components/button/button.js';
   import '@brightspace-ui/core/components/tooltip/tooltip.js';
+  import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
 
   window.addEventListener('load', function () {
     setTimeout(function() {
       var tooltip = document.querySelector('#tooltip');
+      var helpTooltip = document.querySelector('#help-tooltip').shadowRoot.querySelector('d2l-tooltip');
+
       tooltip.showing = true;
+      helpTooltip.showing = true;
     }, 20);
   });
 </script>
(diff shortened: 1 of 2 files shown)


components/tooltip/README.md
```diff
@@ -7,19 +7,32 @@ Tooltips display additional information when users focus or hover on a point of
...
 <d2l-tooltip id="tooltip" for="tooltip-button">
   Tooltip message example
 </d2l-tooltip>
+
+<d2l-tooltip-help id="help-tooltip" text="Helpful label">
+  Contents should elaborate on the label (be short and concise)
+</d2l-tooltip-help>
 ```
+TODO: Look more into this bug and fix it
```
dcostiuc, line 32: I confirmed locally that this doesn't show up on the rendered paged.

components/tooltip/README.md
```diff
@@ -7,19 +7,32 @@ Tooltips display additional information when users focus or hover on a point of
...
+<!-- docs: end hidden content -->
+Note: there is a known defect with tooltips where they are sometimes rendered offset by some amount unexpectedly, the first time that they are shown.
```
dcostiuc, line 34: Can read [here ](https://d2l.slack.com/archives/C03H8KH2FKN/p1656618501840279?thread_ts=1656085775.322529&cid=C03H8KH2FKN)for more context.

components/tooltip/README.md
```diff
@@ -67,7 +80,13 @@ The `d2l-tooltip` component is used to display additional information when users
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/inputs/input-text.js';
-  import '@brightspace-ui/core/components/tooltip/tooltip.js';
+  import '@brightspace-ui/core/components/tooltip/tooltip.js';  <!-- docs: start hidden content --> window.addEventListener('load', function () {
```
dcostiuc, line 83: Note: starting `<!-- docs: start hidden content -->` and even actual code on the same line as the import is done intentionally.

In my local testing, I would find that doing it as 

```js
import '@brightspace-ui/core/components/tooltip/tooltip.js';  
window.addEventListener('load', function () {
```
would cause a newline (whitespace) to render on the page, making that look at bit awkward/unprofessional. Since this code is hidden anyway, I figured this is a reasonable workaround for still having any live demo code displayed in a nice way.

components/tooltip/README.md
geurts: The defect occurs all the time, not just the first time they're shown (at least for me). So might be good to remove that part.

components/tooltip/README.md
```diff
@@ -163,15 +189,21 @@ The `d2l-tooltip-help` component is used to display additional information when
...
+      tooltip.showing = true;
+    }, 20);
+  });
+<!-- docs: end hidden content -->
 </script>
 
-<p class="d2l-body-standard">
+<p class="d2l-body-compact">
   This is some sample text.
-  <d2l-tooltip-help text="Helpful label" inherit-font-style>Contents should elaborate on the label (be short and concise)</d2l-tooltip-help>
```
dcostiuc, line 174: After discussing with Margaree and Jeff, we decided we didn't want this demo to override the default styling of the opener, even if the text next to it is styled differently

components/tooltip/README.md
```diff
@@ -163,15 +189,21 @@ The `d2l-tooltip-help` component is used to display additional information when
...
 <!-- docs: end best practices -->
 
-<!-- docs: demo live name:d2l-tooltip-help autoSize:false size:small -->
+<!-- docs: demo live name:d2l-tooltip-help autoSize:false size:medium -->
 ```html
 <script type="module">
-  import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
+  import '@brightspace-ui/core/components/tooltip/tooltip-help.js'; <!-- docs: start hidden content --> window.addEventListener('load', function () {
+    setTimeout(function() {
+      var tooltip = document.querySelector('d2l-tooltip-help[text="Helpful label"]:not(#help-tooltip)').shadowRoot.querySelector('d2l-tooltip');
```
dcostiuc, line 197: Perhaps a bit overkill since I could just add an `id` and select that.
However, since _this_ code is hidden, whereas the below other code is shown on the rendered page, I figured it is both unnecessary and potentially a bit confusing to have this help tooltip have a random `id` on it. 

If anyone has any thoughts/disagrees, feel free to let me know.

components/tooltip/README.md
dcostiuc: Discussed and clarified this more [here](https://d2l.slack.com/archives/C03H8KH2FKN/p1656947519933479?thread_ts=1656085775.322529&cid=C03H8KH2FKN).

components/tooltip/README.md
```diff
@@ -7,20 +7,30 @@ Tooltips display additional information when users focus or hover on a point of
 <script type="module">
   import '@brightspace-ui/core/components/button/button.js';
   import '@brightspace-ui/core/components/tooltip/tooltip.js';
+  import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
 
   window.addEventListener('load', function () {
     setTimeout(function() {
       var tooltip = document.querySelector('#tooltip');
+      var helpTooltip = document.querySelector('#help-tooltip').shadowRoot.querySelector('d2l-tooltip');
```
dlockhart, line 15: We should be using `const` instead of `var`, even in README examples.

dcostiuc: Yeah I agree, I thought it a bit odd that the existing code for that was using `var`. Assumed perhaps there was a compatibility reason or something (needed to use ES5 etc). 

Tried it out right now with `const`  locally and everything seems to still work fine, so updating this.

components/tooltip/README.md
eKoopmans: Why is the code hidden in the first place? Is the code not necessary to reproduce the state that's being shown in the demo?

components/tooltip/README.md
```diff
@@ -7,20 +7,30 @@ Tooltips display additional information when users focus or hover on a point of
 <script type="module">
   import '@brightspace-ui/core/components/button/button.js';
   import '@brightspace-ui/core/components/tooltip/tooltip.js';
+  import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
 
   window.addEventListener('load', function () {
     setTimeout(function() {
-      var tooltip = document.querySelector('#tooltip');
+      const tooltip = document.querySelector('#tooltip');
+      const helpTooltip = document.querySelector('#help-tooltip').shadowRoot.querySelector('d2l-tooltip');
```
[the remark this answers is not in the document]
dlockhart: Agreed yeah. Doesn't look like it, but I think it would be worth adding a `showing` attribute to it that maps down to `d2l-tooltip`, even if it's only helpful for demo/test purposes.

components/tooltip/README.md
dcostiuc: This code isn't relevant to the demos that are shown, at least in my opinion. 

It's essentially there to allow for all the tooltips to automatically open when the consumer visits the Tooltip documentation page. However, normally, you have to hover/click/tab to the opener for the tooltip to show up (unless you specifically set it to `always-on`). So that code is essentially just some temp code to improve the documentation page experience for the consumer.

Ah, I see what you mean. A user might get confused when they see that, on Daylight, the tooltips start off opened, whereas when they use it, they can't reproduce that effect with our provided API.

I suppose another reason why it's hidden is because it would make the demo code that's displayed longer/noisier.

However, do you think we should still show it to the consumer?

components/tooltip/README.md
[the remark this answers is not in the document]
dcostiuc: Yeah, I agree that accessing the `shadowRoot`, and having that show in the README, felt off. `showing` attribute added 👍

components/tooltip/tooltip-help.js
```diff
@@ -44,6 +48,7 @@ class TooltipHelp extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
 			#d2l-tooltip-help-text {
 				background: none;
 				border: none;
+				cursor: text;
```
dcostiuc, line 51: This is a regression fix. Originally we had this automatically with a `span` opener, but it changed to the regular cursor after we started using a `button` for the opener. 

Let me know if I should revert the commit and make a separate PR for it.

components/tooltip/README.md
```diff
@@ -163,15 +193,24 @@ The `d2l-tooltip-help` component is used to display additional information when
...
 <script type="module">
   import '@brightspace-ui/core/components/tooltip/tooltip-help.js';
 </script>
-
-<p class="d2l-body-standard">
+<!-- docs: start hidden content --> 
+<script>
+  window.addEventListener('load', function () {
+    setTimeout(function() {
+      const tooltip = document.querySelector('d2l-tooltip-help[text="Helpful label"]:not(#help-tooltip)');
```
dcostiuc, line 205: So I realized that we don't actually need to stress with selectors here, and can literally just have 
`const tooltip = document.querySelector('d2l-tooltip-help');` or 
`const tooltip = document.querySelector('d2l-tooltip');` for all the demos and still have this code work (the tooltips auto-show on page load). 

Do we want to do that? Or would it be potentially confusing for someone reading the raw README?

components/tooltip/README.md
```diff
@@ -69,7 +80,16 @@ The `d2l-tooltip` component is used to display additional information when users
   import '@brightspace-ui/core/components/inputs/input-text.js';
   import '@brightspace-ui/core/components/tooltip/tooltip.js';
 </script>
-
+<!-- docs: start hidden content --> 
+<script>
+  window.addEventListener('load', function () {
```
dlockhart, line 85: If I'm nitting on the code samples/demos, these should all use arrow functions `() => {` but there's likely lots of existing stuff that isn't so 🤷 .

components/tooltip/README.md
dlockhart: That's because they're all iframed ya, so not colliding with each other?

components/tooltip/tooltip-help.js
```diff
@@ -25,6 +25,10 @@ class TooltipHelp extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
 			 * @type {'top'|'bottom'|'left'|'right'}
 			 */
 			position: { type: String },
+			/**
+			 * @ignore
+			 */
+			showing: { type: Boolean, reflect: true },
```
dlockhart, line 31: This should default to `false` in the constructor, otherwise it's `undefined` which is still `false`-y but not as nice. Don't you love JavaScript?!

components/tooltip/tooltip-help.js
```diff
@@ -97,7 +102,7 @@ class TooltipHelp extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
 			<button id="d2l-tooltip-help-text" class="${classMap(classes)}">
 				${this.text}
 			</button>
-			<d2l-tooltip for="d2l-tooltip-help-text" delay="0" offset="13" position="${ifDefined(this.position)}">
+			<d2l-tooltip for="d2l-tooltip-help-text" delay="0" offset="13" position="${ifDefined(this.position)}" showing="${ifDefined(this.showing)}">
```
dlockhart, line 105: Boolean attributes need the `?`:

```suggestion
			<d2l-tooltip for="d2l-tooltip-help-text" delay="0" offset="13" position="${ifDefined(this.position)}" ?showing="${ifDefined(this.showing)}">
```

components/tooltip/README.md
dcostiuc: > but there's likely lots of existing stuff that isn't so 🤷 .

Yeah so I guess your point is consistency, huh?

I mean, we could always just use arrow functions going forward at least. 
Perhaps we could add a chore-y Dragonslayer item for updating any other such code snippets to use arrow functions?

dlockhart: Yeah don't worry about it for now. I might do a search & replace tomorrow.

dlockhart (approved):
LGTM!

components/tooltip/README.md
dcostiuc: Yup [looks like it](https://github.com/BrightspaceUI/documentation/blob/main/pages/_components/demo-snippet.js#L67-L68). 

In terms of readability (someone reading the raw README), are there any concerns? 
Or should I just make the changes to simplify the selector queries?

dlockhart: I'd go for simplicity of the selectors and wouldn't worry about raw README readers haha.

PR #2605, opened 2022-07-04
Bump MathJax to 3.2.1 and simplify shadowDOM rendering logic

grant-cleary:
This is the first of a couple changes I'm planning to make to try to simplify the way the html-block control renders and (hopefully) actually use Lit fully to do said rendering.

This first step is to clean up the MathJax renderer by bumping MathJax to 3.2.1, which seems to solve the vast majority of the problems we've encountered trying to get it working in shadowDOM. The only change I had to make was to disable their new adaptiveCSS feature (a perf optimization that allows MathJax to generate its CSS stylesheets on the fly at render time) because we have to clone that stylesheet into the shadowDOM of wherever we're working in to get it rendering properly. Their adaptive stuff interferes with that because there's no guarantee the stylesheet is complete by the time we need to clone it.


helpers/mathjax.js
```diff
@@ -60,32 +60,25 @@ export class HtmlBlockMathRenderer {
...
+			elm.setAttribute('style', 'display: block; height: 0.5rem;');
+		});
+
+		await window.MathJax.startup.promise;
+		await window.MathJax.typesetPromise([elem]);
 
-		const inner = elem.innerHTML.replace(/<mspace linebreak="newline">/gi, `<mspace linebreak="newline" style="${lineBreakStyle}">`);
+		// If we're opting out of deferred rendering, we can rely
+		// on the global MathJax styles for rendering.
+		if (options.noDeferredRendering) return elem;
```
grant-cleary, line 72: Hmm. I suspect I need to do something more clever here. If we're in a nested shadowDOM this will break because the styles won't get set properly.

helpers/mathjax.js
```diff
@@ -60,32 +60,25 @@ export class HtmlBlockMathRenderer {
...
-		temp.attachShadow({ mode: 'open' });
-		temp.shadowRoot.innerHTML = `<div><mjx-doc><mjx-head></mjx-head><mjx-body>${inner}</mjx-body></mjx-doc></div>`;
+		const styleElm = window.MathJax.chtmlStylesheet().cloneNode(true);
+		styleElm.id = 'd2l-mathjax-styles';
 
-		elem.appendChild(temp);
-		await window.MathJax.startup.promise;
-		window.MathJax.typesetShadow(temp.shadowRoot);
+		if (!elem.querySelector(`#${styleElm.id}`)) {
+			elem.appendChild(styleElm);
```
grant-cleary, line 78: My gut feeling is that we want to append this to the nearest shadowRoot or the document root directly. Thoughts?

grant-cleary:
Visual diff failures are slight layout differences in some of the rendered math, presumably because of the MathJax upgrade. Going to merge in those new goldens.

helpers/mathjax.js
dbatiste: Yeah I think it would have to be appended to the root of the DOM scope in which the math will be rendered, in order for it to work in the case where it's not the global scope. 

I don't feel super great about the idea of appending a stylesheet to arbitrary shadowRoots though. It would effectively be modifying the DOM outside of its normal render - would we encounter problems where Lit could blow away that stylesheet? 

Ideally such stylesheet would be constructed and adopted at the document level, and as necessary within shadowRoots. [Constructable Stylesheets](https://web.dev/constructable-stylesheets/) - Lit handles this for us without us needing to worrying about browser support.

Hence the desire to keep all the rendering within `d2l-html-block`'s shadowDOM / local scope.

Also, is the stylesheet always the same? For instance, if there were two `d2l-html-block` components with `noDeferredRendering`, the one stylesheet would work from them both, even though their math could be vastly different?

grant-cleary: That's a good point, yeah. I mean I think it comes down to whether we're expecting `noDeferredRendering` to be used that often. If it's limited to the LMS where we can broadly depend on MathJax being able to use its global stylesheet, that's probably fine? Hard to say for sure though. That use case has always been strange.

And yes, with adaptiveCSS disabled, the stylesheets are always the same. So all it takes is for one to load in the `noDeferredRendering` case and we're good. MathJax does sprout its styles at the document level whether you like it or not, so in _most_ cases I suspect we'll just be covered by that.

grant-cleary: As for why I'm sprouting it off `elem` (for context), right now `elem` is just a temp element created for stamping purposes. In my change in how this rendering works, however, `elem` will be the render container itself. So realistically this amounts to sprouting the styles inside the render container div.

grant-cleary: The `noDeferredRendering` case just bothers me. As the code is currently written, we'll get potentially weird results if another web component loads solely non-deferred html-blocks in its shadowDOM. On the other hand, if we just blindly append styles the way we're doing here, we could end up with lots of duplicate stylesheets spread around the same DOM scope, which also doesn't feel great.

dbatiste: Yeah it bothers me as well. If we're confident that we will only support `noDeferredRendering` in the global scope then I think we're "ok". If any tool that relies on it is ever converted to an app web component, we'd need to find a way to address that.

grant-cleary: That sounds fine to me. I can make sure this is tested well in places that could be problem areas and we can always opt to cross this particular bridge later.

Given that, is the current suggested implementation okay, or would you prefer to see it implemented in some other fashion?

dbatiste: Looks fine.

dbatiste (approved):

helpers/mathjax.js
grant-cleary: Cool beans. T-10s until I break everything.

PR #2607, opened 2022-07-04
fix: Remove focus-visible from link

eKoopmans:
Reverting #2478 (except for a few fixes that weren't related to `focus-visible`). This fixes an issue where components using link styles (like Breadcrumbs) were no longer getting focus styling, because they don't explicitly load the `focus-visible` polyfill.

Rally: https://rally1.rallydev.com/#/?detail=/defect/640887963745&fdp=true


dlockhart (approved):
LGTM

PR #2608, opened 2022-07-04
fix: filter-tags and tag-list: remove 100% width on filter-tags and fix width flicker

margaree:
The problem I am seeing is that when `d2l-filter-tags` is used in Nova, which wraps filter-tags in a flex box in order to right align, we need to remove the `width: 100%` in order for the flex to actually work, but then there is flicker caused by continuous re-triggering of the ResizeObserver when a certain number of filters are hit which could be grouped into the +X more hidden group (it's a bit of a chicken and egg problem).

Alternative solution to https://github.com/BrightspaceUI/core/pull/2592 - let me know thoughts on one vs the other, if there's something else I should try, and if there's any other considerations I may have missed.

This solution uses `getOffsetParent` in order to get the width of the closest positioned parent and use that in the `ResizeObserver` for getting the available width and actually triggering the `ResizeObserver`. If the body is the result of `getOffsetParent` then we will use the `parentNode` to actually measure the available width, since body width and actual available width can differ (not sure if this is best/correct choice here). 

To do:
- [x] ensure visual diffs pass or change in an expected way
- [X] test with user-profile-card-tag-list-item (including consistent-eval usage) - tested at regular two-panel width and mobile width
- [X] test Daylight tag list and filter tags examples - the example was updated and seems to work well when resized
- [X] test with Nova backport - particular issue was seen on the Requests page. Can no longer reproduce.


components/filter/README.md
```diff
@@ -222,47 +222,54 @@ A tag-list allowing the user to see (and remove) the currently applied filters.
   import '@brightspace-ui/core/components/filter/filter-tags.js';
 </script>
 <style>
-	d2l-filter-tags {
-		max-width: calc(100% - 300px);
-	}
-
-	@media(max-width: 600px) {
-      d2l-filter-tags {
+  .filter-wrapper {
```
margaree, line 225: Updated the example to use flex rather than a set max-width, which seems more ideal/like a potential usecase based on designs.

components/filter/README.md
```diff
@@ -222,47 +222,54 @@ A tag-list allowing the user to see (and remove) the currently applied filters.
...
-    <d2l-filter-dimension-set-value text="Option 1 - 4" key="4"></d2l-filter-dimension-set-value>
-  </d2l-filter-dimension-set>
-  <d2l-filter-dimension-set key="2" text="Dim 2" value-only-active-filter-text>
-    <d2l-filter-dimension-set-value selected text="Option 2 - 1" key="1"></d2l-filter-dimension-set-value>
-    <d2l-filter-dimension-set-value text="Option 2 - 2" key="2"></d2l-filter-dimension-set-value>
-    <d2l-filter-dimension-set-value text="Option 2 - 3" key="3"></d2l-filter-dimension-set-value>
-  </d2l-filter-dimension-set>
-</d2l-filter>
+<div class="filter-wrapper">
+  <d2l-filter-tags filter-ids="core-filter core-filter-2"></d2l-filter-tags>
```
margaree, line 243: Diff below is just indentation

components/filter/demo/filter-tags.html
```diff
@@ -11,6 +11,11 @@
 			import '../filter-dimension-set-value.js';
 			import '../filter-tags.js';
 		</script>
+		<style>
```
margaree, line 14: This file updates the example to one that reproduces the issue seen in Nova. It's no longer seen with the other changes in this PR, but can be seen on main with these changes and with the `width: 100%` removed from `filter-tags`

components/filter/filter-tags.js
```diff
@@ -34,7 +34,6 @@ class FilterTags extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		return [bodyCompactStyles, css`
 			:host {
 				display: inline-block;
-				width: 100%;
```
margaree, line 37: This required that any consumer have this take up 100% of width, which was causing the issue in Nova where it could not be right aligned.

components/tag-list/tag-list.js
```diff
@@ -133,9 +134,13 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		});
 		this._clearButtonResizeObserver.observe(clearButton);
 
-		const container = this.shadowRoot.querySelector('.tag-list-outer-container');
-		this._resizeObserver = new ResizeObserver((e) => requestAnimationFrame(() => this._handleResize(e)));
-		this._resizeObserver.observe(container);
+		const offsetParent = getOffsetParent(this);
+		this._resizeObserver = new ResizeObserver((e) => requestAnimationFrame(() => {
+			if (offsetParent.tagName === 'BODY') this._availableWidth = Math.floor(this.parentNode.offsetWidth);
```
margaree, line 139: I put this `if/else` block here because it's easy to access `offsetParent` here for this check, but it could make sense to do a bit of refactoring and put all the logic back in `_handleResize`

components/filter/test/filter-tags.visual-diff.js
```diff
@@ -64,13 +68,9 @@ describe('d2l-filter-tags', () => {
...
 
-		before(async() => {
-			await page.$eval(selector, async(elem) => {
-				elem.parentNode.style.width = '1500px';
-				elem._showHiddenTags = false;
-				await elem.updateComplete;
-			});
-			await page.waitForTimeout(2000);
+		beforeEach(async() => {
+			await page.reload();
```
margaree, line 72: I don't love doing a page reload before each, but this is the simplest way to reset the components back to their initial state so they aren't dependant on each other from one to the next (i.e., currently test 1 deletes last item, test 2 deletes second item, test 2's visual diff appearance relies on result of test 1)

components/filter/test/filter-tags.visual-diff.js
```diff
@@ -64,13 +68,9 @@ describe('d2l-filter-tags', () => {
 	describe('clearable behavior', () => {
 		const selector = '#basic';
 
-		before(async() => {
-			await page.$eval(selector, async(elem) => {
-				elem.parentNode.style.width = '1500px';
-				elem._showHiddenTags = false;
```
margaree, line 70: Tested this out and this wasn't actually doing anything

components/filter/test/filter-tags.visual-diff.js
margaree: Yup that's a good idea

eKoopmans (approved):

PR #2613, opened 2022-07-05
Prism helper updates

dbatiste:
This PR includes a couple of small updates to prepare for release.

* include Regex as a language that can be formatted
* light and dark theme color updates from Design
* include css-extras language for improved CSS formatting
* include inline-color plugin for color swatches in formatted CSS

With css-extras, we have the ability to be more granular in formatting CSS - specifically we could have special formatting for pseudo-classes, pseudo-elements, combinators, units, ids, n-nth selectors, and css variables. This was briefly considered but decided against for the time-being.


helpers/prism.js
```diff
@@ -28,6 +28,7 @@ codeLanguages.set('plain', 'Plain Text');
 codeLanguages.set('python', 'Python');
 codeLanguages.set('r', 'R');
 codeLanguages.set('racket', 'Racket');
+codeLanguages.set('regex', 'Regex');
```
grant-cleary, line 31: Mm, did we formally decide what we were going to do about languages CM doesn't support? I know for examples like Racket, the thought was "Lisp is probably good enough", but for sure there's no equivalent like that for regex.

helpers/prism.js
```diff
@@ -255,19 +256,21 @@ export const codeStyles = css`
...
-	pre[class*="language-"].line-numbers > code {
+	pre[class*="language-"].d2l-code.line-numbers > code {
 		position: relative;
 		white-space: inherit;
 	}
 
-	.line-numbers .line-numbers-rows {
-		border-right: 1px solid var(--d2l-code-line-numbers-separator);
+	.d2l-code.line-numbers > code > .line-numbers-rows {
+		background-color: var(--d2l-code-line-numbers-background);
```
dbatiste, line 271: I didn't call it out in the PR description, but the design changed for the line-number formatting from:
image
to:
image

helpers/prism.js
grant-cleary: Yeah, plain text is certainly fine. I know there was some concern about trying to prioritize consistency over breadth. I don't personally care and I think supporting more stuff in Prism than we do in CM seems fine, but... I can't remember if that's where we actually landed.

dbatiste: Circling back to this after discussion with designers, if CodeMirror does not support the language, we will format it as plain text in CodeMirror, while Prism will apply the expected formatting. We will not exclude those languages from the list.  If CodeMirror has reasonable formatting based on a different language, I think we should try to use it (ex. Arduino -> C++/C/C-like).

grant-cleary: Yeah, seems good to me!

dlockhart (approved):

PR #2615, opened 2022-07-05
fix: filter: Add demo for "Filter Dimension: Set Value" and fix links

margaree:
The links don't work in the README and also don't work in Daylight. With this change they will at least work in Daylight.

This adds a simple demo for `Filter Dimension: Set Value` and hides its properties table. It was originally not included (see https://github.com/BrightspaceUI/core/pull/1776#discussion_r722613522 for context) but my feeling is at this point it's useful for consistency in how we communicate properties and for having the single source of truth for our properties.


oelhanafey (approved):
LGTM

PR #2618, opened 2022-07-06
Add autofocus feature to dialogs

KaiPrince:
[US140739](https://rally1.rallydev.com/#/?detail=/userstory/641721547587&fdp=true): UI Core dialog - Add autofocus feature

image

image

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -3,7 +3,7 @@ import '../../helpers/viewport-size.js';
 import { allowBodyScroll, preventBodyScroll } from '../backdrop/backdrop.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
-import { forceFocusVisible, getComposedActiveElement, getNextFocusable, tryApplyFocus } from '../../helpers/focus.js';
+import { forceFocusVisible, getComposedActiveElement, getNextFocusable, isFocusable, tryApplyFocus } from '../../helpers/focus.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { html } from 'lit';
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		}
 	}
 
+	_findAutofocusElement(node) {
+		if (this._useNative) {
+			// Do not override native autofocus attribute implementation
+			return null;
+		}
+
+		if (!node.querySelector('slot')) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
+
+		const content = node.querySelector('slot').assignedElements({ flatten: true });
+		const el = content.find(el => el.autofocus)
+			?? content.find(el => el.querySelector('[autofocus]'))?.querySelector('[autofocus]');
+		return el;
+	}
+
 	_focusFirst() {
 		if (!this.shadowRoot) return;
 		const content = this.shadowRoot.querySelector('.d2l-dialog-content');
 		if (content) {
+			const firstAutofocus = this._findAutofocusElement(content);
+			if (isFocusable(firstAutofocus, false, false)) {
+				firstAutofocus.focus();
+				return;
+			}
+
 			const firstFocusable = getNextFocusable(content);
 			if (isComposedAncestor(this.shadowRoot.querySelector('.d2l-dialog-inner'), firstFocusable)) {
 				forceFocusVisible(firstFocusable);
--- a/components/dialog/test/dialog.test.js
+++ b/components/dialog/test/dialog.test.js
@@ -37,6 +37,27 @@ describe('d2l-dialog', () => {
 			expect(getComposedActiveElement()).to.equal(button);
 		});
 
+		it('should focus on an autofocus element', async() => {
+			const el = await fixture(html`<d2l-dialog opened><p autofocus tabindex="-1">focus</p><button>focus</button></d2l-dialog>`);
+			const paragraph = el.querySelector('p');
+			await oneEvent(el, 'd2l-dialog-open');
+			expect(getComposedActiveElement()).to.equal(paragraph);
+		});
+
+		it('should focus on a descendant autofocus element', async() => {
+			const el = await fixture(html`<d2l-dialog opened><div><p autofocus tabindex="-1">focus</p></div><button>focus</button></d2l-dialog>`);
+			const paragraph = el.querySelector('p');
+			await oneEvent(el, 'd2l-dialog-open');
+			expect(getComposedActiveElement()).to.equal(paragraph);
+		});
+
+		it('should not focus on an autofocus element that had not been made focusable', async() => {
+			const el = await fixture(html`<d2l-dialog opened><div><p autofocus>focus</p></div><button>focus</button></d2l-dialog>`);
+			const button = el.querySelector('button');
+			await oneEvent(el, 'd2l-dialog-open');
+			expect(getComposedActiveElement()).to.equal(button);
+		});
+
 	});
 
 });
--- a/components/dialog/test/dialog.test.js
+++ b/components/dialog/test/dialog.test.js
@@ -14,7 +14,7 @@ describe('d2l-dialog', () => {
 
 	});
 
-	describe.skip('focus management', () => {
+	describe('focus management', () => {
 
 		it('should focus on close button if no focusable elements inside', async() => {
 			const el = await fixture(html`<d2l-dialog opened>not focusable</d2l-dialog>`);


components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+
+		if (!node.querySelector('slot')) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
+
+		const content = node.querySelector('slot').assignedElements({ flatten: true });
+		const el = content.find(el => el.autofocus)
+			?? content.find(el => el.querySelector('[autofocus]'))?.querySelector('[autofocus]');
+		return el;
```
KaiPrince, line 194: I had a lot of trouble with this bit. `querySelector('[autofocus]')` will not traverse `slot`s. `getComposedChildren` also did not work.

components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		}
+
+		if (!node.querySelector('slot')) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
+
+		const content = node.querySelector('slot').assignedElements({ flatten: true });
+		const el = content.find(el => el.autofocus)
+			?? content.find(el => el.querySelector('[autofocus]'))?.querySelector('[autofocus]');
```
KaiPrince, line 193: This part is messy. Because `assignedElements` returns an array instead of a node, I have to handle the case of the immediate (depth=0) element being autofocus-able separately from its children.

e.g.
```
[
  <p autofocus> ... </p>
]
```
won't have a querySelector method.

whereas:
```
[
  <div>
    <p autofocus> ... </p>
  </div>
]
```
will need to use querySelector.

components/dialog/test/dialog.test.js
```diff
@@ -14,7 +14,7 @@ describe('d2l-dialog', () => {
 
 	});
 
-	describe.skip('focus management', () => {
+	describe('focus management', () => {
```
KaiPrince, line 17: I'm not sure why this was disabled. I enabled it and all tests pass.

KaiPrince:
Ah, what the heck? I had all tests passing :(

components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		}
+
+		if (!node.querySelector('slot')) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
+
+		const content = node.querySelector('slot').assignedElements({ flatten: true });
+		const el = content.find(el => el.getAttribute('autofocus') === '' || el.getAttribute('autofocus') === 'true')
+			?? content.find(el => el.querySelector('[autofocus]'))?.querySelector('[autofocus]');
```
KaiPrince, line 193: This is messy. I am open to suggestions in how to clean it up.

components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+	_findAutofocusElement(node) {
+		if (this._useNative) {
+			// Do not override native autofocus attribute implementation
+			return null;
+		}
+
+		if (!node.querySelector('slot')) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
```
KaiPrince, line 189: I don't really like that this is here, because it limits future use cases. I have left it in because it currently suits our needs and there is no use case of `d2l-dialog` that wouldn't pass this guard. The alternative is to add extra logic for composing slots conditionally.

components/dialog/test/dialog.test.js
KaiPrince: Ah everything is failing on webkit. I see. I will disable again.

KaiPrince: Oh I see: https://github.com/BrightspaceUI/core/pull/1351#issuecomment-843579774

components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		}
 	}
 
+	_findAutofocusElement(node) {
+		if (this._useNative) {
+			// Do not override native autofocus attribute implementation
+			return null;
+		}
+
+		if (!node.querySelector('slot')) {
```
dbatiste, line 186: I think I would optimize this a little to only get the slot once.

``` javascript
const slot = node.querySelector('slot');
if (!slot) return null;

const content = slot.assignedElements({ flatten: true });
...
```

I think it should be possible to use `el.hasAttribute('autofocus')`.

components/dialog/test/dialog.test.js
dbatiste: Yeah not sure... looks like they've been skipped since @emil-furniss created [in this change](https://github.com/BrightspaceUI/core/commit/cd63c0567faeb364f19cf47e71111537e6b53be6#diff-12b7c6cce35edaf42462efa513d894e23b91ac2d9277617684b570b52a31669d).

components/dialog/dialog-mixin.js
```diff
@@ -177,10 +177,33 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		return el;
+	}
+
 	_focusFirst() {
 		if (!this.shadowRoot) return;
 		const content = this.shadowRoot.querySelector('.d2l-dialog-content');
 		if (content) {
+			const firstAutofocus = this._findAutofocusElement(content);
+			if (isFocusable(firstAutofocus, false, false)) {
+				firstAutofocus.focus();
```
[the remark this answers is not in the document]
KaiPrince: I was thinking about that, but `forceFocusVisible` doesn't allow `tabindex="-1"`

components/dialog/dialog-mixin.js
KaiPrince: Wouldn't `hasAttribute` give a false positive if we get something like `<p autofocus="false">`?

dlockhart: That's not a false positive -- the presence or absence of boolean attributes make them true or false, not the attribute value. `autofocus="false"` is the same as `autofocus` is the same as `autofocus="yes please"` -- the attribute value doesn't matter.

components/dialog/dialog-mixin.js
dbatiste: Yeah this is a little awkward.  `find` doesn't seem to be the right tool for the job here. `filter` would work, but we want to stop searching once we find the element.

My first inclination is to do something like this, as it seems easy to follow, stops as soon as it finds the element, and eliminates the double query.  But yeah, if you can get something nice working for `reduce`, it may be cleaner.

``` javascript
let el;
for (let i = 0; i < content.length; i++) {
    if (content[i].hasAttribute('autofocus')) {
        el = content[i];
        break;
    }
    const temp = content[i].querySelector('[autofocus]');
    if (temp) {
        el = temp;
        break;
    }
}
```

components/dialog/dialog-mixin.js
KaiPrince: Fixed with https://github.com/BrightspaceUI/core/pull/2618/commits/1b91e5b1b0dcf9ef7eb2bcf1711a5f2378e2f53a

components/dialog/dialog-mixin.js
```diff
@@ -177,13 +177,37 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		if (!slot) {
+			// We are in a confirm dialog, autofocus attribute will never be set
+			return null;
+		}
+
+		const content = slot.assignedElements({ flatten: true });
+
+		const searchOneElement = (el) => {
+			const foundElementOrNull = el.hasAttribute('autofocus') ? el : el.querySelector('[autofocus]');
+			return isFocusable(foundElementOrNull, false, false) ? foundElementOrNull : null;
```
KaiPrince, line 196: Using `isFocusable` enforces a `tabindex` of `1` or `0`, or semantic goodlist. This also means that custom elements will be required to add `tabindex` to use `autofocus`.

dbatiste (approved):

components/dialog/dialog-mixin.js
dlockhart: That doesn't sound like something we want then? Maybe we just don't worry about `isFocusable` and if someone puts `autofocus` on something, we focus on it no matter what.

components/dialog/test/dialog.test.js
```diff
@@ -37,6 +37,27 @@ describe('d2l-dialog', () => {
 			expect(getComposedActiveElement()).to.equal(button);
 		});
 
+		it('should focus on an autofocus element', async() => {
+			const el = await fixture(html`<d2l-dialog opened><p autofocus tabindex="-1">focus</p><button>focus</button></d2l-dialog>`);
+			const paragraph = el.querySelector('p');
+			await oneEvent(el, 'd2l-dialog-open');
+			expect(getComposedActiveElement()).to.equal(paragraph);
```
dlockhart, line 44: Do these tests actually pass? If so, we could unskip the parent and just add `it.skip` to the ones that are broken.

components/dialog/dialog-mixin.js
```diff
@@ -177,13 +177,37 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+		};
+
+		let autofocusElement = null;
+		content.some(x => (autofocusElement = searchOneElement(x)));
```
KaiPrince, line 200: Alternatively, I could do:

```suggestion
		for (const el of content) {
			const foundElement = searchOneElement(el);
			if (foundElement) {
				autofocusElement = foundElement;
				break;
			}
		}
```

components/dialog/dialog-mixin.js
KaiPrince: Updated in https://github.com/BrightspaceUI/core/pull/2618/commits/3b75d6c506f9e906e61907dc733635e737af3a1c

dbatiste (approved):

PR #2623, opened 2022-07-07
Removed middle vertical alignment from subtle button span

oelhanafey:
Small alignment change to the subtle button. This will align the text to the baseline and also make it inline with the other button elements.


eKoopmans (approved):

eKoopmans:
I'm assuming the one failed Firefox test is flake, I triggered it to re-run.

dbatiste (approved):

components/filter/test/filter.visual-diff.js
```diff
@@ -107,10 +107,16 @@ describe('d2l-filter', () => {
...
 					await open(page, selector);
-					await page.$eval(selector, async(filter) => {
-						filter.shadowRoot.querySelector('d2l-selection-select-all').shadowRoot.querySelector('d2l-input-checkbox').shadowRoot.querySelector('input').click();
+					await page.$eval(selector, async(filter, type) => {
+						const checkbox = filter.shadowRoot.querySelector('d2l-selection-select-all').shadowRoot.querySelector('d2l-input-checkbox');
+						if ((type === 'press-unselect-all' && checkbox.checked)
+							|| (type === 'press-select-all' && !checkbox.checked)
+						) {
+							// NOT visual-diff retry
+							checkbox.shadowRoot.querySelector('input').click();
```
margaree, line 116: The checkbox was being re-clicked each retry, so this changes it to only click the first time

components/tag-list/test/tag-list.visual-diff.js
```diff
@@ -32,6 +32,12 @@ describe('d2l-tag-list', () => {
 	describe('tag list item style behaviour', () => {
 
 		it('is correct on focus on tag list item', async function() {
+			await page.$eval('#default', async(elem) => {
```
margaree, line 35: This test shows the keyboard tooltip, so if we're retrying we need to make sure that would still get shown

components/tag-list/test/tag-list.visual-diff.js
```diff
@@ -73,6 +79,17 @@ describe('d2l-tag-list', () => {
 					await page.waitForTimeout(500);
 				});
 
+				beforeEach(async() => {
+					await page.$eval(selector, async(elem) => {
```
margaree, line 83: The impact of moving this is that "Show Less" becomes focused in the test `is correct when show more button clicked if applicable`, which seems to be the expected behaviour (based on how it behaves in the `filter-tags` tests). I'm not totally sure why it wouldn't have been focusing before.

components/tag-list/test/tag-list.visual-diff.js
```diff
@@ -112,33 +125,28 @@ describe('d2l-tag-list', () => {
...
-		before(async() => {
-			await page.$eval(selector, async(elem) => {
-				elem.parentNode.style.width = '1200px';
-				elem._showHiddenTags = false;
-				elem._hasShownKeyboardTooltip = true;
-				await elem.updateComplete;
-			});
-			await page.waitForTimeout(2000);
+		beforeEach(async() => {
+			await page.reload();
```
margaree, line 130: There isn't really a great way to do this without a reload. Another option would be to add a new item before each tests get run for them to delete, but then that seems like it could have a bit of risk of failing silently (the before case would be the same as the after)

margaree (commented):
Added some comments to the visual diff tests I updated for more context

components/filter/test/filter.visual-diff.js
eKoopmans: I'm wondering if there might be a more general solution to "only run on first try" - if the window is persistent, then could we set a boolean flag on the window, something like `setupComplete`? Might make sense for it to be per-test, so like:

```js
// In a before block for the whole test:
window.setupComplete = {};

// In an individual test:
if (!window.setupComplete[this.test.fullTitle()] {
  // Do setup, then:
  window.setupComplete[this.test.fullTitle()] = true;
}

Comment (eKoopmans) [reply components/filter/test/filter.visual-diff.js:116]: Not sure whether this is worth the effort ☝️ ... Looking at some of the other changes, it seems like what we really need is cleanup after each test, to reset it to its initial state - or even better, starting fresh each time.

File: components/tag-list/test/tag-list.visual-diff.js (hunk 4 above)
Comment (eKoopmans) [reply components/tag-list/test/tag-list.visual-diff.js:130]: If we're reloading before each, is there still a benefit of adding `#clearable2`?

Comment (margaree) [reply components/tag-list/test/tag-list.visual-diff.js:130]: It's overall not needed, just separates the tests, which would be useful if it was switched to the second option.

File: components/filter/test/filter.visual-diff.js (hunk 1 above)
Comment (margaree) [reply components/filter/test/filter.visual-diff.js:116]: Yup that could work, though agreed that it might not be worth the effort - the tests need an overhaul.

**Review by eKoopmans — APPROVED** [review]

**Comment by eKoopmans** [comment]
> Arabic in order to activate rtl

Right - it's possible the RTL test itself wants to use Arabic text, hadn't thought of that. Anyway I was just investigating flake, but I'm guessing if you just re-run the visual diff it'll pass.

File: components/filter/test/filter.visual-diff.js [hunk 5]
```diff
@@ -226,7 +233,12 @@ describe('d2l-filter', () => {
 				await page.$eval(selector, async(filter) => {
 					await new Promise((resolve) => {
 						filter.addEventListener('d2l-localize-resources-change', resolve, { once: true });
-						document.querySelector('html').setAttribute('lang', 'ar');
+						if (document.querySelector('html').getAttribute('lang') !== 'ar') {
```
oelhanafey, line 236: > > Arabic in order to activate rtl
> 
> Right - it's possible the RTL test itself wants to use Arabic text, hadn't thought of that. Anyway I was just investigating flake, but I'm guessing if you just re-run the visual diff it'll pass.

That's true, I decided to keep the test in Arabic because they may have intended for it to test that. The reason the promise was not resolving was because it was waiting for the language change event to resolve on retries but the language had already changed so it times out. I added a check so that if the language had already changed it resolves the Promise

PR #2635, opened 2022-07-12
Add load-more paging to d2l-list

dbatiste:
This PR adds load-more paging functionality to `d2l-list`. Specifically:

* new `d2l-pager-load-more` component that can be used with lists and eventually tables
* new `PageableMixin` mixins to assist in wire-up of paging component
* new `d2l-list` `pager` slot where consumers render the paging control

Spike Table Wrapper PR: https://github.com/BrightspaceUI/core/pull/2636 to support using `d2l-pager-load-more`.

image


components/list/README.md
```diff
@@ -158,6 +158,12 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
 
 The `d2l-list` supports selectable items within a list, including both single and multi selection. Selection is enabled when `d2l-list-item`s have the `selectable` attribute. When items are selectable, multiple selection is the default behaviour, however the `selection-single` attribute can be applied to the `d2l-list` to enable single selection. A `d2l-list-header` component can be added to `d2l-list`'s `header` slot to provide select-all and bulk actions.
 
+## Pageable Lists
+
+Load-More paging functionality can be implemented in lists by placing a `d2l-pager-load-more` in `d2l-list`'s `pager` slot. The consumer must handle the `d2l-pager-load-more` event by loading more items, updating the pager state, and signalling completion by calling `complete()` on the event detail. Focus will be automatically moved on the first new item once complete. See [Paging](../../components/paging) for more details.
```
dbatiste, line 163: Curious what people's thoughts are - I would like to have a full working Daylight demo for the load-more paging, bit there's a fair amount of extra coding that's needed to set it up, and eventually `d2l-table-wrapper` will also support load-more, so perhaps it makes sense to place the demo in the Paging readme, but then it's not in the list/table documentation where people might look for it.

components/list/demo/list.html
```diff
@@ -155,22 +156,63 @@ <h2>List (with all the stuff)</h2>
...
+							return item;
+						};
+
+						let newItemCount = 0;
+						document.querySelector('d2l-selection-action[text="Add"]').addEventListener('d2l-selection-action-click', () => {
+							list.appendChild(createItem(`${++newItemCount}`, `New Item ${newItemCount}`, 'https://s.brightspace.com/course-images/images/38e839b1-37fa-470c-8830-b189ce4ae134/tile-high-density-max-size.jpg', 'Supporting Info'));
+						});
+
+						let lastLoadedIndex = 2;
+						list.querySelector('d2l-pager-load-more').addEventListener('d2l-pager-load-more', e => {
```
dbatiste, line 196: The code for a component handling load-more would obviously be different, but this illustrates what they would be doing: loading more items, updating the state of `d2l-pager-load-more`, and calling `complete()`.

components/list/list-item-generic-layout.js
```diff
@@ -530,6 +530,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		if (!this.gridActive) return;
 		const slot = (e.path || e.composedPath()).find(node =>
 			node.nodeName === 'SLOT' && node.classList.contains('d2l-cell'));
+		if (!slot) return;
```
dbatiste, line 533: The following line results in a JavaScript error in nested lists when focusing on `d2l-pager-load-more` because the `slot` would be undefined in that case.

components/list/list.js
```diff
@@ -175,6 +188,35 @@ class List extends SelectionMixin(LitElement) {
...
+	async _getItemsShowingCount() {
+		if (this.slot === 'nested') return this._itemsShowingCount;
+		else return this._getListItemsShowingTotalCount(false);
+	}
+
+	_getLastItemIndex() {
+		return this._itemsShowingCount - 1;
+	}
+
+	async _getListItemsShowingTotalCount(refresh) {
```
dbatiste, line 206: This helper method attempts to reduce the cost of calculating the root showing count by using a cached valued when possible, thereby avoiding some tree walking. 

At first I was going to have the consumer provide this value directly to the `d2l-pager-load-more` component, but future expand/collapse functionality will affect this count so that it will only include items in expanded items (no items in collapsed lists), and would require the consumer to maintain that count based on user interaction with the list. The list is in the best position to do that, but unfortunately needs to deal with the complexity of getting the total root count including nested lists.

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+			<span class="action">${this.localize('components.pager-load-more.action', { count: this.pageSize })}</span>
+			${this.itemCount > -1 ? html`
+				<span class="separator"></span>
+				<span class="info">${this.localize('components.pager-load-more.info', { showingCount: this.itemShowingCount, totalCount: this.itemCount })}</span>
+			` : null}
+			<d2l-loading-spinner size="24"></d2l-loading-spinner>
+		</button>`;
+	}
+
+	async _handleClick() {
```
dbatiste, line 124: The general order of operations here is:

1. get the pageable (ex. `d2l-list`) component
2. get the last item index as so we can figure out where to move focus to after
3. dispatch event with `complete` method, so consumers can load items
4. wait for consumer resolve the promise (by calling `complete`)
5. find (using the last item index above) the first new item so we can move focus to it
6. if it has a custom focus implementation, then call it, otherwise find first focusable descendant and focus it

lang/en.js
```diff
@@ -91,6 +91,8 @@ export default {
 	"components.more-less.less": "less",
 	"components.more-less.more": "more",
 	"components.overflow-group.moreActions": "More Actions",
+	"components.pager-load-more.action": "Load {count} More",
```
dbatiste, line 94: I need to copy these to the other lang files.

components/list/demo/list.html
dlockhart: I'm not sure if it might be easier/more typical to do this from a Lit demo component? Then you'd be "fetching" (with your timeout), adding those new items to a local state collection, which will trigger a re-render.

What's unclear to me is how calling `e.detail.complete()` would fit into that kind of workflow... I'm worried that it'll be awkward for the consumer to hold onto the promise and trigger it at the right time.

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+			/**
+			 * The number of additional items to load.
+			 * @type {number}
+			 */
+			pageSize: { type: Number, attribute: 'page-size', reflect: true },
+			/**
+			 * The number of items showing. Assigned by PageableMixin.
+			 * @ignore
+			 */
+			itemShowingCount: { type: Number },
```
dlockhart, line 42: If you're not expecting consumers to be setting this:

```suggestion
			itemShowingCount: { attribute: false, type: Number },
```

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+			_loading: { state: true }
```
dlockhart, line 43: I think these still have a `type`?
```suggestion
			_loading: { state: true, type: Boolean }
```

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+		return 'button';
+	}
+
+	render() {
+		if (!this.hasMore) return;
+		const classes = {
+			'd2l-label-text': true,
+			'loading': this._loading
+		};
+		return html`<button class="${classMap(classes)}" @click="${this._handleClick}">
```
dlockhart, line 114: Otherwise you get a submit button by default!

```suggestion
		return html`<button class="${classMap(classes)}" @click="${this._handleClick}" type="button">
```

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+			<span class="action">${this.localize('components.pager-load-more.action', { count: this.pageSize })}</span>
+			${this.itemCount > -1 ? html`
+				<span class="separator"></span>
+				<span class="info">${this.localize('components.pager-load-more.info', { showingCount: this.itemShowingCount, totalCount: this.itemCount })}</span>
+			` : null}
```
dlockhart, line 119: Nit: we've been using `nothing` in these cases as I believe it allows Lit to make some optimizations

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,155 @@
...
+			<d2l-loading-spinner size="24"></d2l-loading-spinner>
+		</button>`;
+	}
+
+	async _handleClick() {
+		if (this._loading) return;
```
dlockhart, line 125: Does it make sense for the `button` to disable itself while things are loading to avoid this?

components/list/demo/list.html
dbatiste: Ok, I can set up a demo component that would be akin to what they would do.  I was sorta hoping to keep the demo component as is, since the snippet shows so many other things related to the component.  I guess it doesn't really matter too much if we don't show the snippet here, since devs will be looking at the Daylight site for examples.  Having said that, we'll have the same challenge there... showing appropriate example code without burying important aspects in the demo component.

components/paging/pager-load-more.js
dbatiste: I don't _think_ so. The [Lit docs](https://lit.dev/docs/components/properties/#internal-reactive-state) only show examples that specify `state`, and they say the only other valid option is `hasChanged`.

components/paging/pager-load-more.js
dbatiste: Thanks... yeah I saw this earlier today in one of your nav PRs. 

@dcostiuc, I think `d2l-tooltip-help` should do the same.

components/paging/pager-load-more.js
dbatiste: Yeah, it's probably good to get in the habit of this even if the result is the same.

In Lit 1, `null`, `undefined`, and an empty string all rendered an empty text node, which had the potential to suppress the default content of a `slot`. That was the primary use-case for using `nothing` then. (ex. `<my-element>${stuff}</my-element>` where stuff is `null | undefined` would suppress my-element's default slot content)

For Lit 2, the primary use-case now appears to be to normalize the binding behaviour for properties and attributes - `nothing` removes the attribute, whereas `null` and `undefined` render an empty string.  The Lit docs are slightly contradictory for [nothing](https://lit.dev/docs/api/templates/#nothing) and [Rendering Nothing](https://lit.dev/docs/templates/conditionals/#rendering-nothing) - `nothing` seems to be the more accurate of the two, as I've confirmed that that `null`, `undefined`, and empty string really do render no nodes in child contexts (the slot scenario above), so default slot content renders regardless of which value of those values is used.

components/paging/pager-load-more.js
dbatiste: That's a good question - I considered this but decided not to, since disabling it may cause focus to move to the `body` which might be announced to the user. I can try it to see.

components/paging/pager-load-more.js
dlockhart: Cool yep, my misunderstanding!

components/paging/pager-load-more.js
dcostiuc: Ah, good point, thanks!

components/paging/pager-load-more.js
dbatiste: Result of some exploration: either approach is viable. Leaving the button enabled offers the potential for the user to move focus and return and get the loading status again, albeit with inconsistent screen reader experience (VoiceOver does not re-announce the alert, NVDA does). Disabling the button causes focus to be lost but we do still get the initial status message announced, but seemingly the browser knows where to move focus if the user hits tab. For now, Design feels disabling the button would be the smoothest experience, so I've changed it to that approach.

components/list/demo/list.html
dbatiste: I've set up a demo component to show how this would work for our typical component consumers.  Obviously the data is mocked up client-side. I suspect we will want to further build this out to include some filtering, since some folks are looking for guidance on how to persist selection when filters change.

components/list/demo/demo-list.js
```diff
@@ -0,0 +1,242 @@
...
+
+	_handlePagerLoadMore(e) {
+		const pager = e.target;
+
+		// mock delay consumers might have
+		setTimeout(() => {
+			this._lastItemLoadedIndex += this._pageSize;
+
+			if (this._lastItemLoadedIndex >= this.items.length - 1) {
+				pager.hasMore = false;
```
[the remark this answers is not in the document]
dbatiste: Yep... you're right.  This was an oversight from copying this stuff over from the non-component version of this code.

components/paging/pager-load-more.js
```diff
@@ -0,0 +1,146 @@
...
+			this._loading = true;
+			this.dispatchEvent(new CustomEvent('d2l-pager-load-more', {
+				bubbles: false,
+				composed: false,
+				detail: { complete: resolve }
+			}));
+		});
+		this._loading = false;
+
+		const item = pageable._getItemByIndex(lastItemIndex + 1);
```
dlockhart, line 131: Is there any concern that when they resolve the promise that their collection hasn't actually re-rendered yet and then this line wouldn't find the new item(s)?

Seems like the typical workflow will be for the consumer to do a `fetch` call, get some new data, update a local collection and then resolve this promise. But then their re-render could easily happen afterwards. 🤔

dbatiste: Possibly yeah. Perhaps it would be best practice that they would `await this.updateComplete;` before calling `complete()`.  Even at that point, it is possible that the list-item itself has not been updated, but `d2l-pager-load-more` will wait for it's `updateComplete` if defined.  At that point it should be ready to be focused... unless... the item isn't draggable, isn't selectable, doesn't define a primary action (navigation or button), doesn't provide any secondary actions, and is also a hypermedia component that is yet to load and render. I'm not sure how likely it is to have that scenario... but if we did, maybe we'd wait some amount of time or check on an interval for a focusable element. Thoughts?

dlockhart: What if `PageableMixin` provided a way for the pager to register with it to be notified whenever new items were added? I guess List doesn't actually know that though eh... I was thinking the items fired an event that List was listening for but that's just the selection observer stuff. To finish my thought: if such a thing existed, the pager could just wait for a new item to be added _after_ it had been put into "load more" state.

dbatiste: Yeah I was thinking about that a bit. Items can be added to the list for a couple different reasons - via drag & drop or by bulk actions, and the list doesn't really know why they were added.  We could require the items be added with a special attribute to identify them, but it would require removing that attribute, probably outside of render.  Not sure if I like that any better.

dlockhart: For sure it's definitely not straightforward. We could block adding while load-more is in progress or something, but yeah... complex.

dlockhart (approved):
Approved so this can get in before you're off next week -- we can make any necessary tweaks later pretty easily I think.

PR #2643, opened 2022-07-13
fix: clean up aria-describedby/labelledby on tooltip target

dlockhart:
Ran into this issue in a scenario where a tooltip is conditionally rendered. When the tooltip is initially rendered, it adds `aria-labelledby` (or `aria-describedby`) to the target pointing back at the tooltip -- all good.

But then when it's removed and then re-added later, now the target will have TWO values for `aria-labelledby`, only one of which is now valid.

### DIFF
--- a/components/tooltip/test/tooltip.test.js
+++ b/components/tooltip/test/tooltip.test.js
@@ -32,12 +32,29 @@ describe('d2l-tooltip', () => {
 		it('should add aria-labelledby to its target if for-type is \'label\'', async() => {
 			const tooltipLabelFixture = await fixture(labelFixture);
 			const target = tooltipLabelFixture.querySelector('#label-target');
-			expect(target.hasAttribute('aria-labelledby')).to.be.true;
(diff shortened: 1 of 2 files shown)


components/tooltip/test/tooltip.test.js
```diff
@@ -32,12 +32,29 @@ describe('d2l-tooltip', () => {
 		it('should add aria-labelledby to its target if for-type is \'label\'', async() => {
 			const tooltipLabelFixture = await fixture(labelFixture);
 			const target = tooltipLabelFixture.querySelector('#label-target');
-			expect(target.hasAttribute('aria-labelledby')).to.be.true;
+			const tooltip = tooltipLabelFixture.querySelector('d2l-tooltip');
+			expect(target.getAttribute('aria-labelledby')).to.equal(tooltip.id);
```
dlockhart, line 36: Not really part of this change, but figured it was better to check that the value it set was actually... correct.

dbatiste (approved):

dbatiste:
Nice find!

PR #2647, opened 2022-07-14
Definitely Do Not Merge

svanherk:
Fail every visual-diff test.


svanherk:
Fixed officially in https://github.com/BrightspaceUI/core/pull/2709 and https://github.com/BrightspaceUI/core/pull/2714

PR #2649, opened 2022-07-14
feat: Add trim-whitespace mixin

eKoopmans:
This PR adds `TrimWhitespaceMixin`, which is required for the `object-property-list` component. A few details:

- I've separated the actual trimming logic (`trim-whitespace-core.js`) from the mixin (`trim-whitespace-mixin.js`)
- I also made a directive version, but... I'm thinking of just deleting it in favour of the mixin
- both the mixin and the directive use an additional `noTrim()` directive to prevent children from being trimmed
- I've detailed some of the limitations in the readme

I've also included small fixes to two other readmes (`async-container` and `focus-mixin`).

Please weigh in on whether you think it's worth keeping `trim-whitespace-directive`! I'm thinking no - one reason is it currently doesn't `stop()` the observer when disconnected, which seems inefficient. To add that I'd need to use an AsyncDirective, which I've read adds "[runtime and payload costs](https://github.com/lit/lit/pull/1432)".

Rally: https://rally1.rallydev.com/#/?detail=/userstory/641658871145&fdp=true


mixins/trim-whitespace/trim-whitespace-mixin.js
```diff
@@ -0,0 +1,32 @@
+import { TrimWhitespaceCore } from './trim-whitespace-core.js';
+
+export const TrimWhitespaceMixin = superclass => class extends superclass {
+	static get properties() {
+		return {
+			trimWhitespaceDeep: { type: Boolean, attribute: 'trim-whitespace-deep', reflect: true },
```
dlockhart, line 6: What's the motivation for this being a reactive property that would be exposed to consumers of a component using this mixin vs. just being a regular property that the component itself would decide when to use?

Just thinking in that scenario, the component would set it in its constructor and then you'd defer creating `TrimWhitespaceCore` until `connectedCallback`. There'd be no ability to change the value later obviously.

Was there a scenario where the component itself doesn't know if it wants deep whitespace trimming and it would need to be left up to the consumer?

eKoopmans: No particularly good reason 😄. I asked myself the same question when I started writing the readme. I'd be happy to remove the reactive property - would you think:

1. Still keep it as a property (`.trimWhitespaceDeep`) on the object, and still set it in the constructor as the readme shows?
2. Or change it to a `static get trimWhitespaceDeep()` which can then be set on the inheriting object?

The advantage of number 2 is that it would be less likely someone would try to change the property after initializing, I think.

mixins/trim-whitespace/trim-whitespace-core.js
```diff
@@ -0,0 +1,58 @@
...
+	const attachShadow = node.attachShadow.bind(node);
+
+	node.attachShadow = (...args) => {
+		const shadowRoot = attachShadow(...args);
+		callback(shadowRoot);
+		return shadowRoot;
+	};
+};
+
+export class TrimWhitespaceCore {
```
dbatiste, line 14: Although it's called out in the readme, I'm a little worried about the performance. Some thoughts...

* Consumers could use it liberally... thinking, "isn't is always better to trim this whitespace", which is probably usually true, cost aside.  That became a standard practice back in the Polymer days when they supported `trim-whitespace` on `templates`, though I recall someone from Google saying... "it's performant". 
* It would be possible for an ancestor and a descendant to both be trimming whitespace, resulting in multiple observers observing the same sub-tree. Consumers could end up composing this scenario unknowingly as they use/nest other components that happen to be trimming whitespace.  Maybe there is opportunity for optimization here.
* Do we have a use-case for `deep`? Aside from the same point above applying to an even larger sub-tree, adding more observers, I am also a little worried about modifying other DOM scopes. I know we're just trimming whitespace, but it still doesn't feel right to me.
* This is effectively updating the DOM outside of `render`. Just whitespace, but maybe there is a way to use a directive that doesn't go against that best practice?

Maybe we can find a way to reduce the scope of nodes being observed so that consumers can't apply this to a huge sub-tree?  For example, a directive that only applied to immediate children, or direct children of a slot.

eKoopmans: So, some answers:

> > directive that only applied to immediate children
>
> I'll investigate whether this is viable 👍.

So, yes it's doable, but it's not great. Usage would look like:

```js
return html`${trimShallow()}|
	<span ${trimShallow()}> ${this.spanText} </span>
	|
	<span ${trimShallow()}> ${this.enableSpanContent ? ' (2 Span Content 2) ' : nothing} </span>
	|
	${this.enableSpanElement ? html`<span ${trimShallow()}> (3 Span Element 3) </span>` : nothing}
	|
	<slot></slot>
|`;
```

Some downsides:

- need to be very verbose with adding the directive to every element
- the directive still needs to use a `MutationObserver`, it just doesn't watch on the full `subtree` in this version
  - unfortunately, the directive `update` method gets called before the DOM is updated, so the trimming can't happen there instead of using a `MutationObserver`

I'll investigate more whether there are other optimizations to be made...

PR #2652, opened 2022-07-15
fix: tag-list: also check d2l-backdrop-role to get list items

margaree:
**Issue:**
At mobile widths, filter appears at the side and uses `d2l-backdrop`. `d2l-backdrop` changes the role of everything behind it to be `presentation`. When a filter change would happen (and `d2l-filter` would remain open), `d2l-tag-list` would look for `listitem`s to do the necessary chomping to 3 lines, but the problem was they would have a role of `presentation` instead of `listitem` and so the `d2l-tag-list` would do no such chomping.

**Solution Notes:**
When `d2l-backdrop` is enabled, it stores the previous role as `d2l-backdrop-role`. This solution uses that to get the list items.

This solution knows a bit too much about how `d2l-backdrop` is working and so is intended to be temporary. One suggestion for a more permanent solution by Stacey is having a `d2l-role` on the list items that reflects the `role` and that this, menu, etc look at that in cases like this rather than relying on `role`.

### DIFF
--- a/components/backdrop/backdrop.js
+++ b/components/backdrop/backdrop.js
@@ -3,6 +3,8 @@ import { css, html, LitElement } from 'lit';
 import { cssEscape, getComposedChildren, getComposedParent, isVisible } from '../../helpers/dom.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 
+export const BACKDROP_ROLE = 'd2l-backdrop-role';
+
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 
 const scrollKeys = [];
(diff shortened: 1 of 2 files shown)


dlockhart (approved):

svanherk (approved):

dbatiste:
I think the reason why a generic data attribute was not used in this case was to ensure that if some other logic also wanted to replace the `role` or `aria-hidden` for another purpose, they would not interfere with each other. Maybe not something to actually worry about, but I think that was the rationale.

I too an eager to be able to use the `inert` behaviour provided by the browser - it will at least avoid the `role` and `aria-hidden` replacement shenanigans. It could also simplify some of our focus trapping logic.  I think `inert` targets the node and all of its descendants, and doesn't provide a way for descendants to opt out, so _some_ tree walking (siblings and siblings of ancestor elements) will still be required to manage the `inert` state.  

```html
<div>
    <div inert><!-- other content --></div>
    <div>
        <div inert><!-- other content --></div>
        <d2l-dialog opened></d2l-dialog> / <d2l-backdrop shown></d2l-backdrop>
    </div>
</div>
<div inert>
    <div><!-- other content --></div>
</div>
```
The `inert` might also benefit legacy/MVC dialogs.

PR #2654, opened 2022-07-18
fix: List paging documentation: Small tweaks

margaree:
Changes:
- The `Pageable Lists` section was placed within the `Selection Lists` section so I moved that to be below `Selection Lists`
- Fixed the best practices block in `Paging`
- Removed hiding the properties and events in `Paging` just temporarily until there's an interactive demo that exhibits these so they aren't missing on the page.

I also updated the `Paging` issue in the documentation repo so that it will properly use the new Paging readme.


dlockhart (approved):
🎉

dbatiste:
Thanks Margaree.  😄

PR #2665, opened 2022-07-20
feat: Visibility Switch > Update to include conditions met

dcostiuc:
[Rally Story](https://rally1.rallydev.com/#/?detail=/userstory/637810377847&fdp=true) (includes link to design for more context)

This PR updates the Visibility Switch to include a 3rd new state that is displayed when "conditions" are passed into it. 

Also includes other updates related to the design spec, such as having the Visibility Switch label be dynamic, rather than static. 

Note that there were various accessibility (screen reader-related) issues/bugs that were identified, and these will be looked more into in [this story](https://rally1.rallydev.com/#/?detail=/userstory/642704943581&fdp=true) (and potentially other stories/tickets).


components/switch/README.md
```diff
@@ -62,6 +62,7 @@ To make your usage of `d2l-switch` accessible, use the following property:
 ## Visibility Switch [d2l-switch-visibility]
 
 The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions ("Visible. Conditions must be met.").
```
dcostiuc, line 65: Particularly pen to any feedback/revision ideas for this.

components/switch/switch-visibility.js
```diff
@@ -32,6 +41,35 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
...
+			'conditions-show': this.on && this._hasConditions
+		};
+
+		const conditions = html`
+			<d2l-tooltip-help class="${classMap(tooltipHelpClasses)} d2l-switch-text" id="conditions-help" text="${this.localize('components.switch.conditions')}" inherit-font-style>
+				<slot @slotchange="${this._handleConditionsSlotChange}"></slot>
+			</d2l-tooltip-help>
+			`;
+
+		return html`<span id="${this._textId}" class="d2l-switch-text">${this.text}${conditions}</span>`;
```
dcostiuc, line 67: Is there a way to re-use what we return in `super().render()` and insert the conditions inside of that html literal?

components/switch/switch-visibility.js
```diff
@@ -32,6 +41,35 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
 		return html`<d2l-icon icon="tier1:visibility-show"></d2l-icon>`;
 	}
 
+	get text() {
```
dcostiuc, line 44: I don't think we need documentation here anymore (i.e. [what we had before](https://github.com/BrightspaceUI/core/pull/2665/files#diff-43ec62628ec01309e32f82bffb749c81f3ef88699064f17e97180a4d261fed33L11-L14)) because we no longer allow the consumer to change this text anyway.

components/switch/test/switch-visibility.test.js
```diff
@@ -7,4 +8,9 @@ describe('d2l-switch-visibility', () => {
 		runConstructor('d2l-switch-visibility');
 	});
 
+	it('should have conditions if conditions are passed in', async() => {
+		const switchFixture = await fixture(html`<d2l-switch-visibility on>Conditions</d2l-switch-visibility>`);
+		expect(switchFixture._hasConditions).to.be.true;
```
dcostiuc, line 13: Not sure if this test is redundant given the new ["on-with-conditions"](https://github.com/BrightspaceUI/core/pull/2665/files#diff-a232a1a1541f68920c062087554cefc14fbd4897c80da754c035fc4337e416b6R32-R42) vdiff test that we will have.

components/switch/switch-visibility.js
```diff
@@ -32,6 +41,35 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
...
+			return this.localize('components.switch.visible');
+		}
+		else {
+			return this.localize('components.switch.hidden');
+		}
+	}
+
+	_createLabel() {
+		const tooltipHelpClasses = {
+			'conditions-show': this.on && this._hasConditions
```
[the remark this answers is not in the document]
dcostiuc: Just wondering, how come it would be better to specify `switch-visibility` there? 
Wouldn't it already be possible to infer that it has to do with just the visibility switch?

components/switch/switch-visibility.js
dcostiuc: Good question! Based on the search results for [d2l-switch-visibility](https://search.d2l.dev/search?project=Brightspace-Configs&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22d2l-switch-visibility%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5) and [SwitchVisibility](https://search.d2l.dev/search?project=lms&full=%22SwitchVisibility%22&defs=&refs=&path=&hist=&type=&xrd=&nn=1), as well as searching for SwitchVisibility in the MVC and Testing solutions in Visual Studio, it looks like that was never used. 

In the case of backporting to MVC in the lms, if some place _was_ overriding the `text`, and we got rid of that within the wrapper class, would we expect an obvious error (like a compile-time error) to pop up?

components/switch/test/switch-visibility.test.js
dcostiuc: Would you want me to remove it then?

components/switch/test/switch-visibility.visual-diff.html
```diff
@@ -29,5 +29,16 @@
 		<d2l-switch-visibility id="on" on></d2l-switch-visibility>
 	</div>
 
+	<div class="visual-diff">
```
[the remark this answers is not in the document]
dcostiuc: Yup, very good point, thanks!

components/switch/test/switch-visibility.visual-diff.js
```diff
@@ -25,11 +25,25 @@ describe('d2l-switch-visibility', () => {
 
 			[
 				{ name: 'off', selector: '#off' },
-				{ name: 'on', selector: '#on' }
+				{ name: 'on', selector: '#on' },
+				{ name: 'on with conditions', selector: '#on-with-conditions' },
+				{ name: 'on with conditions, conditions focused', selector: '#on-with-conditions' }
```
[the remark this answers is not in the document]
dcostiuc: Yeah, you're right, some obvious code smells with those if statements haha. 
Also thanks for letting me know about that test name constraint!

components/switch/switch-visibility.js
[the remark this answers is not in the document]
margaree: It's generally been our best practice to preface classes with the component name. It also makes the class more searchable for someone who needs to find it in the future (e.g., perhaps `conditions-show` was used within something else, `switch-visibility-conditions-show` is less likely to have been)

components/switch/switch-visibility.js
margaree: If the MVC control had a `text` property which you then removed, then yes there would likely be a compile time error (more likely from a full all than just a build of `lp` since it would likely be in an `le` project)

components/switch/test/switch-visibility.test.js
margaree: Yup I think that's safe.

components/switch/switch-visibility.js
dcostiuc: I see. I also searched for `SwitchVisibility` in the `le` folder through Visual Studio and the only use case that came up wasn't making use of the `text` properly either

components/switch/switch-visibility.js
dcostiuc: Thoughts on having that be a method, rather than a getter? I was thinking of doing 

```js
return html`${super.labelContent()}${conditions}`;
```

So that we're not copy-pasting `this.text`, in case in the future we want to change that. 
But to make use of `super()` within `labelContent()`, we would need it to be a method. 

Admittedly, it would probably also be fine we just copy-pasted `${this.text}` and kept it as a getter,l as suggested above.

components/switch/switch-visibility.js
dcostiuc: Update: I _did_ find [a place or two (same use case) ](https://d2l.slack.com/archives/C03FC0GDVB7/p1658435070006839)where the Visibility Switch `text` property is being overridden.

components/switch/switch-visibility.js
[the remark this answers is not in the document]
dlockhart: I'd be OK if we wanted to start relaxing this practice a bit. It was originally a thing because IE11 and legacy-Edge didn't support shadow DOM, so all these classes were in the light DOM and could collide with each other. Not an issue anymore now that we hard-block them.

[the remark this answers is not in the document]
dcostiuc: So is there still enough of a benefit in prefacing classes with the component name? 
As in, are we now more lenient with not doing that, or do we _prefer_ to not preface them?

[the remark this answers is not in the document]
dlockhart: The way I would go with it is if the class is defined in a shareable `whatever-styles.js` that gets imported into your component (e.g. [typography](https://github.com/BrightspaceUI/core/blob/main/components/typography/typography.js), [link](https://github.com/BrightspaceUI/core/blob/main/components/link/link.js)), then it makes a lot of sense to prefix with `d2l-whatever` since we would still need to worry about collisions from multiple things being imported and used in a component.

But if it's just 100% internal to your component (the class & style is defined in your component and used in your component), then the benefits Margaree mentioned about searchability go away.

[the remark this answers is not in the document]
dcostiuc: > The way I would go with it is if the class is defined in a shareable whatever-styles.js that gets imported into your component (e.g. [typography](https://github.com/BrightspaceUI/core/blob/main/components/typography/typography.js), [link](https://github.com/BrightspaceUI/core/blob/main/components/link/link.js))

So essentially if we're `export`ing any style classes, then we should prefix. 
Otherwise it's not necessary. 

Going forward though, do we want that to be more strict (i.e. should I remove the prefixing for that `conditions-show` class styling), or do we want to be loose with it (whether you do it or not in a non-shareable case, doesn't matter)?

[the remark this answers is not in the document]
dlockhart: Correct. I'm OK with not having a hard and fast rule here, so whatever you prefer.

components/switch/demo/switch.html
```diff
@@ -49,7 +49,14 @@ <h2>Visibility</h2>
 
 			<d2l-demo-snippet>
 				<template>
-					<d2l-switch-visibility></d2l-switch-visibility>
```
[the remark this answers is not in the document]
margaree: Ah sorry for this I meant in separate demo snippets, like with the other demos above (e.g., first one is plain visibility, next is something like `<h2>Visibility with conditions</h2>`

components/switch/switch-visibility.js
margaree: We should probably add a visual diff test that overrides `text` in `switch-visibility` (and then can remove that once it's no longer needed)

components/switch/switch-visibility.js
```diff
@@ -32,6 +64,25 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
...
 
+	get _labelContent() {
+		if (this._text) return super._labelContent;
+
+		const tooltipHelpClasses = {
+			'switch-visibility-conditions-show': this.on && this._hasConditions
+		};
+
+		const conditions = html`
+			<d2l-tooltip-help class="${classMap(tooltipHelpClasses)} d2l-switch-text" id="conditions-help" text="${this.localize('components.switch.conditions')}" inherit-font-style>
```
margaree, line 75: nit: can probably wrap this line over multiple lines

components/switch/switch-visibility.js
dcostiuc: Yeah I was gonna bring that up and ask if we would like to add tests for that (which I think makes sense, even though it's a bit weird since we're hoping to eventually remove this backwards-compatibility feature).

components/switch/switch-mixin.js
```diff
@@ -21,8 +21,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 			 */
 			on: { type: Boolean, reflect: true },
 			/**
-			 * REQUIRED: The text that is displayed for the switch label.
-			 * @type {string}
+			 * @ignore - Need to add documentation in each component that uses this mixin.
```
margaree, line 24: Does `Need to add documentation in each component that uses this mixin` get added to custom-elements.json at all or does the `@ignore` get priority?

components/switch/switch-visibility.js
```diff
@@ -32,6 +64,29 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
...
+		if (this._text) return super._labelContent;
+
+		const tooltipHelpClasses = {
+			'switch-visibility-conditions-show': this.on && this._hasConditions
+		};
+
+		const conditions = html`
+			<d2l-tooltip-help 
+				class="${classMap(tooltipHelpClasses)} d2l-switch-text" 
+				id="conditions-help"
```
margaree, line 77: nit: (only if you're changing other things) attributes could be in alphabetical order

components/switch/switch.js
```diff
@@ -9,6 +9,16 @@ import { SwitchMixin } from './switch-mixin.js';
  */
 class Switch extends SwitchMixin(LitElement) {
 
+	static get properties() {
+		return {
+			/**
+			 * REQUIRED: The text that is displayed for the switch label.
+			 * @type {string}
+			 */
+			text: { type: String, reflect: true }
```
margaree, line 18: So since this is already in `SwitchMixin` you don't need to do this part, just add it to the top comment block in the file:
`@attr text - REQUIRED: The text that is displayed for the switch label.`

components/switch/test/switch-visibility.visual-diff.html
```diff
@@ -29,5 +30,79 @@
...
+			These are some conditions that must be met for the activity to be visible.
+			<ul>
+				<li> Condition 1 </li>
+				<li> Condition 2 </li>
+				<li> Condition 3 </li>
+			</ul>
+		</d2l-switch-visibility> 
+	</div>
+	
+	<div class="visual-diff">
```
margaree, line 85: Since it's the usecase in LCS, perhaps also having a case for `text-position="hidden"` overriding `text` would be good as well.

components/switch/switch-mixin.js
dcostiuc: So to directly answer your question, it seems `ignore` gets priority and I can get away with `Need to add documentation in each component that uses this mixin` as an extra comment for future developers.

components/switch/test/switch-visibility.visual-diff.html
dcostiuc: Yeah fair point. I was thinking, without considering the LCS usecase, that technically things shouldn't matter visually in terms of whether the visibility switch `has conditions` or not, whether the `text` is being overridden or not etc.

But yeah since it _is_ a use case, it would be good to have, for example, for regression checks.

components/switch/switch.js
dcostiuc: Hmm I'm not sure if that works as well, or if it's fine how it is, because I noticed that with 
`@attr text - REQUIRED: The text that is displayed for the switch label.`, we get 

```
        {
          "name": "text",
          "description": "REQUIRED: The text that is displayed for the switch label."
        }
```
for the `attributes`, and nothing for the `properties`.

dcostiuc: Technically, the above still renders in the properties table as:

text |   | REQUIRED: The text that is displayed for the switch label. | --
-- | -- | -- | --

But in the above, it's missing it's type of `string` since that wasn't generated.

components/switch/switch-mixin.js
margaree: Great, yeah Daylight pulls from `attributes` and not `properties`

components/switch/switch.js
dcostiuc: I had also previously tried doing
`@property text - REQUIRED: The text that is displayed for the switch label. @type {string}`

but that seems to be the opposite of `@attr`, namely it generates
```
        {
          "name": "text",
          "description": "REQUIRED: The text that is displayed for the switch label."
        },
```
for the `properties`, and nothing for the `attributes`.

margaree: Yup so in that case we can change it to `@attr {string} text - REQUIRED: The text that is displayed for the switch label.`. The analyzer generally works with standard jsdoc comment format, so there's pretty good documentation on that out there (e.g., https://jsdoc.app/tags-property.html).

If we did want it to appear in properties as well we could also add a `@property` line but I don't think we need that too since Daylight just needs the attributes.

dcostiuc: Ah, nice. Looks like that works!

components/switch/test/switch-visibility.visual-diff.html
```diff
@@ -29,5 +30,95 @@
...
+			<ul>
+				<li> Condition 1 </li>
+				<li> Condition 2 </li>
+				<li> Condition 3 </li>
+			</ul>
+		</d2l-switch-visibility> 
+	</div>
+
+	<div class="visual-diff">
+		<d2l-switch-visibility id="on-with-conditions-text-position-start-text-overridden" on text="Label text has been overridden.">
```
margaree, line 97: I think the name for this one and the above one are swapped

margaree (commented):
Just the comment on the ids for the visual diff tests, otherwise LGTM!

margaree (approved):

PR #2672, opened 2022-07-21
fix: tag-list: handle focus on delete even if entire list is not clearable

margaree:
I noticed when using a `tag-list` in Nova, where the use case is that only select items of the list are clearable, that the focus behaviour on deletion of the last item wasn't behaving as expected (focus was just going outside of the list) which was caused by this.

I'm not sure why this was originally the case. The PR where `clearable` was added to `tag-list` doesn't seem to provide context (https://github.com/BrightspaceUI/core/pull/2319). It states that this behaviour was modelled off of `multi-select-list` but that doesn't require the whole list to be deletable (https://github.com/BrightspaceUILabs/multi-select/blob/main/multi-select-list.js#L322).


margaree:
The table visual-diff change is unexpected and unrelated to this. It seems okay though? Did a quick look at it and not really sure why it would have changed like this.

svanherk:
> The table visual-diff change is unexpected and unrelated to this. It seems okay though? Did a quick look at it and not really sure why it would have changed like this.

Looks like Puppeteer bumped the Chromium version 23 hours ago: https://github.com/puppeteer/puppeteer/releases/tag/v15.5.0
Timing seems to line up with when we started seeing this, I agree that they seem ok

svanherk (approved):

PR #2676, opened 2022-07-22
Bump MathJax to v3.2.2 and re-align rendering to the new version

grant-cleary:
MathJax 3.2 introduces some features/fixes that may make it a bit easier to simplify the html-block control. As such, we want to start off by getting it updated and making sure our custom shadow adaptor still works properly.


helpers/mathjax.js
```diff
@@ -70,7 +72,8 @@ export class HtmlBlockMathRenderer {
 			});
 
 			await window.MathJax.startup.promise;
-			window.MathJax.typesetShadow(elem.getRootNode(), elem);
+			renderingPromise = renderingPromise.then(() => window.MathJax.typesetShadow(elem.getRootNode(), elem));
```
grant-cleary, line 75: MathJax seems to have made more of its package loading asynchronous, so we now need the shadow handler typesetting to also be asynchronous. This does cause problems if we're typesetting multiple html-blocks on a page, however, because MathJax relies on internal state tracking during rendering, order matters here and we need to make sure only one html-block is ever rendering at a time.

helpers/mathjax.js
```diff
@@ -188,14 +189,16 @@ export function loadMathJax(mathJaxConfig) {
...
-				window.MathJax.typesetShadow = function(root, elem) {
+				window.MathJax.typesetShadow = async function(root, elem) {
 					const InputJax = startup.getInputJax();
 					const OutputJax = startup.getOutputJax();
 					const html = mathjax.document(root, { InputJax, OutputJax });
 
 					if (elem) html.options.elements = [elem];
 
-					html.render().typeset();
+					await mathjax.handleRetriesFor(() => html.render());
```
grant-cleary, line 199: `handleRetriesFor` is MathJax's way of enabling asynchronous actions. If it tries to render and something isn't loaded, it will retry after that loading has finished.

helpers/mathjax.js
```diff
@@ -94,21 +99,17 @@ export function loadMathJax(mathJaxConfig) {
 
 	if (mathJaxLoaded) return mathJaxLoaded;
 
-	const loadOptions = ['ui/menu'];
-	if (mathJaxConfig && mathJaxConfig.renderLatex) {
-		loadOptions.push('[tex]/all-packages');
```
grant-cleary, line 99: One benefit of forcing typesetting to queue is that we no longer need to bulk import all of the TeX packages. MathJax will dynamically import whatever packages are necessary based on the TeX its processing and now that we're only processing one html-block at a time, it's guaranteed that we'll have the packages loaded properly before we typeset.

helpers/mathjax.js
```diff
@@ -94,21 +98,17 @@ export function loadMathJax(mathJaxConfig) {
...
-	}
-
 	window.MathJax = {
 		chtml: {
+			adaptiveCSS: false,
```
grant-cleary, line 103: `adaptiveCSS` is a neat feature MathJax introduced in 3.2 that allows its CSS stylesheets to be constructed on the fly as math is processed (so not _all_ of the styles need to be appended always). Unfortunately, we can't make use of it. MathJax keeps a single internal stylesheet reference for all referenced math, so if we turn this on, we're going to end up with inconsistent/incomplete styles in some components.

grant-cleary:
Visual diff failures are minor formatting differences in the math, presumably as a result of the version bump. I'll avoid merging in the new goldens until we're happy with this PR just to keep churn to a minimum.

helpers/mathjax.js
dbatiste: Previously the `elem` would have been returned as the promise result of the `render()` call... but now it looks like nothing is returned for the resolved promise. Does this cause any issues for the caller of `render()`?

helpers/mathjax.js
dbatiste: It is unfortunate that it doesn't just append more styles to it, which could then be used in the various scopes.

helpers/mathjax.js
dbatiste: Specifically... https://github.com/BrightspaceUI/core/blob/main/components/html-block/html-block.js#L246

Looks like the code relies on the updated element being returned, so it can subsequently pass it to other renderers.

I'd expect something to be breaking though.

grant-cleary: Ah, it doesn't break because MathJax does its rendering in place. We still return the element to the outer html-block rendering call, it's just implicitly modified by MathJax's rendering. Because we're passing in a temp element, it stays stable even if MathJax is fully replacing its inner HTML.

helpers/mathjax.js
grant-cleary: Yeah, MathJax seems to rely heavily on globally scoping everything and sort of just building out whatever it needs when it needs it. Which is fine most of the time, but completely breaks in a situation like ours where we effectively have multiple documents with different math needs. 😞

helpers/mathjax.js
grant-cleary: I worded that poorly. The element being passed around here is already attached to the DOM and MathJax does its work directly on it. There's no need to retrieve an element from MathJax's `render()` function. I believe that was always redundant.

dbatiste: Doesn't it return the result of `typesetShadow` and assign it to `elem`?

grant-cleary: I mean technically it does, but it's redundant. So if anything, changing `typesetShadow` to not return anything makes more sense here. MathJax's actual rendering happens as a side effect.

dbatiste (approved):

PR #2681, opened 2022-07-25
[ignore for now] fix: tag-list resizing algorithm

margaree:
This updates the resizing algorithm to be more like the multi-select list one (https://github.com/BrightspaceUILabs/multi-select/blob/main/multi-select-list.js#L372).

Issues that were happening:
- A fun dance at certain widths where tag list items would be constantly re-triggering the resize observer (mostly when wrapped in a flex container)
- Chomping was incorrect at small widths

The idea is that the max available width is equivalent to the max width the `flex` container that wraps the items (`div.tag-list-container`) is when all the items are visible.

Testing:
- [ ] Visual-diff tests
- [ ] Nova (2 filter-tags usages, 1 tag-list usage)
- [ ] Consistent-eval

### DIFF
--- a/components/tag-list/tag-list.js
+++ b/components/tag-list/tag-list.js
@@ -135,8 +135,7 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		});
 		this._clearButtonResizeObserver.observe(clearButton);
 
-		let container = getOffsetParent(this);
-		if (!container || container.tagName === 'BODY') container = this.shadowRoot.querySelector('.tag-list-outer-container');
+		const container = this.shadowRoot.querySelector('.tag-list-outer-container');
 		this._resizeObserver = new ResizeObserver((e) => requestAnimationFrame(() => this._handleResize(e)));
 		this._resizeObserver.observe(container);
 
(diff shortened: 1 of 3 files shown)


components/tag-list/tag-list.js
```diff
@@ -297,7 +298,7 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 
 			return {
 				isHidden: computedStyles.display === 'none',
-				width: Math.ceil(parseFloat(computedStyles.width) || 0)
```
margaree, line 300: The calculations become more exact so removing the `Math.ceil` in a couple places

components/tag-list/tag-list.js
```diff
@@ -370,8 +371,11 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		this._hasShownKeyboardTooltip = true;
 	}
 
-	async _handleResize(entries) {
-		this._availableWidth = Math.floor(entries[0].contentRect.width);
+	async _handleResize() {
+		this._chompIndex = 10000;
```
margaree, line 375: This resets the container to get max available width. I haven't noticed a flicker of list items but in theory it's possible.

components/tag-list/tag-list.js
```diff
@@ -389,17 +395,21 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		await this.updateComplete;
 
 		this._items = await this._getTagListItems();
+		this._chompIndex = 10000;
```
margaree, line 398: Moved this out of the if to always reset the container size to get max available width

dlockhart (approved):
These changes LGTM

PR #2688, opened 2022-07-26
feat: Description List Component

jameswklassen:
## Description
A wrapper component around native [`dl`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) elements

## Checklist
- [x] Rough implementation
- [x] Final design approval
- [x] Visual diff tests
- [x] Accessibility tests
- [x] Unit tests
- [x] README documentation

## Visual Diff
The visual diff test results can be found in #2717
- #2717


components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this.stacked = false;
+		this.breakpoint = 350;
+		this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
+		this._checkIfShouldStack();
```
dlockhart, line 64: I wouldn't call this from the constructor, since it won't be attached to the DOM yet -- do this in `connectedCallback` or `firstUpdated`, or if the resize observer always fires when `observe()` is first called (I forget if that's true), then you may not even need to call it.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+];
+
+/**
+ * Wraps a native <dl> element, providing styling and resize behavior.
+ * @slot - Content to wrap
+ */
+class DescriptionListWrapper extends LitElement {
+	static get properties() {
+		return {
+			stacked: { type: Boolean, reflect: true },
```
dlockhart, line 46: It seems like we're not expecting consumers to set `stacked`, right? Like even if they do, the value will get clobbered by the the breakpoint handling code. So I'd suggest turning this into:

```suggestion
			_stacked: { state: true },
```

When then will result in the attribute no longer getting reflected to the host. That's both good (hosts "sprouting" attributes is kind of gross) and bad because then you can't target it using a CSS selector. To get around that, I'd suggest wrapping the `<slot>` in a `<div class="stacked">` (when it's stacked) and target that instead.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+			breakpoint: { type: Number, reflect: true },
+		};
+	}
+
+	static get styles() {
+		return css`
+			:host {
+				display: block;
+			}
```
dlockhart, line 55: Nit: if you modify the `display`, you need to specify the `hidden` behaviour:
https://daylight.d2l.dev/developing/best-practices/styles/#do%3A-style-the-%22display%22-value-of-the-host

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._resizeObserver.disconnect();
+	}
+
+	render() {
+		return html`
+			<slot @slotchange="${this._handleSlotChange}"></slot>
```
dlockhart, line 79: Did you find that the slot change handler was necessary? I'd be surprised if there were cases where the slot changed that required a recalculation but the resize observer on the host didn't fire...

components/typography/styles.js
```diff
@@ -41,6 +41,27 @@ export const bodyStandardStyles = css`
 	}
 `;
 
+export const bodyCompactStylize = (tag) => css`
```
dlockhart, line 44: Nit: I'd rename `tag` -> `selector` since technically it can be any CSS selector.

components/typography/styles.js
```diff
@@ -41,6 +41,27 @@ export const bodyStandardStyles = css`
...
+	}
+	:host([skeleton]) ${tag}.d2l-skeletize-paragraph-3 {
+		max-height: 3.6rem;
+	}
+	:host([skeleton]) ${tag}.d2l-skeletize-paragraph-5 {
+		max-height: 6rem;
+	}
+`;
+
 export const bodyCompactStyles = css`
```
dlockhart, line 65: Can this just call:

```css
bodyCompactStylize(css`.d2l-body-compact`)
```

components/description-list/description-list-wrapper.js
jameswklassen: Good point. I initially had the consumer set `stacked`, but when I added the breakpoint, it clobbers this.

I'm guessing the details of this implementation will depend on how we want "default" behavior to look (e.g. break at `__px`, don't break unless consumer specifies breakpoint, etc..)

components/description-list/description-list-wrapper.js
jameswklassen: Copied from the table wrapper - I think you're right 😄

components/typography/styles.js
jameswklassen: Yup! Still need to figure out why Lit syntax highlighting and style linting doesn't like this though.

components/description-list/description-list-wrapper.js
dlockhart: For sure yeah... I think `__px` is a reasonable place to start, and then we can iterator a bit later to see if there's a better approach. Nailing the default will be key too so that consumers almost never have to override.

components/description-list/description-list-wrapper.js
dlockhart: Table wrapper has the extra complexity in that it needs to apply CSS classes to the first/last rows and cells, so yeah it needs to listen for `slotchange` too.

components/description-list/description-list-wrapper.js
jameswklassen: I'm not sure what you mean by the following:

> if the resize observer always fires when observe() is first called

dlockhart: Like in `connectedCallback()` you're calling `.observe()` on it -- if that immediately triggers the callback, then you don't need to also call `_checkIfShouldStack()` when initializing since it'll do that for you. But I don't remember if that happens initially or if it only fires when a future-resize event happens.

jameswklassen: Gotcha. It does look like it fires initially, so I'll remove it from the constructor.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,97 @@
...
+
+	_checkIfShouldStack() {
+		if (this.clientWidth < this.breakpoint && !this._stacked) {
+			this._stacked = true;
+		} else if (this.clientWidth >= this.breakpoint && this._stacked) {
+			this._stacked = false;
+		}
+	}
+
+	_onResize() {
```
dbatiste, line 92: I know this is a WIP but figure I thought would inquire anyway - any chance we can use the [ResizeObserverEntry](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry) that the `ResizeObserver` provides to the handler? Using those values would always be preferred if possible, as opposed to potentially forcing a layout by calling `this.clientWidth` above. After all, that's one of the advantages of using RO. 

If trying this... be aware that we're limited to using [contentRect](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentRect) since `borderBoxSize` and `contentBoxSize` were not added to the entry API until much later, and so the RO polyfill does not support it. The Mozilla doc shows how it can be detected and adapted, but we haven't bothered so far.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	render() {
+		const classes = {
+			'stacked': this._stacked,
+		};
+		return html`<slot class="${classMap(classes)}"></slot>`;
+	}
+
+	_checkIfShouldStack() {
```
dbatiste, line 84: Nit: maybe `_updateStacked` is more accurate.  🤷‍♂️

components/typography/styles.js
```diff
@@ -41,26 +41,31 @@ export const bodyStandardStyles = css`
...
-		max-height: 2.4rem;
-	}
-	:host([skeleton]) .d2l-body-compact.d2l-skeletize-paragraph-3 {
-		max-height: 3.6rem;
-	}
-	:host([skeleton]) .d2l-body-compact.d2l-skeletize-paragraph-5 {
-		max-height: 6rem;
-	}
-`;
+export const bodyCompactStylize = (selector) => {
```
dbatiste, line 44: Nit: I'm not too particular about it, but maybe `generateBodyCompactStyles` is more accurate.  🤷‍♂️   That's the naming convention I used recent in the Prism helper when I did this sort of thing.

jameswklassen: I do like that naming as well.

I'm working on a draft PR specifically for these style changes, since it's a pretty big shift away from what `styles.js` currently has. I don't want conversations around those changes to get buried as in this PR.

components/description-list/description-list-wrapper.js
jameswklassen: Tried this out in 31deda23fb5fb35fd617959fe3c3af2f932f43dc - looks like it works!

I'm open to feedback on the best way to implement `ResizeObserverEntry`. I based my implementation off this:

https://github.com/BrightspaceUI/core/blob/ce0150cabc808de416cd0a9e1f77f64038d28c15/components/empty-state/empty-state-illustrated.js#L120-L125

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+			'stacked': this._stacked,
+		};
+		return html`<slot class="${classMap(classes)}"></slot>`;
+	}
+
+	_onResize(entries) {
+		if (!entries || entries.length === 0) return;
+		const entry = entries[0];
+		this._contentWidth = entry.contentRect.width;
+		this._updateStacked();
```
dlockhart, line 88: Since `_updateStacked()` is now only called from here, might be reasonable to just move it into here which would eliminate the need for the `_contentWidth` member variable.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+				--d2l-dt-max-width: unset;
+				--d2l-description-list-dd-margin: 0 0 0.5rem 0;
+			}
+		`;
+	}
+
+	constructor() {
+		super();
+		this._stacked = false;
+		this.breakpoint = 350;
```
dlockhart, line 63: Nit: typically we'll order these alphabetically starting with the public properties, then the private.

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,100 @@
...
+import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
+
+export const descriptionListStyles = [
+	_generateLabelStyles('dt'),
+	_generateBodyCompactStyles('dd'),
+	css`
+		:host {
+			--d2l-dt-min-width: min-content;
+			--d2l-dt-max-width: 10rem;
+			--d2l-dd-min-width: 50%;
```
dlockhart, line 13: Since these end up in the global namespace, I'd suggest prefixing them with `--d2l-description-list-*` just to avoid any future collisions or confusion.

dlockhart: Or `d2l-dl-wrapper-*` I guess to match the name of the wrapper.

jameswklassen: Gah, thought I did this. Sorry

components/description-list/description-list-wrapper.js
dbatiste: Looks great!

components/description-list/description-list-wrapper.js
```diff
@@ -0,0 +1,95 @@
...
+			'stacked': this._stacked,
+		};
+		return html`<slot class="${classMap(classes)}"></slot>`;
+	}
+
+	_onResize(entries) {
+		if (!entries || entries.length === 0) return;
+		const entry = entries[0];
+
+		if (entry.contentRect.width < this.breakpoint && !this._stacked) {
```
dbatiste, line 87: Not sure there is value in checking the `_stacked` state in these conditions.  I suppose it avoids calling the property setter, at the expense of calling the getter. It would be no-op either way.  🤷‍♂️

jameswklassen: I assumed a get operation would be cheaper than a set? But maybe this isn't the case. (?) 🤷

dbatiste: For the built-in Lit setter, I would guess that it checks the value and bails right away.  I suspect it is inconsequential but I'm not really sure. No harm in doing the check.

jameswklassen: I did some local logging/testing with [`updated(changedProperties)`](https://lit.dev/docs/components/lifecycle/#updated).

It _seeems_ like `changedProperties` only contains an update when the value changes, so I removed the `_stacked` check altogether.

components/description-list/demo/description-list-test.js
```diff
@@ -0,0 +1,136 @@
+import '../../icons/icon.js';
```
dbatiste, line 1: Is this still used?  Missing a test you wanted to include?

jameswklassen: good eye! Not used anymore since I replaced that demo

components/description-list/test/description-list.visual-diff.js
```diff
@@ -0,0 +1,44 @@
...
+		'activity-display',
+		'bulk-course-import',
+		'slotted',
+	].forEach((name) => {
+		const selector = `#${name}`;
+		[
+			799,
+			599,
+			299,
+			199,
```
jameswklassen, line 31: Since the default breakpoint to stack is <200 at the moment, need to make the test width 199 to show the stacked width.

components/description-list/demo/description-list.html
```diff
@@ -0,0 +1,51 @@
...
+			<d2l-demo-snippet fullscreen>
+				<dl>
+					<dt>Course Name</dt>
+					<dd>Brightspace 101B A</dd>
+				</dl>
+			</d2l-demo-snippet>
+
+			<h2>Short</h2>
+			<d2l-demo-snippet fullscreen>
+				<d2l-test-dl></d2l-test-dl>
```
eKoopmans, line 26: Really unfortunate that we can't show actual usage in these demos - as I understand it, the blocker is that we can't apply styles to nested slotted children, so we need to import the styles and use them in a (Lit) component.

components/description-list/README.md
```diff
@@ -0,0 +1 @@
+# Description List
```
eKoopmans, line 1: I just verified locally, it looks like the grid approach only supports 1-to-1 `dt`/`dd` pairing, right? Will be worth calling that out in the readme when you've had time to write one up.

components/description-list/demo/description-list.html
jameswklassen: Yeah, this is unfortunately one of the biggest drawbacks to using a wrapper component like we do here. This is how `d2l-table-wrapper` handles its demos as well:

https://github.com/BrightspaceUI/core/blob/8dc2e75b16f6ab9b95680667faaaae6cbfb557a8/components/table/demo/table.html#L19

components/description-list/test/description-list.test.js
```diff
@@ -0,0 +1,42 @@
...
+					<dl>
+						<dt>Title</dt>
+						<dd>Details</dd>
+					</dl>
+				</d2l-dl-wrapper>
+			`);
+
+			expect(elem.breakpoint).to.equal(300);
+		});
+	});
```
eKoopmans, line 40: I wonder if it would be useful to test the `.stacked` class in these tests - e.g. manipulate the component's width, and check that the `.stacked` class is being added/removed. Then again, the visual diffs do already test that behaviour.

eKoopmans (approved):
Looks great!

components/description-list/test/description-list.test.js
dlockhart: James and I had chatted about that and just opted to rely on the visual diff tests.

components/description-list/demo/description-list-test.js
```diff
@@ -42,6 +42,7 @@ class TestDescriptionList extends LitElement {
 	constructor() {
 		super();
 		this.type = 'default';
+		this.breakpoint = 200;
```
eKoopmans, line 45: Rather than explicitly defining this here (as well as in `d2l-dl-wrapper`), you could leave it without a default and use `ifDefined(this.breakpoint)` in the renderer.

jameswklassen: Oh yeah, whoops forgot about this! 😓

components/description-list/test/description-list.test.js
jameswklassen: Yeah, the visual diff tests should cover default/consumer overridden breakpoint behavior, so unit tests checking the same thing would probably be redundant.

components/description-list/demo/description-list.html
dbatiste: Yeah it's unfortunate. These demos are perhaps viewed less by others now that we have the Daylight site, but if we were really inclined to show something informative in the core demo page, we could:

```html
<d2l-demo-snippet code-view-hidden>
    <!-- the stuff -->
</d2l-demo-snippet>
<d2l-code-view language="javascript">
    // the stuff we really want to show
</d2l-code-view>
```

We could even add slot for it if we wanted it to look exactly the same in terms of layout.  But I feel like it's not really worth it because it becomes a maintenance issue and these demos are more for our own debugging/development purposes now.

components/description-list/description-list-wrapper.js
```diff
@@ -64,7 +64,8 @@ class DescriptionListWrapper extends LitElement {
 
 	constructor() {
 		super();
-		this.breakpoint = 200;
+		// TODO: can this be REM?
```
dbatiste, line 67: The value it's compared to is in pixels. Accepting a rem value would require converting it using root font-size that would have to be obtained from the document. Looks like our other components just accept pixels.

jameswklassen: Yeah, this was mostly a todo for myself next week to see if this was possible/feasible. Seems like PX is the convention (and what makes sense here).

components/description-list/README.md
```diff
@@ -0,0 +1,176 @@
...
+A Description List displays information in a
+vertical list of key-value pairs. Common usages could include terms & definitions, or fields & values.
```
dbatiste, line 3: Looks like there was an unintended return here.

dbatiste (approved):
Looks great!

eKoopmans (approved):
LGTM!

PR #2690, opened 2022-07-26
feat: Switch > label hover and click improvements

dcostiuc:
[Rally Story](https://rally1.rallydev.com/#/?detail=/userstory/629821423167&fdp=true)

Similar to our checkbox or radio options, we'd like the Switch component to allow users to click on the label to toggle it on/off. Hovering over the label should also trigger the hover state. 

This PR also updates the cursor to be the default one for both the switch icon/graphic and it's label (including the conditions met opener), as discussed [here](https://d2l.slack.com/archives/C03FC0GDVB7/p1658771035474229?thread_ts=1658519924.626809&cid=C03FC0GDVB7).

Note that the new `click`/`hover` changes specifically _do not_ apply to the `conditions met` opener for the visibility switch, which was how we wanted the requirements for this story to apply.

Also worth noting that when the user clicks on the label to toggle the switch, there is currently no expected output from the screen reader (unlike when the user toggles the switch icon directly). 
Adding support for some screen reader announcements was initially considered, but was deemed a use case that wasn't worth supporting (i.e. sighted users using the mouse + screen reader).

### DIFF
--- a/components/switch/switch-mixin.js
+++ b/components/switch/switch-mixin.js
@@ -215,7 +215,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 	}
 
 	get _labelContent() {
-		return html`<span @click='${this._handleClick}' @mouseover='${this._handleSwitchHover}' @mouseout='${this._handleSwitchHoverLeave}'>${this.text}</span>`;
+		return html`<span @click='${this._handleClick}' @mouseenter='${this._handleSwitchHover}' @mouseleave='${this._handleSwitchHoverLeave}'>${this.text}</span>`;
 	}
 
 	_handleClick() {
--- a/components/switch/switch-mixin.js
+++ b/components/switch/switch-mixin.js
@@ -1,11 +1,11 @@
 import '../colors/colors.js';
 import { css, html } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
 import { FocusMixin } from '../../mixins/focus-mixin.js';
 import { FocusVisiblePolyfillMixin } from '../../mixins/focus-visible-polyfill-mixin.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
-
 export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(FocusVisiblePolyfillMixin(superclass))) {
 
 	static get properties() {
@@ -29,7 +29,8 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 			 * @type {'start'|'end'|'hidden'}
 			 * @default end
 			 */
-			textPosition: { type: String, attribute: 'text-position', reflect: true }
+			textPosition: { type: String, attribute: 'text-position', reflect: true },
+			_labelHovering: { type: Boolean, state: true }
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


dcostiuc:
Technically these changes aren't specific to the `Visibility Switch` (though the Rally Story is grouped under the `Conditions Met` feature), so do we want to just merge this directly into `main`, or pool it with the other new visibility switch changes?

I think what really determines this is if @geurts would need to notify downstream teams of this change. 

Since the `cursor` for the switch icon _does_ now, in particular, change from its previous `pointer` to the new `regular`, users would potentially notice this change (or feel a slight disturbance in the Force) if a switch with these changes is used.

dcostiuc:
As a side note, I think it is interesting that none of our tests (e.g. visual diff) fail with these changes - though this is understandable given that the changes are related to `clicking`/mouse `hovering` and the `cursor`. 

Still, currently, without manual testing, we wouldn't be able to spot regressions/bugs related to this.

components/switch/switch-mixin.js
```diff
@@ -208,7 +214,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 	}
 
 	get _labelContent() {
-		return html`${this.text}`;
+		return html`<span @click='${this._handleClick}' @mouseover='${this._handleSwitchHover}' @mouseout='${this._handleSwitchHoverLeave}'>${this.text}</span>`;
```
dlockhart, line 217: Could the hover effects not have just been applied whenever the entire host element is hovered, vs. listening for mouse over/out for both the inner container and the label?

dcostiuc: So if we were to do that, i.e.
```css
:host(:hover) .d2l-switch-inner {
				border-color: var(--d2l-color-celestine);
				box-shadow: 0 0 0 1px var(--d2l-color-celestine) inset;
			}
```
then the hover styling would also apply to the entire label, including the `conditions met` part of the label in the visibility switch. 

However, I got [confirmation from Jeff](https://d2l.slack.com/archives/C03FC0GDVB7/p1658932976745289?thread_ts=1658519924.626809&cid=C03FC0GDVB7) that we don't want that to happen.

dlockhart: Gotcha. I think you can still remove the mouseover/mouseout on the `d2l-switch-inner` and modify its CSS a bit to handle its own hover state:

```css
.d2l-switch-inner:hover,
.switch-hover {
	border-color: var(--d2l-color-celestine);
	box-shadow: 0 0 0 1px var(--d2l-color-celestine) inset;
}
```

That way you only have the one set of event listeners.

components/switch/switch-mixin.js
```diff
@@ -208,7 +215,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
...
```
dbatiste, line 218: Generally, I think [mouseenter](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event) and [mouseleave](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event) would be better events to use for this purpose... combined, they act more similar to `:hover` in CSS than when using `mouseover` and `mouseout`.  There can be performance implications with deep hierarchies as well, as described in the Mozilla docs.

In this particular case it doesn't make any difference (in terms of the number of times the handlers are called) since the element only contains plain text, but it would if someone happened to add nested span to the label. It would still be minor, but in other contexts it can become more meaningful.

dbatiste (approved):
Looks good. You may or may not elect to use the other events.

PR #2694, opened 2022-07-27
Empty State Action Components

oelhanafey:
Rally tickets:
[Empty State Component Changes](https://rally1.rallydev.com/#/?detail=/userstory/642950814113&fdp=true)
[Empty State Action Components](https://rally1.rallydev.com/#/?detail=/userstory/642950820915&fdp=true)
[Documentation Updates](https://rally1.rallydev.com/#/?detail=/userstory/642951254481&fdp=true)

This PR alters the existing empty state API. It removes the previous empty state components and introduces 4 new components:

- empty-state-illustrated
- empty-state-simple
- empty-state-action-button
- empty-state-action-link

Rather than having separate empty state components for button and link actions, the new action components can be placed within the default slot of `empty-state-simple` or `empty-state-illustrated` to add the appropriate action to the component. This change was made primarily in order to support the case where an empty state has no action, but it also reduces code repetition and makes it easier to add additional action types in the future.


components/empty-state/empty-state-styles.js
```diff
@@ -40,12 +45,16 @@ export const emptyStateIllustratedStyles = css`
...
-		margin-top: 0.5rem;
+	.illustration-slot::slotted(*) {
+		display: none;
 	}
 
-	.d2l-empty-state-description {
-		margin: 0 auto 0.3rem;
+	.illustration-slot::slotted(svg:first-child) {
+		display: inline-block;
+	}
```
oelhanafey, line 54: `::slotted` can only be used to specify top level elements so the only way to control what goes in the slot is to block all children that are not svgs. This means that the `svg` itself needs to be the slot and it cannot be placed within a container like a `span` or a `div`. 

Ex:
```
// This will not render an svg
<div slot="illustration">
    <svg>
        ....
    </svg>
</div>
```

components/empty-state/empty-state-styles.js
```diff
@@ -40,12 +45,16 @@ export const emptyStateIllustratedStyles = css`
...
-		margin-top: 0.5rem;
+	.illustration-slot::slotted(*) {
+		display: none;
 	}
 
-	.d2l-empty-state-description {
-		margin: 0 auto 0.3rem;
+	.illustration-slot::slotted(svg:first-of-type) {
+		display: inline-block;
+	}
```
oelhanafey, line 54: `::slotted` can only be used to specify top level elements so the only way to control what goes in the slot is to block all children that are not svgs. This means that the `svg` itself needs to be the slot and it cannot be placed within a container like a `span` or a `div`. 

Ex:
```
// This will not render an svg
<div slot="illustration">
    <svg>
        ....
    </svg>
</div>
```

components/empty-state/empty-state-action-button.js
```diff
@@ -0,0 +1,101 @@
...
+							text=${this.text}>
+						</d2l-button-subtle>`;
+		}
+		return html`${actionButton}`;
+	}
+
+	_handleActionClick(e) {
+		e.stopPropagation();
+		this.dispatchEvent(new CustomEvent('d2l-empty-state-action', {
+			bubbles: true
```
dlockhart, line 81: Is bubbling actually required? I'd expect almost all consumers to wire up directly to the action.

components/empty-state/empty-state-action-link.js
```diff
@@ -0,0 +1,75 @@
...
+		super.firstUpdated(changedProperties);
+		this._validateAttributes();
+	}
+
+	render() {
+		const actionLink = this.text && this.href
+			? html`
+				<a class="d2l-body-compact d2l-link" href=${this.href}>
+					${this.text}
+				</a>`
```
dlockhart, line 46: Nit: I'd suggest putting this all on one line if it doesn't hurt readability too much, just to avoid the extra empty text nodes (and potential spacing impact they'd have):

```suggestion
				<a class="d2l-body-compact d2l-link" href=${this.href}>${this.text}</a>`
```

components/empty-state/empty-state-action-link.js
```diff
@@ -0,0 +1,75 @@
...
+			if (!hasHref && !this._missingHrefErrorHasBeenThrown) {
+				this._missingHrefErrorHasBeenThrown = true;
+				setTimeout(() => { throw new Error('<d2l-empty-state-action-link>: missing required "href" attribute.'); });
+			}
+
+			if (!hasText && !this._missingTextErrorHasBeenThrown) {
+				this._missingTextErrorHasBeenThrown = true;
+				setTimeout(() => { throw new Error('<d2l-empty-state-action-link>: missing required "text" attribute.'); });
+			}
+		}, 3000);
```
dlockhart, line 70: For future-Gaudi work: it'd be cool to have some kind of helper/controller/mixin/something to make doing this kind of validation easier... need to think more about how though.

components/empty-state/empty-state-action-button.js
```diff
@@ -0,0 +1,101 @@
...
+	static get styles() {
+		return css`
+					.d2l-empty-state-action {
+						vertical-align: top;
+					}
+				`;
+	}
+
+	constructor() {
+		super();
```
dlockhart, line 36: Probably want to default `this._illustrated` to `false` here?

components/empty-state/empty-state-simple.js
```diff
@@ -35,32 +30,31 @@ class EmptyStateSimpleButton extends RtlMixin(LitElement) {
 		this._validatingDescriptionTimeout = null;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('d2l-empty-state-illustrated-check', this._handleEmptyStateIllustratedCheck);
```
dlockhart, line 35: Could we just not wire up the event handler here? It'll bubble all the way to the top of the document and just get ignored, but as long as the illustrated code handles an `undefined` result (or maybe it could default it to `false` before firing the event), it should be fine.

components/empty-state/test/empty-state-illustrated.test.js
```diff
@@ -0,0 +1,68 @@
...
+					primary>
+				</d2l-empty-state-action-button>
+			</d2l-empty-state-illustrated>
+		`);
+		const button = el.querySelector('d2l-empty-state-action-button');
+
+		// Wait for _illustrated attribute to be set in d2l-empty-state-action-button
+		setTimeout(() => {
+			button.shadowRoot.querySelector('d2l-button').click();
+		}, 1000);
```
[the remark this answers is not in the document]
dlockhart: One option would be to implement `shouldUpdate` in the illustrated component and have it return `false` until that event has fired and been handled. That way the `await fixture()` call above won't actually return until everything is ready.

components/empty-state/empty-state-simple.js
oelhanafey: Yeah, I considered this as well but thought it better to not have the event bubble up to the rest of the document. If the event bubbling isn't a big deal, then it makes sense to simplify it.

dlockhart: It's really low cost, and actually probably cheaper than handling it here. Mostly I'm thinking if we have a bunch of other empty-state types or actions in the future, we wouldn't want to filth them all up with this event handling code that really only makes sense in the illustrated + button case.

components/empty-state/empty-state-action-button.js
oelhanafey: You're right, that makes more sense. I wrote the documentation that way but when I was writing this I was thinking that consumers would be listening for events on the container empty state component.

components/empty-state/test/empty-state-illustrated.test.js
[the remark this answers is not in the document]
oelhanafey: Ah, I like that! That would be a nice way to do it

components/empty-state/empty-state-action-link.js
oelhanafey: Yeah, I have repeated this logic on all of my empty state components and this validation would be a useful addition to all of our components with required attributes. 

I will think more about this.

dlockhart (approved):
This is looking good to me!

PR #2698, opened 2022-07-27
Add dynamic typography style generation helpers

jameswklassen:
## Description

The current approach taken in #2688 requires elements to be styled without adding a class.

This can be achieved by generating styles for a specific CSS selector. ([Lit docs](https://lit.dev/docs/components/styles/#expressions) on using expressions in styles)

Looking for any feedback/thoughts/concerns/gotchas 😄 

## Known limitations/issues
- Wrapping the selector in `unsafeCSS`
  - Ideally, we'd use `css` string (like in the [docs](https://lit.dev/docs/components/styles/#expressions)), but Stylelint can't parse since `css'.class'` is not _technically_ a valid css block.
  - We could mitigate potential issues with `unsafeCSS` by writing validation to ensure the string is safe

### DIFF
--- a/components/typography/styles.js
+++ b/components/typography/styles.js
@@ -1,6 +1,13 @@
 import '../colors/colors.js';
 import { css, unsafeCSS } from 'lit';
 
+export const _isValidCssSelector = (selector) => {
+	const re = /[#.]?([a-zA-Z0-9-_]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_]+)?/g;
+	const match = selector.match(re);
+
+	return !!match && match.length === 1 && match[0].length === selector.length;
+};
+
 export const bodyStandardStyles = css`
 	.d2l-body-standard {
 		font-size: 0.95rem;
@@ -45,6 +52,8 @@ export const bodyStandardStyles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateBodyCompactStyles = (selector) => {
+	if (!_isValidCssSelector(selector)) return;
+
 	selector = unsafeCSS(selector);
 	return css`
 		${selector} {
@@ -224,6 +233,8 @@ export const heading4Styles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateLabelStyles = (selector) => {
+	if (!_isValidCssSelector(selector)) return;
+
 	selector = unsafeCSS(selector);
 	return css`
 		${selector} {
--- a/components/typography/styles.js
+++ b/components/typography/styles.js
@@ -1,5 +1,5 @@
 import '../colors/colors.js';
-import { css } from 'lit';
+import { css, unsafeCSS } from 'lit';
 
 export const bodyStandardStyles = css`
 	.d2l-body-standard {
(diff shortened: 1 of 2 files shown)


components/typography/styles.js
```diff
@@ -41,26 +41,31 @@ export const bodyStandardStyles = css`
...
-		max-height: 2.4rem;
-	}
-	:host([skeleton]) .d2l-body-compact.d2l-skeletize-paragraph-3 {
-		max-height: 3.6rem;
-	}
-	:host([skeleton]) .d2l-body-compact.d2l-skeletize-paragraph-5 {
-		max-height: 6rem;
-	}
-`;
+export const generateBodyCompactStyles = (selector) => {
```
dlockhart, line 44: As we discussed outside this review, I think I'd like to see this go through a `validateSelector()` step before being passed to `unsafeCSS`. It could use a simple regular expression to make sure the value is either a CSS class, id or tag name.

dbatiste: Being explicit might be the safest.  If we need to support complex selectors including combinators (ex. `.foo .bar`), [this](https://stackoverflow.com/questions/12575845/what-is-the-regex-of-a-css-selector) might provide some ideas on what to look for.

dbatiste:
I wonder if there is a difference in terms of the amount of CSS added to the page if a consumer imports and uses `bodyCompactStyles` vs `generateBodyCompactStyles` for their specific selector.  i.e. if three different web components import and include `bodyCompactStyles`, does one instance of those styles get added to the page and adopted by each component, or do they get included three times. Just thinking of future guidance we would give... whether one approach would be preferred.

components/typography/styles.js
dbatiste: Just thinking a little more on the complex selector case.

Give this:
```css
:host([skeleton]) ${selector}.d2l-skeletize::before
```

This would be no good if the consumer tried to do:
```javascript
generateLabelStyles(':host .some-element')
```

jameswklassen:
> I wonder if there is a difference in terms of the amount of CSS added to the page if a consumer imports and uses `bodyCompactStyles` vs `generateBodyCompactStyles` for their specific selector. i.e. if three different web components import and include `bodyCompactStyles`, does one instance of those styles get added to the page and adopted by each component, or do they get included three times. Just thinking of future guidance we would give... whether one approach would be preferred.

This is a good point. I'm not sure how many times the styles get added, I'll try and investigate this a bit.

dlockhart:
I'd theorize that it would get added once for each unique string of CSS...

components/typography/styles.js
dlockhart: Good point -- agreed.

jameswklassen:
> I'd theorize that it would get added once for each unique string of CSS...

That would be my guess too, but I haven't been able to produce a test that confirms/denies this hypothesis.

components/typography/test/styles.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+	it('should not support complex selectors', () => {
+		expect(_isValidCssSelector('dl dd')).to.be.false;
+		expect(_isValidCssSelector('dt.class')).to.be.false;
+		expect(_isValidCssSelector('dt#id')).to.be.false;
+		expect(_isValidCssSelector('dl .class')).to.be.false;
+		expect(_isValidCssSelector('dl #id')).to.be.false;
+	});
+
+	it('should not support invalid selectors', () => {
```
dbatiste, line 36: Worth a test that checks for a CSS block `.some-class{...}`?  Looks like they can't do anything crazy like passing in `@import` or `@media`, and the `@` case sorta pokes at those cases.

dbatiste:
Looking pretty good!

dlockhart (approved):
I'm happy with this!

components/typography/test/styles.test.js
jameswklassen: Good call. I've added a few more tests.

dbatiste (approved):

PR #2699, opened 2022-07-27
Simplify html-block and MathJax rendering to use Lit and avoid temp element creation

grant-cleary:
The goal with this PR is to try to simplify some of the core html-block rendering logic (and the MathJax renderer logic) to align more with our standard Lit components and avoid creating chains of temp elements that could cause trouble for us.

To do this, we use Lit to render the barebones render container and the slot we use for stamping, but we still do all of the rendering itself manually outside of the Lit rendering lifecycle. This is required because of how MathJax in particular needs the elements it's working on to be attached to the DOM when it does its work. i.e. We need the html-block contents to be stamped into the shadowRoot before we can hand off to MathJax for rendering.


components/html-block/html-block.js
```diff
@@ -209,35 +214,51 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+		`;
 	}
 
-	updated() {
-		super.updated();
-		if (this._contextObserverController && this._contextObjectHasChanged()) this._render();
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (this._contextChanged()) {
+			this._render();
```
grant-cleary, line 241: There's some jank here because (again) things get really complicated if we need to dynamically change MathJax's configuration. So this logic works in a general sense, but once the MathJax context object is set to something valid the first time, subsequent changes won't accomplish anything.

This isn't ideal, but it's not a change in behaviour and I don't want to muddy this PR with additional work to support dynamically changing MathJax's configuration.

components/html-block/html-block.js
```diff
@@ -274,20 +293,14 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
-				let temp = document.createElement('div');
-				temp.style.display = 'none';
-				nodes.forEach(node => temp.appendChild(node.cloneNode(true)));
+		const renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
 
-				this._renderContainer.appendChild(temp);
-				temp = await this._processRenderers(temp);
-				this._renderContainer.innerHTML = temp.innerHTML;
+		const stampHTML = async nodes => {
+			renderContainer.innerHTML = '';
```
grant-cleary, line 299: Wipe the render container before we start stamping. We can't assume that whatever was rendered here before stamping is still valid. Prism and MathJax both effectively need us to start from a clean slate, for example.

components/html-block/html-block.js
```diff
@@ -274,20 +293,14 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+			if (!nodes || nodes.length === 0) return;
 
-			} else {
-				this._renderContainer.innerHTML = '';
-			}
+			nodes.forEach(node => renderContainer.appendChild(node.cloneNode(true)));
+			await this._processRenderers(renderContainer);
```
grant-cleary, line 303: It's important that we stamp the slot contents into the render container **before** we run any of the renderers. MathJax in particular requires that these be attached to the DOM or they won't typeset properly.

helpers/mathjax.js
```diff
@@ -129,12 +113,18 @@ export function loadMathJax(mathJaxConfig) {
 				const AbstractHandler = window.MathJax._.core.Handler.AbstractHandler.prototype;
 				const startup = window.MathJax.startup;
 
+				const getFirstChild = doc => {
+					const child = doc.firstChild;
+					if (!child || child.nodeType === Node.ELEMENT_NODE) return child;
+					else return child.nextElementSibling;
+				};
```
grant-cleary, line 120: This handles the Lit comment problem. The reason that exploded in our custom adapter here is because the adaptor is expecting a structure that looks exactly like this:
```
<div class="d2l-html-block-rendered">
   <mjx-doc>
      <mjx-head>...</mjx-head>
      <mjx-body>...</mjx-body>
   </mjx-doc>
</div>
```
Effectively, we use these `mjx-{structure}` nodes to simulate an HTML document for MathJax to process. Lit messes things up here because it produces a structure like this (note the comment):
```
<div class="d2l-html-block-rendered">
   <!--lit$1425314-->
   <mjx-doc>...</mjx-doc>
</div>
```
So the solution implemented here is to just deliberately ignore non-element nodes when our adaptor parses the document structure.

helpers/mathjax.js
```diff
@@ -61,35 +61,19 @@ export class HtmlBlockMathRenderer {
...
-		temp.attachShadow({ mode: 'open' });
-		temp.shadowRoot.innerHTML = `<div><mjx-doc><mjx-head></mjx-head><mjx-body>${inner}</mjx-body></mjx-doc></div>`;
+		elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
+			elm.setAttribute('style', 'display: block; height: 0.5rem;');
+		});
 
-		elem.appendChild(temp);
+		// If we're using deferred rendering, we need to create a document structure
+		// within the element so MathJax can appropriately process math.
+		if (!options.noDeferredRendering) elem.innerHTML = `<mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc>`;
```
grant-cleary, line 70: Unfortunately we can't really avoid tampering with the element's contents again here because we do need to build this structure so our custom adaptor has a "document" to hook onto.

helpers/mathjax.js
```diff
@@ -129,12 +113,18 @@ export function loadMathJax(mathJaxConfig) {
...
+				const getFirstChild = doc => {
+					const child = doc.firstChild;
+					if (!child || child.nodeType === Node.ELEMENT_NODE) return child;
+					else return child.nextElementSibling;
+				};
+
 				//
 				//  Extend HTMLAdaptor to handle shadowDOM as the document
 				//
 				class ShadowAdaptor extends HTMLAdaptor {
```
grant-cleary, line 125: This implementation does keep our custom adaptor. Brief forays into removing it have so far proven fruitless, so I'm not optimistic we can ditch it. MathJax just stores too much document-level state. However, I will keep poking at it to see if there's a path forward where we can just move to their default APIs.

components/html-block/test/html-block.test.js
```diff
@@ -52,7 +52,7 @@ class TestAsyncRenderer {
 
 class TestNoInlineRenderer {
 	async render(elem, options) {
-		if (options.noDeferredRendering) return;
+		if (options.noDeferredRendering) return elem;
```
grant-cleary, line 55: This wasn't failing any tests because it was the last renderer in the chain, but not returning the element in this case definitely breaks things now that I've added the extra renderer. 😅

helpers/mathjax.js
grant-cleary: Actually, this isn't true now that I'm wiping out the container's inner HTML before stamping. Let me see if I can remove this.

components/html-block/html-block.js
```diff
@@ -161,16 +163,19 @@ class HtmlBlock extends RtlMixin(LitElement) {
 				text-align: left;
 			}
 			:host([inline]),
-			:host([inline]) div.d2l-html-block-rendered {
+			div.d2l-html-block-inline {
```
dbatiste, line 166: Are these selectors overqualified? Could they just be:

```css
.d2l-html-block-inline {
   ...
}
.d2l-html-block-no-deferred-rendering {
   ...
}
.d2l-html-block-deferred-rendering {
   ...
}
```
Admittedly though, the `slot` element selector provides some hint/clarity.

grant-cleary: Yeah, they could be simplified. I always struggle at least a bit with how specific to get to avoid potential issues or pitfalls later. I'm content with whatever you would prefer.

dbatiste:
Just an FYI regarding `_renderInline` - it is now safe to use [assignedElements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements). You'd still need to find the `DIV`, but filtering on `Node.ELEMENT_NODE` would no longer be necessary if using the element-specific API.

grant-cleary:
> Just an FYI regarding `_renderInline` - it is now safe to use [assignedElements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements). You'd still need to find the `DIV`, but filtering on `Node.ELEMENT_NODE` would no longer be necessary if using the element-specific API.

Hmm, is that true? A few weeks ago I was told we still didn't want to break Safari 11.

components/html-block/html-block.js
dbatiste: I'm good either way re the use of the `:host([no-deferred-rendering])` - there is a pro each way imo.

If keeping with the way it is, I'd reduce it to the minimum specificity that's required to avoid unnecessary selector checks (since `element.class-name` is in theory, two selector checks, ignoring possible optimizations the browser may or may not be doing). 🤷‍♂️

components/html-block/html-block.js
dbatiste: Sounds good.  Out of curiosity, do we envision there ever being a use case for it?

components/html-block/html-block.js
grant-cleary: Seems good. The slot has the opposite condition, which I could either accomplish with `:host(:not([no-deferred-rendering])) slot` or by keeping it as is (or I guess by using two separate selectors). Do you have a particular preference there? I think using the generated conditional class name means one selector and otherwise we need two, but... 🤷

components/html-block/html-block.js
grant-cleary: I have a hard time thinking of one. The big reason we have to handle this via a controller is because in FRAs in particular, it's possible for the component to load before ifrau syncs down the attributes from the host page. However, because the MathJax renderer doesn't instantiate MathJax at all if there's no context object present, that case is covered for us and MathJax will init properly when it's first populated.

So... The use case I think we actually care about now isn't impacted by this long-term "caching" of sorts and I'm not sure there's a use case for adjusting the config on the fly.

components/html-block/html-block.js
dbatiste: Almost worth include a comment to that effect... `// must attach nodes before processing renderer plugins since some require connected nodes (ex. MathJax)"`.

components/html-block/html-block.js
grant-cleary: I have a hard time thinking of one. The big reason we have to handle this via a controller is because in FRAs in particular, it's possible for the component to load before ifrau syncs down the attributes from the host page. However, because the MathJax renderer doesn't instantiate MathJax at all if there's no context object present, that case is covered for us and MathJax will init properly when it's first populated.

So... The use case I think we actually care about now isn't impacted by this long-term "caching" of sorts and I'm not sure there's a use case for adjusting the config on the fly.

components/html-block/html-block.js
grant-cleary: Good call, yeah, otherwise this requirement isn't at all obvious and someone could easily come along and switch the order without realizing it's going to subtly break stuff.

components/html-block/test/html-block.test.js
```diff
@@ -135,4 +161,16 @@ describe('d2l-html-block', () => {
 		expect(spans[1].innerHTML).to.equal('second');
 	});
 
+	it('should do multiple renderer replacements', async() => {
```
dbatiste, line 164: 🎉 ... I think this is a good additional test.

What are your thoughts on a unit or visual-diff test that includes both the code formatting and the math rendering? Such a test would at least ensure they are properly returning the element, which is currently a requirement of `d2l-html-block` unless we change the API such that plugin renderers are always required to operate on the same element, and not create a new one... not sure if that is possible though. Of course, renderers that are external to core (like @mentions) could still blow things up and would not get caught here, as previously mentioned elsewhere.

helpers/mathjax.js
```diff
@@ -61,35 +61,19 @@ export class HtmlBlockMathRenderer {
...
-		}
-
-		const inner = elem.innerHTML.replace(/<mspace linebreak="newline">/gi, `<mspace linebreak="newline" style="${lineBreakStyle}">`);
-
-		const temp = document.createElement('div');
-		temp.style.display = 'none';
-		temp.attachShadow({ mode: 'open' });
-		temp.shadowRoot.innerHTML = `<div><mjx-doc><mjx-head></mjx-head><mjx-body>${inner}</mjx-body></mjx-doc></div>`;
+		elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
+			elm.setAttribute('style', 'display: block; height: 0.5rem;');
```
dbatiste, line 65: I know this is what we were already doing, but is there a chance that we're blowing away properties that already exist on `mspace`? If so, is it intentional, or should we be using `elm.style.display = 'block';` etc.?

helpers/mathjax.js
```diff
@@ -61,35 +61,19 @@ export class HtmlBlockMathRenderer {
...
+		// within the element so MathJax can appropriately process math.
+		if (!options.noDeferredRendering) elem.innerHTML = `<mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc>`;
 
 		await window.MathJax.startup.promise;
-		renderingPromise = renderingPromise.then(() => window.MathJax.typesetShadow(temp.shadowRoot));
+		renderingPromise = renderingPromise.then(() => window.MathJax.typesetShadow(elem.getRootNode(), elem));
 		await renderingPromise;
 
-		return temp.shadowRoot.firstChild;
+		return elem;
```
dbatiste, line 76: If I follow this correctly, this now returns the same element that was passed in. Is that correct?

helpers/mathjax.js
dbatiste: I think the comment node is what was blowing up when I first tried to use Lit rendering with MathJax in the very beginning.  Glad we found a solution.  :)

components/html-block/html-block.js
dbatiste: I don't feel strongly either way. I'm mostly focused on reducing the selectors to what's necessary, whichever way it's written. I do like relying on single source of truth (the host attribute) and avoiding conditional rendering which adds very slight complexity at the expense of perhaps an extra bit of CSS. I think the `:not()` could be avoided, but I really think it's splitting hairs we can't even measure.

helpers/mathjax.js
grant-cleary: It feels a bit fragile, in the sense that if Lit ever decides to put comments elsewhere this might explode... But one of the things that appealed to me about MathJax 3.2 is that they added proper handling for comments _within_ the document body they're parsing. So I think this special handling for direct children of the shadowRoot is all we need.

components/html-block/html-block.js
grant-cleary: Yeah, fair enough.

helpers/mathjax.js
grant-cleary: `mspace` shouldn't have any additional properties according to spec, but I think your suggestion is safer regardless so I'll just go with that.

helpers/mathjax.js
grant-cleary: That's right, yeah.

components/html-block/test/html-block.test.js
grant-cleary: I certainly think there's value here. Unit test might be harder because I'm not quite sure how easy it is for us to actually await the rendering before asserting anything. The html-block doesn't currently broadcast when it's done its work. Visual diff is easier since we're inevitably tossing waits around anyways.

grant-cleary:
Going to introduce the visual diff test in a separate commit.

helpers/mathjax.js
dbatiste: If it's invalid otherwise then it's all good. If there is a chance to unexpected clobber, then we should play it safe.

grant-cleary: Yeah, it shouldn't be valid for there to be styling on there because MathJax will just wipe it out anyways, but it's a trivial change and strictly safer, so... 🤷

helpers/mathjax.js
grant-cleary: Technically I suppose that makes the return value completely pointless, since the reference to it in the html-block rendering code will just implicitly update.

dbatiste: Yeah, and also a point of failure as we've been chatting about.  If all the renderers currently return the element it was passed, maybe `d2l-html-block` shouldn't expect it to be returned, and instead assume that all processing will be done on the element it passes. Not sure if this would be limiting in the future, as the previous version of MathJax renderer required the ability to return a new element. 🤔

components/html-block/test/html-block.test.js
dbatiste: If visual-diff covers it, I wouldn't create a unit test.

helpers/mathjax.js
grant-cleary: Yeah, that was my main thought. Certainly right now the existing renderers act on the elements as they are, and enforcing that in the component makes it very obvious that we're expecting renderers to run inline. I don't think it's particularly difficult to reverse course later if we want. Realistically we would just switch the renderers to start returning the element and then adjust the html-block to start consuming it again. I'm certainly fine removing the return value for now though.

grant-cleary: Ok, I made this change in the core renderers and the html-block component itself. It will need to be reflected in the @mentions renderer as well, but because that's in a different repo that change will have to wait until post-merge.

dbatiste: Cool.  Thanks for updating that.  It would be quite unfortunate if a plugin renderer forgot to return the element and in effect broken the component or other renderers.

grant-cleary: Haha yeah, I was trying to think if any renderer could have a valid use case for wanting to return something, but realistically if a renderer needs to detach the element from the DOM or something, it's also going to need to add it back as well. At least this way the requirement is super clear.

helpers/mathjax.js
```diff
@@ -61,35 +61,24 @@ export class HtmlBlockMathRenderer {
...
-
-		elem.appendChild(temp);
+		elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
+			elm.style.display = 'block';
+			elm.style.height = '0.5rem';
+		});
 
 		await window.MathJax.startup.promise;
-		renderingPromise = renderingPromise.then(() => window.MathJax.typesetShadow(temp.shadowRoot));
+		renderingPromise = renderingPromise.then(() => window.MathJax.typesetPromise([elem]));
```
grant-cleary, line 70: I was curious to see if this works and it seems to now. I think the trick is the promise. MathJax grabs the root node of whatever elements it's working on and stores that information globally. I think if multiple renders collided, you could get into a situation where that document reference could change mid-render and MathJax would start typesetting things it shouldn't have typeset.

helpers/mathjax.js
```diff
@@ -61,35 +61,24 @@ export class HtmlBlockMathRenderer {
...
-		return temp.shadowRoot.firstChild;
+		// If we're using deferred rendering, we need to manually attach MathJax's styles to the DOM.
+		if (options.noDeferredRendering) return;
+
+		const styleElm = window.MathJax.chtmlStylesheet().cloneNode(true);
+		styleElm.id = 'd2l-mathjax-styles';
+
+		if (!elem.querySelector(`#${styleElm.id}`)) {
+			elem.appendChild(styleElm);
+		}
```
grant-cleary, line 81: @dbatiste I was musing about Lit's `adoptStyles` here when you sent me that DM about it. In theory we could use that here to attach those styles directly to the `shadowRoot` if we so wished.

dbatiste:
This is looking pretty good.  Is there more MathJax stuff that you are hoping to change?

grant-cleary:
> This is looking pretty good. Is there more MathJax stuff that you are hoping to change?

I'm taking a brief look at whether it's possible to get rid of the `mjx-doc` structure (though I'm not optimistic). Beyond that, I think this will probably be good to go.

grant-cleary:
> I'm taking a brief look at whether it's possible to get rid of the `mjx-doc` structure (though I'm not optimistic). Beyond that, I think this will probably be good to go.

Yeah, this was a wash. As far as I'm concerned this PR is likely in its end state.

grant-cleary:
Deliberately not going to merge this until 20.22.10 branching so the MathJax version bump has a month to soak first.

dbatiste (approved):

PR #2702, opened 2022-07-28
Localization fixes for d2l-pager-load-more

dbatiste:
This PR updates the `d2l-pager-load-more` component to use `formatNumber` for formatting counts.  It also updates the lang-terms to use the plural type for formatting.

### DIFF
--- a/components/paging/pager-load-more.js
+++ b/components/paging/pager-load-more.js
@@ -107,7 +107,7 @@ class LoadMore extends FocusMixin(FocusVisiblePolyfillMixin(LocalizeCoreElement(
 				${this.itemCount > -1 ? html`
 					<span class="d2l-offscreen">${getSeparator({ nonBreaking: true })}</span>
 					<span class="separator"></span>
-					<span class="info">${this.localize('components.pager-load-more.info', { showingCount: formatNumber(this.itemShowingCount), totalCount: formatNumber(this.itemCount) })}</span>
+					<span class="info">${this.localize('components.pager-load-more.info', { showingCount: formatNumber(this.itemShowingCount), totalCount: this.itemCount, totalCountFormatted: formatNumber(this.itemCount) })}</span>
 				` : nothing}
 			`}
 		</button>
--- a/components/paging/pager-load-more.js
+++ b/components/paging/pager-load-more.js
@@ -5,6 +5,7 @@ import { buttonStyles } from '../button/button-styles.js';
 import { findComposedAncestor } from '../../helpers/dom.js';
 import { FocusMixin } from '../../mixins/focus-mixin.js';
 import { FocusVisiblePolyfillMixin } from '../../mixins/focus-visible-polyfill-mixin.js';
+import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
 import { getFirstFocusableDescendant } from '../../helpers/focus.js';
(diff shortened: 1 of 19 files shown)


components/paging/pager-load-more.js
```diff
@@ -102,11 +103,11 @@ class LoadMore extends FocusMixin(FocusVisiblePolyfillMixin(LocalizeCoreElement(
 				<span class="d2l-offscreen" role="alert">${this.localize('components.pager-load-more.status-loading')}</span>
 				<d2l-loading-spinner size="24"></d2l-loading-spinner>
 			` : html`
-				<span class="action">${this.localize('components.pager-load-more.action', { count: this.pageSize })}</span>
+				<span class="action">${this.localize('components.pager-load-more.action', { count: formatNumber(this.pageSize) })}</span>
```
dbatiste, line 106: While this produces the result that I want because I am only ever passing integer values greater than zero, it looks like passing formatted numbers into the localization lib can potentially be problematic. Given a lang-term that is using the plural type syntax:
```
{totalCount, plural, one {{showingCount} of {totalCount} item} other {{showingCount} of {totalCount} items}}
```

In Danish:
* `this.localize('components.pager-load-more.action', { count: 0.1 }` yields the singular format (correct)
* `this.localize('components.pager-load-more.action', { count: formatNumber(0.1) })` yields the other format (incorrect)

This is because `formatNumber` for Danish yields "0,1", which does not match the singular values:

image

As mentioned, this isn't problematic for the pager per se, but passing formatted values seems incompatible with plural syntax types in ICU messages @dlockhart , @bearfriend , do you have any insight into how such a scenario should be handled?

dlockhart: Yikes yeah, that's not great. I'm guessing with larger numbers that might result in group separators (e.g. `1,000,000`) we could be in trouble too.

I wonder if the `pluralize` function itself does any number formatting on its own? We don't use [the number type](https://formatjs.io/docs/core-concepts/icu-syntax#number-type) because obviously we have our own number formatting rules, but `pluraize` might use it under the cover.

bearfriend: I think passing in formatted numbers to the subject of the `plural` is probably a bad idea. Let it do its logic on `totalCount` but use a formatted variable for rendering:
```
{totalCount, plural, one {{showingCount} of {totalCountFormatted} item} other {{showingCount} of {totalCountFormatted} items}}
```
Then:
```
this.localize('components.pager-load-more.info', { showingCount: ..., totalCount: 0.1, totalCountFormatted: formatNumber(0.1) })
```

dlockhart: That's an excellent suggestion!

bearfriend: It's really just the same as this, but up one layer since we have to use our own formatting:
```
{totalCount, plural, one {{showingCount} of {totalCount, number} item} other {{showingCount} of {totalCount, number} items}}
```

We probably could make formatters from our custom stuff to use this way.

dbatiste: Yep, that's excellent.  Much thanks! 

Adapting our custom formatting to something it can use might be an interesting thing to look into in the future.

dlockhart (approved):

PR #2705, opened 2022-07-29
Removed focus-visible polyfill from menu component

oelhanafey:
@grant-cleary brought up that the focus-visible polyfill is causing performance issues by dispatching too many `visibilitychange` events. The polyfill was used since old versions of Safari do not support the `focus-visible` pseudo-selector. As of Safari 15.4 (March 2022 release), `focus-visible` is now supported.

This PR removes the `focus-visible` polyfill and replaces it with the pseudo-selector. This will cause focus issues with keyboard navigation on older version of Safari. The number of people affected by this change would be relatively small and will continue to decrease, so we decided that it was worth the tradeoff for performance improvements.


components/menu/menu-item-styles.js
```diff
@@ -27,7 +25,23 @@ export const menuItemStyles = css`
 		z-index: 2;
 	}
 
-	:host([disabled]), :host([disabled]:hover), :host([disabled].focus-visible) {
+	/** separated because safari is having trouble parsing these */
```
dlockhart, line 28: Maybe mention in these comments that it's just Safari <15.4 that has trouble with these entries? That way when/if we eventually stop caring about those versions we can decide to re-combine them.

dlockhart (approved):
LGTM!

components/menu/menu-item-styles.js
oelhanafey: Yep, good point!

grant-cleary (approved):
Thanks much, Oscar! :)

eKoopmans:
Associated Rally:
https://rally1.rallydev.com/#/?detail=/userstory/643401699741&fdp=true

PR #2709, opened 2022-08-02
Fixup more-less tests

svanherk:
Putting this up separately from the other tests because I expect visual-diff changes here.  I noticed the expanded cases weren't actually testing when the `more-less` was `expanded`.   This PR:
- Fixes the `expanded` tests
- Adds a couple missing scenarios
- Makes these tests rerunnable


dlockhart (approved):
Nice!

PR #2711, opened 2022-08-02
feat: OverflowGroupMixin

margaree:
This rips out shareable overflow group functionality into an `OverflowGroupMixin`. The original `d2l-overflow-group` still functions as is for buttons etc.

Some fairly straightforward changes were needed to be made for the original `d2l-overflow-group` functionality. These can be seen in the first commit. The other commits then divide some pieces out into the mixin and leave component specific functionality in `d2l-overflow-group`.

The commit [filter-overflow-group POC](https://github.com/BrightspaceUI/core/commit/ed01015ffae04d63f82456f320914ceb924bc1f7) shows how this will work for filters. I'll be deleting that prior to merge but it will basically be reused for the new component.

To do (both for once this PR seems good to go):
- [x] Update README
- [x] Remove filter POC


components/overflow-group/overflow-group.js
```diff
@@ -428,48 +466,20 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-		</d2l-dropdown-button>`;
+		this._dropdownItems = this._slotItems.map((node) => convertToOverflowItem(node));
 	}
 
-	_getSlotItems() {
-		const nodes = this._buttonSlot.assignedNodes({ flatten: true });
-		const filteredNodes = nodes.filter((node) => {
-			const isNode = node.nodeType === Node.ELEMENT_NODE && node.tagName.toLowerCase() !== 'template';
+	async _getSlotItems() {
+		const filteredNodes = await filterAsync(this._buttonSlot.assignedNodes({ flatten: true }), async(node) => {
```
margaree: This follows the same pattern used with `tag-list` (https://github.com/BrightspaceUI/core/blob/main/components/tag-list/tag-list.js#L311) in order to make sure a node is ready before we size it.

components/overflow-group/overflow-group.js
```diff
@@ -81,7 +90,7 @@ function createMenuItemMenu(node) {
 	</d2l-menu-item>`;
 }
 
-function convertToDropdownItem(node) {
+function convertToOverflowItem(node) {
```
margaree: Given a slightly more generic name since this was a function that would then be created by the components using the mixin

components/overflow-group/overflow-group.js
```diff
@@ -271,7 +280,7 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 	}
 
 	render() {
-		const overflowMenu = this._getOverflowMenu();
+		const overflowMenu = this.getOverflowMenu();
```
margaree: `getOverflowMenu` is the other function that components using the mixin have to create

components/overflow-group/overflow-group.js
```diff
@@ -311,6 +320,35 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
 		}
 	}
 
+	getOverflowMenu() {
```
margaree, line 157: This was moved as is

components/overflow-group/overflow-group.js
```diff
@@ -494,16 +504,16 @@ class OverflowGroup extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-			this._getItems();
+		requestAnimationFrame(async() => {
+			await this._getItems();
 
 			this._slotItems.forEach(item => {
 				const observer = new MutationObserver(this._handleItemMutation);
 				observer.observe(item, {
 					attributes: true, /* required for legacy-Edge, otherwise attributeFilter throws a syntax error */
-					attributeFilter: ['disabled', 'text'],
+					attributeFilter: ['disabled', 'text', 'selected'],
```
margaree: These changes were made due to filter - we want to trigger the mutation observer when the `selected` attribute changes in the `d2l-filter-dimension-set-value`s

components/overflow-group/filter-overflow-group.js
```diff
@@ -0,0 +1,62 @@
...
+
+		return html`
+			<d2l-filter class="${OVERFLOW_DROPDOWN_CLASS}" @d2l-filter-change="${this._handleFilterChange}">
+				${overflowItems}
+			</d2l-filter>
+		`;
+	}
+
+	_handleFilterChange(e) {
+		const target = (e.target.classList && e.target.classList.contains(OVERFLOW_DROPDOWN_CLASS)) ? this : e.target;
```
margaree, line 48: The most complicated piece of this is that when:
a) a `d2l-filter-dimension-set-value` (NOT in the overflow filter) changes its `selected` status, the overflow menu version needs to be updated (mutation observer needs to be triggered). In `filter` the list item `selected` state gets update but not the original `d2l-filter-dimension-set-value` so this updates that.
b) a `d2l-filter-dimension-set-value` (in the overflow filter) changes its `selected` status, the original `d2l-filter-dimension-set-value` needs to have its `selected` status updated

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,351 @@
...
+async function filterAsync(arr, callback) {
+	const fail = Symbol();
+	const results = await Promise.all(arr.map(async item => {
+		const callbackResult = await callback(item);
+		return callbackResult ? item : fail;
+	}));
+	return results.filter(i => i !== fail);
+}
+
+export const OverflowGroupMixin = superclass => class extends RtlMixin(LocalizeCoreElement(superclass)) {
```
margaree, line 27: This file is generally the same as the original `d2l-overflow-group`, just with the changes made in the first commit

components/overflow-group/overflow-group.js
```diff
@@ -61,104 +50,15 @@ function createMenuItemSeparator() {
...
-			 * Set the opener type to 'icon' for a `...` menu icon instead of `More actions` text
-			 * @type {'default'|'icon'}
-			 */
-			openerType: {
-				type: String,
-				attribute: 'opener-type'
-			},
 			/**
 			 * Setting this property will change the style of the overflow menu opener
 			 * @type {'default'|'subtle'}
```
margaree, line 64: `openerStyle` currently only applies to `d2l-overflow-group` and not to the future `d2l-filter-overflow-group` so I've left it only here.

components/overflow-group/filter-overflow-group.js
```diff
@@ -0,0 +1,62 @@
...
+	const dimensionSets = node.querySelectorAll('d2l-filter-dimension-set');
+	const clones = Array.from(dimensionSets).map((set) => set.cloneNode(true));
+	return clones;
+}
+
+/**
+ * A component that can be used to display a group of filters that will be put into an overflow filter when they no longer fit on the first line of their container
+ * @slot - d2l-filters to be added to the container
+*/
+class OverflowGroupFilter extends OverflowGroupMixin(LitElement) {
```
margaree, line 15: This file will be deleted before merge and is to just show an example of how this would work.

Consumers of `OverflowGroupMixin` would need to implement `convertToOverflowItem` (specifies what a chomped item becomes in the overflow "dropdown") and `getOverflowMenu` (specifies the chomp overflow menu)

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+			 */
+			autoShow: {
+				type: Boolean,
+				attribute: 'auto-show',
+			},
+			/**
+			 * @ignore
+			 */
+			chompIndex: {
+				type: Number,
```
dlockhart, line 42: Could this become `{ state: true }` so that it's just completely hidden?

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+				reflect: true,
+				attribute: 'max-to-show',
+			},
+			/**
+			 * @ignore
+			 */
+			mini: {
+				type: Boolean,
+				reflect: true
```
dlockhart, line 67: Ditto here around using state, although then it wouldn't reflect so if that's actually needed, another mechanism would be required.

dlockhart: Hmm... this seems to just get set based on the opener type, so might not even need to be a property at all.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+	async firstUpdated() {
+		super.firstUpdated();
+
+		// selected elements
+		this._buttonSlot = this.shadowRoot.querySelector('slot');
+
+		this._container = this.shadowRoot.querySelector('.d2l-overflow-group-container');
+
+		this._resizeObserver = new ResizeObserver(this._throttledResize);
+		this._resizeObserver.observe(this._container);
```
dlockhart, line 130: I think we normally wire these up in `connectedCallback`, since `firstUpdated` won't get called again when something is removed and then re-added to the DOM.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+	update(changedProperties) {
+		super.update(changedProperties);
+
+		if (changedProperties.get('autoShow')) {
+			this._getItemLayouts(this._slotItems);
+			this._autoDetectBoundaries(this._itemLayouts);
+		}
+
+		if (changedProperties.get('minToShow')
+			|| changedProperties.get('maxToShow')) {
```
dlockhart, line 161: Is the intent here to re-run `_chomp()` if `minToShow` or `maxToShow` changed, or if they changed to a value > 0? `.get` returns the value, whereas `.has` returns whether it's in the collection.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+				<slot @slotchange="${this._handleSlotChange}"></slot>
+				${overflowMenu}
+			</div>
+		`;
+	}
+
+	update(changedProperties) {
+		super.update(changedProperties);
+
+		if (changedProperties.get('autoShow')) {
```
dlockhart, line 155: This will only run if `autoShow` changed to `true` (not from true => false) -- is that the intent?

components/overflow-group/overflow-group-mixin.js
margaree: This is just the original logic from the overflow group component so I don't know offhand what the intention was, but my guess would be that we would want to run `_chomp` when either of them changed and so `.has` would make more sense.

components/overflow-group/overflow-group-mixin.js
margaree: Same situation as above where I don't know what the original intent was but for this case I think it is correct as is. The other case where `_autoDetectBoundaries` is run also is only run when `autoShow` is true

margaree: So did some investigation and `get` actually returns the old value not the new, so this code runs when `autoShow` goes from true to false, which I don't think was the intention. It actually throws in the case where it goes from true to false, so I'll do some fixing there.

components/overflow-group/overflow-group-mixin.js
dbatiste: Agreed... I think `.has` makes more sense. Note though that this would have been ignoring the case where the value changed from `undefined`/`null`/`0`, since `get` is returning the old value. Once these values are set, they probably never change in practice, and the initial chomping must have been invoked through another path - maybe that's why this hasn't been an issue.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,349 @@
...
+		// convert them to layout items (calculate widths)
+		this._itemLayouts = this._getItemLayouts(this._slotItems);
+		// convert to dropdown items (for overflow menu)
+		this.dropdownItems = this._slotItems.map((node) => this.convertToOverflowItem(node));
+	}
+
+	async _getSlotItems() {
+		const filteredNodes = await filterAsync(this._buttonSlot.assignedNodes({ flatten: true }), async(node) => {
+			if (node.nodeType !== Node.ELEMENT_NODE) return false;
+			await node.updateComplete;
```
dbatiste, line 304: This seems to assume that the node has an `updateComplete` property that returns a promise.  What happens if the node is a native element such as a `button` or a `template`?

components/overflow-group/overflow-group-mixin.js
dbatiste: I think we'll just need to be careful of the timing, since `this._container` until the first render has completed. It would be cleaner if we could wire up the observer to `this`... not sure if that's possible. That element might have been chosen specifically for the observer since it's the one that specifies some layout.

components/overflow-group/overflow-group-mixin.js
margaree: It does also set in `_chomp` for if the width becomes quite small. I tried removing it and it gets a bit overly complicated. I don't think we need for it to reflect, so state seems to work.

components/overflow-group/overflow-group-mixin.js
margaree: Ah yup `this._container` doesn't exist in `connectedCallback`, though if we add in a `requestAnimationFrame` to `connectedCallback` (we do this in `hierarchical-view-mixin`) it does exist so we could do that.

I would be a bit afraid of using `this` instead of `this._container` - I've run into a lot of troubles with similar functionality in `tag-list` figuring out what to put the `ResizeObserver` on and I'd worry that there'd be some weird behaviour that would pop up.

components/overflow-group/overflow-group-mixin.js
margaree: So at least with `button` it just seems to skip over and still return the element as expected. However if you console log `node.updateComplete` it is undefined so best to avoid that altogether. Is just adding an `if (node.updateComplete) await node.updateComplete` fine or is there a best practice for this?

components/overflow-group/test/overflow-group.visual-diff.html
```diff
@@ -74,18 +74,26 @@
 		<div id="auto-show-container" class="visual-diff" style="height: 200px; width: 500px;">
 			<d2l-overflow-group id="auto-show" auto-show>
 				<d2l-button>Button 1</d2l-button>
-				<d2l-button class="d2l-overflow-group-show">Button 2</d2l-button>
+				<d2l-button class="d2l-button-group-show">Button 2</d2l-button>
```
margaree, line 77: This class was incorrect so expecting some visual diff changes

components/overflow-group/test/overflow-group.visual-diff.js
```diff
@@ -49,22 +49,27 @@ describe('d2l-overflow-group', () => {
 	const autoShow = [
 		{
 			name: 'auto-show-small',
-			selector: '#auto-show-small',
```
margaree, line 52: These weren't used so removed them

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,355 @@
...
+			/**
+			 * @ignore
+			 */
+			mini: {
+				state: true
+			},
+			/**
+			 * @ignore
+			 */
+			openerType: {
```
margaree, line 71: This case is a bit strange. This is only used in `overflow-group` but there are plans to likely use it with `filter-overflow-group`. The logic is such that it's difficult to fully just move it into `overflow-group.js` so for documentation purposes I've left this here and documented it in `overflow-group.js`

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,355 @@
...
+		if (this._resizeObserver) this._resizeObserver.disconnect();
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+
+		requestAnimationFrame(async() => {
+			await this.updateComplete;
+
+			const container = this.shadowRoot.querySelector('.d2l-overflow-group-container');
```
margaree, line 125: By this point the container seems to consistently exist. Not sure if it's too hacky to have a `requestAnimationFrame` with a `await this.updateComplete` within it though.

components/overflow-group/overflow-group-mixin.js
dbatiste: I did a quick test running `await undefined` and the browser just seems to ignore it, so sounds like what you have is safe.

components/overflow-group/overflow-group-mixin.js
dlockhart: It's unfortunate that there isn't a more reliable easy way to do this eh... searching through core, it looks like we almost always do the `.observe` call in `firstUpdated`. `hierarchical-view-mixin` does it in `connectedCallback`, but also uses a `rAF` (but didn't need `await this.updateComplete`).

The only other way I can think of to avoid the `rAF`/`updateComplete` would be to:
- Create the `ResizeObserver` in the constructor (we could actually do that now)
- Track whether or not it's observing with a member variable (e.g. `_isObserving`)
- Inside `updated()`: `if (!this._isObserving) { this._isObserving = true; this._resizeObserver.observe(this.shadowRoot.querySelector(...); }`
- Inside `disconnectedCallback`, disconnect the observer and set `_isObserving` to `false`

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,355 @@
...
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		if (this._resizeObserver) this._resizeObserver.disconnect();
+	}
+
+	async firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this._buttonSlot = this.shadowRoot.querySelector('slot');
```
dlockhart, line 134: This doesn't seem very expensive, so I'd be tempted to just move this down to the one place it's used inside `_getSlotItems`, which would remove the need for `firstUpdated` and remove a member variable. *shrug*

components/overflow-group/overflow-group-mixin.js
```diff
@@ -0,0 +1,347 @@
...
+		return items.filter(({ isHidden }) => !isHidden);
+	}
+
+	async _getItems() {
+		// get the items from the button slot
+		this._slotItems = await this._getSlotItems();
+		// convert them to layout items (calculate widths)
+		this._itemLayouts = this._getItemLayouts(this._slotItems);
+		// convert to dropdown items (for overflow container)
+		this.dropdownItems = this._slotItems.map((node) => this.convertToOverflowItem(node));
```
dbatiste, line 296: Do you think we should rename this? Is there anything left in the mixin that assumes the subclass will be handling the overflow with a dropdown?  Maybe we want to flip the naming around so this is more generic, and the variable that's used in the subclass (`d2l-overflow-group`) is more type specific?  Thoughts?

dbatiste: I guess there are the class constants above too. 🤔

margaree: Yup I think renaming this is a good idea. I think in general it would be some sort of dropdown but perhaps the mixin shouldn't make that assumption.

dbatiste: Sounds good. If the mixin doesn't have any technical requirement for it to be a dropdown, then I think it makes sense. Maybe review for other dropdown references and use similar naming needed.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -51,26 +45,27 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
...
-				state: true
-			},
 			/**
 			 * @ignore
 			 */
 			openerType: {
 				type: String,
 				attribute: 'opener-type'
+			},
+			_chompIndex: {
```
margaree, line 63: I put these back to how they were as I realized we can just send more useful parameters to `getOverflowContainer`. Can also put these back to being states though.

components/overflow-group/overflow-group.js
```diff
@@ -157,28 +157,27 @@ class OverflowGroup extends OverflowGroupMixin(RtlMixin(LitElement)) {
...
 		const menu = html`<d2l-dropdown-menu>
 			<d2l-menu label="${moreActionsText}">
 				${overflowItems}
 			</d2l-menu>
 		</d2l-dropdown-menu>`;
 
-		if (this.mini) {
-			return html`<d2l-dropdown-more class="${OVERFLOW_MINI_DROPDOWN_CLASS}" text="${moreActionsText}">
+		if (mini) {
+			return html`<d2l-dropdown-more class="${OVERFLOW_MINI_CLASS} d2l-overflow-dropdown-mini" text="${moreActionsText}">
```
margaree, line 169: This class and `d2l-overflow-dropdown` are used in lp so we can't just remove them right now (but after this PR goes in we could switch them to the new OVERFLOW_CLASS and OVERFLOW_MINI_CLASS)

dbatiste (approved):
I think this is looking good.  I think there's just those couple private properties to change to use `state: true`.

PR #2712, opened 2022-08-02
fix: properly inherit font family

dcostiuc:
[Rally Story](https://rally1.rallydev.com/#/?detail=/defect/643700721293&fdp=true)

Fix related to [this](https://d2l.slack.com/archives/C0PHG3QB0/p1659455962940789) slack thread.
Another related slack [thread](https://d2l.slack.com/archives/C03H8KH2FKN/p1659459859371599).

[en]
Before
image
After
image

Also for other languages that seem to support other non-Lato fonts (based on what we have [here](https://github.com/BrightspaceUI/core/blob/main/components/typography/typography.js#L69-L84)):

[ar]
Before
image
After
image

[zh]
Before
image
After
image

[ko]
Before
image
After
image

[ja]
Before
image
After
image


dlockhart (approved):

dcostiuc:
I was thinking about potentially having a vdiff test or few in another language (to have cases for `ko` or `ja`, etc), 

but after talking with Dave B, it seems as though our covention is to not really have vdiff tests in other locales/languages, and it's interesting to see that we only currently have it in for the [input-date component](https://github.com/BrightspaceUI/core/blob/main/components/inputs/test/input-date.visual-diff.js#L102-L115) in core.

dbatiste:
Just to clarify, if we see there is higher chance for a regression when using other languages, then including targeted tests could be worthwhile.  For most things, we rely on the font being inherited from the body and it's not that often that we need to override the user-agent for it. Mostly we just don't want to repeat all the tests for all possible variables due to the time it would take.

dbatiste (approved):

svanherk:
Also, to add to the localization test discussion in case it makes your decision - we can't currently test in `ko`, `ja`, `zh`, or `zh-tw`.  We used to by installing the languages manually, but it was flaky and also added a couple minutes to the visual-diff runs, so we decided to [remove it](https://github.com/BrightspaceUI/core/pull/1762).

PR #2715, opened 2022-08-02
fix: List items not read in Tooltip contents for safari and voiceover

dcostiuc:
[Rally ticket](https://rally1.rallydev.com/#/?detail=/defect/643273080983&fdp=true)

This PR attempts to fix, or at least work around, the issue where ordered (`ol`)/unordered (`ul`)/definition(`dl`) list items are not read in `Safari + VoiceOver` within a `Tooltip`. 

Margaree also found out that it's not just the `li` elements within the list element that aren't being announced - it's also any other content, such as plain text before the `li`s.

After discussing with [Margaree and Dave B](https://d2l.slack.com/archives/C03FC0GDVB7/p1659465104621859?thread_ts=1658950757.652959&cid=C03FC0GDVB7), we decided to go with a work around where we can get the list content _announced_, but with the tradeoff that the list semantics (`"bullet/list item #"` etc) still aren't announced (only for VoiceOver + Safari).

Note that this change was also tested with:
* VoiceOver + Chrome, 
* NVDA + Chrome/Edge/Firefox, 
* and JAWS + Chrome/Edge/Firefox,   

to check that list contents are still properly announced there (with list semantics), along with plain text still being announced normally. 

It was interesting to note that out of the above combinations, only the following actually had list semantics as part of its announcements (i.e. even without this change, and regardless of verbosity setting):
* NVDA + Firefox
* JAWS + Chrome/Edge/Firefox


dcostiuc:
This is another PR to consider merging into `main` directly (rather than the `Visibility Switch changes` feature branch), since technically it doesn't depend on any of the new Visibility Switch changes, and theoretically it shouldn't make a difference to clients yet, unless there are already uses of `lists within the tooltip contents` released.

components/tooltip/tooltip.js
```diff
@@ -470,7 +470,10 @@ class Tooltip extends RtlMixin(LitElement) {
 			<div class="d2l-tooltip-container">
 				<div class="d2l-tooltip-position" style=${styleMap(tooltipPositionStyle)}>
 					<div class="d2l-body-small d2l-tooltip-content">
-						<slot></slot>
+						<div role="text">
+							<!-- Note: role="text" is a workaround for Safari. Otherwise, list-item content is not announced with VoiceOver" -->
```
dbatiste, line 474: I'd be inclined to move this comment into JavaScript above, since the HTML comment is so readily seen (by users) inspecting the DOM.  I like the comment, just rather it be more hidden from view.

dcostiuc:
Planning to merge it into the feature branch for Visibility Switch changes.

dcostiuc:
Yikes, looks like I'll need to make another PR/cherry-pick.

PR #2718, opened 2022-08-03
Fixed empty state docs

oelhanafey:
In this PR, I added `Internal event` to the `empty-state-action-button` component to prevent it from showing in the Daylight docs. 

I also changed the empty state action to use `inline` styles instead of `inline-block` because on the [Daylight docs](http://localhost:8000/components/empty-state/#d2l-empty-state-action-button), it is causing the `empty-state-action-button` to be mis-aligned. I was investigating why this is occurring but have been unable to reproduce it outside of the Daylight site. That being said, using `display: inline` aligns the button in Daylight and doesn't seem to change any of the styles on the visual diffs or anywhere outside of Daylight.


oelhanafey:
This PR reintroduces the English RTL issue discussed in this PR: #2655. Changing the action styling to `display: inline` has likely caused the component to now consider itself as a single block of text which is why the issue resurfaced. 

RTL for the action link has some weird behaviour when the text is english, likely due to some internal styling by browsers, but seems to work as expected in an actual RTL language like Arabic.

components/empty-state/empty-state-action-button.js
```diff
@@ -5,6 +5,7 @@ import { css, html, LitElement, nothing } from 'lit';
 /**
  * `d2l-empty-state-action-button` is an empty state action component that can be placed inside of the default slot of `empty-state-simple` or `empty-state-illustrated` to add a button action to the component.
  * @fires d2l-empty-state-action - Dispatched when the action button is clicked
+ * @fires d2l-empty-state-illustrated-check - Internal event
```
dbatiste, line 8: Do you want this event to show up in the docs?  If not, then it can be ignored by adding `/** @ignore */` before the dispatch statement. For [example](https://github.com/BrightspaceUI/core/blob/main/components/list/list.js#L251).

oelhanafey: This won't work. The `web-component-analyzer` will only parse `/** @ignore */` if you immediately dispatch the event. In my case I saved it as a variable  and the `/** @ignore */` won't work. That's why the `Internal event` case was hardcoded into Daylight in situations where the analyzer does not work as expected. You can see this used in selection-input [here](https://github.com/BrightspaceUI/core/blob/04668717fca0b18c789c8f46c67471b7f2c1a1b9/components/selection/selection-input.js#L16).

That's what prompted me to update the docs in [this PR](https://github.com/BrightspaceUI/documentation/pull/986).

dbatiste (approved):

PR #2722, opened 2022-08-03
feat: d2l-filter-overflow-group component

margaree:
Initial work to add new `d2l-filter-overflow-group` component. This branches off of the branch that adds `OverflowGroupMixin` but doesn't particularly utilize it at this point. That will be done in a follow up PR. This is just to get the basics in like initial component structure and tests.

~My plan was to merge this into main after the OverflowGroupMixin PR gets merged in~ (done), then hook up the overflow group functionality in a follow up PR.


margaree:
This has become a bit of a mess. Will re-open a fresh PR.

components/filter/demo/filter-overflow-group.html
```diff
@@ -0,0 +1,168 @@
...
+						</d2l-filter-dimension-set>
+					</d2l-filter>
+					<d2l-filter>
+						<d2l-filter-dimension-set key="format" text="Format">
+							<d2l-filter-dimension-set-value key="selfpaced" text="Self-Paced"></d2l-filter-dimension-set-value>
+							<d2l-filter-dimension-set-value key="instructor" text="Instructor Lead" selected></d2l-filter-dimension-set-value>
+						</d2l-filter-dimension-set>
+					</d2l-filter>
+					<d2l-filter>
+						<d2l-filter-dimension-set key="language" text="Language" select-single>
```
svanherk, line 62: Should this be `selection-single` or is this a new thing I missed in a previous review?

components/filter/filter-overflow-group.js
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * A component that can be used to display a group of filters that will be put into an overflow filter when they no longer fit on the first line of their container
+ * @slot - d2l-filters to be added to the container
+*/
+class FilterOverflowGroup extends OverflowGroupMixin(LitElement) {
+
+	static get styles() {
+		return [super.styles, css`
+			::slotted(d2l-filter) {
+				margin-right: 0.3rem;
```
svanherk, line 13: Looks like RTL styles are added in the next PR 👍

components/filter/filter-overflow-group.js
```diff
@@ -0,0 +1,26 @@
+import { css, LitElement } from 'lit';
+import { OverflowGroupMixin } from '../overflow-group/overflow-group-mixin.js';
+
+/**
+ * A component that can be used to display a group of filters that will be put into an overflow filter when they no longer fit on the first line of their container
+ * @slot - d2l-filters to be added to the container
```
svanherk, line 6: What happens if they try to put something other than `d2l-filter` in the `slot`?  From the next PR it looks like we won't add it to the overflow group, but it'll be rendered until it overflows and then disappear?  It's an invalid state so maybe this is "good enough" to tell the consumer they messed up, as long as no one uses it as feature

components/filter/filter-overflow-group.js
```diff
@@ -0,0 +1,26 @@
...
+				margin-right: 0.3rem;
+			}
+		`];
+	}
+
+	convertToOverflowItem() {
+	}
+
+	getOverflowContainer() {
+	}
```
svanherk, line 22: Are these required to be set for the overflow mixin to work?  Just wondering if there was any discussion in that PR about defining them in the mixin and throwing an error, like https://github.com/BrightspaceUI/core/blob/main/components/form/form-mixin.js#L61-L71.  Seems like sometimes we do that, sometimes we don't.

components/filter/filter-overflow-group.js
margaree: Yup that's correct as to how it currently behaves. We could also add a `console.warn` in `convertToOverflowItem` about how this item is invalid here and will just disappear once space is no longer available for it.

components/filter/filter-overflow-group.js
margaree: There wasn't any discussion about that but that's a good idea in order to provide more information.

svanherk (approved):
Looks good to me!

PR #2726, opened 2022-08-04
feat: add "d2l-dropdown-opener-click" event

dlockhart:
For many of the LMS dropdowns in the navbar, they actually use the `no-auto-open` property to prevent the default open/close behaviour when the opener is clicked. That's because they have some logic where on the first open they want to do a partial render before actually opening the dropdown.

Unfortunately, when I switched from using a `<d2l-dropdown>` with a custom opener to having `<d2l-navigation-dropdown-button-icon>` actually be the opener (which is better), there's now no way to distinguish between a click on the opener and a click inside the dropdown to manually open/close it.

So this PR introduces a `d2l-dropdown-opener-click` event that consumers who are using `no-auto-open` can listen for to know when the opener was clicked (or ENTER was pressed).

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -174,6 +174,13 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this.dropdownOpened = !this.dropdownOpened;
 	}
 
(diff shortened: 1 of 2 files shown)


dbatiste (approved):
Looks good!

PR #2727, opened 2022-08-04
fix: SubscriberControllers: fix error caused by space on end

margaree:
When ids ended with a space the error `SyntaxError: Failed to execute 'querySelector' on 'Document': '#' is not a valid selector.` was being thrown.


svanherk (approved):
Nice, thanks @margaree!

PR #2729, opened 2022-08-05
feat: Filter overflow group functionality

margaree:
Final PR for initial `d2l-filter-overflow-group` component. Adds documentation and actual overflow functionality.


components/filter/filter-overflow-group.js
```diff
@@ -1,24 +1,62 @@
...
+	getOverflowContainer(overflowItems) {
+		return html`
+			<d2l-filter class="${OVERFLOW_CLASS}" @d2l-filter-change="${this._handleFilterChange}">
+				${overflowItems}
+			</d2l-filter>
+		`;
 	}
 
-	getOverflowContainer() {
+	_handleFilterChange(e) {
```
margaree, line 49: This function is to keep the `selected` state in sync whether selected in the slotted `d2l-filter` or in the overflow `d2l-filter`

components/overflow-group/overflow-group-mixin.js
```diff
@@ -28,8 +29,7 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 	static get properties() {
 		return {
 			/**
-			 * Use predefined classes on slot elements to set min and max slotted items to show
-			 * @type {boolean}
+			 * @ignore
 			 */
 			autoShow: {
```
margaree, line 34: Not (at least initially) supporting autoShow in filter-overflow-group, so moving documentation to just be in `overflow-group`

components/overflow-group/overflow-group-mixin.js
```diff
@@ -130,8 +131,13 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 			}
 		});
 
+		const containerStyles = {
```
margaree, line 134: This sets the min and max height of the container to be the tallest slotted item as calculated in _getItemLayouts. I should also be checking overflowContainer height as it's possible that could be taller.

This is because the overflow groups always appear on one line and this prevents a momentary flash where the contents wrap to two line, which can be jarring.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -311,7 +320,9 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 
 		this._updateOverflowItemsRequested = true;
 		setTimeout(() => {
-			this._overflowItems = this._slotItems.map(node => this.convertToOverflowItem(node));
+			this._itemLayouts = this._getItemLayouts(this._slotItems);
+			this._overflowItems = this._slotItems.map((node) => this.convertToOverflowItem(node));
+			this._chomp();
```
margaree, line 325: When selection status changes, the width of an item can also change and that can affect chomping. I put this here in the mixin since in theory mutations to other components using the mixin could also cause layout changes.

margaree:
Opening a fresh PR for this

PR #2736, opened 2022-08-09
feat: announce conditions met on hidden label

dcostiuc:
[Rally Story](https://rally1.rallydev.com/#/?detail=/userstory/644021306285&fdp=true)

Accounting for an edge case as discussed in [this](https://d2l.slack.com/archives/C03FC0GDVB7/p1659034537026399) slack thread.


components/switch/switch-mixin.js
```diff
@@ -193,7 +196,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 			? this.textPosition : 'end');
 
 		return html`
-			${textPosition === 'start' ? switchLabel : ''}
+			${textPosition === 'start' || textPosition === 'hidden' ? switchLabel : ''}
```
margaree, line 199: Does the switchLabel need to be rendered both before and after the div or would it work to use just one? It would also likely be useful to add a quick comment on why we're rendering it even though there's an `aria-label` on the switch div.

dcostiuc: Feedback addressed in deaf5d005d77f6505875cbca382f934384fcd016 and 4b9b80f9f12f0527d3dd780250e36ba7db0cb3db

components/switch/switch-mixin.js
```diff
@@ -192,6 +195,7 @@ export const SwitchMixin = superclass => class extends FocusMixin(RtlMixin(Focus
 		const textPosition = (this.textPosition === 'start' || this.textPosition === 'hidden'
 			? this.textPosition : 'end');
 
+		// Note: we render the switchLabel in the case of textPosition === 'hidden' so that any slot handlers can pick up on content being passed in
```
dcostiuc, line 198: In this case, I'm specifically referring to the Visibility Switch's `_handleConditionsSlotChange`, which is necessary for the `_hasConditions` flag to be properly set.

margaree (approved):

PR #2740, opened 2022-08-09
fix: Added h-align to empty-state-action-button

oelhanafey:
Small fix to `empty-state-action-button`. Added `h-align: text` styling into the `empty-state-action-button` when rendered within `empty-state-simple`. This was missed when converting from the previous iteration of the empty state API.

Also changed the visual diffs slightly to add the case where the buttons and links are the first elements on a new line.


dlockhart (approved):
This and the visual-diff changes LGTM

PR #2742, opened 2022-08-10
feat: Filter overflow group functionality

margaree:
Final PR for initial d2l-filter-overflow-group component. Adds documentation and actual overflow functionality.


components/overflow-group/overflow-group-mixin.js
```diff
@@ -28,8 +29,7 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 	static get properties() {
 		return {
 			/**
-			 * Use predefined classes on slot elements to set min and max slotted items to show
-			 * @type {boolean}
+			 * @ignore
```
margaree, line 32: Not (at least initially) supporting autoShow in `filter-overflow-group`, so moving documentation to just be in `overflow-group`

components/overflow-group/overflow-group-mixin.js
```diff
@@ -129,8 +130,13 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 			}
 		});
 
+		const containerStyles = {
```
margaree, line 133: This sets the min and max height of the container to be the tallest slotted item as calculated in _getItemLayouts. I should also be checking overflowContainer height as it's possible that could be taller.

This is because the overflow groups always appear on one line and this prevents a momentary flash where the contents wrap to two line, which can be jarring.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -318,7 +327,9 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 
 		this._updateOverflowItemsRequested = true;
 		setTimeout(() => {
-			this._overflowItems = this._slotItems.map(node => this.convertToOverflowItem(node));
+			this._itemLayouts = this._getItemLayouts(this._slotItems);
+			this._overflowItems = this._slotItems.map((node) => this.convertToOverflowItem(node));
+			this._chomp();
```
margaree, line 332: When selection status changes, the width of an item can also change and that can affect chomping. I put this here in the mixin since in theory mutations to other components using the mixin could also cause layout changes.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -318,18 +329,27 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
...
+		if (!this._hasResized) {
+			this._hasResized = true;
+			await this._handleSlotChange();
+		} else {
+			this._chomp();
+		}
 	}
 
 	_handleSlotChange() {
+		if (!this._hasResized) return;
```
margaree, line 352: This is used similarly in tag-list; we don't want to get all the information about the slot content and run a chomp until a resize has happened (and the fonts have loaded)

components/overflow-group/overflow-group-mixin.js
```diff
@@ -318,18 +328,28 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
...
+			this._overflowItems = this._slotItems.map((node) => this.convertToOverflowItem(node));
+			this._chomp();
 			this._updateOverflowItemsRequested = false;
 			this.requestUpdate();
 		}, 0);
 	}
 
-	_handleResize(entries) {
+	async _handleResize(entries) {
+		await (document.fonts ? document.fonts.ready : Promise.resolve());
```
margaree, line 340: I found that if we didn't wait for this then the item widths calculated would usually not be correct. It wouldn't be crazy off, but for test cases that were exact it was noticeable and caused flake.

I put this here as I wanted to wait for it before the resize observer got used. It could also have a bool where we only await it if they haven't yet loaded.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -129,15 +131,20 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
...
-			<div class="d2l-overflow-group-container">
+			<div class="d2l-overflow-group-container" style="${styleMap(containerStyles)}">
 				<slot @slotchange="${this._handleSlotChange}"></slot>
 				${overflowContainer}
 			</div>
 		`;
 	}
 
-	update(changedProperties) {
+	async update(changedProperties) {
```
svanherk, line 147: Why was this change needed?

components/overflow-group/overflow-group-mixin.js
svanherk: Makes sense to me, but maybe a comment here would be helpful so we remember why we did this

components/overflow-group/overflow-group-mixin.js
svanherk: Right, like a change to `text` (which it looks like the observer also listens to) could potentially cause a width change as well.  Are noticing a perf hit with this at all, since it has to re-calculate for each item?

components/overflow-group/overflow-group-mixin.js
svanherk: If they _do_ set `auto-show` on the `d2l-filter-overflow-group` what happens?  Chaos?

margaree: haha no chaos, it does actually work as is. This made me realize I need to make a tweak to max height though since wrapping could technically happen with `auto-show` and a `d2l-button-group-show` on an end element.

components/overflow-group/overflow-group-mixin.js
margaree: Good point, I can see if I can change this to only do the whole workflow if selection changes

components/filter/filter-overflow-group.js
```diff
@@ -1,26 +1,63 @@
-import { css, LitElement } from 'lit';
-import { OverflowGroupMixin } from '../overflow-group/overflow-group-mixin.js';
+import './filter.js';
+import { css, html, LitElement } from 'lit';
+import { OVERFLOW_CLASS, OverflowGroupMixin } from '../overflow-group/overflow-group-mixin.js';
+import { RtlMixin } from '../../mixins/rtl-mixin.js';
+
+function createFilterItem(node) {
+	const dimensionSets = node.querySelectorAll('d2l-filter-dimension-set');
```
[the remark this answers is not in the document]
margaree: I believe so but I'll confirm

components/overflow-group/overflow-group-mixin.js
margaree: Ah forgot to remove this. I'd added it when trying to get something to work.

components/filter/filter-overflow-group.js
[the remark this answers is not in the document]
margaree: This is correct, the order is the same in the overflow filter

components/overflow-group/overflow-group-mixin.js
```diff
@@ -129,8 +131,13 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 			}
 		});
 
+		const containerStyles = {
+			minHeight: this.autoShow ? 'none' : `${this._itemHeight}px`,
+			maxHeight: this.autoShow ? 'none' : `${this._itemHeight}px`
```
svanherk, line 136: Since this is a `styleMap`, could you switch the `'none'` to `false` and it just won't add the `minHeight` and `maxHeight` styles at all?  Unless you do want them to be set to `none`

margaree: Just tried that and it looks like then it sets the values to `false` and still adds them to the div's styles, so I'm guessing false only really works for `classMap` (which I believe we wouldn't want to use in this case since the possible min/max height is a variable)

svanherk: Huh weird, yeah I figured they just worked the same because they're named similarly 🤷

svanherk (approved):

PR #2738, opened 2022-08-09
feat: Visibility Switch > indication of availability conditions and related changes

dcostiuc:
[Rally Feature](https://rally1.rallydev.com/#/?detail=/portfolioitem/feature/636986434285&fdp=true)

This PR merges in the `Visibility Switch - Indication of avaibility conditions` feature branch into `main`, where this feature branch contains several new additions/improvements/tweaks/fixes to the `Visibility Switch`, as well as the general `Switch`, `Tooltip`, and `Help Tooltip`, in some cases.

Specifically, it includes the following changes:
* #2665
* #2720 
* #2690 
* #2736 
* #2750 

Each of these individual PRs have already been reviewed.

#### Accessibility
* [Visibility Switch/Help Tooltip > tooltip contents being read on help tooltip close/tab back to switch](https://rally1.rallydev.com/#/?detail=/defect/643715796967&fdp=true)
* [Visibility Switch > switch state toggle does not announce labels (only on/off) for JAWS](https://rally1.rallydev.com/#/?detail=/defect/643727769011&fdp=true)
* [Help Tooltip > tooltip contents being repeated more than once for JAWS](https://rally1.rallydev.com/#/?detail=/defect/643729317373&fdp=true)
* [[mobile] Visibility Switch > switch state toggle does not announce labels (only on/off)](https://rally1.rallydev.com/#/?detail=/defect/644288372079&fdp=true)
* [[mobile] Help Tooltip/Visibility Switch > tooltip contents not announced on tooltip open](https://rally1.rallydev.com/#/?detail=/defect/644289999437&fdp=true)
* [Help Tooltip/Tooltip > tooltip conditions announced without list semantics on browser window tab back](https://rally1.rallydev.com/#/?detail=/defect/644298190871&fdp=true)
* [Help Tooltip > occasionally tooltip contents not announced when switching into window](https://rally1.rallydev.com/#/?detail=/defect/644299289151&fdp=true)

#### UI/UX
* [(Visibility) Switch > label occasionally highlighted on switch toggle](https://rally1.rallydev.com/#/?detail=/defect/644299602729&fdp=true)

PR #2745, opened 2022-08-10
fix: Tooltip > tweak styling for paragraphs

dcostiuc:
Note: planning to merge this into `main` as a separate release, to not block [the other PRs](https://github.com/BrightspaceUI/core/pull/2738) from getting merged in.

[Rally Story](https://rally1.rallydev.com/#/?detail=/userstory/644180344935&fdp=true)

Essentially just adding in a bit of styling to the tooltip so that we don't have extra margins on the top or bottom of the tooltip contents, although we can only do this for a certain case of the consumer passing in content into the slot*.

*Note that as soon as the consumer wraps their paragraph elements in a div, or pass in content structured in a different way, the CSS in this PR will not work. 
However, Jeff and Dave B still agreed that it wouldn't hurt to add this in.


components/switch/README.md
```diff
@@ -62,6 +62,7 @@ To make your usage of `d2l-switch` accessible, use the following property:
 ## Visibility Switch [d2l-switch-visibility]
 
 The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions ("Visible. Conditions must be met.").
```
eKoopmans, line 65: This looks like a duplicate of Line 66.

components/tooltip/tooltip.js
```diff
@@ -351,6 +351,14 @@ class Tooltip extends RtlMixin(LitElement) {
 				padding-right: 1rem;
 			}
 
+			::slotted(p:first-child) {
+				margin-top: 0 !important;
+			}
+
+			::slotted(p:last-child) {
+				margin-bottom: 0 !important;
```
eKoopmans, line 359: Looks like these changes are for all tooltips, not just the new `d2l-help-tooltip` - is there any risk of this negatively affecting existing users of the tooltip?

components/tooltip/demo/tooltip.html
```diff
@@ -163,7 +156,7 @@ <h2>Help Tooltip</h2>
 		<d2l-demo-snippet>
 			<template>
 				<d2l-tooltip-help text="Helpful label">Contents should elaborate on the label (be short and concise)</d2l-tooltip-help>
-				<p class="d2l-body-small">
+				<p class="d2l-body-small" style="padding-top: 20px;">
```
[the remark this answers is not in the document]
dcostiuc: Yep, this change is unrelated but I just tacked it onto this PR anyway while I was at it.

I added a new vdiff test, since then we could capture the "margins being removed correctly" in automation. Or are you suggesting to also add that to the demo page?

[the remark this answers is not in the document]
eKoopmans: Yeah I was saying demoing paragraphs nested in the tooltip, if that's a supported use case.

components/tooltip/tooltip.js
dcostiuc: Yup, my understanding was that this was something that we wanted to change for the Tooltip in general, as per [here](https://d2l.slack.com/archives/C03FC0GDVB7/p1658338548469239?thread_ts=1658257409.018289&cid=C03FC0GDVB7).

However, theoretically if other consumers/teams have vdiff tests of their own, specifically with paragraphs slotted in, then they may notice those tests failing. Otherwise there shouldn't be any other risk.

components/tooltip/demo/tooltip.html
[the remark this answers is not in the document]
dcostiuc: I think it's more of an edge case, to be honest.

I believe the biggest use case for a tooltip is just passing in plain text or so. 
However, it could be possible (since we use slotting) that a consumer passes in a paragraph, and this was an attempt to improve that a bit since we don't desire the margins that paragraphs come with by default.

To me, it feels a bit weird to create a demo for it in the demo page, since it's less of a distinct "thing" compared to what we showcase in the other demos, I suppose. 

If someone on our team wanted to double-check that, they should be able to just generate the vdiff golden. 
Though I suppose for ease of use a demo in that case would make sense.

components/switch/README.md
dcostiuc: Ah, yup, my bad.

dcostiuc:
Closing this PR.

The reasoning is discussed in a [Rally discussion comment](https://rally1.rallydev.com/#/15545167705d/dashboard?detail=%2Fuserstory%2F644180344935%2Fdiscussion).

I don't think it should be a big deal since that new css would only work in a certain situation anyway (not to mention, I checked, and we don't really seem to have `p` tags passed into any tooltips at the moment either).

Also, this was always considered an optional improvement.

PR #2747, opened 2022-08-10
Reintroduce support for templates within html-blocks

grant-cleary:
So... I think in hindsight there are a few unfortunate flaws that came up when we removed templates from the html-block API. Specifically, there are certain elements that will continue to load even if they're slotted in a `display: none` container. The two issues I'm currently aware of are as follows:
1. Scripts within iframes will still run. This case is only temporary because the iframe won't render after stamping has completed, but depending on the order things load in, if the iframe contents make web requests, those can still end up going out before the component renders. Those requests get canceled by the browser, but we still end up with tons of duplicate requests being made.
2. Audio elements with `autoplay` enabled will play regardless of being in a `display: none` container. This one is arguably worse because the user can't stop the audio without either disabling it in the browser tab or by using the dev console to delete the unstamped version.

So I figure... We may want to consider re-introducing support for templates. This PR makes a first go at that, with the paradigm of maintaining both methods simultaneously. As a reminder, we need to support the non-template case for components that need to render dynamic HTML, as Lit won't render dynamic HTML within `<template>`s. I figure we have two options here...
1. We support both, as in this PR. The norm would be to use `<template`s but components that absolutely need dynamic HTML can opt not to use them.
2. We accept that any component that can't use `<template>`s needs to set `no-deferred-rendering` and import the HTML block styles itself.

I'm open to ideas. I'm leaning a bit towards (2), personally, since it would go the furthest towards maintaining a consistent experience... The issue with (1) is that the two issues noted above would still potentially occur if the html-block is used in this way.

### DIFF
--- a/components/html-block/demo/html-block.html
+++ b/components/html-block/demo/html-block.html
@@ -402,6 +402,120 @@
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>HTML Block (math, template)</h2>
+
+			<d2l-demo-snippet>
+				<template>
+					<d2l-html-block>
+						<template>
+							<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
+								<mi>x</mi>
+								<mo>=</mo>
+								<mrow>
+									<mfrac>
+										<mrow>
+											<mo>&#x2212;</mo>
+											<mi>b</mi>
+											<mo>&#xB1;</mo>
+											<msqrt>
+												<msup>
+													<mi>b</mi>
+													<mn>2</mn>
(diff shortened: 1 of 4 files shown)


grant-cleary:
Note: This is currently based off of my existing refactoring branch (this PR: https://github.com/BrightspaceUI/core/pull/2699) because I'm not interested in doing this work twice for pre-refactor and post-refactor states.

grant-cleary:
Closing in favour of https://github.com/BrightspaceUI/core/pull/2767

PR #2748, opened 2022-08-10
feat: added querySelectorComposed to DOM helpers

dlockhart:
For the "skip nav" link accessibility fixes, we'd like to change the behavior such that it searches for a `<main>` or `[role="main"]` and then falls back to `<h1>` when moving focus. We'd also like for that search to include those elements inside web component shadow roots, since page or app templates [like our primary/secondary template](https://github.com/BrightspaceUI/core/blob/main/templates/primary-secondary/primary-secondary.js#L939) sometimes render those elements.

This adds a new DOM helper that works just like `document.querySelector` except that it also reaches into custom element shadow roots and queries them as well. I considered adding a `querySelectorComposedAll`, but decided to leave that for when/if the use case comes up.

### DIFF
--- a/helpers/dom.js
+++ b/helpers/dom.js
@@ -234,9 +234,9 @@ export function querySelectorComposed(node, selector) {
 	if (elem) return elem;
 
 	const allElems = node.querySelectorAll('*');
-	for (let i = 0; i < allElems.length; i++) {
-		if (allElems[i].shadowRoot) {
-			const nestedElem = querySelectorComposed(allElems[i].shadowRoot, selector);
+	for (const elem of allElems) {
+		if (elem.shadowRoot) {
+			const nestedElem = querySelectorComposed(elem.shadowRoot, selector);
 			if (nestedElem) return nestedElem;
 		}
 	}
--- a/helpers/dom.js
+++ b/helpers/dom.js
@@ -230,6 +230,13 @@ export function isVisible(node) {
 
 export function querySelectorComposed(node, selector) {
 
+	if (!node || (node.nodeType !== 1 && node.nodeType !== 9 && node.nodeType !== 11)) {
+		throw new TypeError('Invalid node. Must be nodeType document, element or document fragment');
+	}
+	if (typeof selector !== 'string') {
+		throw new TypeError('Invalid selector');
+	}
+
 	const elem = node.querySelector(selector);
 	if (elem) return elem;
 
--- a/helpers/dom.js
+++ b/helpers/dom.js
@@ -227,3 +227,19 @@ export function isVisible(node) {
 	return true;
 
 }
+
+export function querySelectorComposed(node, selector) {
+
+	const elem = node.querySelector(selector);
+	if (elem) return elem;
+
+	const allElems = node.querySelectorAll('*');
+	for (let i = 0; i < allElems.length; i++) {
+		if (allElems[i].shadowRoot) {
+			const nestedElem = querySelectorComposed(allElems[i].shadowRoot, selector);
+			if (nestedElem) return nestedElem;
+		}
+	}
+
+	return null;
+}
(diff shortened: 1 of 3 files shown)


helpers/test/dom.test.js
```diff
@@ -757,4 +759,54 @@ describe('dom', () => {
...
+				render() {
+					return html`<div class="shadowDiv"><slot></slot></div>`;
+				}
+			},
+		);
+
+		const nestedElement = defineCE(
+			class extends LitElement {
+				render() {
+					return html`<d2l-button-icon icon="tier1:delete" text="delete"></d2l-button-icon>`;
```
dlockhart, line 775: I didn't want to use button-icon here as there's a bit of leakage, but unfortunately you can't combine two `defineCE` components -- like one can't reference another.

helpers/dom.js
```diff
@@ -227,3 +227,19 @@ export function isVisible(node) {
...
 
 }
+
+export function querySelectorComposed(node, selector) {
+
+	const elem = node.querySelector(selector);
+	if (elem) return elem;
+
+	const allElems = node.querySelectorAll('*');
+	for (let i = 0; i < allElems.length; i++) {
```
eKoopmans, line 237: Nit: `i` isn't used here outside of indexing `allElems`, so you could instead use `for (const elem of allElems)`.

helpers/test/dom.test.js
eKoopmans: Would this not work:

- `nestedElement` rendered just a div with a class like `.nested`
- in the test, you did ``await fixture(`<${slottedElement}><${nestedElement}></${nestedElement}></${slottedElement}>`)``
- then querying for `.nested`?

helpers/dom.js
```diff
@@ -227,3 +227,19 @@ export function isVisible(node) {
 	return true;
 
 }
+
+export function querySelectorComposed(node, selector) {
+
+	const elem = node.querySelector(selector);
+	if (elem) return elem;
+
+	const allElems = node.querySelectorAll('*');
```
eKoopmans, line 236: It's a shame there's no query selector for just custom elements, but I didn't find anything.

eKoopmans (approved):
Looks great! Probably not super performant, but works in a pinch.

helpers/test/dom.test.js
dlockhart: In that case, `nestedElement` would be inside `slottedElement`'s light DOM, so would be found as part of the first `node.querySelector` call. For this test, I want it to query a custom element's shadow root, not find it but find another custom element and then recurse and find it as part of that nested element's shadow root.

helpers/dom.js
dlockhart: Agreed -- fixed!

dlockhart:
> Probably not super performant, but works in a pinch.

Yeah I agree. In the place this will be used (skip-nav link), it's not executed until the user actually clicks the link, so we'll be deferring it and it won't run on page load.

helpers/test/dom.test.js
eKoopmans: Right, gotcha. That's fine then, though would it be an option to actually use `customElements.define()` to create an element that can be nested, and including it as a sibling to the slot in `slottedElement`? Not sure the specific benefits/requirements to using `defineCE`, apart from it seeming to create a nice impermanent element.

dlockhart: Good call! Yep, that worked. Thanks for the suggestion. 🎉

eKoopmans (approved):

helpers/dom.js
```diff
@@ -227,3 +227,19 @@ export function isVisible(node) {
 	return true;
 
 }
+
+export function querySelectorComposed(node, selector) {
+
+	const elem = node.querySelector(selector);
```
eKoopmans, line 233: One note: this will fail if `node` is undefined, null, or otherwise not a node type. Just wondering if you considered doing a check and throwing a more friendly error - but I think it's fine as-is.

dlockhart: Good call -- added type checking for both parameters.

eKoopmans (approved):

PR #2750, opened 2022-08-11
fix: Visibility Switch > Documentation Tweaks

dcostiuc:
Related to this [Rally Story](https://rally1.rallydev.com/#/?detail=/userstory/637810377847&fdp=true)

Added in a separate demo showing the consumer how they can work with the new "conditions" state of the Visibility Switch.

### DIFF
--- a/components/switch/README.md
+++ b/components/switch/README.md
@@ -61,7 +61,8 @@ To make your usage of `d2l-switch` accessible, use the following property:
 
 ## Visibility Switch [d2l-switch-visibility]
 
-The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".  
+The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+
 It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions ("Visible. Conditions must be met.").
 
 <!-- docs: demo live name:d2l-switch-visibility autoSize:false size:small -->
@@ -92,7 +93,8 @@ It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions 
 <!-- docs: end hidden content -->
 
 ### Visibility Switch with Conditions
-It may be helpful to have visibility be dependent on additional conditions being met outside of the Visibility Switch.  
+It may be helpful to have visibility be dependent on additional conditions being met outside of the Visibility Switch.
+
 Additional information can be added to the label by passing it into the `d2l-switch-visibility` element.
 
 <!-- docs: demo code autoSize:false size:large -->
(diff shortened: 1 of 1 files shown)


components/switch/README.md
```diff
@@ -90,3 +90,21 @@ It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions
 - Optional default slot content - Content that will be displayed within the "conditions" opener tooltip when the switch is on.
   
 <!-- docs: end hidden content -->
+
+### Visibility Switch with Conditions
+It may be helpful, depending on the use case, to include additional information in the label, relating to other factors that may affect the availability that the Visibility Switch represents.
+
+<!-- docs: demo code autoSize:false size:large -->
```
dcostiuc, line 97: Manually set the demo container `size` to `large` as that was the smallest container size where there would be enough room for the `conditions met` tooltip to show up on the bottom of the component and not to the side.

components/switch/README.md
```diff
@@ -90,3 +90,21 @@ It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions
 - Optional default slot content - Content that will be displayed within the "conditions" opener tooltip when the switch is on.
   
 <!-- docs: end hidden content -->
+
+### Visibility Switch with Conditions
+It may be helpful, depending on the use case, to include additional information in the label, relating to other factors that may affect the availability that the Visibility Switch represents.
```
dcostiuc, line 95: Let me know if this makes sense or not.

components/switch/README.md
```diff
@@ -61,7 +61,7 @@ To make your usage of `d2l-switch` accessible, use the following property:
 
 ## Visibility Switch [d2l-switch-visibility]
 
-The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
```
oelhanafey, line 64: Seems like you missed some extra whitespace here

components/switch/README.md
oelhanafey: Maybe splitting this into two separate sentences would make it flow a bit smoother?
e.g. "You may need to have visibility be dependent on additional conditions being met outside of the Visibility Switch. You can include additional information ..."

components/switch/README.md
dcostiuc: Oh, yeah, I added that in there so that the sentence after it would properly render on a new line on the page 
(I think it's a markdown quirk where you need at least two spaces to have the next sentence show up on a new line)

eKoopmans: @dcostiuc you should use a blank newline between lines of text if you want them to appear on separate lines.

eKoopmans (approved):

oelhanafey (approved):

PR #2751, opened 2022-08-11
fix: Change filter tags 'no filters' colour for accessibility

oelhanafey:
The text in filter tags components when no filters were set did not have enough contrast to pass aXe tests. Changed the colour to `galena` which is a slightly darker grey (first picture) and added an aXe test for the "no filters" case to catch issues like this in the future.

I opted to keep the colour as close as possible to the original while still passing aXe tests but I put some other options to have a look at if someone thinks a darker colour may be more suitable.

Screen Shot 2022-08-11 at 1 59 08 PM
Screen Shot 2022-08-11 at 1 57 56 PM
Screen Shot 2022-08-11 at 1 47 41 PM


svanherk (approved):
We can check with @geurts once he's back but galena looks good to me! Thanks Oscar!

PR #2754, opened 2022-08-11
feat: filter-overflow-group: tags

margaree:
This change allows a consumer to add the `tags` property to `d2l-filter-overflow-group` and tags for all filters in the group will appear below the filter group. It adds a bit of convenience so the consumer doesn't have to be concerned with id management.

Something to note is that because of the fact that the filters aren't actually rendered within the component but are slotted elements, in order for `d2l-filter-tags` to be able to access them it also has to be outside of the shadow dom of the component. This isn't totally ideal and I'm open to alternative ideas here.


components/filter/filter-overflow-group.js
```diff
@@ -27,16 +40,46 @@ class FilterOverflowGroup extends OverflowGroupMixin(RtlMixin(LitElement)) {
...
+
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
 		this.addEventListener('d2l-filter-change', this._handleFilterChange);
+
+		if (!this.tags) return;
+
+		this._filterTags = document.createElement('d2l-filter-tags');
+		this.insertAdjacentElement('afterend', this._filterTags);
```
margaree, line 57: This adds a new `d2l-filter-tags` element as a sibling of the overflow group so that it can access/subscribe to the slotted filters.

If styles need to be changed, it will need to be done here rather than in `styles` like usual

components/filter/filter-overflow-group.js
```diff
@@ -27,16 +40,46 @@ class FilterOverflowGroup extends OverflowGroupMixin(RtlMixin(LitElement)) {
...
+		if (tagName !== 'd2l-filter') {
+			console.warn(`d2l-filter-overflow-group: ${tagName} is invalid in this group. This group should only contain d2l-filter direct child elements.`);
+			return;
+		}
+
+		if (this.tags && !node.hasAttribute('data-filter-tags-subscribed')) {
+			const filterId = node.id || getUniqueId();
+			node.id = filterId;
+			this._filterIds += ` ${filterId}`;
+			node.setAttribute('data-filter-tags-subscribed', 'data-filter-tags-subscribed');
```
margaree, line 79: I initially had the id addition happening in `firstUpdated`, but it's possible that new filters could be added, so I moved that here and added this `data-filter-tags-subscribed` so we aren't re-adding an id or scanning through `_filterIds` for the id each time this function is run.

If it's thought to be unlikely filters will be added after initial render then this can just be moved to firstUpdated.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -382,6 +382,7 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 			}
 
 			this._chomp();
+			this.requestUpdate();
```
margaree, line 385: It's possible to have situations where the `_chompIndex` doesn't change but we need to re-render (e.g., new filter is added but chomp index is calculated to stay the same we still need to re-render in order to add `data-is-chomped` to applicable elements)

components/filter/filter-overflow-group.js
svanherk: I think I'm missing something - how do filters that aren't an overflow item get their ids added to `this._filterIds` and hooked up to tags?

margaree: So everything that's in the `OverflowGroupMixin` gets converted to an overflow item, and then when they're needed because of a resize they get used in the overflow filter, so this function gets run on all the filters.

svanherk: Ah ok.  So we have a copy of every filter, and either the original or the overflow version will be visible with the other having `display: none;`.  And the tags get hooked up to only the original filters (where an id is added if needed).  If there was already an `id` it doesn't add a new one, and it doesn't matter that id wasn't copied to the overflow filters because its just copying the dimensions, not the outside filter.

I said the above like a statement but it's really a question lol

components/overflow-group/overflow-group-mixin.js
svanherk: Minor: Are there still situations where `this._chomp` doesn't need an update?  Seems like every `_chomp` now requests an update or happens in `update`.  Just wondering if this should go in `_chomp`, but maybe not worth possible issues or regression with that

components/filter/filter-overflow-group.js
```diff
@@ -27,16 +40,46 @@ class FilterOverflowGroup extends OverflowGroupMixin(RtlMixin(LitElement)) {
...
 		const tagName = node.tagName.toLowerCase();
-		if (tagName !== 'd2l-filter') console.warn(`d2l-filter-overflow-group: ${tagName} is invalid in this group. This group should only contain d2l-filter direct child elements.`);
-		else return createFilterItem(node);
+		if (tagName !== 'd2l-filter') {
+			console.warn(`d2l-filter-overflow-group: ${tagName} is invalid in this group. This group should only contain d2l-filter direct child elements.`);
+			return;
+		}
+
+		if (this.tags && !node.hasAttribute('data-filter-tags-subscribed')) {
+			const filterId = node.id || getUniqueId();
```
svanherk, line 76: Maybe we should add an `id` to one of the `d2l-filter`s in the visual-diff tests and demo, just to help catch any possible regressions around that

components/filter/filter-overflow-group.js
margaree: Yup that's mostly correct. We don't have a copy of every filter, just the dimensions, but yup when they're chomped they get a `display: none`. The remainder is all correct.

components/overflow-group/overflow-group-mixin.js
margaree: That's a good point. I think actually with the `_handleResize` case we could get rid of it and that would probably be a good idea for performance. The slot change and mutation cases would be the ones where it would be more likely to be needed.

components/filter/filter-overflow-group.js
svanherk: Would it work if we made this a child instead, so its still in the light dom but gets cleaned up if the filter overflow group is `diconnected`?  I guess that'd probably mess with all the chomping 🤔

margaree: I think I had tried that and it gets pulled into the overflow slot. Ah yeah should this be added/deleted in the connected/disconnected callback?

svanherk: Yeah I think `connected`/`disconnected` should work.  Was trying to think if we could do anything with a new non-default `slot`, but even so that might be weird since it'd be available to consumers but we wouldn't want them to use it

svanherk: @dlockhart @dbatiste Wondering if you have any major concerns here I've missed... since this mainly a convenience feature, I just want to make sure we're not setting ourselves up for future headaches with supporting this.
This allows the consumer to just add the `tags` attribute to this component and it will add the `d2l-filter-tags` after itself and hookup the ids for them, rather than having to do it like the "Filter Overflow Group with Tags (external to d2l-filter-overflow-group)" demo above.

dlockhart: Hmm yeah I don't love this for a couple reasons -- we're modifying the DOM outside of `render()` and the component is essentially "leaking" a sibling, which is pretty unexpected. That shows up in places like the unit and visual diff tests where we had to wrap it in a seemingly unnecessary `<div>` in order to properly capture both.

Could the overflow-group mixin expose another optional method like `getOverflowAdjacentElement()` (or a better name) that it would render after its `<div class="d2l-overflow-group-container">`?

svanherk: Yeah, I can play around with that. I like that better than trying to do something similar with a new `slot`.

svanherk: > Could the overflow-group mixin expose another optional method like getOverflowAdjacentElement() (or a better name) that it would render after its <div class="d2l-overflow-group-container">?

So turns out if we do this, then the `d2l-filter-tags` will be in the `d2l-filter-overflow-group`'s shadow DOM, but the `d2l-filter`s are all in the light DOM.  So they can't see each other.  I'm not seeing a way to get the tags in the light DOM other than what we do here (attaching them after `this`) or having the consumer place them in a `slot` (which is basically the same as having no `slot` at all and the consumers handling it).  If we add them as children, they get handled with all the other chomping calculations.

eKoopmans: > or having the consumer place them in a slot (which is basically the same as having no slot at all and the consumers handling it)

What about having a "silent" slot - it's there, the consumers aren't expected to use it, and the component injects the `d2l-filter-tags` into the slot, instead of as a child or sibling?

svanherk: Originally I was trying to avoid the slot but it's actually a lot cleaner than I expected! I'll put it up as a branched PR to make it easier to see that change, thanks!

eKoopmans (approved):

PR #2767, opened 2022-08-16
Add html property for deferred rendering html-block

grant-cleary:
This is another option for potentially solving the issues currently cropping up with the html-block's stamping process (see https://github.com/BrightspaceUI/core/pull/2747 for background)... Namely, don't stamp at all! This PR introduces an `html` property that the html-block can render directly into its render container, instead of stamping existing DOM elements in. Right now, it still also keeps most of the old way working and continues to respect the slotted content if there is any (for backwards compatibility), but in the future the idea would be to only use the slot for non-deferred rendering.


components/html-block/test/html-block.visual-diff.js
```diff
@@ -34,7 +34,9 @@ describe('d2l-html-block', () => {
 		{ name: 'math (inline)', selector: '#math-inline' },
 		{ name: 'code (block)', selector: '#code-block' },
 		{ name: 'code (inline)', selector: '#code-inline' },
-		{ name: 'math (block) and code (block)', selector: '#math-block-and-code-block' }
+		{ name: 'math (block) and code (block)', selector: '#math-block-and-code-block' },
+		{ name: 'using html property', selector: '#html-property' },
+		{ name: 'using both html property and slot', selector: '#html-property-and-slot' }
```
grant-cleary, line 39: I didn't see a point adding specific unit tests for these cases for now, since eventually the idea is that we'll only have two cases: property (deferred) and slotted (non-deferred), so those tests are going to need a rewrite anyways. In the meantime, I think these visual diff tests will do.

components/html-block/html-block.js
```diff
@@ -230,10 +240,21 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+			renderContainer.innerHTML = this.html;
+			await this._processRenderers(renderContainer);
+		}
 		if (this._contextChanged()) {
-			this._render();
+			if (this._hasSlottedContent) this._render();
+			else if (this.html !== undefined && this.html !== null) {
+				const renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
+				renderContainer.innerHTML = this.html;
+				await this._processRenderers(renderContainer);
```
grant-cleary, line 255: The idea here is to only render the actual results of the `html` property in update. If we try to do it in `render`, MathJax modifies the DOM and Lit can't track whether `html` has actually changed. Doing it this way means we keep Lit's change detection while also preventing the world from exploding post-render.

grant-cleary:
Hmm, need to check on those visual diff tests. Seems like in some cases the property is taking precedence over the slot, which isn't ideal.

grant-cleary:
Ooh, I know why that's happening! @dbatiste this came up the other day and I was kind of curious what your thoughts were... Basically, all of the logic we dumped in here to make sure we're only checking valid elements also means if the html-block only contains non-element nodes, the slotchange event doesn't actually do anything.

I feel like that's wrong and passing pure text nodes into the html-block should be valid, but I also feel like we ran into some kind of issue with that. Do you happen to remember the details?

components/html-block/html-block.js
```diff
@@ -250,8 +271,22 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+		if (this._hasSlottedElements(slottedNodes)) {
+			this._hasSlottedContent = true;
+			await this.updateComplete;
+			await this._render(e.target);
+		} else {
+			this._hasSlottedContent = false;
+		}
+	}
+
+	_hasSlottedElements(slottedNodes) {
```
dbatiste, line 287: Thinking about the element requirement here: the whole purpose of them providing slotted content is to avoid deferred rendering, in which case they _would_ always be providing an element. i.e. if they are just providing text, there's no reason for them not to use the `html` property.

grant-cleary: Right, this check exists because Lit seems to jam an empty text node into the slot. I figured it would be safe enough for the reasons you mentioned... If there's no actual HTML in here, I don't think we care about it because it's not really a valid use case for the slot (at least moving forward).

components/html-block/html-block.js
```diff
@@ -230,10 +240,21 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		`;
 	}
 
-	updated(changedProperties) {
+	async updated(changedProperties) {
 		super.updated(changedProperties);
+		if (changedProperties.has('html') && this.html !== undefined && this.html !== null && !this._hasSlottedContent) {
+			const renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
```
dbatiste, line 246: Looks like these 3 lines could be factored out (since they are duplicated) into an appropriately named method.

components/html-block/html-block.js
```diff
@@ -250,8 +271,22 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
 	async _handleSlotChange(e) {
-		if (!e.target) return;
-		await this._render(e.target);
+		if (!e.target || !this.shadowRoot) return;
+		const slot = this.shadowRoot.querySelector('slot');
+		const slottedNodes = slot.assignedNodes({ flatten: true });
+
+		if (this._hasSlottedElements(slottedNodes)) {
+			this._hasSlottedContent = true;
+			await this.updateComplete;
```
dbatiste, line 280: Under what circumstance do we need to wait for `updateComplete`? Looks like this is a new addition... did you run into an issue without it?

components/html-block/html-block.js
grant-cleary: Oops, you're right.

components/html-block/html-block.js
grant-cleary: Nope! This was a holdover from when I was trying to render the `html` in the `render` method, at which point we did need to wait for that to happen before trying to run the renderers against it. I can remove this.

dbatiste:
So how does this approach play out for the various scenarios we encounter?

* no deferred rendering, containing content that depends on listening to page load events
* avoid running user-authored scripts more than once
* content containing audio (or video?) with auto-play
* dynamic content (avoiding `template` due to Lit limitation), and html-editor inline mode

Have I missed any atypical scenario?

grant-cleary:
> * no deferred rendering, containing content that depends on listening to page load events

So this plays out identically to how it worked before, since we're still using the exact same slotted inline rendering in that case. The way I have things working right now is that if the slot has elements in it, it always takes priority over the `html` property (basically we'll always do non-deferred if consumers dump something in the slot... Right now we're still using the attribute to signal that, but that will change once existing consumers are ported over to the new property).

> * avoid running user-authored scripts more than once
> * content containing audio (or video?) with auto-play

This becomes a non-issue. If it's non-deferred, the same case as the above plays out (we were never double rendering). If it's deferred, we should now only render it once when the component updates and we should avoid the double script issue. Same with the autoplay shenanigans.

> * dynamic content (avoiding `template` due to Lit limitation), and html-editor inline mode

This I haven't yet tested, but in practice this just needs a bit of massaging so that the editor can pass in a string instead of a Lit template.

> Have I missed any atypical scenario?

I think this about covers it!

dbatiste:
>Basically, all of the logic we dumped in here to make sure we're only checking valid elements also means if the html-block only contains non-element nodes, the slotchange event doesn't actually do anything.

>I feel like that's wrong and passing pure text nodes into the html-block should be valid, but I also feel like we ran into some kind of issue with that. Do you happen to remember the details?

Yeah, it seems valid. No I don't recall the details - maybe is has something to do with handing off the nodes for processing by the renderers, since they probably require an element.  

We _want_ people to provide the HTML via the property because that will effectively defer it, which I think we prefer for the original performance reasons that were encountered back when all the rendering was done in the MVC control. With these changes, I think `no-deferred-rendering` no longer has a purpose, correct? Somehow we still want people to default to deferred rendering.

grant-cleary:
> Yeah, it seems valid. No I don't recall the details - maybe is has something to do with handing off the nodes for processing by the renderers, since they probably require an element.

That sounds familiar. Probably not as relevant with the refactored changes because we're always passing off the render container to the other renderers now.

> We _want_ people to provide the HTML via the property because that will effectively defer it, which I think we prefer for the original performance reasons that were encountered back when all the rendering was done in the MVC control. With these changes, I think `no-deferred-rendering` no longer has a purpose, correct? Somehow we still want people to default to deferred rendering.

That's right. The attribute needs to exist until existing consumers are all ported over, but it's pointless now. Using the property effectively does defer rendering by virtue of our components rendering async anyways, and if the slot takes precedence, we have a clear different API that can be used for cases where we still need the non-deferred case. Given this, I'm also not sure I care about processing a slot with only non-element nodes in it. Consumers can use the property if they want that.

dbatiste:
>Given this, I'm also not sure I care about processing a slot with only non-element nodes in it. Consumers can use the property if they want that.

Yeah I am more concerned about them just always dumping their HTML into the slot, which would not be deferred. At least before they had to consciously add `no-deferred-rendering`.

grant-cleary:
> > Given this, I'm also not sure I care about processing a slot with only non-element nodes in it. Consumers can use the property if they want that.
> 
> Yeah I am more concerned about them just always dumping their HTML into the slot, which would not be deferred. At least before they had to consciously add `no-deferred-rendering`.

Hmm, true. We could still require the attribute to even listen to what's in the slot. We can't make that change now (because it will break existing implementations), but it's something we could do once we're ready to switch over.

dbatiste:
>We could still require the attribute to even listen to what's in the slot.

I think I am in favour of that. It means they have to be very intentional about opting out of deferred rendering. As discussed in Slack, if they opt out, they also have to add the html-block styles in the consuming component, and if they just use the slot blindly without knowing what's required, there will be problems. For now, the **only** thing that would be using the slot would be the MVC control in specific cases.

Aside, I'm not convinced that user-authored content containing scripts that depend on page load events will ever work when rendered in the context of other web components that are rendered async. So, the MVC control rendering like this may be the only case of opting out, ever. Ideally, we are able to eventually move completely to sand-boxed widgets.

grant-cleary:
> I think I am in favour of that. It means they have to be very intentional about opting out of deferred rendering. As discussed in Slack, if they opt out, they also have to add the html-block styles in the consuming component, and if they just use the slot blindly without knowing what's required, there will be problems. For now, the **only** thing that would be using the slot would be the MVC control in specific cases.
> 
> Aside, I'm not convinced that user-authored content containing scripts that depend on page load events will ever work when rendered in the context of other web components that are rendered async. So, the MVC control rendering like this may be the only case of opting out, ever. Ideally, we are able to eventually move completely to sand-boxed widgets.

Yep, I think these are some great points. The moment we're rendering anything in another web component, we've added an async layer on top of the initial page render, and at that point it's already a crapshoot whether any scripts will run successfully.

I think subsequent changes do need to wait until consumers are swapped over to use the property, as we can't really change how the slot behaves until then. However, I also don't think that presents any kind of problem.

dbatiste:
>Yep, I think these are some great points. The moment we're rendering anything in another web component, we've added an async layer on top of the initial page render, and at that point it's already a crapshoot whether any scripts will run successfully.

Another problem they would very likely run into is using API's such as `document.querySelector(...)` etc.. If the elements they are interested in are buried in some shadowDOM, rendered by us or the consuming app, those API call are gonna break. Of course, we see that all the time with 3rd-party libraries that aren't developed with shadowDOM in mind. All the more reason to not worry about that use case.

grant-cleary:
> Another problem they would very likely run into is using API's such as `document.querySelector(...)` etc.. If the elements they are interested in are buried in some shadowDOM, rendered by us or the consuming app, those API call are gonna break. Of course, we see that all the time with 3rd-party libraries that aren't developed with shadowDOM in mind. All the more reason to not worry about that use case in `d2l-html-block`.

Agreed. "Here be dragons" if we're actually intending for client scripts to run within arbitrary web components, really.

grant-cleary:
A couple of the visual diff failures seem to be legit. I think the "elements only" check might be too strict right at this moment for figuring out whether to fall back to the slot or not. On the one hand, we already know plain text nodes in the slot don't currently work properly, but on the other hand, I think that was only on initial render, so I don't think it's safe to be _this_ strict quite yet.

components/html-block/html-block.js
```diff
@@ -250,8 +267,21 @@ class HtmlBlock extends RtlMixin(LitElement) {
 	}
 
 	async _handleSlotChange(e) {
-		if (!e.target) return;
-		await this._render(e.target);
+		if (!e.target || !this.shadowRoot) return;
+		const slot = this.shadowRoot.querySelector('slot');
+		const slottedNodes = slot.assignedNodes({ flatten: true });
+
+		if (!this.html && this._hasSlottedElements(slottedNodes)) {
```
grant-cleary, line 274: The addition of the `!this.html` check here will need some rethinking when consumers have implemented the new API. For now, however, it's required to maintain consistent behaviour. While the html-block doesn't currently _fully_ work properly if it contains only non-element nodes, the initial render does still work. Without this additional check, it stops working, and that's not good. While "partially broken" still isn't ideal, "completely broken" is worse.

grant-cleary:
Ok. I'm annoyed now because when I check the visual diffs locally they render as I would expect.

components/html-block/html-block.js
grant-cleary: Scratch this. It wasn't sufficient and I found a better way to handle this.

components/html-block/html-block.js
```diff
@@ -250,8 +267,22 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+			this._hasSlottedContent = true;
+			await this._render(e.target);
+		} else {
+			this._hasSlottedContent = false;
+		}
+	}
+
+	_hasSlottedElements(slottedNodes) {
+		if (!slottedNodes || slottedNodes.length === 0) return false;
+		if (slottedNodes.filter(node => node.nodeType === Node.ELEMENT_NODE || node.textContent).length === 0) return false;
```
grant-cleary, line 284: This second check should catch cases where the node has non-empty text content (i.e. contains only text nodes), but no elements. This is a case we won't want to continue to support once we get rid of the backwards compatibility.

svanherk:
> Ok. I'm annoyed now because when I check the visual diffs locally they render as I would expect.

Is it reproducible locally if you add retries (`--retries=2` )?

grant-cleary:
> Is it reproducible locally if you add retries (`--retries=2` )?

Oh, yeah, it did end up being reproducible and I (hopefully) addressed it in my last commit. I was just looking at two nearly (but not quite) identical html-blocks and misindentifying which was which. 😅

components/html-block/html-block.js
grant-cleary: I'm dumb. It wasn't sufficient, but it _was_ still necessary. 🤦

grant-cleary:
Alrighty! Visual diff tests now look good. Merged in the new goldens. :)

dbatiste:
@dlockhart ... Grant and I are interested in your opinion as a sanity check here. The `d2l-html-block` API has gone through a couple of revisions over the recent past. To recap:

Originally we required consumers to use a `template` element. With the `template` approach, we stamped the content into the shadowDOM where all the html-block styles were (and still are) defined. It also effectively deferred the HTML rendering. 

```html
<d2l-html-block>
   <template> ... user-authored HTML ... </template>
</d2l-html-block>
```
While ideal from a semantic perspective, Lit 2 does not support dynamic expressions in `template` elements so we pivoted from requiring it when we adopted Lit 2. This limitation was problematic for the HTML editor because it requires the ability to do this for its inline mode (toggling back and forth between the edit view and the read-only/rendered view).

When we removed the `template` requirement, consumers were then required to place their user-authored HTML into the slot, which was initially hidden by `d2l-html-block`, and we copied it into the shadowDOM, the end result being similar to the template approach. 

```html
<d2l-html-block>
   ... user-authored HTML ... <-- hidden by slotted selector -->
</d2l-html-block>
```

However, this created at least two issues (that we know of) - a) scripts within such content would run twice, potentially causing performance issues or other issues if such script has side-effects; and b) audio configured with auto-play in such content would still play. So that's where we stand today. 

So this PR: attempts to address those issues by pivoting the API yet again to take the HTML as a property rather than rendered into a slot. The end result is again the same, with the HTML rendered into the `d2l-html-block`'s shadowDOM where html-block styles can target various elements, but it avoids those issues by not rendering the HTML and then copying it. It's not as elegant or semantically correct as the original `template` approach, but achieves the same thing. So this is what the typical usage would look like:

```html
<d2l-html-block html="... user-authored HTML ..."></d2l-html-block>
```

There is another requirement however... non-deferred user-authored HTML. This is HTML that is not sand-boxed and contains script that may depend on page load events, which is therefore unsafe for deferred rendering. This is already supported and used strictly by the LMS MVC control. It is currently enabled by the consumer specifying the `no-deferred-rendering` attribute and placing the HTML into the visible slot. We would plan to continue supporting that, but we don't want consumers using it incorrectly so we would plan to keep the attribute as an explicit way to opt-out of deferred rendering. The LMS MVC control would likely be the only usage of this mode, ever, because the likelihood that we can rely on any user-authored script running successfully in the unsandboxed scope of an app component's shadowDOM is extremely low. So this is what that usage would look like:

```html
<d2l-html-block no-deferred-rendering>
   ... user-authored HTML containing script, as determined by MVC control ...
</d2l-html-block>
```

So the main change here is the pivot to accepting the HTML via the `html` property.

Grant, please jump in if I missed something.

Apologies for the long post - I figured the context might all be useful.

grant-cleary:
> Grant, please jump in if I missed something.

Sums it all up very nicely! It does feel like we're caught between a bunch of not-fantastic options, unfortunately. :(

dlockhart:
Thanks for the recap. This _seems_ like a good compromise to me... we'll obviously have to be careful about HTML-encoding the stuff going into the attribute, but Lit handles that for us. Are newlines in the HTML a concern, as if rendered out as-is they may "break out" of the attribute. Not sure how that works TBH.

grant-cleary:
An interesting side bit to note I suppose... As components expand their domination over our front-end, we'll likely want to re-evaluate the no-deferred-rendering business as a whole. Consider that we're already in a situation where much of our user-authored HTML is rendered asynchronously via other components, which as Dave B. mentioned, already means we're in a case where HTML sensitive to render order will already be wonky in many cases. I'm interested in seeing at what point we pass the line where we can consider just dropping that consideration entirely...

dbatiste:
So in terms of migration - this PR is backwards compatible with current usage of `d2l-html-block`. Those usages will need to switch to use the `html` attribute/property, with the exception of the conditional case rendered from the MVC control. Once they've been updated, we need to update `d2l-html-block` again to remove that support, and enforce using `no-deferred-rendering` in order to utilize the slot. Sound about right?

grant-cleary:
> So in terms of migration - this PR is backwards compatible with current usage of `d2l-html-block`. Those usages will need to switch to use the `html` attribute/property, with the exception of the conditional case rendered from the MVC control. Once they've been updated, we need to update `d2l-html-block` again to remove that support, and enforce using `no-deferred-rendering` in order to utilize the slot. Sound about right?

That's exactly right, yeah. I figure for the actual escalation, I'll backport this change (and the LMS' conditional case), and then in 22.10 I'll fix up existing usages and then clean up the component.

dbatiste (approved):

PR #2772, opened 2022-08-17
Add description property to tag-list-item to better control tooltip - US141912

imcdonald-d2l:
This change will allow us to use the `d2l-tag-list` core component instead of `d2l-labs-multi-select-list` component [in d2l-activity-alignments](https://github.com/Brightspace/d2l-activity-alignments/blob/main/src/components/tags/activity-alignment-tag-list.js#L84). 

We require the tooltip to contain separate info than the text displayed on tag item to give the users context on selected standards aligned to a given activity. This PR addresses this gap by adding a `description` property to `d2l-tag-list-item` to better control what we can display in the tooltip. 

See demo for functionality added:
tag-item-description-property

Rally ticket: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F644105367505

### DIFF
--- a/components/tag-list/demo/tag-list.html
+++ b/components/tag-list/demo/tag-list.html
@@ -86,6 +86,16 @@
 				</div>
 			</d2l-demo-snippet>
 
+			<h2>Tag List with custom tooltip text</h2>
+			<d2l-demo-snippet fullscreen>
+				<d2l-tag-list description="A bunch of example tags">
+					<d2l-tag-list-item text="Longer Example Tag - much much much much much longer" description="display text and description value when truncated"></d2l-tag-list-item>
+					<d2l-tag-list-item text="Example Tag 5" description="custom tooltip text - display description only when not truncated"></d2l-tag-list-item>
+					<d2l-tag-list-item text="Truncated list item without a custom tooltip description"></d2l-tag-list-item>
+					<d2l-tag-list-item text="No tooltip"></d2l-tag-list-item>
+				</d2l-tag-list>
+			</d2l-demo-snippet>
+
 		</d2l-demo-page>
 	</body>
 </html>
--- a/components/tag-list/demo/tag-list.html
+++ b/components/tag-list/demo/tag-list.html
@@ -86,7 +86,7 @@
 				</div>
 			</d2l-demo-snippet>
 
-			<h2>Tag List with custom tooltip text</h2>
+			<h2>Tag List - Custom Tooltip Description</h2>
 			<d2l-demo-snippet fullscreen>
 				<d2l-tag-list description="A bunch of example tags">
 					<d2l-tag-list-item text="Longer Example Tag - much much much much much longer" description="display text and description value when truncated"></d2l-tag-list-item>
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -1,7 +1,7 @@
 import '../button/button-icon.js';
 import '../colors/colors.js';
 import '../tooltip/tooltip.js';
-import { css, html } from 'lit';
+import { css, html, nothing } from 'lit';
 import { announce } from '../../helpers/announce.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { findComposedAncestor } from '../../helpers/dom.js';
@@ -120,6 +120,9 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			.d2l-tag-list-item-tooltip-title-key {
 				font-weight: 600;
 			}
+			d2l-tooltip h4 {
+				margin-top: 0;
+			}
 		`];
 	}
 
@@ -200,14 +203,16 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		));
 	}
 
-	_renderTag(tagContent, isTruncated, description) {
+	_renderTag(tagContent, hasTruncationTooltip, description) {
 		const buttonText = typeof tagContent === 'object'
 			? this.localize('components.tag-list.clear', { value: '' })
 			: this.localize('components.tag-list.clear', { value: tagContent });
 		const hasDescription = !!description;
-		const tooltipTagOverflow = isTruncated || hasDescription ? html`
-				<d2l-tooltip for="${this._id}">
-					${hasDescription ? description : tagContent}
+		const tooltipHeader = hasDescription ? html`<h4>${tagContent}</h4>` : tagContent;
+		const tooltipTagOverflow = hasTruncationTooltip ? html`
+				<d2l-tooltip for="${this._id}" ?show-truncated-only="${!hasDescription}">
+					${tooltipHeader}
+					${hasDescription ? description : nothing}
 				</d2l-tooltip>
 			` : null;
 		const tooltipKeyboardInstructions = this._displayKeyboardTooltip ? html`
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -2,11 +2,11 @@ import '../button/button-icon.js';
 import '../colors/colors.js';
 import '../tooltip/tooltip.js';
 import { css, html, nothing } from 'lit';
+import { heading4Styles, labelStyles } from '../typography/styles.js';
 import { announce } from '../../helpers/announce.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { findComposedAncestor } from '../../helpers/dom.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
-import { labelStyles } from '../typography/styles.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
@@ -39,7 +39,7 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 	}
 
 	static get styles() {
-		return [labelStyles, css`
+		return [labelStyles, heading4Styles, css`
 			:host {
 				display: grid;
 				max-width: 100%;
@@ -120,8 +120,8 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			.d2l-tag-list-item-tooltip-title-key {
 				font-weight: 600;
 			}
-			d2l-tooltip h4 {
-				margin-top: 0;
+			.d2l-heading-4 {
+				margin: 0 0 0.5rem 0;
 			}
 		`];
 	}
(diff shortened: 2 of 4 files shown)


components/tag-list/tag-list-item-mixin.js
```diff
@@ -200,13 +200,14 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		));
 	}
 
-	_renderTag(tagContent, hasTruncationTooltip) {
+	_renderTag(tagContent, hasTruncationTooltip = false, description = '') {
```
devpow112, line 203: Do you need the default parameters here?

components/tag-list/tag-list-item.js
```diff
@@ -9,12 +10,47 @@ class TagListItem extends TagListItemMixin(LitElement) {
...
-			text: { type: String }
+			text: { type: String },
+			/**
+			 * Optional: Text to display in tooltip.
+			 * Tooltip will also include text property value if truncated.
+			 * @type {string}
+			 */
+			description: { type: String },
+			_isTruncated: { attribute: false },
+			_resizeObserver: { attribute: false }
```
devpow112, line 21: I wouldn't add the `_resizeObserver` as an property . Might be better to also just do `this.RequestUpdate()` when `_isTruncated` changes and not have it as an property either.

components/tag-list/tag-list-item-mixin.js
imcdonald-d2l: Nope 👍🏻

components/tag-list/tag-list-item.js
imcdonald-d2l: Can remove `_resizeObserver`, will leave `_isTruncated` as a property so I don't manually have to manually `this.RequestUpdate()` on changes unless Gaudi folks think this should be done.

components/tag-list/demo/tag-list.html
```diff
@@ -86,6 +86,16 @@ <h2>Tag List with Interactive</h2>
 				</div>
 			</d2l-demo-snippet>
 
+			<h2>Tag List with custom tooltip text</h2>
+			<d2l-demo-snippet fullscreen>
+				<d2l-tag-list description="A bunch of example tags">
```
imcdonald-d2l, line 91: Tried testing this component with NVDA and it didn't seem very accessible. Was only reading out the first one and then `list item 2 of 7, list item 3 of 7 etc`. This is with or without my changes. Is this expected?

components/tag-list/tag-list-item.js
```diff
@@ -9,12 +10,46 @@ class TagListItem extends TagListItemMixin(LitElement) {
...
 			 */
-			text: { type: String }
+			text: { type: String },
+			/**
+			 * Optional: Text to display in tooltip.
+			 * Tooltip will also include text property value if truncated.
+			 * @type {string}
+			 */
+			description: { type: String },
+			_isTruncated: { attribute: false }
```
dlockhart, line 20: The Lit 2 way to do this is with state -- all the benefits of properties, but they're not settable by consumers.

```suggestion
			_isTruncated: { state: true }
```

components/tag-list/tag-list-item.js
```diff
@@ -9,12 +10,46 @@ class TagListItem extends TagListItemMixin(LitElement) {
...
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		if (this.description) {
+			// observe resize events so we can update tooltip text when truncated
+			const itemContent = this.shadowRoot.querySelector('.tag-list-item-content');
+			this._onListItemResize();
+			this._resizeObserver = new ResizeObserver(() => this._onListItemResize());
+			this._resizeObserver.observe(itemContent);
```
dlockhart, line 41: The tooltip [has some pretty complex](https://search.d2l.dev/xref/BrightspaceUI/core/components/tooltip/tooltip.js?r=56fecdc9#909) logic already to determine if the target is truncating... so it's unfortunate that in this case we have to run it twice. Tooltips of course have to deal with any kind of target, whereas this code is very specific to tag list items.

How terrible would it be if we just always showed the text + description in the tooltip? Then all this code could go away and we'd just rely on tooltip to show when things are truncated if there's no description, or always show if there is a description.

imcdonald-d2l: Ideally we keep them separate when it is not truncated as that follows our current design. This shouldn't really impact performance I would think as it is a simple resize observer. Would you block this change if we kept it this way?

components/tag-list/demo/tag-list.html
svanherk: We do have an open defect about that: [DE49511](https://rally1.rallydev.com/#/15545167705ud/search?detail=%2Fdefect%2Ff1adaca6-2151-4f51-b6c8-9dbceffa3705&keywords=tag%20list&searchScopeType=project). I do think that this would play into the fix for that - I assume we'd want to always read both the text and the description to the user

devpow112: Yeah we'd want to read both for sure. From looking at the defect it looks like the behaviour noted in the defect matches what we are seeing.

components/tag-list/tag-list-item.js
svanherk: Yeah, tooltip also delays the truncation calculations until the moment when we are about to show the tooltip (or not), rather than on every resize.  That logic was originally added for this use case, I wonder if we can instead just expose that calculation instead of doing it twice.

> How terrible would it be if we just always showed the text + description in the tooltip? Then all this code could go away and we'd just rely on tooltip to show when things are truncated if there's no description, or always show if there is a description.

I was wondering that too, it results in a slightly repetitive tooltip but might be more consistent of an experience for the user.  @geurts do you have thoughts here on how general `d2l-tag-list-items` should behave when their tooltip text is customized?

components/tag-list/demo/tag-list.html
svanherk: Yeah I'm not sure where along the way this broke (it definitely _used_ to work) but I'm surprised we haven't tackled it yet, seems pretty bad. I'll poke about that

components/tag-list/tag-list-item.js
imcdonald-d2l: Potential solution here: https://github.com/BrightspaceUI/core/pull/2772/commits/11b94cf6aeff2258d34d06c5a35e09e709b53528
Just remove the `?show-truncated-only` tooltip check and use the new truncated check I added in `d2l-list-item` to control whether a tooltip is added

dbatiste: I think it does the check regardless of whether the user interacts though ya? I think we would like to defer as much work as possible until needed, especially for components that could be rendered on the page many times. In the future we might even want to avoid rendering such aspects of the component until necessary, but that's a much bigger thing.

imcdonald-d2l:
@dlockhart @svanherk do you know why `ResizeObserver` is causing the following error? When I removed it the error disappeared. Also getting a separate timeout issue I haven't been able to determine the cause of yet. 
image

dlockhart:
> do you know why `ResizeObserver` is causing the following error?

Yep, I've seen it a lot with resize observers -- typically disconnecting them when the component is disconnected solves things but I've also seen that not work.

svanherk:
> > do you know why `ResizeObserver` is causing the following error?
> 
> Yep, I've seen it a lot with resize observers -- typically disconnecting them when the component is disconnected solves things but I've also seen that not work.

Yeah I ran into this in another repo recently where the `ResizeObserver` was already being disconnected properly, but the component just goes into a resizing loop and re-renders too many times and Safari throws an error.  I didn't dig into fixing it because it was a deprecated labs component

components/tag-list/tag-list-item.js
```diff
@@ -9,12 +9,23 @@ class TagListItem extends TagListItemMixin(LitElement) {
...
+			description: { type: String }
 		};
 	}
 
 	render() {
-		return this._renderTag(this.text, true);
+		return this._renderTag(this.text, true, this.description);
+	}
+
+	_onListItemResize() {
```
dlockhart, line 26: This can go away now I think!

imcdonald-d2l: Good catch, missed that! 👍🏻

dlockhart (approved):
This LGTM!

components/tag-list/tag-list-item-mixin.js
```diff
@@ -200,13 +203,16 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		));
 	}
 
-	_renderTag(tagContent, hasTruncationTooltip) {
+	_renderTag(tagContent, hasTruncationTooltip, description) {
 		const buttonText = typeof tagContent === 'object'
 			? this.localize('components.tag-list.clear', { value: '' })
 			: this.localize('components.tag-list.clear', { value: tagContent });
+		const hasDescription = !!description;
+		const tooltipHeader = hasDescription ? html`<h4>${tagContent}</h4>` : tagContent;
```
dbatiste, line 211: Note: with Chrome + VO, the tooltip content is not announced when the tooltip contains the `h4` element. I think @dcostiuc ran into similar behaviour with `d2l-tooltip-help` when lists were placed inside. Perhaps we just want to use one of our heading styles, and forego the semantic heading. From what I gather, this might be due to VO strictly adhering to the idea that [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) element references should only contain text (no complex elements).

imcdonald-d2l: Sounds good, I'll swap it out for a span with the heading styles 👍🏻

components/tag-list/tag-list-item-mixin.js
```diff
@@ -200,13 +203,16 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		));
 	}
 
-	_renderTag(tagContent, hasTruncationTooltip) {
+	_renderTag(tagContent, hasTruncationTooltip, description) {
 		const buttonText = typeof tagContent === 'object'
 			? this.localize('components.tag-list.clear', { value: '' })
 			: this.localize('components.tag-list.clear', { value: tagContent });
+		const hasDescription = !!description;
+		const tooltipHeader = hasDescription ? html`<div class="d2l-heading-4">${tagContent}</div>` : tagContent;
```
imcdonald-d2l, line 211: @dbatiste using a `div` here to get a line break and adjusted margin to be more reasonable. See before/after margin changes:
| Default `margin-bottom` | Adjusted Margin | 
| --- | --- |
| tag-list-before | tag-list-after |

dbatiste: I think it should be good.  I will test it out just to make sure.

dbatiste: So it doesn't announce when the `div` is used either. I also tried a `span` with `display: block` and it does't get announced either. Need to play around with it more.

imcdonald-d2l: @dbatiste saw the conversation on slack regarding this issue, given that there is no immediate solution can this be merged as is since there are existing accessibility issues with this component? (see [resolved comment here](https://github.com/BrightspaceUI/core/pull/2772#discussion_r948250198) regarding the existing issue)

dbatiste: Yeah I think so.

dbatiste (approved):

PR #2775, opened 2022-08-17
fix: don't show conditions opener on whitespace-only slot content

dcostiuc:
Fix for the defect identified in [this](https://d2l.slack.com/archives/C03FC0GDVB7/p1660754175274559) slack thread, 

where it was noticed that the `Visibility Switch` was picking up pure whitespace between the two opening and closing tags as valid "conditions" default slot content.


components/switch/switch-visibility.js
```diff
@@ -88,7 +88,10 @@ class VisibilitySwitch extends LocalizeCoreElement(SwitchMixin(LitElement)) {
 	}
 
 	_handleConditionsSlotChange(e) {
-		this._hasConditions = e.target.assignedNodes({ flatten: true }).length > 0;
+		const nodes = e.target.assignedNodes({ flatten: true });
+		if (nodes[0]?.textContent?.trim()) {
```
dbatiste, line 92: Is there a scenario where `_hasConditions` would not get updated when it should? The code kinda assumes that it would be in its initial state of `false`. Perhaps this should be:
```javascript
this._hasConditions = (nodes[0]?.textContent?.trim() && nodes.length > 0);
```

dcostiuc: We do set the initial value to `false` in the constructor, but you are right, we would want the bool to update in cases where, for example, we initially had slot contents, but then, for whatever reason, they were deleted.

Though, would we even need the `nodes.length > 0` at that point?

dlockhart (approved):

components/switch/switch-visibility.js
dbatiste: Probably not - I think that's already covered by the first check.

dbatiste (approved):

PR #2781, opened 2022-08-18
fix: Visibility Switch > additional documentation tweaks

dcostiuc:
Just some tweaks and improvements to the documentation page for the Visibility Switch, after checking in with Jeff.

Note that the new changes are from Jeff.

### DIFF
--- a/components/switch/README.md
+++ b/components/switch/README.md
@@ -61,9 +61,13 @@ To make your usage of `d2l-switch` accessible, use the following property:
 
 ## Visibility Switch [d2l-switch-visibility]
 
-The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+The `d2l-switch-visibility` component is a special variant for toggling the visibility of activities. Activities can be set to "visible" but still not appear for users due to availability conditions, so the Visibility Switch has three states instead of just two:
 
-It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions ("Visible. Conditions must be met.").
+- Hidden
+- Visible
+- Visible. Conditions must be met
+
+See [Visibility Switch with Conditions](https://daylight.d2l.dev/components/switch/#visibility-switch-with-conditions) for information about the special third state.
 
 <!-- docs: demo live name:d2l-switch-visibility autoSize:false size:small -->
 ```html
@@ -93,9 +97,7 @@ It consists of 3 states: On ("Visible"), Off ("Hidden"), and On with Conditions 
 <!-- docs: end hidden content -->
 
 ### Visibility Switch with Conditions
-It may be helpful to have visibility be dependent on additional conditions being met outside of the Visibility Switch.
-
-Additional information can be added to the label by passing it into the `d2l-switch-visibility` element.
+If an activity is set to `Visible` but also has other conditions affecting its visibility, information about the conditions can be passed in the default slot so it will be available in a tooltip.
 
 <!-- docs: demo code autoSize:false size:large -->
 ```html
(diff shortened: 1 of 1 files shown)


components/switch/README.md
```diff
@@ -61,9 +61,13 @@ To make your usage of `d2l-switch` accessible, use the following property:
 
 ## Visibility Switch [d2l-switch-visibility]
 
-The `d2l-switch-visibility` element is a variant of the generic switch configured with special icons and default text for toggling "visibility".
+The `d2l-switch-visibility` component is a special variant for toggling the visibility of activities. Activities can be set to "visible" but still not appear for users due to availability conditions, so the Visibility Switch has three states instead of just two:
```
dbatiste, line 64: Not sure if we want the documentation to be "activity" specific. @geurts , thoughts?

I think it's good to add clarity about the conditions, but I might reword to avoid the ambiguity around the term "states". I think there is a mismatch in terms of the states that are perceived by the user (hidden, visible, visible with conditions), the states that are assigned by the user (hidden, visible), and the states that are provided by the generic switch mixin API via the `on` property (on/off). Switches don't have the concept of an indeterminate state, but in this case we've kinda tried to present it even though the browser doesn't support it (which unfortunately has also caused us some a11y headaches). In any case, we don't want developers thinking they can get back three different selected states.

Possible wording, assuming the activity-specific verbiage is kept:

>The `d2l-switch-visibility` component is a special variant for toggling the visibility of activities. Activities can be set to "on" but still not appear for users due to availability conditions. Therefore, the `d2l-switch-visibility` can display as "Hidden", "Visible", or "Visible. Conditions must be met" if a description of the conditions is provided.

dcostiuc: >  Switches don't have the concept of an indeterminate state

Kind of a side fun fact - note that technically `aria-checked` (which we use in our switch) _does_ support [a third, indeterminate state](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#:~:text=element%20is%20checked.-,mixed,-for%20checkbox%20and), though just not for the `switch` aria role.

dcostiuc: > The d2l-switch-visibility component is a special variant for toggling the visibility of activities. Activities can be set to "on" but still not appear for users due to availability conditions. Therefore, the d2l-switch-visibility can display as "Hidden", "Visible", or "Visible. Conditions must be met" if a description of the conditions is provided.

I like this wording too 👍

dbatiste: >Kind of a side fun fact - note that technically aria-checked (which we use in our switch) does support [a third, indeterminate state](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#:~:text=element%20is%20checked.-,mixed,-for%20checkbox%20and), though just not for the switch aria role.

Yeah that's interesting - we do support it on our [d2l-input-checkbox](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-checkbox.js#L189) component. The [Mozilla aria switch doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role#description) mentions the `aria-checked` limitation as well.

dcostiuc: Added this in.

Note that I also tweaked [this line](https://github.com/BrightspaceUI/core/pull/2781/commits/2b9252770628b6625b9fa567ca3da083e3f4040e#diff-5e8dfbf840b68fd9dc7a651864d2f6dcc22d1cb02f19e2ceeb56b50a94df5c03R66) so that it doesn't use "state".

dbatiste (approved):

PR #2783, opened 2022-08-19
Add docs ignore on README for `d2l-tag-list-item` properties

imcdonald-d2l:
Daylight sight creates property table from code so adding ignore comment on README so it does not get duplicated.

### DIFF
--- a/components/tag-list/README.md
+++ b/components/tag-list/README.md
@@ -77,9 +77,11 @@ The `d2l-tag-list-item` provides the appropriate `listitem` semantics and stylin
 </d2l-tag-list>
 ```
 
+<!-- docs: start hidden content -->
 ### `d2l-tag-list-item` properties
 
 | Property | Type | Description |
 |--|--|--|
 | `text` | String, required | Text displayed on the tag item. Will truncate automatically if text exceeds a certain width and will display full `text` value in a tooltip. |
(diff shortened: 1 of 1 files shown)


imcdonald-d2l:
Do we want to release this as `fix:` or just let the next release include it? 🤔

devpow112:
> Do we want to release this as `fix:` or just let the next release include it? 🤔

I'd just let the next release have it since it's only an update to docs (unless this causes issues with the daylight site @svanherk)

devpow112 (approved):

dlockhart (approved):

svanherk:
> > Do we want to release this as `fix:` or just let the next release include it? 🤔
> 
> I'd just let the next release have it since it's only an update to docs (unless this causes issues with the daylight site @svanherk)

Deployment of the Daylight site is technically blocked unless we accept the double properties table (which we can just do for now, it's not a huge issue).  But if you did release I could retrigger the Daylight update job and get this fix in before accepting a new deployment

imcdonald-d2l:
> Deployment of the Daylight site is technically blocked unless we accept the double properties table (which we can just do for now, it's not a huge issue). But if you did release I could retrigger the Daylight update job and get this fix in before accepting a new deployment

I'll release a `fix:` patch then 👍🏻

PR #2785, opened 2022-08-19
fix: switch away from using `lit-element`

devpow112:
Looks like this was just missed. Came up during failing build of FRA.


dlockhart (approved):
Thanks!

PR #2791, opened 2022-08-22
Removed deprecated d2l-alert events

oelhanafey:
[Rally ticket](https://rally1.rallydev.com/#/?detail=/userstory/644271182239&fdp=true)

In this PR I removed the `d2l-alert-closed` and `d2l-alert-button-pressed` events that were deprecated in favour of `d2l-alert-close` and `d2l-alert-button-press` respectively. 

All existing uses have been removed:
[d2l-alert-closed](https://search.d2l.dev/search?project=Brightspace&project=Brightspace-Configs&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&project=Scratch&full=%22d2l-alert-closed%22&defs=&refs=&path=&hist=&type=&xrd=&nn=7&si=full&si=full)
[d2l-alert-button-pressed](https://search.d2l.dev/search?project=Brightspace&project=Brightspace-Configs&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&project=Scratch&full=%22d2l-alert-button-pressed%22&defs=&refs=&path=&hist=&type=&xrd=&nn=7&si=full&si=full)
- The `capture-central` repo is no longer used and set to be graveyarded


components/alert/alert-toast.js
```diff
@@ -15,6 +15,7 @@ const states = {
 /**
  *  A component for communicating important information relating to the state of the system and the user's work flow, displayed as a pop-up at the bottom of the screen that automatically dismisses itself by default.
  * @slot - Default content placed inside of the component
+ * @fires d2l-alert-button-press - Dispatched when the alert's action button is clicked
```
dlockhart, line 18: > ...but preventing the original event from bubbling may break existing usages and these would need to be changed

Could you do a quick audit and see if any of the places listening for this are listening on/from a `d2l-alert-toast`? If it's only a few places, then it might be cleanest to trap it and rename it as you said, but if it's a lot then maybe leaving it and documenting it is better.

oelhanafey: Yeah, it looks like it is only used in a single place, which I needed to change anyway so we can go with a `d2l-alert-toast-button-press` instead.

dlockhart: Nice!

eKoopmans: So this will be updated to `d2l-alert-toast-button-press`?

eKoopmans (commented):
All looks good, I think just waiting on the switch from `d2l-alert-button-press` to `d2l-alert-toast-button-press`.

components/alert/alert-toast.js
oelhanafey: Yep, sorry I hadn't updated this PR yet since I'm still waiting on my PR in another repo to be merged. Fixed now!

dbatiste (approved):
Looks good!

eKoopmans (approved):

PR #2795, opened 2022-08-23
Added d2l-alert-toast-button-press event

oelhanafey:
Added the `d2l-alert-toast-button-press` event to the `<d2l-alert-toast>` component. The component will listen for the `d2l-alert-button-press` event, stop it from propagating and then dispatch it's own event.

The only existing usage of a button press event in a `d2l-alert-toast` is listening for the `d2l-alert-button-pressed` event which I did not stop from propagating so this change should not break any existing uses of the component. In a later PR, I plan to remove that event from the `<d2l-alert>` component after existing uses have been changed.

### DIFF
--- a/components/alert/test/alert-toast.test.js
+++ b/components/alert/test/alert-toast.test.js
@@ -12,6 +12,17 @@ describe('d2l-alert-toast', () => {
 
 	});
 
+	describe('button-press', () => {
+
+		it('should fire "d2l-alert-toast-button-press" event when alert button is pressed', async() => {
+			const el = await fixture(html`<d2l-alert-toast button-text="Click Me" open>message</d2l-alert-toast>`);
+			const alert = el.shadowRoot.querySelector('d2l-alert');
+			setTimeout(() => alert.dispatchEvent(new CustomEvent('d2l-alert-button-press', { bubbles: true, composed: true })));
+			await oneEvent(el, 'd2l-alert-toast-button-press');
+		});
+
+	});
+
 	describe('close', () => {
 
 		it('should close when close button is clicked', async() => {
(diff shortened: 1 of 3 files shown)


components/alert/test/alert-toast.test.js
```diff
@@ -12,6 +12,17 @@ describe('d2l-alert-toast', () => {
 
 	});
 
+	describe('button-press', () => {
+
+		it('should fire "d2l-alert-toast-button-press" event when alert button is pressed', async() => {
+			const el = await fixture(html`<d2l-alert-toast button-text="Click Me" open>message</d2l-alert-toast>`);
+			const alert = el.shadowRoot.querySelector('d2l-alert');
+			setTimeout(() => alert.dispatchEvent(new CustomEvent('d2l-alert-button-press', { bubbles: true, composed: true })));
```
dbatiste, line 20: This test assumes that `d2l-alert` continues to dispatch `d2l-alert-button-press`. If that changed, this test would still pass but the toast component would no longer be working correctly. A more comprehensive test would call `click()` on the underlying button, but it requires poking even further into the component's local DOM.

dbatiste (approved):

components/alert/test/alert-toast.test.js
oelhanafey: Yeah, both ways are subject to being broken if the component is changed in the future (e.g. if the button is changed). My train of thought was that this test should not be testing the inner workings of `<d2l-alert>`, rather it should assume that `<d2l-alert>` works and use the API to test the `<d2l-alert-toast>` component. But you're right that if the `<d2l-alert>` API is changed, then it would not break this test.

dbatiste: Yep that's fair.

PR #2797, opened 2022-08-24
Fix the grid focus logic.

dbatiste:
This PR fixes an issue with the `grid` focus logic - specifically when focus is being moved backwards in the list (pressing `[up arrow]` key) and the previous adjacent list-item does not have a focusable element in the corresponding cell. 

The logic would simply use `getPreviousFocusable` with respect to the adjacent list-item as a fallback, which would cause focus to be applied to **any** previous focusable **before** the adjacent list-item, effectively skipping the item. What we really want is for focus to be applied to another cell in that list-item, preferably the primary action, and if that still doesn't work, then do the ultimate generic fallback.

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -208,17 +208,13 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		if (!cell) return;
 
 		const firstFocusable = getFirstFocusableDescendant(cell);
-		if (!firstFocusable) {
-			const listItem = findComposedAncestor(this, node => node.role === 'rowgroup');
-			if (listItem) {
-				const nextFocusable = previous ? getPreviousFocusable(listItem) : getNextFocusable(listItem);
-				if (nextFocusable) nextFocusable.focus();
-			}
-			return;
-		}
+		if (!firstFocusable) return;
 
-		if (itemNum === 1 || !this._focusNextWithinCell(firstFocusable, itemNum)) {
+		if (itemNum === 1) {
 			firstFocusable.focus();
+			return firstFocusable;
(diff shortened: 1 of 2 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -208,17 +208,13 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		if (!cell) return;
 
 		const firstFocusable = getFirstFocusableDescendant(cell);
-		if (!firstFocusable) {
-			const listItem = findComposedAncestor(this, node => node.role === 'rowgroup');
-			if (listItem) {
-				const nextFocusable = previous ? getPreviousFocusable(listItem) : getNextFocusable(listItem);
```
dbatiste, line 214: This was the culprit - if the adjacent list-item did not have a focusable in the corresponding cell, then it would fall back to this very general method of locating the next/previous focusable.

components/list/list-item-generic-layout.js
```diff
@@ -208,17 +208,13 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
-				if (nextFocusable) nextFocusable.focus();
-			}
-			return;
-		}
+		if (!firstFocusable) return;
 
-		if (itemNum === 1 || !this._focusNextWithinCell(firstFocusable, itemNum)) {
+		if (itemNum === 1) {
```
dbatiste, line 213: This logic has been updated to return the element that focus was applied to. I retained the naming of the method (as opposed to introducing a try-method) in effort to be consistent with other methods here. The consuming method can then choose the appropriate course of action.

dlockhart (approved):

PR #2800, opened 2022-08-25
Spike - table-wrapper with selection components/bulk actions

dbatiste:
This spike shows what would generally be required to support using the selection components similar to `d2l-list`. It includes the following, which automatically handles wire-up for the standard action behaviours.

* `d2l-selection-input`
* `d2l-selection-select-all`
* `d2l-selection-summary`
* `d2l-selection-action`
* `d2l-selection-action-dropdown`
* `d2l-selection-action-menu-item`
* `SelectionMixin` - required to be extended by `d2l-table-wrapper`

What's not covered here is:

* layout as per Design - might require a header component like `d2l-list-header` (ex. spacing, overflow, etc)
* probably require a new slot for the actions header so we can control layout
* stickiness - likely we want the header actions to stick along with the actual table header
* `d2l-selection-select-all-pages` - likely need a new component to manage things like `d2l-list-header`
* what happens to the actions header when scroll-wrapper is in play and the table is overflowing?

### DIFF
--- a/components/table/demo/table-test.js
+++ b/components/table/demo/table-test.js
@@ -1,3 +1,14 @@
+import '../../dropdown/dropdown-button-subtle.js';
+import '../../dropdown/dropdown-menu.js';
+import '../../menu/menu.js';
+import '../../menu/menu-item.js';
+import '../../selection/selection-action.js';
+import '../../selection/selection-action-dropdown.js';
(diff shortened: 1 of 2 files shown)


dbatiste:
Closing this PR since it was just a spike and we no longer need to reference anything here.

PR #2807, opened 2022-08-30
Removed empty state illustration slotted styles

oelhanafey:
This is a fix after my previous empty state change #2806. I removed all `::slotted` styles from the illustration slot. `::slotted` styles only work on top level elements so to make an `svg` or `img` visible they would need to be top level elements. In order to get this working in the LMS without directly editing the `Image` control, the image needs to be wrapped in a container like a `div` and so it will no longer be a top level element and would have `display: none`. 

e.g.
```
w.WriteTagBegin( "div" );
  w.WriteAttribute( "slot", "illustration" );
  w.WriteTagBeginRightChar();
  viewData.IllustrationView.Render( rc );
w.WriteTagEnd( "div" );
```

There doesn't seem to be another way to control what goes in the slot so it looks like we will just need to trust the consumer to use it correctly.

### DIFF
--- a/components/empty-state/empty-state-styles.js
+++ b/components/empty-state/empty-state-styles.js
@@ -48,15 +48,6 @@ export const emptyStateIllustratedStyles = css`
 		text-align: center;
 	}
 
-	.illustration-slot::slotted(*) {
-		display: none;
-	}
-
-	.illustration-slot::slotted(img:first-of-type),
-	.illustration-slot::slotted(svg:first-of-type) {
-		display: inline-block;
-	}
-
 	svg {
 		height: 100%;
 		max-width: 500px;


dbatiste:
Could the EmptyState MVC control accept an `IImage` property. Then, in the renderer for EmptyState MVC control we render the `img` tag like the `Image` control does, except we have the ability to assign a `slot` attribute without updating `Image`? With that approach, we can avoid the wrapper `div` and keep the web component locked down.

oelhanafey:
> Could the EmptyState MVC control accept an `IImage` property? Then, in the renderer for EmptyState MVC control we render the `img` tag like the `Image` control does, except we have the ability to assign a `slot` attribute without updating `Image`? With that approach, we can avoid the wrapper `div` and keep the web component locked down.

Hmm I'm not sure I understand. If we render the `img` tag like the `Image` control like this: `viewData.IllustrationView.Render( rc )`, then it would render the whole `img` tag and there would be no opportunity to add the `slot`.

Are you suggesting we copy the `Image` renderer code into the `EmptyState` renderer and augment it with the added `slot` attribute? The problem with that is that now `Image` would need to be updated in 2 places.

dbatiste:
> Are you suggesting we copy the `Image` renderer code into the `EmptyState` renderer and augment it with the added `slot` attribute? The problem with that is that now `Image` would need to be updated in 2 places.

Yes. It would render ``, but EmptyState MVC control would accept an `IImage` (data), not an `Image` (view/control). The parts of the rendering that are required could be copied (pretty minimal), or we could find a way to re-use it.

I'm not a super huge fan of adding a `Slot` property to `Image` control because it begins leaking HTML API on our MVC control API, which we usually kind of avoid. MVC controls have their own way of being used by other controls, so introducing a `Slot` property feels pretty weird to me.

PR #2811, opened 2022-09-02
Add support for async renderer fetching in html-block component

grant-cleary:
So there's a bit of an issue with how the non-default renderers load in the html-block component. Specifically, they're _very_ order-dependent and we currently don't have a fantastic way to guarantee that ordering.

This PR is the first of a few meant to bridge that gap - it refactors the html-block component to be resilient to asynchronously loaded renders, while leaving the existing logic in place to cover our bases until existing renderers are switched over.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -1,9 +1,9 @@
 import '../colors/colors.js';
-import { codeStyles, HtmlBlockCodeRenderer } from '../../helpers/prism.js';
+import { codeStyles, createHtmlBlockRenderer as createCodeRenderer } from '../../helpers/prism.js';
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { createHtmlBlockRenderer as createMathRenderer } from '../../helpers/mathjax.js';
 import { HtmlAttributeObserverController } from '../../controllers/attributeObserver/htmlAttributeObserverController.js';
-import { HtmlBlockMathRenderer } from '../../helpers/mathjax.js';
 import { requestInstance } from '../../mixins/provider-mixin.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
@@ -118,10 +118,15 @@ export const htmlBlockContentStyles = css`
 
 let renderers;
 
-const getRenderers = () => {
+const getRenderers = async() => {
 	if (renderers) return renderers;
-	const tempRenderers = requestInstance(document, 'html-block-renderers');
-	const defaultRenderers = [ new HtmlBlockMathRenderer(), new HtmlBlockCodeRenderer() ];
+	let tempRenderers = requestInstance(document, 'html-block-renderers');
+	if (!tempRenderers) {
+		const tempRendererLoader = requestInstance(document, 'html-block-renderer-loader');
+		tempRenderers = await tempRendererLoader.renderers();
+	}
+
+	const defaultRenderers = [ createMathRenderer(), createCodeRenderer() ];
 	renderers = (tempRenderers ? [ ...defaultRenderers, ...tempRenderers ] : defaultRenderers);
 	return renderers;
 };
@@ -193,13 +198,12 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		this.noDeferredRendering = false;
 		this._hasSlottedContent = false;
 
-		const rendererContextAttributes = getRenderers().reduce((attrs, currentRenderer) => {
+		getRenderers().then(renderers => renderers.reduce((attrs, currentRenderer) => {
 			if (currentRenderer.contextAttributes) currentRenderer.contextAttributes.forEach(attr => attrs.push(attr));
 			return attrs;
-		}, []);
-
-		if (rendererContextAttributes.length === 0) return;
-		this._contextObserverController = new HtmlAttributeObserverController(this, ...rendererContextAttributes);
+		}, [])).then(rendererContextAttributes => {
+			this._contextObserverController = new HtmlAttributeObserverController(this, ...rendererContextAttributes);
+		});
 	}
 
 	connectedCallback() {
@@ -257,6 +261,10 @@ class HtmlBlock extends RtlMixin(LitElement) {
 
 	_contextChanged() {
 		if (!this._contextObserverController) return false;
+		if (!this._contextKeys) {
+			this._updateContextKeys();
+			return true;
+		}
 
 		if (this._contextKeys.size !== this._contextObserverController.values.size) return true;
 		for (const [attr, val] of this._contextKeys) {
@@ -286,7 +294,8 @@ class HtmlBlock extends RtlMixin(LitElement) {
 	}
 
 	async _processRenderers(elem) {
-		for (const renderer of getRenderers()) {
+		const renderers = await getRenderers();
+		for (const renderer of renderers) {
 			if (this._contextObserverController && renderer.contextAttributes) {
 				const contextValues = new Map();
 				renderer.contextAttributes.forEach(attr => contextValues.set(attr, this._contextObserverController.values.get(attr)));
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -121,7 +121,7 @@ let renderers;
 const getRenderers = async() => {
 	if (renderers) return renderers;
 	const rendererLoader = requestInstance(document, 'html-block-renderer-loader');
-	const tempRenderers = await rendererLoader.getRenderers();
+	const tempRenderers = rendererLoader ? await rendererLoader.getRenderers() : undefined;
 	const defaultRenderers = [ createMathRenderer(), createCodeRenderer() ];
 	renderers = (tempRenderers ? [ ...defaultRenderers, ...tempRenderers ] : defaultRenderers);
 	return renderers;
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -121,7 +121,7 @@ let renderers;
 const getRenderers = async() => {
 	if (renderers) return renderers;
 	const rendererLoader = requestInstance(document, 'html-block-renderer-loader');
-	const tempRenderers = await rendererLoader.renderers();
+	const tempRenderers = await rendererLoader.getRenderers();
 	const defaultRenderers = [ createMathRenderer(), createCodeRenderer() ];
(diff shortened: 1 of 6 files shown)


components/html-block/html-block.js
```diff
@@ -118,10 +118,15 @@ export const htmlBlockContentStyles = css`
 
 let renderers;
 
-const getRenderers = () => {
+const getRenderers = async() => {
 	if (renderers) return renderers;
-	const tempRenderers = requestInstance(document, 'html-block-renderers');
-	const defaultRenderers = [ new HtmlBlockMathRenderer(), new HtmlBlockCodeRenderer() ];
+	let tempRenderers = requestInstance(document, 'html-block-renderers');
```
grant-cleary, line 123: You know, I think I'm being overly zealous here. Because we never hooked up the @mentions renderer, nothing is actually providing any of these here. I think I might just remove the backwards-compatible stuff.

controllers/attributeObserver/htmlAttributeObserverController.js
```diff
@@ -14,9 +14,9 @@ export class HtmlAttributeObserverController {
 		);
 
 		this._host = host;
-		host.addController(this);
 		this._attributes = attributes;
 		this.values = new Map();
+		host.addController(this);
```
grant-cleary, line 19: I'm going to re-test to see if this reordering is still necessary, or if it's something I can just drop because I'm being more careful about order-of-operations elsewhere.

helpers/mathjax.js
```diff
@@ -77,6 +77,10 @@ export class HtmlBlockMathRenderer {
 
 }
 
+export function createHtmlBlockRenderer() {
```
grant-cleary, line 80: Given the new way we want to do renderer registration, I figured it would be nice if all of the renderer imports had a consistent API that could be used to instantiate the renderer. That way the provider can do something generic like:
```
import(path).then(renderer => renderer.createHtmlBlockRenderer())
```

components/html-block/html-block.js
```diff
@@ -257,6 +261,10 @@ class HtmlBlock extends RtlMixin(LitElement) {
 
 	_contextChanged() {
 		if (!this._contextObserverController) return false;
+		if (!this._contextKeys) {
```
grant-cleary, line 264: Given the asynchronous nature of the controller instantiation now, it's possible that the controller won't yet exist when `firstUpdated` runs and we can't guarantee this will be initialized yet.

grant-cleary:
Oh shoot. I forgot that the tests are going to fail because the test renderers aren't using the new function.

grant-cleary:
Visual diff failures are surprising because the demo works fine. Let me dig in; might be a timing problem now that the rendering is delayed a bit.

components/html-block/html-block.js
```diff
@@ -118,10 +118,15 @@ export const htmlBlockContentStyles = css`
...
 let renderers;
 
-const getRenderers = () => {
+const getRenderers = async() => {
 	if (renderers) return renderers;
-	const tempRenderers = requestInstance(document, 'html-block-renderers');
-	const defaultRenderers = [ new HtmlBlockMathRenderer(), new HtmlBlockCodeRenderer() ];
+	let tempRenderers = requestInstance(document, 'html-block-renderers');
+	if (!tempRenderers) {
+		const tempRendererLoader = requestInstance(document, 'html-block-renderer-loader');
```
dbatiste, line 125: Nit: this could be named this variable `rendererLoader`.

components/html-block/html-block.js
```diff
@@ -193,13 +198,12 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+		getRenderers().then(renderers => renderers.reduce((attrs, currentRenderer) => {
 			if (currentRenderer.contextAttributes) currentRenderer.contextAttributes.forEach(attr => attrs.push(attr));
 			return attrs;
-		}, []);
-
-		if (rendererContextAttributes.length === 0) return;
-		this._contextObserverController = new HtmlAttributeObserverController(this, ...rendererContextAttributes);
+		}, [])).then(rendererContextAttributes => {
+			this._contextObserverController = new HtmlAttributeObserverController(this, ...rendererContextAttributes);
+		});
```
dbatiste, line 206: Should there be an extra `then` statement here that includes a call to `requestUpdate`? Or put another way, are we relying on `HtmlAttributeObserverController` calling `requestUpdate` due to the context attributes being updated, and is that good/safe? FWIW, I'm not sure calling `requestUpdate` in the constructor is safe either, even though it would be several micro-tasks later... a full task might be necessary to call it.

components/html-block/html-block.js
```diff
@@ -286,7 +294,8 @@ class HtmlBlock extends RtlMixin(LitElement) {
 	}
 
 	async _processRenderers(elem) {
-		for (const renderer of getRenderers()) {
+		const renderers = await getRenderers();
```
dbatiste, line 297: So in the ideal case, the renderers will already be loaded, and the `Promise` returned from `getRenderers` here will already be resolved. But they might not be loaded yet, so we need to wait on that promise. Sound right?

helpers/mathjax.js
dbatiste: I see, so that's why the exported methods are all named the same.

components/html-block/html-block.js
grant-cleary: That's the idea, yeah.

components/html-block/html-block.js
grant-cleary: Right, so I originally did do something like that, but was leery about it because (like you said) who knows if the component would be rendered by then. The controller will always request an update when the attributes change, including when they're first set, and it kind of has to work that way for the controller to function, so my inclination is that this change is safe. Whether it's _good_ is a different question of course. If there's concern there, I can look at forcing an update explicitly in the component.

components/html-block/html-block.js
grant-cleary: Oh yeah, good point. Just slapped `loader` at the end but I guess it's not really `temp` anymore. 😆

helpers/mathjax.js
grant-cleary: Yeah. I don't feel super strongly about this. The alternative is that the provider would need to do something like:
```
import(path).then(renderer => new renderer.WhateverTheRendererClassIs())
```
I figured the way I did it was marginally cleaner, but I could see benefits to both. Needing to alias the default imports is certainly a bit cumbersome.

components/html-block/html-block.js
dbatiste: Nit: maybe the loader's method should be `getRenderers()` instead of `renderers()`?

helpers/mathjax.js
dbatiste: So the loader module would look something like...?
```javascript
provideInstance(document, 'html-block-renderer-loader', {
	getRenderers() {
		return Promise.all(
			['rendererPath1', `rendererPath2`, ...].map(path => import(path).then(renderer => renderer.createHtmlBlockRenderer()))
		);
	}
});
```
vs.
```javascript
provideInstance(document, 'html-block-renderer-loader', {
	getRenderers() {
		return Promise.all([
			import('rendererPath1').then(renderer => renderer.WhateverTheRenderer1MethodIs()),
			import('rendererPath2').then(renderer => renderer.WhateverTheRenderer2MethodIs()),
			...
		]);
	}
});
```
Yeah it's marginally cleaner. The aliasing of the imports is only something that needs to be done for the default renderers that html-block references directly. Unless I am missing something.

Also, I _think_ that import chain in the first case would resolve to the renderers, but I'm not 100% sure. Either way, looking good on the html-block end of things.

grant-cleary: Yeah, what I have mocked up in my local provider looks like this:
```
provideInstance(document, 'html-block-renderer-loader', {
	_renderers: [
		() => import('rendererPath').then(renderer => renderer.createHtmlBlockRenderer())
	],
	getRenderers() {
		return Promise.all(this._renderers.map(action => action()));
	}
});
```
The structure is a bit strange, but the `import` statements need to live outside of `getRenderers` or the scope will be wrong and they won't resolve properly. It's a bit of a weird structure but it works and does keep things deferred. 😆 

And yeah, I've confirmed locally that this does work as expected, so in terms of functionality I think we're good.

components/html-block/html-block.js
grant-cleary: Yeah, I think that's a better idea. I originally had it implemented as a local getter, which is where the name came from, but that had some quirks I didn't like.

controllers/attributeObserver/htmlAttributeObserverController.js
grant-cleary: Ah, I know why this is necessary now. Turns out reactive controllers _can_ be hooked up after the host is connected, but if that happens Lit immediately calls `hostConnected` when the controller is added to the host. With the new async workflow, the controller now instantiates after the host is connected and the previous code explodes because `this._attributes` is undefined.

components/html-block/test/html-block.visual-diff.js
```diff
@@ -42,7 +42,9 @@ describe('d2l-html-block', () => {
 		it(info.name, async function() {
 			const rect = await visualDiff.getRect(page, info.selector);
 			if (info.action) await info.action(info.selector);
-			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+			setTimeout(async() => {
```
dbatiste, line 45: I'm not sure the test will actually wait for the screenshot.  You might need something like:
```javascript
await new Promise(resolve => setTimeout(resolve, 0));
```

components/html-block/test/html-block.test.js
```diff
@@ -143,7 +144,7 @@ describe('d2l-html-block', () => {
 		expect(spans[1].innerHTML).to.equal('2');
 	});
 
-	it('should do async replacements', async() => {
+	it.only('should do async replacements', async() => {
```
dbatiste, line 147: Is this `only` intentional/temp?

grant-cleary: Shoot, no. Definitely just slipped in by accident from my debugging.

components/html-block/test/html-block.visual-diff.js
grant-cleary: Ah, of course.

dbatiste (approved):

PR #2820, opened 2022-09-06
fix: Description List - Update README

jameswklassen:
## Description
Update the README for `d2l-dl-wrapper`

### DIFF
--- a/components/description-list/README.md
+++ b/components/description-list/README.md
@@ -1,5 +1,5 @@
 # Description List
-A Description List displays information in a vertical list of key-value pairs. Common usages could include terms & definitions, or fields & values.
+A Description List displays static pairs of terms and definitions in a vertical list. In the standard HTML element `dl` these are called "terms" `dt` and "definitions" `dd`.
 
 <!-- docs: demo -->
 ```html
@@ -10,7 +10,6 @@ A Description List displays information in a vertical list of key-value pairs. C
 ```
 
 ## Best Practices
-Use a Description List to display static information that pairs property names and values. In the standard HTML description list `<dl>` element these are called terms `<dt>` and definitions `<dd>`.
 
 <!-- docs: start best practices -->
 <!-- docs: start dos -->
--- a/components/description-list/README.md
+++ b/components/description-list/README.md
@@ -10,16 +10,20 @@ A Description List displays information in a vertical list of key-value pairs. C
 ```
 
 ## Best Practices
+Use a Description List to display static information that pairs property names and values. In the standard HTML description list `<dl>` element these are called terms `<dt>` and definitions `<dd>`.
+
 <!-- docs: start best practices -->
 <!-- docs: start dos -->
-* Terms and descriptions are 1-to-1
+* Keep terms short and scannable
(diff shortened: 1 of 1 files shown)


components/description-list/README.md
```diff
@@ -10,16 +10,20 @@ A Description List displays information in a vertical list of key-value pairs. C
 ```
 
 ## Best Practices
+Use a Description List to display static information that pairs property names and values. In the standard HTML description list `<dl>` element these are called terms `<dt>` and definitions `<dd>`.
```
jameswklassen, line 13: cc: @geurts 

This is the first time we've had text preceding the best practices.

This surfaced a minor spacing issue which I've tried addressing in https://github.com/BrightspaceUI/documentation/pull/1048.

image

components/description-list/README.md
```diff
@@ -60,9 +67,9 @@ When the component width is less than the `breakpoint` property, the side-by-sid
 
 ## Description List Wrapper [d2l-dl-wrapper]
 
-The `d2l-dl-wrapper` can be combined with `descriptionListStyles` to apply styling and resize behavior to native `dl` elements.
+The `d2l-dl-wrapper` component can be combined with `descriptionListStyles` to apply styling and resize behavior to native `dl` elements.
 
-<!-- docs: demo live name:d2l-test-dl -->
+<!-- docs: demo live name:d2l-test-dl display:block -->
```
oelhanafey, line 72: Does the `display: block` do anything? I haven't seen this used before

components/description-list/README.md
oelhanafey: I wonder if the text in _Best Practices_ would fit better under another section. I'll let @geurts chime in on that.

oelhanafey (commented):
`feat:` will trigger a minor version bump for the core repo. For docs changes, we often use `docs:` which doesn't trigger a release at all or if you need the docs to update sooner you can trigger a patch version bump with `fix:`.

components/description-list/README.md
jameswklassen: Yeah, this changes the `display` selector in the demo (defaults to flex). This felt like the cleanest way to resolve some of the resize issues I was seeing in the live demo.

jameswklassen:
> `feat:` will trigger a minor version bump for the core repo. For docs changes, we often use `docs:` which doesn't trigger a release at all or if you need the docs to update sooner you can trigger a patch version bump with `fix:`.

Yeah I could do `docs:`, but I think a new version needs to be released for daylight to pick it up (?).

components/description-list/README.md
oelhanafey: oo good to know, thanks!

oelhanafey:
> Yeah I could do `docs:`, but I think a new version needs to be released for daylight to pick it up (?).

Yeah if you want to trigger a release, then `fix:` would be better since it just triggers a patch version bump instead of a minor version bump

dbatiste (approved):

oelhanafey (approved):

eKoopmans (approved):

dbatiste (approved):

PR #2824, opened 2022-09-07
Make sure context attribute observer is loaded before rendering

grant-cleary:
We always expect this observer to load eventually, so it seems reasonable for us to wait for it. If we don't, we can get into an unfortunate race condition because the observer doesn't trigger a host update when it first registers.

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -195,7 +195,7 @@ class HtmlBlock extends RtlMixin(LitElement) {
 		this._hasSlottedContent = false;
 
 		this._contextObserverControllerResolve = undefined;
-		this._contextObserverControllerLoaded = new Promise(resolve => {
+		this._contextObserverControllerInitialized = new Promise(resolve => {
 			this._contextObserverControllerResolve = resolve;
 		});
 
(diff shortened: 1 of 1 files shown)


components/html-block/html-block.js
```diff
@@ -194,11 +194,17 @@ class HtmlBlock extends RtlMixin(LitElement) {
...
+		this._contextObserverControllerResolve = undefined;
+		this._contextObserverControllerLoaded = new Promise(resolve => {
+			this._contextObserverControllerResolve = resolve;
+		});
+
 		getRenderers().then(renderers => renderers.reduce((attrs, currentRenderer) => {
 			if (currentRenderer.contextAttributes) currentRenderer.contextAttributes.forEach(attr => attrs.push(attr));
 			return attrs;
 		}, [])).then(rendererContextAttributes => {
 			this._contextObserverController = new HtmlAttributeObserverController(this, ...rendererContextAttributes);
```
grant-cleary, line 206: I could have wrapped this in a promise directly, but I think it's still ideal if the renderer promise starts resolving ASAP, not just when we first try to access the controller.

PR #2827, opened 2022-09-08
fix: Add outline to tooltips for accessibility

eKoopmans:
This PR adds a semi-transparent outline to all tooltips, to help them stand out against dark backgrounds.

Note that for the "pointer" of the tooltip, it was necessary to create a duplicate element explicitly for the outline, because it needs to be clipped differently than the pointer itself.

Rally: https://rally1.rallydev.com/#/?detail=/defect/645635571497&fdp=true


eKoopmans:
Screenshots of how it looked when using a single "pointer" vs a separate element for the outline (border set to green for emphasis):

Single pointer:


Separate pointer element for the outline:

components/tooltip/test/tooltip.visual-diff.js
```diff
@@ -62,4 +62,21 @@ describe('d2l-tooltip', () => {
...
 
+	[
+		'position-top',
+		'position-bottom',
+		'position-right',
+		'position-left'
+	].forEach((testName) => {
+
+		it(`dark-background-${testName}`, async function() {
+			await page.addStyleTag({ content: 'body { background-color: black }' });
```
dbatiste, line 73: If these sets of tests were in their own `describe` block, then the style could be added just once in `before(...)`, and it could be cleaned up in `after(...)`. Though Puppeteer is probably smart enough not to add it twice, and it would be a no-op anyway. Not a big deal since there are no other tests that follow this set.

eKoopmans: The tests have an `afterEach(async() => await page.reload());`, so I still need to apply the styles before each test - but I like moving them into a describe block and doing that styling with a `beforeEach`.

dbatiste: Ah cool... 👍 .  I didn't notice the reload. 😄

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -34,14 +41,14 @@ describe('d2l-list-item-drag-handle', () => {
 			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 		});
 
-		it('keyboard-mode', async function() {
+		itWithReload('keyboard-mode', async function() {
```
dbatiste, line 44: Interesting... what was happening here? Was this always kinda flakey?

eKoopmans: Just another retry issue - first attempt is fine, but retries are different. Since there's an intended diff here, it was retrying and ending up wrong (tooltip was no longer displaying).

I created the `itWithReload` function to just do a page reload after each attempt, I could have added `page.reload()` on an afterEach for all the tests, or stuck this in its own describe block and done an afterEach there, but this seemed cleaner.

eKoopmans: Also, there might have been a less nuclear option than `page.reload()` to get retries to work, but I didn't want to dig into the specifics of what would and wouldn't put it back into a "clean" state 😄.

dbatiste (approved):

components/list/test/list-item-drag-handle.visual-diff.js
dbatiste: I think that keyboard tooltip only gets displayed once per page load per list, or once per page load, so the reload is probably the safest.

PR #2831, opened 2022-09-08
fix: Description List - use more specific css selectors

jameswklassen:
## Description

Since `descriptionListStyles` are being applied as a global stylesheet, the CSS selectors should be more specific.

In order to support complex CSS selectors in our style generation functions, the regex for `_isValidCssSelector` had to be updated

### DIFF
--- a/components/typography/styles.js
+++ b/components/typography/styles.js
@@ -2,7 +2,7 @@ import '../colors/colors.js';
 import { css, unsafeCSS } from 'lit';
 
 export const _isValidCssSelector = (selector) => {
-	const re = /[#.]?([a-zA-Z0-9-_]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_]+)?/g;
+	const re = /([a-zA-Z0-9-_ >.#]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_ >.#]+)?/g;
 	const match = selector.match(re);
 
 	return !!match && match.length === 1 && match[0].length === selector.length;
@@ -52,7 +52,7 @@ export const bodyStandardStyles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateBodyCompactStyles = (selector) => {
-	// if (!_isValidCssSelector(selector)) return;
+	if (!_isValidCssSelector(selector)) return;
 
 	selector = unsafeCSS(selector);
 	return css`
@@ -233,7 +233,7 @@ export const heading4Styles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateLabelStyles = (selector) => {
-	// if (!_isValidCssSelector(selector)) return;
+	if (!_isValidCssSelector(selector)) return;
 
 	selector = unsafeCSS(selector);
 	return css`
--- a/components/typography/styles.js
+++ b/components/typography/styles.js
@@ -52,7 +52,7 @@ export const bodyStandardStyles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateBodyCompactStyles = (selector) => {
-	if (!_isValidCssSelector(selector)) return;
+	// if (!_isValidCssSelector(selector)) return;
 
 	selector = unsafeCSS(selector);
 	return css`
@@ -233,7 +233,7 @@ export const heading4Styles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateLabelStyles = (selector) => {
-	if (!_isValidCssSelector(selector)) return;
+	// if (!_isValidCssSelector(selector)) return;
 
 	selector = unsafeCSS(selector);
 	return css`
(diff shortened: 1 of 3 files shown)


components/typography/styles.js
```diff
@@ -52,7 +52,7 @@ export const bodyStandardStyles = css`
  * A private helper method that should not be used by general consumers
  */
 export const _generateBodyCompactStyles = (selector) => {
-	if (!_isValidCssSelector(selector)) return;
+	// if (!_isValidCssSelector(selector)) return;
```
dbatiste, line 55: Do you recall why we explicitly didn't allow this the complex selector case?

https://github.com/BrightspaceUI/core/blob/main/components/typography/test/styles.test.js#L28

dbatiste: Oh it was [this](https://github.com/BrightspaceUI/core/pull/2698#discussion_r931527209). Maybe be we can allow complex selectors, but not if they include `:host`.

jameswklassen: There was no explicit reason we didn't include more complex cases. I think we wanted to start with a minimum viable solution, and not build more than we needed at the time.

Just curious, what's the danger of allowing complex selectors that include `:host`?

components/typography/styles.js
```diff
@@ -2,7 +2,7 @@ import '../colors/colors.js';
 import { css, unsafeCSS } from 'lit';
 
 export const _isValidCssSelector = (selector) => {
-	const re = /[#.]?([a-zA-Z0-9-_]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_]+)?/g;
+	const re = /([a-zA-Z0-9-_ >.#]+)(\[[a-zA-Z0-9-_ >.#]+\])?([a-zA-Z0-9-_ >.#]+)?/g;
```
jameswklassen, line 5: Rough idea:
- 0 or more alphanumeric (plus `-`, `_`, ` `, `>`, `.`, `#`
- optional `[ ... ]` (alphanumeric plus chars)
- 0 or more alphanumeric (plus `-`, `_`, ` `, `>`, `.`, `#`

components/typography/styles.js
```diff
@@ -2,7 +2,7 @@ import '../colors/colors.js';
 import { css, unsafeCSS } from 'lit';
 
 export const _isValidCssSelector = (selector) => {
-	const re = /[#.]?([a-zA-Z0-9-_]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_]+)?/g;
+	const re = /([a-zA-Z0-9-_ >.#]+)(\[[a-zA-Z0-9-_]+\])?([a-zA-Z0-9-_ >.#]+)?/g;
```
jameswklassen, line 5: Rough idea:
- 0 or more alphanumeric (plus `-`, `_`, ` `, `>`, `.`, `#`
- optional `[ ... ]` (inside brackets can be alphanumeric plus `-` and `_`)
- 0 or more alphanumeric (plus `-`, `_`, ` `, `>`, `.`, `#`

components/typography/test/styles.test.js
```diff
@@ -25,12 +25,14 @@ describe.only('_isValidCssSelector', () => {
...
-		expect(_isValidCssSelector('dl #id')).to.be.false;
+	it('should support complex selectors', () => {
+		expect(_isValidCssSelector('dl dd')).to.be.true;
+		expect(_isValidCssSelector('dt.class')).to.be.true;
+		expect(_isValidCssSelector('dt#id')).to.be.true;
+		expect(_isValidCssSelector('dl .class')).to.be.true;
+		expect(_isValidCssSelector('dl #id')).to.be.true;
+		expect(_isValidCssSelector('dl > dt')).to.be.true;
+		expect(_isValidCssSelector('dl > dt > dd')).to.be.true;
 	});
```
dbatiste, line 36: Do you think we should add a test that explicitly tests something like `:host ...` or `:host(...) ...`?

jameswklassen: Hmm yeah, good call.

This does remind me though, that  I didn`t add `:` to my admittedly not-very-smart regex. (But I think that's okay?).

dbatiste: I can't imagine them wanting to target with pseudo-class or this - seems like a weird case that they'd want special typography applied only when a pseudo-class is in play. I'm not sure if it would ever come up, but in theory they might target a pseudo-element, for example `::before` or `::after`.

jameswklassen: Yeah, pseudo-classes + this dynamic css generation seems very niche and something we could tackle if it ever came up.

dbatiste: But ultimately we only _really_ want this used when absolutely needed. Consumers hopefully won't even need these generator functions, so if we want to exclude the `:` cases for now, I think that's fine.

dbatiste (approved):

PR #2837, opened 2022-09-10
allow specifying includeTabbablesOnly in getFirstFocusableDescendant

mdulat:
We were having difficulty doing validation/focus on the htmleditor on save (as opposed to on blur) because it was not showing up as a focusable element, even though it has tabindex=-1. We were checking for the presence of any "focusable" children that are invalid (i.e. the predicate). This fix will allow us to include the htmleditor as an element to check validation and focus on.


helpers/focus.js
```diff
@@ -37,13 +37,13 @@ export function getComposedActiveElement() {
...
+export function getFirstFocusableDescendant(node, includeHidden, predicate, includeTabbablesOnly) {
 	if (predicate === undefined) predicate = () => true;
 
 	const composedChildren = getComposedChildren(node);
 
 	for (let i = 0; i < composedChildren.length; i++) {
-		if (isFocusable(composedChildren[i], includeHidden) && predicate(composedChildren[i])) return composedChildren[i];
+		if (isFocusable(composedChildren[i], includeHidden, includeTabbablesOnly) && predicate(composedChildren[i])) return composedChildren[i];
 
 		const focusable = getFirstFocusableDescendant(composedChildren[i], includeHidden, predicate);
```
dbatiste, line 48: This method is recursive and calls itself, so I think we'd want `includeTabbablesOnly` passed in here as well, yes?

dbatiste:
Looks good. I think it would be good to include a unit test or two for this case.

dbatiste (approved):

dbatiste:
I was a little quick to approve. Were you expecting that visual-diff change?

mdulat:
> I was a little quick to approve. Were you expecting that visual-diff change?

Oh sorry no, I didn't take a close look at that. I reverted it , but not sure what would have caused that.

helpers/test/focus.test.js
```diff
@@ -144,11 +150,18 @@ describe('focus', () => {
 				.to.equal(elem.querySelector('#light1'));
 		});
 
+		it('returns focusable not tabblable child', async() => {
+			const elem = await fixture(focusableNotTabbableFixture);
+			expect(getFirstFocusableDescendant(elem.getContent(), false, true, false))
```
dbatiste, line 155: Looks like this should just be `elem` instead of `elem.getContent()`...? Same with the other test.

PR #2841, opened 2022-09-12
Don't set Prism to manual mode unless we're loading it ourselves

grant-cleary:
Because we always load the Prism loader, we always set Prism to manual mode whether or not we're actually loading Prism. This causes problems in HTML files where Prism is already loaded and our code samples aren't being used.

### DIFF
--- a/helpers/prism.js
+++ b/helpers/prism.js
@@ -1,8 +1,5 @@
 import { css, unsafeCSS } from 'lit';
 
-window.Prism = window.Prism || {};
-Prism.manual = true;
-
 const prismLocation = 'https://s.brightspace.com/lib/prismjs/1.28.0';
 //const prismLocation = '/node_modules/prismjs'; // for local debugging
 
@@ -406,6 +403,12 @@ let prismLoaded;
 
 const loadPrism = () => {
 	if (prismLoaded) return prismLoaded;
+
+	// Set Prism to manual mode before loading to make sure
+	// we don't automatically highlight before we finish
+	// configuring it.
+	window.Prism = window.Prism || {};
+	Prism.manual = true;
 
 	prismLoaded = Promise.all([
 		new Promise(resolve => {


helpers/prism.js
```diff
@@ -407,6 +404,12 @@ let prismLoaded;
 const loadPrism = () => {
 	if (prismLoaded) return prismLoaded;
 
+	// Set Prism to manual mode before loading to make sure
+	// we don't automatically highlight before we finish
+	// configuring it.
+	window.Prism = window.Prism || {};
+	Prism.manual = true;
```
grant-cleary, line 411: I feel like all bets are off if somebody comes in and changes this afterward or loads their own version of Prism. Caveat emptor.

dbatiste: So if a user has loaded their own version of Prism and it initialized first, does their implementation affect our code samples in any way? For instance, if Prism were to have a MutationObserver on the document that was trying to process newly added code elements? In such a scenario, setting `Prism.manual = true` here may be too late.

grant-cleary: If they load their own version of Prism, we're going to get strange behaviour here anyways because we're loading it a second time. I'm not sure how Prism would opt to behave in that case; my guess is that it will depend on whether the versions are compatible or not.

grant-cleary: I think there could be a broader discussion to be had about what we should do if clients opt to load their own versions of MathJax or Prism, but I think there are more complexities at play than this particular property.

dbatiste: Ok. So this change just makes it so we aren't affecting their usage of Prism, in the absence of code samples created from the editor. I think that makes sense, but also we should figure out what to do if we detect Prism on the page that wasn't loaded by us, so we can fail gracefully in such a way the user know's what the issue is.

dbatiste: ^ but I don't think that _has_ to be done in this PR.

grant-cleary: Agreed on both counts. I also want to extend that discussion to MathJax because we do run into the same problem there.

grant-cleary: I do like the idea of just displaying an error message instead of trying to render anything. Certainly that's what MathJax itself does if it detects weirdness.

dbatiste: Having said all that, it's also possible that their version loads after ours, and that could also get nasty.

dbatiste (approved):

helpers/prism.js
grant-cleary: Yeah, realistically version differences mean we can't always get away with that. Nothing stops the user from using a version that just isn't compatible with ours, unfortunately (MathJax 2 vs. 3 comes to mind), or, like you said, deliberately loading theirs later. There's definitely a line we have to draw somewhere in terms of what we're willing/able to support.

PR #2846, opened 2022-09-13
US137419 - Select courses - Add override for selected count 

JeremyWinkler:
https://rally1.rallydev.com/#/255540250444d/iterationstatus?detail=%2Fuserstory%2F630445451251

Related PRs:
- https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/522
- https://github.com/Brightspace/ipsis-sis-course-merge/pull/28

### DIFF
--- a/components/selection/selection-mixin.js
+++ b/components/selection/selection-mixin.js
@@ -48,7 +48,12 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 			 * Whether to render with single selection behaviour. If `selection-single` is specified, the nested `d2l-selection-input` elements will render radios instead of checkboxes, and the selection component will maintain a single selected item.
 			 * @type {boolean}
 			 */
-			selectionSingle: { type: Boolean, attribute: 'selection-single' }
+			selectionSingle: { type: Boolean, attribute: 'selection-single' },
+			/**
+			 * Number of selected items. Optional parameter used to override existing count.
+			 * @type {number}
+			 */
+			selectedCount: { type: Number, attribute: 'selected-count' }
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


dlockhart:
Hi @JeremyWinkler -- thanks for the PR! Would you be able to provide a bit of background on why overriding the built-in selection counts is needed? Thanks!

JeremyWinkler:
> Hi @JeremyWinkler -- thanks for the PR! Would you be able to provide a bit of background on why overriding the built-in selection counts is needed? Thanks!

Sure, so we're using the list along with filters and searches. We want it so that if a user selects an item then applies a filter or search the count says `1 selected` instead of reverting back to `Select All`. We keep track of the selected items in our backend so we know the expected number that should be there. 

I know that team Gaudi has in their backlog a story for persistent selection that would solve this, but it won't be ready in time for our release so we have this in place for now.

zwilson2 (approved):

components/selection/selection-mixin.js
```diff
@@ -48,7 +48,12 @@ export const SelectionMixin = superclass => class extends RtlMixin(superclass) {
 			 * Whether to render with single selection behaviour. If `selection-single` is specified, the nested `d2l-selection-input` elements will render radios instead of checkboxes, and the selection component will maintain a single selected item.
 			 * @type {boolean}
 			 */
-			selectionSingle: { type: Boolean, attribute: 'selection-single' }
+			selectionSingle: { type: Boolean, attribute: 'selection-single' },
+			/**
+			 * Number of selected items. Optional parameter used to override existing count.
+			 * @type {number}
+			 */
+			selectedCount: { type: Number, attribute: 'selected-count' }
```
dlockhart, line 56: Can you rename this to `selectionCountOverride`? Just to be explicit about its purpose.

JeremyWinkler:
Not sure why visual diff failed. Re-ran and it succeeded

dlockhart (approved):
We're going to temporarily merge this to unblock Jeremy until we can come up with a more long-term solution to this issue.

PR #2858, opened 2022-09-21
US143322 - Fix flaky list tests in Firefox

svanherk:
I was not able to reproduce the failure locally, but I was able to get it to fail in Firefox on a Sauce Labs machine. I was seeing that any list test that clicks an input too early could have the issue, so I applied the fix to all those tests.  The below fixed the issue on Sauce Labs, but we'll have to monitor in CI to see if its actually fixed.


dlockhart (approved):
🤞

eKoopmans (approved):

PR #2864, opened 2022-09-21
US143412 - Try to fix flaky list visual-diff test

svanherk:
So I'm not sure if this will totally fix the flake for this visual-diff test. I can't reproduce locally and I was initially able to through ci, and then after I'd fixed it I couldn't get it to break again even after removing the fix 🤷.  But it won't make things worse so let's see... 🤞

### DIFF
--- a/components/list/test/list.visual-diff.js
+++ b/components/list/test/list.visual-diff.js
@@ -62,9 +62,10 @@ describe('d2l-list', () => {
 		}, y);
 	};
 
-	const wait = (milliseconds) => {
-		return page.evaluate((milliseconds) => {
-			return new Promise(resolve => setTimeout(resolve, milliseconds));
+	const wait = (selector, milliseconds) => {
+		return page.$eval(selector, async(elem, milliseconds) => {
+			await elem.updateComplete;
+			await new Promise(resolve => setTimeout(resolve, milliseconds));
 		}, milliseconds);
 	};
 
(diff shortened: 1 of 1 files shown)


components/list/test/list.visual-diff.js
```diff
@@ -62,9 +62,10 @@ describe('d2l-list', () => {
 		}, y);
 	};
 
-	const wait = (milliseconds) => {
-		return page.evaluate((milliseconds) => {
-			return new Promise(resolve => setTimeout(resolve, milliseconds));
+	const wait = (selector, milliseconds) => {
+		return page.$eval(selector, async(elem, milliseconds) => {
+			await elem.updateComplete;
+			await new Promise(resolve => setTimeout(resolve, milliseconds));
```
eKoopmans, line 68: Is it possible the `elem.updateComplete` would be enough, and waiting 100ms isn't necessary anymore?

eKoopmans (approved):

components/list/test/list.visual-diff.js
svanherk: It is definitely possible, but I just can't get it to reproduce consistently enough to check 😞

dlockhart (approved):

components/list/test/list.visual-diff.js
svanherk: Gonna go with this for now to give us our best shot at no flake.  Note to self - if we keep seeing this flake, I think the next step is to dive into the mechanics of the subscription model to see if there's any chance of a rare race condition where a subscription is simply missed, and no amount of waiting would help us.

PR #2867, opened 2022-09-25
Added ability to use sticky headers and horizontal scrolling simultaneously in d2l-table-wrapper

alexvlsem:
[US137261](https://rally1.rallydev.com/#/?detail=/userstory/630015090497&fdp=true): [[A11Y] ] Advanced Evaluation Settings Dialog > Table is not accessible


awikkerink:
I wasn't able to get the demo page to work with RTL set. I used Firefox and Chrome.

In Firefox, the sticky column doesn't work at all. In Chrome, the sticky column is in the wrong place (too far to the right and it scrolls out of view)

eKoopmans:
Hi @alexvlsem , thanks for the PR, and sorry we haven't taken any action with this yet! I've just begun some work on `d2l-table-wrapper` and sticky is part of it, so this is definitely on my radar. We just need to see how this best fits with the other changes we're making.

alexvlsem:
> I wasn't able to get the demo page to work with RTL set. I used Firefox and Chrome.
> 
> In Firefox, the sticky column doesn't work at all. In Chrome, the sticky column is in the wrong place (too far to the right and it scrolls out of view)

I made changes to work with RTL. It works in Firefox. In Chrome it does not work because of [this bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1140374).

eKoopmans:
Hi again @alexvlsem! Thanks again for the PR, and sorry it took a while to follow up. There's some very useful things going on in this PR:

- how you've managed to get the "split overflows but linked scrolling" behaviour (e.g. as described [here](https://uxdesign.cc/position-stuck-96c9f55d9526)) working without breaking up the table itself
- hiding the scroll-wrapper actions when they reach the bottom edge of the container ([Lines 463-469](https://github.com/BrightspaceUI/core/pull/2867/files#diff-dbc548a87db2f9f974ba8e30e0060a0e802e689c2bd2d3bbc4efe446a169f8f7R463-R469))
- the added utility of the `sticky` attribute to make columns sticky (though this may be best as a separate change)

However, there's some things we'd want to improve:

- the main purpose of this PR is to enable a sticky `<thead>` to work in combination with the scroll-wrapper, with all the styling and actions (left/right buttons) that come with the scroll-wrapper
- I can see that you've accomplished this by duplicating a lot of the scroll-wrapper's styles and logic in the table-wrapper itself
- while this works as a great proof-of-concept, it's not how we'd ultimately want this code to be implemented
    - duplication of this code makes it harder to reason with, and also means any future changes in either codebase need to be duplicated in both locations or else we'll get code drift

From my investigation, this seems like a feature we'd want to add into `scroll-wrapper` itself, which then `table-wrapper` could make use of. The core functionality would be to have a scroll-wrapper with multiple "independent" overflow targets, which then have their scrolling synced.

So - our team will pick up that work. I have a bit of a draft of what it would look like already, but not sure whether I can prioritize this immediately. What's your team's timeline for needing this in core?

sarahboyd:
Hey @eKoopmans ! Thanks for taking a look at this - we really appreciate it and that's great that you think the feature could fit in with what you're currently working on. 

For a timeline, we were originally hoping to have it in core by December - do you have an idea on what your own timeline might be for this work? We can certainly adjust things a bit if needed

alexvlsem:
Hi @eKoopmans! Thanks for your feedback! Of course this PR is POC in which I applied the approach from the [mentioned source](https://uxdesign.cc/position-stuck-96c9f55d9526). I will be glad if my efforts help you with table-wrapper. I agree that  scroll-wrapper functionality should be separated and enhanced.

eKoopmans:
Hi again!

> we were originally hoping to have it in core by December

Do you mean merged in December, or as part of the December release? Hopefully not the latter because that window passed on Friday! 😄 

We won't be putting any work into this during Inspiration, so the soonest we'd get started is the following sprint. There's a chunk of other table work in the process, so I'm not sure where this fits in terms of the other priorities - it would be great if your team could coordinate with our dev manager, @nathankoop. Thanks!

eKoopmans:
Hi team! This feature is now supported as part of `core@2.114.0`. Basic usage:

```js
<d2l-table-wrapper sticky-headers sticky-headers-scroll-wrapper>
  <table>...</table>
</d2l-table-wrapper>
```

No need for anything extra - if you want sticky columns, you use the same `sticky` attributes as before. The only catch is that with `sticky-headers-scroll-wrapper`, all header rows must be in `thead`, they can't be in the table body with a `header` attribute or `.d2l-table-header` class.

Let us know if you run into any problems, thanks!

PR #2872, opened 2022-09-26
DE50056: disable skeleton pulse animation in Safari

dlockhart:
DE50056:
https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fdefect%2F647111877577

For an unknown reason, starting with Safari 16 the pulsing skeleton animation causes FACE (assignments, quizzing) and possibly other pages that make use of skeletons to render a blank page.

I tried a few different alternatives, but in the end disabling the animation was the only thing that worked reliably. I'm hoping that Apple fixes this relatively quickly and we can restore things.

This will be cert-fixed to `20.22.10` and possibly hotfixed on `20.22.09`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on: pull_request
 jobs:
   lint:
     name: Lint
-    timeout-minutes: 2
+    timeout-minutes: 3
     runs-on: ubuntu-latest
     steps:
       - uses: Brightspace/third-party-actions@actions/checkout
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -2,6 +2,7 @@
   "extends": "@brightspace-ui/stylelint-config",
 	"ignoreFiles": [
     "components/demo/code-dark-plus-styles.js",
-    "components/demo/code-tomorrow-styles.js"
+    "components/demo/code-tomorrow-styles.js",
+    "components/skeleton/skeleton-mixin.js"
   ]
 }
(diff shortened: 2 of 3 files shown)


dlockhart:
Ugh... need to figure out a way to disable stylelint.

.stylelintrc.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@brightspace-ui/stylelint-config",
 	"ignoreFiles": [
     "components/demo/code-dark-plus-styles.js",
-    "components/demo/code-tomorrow-styles.js"
+    "components/demo/code-tomorrow-styles.js",
+    "components/skeleton/skeleton-mixin.js"
```
dlockhart, line 6: Stylelint hates the inline CSS literals on line 9 of this file. There's [another plugin](https://www.npmjs.com/package/postcss-lit) for Lit specifically that we might have better luck with -- I'll try to look into it separate from this fix.

mdgbayly (approved):
Looks good to me. Glad you had to come up with that Regex and not me!
Tested out the regex against some latest Safari user agent strings for Mac/iPad/iPhone and looks good.
Also tested against latest Chrome to make sure it gets a miss and also looks good!
Thanks for the speedy resolution.

.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ on: pull_request
 jobs:
   lint:
     name: Lint
-    timeout-minutes: 2
+    timeout-minutes: 3
```
dlockhart, line 6: Had 3 timeout failures in a row, so bumping this.

dlockhart:
> Glad you had to come up with that Regex and not me!

Haha I didn't, it was what I considered the best out of several Stack Overflow provided options! :)

> Tested out the regex...

Thanks for the extra testing, appreciate it!

dbatiste (approved):

oelhanafey (approved):

PR #2875, opened 2022-09-27
Taglist accessibility fixes.

dbatiste:
This PR makes some fixes so that tags and tag tooltips are announced as expected. It updates the focus management of tag-list items so that focus is redirected to the internal element that is wired up to tooltips via `aria-describedby`, or the dropdown opener. Further, in order for the tooltips to be properly announced in all browsers, the target **must** have an interactive role. To make the semantics clearer in announcements, `aria-roledescription` is used. Lastly, it cleans up some unneeded code and renames a couple of variables and methods for clarity, given that there are different types of tooltips that can be displayed (keyboard instructions, truncation tooltip, description tooltip).

These change will necessitate a change to the user profile card custom tag-list item (https://github.com/BrightspaceHypermediaComponents/users/pull/83).

**Background:**

The use of tooltips in tag-list was atypical and problematic. Previously, tag-list items focused on the host element, and focus was not redirected  internally to the element referenced by tooltips. It would force the tooltip to become visible by dispatching a fake `focus` event on an internal element while keeping focus on the host. As such, the tooltip would not get announced via `aria-describedby` (referencing the internal element) as one would expect, but rather, as part of the label since the tooltip contents were effectively baked into the focused host element. This is problematic because it would cause the tags and tooltips to not be announced correctly - either partially or not at all, and inconsistent across browser+screen reader combinations. Sometimes only the first element within the tag would be announced, which would happen to be the tooltip, and if the tooltip contained more than one element, sometimes only the first of those elements would be announced. This is also why the keyboard instructions would be announced for the first item, but the first item itself would not be announced. Further, this is also why tag-list items that open dropdowns were not being announced properly, since the opener itself was never being focused.

Aside, debugging was sort of misleading since the tag content is copied and rendered as the first element within the tooltip when a description is present, or when truncating tooltips are enabled. The addition of the `description` property really highlighted some of the issues because it would contain additional content that was not being properly announced.


components/dropdown/dropdown.js
```diff
@@ -22,9 +22,15 @@ class Dropdown extends DropdownOpenerMixin(LitElement) {
 	 */
 	getOpenerElement() {
 		if (!this.shadowRoot) return undefined;
-		return this.shadowRoot.querySelector('slot')
```
dbatiste, line 25: This change is necessary to enable the generic dropdown to find the opener if it's not an immediate child.

components/tag-list/demo/tag-list.html
```diff
@@ -42,31 +42,47 @@ <h2>Basic Tag List</h2>
 			</script>
 			<d2l-demo-snippet fullscreen>
 				<d2l-tag-list description="A bunch of example tags">
-					<d2l-tag-list-item text="Example Tag"></d2l-tag-list-item>
```
dbatiste, line 45: I changed the text in these items to make unit testing the announcements easier, given the role descriptions that are announced (i.e. repetitive "Tag" is somewhat confusing).

components/tag-list/tag-list-item-mixin.js
```diff
@@ -131,52 +131,55 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
 			this._displayKeyboardTooltip = false;
-			/** @ignore */
-			container.dispatchEvent(new FocusEvent('blur', { bubbles: false, cancelable: true }));
 		});
+
 		this.addEventListener('keydown', this._handleKeydown);
 	}
 
-	handleClearItem(e, clearAll) {
+	_handleClearItem() {
```
dbatiste, line 161: The only thing that called `handleClearItem` outside this class was `d2l-tag-list`, which passed `true` for `clearAll`, which resulted in nothing actually being done. Further, when dispatched from the item's clear button, `handleFocus` would always be `true`, but it wasn't actually being used. As a result, this method could be made private and cleaned up.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -131,52 +131,55 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
 
 		/** Dispatched when a user selects to delete an individual tag list item. The consumer must handle the actual element deletion and focus behaviour if there are no remaining list items. */
 		this.dispatchEvent(new CustomEvent(
 			'd2l-tag-list-item-clear',
-			{ bubbles: true, composed: true, detail: { value: this.text, handleFocus } }
+			{ bubbles: true, composed: true, detail: { value: this.text } }
+		));
+	}
+
+	_handleKeyboardTooltipHide() {
```
dbatiste, line 173: These are just renamed so it's clear that they are just for the keyboard instructions tooltip.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -187,55 +190,72 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
 		const buttonText = typeof tagContent === 'object'
 			? this.localize('components.tag-list.clear', { value: '' })
 			: this.localize('components.tag-list.clear', { value: tagContent });
-		const hasDescription = !!description;
-		const tooltipHeader = hasDescription ? html`<div class="d2l-heading-4">${tagContent}</div>` : tagContent;
-		const tooltipTagOverflow = hasTruncationTooltip ? html`
+
+		const hasDescription = !!options.description;
+
+		let tooltip = nothing;
```
dbatiste, line 215: I refactored this slightly so that only one template result is created for whichever tooltip is going to be rendered, instead of both.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -187,55 +190,72 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
 
-	_handleTooltipShow() {
-		this._tooltipShown = true;
-		/** @ignore */
-		this.dispatchEvent(new CustomEvent(
-			'd2l-tag-list-item-tooltip-show',
-			{ bubbles: true, composed: true }
-		));
-	}
+	_renderTag(tagContent, options) {
```
dbatiste, line 206: This change to `_renderTag(...)` will require change to the user's tag-list-item to pass in the appropriate options.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -187,55 +190,72 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
+			'tag-list-item-content': true
+		};
+		if (options.focusableClass) focusableClasses[options.focusableClass] = true;
+
 		return html`
-			${tooltipKeyboardInstructions || tooltipTagOverflow}
+			${tooltip}
 			<div class="${classMap(containerClasses)}">
-				<div class="tag-list-item-content" id="${this._id}" tabindex="-1">${tagContent}</div>
+				<div aria-label="${ifDefined(options.label)}"
```
dbatiste, line 247: This element will get the focus. In order for the tooltips to be announced correctly, it must have an interactive role. To make the semantics clearer, `aria-roledescription` is used.

The addition of `aria-label` option is required by the user's tag-list-item, since it provides some special instruction for screen reader users.

components/tag-list/tag-list.js
```diff
@@ -193,18 +192,17 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 			<div
 				aria-label="${this.description}"
 				class="${classMap(containerClasses)}"
-				role="list"
+				role="group"
```
dbatiste, line 195: Since we are moving focus internally, we lose the list-specific announcements that were announced with NVDA. To make the semantics clearer, we can use the `group` role and provide a custom description for the role. Should we choose to, we could append counts to the description, but these they were not announced consistently across browsers+screen readers previously anyway.

components/tag-list/tag-list.js
```diff
@@ -347,15 +341,9 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		));
 
 		announce(this.localize('components.tag-list.cleared-all'));
-
```
dbatiste, line 350: This is no longer needed, since as mentioned previously, it wasn't doing anything.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -76,12 +76,12 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 				padding: 0.25rem 0.6rem;
 				text-overflow: ellipsis;
 			}
-			:host(:focus-visible) .tag-list-item-container,
-			:host(:focus-visible:hover) .tag-list-item-container {
+			:host(:focus-within) .tag-list-item-container,
+			:host(:focus-within:hover) .tag-list-item-container {
```
dlockhart, line 80: Seems safe enough to use browser-compatibility-wise. 👍

components/tag-list/tag-list-item-mixin.js
```diff
@@ -131,52 +131,55 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		this.clearable = false;
 		/** @ignore */
 		this.keyboardTooltipItem = false;
-		/** @ignore */
-		this.role = 'listitem';
```
dlockhart, line 135: Can the property definition for `role` go away?

components/tag-list/tag-list-item-mixin.js
dlockhart: This `_displayKeyboardTooltip` could probably be converted to a state property.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -187,55 +190,72 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		const clearKeys = e.keyCode === keyCodes.BACKSPACE || e.keyCode === keyCodes.DELETE;
 		if (!this.clearable || !clearKeys) return;
 		e.preventDefault();
-		this.handleClearItem(e);
+		this._handleClearItem(e);
```
dlockhart, line 193: I don't think `e` needs to be passed anymore.

components/tag-list/tag-list-item-mixin.js
dlockhart: Cool, hadn't seen `aria-roledescription` before!

components/tag-list/tag-list-item-mixin.js
dbatiste: Yep, thanks!

dlockhart (approved):

components/tag-list/tag-list.js
```diff
@@ -313,10 +311,6 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 			if (node.nodeType !== Node.ELEMENT_NODE) return false;
 			await node.updateComplete;
 
-			const role = node.getAttribute('role');
-			const backdropRole = node.getAttribute(BACKDROP_ROLE);
-			if (role !== 'listitem' && backdropRole !== 'listitem') return false;
```
svanherk, line 318: Now that we're no longer checking the role, are we getting any additional nodes?  Or was this just being cautious to begin with?

components/tag-list/tag-list.js
```diff
@@ -2,7 +2,6 @@ import '../button/button-subtle.js';
 import { css, html, LitElement } from 'lit';
 import { announce } from '../../helpers/announce.js';
 import { ArrowKeysMixin } from '../../mixins/arrow-keys-mixin.js';
-import { BACKDROP_ROLE } from '../backdrop/backdrop.js';
```
svanherk, line 5: Minor: This class was exported specifically for the fix here - if we no longer need it, we could technically make it a private constant again.  Although it may be needed in the future somewhere else 🤷

PR #2877, opened 2022-09-28
add getFirstDescendant function with predicate

mdulat:
A previous bug fix broke accessibility of the html editor in FACE pages. This addition allows us to find which child node has the invalid attribute (i.e. our predicate) so that we can handle the error, without changing the editor's tabindex.

https://rally1.rallydev.com/#/?detail=/defect/650517971645&fdp=true
Goes with https://github.com/BrightspaceHypermediaComponents/activities/pull/2978

### DIFF
--- a/helpers/dom.js
+++ b/helpers/dom.js
@@ -159,6 +159,21 @@ export function getPreviousAncestorSibling(node, predicate = () => true) {
 	return null;
 }
 
+export function getFirstDescendant(node, predicate) {
+	if (predicate === undefined) predicate = () => true;
+
+	const composedChildren = getComposedChildren(node);
+
+	for (let i = 0; i < composedChildren.length; i++) {
+		if (predicate(composedChildren[i])) return composedChildren[i];
+
+		const found = getFirstDescendant(composedChildren[i], predicate);
+		if (found) return found;
+	}
+
+	return null;
+}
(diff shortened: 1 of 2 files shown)


helpers/dom.js
```diff
@@ -159,6 +159,21 @@ export function getPreviousAncestorSibling(node, predicate = () => true) {
 	return null;
 }
 
+export function getFirstDescendant(node, predicate) {
```
dbatiste, line 162: Could `querySelectorComposed(...)` (defined below) serve your purpose?

If not, I propose we name this `getComposedFirstDescendant` to be consistent with the other helper methods that walk the composed DOM. The two immediately above are not composed, but ones that are, are named accordingly.

Also, I might suggest we default the predicate as is done in a couple others above (i.e. `(node, predicate = () => true)`), just for consistency.

I think we need be very careful with these methods that recursively walk deeper into the sub-tree, which includes the use of `querySelectorComposed`. Walking up the tree too, but I am slightly less concerned. They could be detrimental to performance if used in certain scenarios. I am somewhat worried that consumers will not consider this.

mdulat: I can make it work with `querySelectorComposed`, thanks Dave. I'll close this PR.

PR #2883, opened 2022-09-30
fix: Add aria-hidden to object-property-list icons

eKoopmans:
Small fix to add aria-hidden to icons.

Relevant Slack thread: https://d2l.slack.com/archives/C03FB9DKVLL/p1661876160090209?thread_ts=1661875106.547899&cid=C03FB9DKVLL


dlockhart (commented):
I just read the Slack thread... I'm surprised this is necessary at all I guess? Like `<d2l-icon>` to a screen reader is meaningless, so I would have expected it to just skip over it. We don't put `aria-hidden` on any other uses of `<d2l-icon>` that I'm aware of.

PR #2886, opened 2022-10-03
US140730 add check-user icon

BhamelinD2L:
[US140730](https://rally1.rallydev.com/#/?detail=/userstory/641719318519&fdp=true): Org-Level Updates Widget Notifications for Approvers [Note that the example icons in the story don't look like this, but the designer eventually settled on this one]

Adds a new `check-user` icon for use in Discover's pending request widget as well as Discover's Manage requests button.

I manually changed the URL in the catalogue and it should look like this:
image

You can also view it here: https://raw.githubusercontent.com/BrightspaceUI/core/US140730-add-icon/components/icons/images/tier1/check-user.svg


dlockhart (approved):
I pushed some changes to the SVG -- previously color overrides wouldn't get applied due to how the `fill` was set on the various elements.

PR #2896, opened 2022-10-06
fix: (DE50257) disable skeleton pulse animation in Safari

dlockhart:
Hotfix version of #2872, which is now being requested.

Rally:
https://rally1.rallydev.com/#/15545167705d/custom/21568985922


svanherk (approved):
You might get some visual diff changes, there's been a puppeteer bump since

PR #2904, opened 2022-10-12
changes made for filters to now display result counts

ksanctis:
US139257 - Filter > Result Counts

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -16,10 +16,11 @@ import '../selection/selection-select-all.js';
 import '../selection/selection-summary.js';
 
 import { bodyCompactStyles, bodySmallStyles, bodyStandardStyles } from '../typography/styles.js';
-import { css, html, LitElement } from 'lit';
+import { css, html, LitElement, nothing } from 'lit';
 import { announce } from '../../helpers/announce.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { FocusMixin } from '../../mixins/focus-mixin.js';
+import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
@@ -120,12 +121,15 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				white-space: nowrap;
 			}
 
-			.d2l-filter-dimension-set-value-text {
+			.d2l-filter-dimension-set-value {
 				color: var(--d2l-color-ferrite);
+				display: flex;
+				gap: 0.45rem;
 				line-height: unset;
+				overflow: hidden;
 			}
 
-			d2l-list-item[disabled] .d2l-filter-dimension-set-value-text {
+			d2l-list-item[disabled] .d2l-filter-dimension-set-value {
 				color: var(--d2l-color-chromite);
 			}
 
@@ -415,9 +419,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 						?hidden="${item.hidden}"
 						key="${item.key}"
 						label="${item.text}"
+						number="${item.count}"
 						selectable
 						?selected="${item.selected}">
-						<div class="d2l-filter-dimension-set-value-text d2l-body-compact">${item.text}</div>
+						<div class="d2l-filter-dimension-set-value d2l-body-compact">
+							<div class="d2l-filter-dimension-set-value-text">${item.text}</div>
+							${item.count !== undefined ? html`<div class="d2l-body-small">(${formatNumber(item.count)})</div>` : nothing}
+						</div>
 					</d2l-list-item>
 				`)}
 			</d2l-list>
--- a/components/filter/test/filter-dimension-set-value.test.js
+++ b/components/filter/test/filter-dimension-set-value.test.js
@@ -4,13 +4,55 @@ import { runConstructor } from '../../../tools/constructor-test-helper.js';
 import { spy } from 'sinon';
 
 const valuefixture = html`
-	<d2l-filter-dimension-set-value key="value" text="Value"></d2l-filter-dimension-set-value>
+	<d2l-filter-dimension-set-value key="value" text="Value" count="0"></d2l-filter-dimension-set-value>
 `;
 
 describe('d2l-filter-dimension-set-value', () => {
 
 	it('should construct', () => {
 		runConstructor('d2l-filter-dimension-set-value');
+	});
+
+	describe('data change', () => {
+		it('fires data change event when its count data changes', async() => {
+			const elem = await fixture(valuefixture);
+			const eventSpy = spy(elem, 'dispatchEvent');
+			elem.count = 1;
+
+			const e = await oneEvent(elem, 'd2l-filter-dimension-set-value-data-change');
+			expect(e.detail.valueKey).to.equal('value');
+			expect(e.detail.changes.size).to.equal(1);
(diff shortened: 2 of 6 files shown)


components/filter/filter.js
```diff
@@ -415,9 +419,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 						?hidden="${item.hidden}"
 						key="${item.key}"
 						label="${item.text}"
+						number="${item.count}"
```
dbatiste, line 422: This looks a little odd. Are you sure you meant to include this?

components/filter/filter.js
```diff
@@ -120,12 +121,15 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				white-space: nowrap;
 			}
 
-			.d2l-filter-dimension-set-value-text {
+			.d2l-filter-dimension-set-value {
 				color: var(--d2l-color-ferrite);
```
svanherk, line 125: What colour are the counts supposed to be?  The design looks like they're supposed to be lighter than `ferrite`, but you may have a newer copy than the one linked in the story.  The disabled case also looks a bit odd, wondering if we should at least make it lighter in that case

dbatiste:
We should update the [readme](https://github.com/BrightspaceUI/core/tree/main/components/filter#filter-dimension-set-value-d2l-filter-dimension-set-value) to include the property along with the others until we decide to get rid of the duplication. Also, we should add some usage of `count` to the readme demo as well - it gets used for the [Daylight docs](https://daylight.d2l.dev/components/filters/). Jeff might have some preference as to where/which demo he wants to see the counts.

components/filter/filter.js
dbatiste: The `d2l-body-small` styles includes setting color to `tungsten`.

svanherk: Ah gotcha!  But yeah I wonder if when disabled, we match the text

dbatiste: @ksanctis , can you follow up with Jeff to find out what he wants? I suspect we want to match the disabled color when it's disabled as Stacey mentioned.

components/filter/filter.js
ksanctis: It isn't meant to be there, have taken it off.

components/filter/filter.js
ksanctis: I will confirm with Jeff and make changes accordingly.

components/filter/filter.js
```diff
@@ -415,9 +419,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 						?hidden="${item.hidden}"
 						key="${item.key}"
 						label="${item.text}"
+						number="${item.count}"
 						selectable
 						?selected="${item.selected}">
-						<div class="d2l-filter-dimension-set-value-text d2l-body-compact">${item.text}</div>
+						<div class="d2l-filter-dimension-set-value d2l-body-compact">
+							<div class="d2l-filter-dimension-set-value-text">${item.text}</div>
+							${item.count !== undefined ? html`<div class="d2l-body-small">(${formatNumber(item.count)})</div>` : nothing}
```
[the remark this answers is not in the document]
ksanctis: I have tried testing this out using the demo pages and when I pass a string into "count" the filter dropdown just breaks and does not render at all even if I have the `Number.isNaN(item.count)` check.

components/filter/filter.js
ksanctis: Got confirmation from Jeff, will make the colour for the count the same as the text when the filter is disabled.

eKoopmans:
Ah another detail about the PR itself - the title should start with `feat: ` or `fix: `, which will tell semantic-release whether to bump the minor or patch version when you merge. In this case since you're adding a new feature, it should be `feat:`.

Technically it doesn't need to be in the PR title, just the commit name when you squash-merge, but having it in the PR title is good practice and will auto-fill the squash-merge message.

components/filter/filter.js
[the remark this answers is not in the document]
dbatiste: Just updating here with what I already mentioned in Slack. I think this can be done a couple different ways.

My preference is to validate closer to the assignment on `d2l-filter-dimension-set-value` with something like...
```javascript
get count() {
	return this._count;
}

set count(val) {
	if (Number.isNaN(val)) val = undefined;
	const oldVal = this._count;
	if (oldVal !== val) {
		this._count = val;
		this.requestUpdate('count', oldVal);
	}
}
```

But validating on usage in `d2l-filter` can in fact work as well:
```javascript
${(item.count !== undefined && !Number.isNaN(item.count)) ? html`<div class="d2l-body-small">(${formatNumber(item.count)})</div>` : nothing}
```

I've tested both these approaches. It's a worthwhile check. I would not expect consumers to actually set a string value purposefully, but if the consumer is provided an invalid count, we'd like the UI component to not produce JavaScript errors or the invalid value.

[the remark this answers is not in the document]
ksanctis: Made the change for an additional check on string inputs for count.

components/filter/README.md
```diff
@@ -220,12 +221,43 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
 
 | Property | Type | Description |
 |---|---|---|
+| `count` | Number, default: `undefined` | Count value for the value in the list |
```
dbatiste, line 224: For this one here, I would exclude the default, and in the description I would say "Count for the value in the list. If no count is provided, no count will be displayed." I would update the comment in the code to be this as well, since that's what gets used when the docs are generated.

ksanctis: Changes made.

dbatiste (approved):

components/filter/test/filter-dimension-set-value.test.js
```diff
@@ -14,16 +14,47 @@ describe('d2l-filter-dimension-set-value', () => {
...
+			{ property: 'text', value: 'Test' }
+		].forEach(condition => {
+			it(`fires data change event when its ${condition.property} data changes`, async() => {
+				const elem = await fixture(valuefixture);
+				const eventSpy = spy(elem, 'dispatchEvent');
+
+				if (condition.property === 'count') elem.count = condition.value;
+				else if (condition.property === 'disabled') elem.disabled = condition.value;
+				else if (condition.property === 'selected') elem.selected = condition.value;
+				else if (condition.property === 'text') elem.text = condition.value;
```
eKoopmans, line 30: These can be simplified as `elem[condition.property] = condition.value`.

eKoopmans (approved):
Reminder to update the PR title to start with `feat: `.

components/filter/test/filter-dimension-set-value.test.js
ksanctis: Made the change.

components/filter/README.md
```diff
@@ -220,12 +221,43 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
...
 | `selected` | Boolean, default: `false` | Whether the value in the filter is selected or not |
 <!-- docs: end hidden content -->
 
+## Counts [d2l-filter-dimension-set-value]
+
+The `count` property displays a count next to each filter value to indicate the number of results a value will yield. This helps users more effectively explore data and make selections, so it’s a good idea to provide these counts if it can be done performantly.
+
+Note that when using multiple filter dimensions, the counts should be updated when selections are made across dimensions so that they always reflect the number of results a filter will yield.
+
+<!-- docs: demo live align:start autoOpen:true autoSize:false size:large -->
```
oelhanafey, line 237: You probably want to be using `demo code` here instead of `demo live`. `demo live` displays a web component's properties, slots and events in addition to the code and demo. Though it may be displaying correctly at the moment since it was not given a web component `name`. [See here](http://localhost:8000/about/adding-component/#readme%2Fmarkdown-document)

PR #2910, opened 2022-10-13
fix: wait for "invalid" to be updated before firing change event

dlockhart:
There's some logic in the LMS that needs to listen for a number input's `change` event, and then perform different actions based on whether the value the user entered is valid or not. And by invalid, I mean if a min/max is set and the user enters a number outside that range.

The issue is that the input's `invalid` property gets updated in the `updated()` lifecycle callback, which was happening _after_ the event was fired. This change just waits for that to run before firing the event.

### DIFF
--- a/components/inputs/test/input-number.test.js
+++ b/components/inputs/test/input-number.test.js
@@ -387,6 +387,14 @@ describe('d2l-input-number', () => {
 			await aTimeout(1);
 
 			expect(fired).to.be.false;
+
+		});
+
+		it('should have correct "invalid" state when change event fires', async() => {
+			const elem = await fixture(minMaxFixture);
+			setTimeout(() => setInnerInputValue(elem, '0'));
+			await oneEvent(elem, 'change');
+			expect(elem.invalid).to.be.true;
 		});
 	});
 
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -438,6 +438,7 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 		}
 
 		if (dispatchEvent) {
+			await this.updateComplete; // wait for validity logic to re-run
 			this.dispatchEvent(new CustomEvent(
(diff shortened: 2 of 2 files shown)


components/inputs/test/input-number.test.js
```diff
@@ -387,6 +387,12 @@ describe('d2l-input-number', () => {
 			await aTimeout(1);
 
 			expect(fired).to.be.false;
+
```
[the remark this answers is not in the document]
eKoopmans: Rather than commenting out tests, I think you can use `it.only(...)` (Jest uses `fit(...)`, but I noticed that doesn't work for our setup).

eKoopmans (dismissed):

components/inputs/test/input-number.test.js
```diff
@@ -387,6 +387,14 @@ describe('d2l-input-number', () => {
...
+		});
+
+		it('should have correct "invalid" state when change event fires', async() => {
+			const elem = await fixture(minMaxFixture);
+			setTimeout(() => setInnerInputValue(elem, '0'));
+			await oneEvent(elem, 'change');
```
eKoopmans, line 396: I'm just noticing that `setInnerInputValue` itself fires a change event on the element:

```js
setTimeout(() => dispatchEvent(inputTextElement, 'change'));
```

Is this line waiting for that event, or for the one fired in input-number's `_handleChange()`? Seems like there could be a race condition there.

dlockhart: It's a different event. We're simulating the user modifying the underlying `<d2l-input-text>`'s value (which in turn sets an underlying `<input type="text">`) and firing a change on that `<d2l-input-text>`. That's what [the number input is listening for](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-number.js#L326) and will trigger the logic in `_handleChange` to run, which in turn fires a `change` event on the actual `d2l-input-number`, which is what the test is waiting for here.

PR #2912, opened 2022-10-13
fix: using an outline when table column sort button is focused

dlockhart:
This was just using an underline previously, which doesn't pass the WCAG minimum requirements for focus styles. It's now using `:focus-visible` to apply our standard ring, but with `4px` radius instead of `6px`.

Screen Shot 2022-10-13 at 5 06 03 PM

Expecting some visual-diff failures as there are tests specifically for the focus state of the sort buttons.


components/table/table-col-sort-button.js
```diff
@@ -51,11 +52,14 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 			button:disabled {
 				opacity: 0.5;
 			}
-			button:hover,
-			button:focus {
-				outline-style: none;
+			button:hover {
 				text-decoration: underline;
 			}
+			button:focus-visible {
```
dlockhart, line 58: I'm not using the mixin on purpose, which means Safari <15.4 won't work. We're OK with that at this point.

dbatiste (approved):
Looks good!

PR #2915, opened 2022-10-14
Spike - list-header bulk-action mode

dbatiste:
This spike simply shows how bulk-action mode could be implemented for the list-header, where the header is only displayed when one or more items are selected. The exact requirements are uncertain for this story, and some questions are outlined in [US144294](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F660025810659&fdp=true).

### DIFF
--- a/components/list/list-header.js
+++ b/components/list/list-header.js
@@ -7,12 +7,15 @@ import { classMap } from 'lit/directives/class-map.js';
 import { findComposedAncestor } from '../../helpers/dom.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { SelectionObserverMixin } from '../selection/selection-observer-mixin.js';
+
+const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 
 /**
  * A header for list components containing select-all, etc.
  * @slot - Responsive container using `d2l-overflow-group` for `d2l-selection-action` elements
  */
-class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
+class ListHeader extends SelectionObserverMixin(RtlMixin(LocalizeCoreElement(LitElement))) {
 
 	static get properties() {
 		return {
@@ -27,12 +30,18 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			 */
 			noSticky: { type: Boolean, attribute: 'no-sticky' },
 			/**
+			 * Whether selection is required to show the bulk-action header
+			 * @type {boolean}
+			 */
+			requiresSelection: { type: Boolean, attribute: 'requires-selection', reflect: true },
+			/**
 			 * Whether all pages can be selected
 			 * @type {boolean}
 			 */
 			selectAllPagesAllowed: { type: Boolean, attribute: 'select-all-pages-allowed' },
 			_hasActions: { state: true },
-			_scrolled: { type: Boolean, reflect: true }
+			_scrolled: { type: Boolean, reflect: true },
+			_state: { state: true }
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


components/list/list-header.js
```diff
@@ -66,14 +75,32 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
 			:host([hidden]) {
 				display: none;
 			}
-			.d2l-list-header-container {
+			:host([requires-selection]) .d2l-list-header-container {
```
dbatiste, line 78: There are no defined requirements (yet) for the animation. This spike simply animates the `opacity` and `max-height` as a proof of concept. Animating other properties may be required depending on the required animation - we may even need to measure and animate `height`, which could complicate things slightly. Further, hiding `overflow` may be required, which also complicates things because we would not want the `overflow` hidden in the visible end-state. It would just mean having another end-state class that is applied after the animation completes. We can cross that bridge when the time comes.

components/list/list-header.js
```diff
@@ -66,14 +75,32 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+				display: none;
+				max-height: 0;
+				opacity: 0;
+				/* transform: translate(0, -10px); */
+				transition: opacity 100ms linear, max-height 100ms linear;
+			}
+			:host([requires-selection]) .d2l-list-header-container-pre-showing,
```
dbatiste, line 85: We need the header to be visible in the DOM before animation begins (i.e. we cannot animate starting from `display: none;`.

components/list/list-header.js
```diff
@@ -66,14 +75,32 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+			:host([requires-selection]) .d2l-list-header-container-hiding {
+				display: block;
+			}
+			:host([requires-selection]) .d2l-list-header-container-showing {
+				display: block;
+				max-height: 70px;
```
dbatiste, line 91: This assumes a `max-height`, which is probably not a safe assumption since consumers can potentially put all sorts of things into the header. This assumption is probably worth discussing with Design, but ultimately, a height measurement may be more appropriate if we are required to animate the height aspect.

components/list/list-header.js
```diff
@@ -129,28 +163,88 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+			'd2l-list-header-container-showing': (this.requiresSelection && this._state === 'showing'),
+			'd2l-list-header-container-hiding': (this._state === 'hiding')
+		};
+		const classesInner = {
+			'd2l-list-header-container-inner': true,
+			'd2l-list-header-slim': (!this._hasActions && !this.selectAllPagesAllowed),
 			'd2l-list-header-extend-separator': this._extendSeparator
 		};
 		return html`
 			<div class="${classMap(classes)}">
```
dbatiste, line 176: We need to move the element with `min-height` into an inner element, so that it doesn't mess up height animation. Note: I think the `min-height` was being applied for the case where there are no actions in the header... just select-all. i.e. in order to ensure we have a consistent height regardless of the header contents.

components/list/list-header.js
```diff
@@ -129,28 +163,88 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (!this.requiresSelection || !changedProperties.has('selectionInfo')) return;
+
+		const hasSelection = (this.selectionInfo.state !== 'none');
+
+		if (hasSelection) {
```
dbatiste, line 202: This code manages the animation state. It tries to handle cases where the state is changed midway through animation (ex. quickly toggling the selection). It bumps the animation by a couple frames so that the header is visible in the DOM before animation begins. It handles the reduced motion case as well.

I'm not sure why two frame are being used, since one frame is what we'd normally expect. This code what copied and modified from alert-toast, so presumably it is needed.

Ignore the commented code for now. It more applies if we decided to use animations instead of transitions.

components/list/list-header.js
eKoopmans: What are you picturing for a height measurement? I'll probably need to measure this component's height to get sticky table headers working too.

components/list/list-header.js
```diff
@@ -129,28 +157,88 @@ class ListHeader extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
+			if (this._state === 'hiding') {
+				this._state = 'showing';
```
eKoopmans, line 198: Do we not need to cancel the 'transitionend' listener in this case?

components/list/list-header.js
dbatiste: In this case, if we needed the height of the contents, in `updated()` I would probably try to use `getBoundingClientRect()` on the inner `div` after `display: block` is applied to the outer `div` but before beginning the animation. 

While the outer `div` would have a `height` of 0 (because `max-height: 0`), the inner `div` would be able to provide the height of the contents. I would then assign it to a  `_contentHeight` property, which could be used in `render()` to specify either `height` or `max-height` style property on the outer `div`, and the `transition` would kick in at that point.

If we measure before beginning each `showing` animation, then we don't need to worry if the consumer adds/removes something from the header that affects the height. We'd never expect the contents of the header's slot to be changed during the animation. The explicit height assignment would be removed after the animation is complete.

So those are my initial thoughts.

components/list/list-header.js
dbatiste: I tried to break it in testing, but I was unsuccessful. But yeah, probably to be safe, because it does seem like there is an edge case where we could see some very buggy behaviour if the state was toggled at the exact "wrong" time.

dbatiste:
Closing this PR since it was for PoC only. We'll have the branch if we decide to pursue this further in the future.

PR #2920, opened 2022-10-18
Add support for localizing based on browser settings

bearfriend:
The `awards-public-ui` repo renders a component on a public-facing page that should be localized based on the viewing-user's preferences, not the creator's account settings or the default tenant settings.

- Allows the `localize-dynamic-mixin` to get its `possibleLanguages` from the browser with `useBrowserLangs`
- Moves `documentLocaleSettings` to a static property
- Refactors `generatePossibleLanguages`, moves it to a static method, and calls it only where needed
- Replaces `hasOwnProperty` calls
- Adds `useBrowserLangs` docs

[DE50422](https://rally1.rallydev.com/#/?detail=/defect/662854795259&fdp=true): [awards-public-ui] Spanish is not getting translated properly

### DIFF
--- a/mixins/test/localize-mixin.test.js
+++ b/mixins/test/localize-mixin.test.js
@@ -321,13 +321,13 @@ describe('LocalizeMixin', () => {
 			localizedTerm: 'Labour Day',
 			it: 'should localize text based on browser settings'
 		},
-	  {
+		{
 			browserLangs: ['fr-CA', 'fr-BE', 'fr-FR', 'fr'],
 			resolvedLang: 'fr',
 			localizedTerm: 'Fête du travail',
 			it: 'should loop until an alternative is found'
 		},
-	  {
+		{
 			browserLangs: ['fr-CA', 'fr-BE', 'fr-FR'],
 			resolvedLang: 'en',
 			localizedTerm: 'Labor Day',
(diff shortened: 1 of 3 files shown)


dlockhart (commented):
Very cool! Could you add some test coverage for this new feature in the existing unit tests for the mixin?

mixins/test/localize-mixin.test.js
```diff
@@ -269,6 +300,53 @@ describe('LocalizeMixin', () => {
 
 	});
 
+	describe('browser language settings in dynamic mixin', () => {
+
+		const browserLangsFixture = `<${browserLangsTag}></${browserLangsTag}>`;
```
bearfriend, line 305: These are sort of oddball tests, but `navigator` isn't able to be stubbed which causes some timing weirdness.

mixins/test/localize-mixin.test.js
```diff
@@ -104,14 +104,45 @@ const test2LocalizeDynamicMixn = superclass => class extends LocalizeDynamicMixi
 	}
 };
 
+const Test3LocalizeDynamicMixn = superclass => class extends LocalizeDynamicMixin(superclass) {
+
+	static translations = {
+		'en': { laborDay: 'Labor Day' },
+		'en-ca': { laborDay: 'Labour Day' },
+		'fr': { laborDay: 'Fête du travail' }
```
dlockhart, line 112: Work party! 🎉

dlockhart (approved):

PR #2924, opened 2022-10-19
proof of concept of how to track heading levels

dlockhart:
This is a really rough draft to demonstrate an idea to solve the problem of arbitrary components that need to render a heading, but don't know what semantic level (`h1`, `h2`, etc.) to use because they don't have the necessary context.

Similar to the MVC framework, this introduces a `<heading-stack>` component. You'd use this whenever you render a heading and have some nested views who might need to render a heading at a higher level. I've consumed this inside our `<demo-page>` component since it does this.

In addition, there's a `HeadingLevelController`. Any place that needs to render a heading but doesn't know what level to use would have one of these. I've created a test component in the demo page that does this.

The code to actually render out the heading here is pretty gnarly. It uses Lit's new [static expressions](https://lit.dev/docs/templates/expressions/#static-expressions) stuff, but because those values need to be static I had to resort to a `switch` statement, which I don't love. We could instead use [non-literal static](https://lit.dev/docs/templates/expressions/#non-literal-statics) stuff since we "trust" this value, but yeah.. kind of gross.

Another alternative would be to introduce actual `<d2l-heading>` controls that do all this for the consumer. I worry though that it would involve giving up too much control to style/position the actual heading elements?

### DIFF
--- a/components/dialog/dialog.js
+++ b/components/dialog/dialog.js
@@ -1,4 +1,5 @@
 import '../button/button-icon.js';
+import '../heading/heading-stack.js';
 import '../loading-spinner/loading-spinner.js';
 import { AsyncContainerMixin, asyncStates } from '../../mixins/async-container/async-container-mixin.js';
 import { css, html, LitElement } from 'lit';
@@ -149,7 +150,7 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 						<d2l-button-icon icon="tier1:close-small" text="${this.localize('components.dialog.close')}" @click="${this._abort}"></d2l-button-icon>
 					</div>
 				</div>
-				<div class="d2l-dialog-content" @pending-state="${this._handleAsyncItemState}">${content}</div>
+				<d2l-heading-stack class="d2l-dialog-content" override-level="3" @pending-state="${this._handleAsyncItemState}" style="display: block;">${content}</d2l-heading-stack>
 				<div class="${classMap(footerClasses)}">
 					<slot name="footer" @slotchange="${this._handleFooterSlotChange}"></slot>
 				</div>
--- a/components/heading/demo/heading-stack-test.js
+++ b/components/heading/demo/heading-stack-test.js
@@ -1,0 +1,41 @@
+import { html, literal } from 'lit/static-html.js';
+import { HeadingLevelController } from '../heading-stack.js';
+import { LitElement } from 'lit';
+
+class TestHeadingStack extends LitElement {
+
+	constructor() {
+		super();
+		this._headingLevelController = new HeadingLevelController(this);
+	}
+
+	render() {
+		let headingLevel;
+		switch (this._headingLevelController.level) {
+			case 1:
+				headingLevel = literal`h1`;
+				break;
+			case 2:
+				headingLevel = literal`h2`;
+				break;
+			case 3:
+				headingLevel = literal`h3`;
+				break;
+			case 4:
+				headingLevel = literal`h4`;
+				break;
+			case 5:
+				headingLevel = literal`h5`;
+				break;
+			case 6:
+				headingLevel = literal`h6`;
+				break;
+		}
+		return html`
+			<${headingLevel}>heading</${headingLevel}>
+			<slot></slot>
+		`;
+	}
+
+}
+customElements.define('d2l-test-heading-stack', TestHeadingStack);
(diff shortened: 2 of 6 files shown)


components/heading/demo/heading-stack-test.js
```diff
@@ -0,0 +1,41 @@
...
+			case 4:
+				headingLevel = literal`h4`;
+				break;
+			case 5:
+				headingLevel = literal`h5`;
+				break;
+			case 6:
+				headingLevel = literal`h6`;
+				break;
+		}
```
dlockhart, line 33: Alternate way to do this:

```javascript
const headingLevel = `h${this._headingLevelController.level}`;
return html`
			<${unsafeStatic(headingLevel)}>heading</${unsafeStatic(headingLevel)}>
			<slot></slot>
`;
```

dbatiste: Interesting, so it doesn't let you do the following eh?
```javascript
 literal`h${level}`
```
These aren't a huge deal if they are within our components. The switch statement could be factored away into a helper function that could be shared. Nevertheless, I agree about, and I'd rather consumers not have to be concerns with any of this gnarling. 😬

components/heading/heading-stack.js
```diff
@@ -0,0 +1,41 @@
+import { html, LitElement } from 'lit';
+import { ProviderMixin, requestInstance } from '../../mixins/provider-mixin.js';
+
+export class HeadingLevelController {
```
dbatiste, line 4: I wonder if there are cases where the consumer would need to reset the level. For instance, in a dialog...? Maybe we would want an optional `level` property to override.

components/heading/demo/heading-stack-test.js
dlockhart: Yeah, the stuff you pass to `literal` has to "literally" be a literal -- no dynamic stuff, unless you use `unsafeStatic` like in the sample above.

I might mock out in this draft what a `<d2l-heading>` component might look like...

dbatiste: Yeah... I think I like the idea of the `<d2l-heading>` component - it seems like it would hide all this stuff from the consumer. They would just need to use `<d2l-heading>` and `<d2l-heading-stack>`. I think it could still support all the positioning flexibility that `<h#>` provides, and we'd just have to expose a property for the style level, and perhaps a property for the semantic level so they can override. They'd no longer need to import the heading level styles, so that's a ➕ .

It seems like the viability of this depends on everything on the page using it, ya?

components/heading/heading-stack.js
dlockhart: That's a really good point yeah -- like either a `level` property like you said, or a boolean `reset` property. If those are set, it would just return that value (or `1` for the `reset`) instead of asking upwards. I'll add this in!

components/heading/demo/heading-stack-test.js
dlockhart: For sure yeah, it would have to be all-or-nothing pretty much. Of course, the manually created headings with styles applied would still work in the meantime. I'll continue to flesh this out a bit.

components/heading/heading-stack.js
dlockhart: I've added `override-level` and am now using it in dialogs as an example.

components/dialog/dialog.js
```diff
@@ -149,7 +150,7 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 						<d2l-button-icon icon="tier1:close-small" text="${this.localize('components.dialog.close')}" @click="${this._abort}"></d2l-button-icon>
 					</div>
 				</div>
-				<div class="d2l-dialog-content" @pending-state="${this._handleAsyncItemState}">${content}</div>
+				<d2l-heading-stack class="d2l-dialog-content" override-level="3" @pending-state="${this._handleAsyncItemState}" style="display: block;">${content}</d2l-heading-stack>
```
jameswklassen, line 153: So just to confirm, this is how a consumer would render the heading?

dlockhart: No, this is an example of needing to "reset" the heading stack, since when dialogs open they blank out the rest of the page and right above here render a `<h2>` -- which is why this is resetting things such that if there's a nested heading inside the dialog, it'll be a `h3`.

There's an example of consuming the heading level inside `heading-stack-test.js`, but as discussed in there it's a bit awkward to do. So we're considering also adding a `<d2l-heading>` control that would reduce the complexity there.

dbatiste: No, the `d2l-heading-stack` just initializes or increments the heading level variable for that DOM subtree. Inside the stack, they would either have to render an `<h#>` using the controller to resolve the current level as Dave did below in `heading-stack-test.js`, or we'd provide a `<d2l-heading>` component renders the actual heading and hides the gnarly switch statement away.

dbatiste: Apologies. I shoulda refreshed the page before submitting my reply.

PR #2926, opened 2022-10-20
Add fullscreen-view switch to demo-snippet

bearfriend:
Adds a `d2l-switch` to demo snippets ~that have the `fullscreen` attribute~ to enable a more truly full-screen experience.

[US144714](https://rally1.rallydev.com/#/?detail=/userstory/665191647783&fdp=true): Add fullscreen switch to d2l-demo-snippet

fullscreen-demo-snippet


components/demo/demo-page.js
```diff
@@ -59,11 +65,15 @@ class DemoPage extends LitElement {
...
 			<main>
 				<h1 class="d2l-heading-2">${this.pageTitle}</h1>
-				<div class="d2l-demo-page-content"><slot></slot></div>
+				<div class="d2l-demo-page-content" @fullscreen-toggle="${this._handleFullscreenToggle}"><slot></slot></div>
 			</main>
 		`;
 	}
 
+	_handleFullscreenToggle() {
+		this.classList.toggle('no-scroll');
```
dbatiste, line 74: Another way to do this that doesn't modify the DOM outside of Lit's render cycle would be to set a `noScroll` property that reflects (as `no-scroll`), and change the CSS selector to be `[:host([no-scroll])]`.

components/demo/demo-snippet.js
```diff
@@ -186,6 +202,12 @@ class DemoSnippet extends LitElement {
 		this._applyAttr('dir', this._dir, true);
 	}
 
+	_handleFullscreenViewChange(e) {
+		this.fullscreenView = e.target.on;
+		const event = new CustomEvent('fullscreen-toggle', { bubbles: true, composed: true });
```
dbatiste, line 207: I think this should follow our event naming conventions and be something like `d2l-demo-snippet-fullscreen-toggle`.

components/demo/demo-snippet.js
```diff
@@ -82,6 +96,7 @@ class DemoSnippet extends LitElement {
 
 	render() {
 		const dirAttr = this._dir === 'rtl' ? 'rtl' : 'ltr';
+		const fullscreen = this.fullscreen ? html`<d2l-switch text="Fullscreen" ?on="${this.fullscreenView}" @change="${this._handleFullscreenViewChange}"></d2l-switch><br />` : null;
```
dbatiste, line 99: Maybe we want this switch to always be present? This existing `fullscreen` property just sorta means full width (unfortunate name), and I can imagine it being useful to toggle fullscreen even when the snippet is standard width.

components/demo/demo-snippet.js
```diff
@@ -9,6 +9,7 @@ class DemoSnippet extends LitElement {
 		return {
 			codeViewHidden: { type: Boolean, reflect: true, attribute: 'code-view-hidden' },
 			fullscreen: { type: Boolean, reflect: true },
+			fullscreenView: { type: Boolean, reflect: true, attribute: 'fullscreen-view' },
```
dbatiste, line 12: Looks like this property should be private, ya? If so, then I think this should be defined like:
```javascript
_fullscreenView: { state: true }
```
And conditionally render an internal class to apply the styles instead of using `:host([fullscreen-view])`. Generally, we prefer not sprouting private stuff on the host.

components/demo/demo-snippet.js
bearfriend: Yeah, I considered this. I'll gladly make that change.

bearfriend: Though, I guess we have to keep it named differently so as not to be breaking. There are only a couple uses of the existing `fullscreen`. Perhaps we can get those updated in tandem.

dbatiste: Yeah, if we can easily identify them, I'm all for updating the property name.

components/demo/demo-page.js
```diff
@@ -59,11 +66,22 @@ class DemoPage extends LitElement {
...
 	}
 
+	async _handleFullscreenToggle() {
+		if (this.noScroll) {
+			this.noScroll = false;
+			await this.updateComplete;
+			document.documentElement.scrollTop = this._previousScrollTop;
+		}
+		else {
+			this._previousScrollTop = document.documentElement.scrollTop;
```
bearfriend, line 81: Should `_previousScrollTop` be declared in some way?

bearfriend:
There [appears](https://cs.github.com/?scopeName=%40d2l&scope=org%3ABrightspace+OR+org%3ABrightspaceUI+OR+org%3ABrightspaceUILabs+OR+org%3ABrightspaceHypermediaComponents&q=%2Fd2l-demo-snippet%5Ba-zA-Z0-9%24%7B%7D%5C%5B%5C%5D%22-%3D%5Cs%5D%2B%3Ffullscreen%2F) to be 6 repos using the existing `fullscreen`.

I used this regex to search for instances: `/d2l-demo-snippet[a-zA-Z0-9${}\[\]"-=\s]+?fullscreen/`
(Github didnt like `[^>]`)

- BrightspaceUI/core‎
- ‎Brightspace/d2l-rubric‎
- ‎BrightspaceUILabs/navigation‎
- ‎Brightspace/discovery-fra‎
- ‎Brightspace/unified-content-experience‎
- ~‎Brightspace/blog.d2l.dev‎~ Likely don't want to change this
- ‎Brightspace/nova‎

components/demo/demo-page.js
dbatiste: Unless it needs to trigger the render cycle to run for the component, it doesn't need to be declared as a Lit property. Having said that, sometimes we will initialize these types of member variables in the constructor, but I don't think we're consistent about that - initializing a member variable in the constructor to `undefined` is somewhat pointless, other than perhaps giving readers a little more insight as they dive into the code.

bearfriend:
All repos above use core `>2.50` or `^2`

components/demo/demo-page.js
```diff
@@ -20,7 +20,8 @@ class DemoPage extends LitElement {
 
 	static get properties() {
 		return {
-			pageTitle: { type: String, attribute: 'page-title' }
+			pageTitle: { type: String, attribute: 'page-title' },
+			noScroll: { type: Boolean, reflect: true, attribute: 'no-scroll' }
```
dbatiste, line 24: Should `noScroll` be private?

components/demo/demo-snippet.js
```diff
@@ -84,15 +98,16 @@ class DemoSnippet extends LitElement {
 		const dirAttr = this._dir === 'rtl' ? 'rtl' : 'ltr';
 		const skeleton = this._hasSkeleton ? html`<d2l-switch text="Skeleton" ?on="${this._skeletonOn}" @change="${this._handleSkeletonChange}"></d2l-switch>` : null;
 		return html`
-			<div class="d2l-demo-snippet-demo-wrapper">
+			<div class="d2l-demo-snippet-demo-wrapper ${this._fullscreen ? 'fullscreen' : ''}">
```
dbatiste, line 101: Nit: usually we use the `classMap` directive to render conditional classes, but if this works ok then it seems fine.

dbatiste (commented):
I think this is looking pretty good. Just a minor question about the `noScroll` property. Not sure about the floating buttons visual-diffs... seems flakey.

components/demo/demo-snippet.js
bearfriend: Yeah it seemed excessive for managing one class on a simple boolean.

components/demo/demo-page.js
bearfriend: Changed this to an internal state and class instead.

bearfriend:
> I think this is looking pretty good. Just a minor question about the `noScroll` property. Not sure about the floating buttons visual-diffs... seems flakey.

Yeah, I don't know why they (and only they) would change some antialiasing.

components/demo/demo-snippet.js
```diff
@@ -186,6 +201,12 @@ class DemoSnippet extends LitElement {
 		this._applyAttr('dir', this._dir, true);
 	}
 
+	_handleFullscreenChange(e) {
+		this._fullscreen = e.target.on;
+		const event = new CustomEvent('d2l-demo-snippet-fullscreen-toggle', { bubbles: true, composed: true });
```
eKoopmans, line 206: Is this event being consumed anywhere?

eKoopmans: Oops - missed `demo-page.js` in my review!

dbatiste (approved):

PR #2927, opened 2022-10-20
Change tooltip display:none to visibility:hidden

oelhanafey:
[Rally ticket](https://rally1.rallydev.com/#/?detail=/defect/664018732801&fdp=true)

This PR updates the tooltip to use `visibility: hidden` when not shown instead of `display: none` to fix an accessibility issue in JAWS where tooltip text is not being read until it is opened for the first time. I explored a few different solutions but this was the least invasive solution and shouldn't change any existing usages or appearance. Though I wasn't able to figure out why exactly this fixes the problem.

This shouldn't have any effect on the appearance of the tooltips since they are not a part of the normal flow of the document. Apart from the potential space issues, there doesn't seem to be any other major concerns with this change and it will improve accessibility with JAWS.


components/tooltip/tooltip.js
```diff
@@ -183,9 +183,10 @@ class Tooltip extends RtlMixin(LitElement) {
 				--d2l-tooltip-outline-color: rgba(255, 255, 255, 0.32);
 				box-sizing: border-box;
 				color: white;
-				display: none;
+				display: inline-block;
 				position: absolute;
 				text-align: left;
+				visibility: hidden;
```
dlockhart, line 189: Crazy that it was this simple, but makes sense!

The only side effect I can think of is that these tooltips are now going to show up when a screen reader user is just arrowing through the content of the page using the virtual cursor. For cases where these tooltips are linked to an interactive element via `aria-labelledby`/`aria-describedby` that's totally fine... but for the other scenarios it may not be ideal.

A potential solution to that might be for us to us to distinguish between these two cases -- `display: none` for non-interactive and `visibility: hidden` for interactive ones.

oelhanafey: Yep, I'll do some testing and check if that works

oelhanafey: After looking into this it should not be a concern, [visibility: hidden](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns) removes itself from the accessibility tree so it will not be read out.

dlockhart: Great, thanks!

dlockhart (approved):

oelhanafey:
The display changes to `tooltip` seems to have broken these two visual diff tests by shifting the elements to the right:`d2l-filter-tags rtl basic is correct` and `d2l-filter-tags rtl two-filters is correct`. The `filter-tag` component contains a `tag-list` which uses `d2l-tooltip`.

Removing `display: none` from the tooltips should not have any effect on the appearance since tooltips use `position:absolute`. From testing on Firefox, Safari, Edge and Chrome this does not appear to actually change the appearance on the webpage. The difference is likely due to the visual-diff library getting confused and grabbing an incorrect rectangle of the element due to the hidden tooltips now appearing on the page.

Making the tooltip in taglist have `display: none` will fix this issue but I believe this to be a visual diff issue and not an issue with the component itself so I don't think it warrants making an exception. I think that we could merge the visual diffs as-is even though they appear odd and the visual diffs will still be useful to detect if there have been any changes since the previous iteration. Are there any concerns with this?

oelhanafey:
@dlockhart Adding a timeout actually does fix it. That's really odd and the issue only seems to arise when `dir` is set to `rtl`. 🤔 

What made you suspect it was a timing issue?

dlockhart:
> What made you suspect it was a timing issue?

Just how things were cut off... almost like they weren't finished rendering yet. And the fact that tooltips are now "rendering" (even though they're hidden) whereas previously the browser could have optimized that could make it take slightly longer.

oelhanafey:
@dlockhart Oops sorry 🤦 I had reverted the tooltip changes before standup when I was testing and forgot about it so that is why the tests were passing. The timeout does not seem to fix the issue.

dlockhart:
Haha, ah well worth a shot!

oelhanafey:
So are there any concerns with merging the diffs as-is?

dlockhart:
> So are there any concerns with merging the diffs as-is?

If they're consistent and not flaky, I'm OK with merging these. Like ideally we could get to the bottom of what changed, but since it doesn't affect the actual experience I'd rather get this in and cert fixed and if we want we can circle back later.

PR #2929, opened 2022-10-20
Remove borders in table header when no-column-border is set

oelhanafey:
[Rally ticket](https://rally1.rallydev.com/#/?detail=/defect/664126251235&fdp=true)

This PR removes table headers in the `d2l-table-wrapper` when the `no-column-border` attribute is set. This is a low risk change and shouldn't affect any current users of this style apart from the slight visual differences. 

Screen Shot 2022-10-20 at 3 52 34 PM
Screen Shot 2022-10-20 at 3 52 43 PM


dlockhart (approved):
Yay, that was an easy fix!

svanherk (approved):
Nice, this is what I was expecting the change to look like after investigating as part of that thread!  Glad it worked out so well

PR #2932, opened 2022-10-20
SPIKE - Dropdown alternatives

eKoopmans:
This is a draft PR to demonstrate a spike result.

I investigated dropdown alternatives to solve issues with how the dropdown is positioned, specifically related to clipping contexts etc.:

- Popup API is the most browser-native approach, but it's still experimental, doesn't support anchors yet, etc
    - It's available as an experimental feature on Chrome, and as a polyfill, but neither support anchors
- Popper works well with the `strategy: 'fixed'` option, displaying the dropdown on top of everything using `position: fixed`
    - It's a widely used library, e.g. in [Material-UI](https://mui.com/material-ui/react-popper/)
- Floating-UI seems almost exactly equivalent to Popper, with a slightly different "middleware" philosophy

If you run this locally, the alternatives will be viewable at:
http://localhost:8000/components/dropdown/demo/dropdown-alternatives.html

Rally: https://rally1.rallydev.com/#/?detail=/userstory/663397013693&fdp=true

---

## Videos:

Note: The videos below show Popper, but Floating-UI has similar results.

Main popper usage:
https://user-images.githubusercontent.com/8449639/197071595-23c6c712-e710-406c-8b0d-13c85ac3276a.mp4

Flipping at viewport boundaries:
https://user-images.githubusercontent.com/8449639/197071621-9b129a60-6afd-4978-b66a-a6233ac83ad2.mp4

Clean scrolling when using `placement=right`:
https://user-images.githubusercontent.com/8449639/197071611-51b0825f-7bb3-4175-98f6-948b1d532ad9.mp4


dbatiste:
Just a clarification - The Popup API is in origin trial currently, which means that it is available without experimental flags if one registers their origin to trial and adds the registration token to their pages. I'm not suggesting we do that, but the origin trials are intentionally time-limited, and after which the feature is generally released not long after assuming no major changes are necessary.

When I follow the Popper link to GitHub, it takes me to the Floating-UI repo. I get the impression Floating-UI is lower level, and the most recent version of Popper uses it.  I think we should have more discussion if we're thinking of adopting either of these.

eKoopmans:
> Floating-UI is lower level, and the most recent version of Popper uses it

Yep that was the feeling I got when playing with them, they felt very equivalent (though Popper was easier to use). And you're right, that's confirmed on [the repo](https://github.com/floating-ui/floating-ui): "Popper is now Floating UI!"

> The Popup API is in origin trial currently

Yeah, thanks for the clarification. I still think that without any anchoring logic, the only thing the Popup API buys over `position: fixed` is not having to mess with `z-index` - though it may be worth it just for that. It's probably possible to combine Popup API with one of these libraries.

dbatiste:
Chatted with Erik about this PR - we have the branch and all the details nicely summarized in Rally. Closing since we don'y need the open PR.

PR #2939, opened 2022-10-24
fix: just use a local variable for unique-id tracking

dlockhart:
This came up in the context of how older versions of `frau-framed` used the presence/absence of a global `D2L` to determine whether they're in an iFRA or a normal Brightspace page. Our usage here was tricking it into thinking an iFRA was a Brightspace page.

`frau-framed` has since been updated to additionally look for the aptly named `D2L.IsNotAnIFramedApp`, but while investigating I had this change ready and figured it was still a preferable solution to using globals.

### DIFF
--- a/helpers/uniqueId.js
+++ b/helpers/uniqueId.js
@@ -1,12 +1,6 @@
-window.D2L = window.D2L || {};
-window.D2L.UniqueId = window.D2L.UniqueId || {};
+let uniqueIndex = 0;
 
 export function getUniqueId() {
-
-	if (window.D2L.UniqueId._unique_index === undefined) {
-		window.D2L.UniqueId._unique_index = 0;
-	}
-	window.D2L.UniqueId._unique_index++;
-	return `d2l-uid-${window.D2L.UniqueId._unique_index}`;
-
+	uniqueIndex++;
+	return `d2l-uid-${uniqueIndex}`;
 }


dbatiste (approved):

dbatiste:
> Is there any risk here? It seems like this logic might be shared by other tools: https://search.d2l.dev/search?full=_unique_index&si=full&searchall=true
> 
> Just wondering if switching how these IDs are generated without switching the others might result in ID collisions.

Those appear to use a different variable. It looks like their implementation(s) could have already been the source of collisions.

dlockhart:
Yeah was just going to say the same... they're referencing what appears to be a super old Polymer copy of this code that uses `D2L.Id._unique_index` vs. we were using `D2L.UniqueId._unique_index`.

dbatiste:
They are HTML pages as well... so unless that HTML is somehow being injected into another document, it would be it's own document. 🤔

eKoopmans:
> They are HTML pages as well... so unless that HTML is somehow being injected into another document, it would be it's own document. 🤔

My guess is that they are, because they're all just a single `<script>` in an HTML file. But you're right, different variable so it should be fine. I can still imagine some random third-party-integration breaking though from this change, if they're using this "undocumented feature" 🤦.

dbatiste:
Re the HTML, must have been from the old HTML "imports" which was killed from the specs. Would it even be possible for them to do imports like that anymore?

I don't see any other usages of the `d2l-uid-` or `D2L.UniqueId` qualifiers. Both of these were/are considered private and subject to change, so if someone is using it that shouldn't be, and we can't identify them, I think we need to able to proceed with the change.

dlockhart:
They'd need to be using Polymer 1 with the v0 polyfill in order to use those old-school HTML imports.

PR #2944, opened 2022-10-25
Revert "Change tooltip display:none to visibility:hidden"

oelhanafey:
Reverts BrightspaceUI/core#2927.

The change is causing similar visual diff issues in [other repos](https://github.com/Brightspace/d2l-outcomes-user-progress/pull/493). Plan is to revert this for now but keep the a11y fix in cert which will give us more time to find a solution.


dlockhart (approved):

eKoopmans (approved):
On a surface level, the horizontal shift seems to make sense to me - a display of `inline-block` with `visibility: hidden` is still going to affect layout. I'm a bit confused why the tooltip as-is doesn't shift things when its display is switched on, but I haven't looked too closely...

oelhanafey:
@eKoopmans when the tooltip is closed it is `0x0` pixels though and the position in inspect element doesn't seem to coincide with where it gets shifted. And this shift only seems to happen in visual diffs and not in actual uses.

PR #2945, opened 2022-10-25
Updating Translations 2211 10/25/2022

scowing:
Updating Translation Packs


lang/cy.js
```diff
@@ -110,7 +110,7 @@ export default {
 	"components.tag-list.cleared-item": "Wedi tynnu’r eitem rhestr tag {value}",
 	"components.tag-list.interactive-label": "Rhestr Tag, {count} o eitemau",
 	"components.tag-list.num-hidden": "+ {count} yn rhagor",
-	"components.tag-list.role-description": "Tag List",
+	"components.tag-list.role-description": "Rhestr o Dagiau",
 	"components.tag-list.show-less": "Dangos Llai",
 	"components.tag-list.show-more-description": "Dewis i ddangos eitemau rhestr tag cudd",
 	"components.tag-list-item.role-description": "Tag",
```
dbatiste, line 116: Missed translation.

lang/da.js
```diff
@@ -110,7 +110,7 @@ export default {
 	"components.tag-list.cleared-item": "Fjernede taglisteelement {value}",
 	"components.tag-list.interactive-label": "Tagliste, {count} elementer",
 	"components.tag-list.num-hidden": "+ {count} mere",
-	"components.tag-list.role-description": "Tag List",
+	"components.tag-list.role-description": "Tagliste",
 	"components.tag-list.show-less": "Vis færre",
 	"components.tag-list.show-more-description": "Vælg for at få vist skjulte taglisteelementer",
 	"components.tag-list-item.role-description": "Tag",
```
dbatiste, line 116: Missed translation.

lang/hi-in.js
```diff
@@ -0,0 +1,125 @@
+/* eslint quotes: 0 */
```
dbatiste, line 1: Is this a new language we are supporting?

Also, looks like the tag translations were missed for `hi`.

devpow112: Looks like it was an issue with serge https://d2l.slack.com/archives/C0TF4KSGZ/p1666719158315249

dbatiste: Thanks Devon.

@oelhanafey , so for our translation updates, we'll just need to copy the following translations out of `hi-in.js` into out `hi.js` and then remove `hi-in.js`. This file was created in error due to a config issue that should be corrected for next time, but we still want those two translations.

```
"components.tag-list.role-description": "टैग सूची",
"components.tag-list-item.role-description": "टैग",
```

oelhanafey: > @oelhanafey , so for our translation updates, besides our regular updates, we'll need to copy the following translations out of `hi-in.js` into our `hi.js` and then remove `hi-in.js`.

Okay sounds good thanks. I'll copy those over

lang/nl.js
```diff
@@ -20,7 +20,7 @@ export default {
 	"components.filter.clearDescription": "Filters wissen voor {filterName}",
 	"components.filter.loading": "Laden van filters",
 	"components.filter.filterCountDescription": "{number, plural, =0 {Geen filters toegepast.} one {{number} filter toegepast.} other {{number} filters toegepast.}}",
-	"components.filter.filters": "Filters", // mfv-translated
+	"components.filter.filters": "filters",
```
oelhanafey: The capitalization in this translation was changed. I changed it back to the original `Filters`. Is that an appropriate change?

lang/da.js
dlockhart, line 116: Looks like for a few languages here (CY, DA, DE) "tag" translates to "tag" in those languages as well. Just to avoid the validator complaining at us, can we add `// mfv-translated` to those lines?

```suggestion
	"components.tag-list-item.role-description": "Tag", // mfv-translated
```

eKoopmans (approved):

lang/nl.js
eKoopmans: Seems reasonable, but a bit suspicious that it was changed...

dlockhart (approved):

PR #2946, opened 2022-10-25
Re-add tooltip display fix for a11y

oelhanafey:
[Rally ticket](https://rally1.rallydev.com/#/?detail=/defect/664018732801&fdp=true)

Re-adding the changes from #2927 in addition to adding `overflow: hidden`, `height: 0`, `width: 0` when the tooltip is not shown. The failing visual diff tests seemed to be caused by some text overflowing when the tooltip is hidden. JAWS works as expected after this change.

The additions of `overflow: hidden` and `width: 0`, `height: 0` were messing up `force-show` positioning so I added an exception for that attribute where it will never be hidden

This PR fixes an issue where JAWS was not reading tooltip labels of interactive elements.

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -184,11 +184,13 @@ class Tooltip extends RtlMixin(LitElement) {
 				box-sizing: border-box;
 				color: white;
 				display: inline-block;
+				height: 0;
 				overflow: hidden;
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Nice!

eKoopmans (approved):

PR #2947, opened 2022-10-25
feat: Description List - add `force-stacked` attribute

jameswklassen:
## Description
Add an optional `force-stacked` attribute to Description List.

This will take priority over the `breakpoint` attribute.


components/description-list/description-list-wrapper.js
```diff
@@ -71,12 +77,20 @@ class DescriptionListWrapper extends LitElement {
 
 	connectedCallback() {
 		super.connectedCallback();
-		this._resizeObserver.observe(this);
+
+		if (!this.forceStacked) {
```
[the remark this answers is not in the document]
jameswklassen: ahh okay, so that'd be in addition to the one in `connectedCallback` right? In case they don't pass anything?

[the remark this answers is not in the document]
dlockhart: It defaults to `false`, so I think the normal sequence will be:
- `connectedCallback()` gets called
- `updated()` gets called with `false`, at which point you can connect the observer

Then if `updated()` ever gets called with `true`, you can disconnect the observer if it was connected in the first place. To track whether it was connected or not, you may want to not initialize it in the constructor and then create it + connect it in `updated()`. We you disconnect it, set it back to `undefined`.

components/description-list/description-list-wrapper.js
```diff
@@ -86,6 +90,18 @@ class DescriptionListWrapper extends LitElement {
...
+		if (changedProperties.get('forceStacked')) {
+			if (!this.forceStacked) {
+				this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
+				this._resizeObserver.observe(this);
+			} else {
+				this._resizeObserver = undefined;
+				this._stacked = true;
+			}
+		}
+	}
```
dlockhart, line 103: You're going to want to disconnect it when it changes to `true` as well (if it was connected in the first place).

```suggestion
	updated(changedProperties) {
		if (changedProperties.has('forceStacked')) {
			if (!this.forceStacked) {
				this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
				this._resizeObserver.observe(this);
			} else {
				if (this._resizeObserver) {
					this._resizeObserver.disconnect();
					this._resizeObserver = undefined;
				}
				this._stacked = true;
			}
		}
	}
```

jameswklassen: agh yes 🤦

components/description-list/description-list-wrapper.js
```diff
@@ -41,6 +41,12 @@ class DescriptionListWrapper extends LitElement {
 			 * @type {number}
 			 */
 			breakpoint: { type: Number, reflect: true },
+			/**
+			 * Force the component to always use a stacked layout
+			 * This will override the breakpoint attribute
+			 * @type {boolean}
+			 */
+			forceStacked: { type: Boolean, reflect: true, attribute: 'force-stacked' },
```
dlockhart, line 49: We have been duplicating the attribute list in the README, so probably worth adding this there:
https://github.com/BrightspaceUI/core/tree/main/components/description-list#properties

jameswklassen: Yeah, still need to update documentation. Was gonna do that once the technical things have been worked out in case naming, etc changes 😄

dlockhart: Cool -- what you have works for me!

dlockhart (approved):

dbatiste (approved):

dbatiste:
Might be good to have a visual-diff that shows this in action.

components/description-list/description-list-wrapper.js
```diff
@@ -86,6 +89,21 @@ class DescriptionListWrapper extends LitElement {
...
+		if (changedProperties.has('forceStacked')) {
+			if (!this.forceStacked) {
+				this._resizeObserver = new ResizeObserver(this._onResize.bind(this));
+				this._resizeObserver.observe(this);
+			} else {
+				if (this._resizeObserver) {
+					this._resizeObserver.disconnect();
+					this._resizeObserver = undefined;
+				}
+				this._stacked = true;
```
eKoopmans, line 102: I have this exact same sort of logic in my sticky mixin 😄.

eKoopmans (approved):

dlockhart (approved):

PR #2952, opened 2022-10-26
Replace Object.hasOwn with Object.prototype.hasOwnProperty

bearfriend:
Fixes older Safari errors (in SauceLabs) on `hasOwn`.


dlockhart (approved):
Thanks!

PR #2958, opened 2022-10-28
Skeletenized object-property-list

louiscoste:
Added skeleton mixin to `object-property-list`


components/object-property-list/object-property-list-item.js
```diff
@@ -44,14 +45,21 @@ export class ObjectPropertyListItem extends RtlMixin(LitElement) {
 			:host([dir="rtl"]) .item-icon {
 				margin: -0.1rem 0 0 0.3rem;
 			}
+			.d2l-skeletize {
+				display: inline-block;
+				width: 100px;
+				height: 0.7rem;
```
louiscoste, line 51: I am not sure how to do this any better. I am open to suggestions

components/object-property-list/object-property-list-item.js
```diff
@@ -44,14 +45,21 @@ export class ObjectPropertyListItem extends RtlMixin(LitElement) {
...
 	}
 
 	render() {
-		return html`
+		return !this.skeleton ? html`
 			${this._renderIcon()}
 			<span>${this.text}</span>
 			${this._renderSeparator()}
+		` : html`
+			<div class="d2l-skeletize"></div>
```
louiscoste, line 62: I am creating a separate div so that we always show a skeleton block no matter the length of the text.

I am doing regardless of the text length because while loading we might not have the text available yet

components/object-property-list/object-property-list-item.js
```diff
@@ -4,12 +4,13 @@ import { css, html, LitElement, nothing } from 'lit';
...
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
+import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
 
 /**
  * A single object property, to be used within an object-property-list,
  * with an optional icon.
  */
-export class ObjectPropertyListItem extends RtlMixin(LitElement) {
+export class ObjectPropertyListItem extends SkeletonMixin(RtlMixin(LitElement)) {
```
dlockhart, line 13: Question for discussion: in cases where skeletons are used, do we feel as though consumers know that there will be X number of items in the property list, or might they not know how many they have until their async data has loaded?

If they do know ahead of time, then I think having the items support skeletons makes sense -- you'll get nice skeletons with the dot separators between them. But if they don't know how many, then it may make sense to support skeleton on the list itself... or alternatively I guess they could render some placeholder items that may or may not represent the final real number.

components/object-property-list/object-property-list-item.js
dlockhart: I have a look at the `d2l-skeletize-<width>` helper classes -- they're intended for exactly this purpose.

components/object-property-list/object-property-list-item.js
louiscoste: Based on my understanding from [this](http://proddevutil.desire2learn.d2l/d2c/content/product%20design/design.d2l%20archive/Skeleton%20Page%20Loaders%20%E2%80%93%20Daylight%20Design%20System.html), for lists we would want to display a potentially hard coded number of elements. But this link mostly refers to tables, not this component.

So I would assume that if we do not know how many then we could display 3 (or something like that)

components/object-property-list/object-property-list-item.js
dlockhart: Also, do we maybe want the icon and the separator to show up when it's in skeleton mode? If so, you might be able to just apply `d2l-skeletize-10` (or whatever size looks good) to the text `<span>` directly when in skeleton-mode.

eKoopmans: It seems to me that we would still want the separators (and maybe icons?) rendered when the item is skeletized. Also from the docs it looks like `d2l-skeletize-<width>` is for block elements - the items are `inline` by default. Would it make more sense for consumers to do something like:

`<d2l-object-property-list-item text="Placeholder" skeleton>`

And then the skeleton will just be the width of the placeholder? Not sure what's best for accessibility here, whether we'd want a descriptive placeholder or for it to read as nothing...

components/object-property-list/object-property-list-item.js
eKoopmans: Yeah I was thinking about that too, I think both use cases could make sense.

components/object-property-list/object-property-list-item.js
louiscoste: So according to [this](http://proddevutil.desire2learn.d2l/d2c/content/product%20design/design.d2l%20archive/Skeleton%20Page%20Loaders%20%E2%80%93%20Daylight%20Design%20System.html) we do not want to display the icons.

Also, I am not sure if I am not doing it right, but `d2l-skeletize-<width>` did not help me. It seems to set the height and width to 0 if there is nothing inside the `div`.

components/object-property-list/object-property-list-item.js
eKoopmans: If skeletizing the object-property-list itself, I'd assume just one item with no separators as the safest default.

components/object-property-list/object-property-list.js
```diff
@@ -21,7 +32,10 @@ class ObjectPropertyList extends LitElement {
 	}
 
 	render() {
-		return html`<slot class="d2l-body-small"></slot>`;
+
+		return !this.skeleton || !this.skeletonCount ?
```
dbatiste, line 36: I was sorta thinking these items would be rendered in the slot as default content.

```javascript
const slotCotnent = (this.skeleton && this.skeletonCount > 0 ? [...Array()].map(...) : nothing);

return html`<slot class="d2l-body-small">${slotContent}</slot>`;
```

louiscoste: Looks like the slot overwrites whatever gets passed into `skeletons`.

dbatiste: Any elements provided in the slot by the consumer would override the default content of the slot. Is that what you mean?

louiscoste: That's right

components/object-property-list/object-property-list-item.js
```diff
@@ -44,28 +45,33 @@ export class ObjectPropertyListItem extends RtlMixin(LitElement) {
 			:host([dir="rtl"]) .item-icon {
 				margin: -0.1rem 0 0 0.3rem;
 			}
+			:host([skeleton]) .d2l-skeletize {
+				display: inline-flex;
+				height: 0.7rem;
+				width: 100px;
+			}
```
eKoopmans, line 52: I'm still thinking we should be leaving these spans un-modified (inline display) and expecting placeholder text.

eKoopmans (commented):
It would also be useful to add some visual-diff tests for the skeleton cases - maybe one at the list level, and one with some number of explicit skeleton items.

components/object-property-list/object-property-list-item.js
eKoopmans: @dlockhart You were mentioning approaches like `d2l-skeletize-<width>`, is it standard practice to use empty fixed-width divs for skeletons, or to use placeholder text?

dlockhart: Code searching around, I'm seeing quite a few `<div class="d2l-skeletize-75">${someValueThatWillEventuallyBeThere}</div>`.

It's harder to search for placeholder text. Placeholder text is also interesting in that it probably needs to be localized (even though it's not shown!) and would show up for screen readers which might be a strange experience.

components/object-property-list/object-property-list-item.js
```diff
@@ -44,28 +45,33 @@ export class ObjectPropertyListItem extends RtlMixin(LitElement) {
...
 
 	_renderSeparator() {
-		return html`
+		return !this.skeleton ? html`
 			<span class="separator">
 				<span class="d2l-offscreen">${getSeparator({ nonBreaking: true })}</span>
 				<d2l-icon icon="tier1:bullet" aria-hidden="true"></d2l-icon>
 			</span>
-		`;
+		` : nothing;
```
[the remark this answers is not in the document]
dlockhart: Is that the case? From the Slack conversation:

> Dave L: Because if we render the separators, then all the spacing is just handled by the existing component
> Jeff G: Oh that would be even better if that’s how it works. I didn’t raise the dot separator because I could take it or leave it, but I like letting the component handle its own spacing instead of approximating it with extra css … nice.

My interpretation of that was that Jeff didn't think we needed the level of detail that the dots provide in the skeleton, but once we mentioned that if we DID show the dots then all the spacing would be handled, he was on board.

components/object-property-list/object-property-list-item.js
eKoopmans: Yeah exactly, the screen reader experience is what I was talking about here:

> Not sure what's best for accessibility here, whether we'd want a descriptive placeholder or for it to read as nothing...

Could it also be a problem if it had no screen-reader content while skeletized? Would we want users to know that there's content coming? Should that content get announced when it becomes available?

According to [this doc](http://proddevutil.desire2learn.d2l/d2c/content/product%20design/design.d2l%20archive/Skeleton%20Page%20Loaders%20%E2%80%93%20Daylight%20Design%20System.html), it sounds like we should be setting `aria-busy="true"` on the parent element while there's a skeleton...

components/object-property-list/object-property-list-item.js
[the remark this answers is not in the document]
eKoopmans: Just referring to the [doc](http://proddevutil.desire2learn.d2l/d2c/content/product%20design/design.d2l%20archive/Skeleton%20Page%20Loaders%20%E2%80%93%20Daylight%20Design%20System.html) that Louis has shared a few times:

> When multiple elements of dynamic text metadata are concatenated via dots, do not render the dots until the dynamic text is rendered.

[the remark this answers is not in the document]
dlockhart: I'm not sure how much weight we should be putting in a doc from 2020 on the old design.d2l archive at this point... 

@geurts do you feel comfortable weighing in again and overriding Colin's original design? He wanted the dots suppressed, which we can do but not suppressing them gives us the spacing between items for free.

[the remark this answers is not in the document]
louiscoste: Currently we are getting the spacing for free. This is what it looks like without the dots: 
image

components/object-property-list/object-property-list-item.js
dlockhart: (This is why I'm not putting a ton of weight in that doc. Colin wrote it well before we ended up implementing it, so it didn't get a technical review. I don't even think we used it during implementation.)

components/object-property-list/object-property-list-item.js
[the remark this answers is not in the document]
eKoopmans: > Currently we are getting the spacing for free

Actually, that's my old nemesis "arbitrary whitespace" 😂. It's coming from the newlines/tabs between elements, which includes how `d2l-object-property-list-item` is coded internally. If you remove them you lose the space (see first two items in this screenshot):

image

I've been [wanting to remove the effect of this whitespace](https://github.com/BrightspaceUI/core/pull/2649), but we haven't come to a solution yet.

components/object-property-list/object-property-list-item.js
dbatiste: I was playing around with this a bit on Friday.

I agree with @eKoopmans that we should maintain the display: inline. When I was playing around with it, I was getting pretty good layout when the dot separators showing, but I was using some placeholder text... and I was thinking we might have to add `aria-hidden: true` while in skeleton mode.

dbatiste: One of the nice things about having some placeholder text, is that we get the height set by the browser in an equivalent way. No changing to inline-block necessary to apply a one-off height to _try_ and match layout. Also, no issues trying to match the baseline.

components/object-property-list/object-property-list-item.js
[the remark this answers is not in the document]
dlockhart: Back again! 😆  We should come up with a nickname for him... Arby?

components/object-property-list/object-property-list-item.js
louiscoste: I can try to change the span to be unmodified.

If we want to go with the placeholder text, any thoughts on what it should be?

eKoopmans: I'd say the consumer should specify the placeholder text. In the case of the skeleton at the list level itself - maybe `Item 1` etc?

And yep maybe best just to stick `aria-hidden="true"` on all skeleton items...

components/object-property-list/object-property-list-item.js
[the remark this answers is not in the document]
geurts: > I'm not sure how much weight we should be putting in a doc from 2020 on the old design.d2l archive at this point...
> 
> @geurts do you feel comfortable weighing in again and overriding Colin's original design? He wanted the dots suppressed, which we can do but not suppressing them gives us the spacing between items for free.

Yes, we can override that old decision. That whole doc is being reworked for publishing on the Daylight site. 

Can we render the dots in the same colour as the rest of the skeleton, though? If we can't easily do that, then I would leave them out (they will look super weird if they're rendered so much darker than the rest of the skeleton).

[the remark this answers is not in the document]
dbatiste: >Can we render the dots in the same colour as the rest of the skeleton, though?

Yep, we should be able to do that. And yep, when I was playing around with it, the dots did look a little weird unless they were colored the same as the skeleton.

components/object-property-list/test/object-property-list.visual-diff.html
```diff
@@ -52,5 +52,20 @@
 		</d2l-object-property-list>
 	</div>
 
+	<div class="visual-diff" id="list-skeleton">
+		<d2l-object-property-list skeleton skeleton-count="3">
+			<d2l-object-property-list-item text="Example item"></d2l-object-property-list-item>
+		</d2l-object-property-list>
```
eKoopmans, line 58:
```suggestion
		<d2l-object-property-list skeleton skeleton-count="3"></d2l-object-property-list>
```

eKoopmans:
Rest of the changes look good to me!

For context, Louis and I met today to go over some fiddly details in getting `inline` working. Most of our issues were with long text. In the end we found that:

- skeleton works very poorly (and inconsistently between browsers) on inline text that spans multiple lines
- to resolve it, we forced the skeletons to be `inline-block`, with no-wrap and overflow hidden
- we also applied a max 80% width to prevent the separator from having to wrap onto its own line (except when the container is now very small)
- the overflow also causes problems with parent flex containers, which also need `overflow: hidden` to properly shrink ignoring overflow
    - fortunately, `d2l-demo-snippet` already had a built-in `overflow-hidden` option for this

☝️ These actually seem like problems with the skeleton mixin itself, not specific to `object-property-list`, but we kept our fixes scoped to just `object-property-list`. I think it would be worth planning a future fix to how our skeletons deal with these problems in general.

eKoopmans (approved):
Thanks for the submission Louis, LGTM!

components/object-property-list/object-property-list-item.js
```diff
@@ -67,6 +78,10 @@ export class ObjectPropertyListItem extends RtlMixin(LitElement) {
 			</span>
 		`;
 	}
+
+	_renderText() {
+		return html`<span class="d2l-skeletize" aria-hidden=${this.skeleton ? 'true' : 'false'}>${this.text}</span>`;
```
dlockhart, line 83: Nit:
```suggestion
		return html`<span class="d2l-skeletize" aria-hidden="${this.skeleton ? 'true' : 'false'}">${this.text}</span>`;
```

components/object-property-list/object-property-list.js
```diff
@@ -1,27 +1,45 @@
...
 				--d2l-object-property-list-item-separator-display: none;
 			}
 		`];
 	}
 
 	render() {
-		return html`<slot class="d2l-body-small"></slot>`;
+
+		const slotContents = this.skeleton && this.skeletonCount > 0 ? [...Array(this.skeletonCount)].map(() => html`
+			<d2l-object-property-list-item text=${this.localize('components.object-property-list.item-placeholder-text')} skeleton></d2l-object-property-list-item>
```
dlockhart, line 39:
```suggestion
			<d2l-object-property-list-item text="${this.localize('components.object-property-list.item-placeholder-text')}" skeleton></d2l-object-property-list-item>
```

dlockhart (approved):
LGTM -- just a couple minor notes about quotes around attributes. Thanks Louis!

components/object-property-list/object-property-list-item.js
louiscoste: out of curiosity, is this a style preference, or is there a functional difference?

dlockhart: Like technically all attributes in HTML are strings, so should be quoted. Lit may be doing some magic here to fix things for you, but if you missed quotes and the value resolved to something with a space in it, then you could end up with: 

```html
<d2l-object-property-list-item text=hello there what's up skeleton>
```

In that case, the browser would interpret the value as "hello" and "there", "what's", "up" and "skeleton" as boolean attributes, I think? Lit may be fixing things for you though like I said, but IMO it's cleaner and safer to just use the quotes.

dlockhart (approved):

components/object-property-list/object-property-list-item.js
eKoopmans: Unquoted is valid HTML, in HTML5+:
https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

As mentioned [here](https://stackoverflow.com/a/13056802), there are cases where it was required pre-HTML5. But yes we still want to use them for consistency and to avoid issues with whitespace.

dlockhart: Haha so I guess "technically" was the wrong word to use, thanks for linking to that. So _technically_ they can be unquoted if we're certain that the values don't contain whitespace. In practice, it's easier IMO to just always quote them.

PR #2961, opened 2022-10-31
Skeletized tooltip help component

louiscoste:
Added skeleton mixin to `tooltip-help`


dbatiste (commented):
This looks pretty good. I'll just take it for a test-drive before approving.

dbatiste:
We should include a visual-diff test that captures the skeleton mode.

dbatiste:
Looks like we get some of the underline peeking out from under the skeleton. We probably just need to apply `text-decoration: none;` when in skeleton mode.
image

components/tooltip/tooltip-help.js
```diff
@@ -103,12 +107,14 @@ class TooltipHelp extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
 			'd2l-body-small': !this.inheritFontStyle
 		};
 		return html`
-			<button id="d2l-tooltip-help-text" class="${classMap(classes)}" type="button">
+			<button id="d2l-tooltip-help-text" class="${classMap(classes)} d2l-skeletize" type="button">
```
dlockhart, line 110: Suggest adding this to the `classes` so there isn't a leading space when `d2l-body-small` isn't there:

```javascript
const classes = {
	'd2l-body-small': !this.inheritFontStyle,
	'd2k-skeletize': true
};
```

dlockhart (approved):
LGTM!

dbatiste (approved):

components/tooltip/tooltip-help.js
```diff
@@ -76,6 +77,9 @@ class TooltipHelp extends FocusMixin(FocusVisiblePolyfillMixin(LitElement)) {
 			d2l-tooltip {
 				cursor: text;
 			}
+			:host([skeleton]) button.d2l-skeletize {
```
dbatiste, line 80: If we use `:host([skeleton]) #d2l-tooltip-help-text.d2l-skeletize` selector, we can avoid the `!important`. We try to only rely on `!important` when absolutely necessary.

dbatiste (approved):
Perfect... thanks Louis!

PR #2966, opened 2022-11-01
feat: Add selectionProvider to link an element

eKoopmans:
This PR adds a `.selectionProvider` optional attribute to the `SelectionObserverMixin`. The idea is that it can be used as an alternative to the `selection-for` attribute, when passing an ID isn't sufficient (e.g. children within a Shadow DOM that are in a different scope than the provider).

This will be used by the `d2l-selection-header` component that I'm creating as part of [US144131](https://rally1.rallydev.com/#/?detail=/userstory/656749634867&fdp=true). The header uses `SelectionObserverMixin` itself, and then passes its `._provider` to its children via `.selectionProvider`.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/668445846425&fdp=true


components/selection/selection-observer-mixin.js
```diff
@@ -79,4 +85,9 @@ export const SelectionObserverMixin = superclass => class extends superclass {
 			this.selectionInfo = new SelectionInfo();
 		}
 	}
+
+	_updateProviderFor() {
+		const selectionComponent = this.selectionFor && this.getRootNode().querySelector(`#${cssEscape(this.selectionFor)}`);
```
eKoopmans, line 90: I've left this as `querySelector` from the original logic, but it seems to me that it could be replaced with `getElementById`.

Also I've added the `this.selectionFor &&` check - in the original logic, if `selectionFor` changed from a string to undefined/empty, the querySelector was failing with an error.

components/selection/selection-observer-mixin.js
```diff
@@ -51,26 +56,27 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
-		this._selectionForObserver.observe(this.getRootNode(), {
-			childList: true,
-			subtree: true
-		});
+				this._selectionForObserver.observe(this.getRootNode(), {
+					childList: true,
+					subtree: true
+				});
+			}
+		}
```
eKoopmans, line 71: This is mostly unchanged from the original logic, with two exceptions:

- `unsubscribeObserver` is moved into `_updateProvider`
- `_updateProvider` now requires an element to be passed to it, so we call `_updateProviderFor` here which computes that element first

components/selection/selection-observer-mixin.js
```diff
@@ -16,7 +16,12 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
 			 */
 			selectionInfo: { type: Object },
-			_provider: { type: Object, attribute: false }
+			/**
+			 * HtmlElement of the `SelectionMixin` component this component wants to observe (alternative to selectionFor)
+			 * @type {string}
+			 */
+			selectionProvider: { attribute: false },
+			_provider: { state: true }
 		};
```
eKoopmans, line 25: I didn't add anything to the readme because there's already nothing there about this mixin...

components/selection/selection-observer-mixin.js
```diff
@@ -16,7 +16,12 @@ export const SelectionObserverMixin = superclass => class extends superclass {
 			 * @type {object}
 			 */
 			selectionInfo: { type: Object },
-			_provider: { type: Object, attribute: false }
+			/**
+			 * HtmlElement of the `SelectionMixin` component this component wants to observe (alternative to selectionFor)
+			 * @type {string}
+			 */
+			selectionProvider: { attribute: false },
```
dbatiste, line 23: Do we need both `selectionProvider` and `_provider`? A bunch of stuff uses `_provider`, so eliminating it becomes a little more involved - while other selection components reference it, it's named as a private property since it's internal to the selection components, and not expected to be referenced by anything else.

eKoopmans: It should be possible to only use `_provider`, but then we'd need to wire up a setter on it to handle unsubscribing/resubscribing etc. I'll take a stab at it.

svanherk:
So not to add _more_ complications to this, but a goal for the selection mixins was to _eventually_ use the subscription controllers under the hood, but we haven't gotten around to that story. I created a [POC almost a year ago](https://github.com/BrightspaceUI/core/pull/1901/files#diff-a7412e30193c0f86945842eaebc7ea7709b8f1b5a541f5b42a7bd56026f5735a) when first making the controllers of what that would look like.  The actual merged controllers ended up [here ](https://github.com/BrightspaceUI/core/blob/main/controllers/subscriber/subscriberControllers.js) and probably look a bit different, this just gives an idea.
In that `_provider` actually goes away altogether, and comes from the controller.  I suspect we'd need to add a new `SubscriberController` type to handle this case, or expand the `id` one.  Feel free to ignore this if it totally overcomplicates things and whoever tackles the story can worry about it then. But figured I should raise this in case there's a way to set things up now so they're in a good place for that work, or even _do_ that work if it makes things easier.

svanherk:
(The POC I linked also takes a stab at moving the form and validation stuff which is totally separate from the `selection` work).

eKoopmans:
> subscription controllers under the hood
>  I suspect we'd need to add a new `SubscriberController` type to handle this case

@svanherk I had a look and yep that's my impression too. If anything that would have us moving back in the direction of a more "public" property instead of `_provider`. Probably a decision for the future though.

components/selection/selection-observer-mixin.js
```diff
@@ -16,21 +16,21 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
 		// delay subscription otherwise import/upgrade order can cause selection mixin to miss event
 		requestAnimationFrame(() => {
-			if (this.selectionFor) return;
+			if (this.selectionFor || this._provider) return;
 
 			const evt = new CustomEvent('d2l-selection-observer-subscribe', {
```
eKoopmans, line 35: With the change to a setter, there's a _slight_ change to the logic here:

- when this event resolves and sets `this._provider = evt.detail.provider`, it'll go through the setter logic
- that will trigger an extra `newProvider.subscribeObserver(this)` that wasn't previously being triggered - it should have no effect, but is a bit of extra computation (re-triggers `_updateSelectionObservers`)
- an alternative here would be to set `this.__provider` (the private var) and then manually trigger `this.requestUpdate('_provider', oldProvider)`

components/selection/selection-observer-mixin.js
```diff
@@ -16,21 +16,21 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
-			_provider: { type: Object, attribute: false }
+			_provider: { state: true }
 		};
 	}
 
 	constructor() {
 		super();
 		this.selectionInfo = new SelectionInfo();
-		this._provider = null;
+		this.__provider = null;
```
eKoopmans, line 26: I don't love this double-underscore for the private variable, it means the code now has `_provider` and `__provider` - I'm open if anyone thinks there's a better naming option.

svanherk:
> > subscription controllers under the hood
> > I suspect we'd need to add a new `SubscriberController` type to handle this case
> 
> @svanherk I had a look and yep that's my impression too. If anything that would have us moving back in the direction of a more "public" property instead of `_provider`. Probably a decision for the future though.

Yup makes sense - would be good to enhance the selection mixin observer tests to help catch regressions of your use case when that work does eventually happen

components/selection/selection-observer-mixin.js
```diff
@@ -16,29 +16,29 @@ export const SelectionObserverMixin = superclass => class extends superclass {
...
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
 		// delay subscription otherwise import/upgrade order can cause selection mixin to miss event
 		requestAnimationFrame(() => {
-			if (this.selectionFor) return;
+			if (this.selectionFor || this._provider) return;
```
dbatiste, line 33: It seems like there is an edge case here where if the observing element is disconnected from the DOM and then reconnected elsewhere in the DOM, it wouldn't get a new provider assigned. 

For the event code path immediately below, that could be mitigated by setting the `_provide = null` in the `disconnectedCallback` - I am almost a little surprised it isn't already doing that. We must be assuming this edge case isn't worth dealing with...? I think setting it to `null` in the `disconnectedCallback` complicates the case where the `_provider` is assigned directly though.

components/selection/selection-observer-mixin.js
dbatiste: Yeah it's kinda awkward eh. We _could_ make the property public, but just configure it so it's undocumented. I have mixed feelings either way.

components/selection/selection-observer-mixin.js
eKoopmans: I have a fix for this, but it has its own edge case in the default event-driven behaviour that may cause an issue.

@dbatiste I've got a rewrite in the works using the full event-driven approach we talked about on Friday, rather than using `._provider`, I think it may end up cleaner.

eKoopmans:
Closing in favour of #2979.

PR #2979, opened 2022-11-07
feat: Pass provider to selection-observer children

eKoopmans:
This PR replaces #2966. It improves the `SelectionObserverMixin` to automatically subscribe any of its selection-observer children, if it receives a subscription event from them. This is to support the case where a selection-observer is using `selection-for="some-id"` and has children in its shadow DOM that need to subscribe to that ID, which would otherwise be outside of their visible scope.

The PR also improves a few existing issues identified in #2966, with added test cases to cover them (see the first commit).

There is one possible edge case where the new behaviour may affect existing usages:

- a selection-observer is **nested** inside a selection-provider (e.g. `<selection-provider id="some-id1">`)
- it subscribes to a different provider instead, using `selection-for="some-id2"`
- and it has children that don't specify any `selection-for`
- in this case, the previous behaviour would have those children subscribe to `some-id1`, whereas the new behaviour would subscribe them to `some-id2`

Maintaining that old behaviour would directly contradict the proposed new behaviour - I assume there's no cases where it would have been used that way, but please correct me if I'm wrong!

Rally: https://rally1.rallydev.com/#/?detail=/userstory/668445846425&fdp=true

### DIFF
--- a/components/selection/selection-observer-mixin.js
+++ b/components/selection/selection-observer-mixin.js
@@ -30,7 +30,10 @@ export const SelectionObserverMixin = superclass => class extends superclass {
 		super.connectedCallback();
 		// delay subscription otherwise import/upgrade order can cause selection mixin to miss event
 		requestAnimationFrame(() => {
-			if (this.selectionFor) return;
+			if (this.selectionFor) {
+				this._handleSelectionFor();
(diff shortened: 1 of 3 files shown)


dbatiste (approved):
Looks good!

I think that edge case is acceptable.

PR #2981, opened 2022-11-08
feat: Create generic selection-header for lists and tables

eKoopmans:
This PR is step 2 of creating a "bulk actions header" to the table. It:

- extracts the core functionality of `d2l-list-header` into a separate `d2l-selection-header`
- refactors `d2l-list-header` to inherit from the selection header, sprinkling in its own required changes
- moves the header's sticky logic entirely internal, so it doesn't depend on lists specifically
- updates the readme and demo to use the `d2l-selection-header`

There will be some changes required to the selection-header in order to implement the table-header, but I wanted to keep this PR's scope limited to just selection-header, and have a separate PR for the table header.

Step 1 is #2979, which improves the selection-observer mixin, and is required for the last few examples in this selection demo to work correctly.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/668478189363&fdp=true


eKoopmans:
Okay, I need to improve how I'm detecting the sticky `_scrolled` state - it works fine when stuck to the viewport, but not within a scroll container. Will work on a fix.

components/selection/selection-header.js
```diff
@@ -0,0 +1,186 @@
...
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._stickyObserverDisconnect();
+	}
+
+	render() {
+		const classes = this._getSelectionHeaderContainerClasses();
+		return html`
+			<div class="d2l-sticky-edge"></div>
```
eKoopmans, line 127: This replaces the "d2l-list-top-sentinel" that was previously being rendered in `list.js`, *outside* of the list-header. I've moved it inside the header but with `position: absolute` and `margin-top: -1px` to achieve the same effect - when it scrolls out of view, the rest of the header is treated as stuck (`._scrolled`).

components/selection/selection-header.js
```diff
@@ -0,0 +1,186 @@
...
+	_stickyObserverUpdate() {
+		this._stickyObserverDisconnect();
+
+		if (!this.noSticky && typeof(IntersectionObserver) === 'function') {
+			this._stickyIntersectionObserver = new IntersectionObserver(([e]) => {
+				this._scrolled = !e.isIntersecting;
+			}, { threshold: 1 });
+
+			const target = this.shadowRoot.querySelector('.d2l-sticky-edge');
+			this._stickyIntersectionObserver.observe(target);
```
eKoopmans, line 180: This is mostly the same as the original (from `list.js`). I've added `threshold: 1` which tells the IntersectionObserver to watch for when 100% of the element is visible - though since it's an empty div, it doesn't really make a difference either way.

components/selection/selection-header.js
```diff
@@ -0,0 +1,186 @@
...
+			:host([_scrolled]) .d2l-selection-header-shadow {
+				background-color: var(--d2l-selection-header-background-color, white);
+				bottom: -4px;
+				box-shadow: 0 8px 12px -9px rgba(0, 0, 0, 0.3);
+				clip: rect(30px, auto, 200px, auto);
+				height: 40px;
+				position: absolute;
+				width: 100%;
+				z-index: -1;
+			}
```
eKoopmans, line 63: Note that I've kept the shadow styles the exact same as they were for `d2l-list-header`. I think these styles work fine as-is, but in the future we might want to tweak them a bit if any of these sizes/etc are specific to lists.

components/selection/demo/selection.html
```diff
@@ -56,22 +57,19 @@ <h2>Multiple Selection</h2>
...
+							<d2l-selection-action text="Settings" icon="tier1:gear"></d2l-selection-action>
+							<d2l-selection-action-dropdown text="Actions 1" requires-selection>
+								<d2l-dropdown-menu>
+									<d2l-menu label="Actions">
+										<d2l-selection-action-menu-item text="Action 1"></d2l-selection-action-menu-item>
+										<d2l-selection-action-menu-item text="Action 2"></d2l-selection-action-menu-item>
+									</d2l-menu>
+								</d2l-dropdown-menu>
+							</d2l-selection-action-dropdown>
+						</d2l-selection-header>
```
eKoopmans, line 71: These selection demos (and the ones in the readme) all had a custom-made selection header, so I've replaced those all now with `d2l-selection-header`. I've used the `no-sticky` attribute to keep the behaviour/appearance the same as before.

components/selection/selection-header.js
eKoopmans: I found that the `_scrolled` state was being triggered when scrolling left/right (i.e. when the left edge of the header was obscured).

I've fixed that by removing the `threshold: 1` here and adding `left`/`right` styles to the `.d2l-sticky-edge` - now the entire div needs to be hidden to trigger the `_scrolled` state.

components/selection/selection-header.js
dbatiste: Yeah these styles are very finicky. Jeff, Noah, and I spent a fair amount of time fiddling with them to get them working properly. I would hope we don't need different sets of shadow styles for different contexts.

components/selection/selection-header.js
dbatiste: Previously I think the sentinel element was just a zero-height `div` so layout wasn't an issue. Was there a problem you encountered that necessitates the positioning/negative margin styles?

components/selection/selection-header.js
```diff
@@ -0,0 +1,188 @@
...
+				display: block;
+				position: sticky;
+				top: 0;
+			}
+			:host([no-sticky]) {
+				background-color: transparent;
+				position: static;
+			}
+			.d2l-selection-header-shadow {
+				transition: box-shadow 200ms ease-out;
```
dbatiste, line 52: I notice that we previously did not have a media query to disable the transition. It is super minor, but if the consumer prefers reduced motion we should just avoid it entirely unless necessary (like a spinner), plus it can help avoid issues with visual-diffs. We used to write them the opposite way (using `prefers-reduced-motion: reduce`), but now that we no longer support legacy Edge, we can write the animation as a progressive enhancement. I think we should include it:
```css
	.d2l-selection-header-shadow {
		transition: box-shadow 200ms ease-out;
	}
}
```

components/selection/selection-header.js
eKoopmans: Yep, it's still zero-height.

Previously the sentinel was outside of the header, so it would scroll off the page before the header got stuck. My goal here was to keep the sentinel inside the header, to keep everything encapsulated (rather than having to define the sentinel separately in e.g. `list.js`).

So to accomplish that, I:

- put it inside the (sticky) header
- gave it an absolute position and negative margin to render it 1px above the top edge of the sticky header
    - this way it goes offscreen when the header gets stuck, same as the external sentinel did
    - absolute position also means it doesn't affect the flow of the other elements
- note that `absolute` works a bit differently in terms of horizontal size, so I had to give it `left: 0` and `right: 0` CSS to get it to span the width of the container
    - otherwise it's triggered as fully offscreen when scrolling the left edge of the header offscreen
    - `width: 100%` would also have worked
    - positioning is all relative to the selection-header (nearest [positioned ancestor](https://developer.mozilla.org/en-US/docs/Web/CSS/position#types_of_positioning)) so `left`/`right` should be safe to use

This approach should also work for making things sticky against left/right/bottom as well.

components/selection/selection-header.js
```diff
@@ -0,0 +1,188 @@
...
+		this._stickyObserverDisconnect();
+	}
+
+	render() {
+		const classes = this._getSelectionHeaderContainerClasses();
+		return html`
+			<div class="d2l-sticky-edge"></div>
+			<div class="${classMap(classes)}">
+				${this.noSelection ? null : html`
+					<d2l-selection-select-all></d2l-selection-select-all>
```
dbatiste, line 132: Do you have an eventual plan for the select-all? In the context of `d2l-list`, the select-all is contained within the `d2l-list-header` element, whereas in the context of `d2l-table-wrapper`, I think that consumers will be required to render it within their `<th>` element.

We might need to expose a slot where `d2l-list-header` can provide a `d2l-selection-select-all`. The alternative would be to change this from an inheritance to composition rendering approach where the sub-types have more rendering control.

components/selection/selection-header.js
eKoopmans: Cool. Yeah my concern is that our designs want a drop-shadow now for the sticky `<thead>`, and to accomplish that we'd need to apply the drop shadows to the thead or th's themselves, because it won't be valid to add extra divs in the table.

components/selection/selection-header.js
```diff
@@ -0,0 +1,188 @@
...
+				<div class="d2l-selection-header-actions">
+					<d2l-overflow-group opener-type="icon"><slot @slotchange="${this._handleSlotChange}"></slot></d2l-overflow-group>
+				</div>
+			</div>
+			${!this.noSticky ? html`<div class="d2l-selection-header-shadow"></div>` : null}
+		`;
+	}
+
+	updated(changedProperties) {
+		if (changedProperties.has('noSticky')) {
```
dbatiste, line 149: I think we probably want `super.updated(changedProperties);` here.

components/selection/selection-header.js
eKoopmans: Yeah, I think we can just use the `noSelection` flag - the [design](https://rally1.rallydev.com/#/?detail=/userstory/656749634867&fdp=true) only shows "1-50 of 1,000 items" there, which I believe is just part of the "select all pages" logic. Still investigating that though.

The only issue there is that currently `noSelection` also omits the `d2l-selection-select-all-pages`.

components/selection/selection-header.js
dbatiste: Sounds great. Thanks for the explanation! 😄

eKoopmans: It just occurred to me that I could do `top: -1px` instead of `margin-top`, should be equivalent.

components/selection/selection-header.js
dbatiste: Ok, yeah, so I think we have some options. We can also introduce more granular properties, but at some point it just makes more sense let the user take more rendering control rather than baking it into this component. We can always tweak this in your next PR.

eKoopmans: Yeah I was thinking of tackling that in the next PR, especially since there's still uncertainty about what we actually want displayed there for tables.

components/selection/README.md
```diff
@@ -441,3 +427,59 @@ The `d2l-selection-summary` component may be placed inside the selection control
 | `no-selection-text` | String | Text to display if no items are selected. By default, the "0 selected" message is displayed. |
 | `selection-for` | String | Id of the corresponding `SelectionMixin` component, if not placed within it. |
 <!-- docs: end hidden content -->
+
+## Selection Header [d2l-selection-header]
+
+The `d2l-selection-header` provides a standardized wrapper to display selection information and actions. It includes a a select-all checkbox, summary, a slot for `d2l-selection-action`s, and overflow-group behaviour.
```
dbatiste, line 433:
```suggestion
The `d2l-selection-header` provides a standardized wrapper to display selection information and actions. It includes a select-all checkbox, summary, a slot for `d2l-selection-action`s, and overflow-group behaviour.
```

components/selection/demo/selection.html
dbatiste: The previous demo was more just set up to show the various building blocks, so now that we have a new, somewhat higher level building block, I'd probably just render it with its default sticky mode, or provide two demos... one sticky and one not.

dbatiste (approved):
Just the typo and maybe the demo to update. Otherwise looks good!

PR #2983, opened 2022-11-08
Update dialog readme for clarity around nested dialogs.

dbatiste:


components/dialog/README.md
```diff
@@ -32,13 +32,14 @@ The `d2l-dialog` element is a generic dialog that provides a slot for arbitrary
 * Label primary actions with clear and predictable language. Use verbs like, “Add” or “Save” that indicate the outcome of a dialog rather than, “OK” or “Close”
 * Keep dialog titles concise
 * Maintain a language relationship between the action that triggered the dialog, dialog title, and dialog primary button.
+* Nest a dialog within the DOM of the dialog containing the opener when stacking dialogs
```
eKoopmans, line 35: This reads to me as still encouraging stacking dialogs... Maybe something like:

> When it is necessary to stack multiple dialogs, ensure the stacked dialog is nested within the DOM of the opening dialog

dbatiste: Yeah fair, I can rephrase.

PR #2984, opened 2022-11-08
resolution for more-less component defect which resulted in extra whitespace

ksanctis:
https://rally1.rallydev.com/#/10153049633d/dashboard?detail=%2Fdefect%2F645660620675

### DIFF
--- a/components/more-less/demo/more-less.html
+++ b/components/more-less/demo/more-less.html
@@ -29,8 +29,8 @@
 
 			<d2l-demo-snippet>
 				<template>
-					<d2l-more-less height="6em">
-						<p id="img-target">Hi everybody! <a href="">Hi!</a></p>
+					<d2l-more-less height="8em">
+						<p id="img-target">Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody! <a href="">Hi!</a></p>
 					</d2l-more-less>
 				</template>
 			</d2l-demo-snippet>
@@ -39,7 +39,7 @@
 
 			<d2l-demo-snippet>
 				<template>
-					<d2l-more-less expanded="">
+					<d2l-more-less expanded>
 						<p id="clone-target">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum elementum venenatis arcu sit amet varius. Maecenas posuere magna arcu, quis maximus odio fringilla ac. Integer ligula lorem, faucibus sit amet cursus vel, pellentesque a justo. Aliquam urna metus, molestie at tempor eget, vestibulum a purus. Donec aliquet rutrum mi. Duis ornare congue tempor. Nullam sed massa fermentum, tincidunt leo eu, vestibulum orci. Sed ultrices est in lacus venenatis, posuere suscipit arcu scelerisque. In aliquam ipsum rhoncus, lobortis ligula ut, molestie orci. Proin scelerisque tempor posuere. Phasellus consequat, lorem quis hendrerit tempor, sem lectus sagittis nunc, in tristique dui arcu non arcu. Nunc aliquam nisi et sapien commodo lacinia. <a href="">Quisque</a> iaculis orci vel odio varius porta. Fusce tincidunt dolor enim, vitae sollicitudin purus suscipit eu.</p>
 					</d2l-more-less>
 				</template>
--- a/components/more-less/more-less.js
+++ b/components/more-less/more-less.js
@@ -1,4 +1,5 @@
 import '../button/button-subtle.js';
+import '../expand-collapse/expand-collapse-content.js';
 import { css, html, LitElement } from 'lit';
 import { getComposedChildren, isComposedAncestor } from '../../helpers/dom.js';
 import { classMap } from 'lit/directives/class-map.js';
@@ -276,6 +277,7 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 
 	__init_setBaseHeight() {
 		this.__contentHeight = this.height;
+		this.__maxHeight = 'unset';
 
 		requestAnimationFrame(() => {
 			this.__init_measureBaseHeight();
--- a/components/more-less/more-less.js
+++ b/components/more-less/more-less.js
@@ -199,7 +199,7 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 		}
 
 		if (this.expanded) {
-			this.__contentHeight = `unset`;
+			this.__contentHeight = 'unset';
 			return;
 		}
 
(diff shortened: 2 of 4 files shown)


eKoopmans (commented):
When I tested locally, it looks like the more-less no longer animates as it opens/closes - is that deliberate?

I'm also wondering about the visual diffs, it looks like the goldens have all been deleted...

components/more-less/more-less.js
```diff
@@ -231,7 +239,8 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 
 	__expand() {
 		this.__transitionAdded = true;
-		this.__contentHeight = `${this.__content.scrollHeight}px`;
+		this.__contentHeight = 'unset';
```
oelhanafey, line 242: While it seems to be working fine while I test it, there seems to be some inconsistency with when `__contentHeight` and `__maxHeight` are set. On initialization `__maxHeight` is unset. It would probably be a good idea to initialize that to something. 

After expanding or shrinking `__contentHeight` is unset and never re-set. Does the component still work as expected if you set `__maxHeight` and remove `__contentHeight` entirely?

components/more-less/demo/more-less.html
```diff
@@ -29,8 +29,8 @@ <h2>More-less with height</h2>
 
 			<d2l-demo-snippet>
 				<template>
-					<d2l-more-less height="6em">
-						<p id="img-target">Hi everybody! <a href="">Hi!</a></p>
+					<d2l-more-less height="8em">
+						<p id="img-target">Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody! <a href="">Hi!</a></p>
```
oelhanafey, line 33: Nit: It might look cleaner to replace this with a "Lorem ipsum" sentence to be consistent with the other demos

components/more-less/more-less.js
```diff
@@ -1,4 +1,5 @@
 import '../button/button-subtle.js';
+import '../expand-collapse/expand-collapse-content.js';
```
dbatiste, line 2: Looks like this import statement is left over from when you were trying to use `d2l-expand-collapse`.

components/more-less/more-less.js
```diff
@@ -48,6 +49,7 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 			inactive: { type: Boolean, reflect: true },
 			__blurBackground: { type: String },
 			__contentHeight: { type: String },
+			__maxHeight: { type: String },
```
dbatiste, line 52: As of Lit 2, internal reactive state properties can be declared using Lit's `state`:
```javascript
__maxHeight: { state: true },
```
Feel free to update the others to this as well (`__blurBackground`, `__contentHeight`, `__transitionAdded`). This component was originally written pre-Lit 2.

components/more-less/more-less.js
```diff
@@ -231,7 +240,8 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 
 	__expand() {
 		this.__transitionAdded = true;
-		this.__contentHeight = `${this.__content.scrollHeight}px`;
+		this.__contentHeight = 'unset';
+		this.__maxHeight = 'unset';
```
oelhanafey, line 244: I think this should be set to `this.height` since the component starts out in "shrunken" form.

components/more-less/more-less.js
oelhanafey: To correct my previous comment, I think there is actually a use for `__contentHeight`. The height attribute is used for the animation [here](https://github.com/BrightspaceUI/core/blob/fb51e8b8be04af51722e2e97cabc923b5f97fea4/components/more-less/more-less.js#L65). Changing these to `unset` is likely why the animation has stopped working.

I think what would fix this is to set the height with 2 separate cases:
1. When in shrunken form, set `this.__contentHeight` to the smallest of  `${this.__content.scrollHeight}` and  `this.height`
2. When in expanded form: `this.__contentHeight = ${this.__content.scrollHeight}px`

The problem with that though is it might not be simple to compare the heights in case 1 since the `height` property is a string 🤔

oelhanafey: After thinking about this for a bit, I think there might be an easier way. It should be possible to get the animations working properly by changing the transition animation to use `max-height` instead of `height`. Then the two cases are easier to code:
1. When in shrunken form set: `this.maxHeight = this.height`
2. When in expanded form set: `this.maxHeight = ${this.__content.scrollHeight}px`

And then I think you might be able to remove the `__contentHeight` property as it would be unused.

components/more-less/more-less.js
```diff
@@ -267,6 +277,7 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 
 	__init_setBaseHeight() {
 		this.__contentHeight = this.height;
+		this.__maxHeight = 'unset';
```
oelhanafey, line 280: I think this should be set to `this.height` since the component is initialized in "shrunken" form.

components/more-less/more-less.js
ksanctis: Change has been made, thanks!

components/more-less/more-less.js
ksanctis: Change has been made, thanks!

components/more-less/more-less.js
ksanctis: Change has been made, thanks!

components/more-less/more-less.js
ksanctis: Change has been made, thanks!

eKoopmans (commented):
Looks good to me, but I'm not up to speed on some of @oelhanafey's latest feedback.

components/more-less/more-less.js
```diff
@@ -197,24 +197,29 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 			return;
 		}
 
+		if (this.expanded) {
+			this.__maxHeight = `${this.__content.scrollHeight}px`;
+			return;
+		}
+
```
oelhanafey, line 204: I think this is unnecessary as the max height should already be set in `__adjustToContent_resize()`

components/more-less/test/more-less.test.js
```diff
@@ -6,7 +6,7 @@ export function waitForHeight(elem) {
 	return new Promise((resolve) => {
 		function check() {
 			const content = elem.shadowRoot.querySelector('.d2l-more-less-content');
-			if (content.style.height === '') {
+			if (content.style.maxHeight === content.scrollHeight) {
```
oelhanafey, line 9: Oops my mistake. I think this test is just checking if the attribute has been initialized yet. So I think we want:
`if (content.style.maxHeight === '') {`

oelhanafey (commented):
This looks good! Couple small fixes and should be good to go.

We could initialize `this.__maxHeight = this.height` in the constructor so it is never undefined.

components/more-less/more-less.js
ksanctis: Change has been made, thanks!

components/more-less/test/more-less.test.js
ksanctis: Change has been made, thanks!

oelhanafey (approved):
Nice! 🎉

eKoopmans (approved):
🎉

PR #2988, opened 2022-11-09
Fix the backdrop showAccessible method.

dbatiste:
This PR fixes an issue where non-`form`/`a` elements that previously had `aria-hidden="true|false"` would not have their `aria-hidden` attributes properly restored when "showing" the accessible elements upon hiding the backdrop. 

The root of the issue is that the `showAccessible` method was assuming that any element should have their `aria-hidden` attribute removed, regardless of whether or not `d2l-backdrop-aria-hidden` was set, but this attribute would only be set for `form` and `a` elements.  

Example case:
```html
<button aria-hidden="true">...</button>
<div>
     <div id="target">...</div>
     <d2l-backdrop for-target="target"></d2l-backdrop>
</div>
```

I think it is worth looking into whether we should just be setting `role="presentation"` and `aria-hidden="true"` on all elements. I think this code was copied from elsewhere - specifically, the [LMS Shim.js](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Shim/Shim.js?r=d3403c8c#159). [Looks like Nick Matthews introduced that logic](https://search.d2l.dev/history/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Shim/Shim.js?n=25&start=25&r1=4&r2=3), but there doesn't appear to be any notes about those elements.

### DIFF
--- a/components/backdrop/backdrop.js
+++ b/components/backdrop/backdrop.js
@@ -3,9 +3,9 @@ import { css, html, LitElement } from 'lit';
 import { cssEscape, getComposedChildren, getComposedParent, isVisible } from '../../helpers/dom.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 
-export const BACKDROP_ROLE = 'd2l-backdrop-role';
-
-const BACKDROP_ARIA_HIDDEN = 'd2l-backdrop-aria-hidden';
+export const BACKDROP_ROLE = 'data-d2l-backdrop-role';
+const BACKDROP_HIDDEN = 'data-d2l-backdrop-hidden';
+const BACKDROP_ARIA_HIDDEN = 'data-d2l-backdrop-aria-hidden';
 
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 
@@ -160,7 +160,7 @@ function hideAccessible(target) {
 
 			if (child.tagName === 'SCRIPT' || child.tagName === 'STYLE') continue;
 			if (path.indexOf(child) !== -1) continue;
-			if (child.hasAttribute('d2l-backdrop-hidden')) continue;
+			if (child.hasAttribute(BACKDROP_HIDDEN)) continue;
 
 			const role = child.getAttribute('role');
 			if (role) child.setAttribute(BACKDROP_ROLE, role);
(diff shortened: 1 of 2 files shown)


components/backdrop/backdrop.js
```diff
@@ -209,12 +209,14 @@ function showAccessible(elems) {
 		} else {
 			elem.removeAttribute('role');
 		}
-		const ariaHidden = elem.getAttribute('d2l-backdrop-aria-hidden');
-		if (ariaHidden) {
-			elem.setAttribute('aria-hidden', ariaHidden);
-			elem.removeAttribute('d2l-backdrop-aria-hidden');
-		} else {
-			elem.removeAttribute('aria-hidden');
+		if (elem.nodeName === 'FORM' || elem.nodeName === 'A') {
```
eKoopmans, line 212: Yep I'm curious why we only manipulate aria-hidden on these two types of nodes to begin with - but if that remains unchanged, then this fix makes sense.

components/backdrop/backdrop.js
```diff
@@ -209,12 +209,14 @@ function showAccessible(elems) {
...
+			const ariaHidden = elem.getAttribute('d2l-backdrop-aria-hidden');
```
eKoopmans, line 213: Should we use a const like `BACKDROP_ARIA_HIDDEN` to be consistent with `BACKDROP_ROLE`?

eKoopmans (approved):

components/backdrop/backdrop.js
dbatiste: Sure, yep, that's worthwhile.

components/backdrop/backdrop.js
dbatiste: Yeah, I'd like to at least get this fixed, and then we can look at making a bigger change to this with a bit more investigation. Don't want to risk introducing a separate problem.

components/backdrop/backdrop.js
dlockhart: If we're making changes here, it'd probably be better to use `data-backdrop-aria-hidden`. Sprouting arbitrary invalid attribute names on native elements probably isn't great for HTML validation. (I'm going through a whole accessibility audit of our invalid HTML, so it's top of mind!)

dbatiste: Cool, yeah, I was just about to switch that, as well as the one that's exported.

components/backdrop/backdrop.js
dlockhart: Yeah I'd be tempted to apply this logic to all elements, not just forms and anchors. Our old Bitbucket server is down so I have a feeling we may never find the original pull request when the change was introduced.

dlockhart: Cool, that's fair.

PR #2989, opened 2022-11-10
Update the d2l-input-number component to recheck validity.

dbatiste:
[DE50582](https://rally1.rallydev.com/#/?detail=/defect/666083170587&fdp=true)

This PR updates the `d2l-input-number` component to reset validity if the `min`, `minExclusive`, `max`, or `maxExclusive` properties are updated. i.e. if these properties are updated, they may change the state from invalid to valid or vice versa depending on the current value.

### DIFF
--- a/components/inputs/input-number.js
+++ b/components/inputs/input-number.js
@@ -352,27 +352,35 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 	updated(changedProperties) {
 		super.updated(changedProperties);
 
+		let checkValidity = false;
 		changedProperties.forEach((oldVal, prop) => {
 			if (prop === 'value') {
 				this.setFormValue(this.value);
-
-				let rangeUnderflowCondition = false;
-				if (typeof(this.min) === 'number') {
-					rangeUnderflowCondition = this.minExclusive ? this.value <= this.min : this.value < this.min;
-				}
-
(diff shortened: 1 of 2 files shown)


components/inputs/input-number.js
```diff
@@ -352,27 +352,35 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 	updated(changedProperties) {
 		super.updated(changedProperties);
 
+		let checkValidity = false;
 		changedProperties.forEach((oldVal, prop) => {
 			if (prop === 'value') {
 				this.setFormValue(this.value);
-
```
dbatiste, line 358: This logic is just moved outside the loop so it only gets executed once as needed for this update cycle. The logic is the same other than that.

components/inputs/test/input-number.test.js
```diff
@@ -339,12 +339,72 @@ describe('d2l-input-number', () => {
...
+				value: 10,
+				change: { prop: 'maxExclusive', value: false },
+				expectedError: ''
 			}
 		].forEach((test) => {
 			it(test.name, async() => {
 				const elem = await fixture(test.fixture);
 				if (test.value !== null) elem.value = test.value;
 				await elem.updateComplete;
+				if (test.change) {
```
dlockhart, line 404: Good idea!

dlockhart (approved):
LGTM 👍

PR #2991, opened 2022-11-10
Fix d2l-input-number so that invalid property reflects internal d2l-input-text validation state.

dbatiste:
[DE50568](https://rally1.rallydev.com/#/?detail=/defect/666002656097&fdp=true)

This PR updates the `d2l-input-number` component so that it's `invalid` property properly reflects the validation rules defined for both `d2l-input-number` as well as the underlying `d2l-input-text` element. When the validity of the underlying element changes (ex. due to `required`), we need to update the validity of `d2l-input-number`.

### DIFF
--- a/components/inputs/test/input-number.test.js
+++ b/components/inputs/test/input-number.test.js
@@ -408,6 +408,7 @@ describe('d2l-input-number', () => {
 				const errors = await elem.validate();
 				if (test.expectedError) expect(errors).to.contain(test.expectedError);
 				else expect(errors).to.be.empty;
+				expect(elem.invalid).to.equal(!!test.expectedError);
 			});
 		});
 	});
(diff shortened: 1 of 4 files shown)


components/inputs/test/input-number.test.js
```diff
@@ -408,6 +408,7 @@ describe('d2l-input-number', () => {
 				const errors = await elem.validate();
 				if (test.expectedError) expect(errors).to.contain(test.expectedError);
 				else expect(errors).to.be.empty;
+				expect(elem.invalid).to.equal(!!test.expectedError);
```
dbatiste, line 411: There's already a test for the `required` case - we just need to check the assert the `invalid` state. Good to check this for all the cases. 😄

dlockhart (approved):
Seems valid!

dbatiste:
> Seems valid!

The validation stuff is kinda hairy.

PR #2995, opened 2022-11-11
Add helper to enable waiting for updateComplete on all elements in a sub-tree.

dbatiste:
This new `getUpdateCompleteAll(node)` helper function walks the composed tree and `await`s any elements that have `updateComplete`. The motivation for this is that sometimes in tests we need to wait arbitrary amount of time for all the components to update - it is not enough to just wait for the root element's `updateComplete`. This is kinda costly, so I was reluctant to add this to our component helpers.

### DIFF
--- a/tools/dom-test-helpers.js
+++ b/tools/dom-test-helpers.js
@@ -10,3 +10,18 @@ export function keyDown(element, keycode) {
 	event.code = keycode;
 	element.dispatchEvent(event);
 }
+
+export async function getUpdateCompleteAll(node) {
+
+	if (!node || (node.nodeType !== Node.ELEMENT_NODE && node.nodeType !== Node.DOCUMENT_NODE && node.nodeType !== Node.DOCUMENT_FRAGMENT_NODE)) {
+		throw new TypeError('Invalid node. Must be nodeType document, element or document fragment');
+	}
+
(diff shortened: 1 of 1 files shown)


tools/dom-test-helpers.js
```diff
@@ -10,3 +11,18 @@ export function keyDown(element, keycode) {
...
 }
+
+export async function getUpdateCompleteAll(node) {
+
+	if (!node || (node.nodeType !== Node.ELEMENT_NODE && node.nodeType !== Node.DOCUMENT_NODE && node.nodeType !== Node.DOCUMENT_FRAGMENT_NODE)) {
+		throw new TypeError('Invalid node. Must be nodeType document, element or document fragment');
+	}
+
+	const getUpdateComplete = async elem => {
+		if (elem.updateComplete) await elem.updateComplete;
```
dbatiste, line 22: This code purposefully waits for `updateComplete` before checking its children. My first implementation of this just collected all the promises in a flat array without waiting, but I was observing inconsistent results due to timing issues.

dlockhart (approved):
Looks good. As we discussed on Slack, this should probably just be built into our own wrapper around `fixture`... but I'm OK trying this out for a bit as a separate helper and then making a decision.

PR #2999, opened 2022-11-16
Move switches into a dropdown for full-width, fullscreen demos

bearfriend:
When fullscreen mode is enabled, the settings switches currently remain on the right side, taking up space the component would otherwise use. This can give an inaccurate view of the component when breakpoints are triggered.

1. For `full-width` demos, moves the settings switches into a dropdown and increases the demo width to fill the viewport
2. Fixes a weird edge case where the icon inside a `d2l-button-icon` was preventing click of the parent button
3. Remove unused `d2l-button-icon` from `demo-snippet`
4. Fixes `full-width` attribute-property syncing
5. Fixes synced scrolling of fullscreen demos in browsers like Polypane

full-width-fullscreen

### DIFF
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -121,7 +121,6 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 
 				.d2l-button-icon {
 					height: 0.9rem;
-					pointer-events: none;
 					width: 0.9rem;
 				}
 
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -122,6 +122,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				.d2l-button-icon {
 					height: 0.9rem;
 					width: 0.9rem;
+					pointer-events: none;
 				}
 
 				:host([translucent]) button {
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -142,7 +142,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 						transition: none;
 					}
 				}
-`
+			`
 		];
 	}
 
--- a/components/button/button-icon.js
+++ b/components/button/button-icon.js
@@ -121,8 +121,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 
 				.d2l-button-icon {
 					height: 0.9rem;
-					width: 0.9rem;
 					pointer-events: none;
+					width: 0.9rem;
 				}
 
 				:host([translucent]) button {
@@ -143,7 +143,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 						transition: none;
 					}
 				}
-			`
+`
 		];
 	}
 
--- a/components/demo/demo-page.js
+++ b/components/demo/demo-page.js
@@ -53,6 +53,10 @@ class DemoPage extends LitElement {
 			.d2l-demo-page-content > ::slotted(d2l-demo-snippet) {
 				margin-bottom: 36px;
 			}
+
+			.no-scroll ::slotted(d2l-demo-snippet) {
+				height: 0;
+			}
 		`];
 	}
 
--- a/components/demo/demo-page.js
+++ b/components/demo/demo-page.js
@@ -53,10 +53,6 @@ class DemoPage extends LitElement {
 			.d2l-demo-page-content > ::slotted(d2l-demo-snippet) {
 				margin-bottom: 36px;
 			}
-
-			.no-scroll ::slotted(d2l-demo-snippet) {
-				height: 0;
-			}
 		`];
 	}
 
--- a/components/demo/demo-snippet.js
+++ b/components/demo/demo-snippet.js
@@ -51,7 +51,7 @@ class DemoSnippet extends LitElement {
 			.d2l-demo-snippet-demo {
 				flex: 1 1 auto;
 				position: relative;
-				translate: 0;
+				translate: 0; /* create stacking context to prevent demos from leaking outside their containers */
 			}
 			:host([full-width]) .d2l-demo-snippet-demo-wrapper.fullscreen .d2l-demo-snippet-demo {
 				width: 100vw;
--- a/components/demo/demo-snippet.js
+++ b/components/demo/demo-snippet.js
@@ -51,6 +51,7 @@ class DemoSnippet extends LitElement {
 			.d2l-demo-snippet-demo {
 				flex: 1 1 auto;
 				position: relative;
+				translate: 0;
 			}
 			:host([full-width]) .d2l-demo-snippet-demo-wrapper.fullscreen .d2l-demo-snippet-demo {
 				width: 100vw;
--- a/components/demo/demo-snippet.js
+++ b/components/demo/demo-snippet.js
@@ -45,7 +45,6 @@ class DemoSnippet extends LitElement {
 				background-color: white;
 				height: fit-content;
 				inset: 0;
-				overflow: auto;
 				position: absolute;
 				z-index: 2;
 			}
--- a/components/demo/demo-snippet.js
+++ b/components/demo/demo-snippet.js
@@ -1,6 +1,8 @@
 import './code-view.js';
-import '../button/button-icon.js';
+import '../button/button-subtle.js';
 import '../switch/switch.js';
+import '../dropdown/dropdown.js';
+import '../dropdown/dropdown-content.js';
 import { css, html, LitElement } from 'lit';
 
 class DemoSnippet extends LitElement {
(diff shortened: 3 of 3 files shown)


components/button/button-icon.js
```diff
@@ -122,6 +122,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				.d2l-button-icon {
 					height: 0.9rem;
 					width: 0.9rem;
+					pointer-events: none;
```
bearfriend, line 125: This is for the weird case where the icon within the button was getting a pointer mouse and preventing click.

components/demo/demo-page.js
```diff
@@ -53,6 +53,10 @@ class DemoPage extends LitElement {
 			.d2l-demo-page-content > ::slotted(d2l-demo-snippet) {
 				margin-bottom: 36px;
 			}
+
+			.no-scroll ::slotted(d2l-demo-snippet) {
+				height: 0;
+			}
```
bearfriend, line 59: I think this is not actually doing anything. Left over from a first attempt.
```suggestion
```

components/demo/demo-snippet.js
```diff
@@ -1,20 +1,23 @@
 import './code-view.js';
-import '../button/button-icon.js';
```
bearfriend, line 2: Appears to be unused.

components/demo/demo-snippet.js
```diff
@@ -68,6 +75,23 @@ class DemoSnippet extends LitElement {
 				position: sticky;
 				top: 0;
 			}
+			d2l-dropdown.settings-dropdown {
+				position: fixed;
+				top: -0.25rem;
```
bearfriend, line 80: This probably should match the border radius units, which I think are in `px`. Is there a variable?
```suggestion
				top: -5px;
```

components/demo/demo-snippet.js
```diff
@@ -201,10 +231,14 @@ class DemoSnippet extends LitElement {
 		this._applyAttr('dir', this._dir, true);
 	}
 
-	_handleFullscreenChange(e) {
+	async _handleFullscreenChange(e) {
 		this._fullscreen = e.target.on;
+		this._settingsPeek = this._fullscreen;
 		const event = new CustomEvent('d2l-demo-snippet-fullscreen-toggle', { bubbles: true, composed: true });
 		this.dispatchEvent(event);
+		await this.update;
```
bearfriend, line 239: This isn't technically necessary but seems like good form.

components/demo/demo-snippet.js
```diff
@@ -40,15 +43,19 @@ class DemoSnippet extends LitElement {
 			}
 			.d2l-demo-snippet-demo-wrapper.fullscreen {
 				background-color: white;
+				height: fit-content;
 				inset: 0;
 				overflow: auto;
-				position: fixed;
+				position: absolute;
```
bearfriend, line 49: This pushes the scrolling back into the body box model or whatever, which allows Polypane to sync scrolling in fullscreen mode.

scrolling-sync

components/demo/demo-snippet.js
bearfriend: Never mind. They do use `rem` and `-0.25rem` is about the limit before we start seeing the radius.

components/button/button-icon.js
```diff
@@ -142,7 +143,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 						transition: none;
 					}
 				}
-			`
+`
```
bearfriend, line 146: 😠

components/button/button-icon.js
dbatiste: Is there a way I can reproduce this? Just want to understand better when this is/was coming up.

components/demo/demo-snippet.js
```diff
@@ -68,6 +74,23 @@ class DemoSnippet extends LitElement {
...
 				top: 0;
 			}
+			d2l-dropdown.settings-dropdown {
+				background-color: white;
+				border-radius: 6px;
+				box-shadow: 0 0 0 1px var(--d2l-color-celestine-minus-1);
+				position: fixed;
+				right: 1rem;
+				top: -0.25rem;
+				transition: translate 0.15s, box-shadow 0.15s;
```
dbatiste, line 84: This is just the demo-snippet so it's not a huge deal, but normally we would layer the transition/animation on with:
```css
	d2l-dropdown.settings-dropdown {
		transition: translate 0.15s, box-shadow 0.15s;
	}
}
```

components/demo/demo-snippet.js
dbatiste: Is this meant to be `await this.updateComplete;`?

Also, if I understand correctly, the purpose of `_settingsPeek` is to just show the dropdown opener briefly in the top-right corner upon toggling to fullscreen, is that correct?

components/demo/demo-snippet.js
```diff
@@ -97,6 +120,16 @@ class DemoSnippet extends LitElement {
 	render() {
 		const dirAttr = this._dir === 'rtl' ? 'rtl' : 'ltr';
 		const skeleton = this._hasSkeleton ? html`<d2l-switch text="Skeleton" ?on="${this._skeletonOn}" @change="${this._handleSkeletonChange}"></d2l-switch>` : null;
+		const switches = html`
+			<d2l-switch text="RTL" ?on="${dirAttr === 'rtl'}" @change="${this._handleDirChange}"></d2l-switch><br />
+			<d2l-switch text="Fullscreen" ?on="${this._fullscreen}" @change="${this._handleFullscreenChange}"></d2l-switch><br />
+			${skeleton}`;
+		const settings = this.fullWidth && this._fullscreen ? html`
```
dbatiste, line 127: Is this meant to be `this.fullWidth || this._fullscreen`?

components/demo/demo-snippet.js
```diff
@@ -68,6 +74,23 @@ class DemoSnippet extends LitElement {
 				position: sticky;
 				top: 0;
 			}
+			d2l-dropdown.settings-dropdown {
+				background-color: white;
+				border-radius: 6px;
+				box-shadow: 0 0 0 1px var(--d2l-color-celestine-minus-1);
+				position: fixed;
+				right: 1rem;
```
dbatiste, line 82: I suspect we might want `left: 1rem; right: auto;` when in RTL mode. For many of our core demos, the element of interest would be in the proximity of the opener in RTL mode, and in such case having it flipped might be desirable. Having said, any demo where we are toggling to fullscreen mode probably has lots more elements of interest, and there would always be a chance of those elements being in the same proximity as the dropdown opener.

Note that the opener positioning doesn't work too well with the toolbar of the list (see list demo page). It appears to be a stacking issue... might be able to just tweak your `z-index` to resolve it.
image

components/demo/demo-snippet.js
bearfriend: Whoops, yes on both counts.

components/demo/demo-snippet.js
bearfriend: It's meant to be `&&` so that it only happens when you have opted for `fullWidth` and have enabled `_fullscreen` mode, but we could do it in `_fullscreen` regardless of `fullWidth`.

components/demo/demo-snippet.js
bearfriend: Yeah I'm open to whatever positioning makes most sense. I'll look at the stacking issue.

components/button/button-icon.js
bearfriend: I tried removing this earlier today to reproduce and I wasn't able to. Seems like maybe it was just a coincidence that it worked after adding this. I'm fine to remove it until/unless we can reproduce. It only happened in Polypane.

dbatiste: Sounds good.

components/demo/demo-snippet.js
dbatiste: I don't feel strongly either way. I've only used the fullscreen option a little, and so far it hasn't been an issue having the settings at the side. 🤷‍♂️

components/demo/demo-snippet.js
```diff
@@ -40,14 +43,18 @@ class DemoSnippet extends LitElement {
...
 				inset: 0;
-				overflow: auto;
-				position: fixed;
+				position: absolute;
 				z-index: 2;
 			}
 			.d2l-demo-snippet-demo {
 				flex: 1 1 auto;
 				position: relative;
+				translate: 0;
```
bearfriend, line 54: This should prevent the demos from leaking their stacking contexts outside this container.

components/demo/demo-snippet.js
bearfriend: And for more context, the current behavior was working mostly fine but was not giving us an accurate representation of when our breakpoints triggered.

components/demo/demo-snippet.js
dbatiste: Suggest adding a comment, otherwise we might forget why we added this and might be inclined to think is can be removed.
```suggestion
				translate: 0; /* create stacking context to prevent demos from leaking outside their containers */
```

components/demo/demo-snippet.js
dbatiste: Sounds good. If we find the position in RTL is problematic more often then not, we can add the RTL rule later.

dbatiste (approved):

PR #3002, opened 2022-11-18
Refactor MathJax helpers to allow consumers to typeset math directly, including in different windows

grant-cleary:
We want the HTML editor to be able to use MathJax to render equations within the editor window directly. This requires embedding MathJax within TinyMCE's iframe, so we need to generalize the MathJax helpers to allow external sources to manage typesetting, as well as to specify a different target window for typesetting.

### DIFF
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -30,9 +30,6 @@ const mathjaxFontMappings = new Map([
 	['MJXTEX-VB', 'MathJax_Vector-Bold']
 ]);
 
-let mathJaxLoaded;
-let renderingPromise = Promise.resolve();
-
 class HtmlBlockMathRenderer {
 
 	get contextAttributes() {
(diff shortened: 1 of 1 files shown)


helpers/mathjax.js
```diff
@@ -229,9 +206,42 @@ export function loadMathJax(mathJaxConfig) {
...
 	});
 
-	return mathJaxLoaded;
+	// If we're using deferred rendering, we need to create a document structure
+	// within the element so MathJax can appropriately process math.
+	//if (!options.noDeferredRendering) elem.innerHTML = `<mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc>`;
 
+	await win.MathJax.startup.promise;
+	if (!win.renderingPromise) win.renderingPromise = Promise.resolve();
+	win.renderingPromise = win.renderingPromise.then(() => win.MathJax.typesetShadow(elem.getRootNode(), elem));
```
grant-cleary, line 245: The rendering queue doesn't technically need to be window-specific, but it probably should be. MathJax instances in different windows can't interfere with each other.

grant-cleary:
Interesting. The demo page works fine locally. I'll have to poke at the visual diff tests to see what's going on.

grant-cleary:
Ah shoot. Something breaks in the demo that doesn't break in my local instance. I made some recent tweaks and probably broke a thing.

helpers/mathjax.js
```diff
@@ -229,9 +206,42 @@ export function loadMathJax(mathJaxConfig) {
...
+	// A duplicate that explains our exact issue: https://github.com/mathjax/MathJax/issues/2495
+	elem.querySelectorAll('mspace[linebreak="newline"]').forEach(elm => {
+		elm.style.display = 'block';
+		elm.style.height = '0.5rem';
 	});
 
-	return mathJaxLoaded;
+	// If we're using deferred rendering, we need to create a document structure
+	// within the element so MathJax can appropriately process math.
+	//if (!options.noDeferredRendering) elem.innerHTML = `<mjx-doc><mjx-head></mjx-head><mjx-body>${elem.innerHTML}</mjx-body></mjx-doc>`;
```
grant-cleary, line 241: I'm dumb. Visual diff tests are failing because I forgot to uncomment this line. 🤦

helpers/mathjax.js
```diff
@@ -83,9 +59,10 @@ export function createHtmlBlockRenderer() {
 
 export function loadMathJax(mathJaxConfig) {
 
-	if (mathJaxLoaded) return mathJaxLoaded;
+	const win = (mathJaxConfig && mathJaxConfig.window) || window;
+	if (win.mathJaxLoaded) return win.mathJaxLoaded;
```
dbatiste, line 63: Nit: usually we try to avoid putting something non-D2L specific in the global namespace. FWIW, [we do this for MathJax in BSI](https://github.com/Brightspace/brightspace-integration/blob/main/web-components/mathjax.js#L4). I would at least prefix the variable with d2l to avoid chances of a collision with MathJax itself.

grant-cleary: Yeah, I can prefix this. Or I can create `win.D2L` and use that structure. I'm fine either way.

grant-cleary: I opted for the latter for consistency.

dbatiste (approved):

PR #3006, opened 2022-11-21
feat: Add static build capability

eKoopmans:
This PR adds `npm run build-static`, which generates a static build of core in the `build/` directory.

Details:

- I named it `build-static` instead of `build:static` because I didn't want to include it as a step in `npm run build`
    - rationale: it's not fast (~7 seconds locally), and isn't needed in all cases where we need to build other assets
    - it **should only** ever be run after `build` has been run - the other build assets need to exist
    - I'm happy to change the name and/or include it in `npm run build` if we want
- most of this PR's changes are in the new `rolup.config.js` file
- there will be a separate PR to actually add the CI step of build-and-publish to S3
- the PR also includes 4 fixes to previously broken imports

Rally: https://rally1.rallydev.com/#/?detail=/userstory/670846094437&fdp=true

### Build directory and file size (5.67 MB):
image

### `build/node_modules` (only copies what's required):
image

### Generated code comparison (using `object-property-list.js`):
image


rollup.config.js
```diff
@@ -0,0 +1,34 @@
+import copy from 'rollup-plugin-copy';
+import del from 'rollup-plugin-delete';
+import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
+import glob from 'glob-all';
+import resolve from '@rollup/plugin-node-resolve';
+
+const jsGlob = [
+	'@(components|controllers|directives|helpers|mixins|templates)/**/*.js',
+	'!**/*@(test|axe|visual-diff).js',
+];
```
eKoopmans, line 10: This selects all JS files in our "main" directories, and filters out any `test.js`, `axe.js`, and `visual-diff.js`.

I confirmed that there are no cases where any of the above files are being imported in HTML.

rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+const nonJsGlob = [
+	'@(components|controllers|directives|helpers|mixins|templates)/**/*.*',
+	'*.*',
+	'!**/*.@(js|md|json)',
+	'!**/screenshots/**/*',
+];
```
eKoopmans, line 16: This selects all non-JS files in our "main" directories, also filtering out Markdown, JSON, and screenshots. This saves us a lot of MB (from 50.7MB down to 5.67MB!).

I confirmed that there are no cases where any of the above files are being imported in HTML or JS.

rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default {
+	input: glob.sync(jsGlob),
+	output: { dir: 'build', format: 'es', preserveModules: true },
```
eKoopmans, line 20: `preserveModules` keeps the generated files looking very similar to their originals, and keeps all `node_modules` imported files in their original folder.

The [docs](https://rollupjs.org/guide/en/#outputpreservemodules) recommend not doing this, and explicitly naming all inputs ("entry points") - but I found that by doing **both**, we get the best of both worlds 😄.

mixins/test/focus-mixin.test.js
```diff
@@ -1,7 +1,7 @@
 
 import { defineCE, expect, fixture, html, oneEvent } from '@open-wc/testing';
 import { FocusMixin } from '../focus-mixin.js';
-import { getComposedActiveElement } from '../../../helpers/focus.js';
+import { getComposedActiveElement } from '../../helpers/focus.js';
```
dlockhart, line 4: Was `web-dev-server` or `web-test-runner` just kind of... fixing these for us?

rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+	external: ['puppeteer', '@brightspace-ui/visual-diff', '@open-wc/testing', 'sinon'],
+	plugins: [
+		del({ targets: 'build' }),
+		copy({
+			targets: [{
+				src: nonJsGlob,
+				dest: 'build',
+				rename: (_name, _extension, fullpath) => fullpath,
```
eKoopmans, line 28: This just makes sure the copied targets keep their original directory structure - there's a `flatten: false` option that should do this too, but it doesn't quite work the same (it always removes the top-level directory regardless...).

package.json
```diff
@@ -13,10 +13,11 @@
     "build:illustrations": "node ./cli/empty-state-illustration-generator.js",
     "build:sass": "node-sass --output-style expanded ./test/sass.scss > ./test/sass.output.css",
     "build": "npm run build:clean && npm run build:icons && npm run build:illustrations && npm run build:sass",
+    "build-static": "rollup -c",
     "lint": "npm run lint:eslint && npm run lint:style && npm run lint:lit",
-    "lint:eslint": "eslint . --ext .js,.html",
+    "lint:eslint": "eslint . --ext .js,.html --ignore-pattern build",
```
dlockhart, line 18: Would the `--ignore-path .gitignore` flag make sense here?

package.json
```diff
@@ -46,16 +47,22 @@
...
     "chalk": "^5",
     "eslint": "^8",
     "eslint-config-brightspace": "^0.19",
+    "glob-all": "^3",
     "lit-analyzer": "^1",
     "messageformat-validator": "^2",
     "node-sass": "^7",
+    "rollup": "^2",
+    "rollup-plugin-copy": "^3",
+    "rollup-plugin-delete": "^2",
```
dlockhart, line 65: Kind of unfortunate that we need to install these in dev and in all the other jobs that _don't_ need them... can't think of a great alternative though other than manually installing them just in the eventual workflow that runs this.

rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+	'@(components|controllers|directives|helpers|mixins|templates)/**/*.*',
+	'*.*',
+	'!**/*.@(js|md|json)',
+	'!**/screenshots/**/*',
+];
+
+export default {
+	input: glob.sync(jsGlob),
+	output: { dir: 'build', format: 'es', preserveModules: true },
+	external: ['puppeteer', '@brightspace-ui/visual-diff', '@open-wc/testing', 'sinon'],
```
eKoopmans, line 21: By marking these as external, rollup just leaves them alone - which means that if we tried to actually use a file with one of these imports, it would break.

Technically, only `'@brightspace-ui/visual-diff'` now needs to be marked external - it's used in `tooltip-helper.js`, `dialog-helper.js`, and `dropdown-helper.js`. None of those helpers are imported by anything that'll display in a webpage - they're only used as helpers for `visual-diff.js` files.

The other externals here were from before I excluded `.test.js` etc. But it probably doesn't hurt to keep them in.

mixins/test/focus-mixin.test.js
eKoopmans: Not really, it's a browser-level thing. When you try to `..` above the root it does nothing, so you could throw as many `../../../../` in there as you want, it'll always just end up at the root.

When doing rollup, we're not serving the files on the web, so root is no longer project root, it's your file-system's root - which is obviously deeper.

package.json
eKoopmans: Sounds good to me! And yeah it works on my end... the extra things it'll exclude are:

- `generated/` (only relevant if you lint after building...)
- `test/sass.output.css` (same as above)

Now, in our current pipelines we do in fact lint after building, so I guess the above are relevant.

package.json
eKoopmans: Yeah... `rollup` and `@rollup/plugin-node-resolve` were already being indirectly included, though `@rollup/plugin-node-resolve` now ends up being installed twice with different versions (v13 and v15).

mixins/test/focus-mixin.test.js
dlockhart: Ah, makes sense!

PR #3009, opened 2022-11-23
demo page language selection

josephsaffron:
@svanherk do you think I even need the property?


svanherk:
> @svanherk do you think I even need the property?

I don't _think_ you need from it a tell-Lit-to-update-the-demo POV, because `intl` watches for any changes to the `lang` and the `localize-mixin` will then update the `__resources`.  That should cause every component using the `LocalizeMixin` to update, so you might be able to do what `rtl` is doing instead.

Why only French?  Is the goal to just smoke test that langterms update properly everywhere and aren't being incorrectly stored, or to test how different languages affect layout?

josephsaffron:
Yep, that's the reason... just as a smoke test to show another language. If we had had that in our demo, we could have caught that we were relying on some hypermedia terms that will always be in english, that we needed to have translated.

josephsaffron:
> I don't _think_ you need from it a tell-Lit-to-update-the-demo POV, because `intl` watches for any changes to the `lang` and the `localize-mixin` will then update the `__resources`. That should cause every component using the `LocalizeMixin` to update, so you might be able to do what `rtl` is doing instead.

How about from a clear code perspective?

components/demo/demo-snippet.js
```diff
@@ -108,6 +110,7 @@ class DemoSnippet extends LitElement {
 				<div class="d2l-demo-snippet-settings">
 					<d2l-switch text="RTL" ?on="${dirAttr === 'rtl'}" @change="${this._handleDirChange}"></d2l-switch><br />
 					<d2l-switch text="Fullscreen" ?on="${this._fullscreen}" @change="${this._handleFullscreenChange}"></d2l-switch><br />
+					<d2l-switch text="French" ?on="${this._isFrench}" @change="${this._handleLanguageChange}"></d2l-switch><br />
```
dlockhart, line 113: Instead of a switch... what about a dropdown that contains all the languages we support? You could [grab that list from here](https://github.com/BrightspaceUI/intl/blob/main/lib/common.js#L3).

components/demo/demo-snippet.js
```diff
@@ -13,6 +13,7 @@ class DemoSnippet extends LitElement {
 			overflowHidden: { type: Boolean, reflect: true, attribute: 'overflow-hidden' },
 			_code: { type: String },
 			_dir: { type: String, attribute: false },
+			_isFrench: { type: Boolean, reflect: false },
```
dlockhart, line 16: This was all written before Lit 2, but all of these `_` properties could use the new state stuff since they don't need to be "public" properties.

```suggestion
			_isFrench: { state: true },
```

components/demo/demo-snippet.js
dbatiste: The placement of this function doesn't seem quite right to me. The code attached to the switch acts at the document level (good), but there is a switch for each snippet and their state isn't sync'd, which results in some kinda semi-broken behaviour depending on the switch that is subsequently clicked and the current state of the document. See screenshot. 

I've been thinking lately that it would be good to have a page-level set of controls (RTL, language) that operates on the whole document vs snippet-level controls (fullscreen). Maybe part of `d2l-demo-page`. There is an edge case where the RTL switch doesn't work well as coded - happy to go into the details for anyone interested, but I think pulling out the RTL switch is the proper solution (though I have another sol'n in my back pocket). Don't need to address RTL now, but creating that page-level control container may put us on the right path.

image

dlockhart: I like the idea of a page-level thing for RTL and the language even better. Technically if we had a language switcher, choosing Arabic would come with RTL so we might get away with dropping that entirely.

dlockhart: I'll put up a draft that attempts a page-level dropdown.

josephsaffron: HMm..... I'll do my best to get to it.. there hasn't been a lot of time for me to code this sprint. =p

dlockhart: Done: https://github.com/BrightspaceUI/core/pull/3015

dbatiste (approved):

PR #3011, opened 2022-11-23
feat: Publish to S3 for PRs and main branch

eKoopmans:
This PR adds a publish step to the CI and release workflows. This builds and publishes a static version of core to S3.

I've created two separate actions for this purpose:
- publish-to-s3
- comment-on-pr

I've designed both actions to be theoretically reusable in the future, but keeping them internal for now. No readmes for them at the moment as they're just internal.

For publishing to S3, I made these decisions:
- no caching, so that changes can be seen immediately without refreshing the cache
    - though I'd be open to putting say a default 2-minute cache on everything if we prefer
- yes to the default compression

Rally: https://rally1.rallydev.com/#/?detail=/userstory/670846679185&fdp=true

### DIFF
--- a/.github/actions/comment-on-pr/action.yml
+++ b/.github/actions/comment-on-pr/action.yml
@@ -1,0 +1,41 @@
+name: Comment on PR
+description: Adds a comment on the current PR
+
+inputs:
+  MESSAGE:
+    description: The message to display
+    required: true
+  ONLY_POST_ONCE:
+    description: Check whether the current message has already been posted before posting again
+    default: false
+    required: false
+  UNIQUE_KEY:
+    description: A unique key attached invisibly to the comment, for use with the ONLY_POST_ONCE option
+    default: gh-actions--comment-on-pr
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Find Previous Comment
+      if: ${{ inputs.ONLY_POST_ONCE != 'false' }}
+      uses: Brightspace/third-party-actions@peter-evans/find-comment
+      id: fc
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        body-includes: ${{ inputs.UNIQUE_KEY }}
+
+    - name: PR Deployment URLs
+      if: ${{ inputs.ONLY_POST_ONCE == 'false' || steps.fc.outputs.comment-id == '' }}
+      uses: Brightspace/third-party-actions@actions/github-script
+      env:
+        MESSAGE: ${{ inputs.MESSAGE }}
+        UNIQUE_KEY: ${{ inputs.UNIQUE_KEY }}
+      with:
+        script: |
+          github.issues.createComment({
+            issue_number: context.issue.number,
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: `<!-- ${process.env.UNIQUE_KEY} -->\n\n${process.env.MESSAGE}`
+          })
--- a/.github/actions/publish-to-s3/action.yml
+++ b/.github/actions/publish-to-s3/action.yml
@@ -1,0 +1,97 @@
+name: Publish to S3
+description: Publishes the specified directory using aws s3 sync
+
+inputs:
+  AWS_REGION:
+    description: The region of your AWS bucket
+    default: ca-central-1
+    required: false
+  AWS_ROLE:
+    description: The AWS role to assume. This role must have write permission for your bucket
+    required: true
+  BUCKET_PATH:
+    description: The full path of your bucket, including subdirectories, to which to publish
+    required: true
+  BUILD_DIRECTORY:
+    description: The directory within your repo to publish to S3 (e.g. "./build/")
+    required: true
+  CACHE:
+    description: An optional comma-separated list of all file extensions you wish to have cached for 1 year (e.g. "js,css")
+    default: ""
+    required: false
+  CACHE_DEFAULT:
+    description: An optional default caching policy to apply to all files (e.g. "--cache-control 'max-age=120'")
+    default: ""
+    required: false
+  COMPRESS:
+    description: Whether to compress eligible files (if "true", will compress all of html/css/js/json/svg/xml)
+    default: false
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Parse files
+      id: parse-files
+      env:
+        CACHE: ${{ inputs.CACHE }}
+        COMPRESS: ${{ inputs.COMPRESS }}
+      run: |
+        typesToCompress=()
+        [[ $COMPRESS != 'false' ]] && typesToCompress=("html" "css" "js" "json" "svg" "xml")
+        typesToCache=(${CACHE//,/ })
+
+        compressAndCache=""
+        compressOnly=""
+        cacheOnly=""
+        neither=""
+
+        for type in ${typesToCompress[@]}; do
+          if [[ " ${typesToCache[*]} " =~ " ${type} " ]]; then
+            compressAndCache+="--include '*.${type}' "
+          else
+            compressOnly+="--include '*.${type}' "
+          fi
+          neither+="--exclude '*.${type}' "
+        done
+
+        for type in ${typesToCache[@]}; do
+          if [[ ! " ${typesToCompress[*]} " =~ " ${type} " ]]; then
+            cacheOnly+="--include '*.${type}' "
+            neither+="--exclude '*.${type}' "
+          fi
+        done
+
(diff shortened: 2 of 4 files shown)


.github/actions/publish-to-s3/action.yml
```diff
@@ -0,0 +1,97 @@
...
+    description: The region of your AWS bucket
+    default: ca-central-1
+    required: false
+  AWS_ROLE:
+    description: The AWS role to assume. This role must have write permission for your bucket
+    required: true
+  BUCKET_PATH:
+    description: The full path of your bucket, including subdirectories, to which to publish
+    required: true
+  BUILD_DIRECTORY:
```
dlockhart, line 15: Nit: maybe `PUBLISH_DIRECTORY` instead of "build", which is more specific to this use case?

.github/actions/publish-to-s3/action.yml
```diff
@@ -0,0 +1,107 @@
...
+    default: ""
+    required: false
+  CACHE_DEFAULT:
+    description: An optional default caching policy to apply to all files (e.g. "--cache-control 'max-age=120'")
+    default: ""
+    required: false
+  COMPRESS:
+    description: Whether to compress eligible files (if "true", will compress all of html/css/js/json/svg/xml)
+    default: false
+    required: false
```
dlockhart, line 29: I can't think of a reason we wouldn't want to compress... I think I'd personally not even make this an option.

.github/actions/publish-to-s3/action.yml
```diff
@@ -0,0 +1,107 @@
...
+        neither=""
+
+        for type in ${typesToCompress[@]}; do
+          if [[ " ${typesToCache[*]} " =~ " ${type} " ]]; then
+            compressAndCache+="--include '*.${type}' "
+          else
+            compressOnly+="--include '*.${type}' "
+          fi
+          neither+="--exclude '*.${type}' "
+        done
```
dlockhart, line 66: This is super cool! If we're not going to use caching out of the gate though (which I agree it doesn't make sense here), is this maybe overly complex? I'd be sad to lose it though when/if this becomes not-internal, so maybe it's fine.

eKoopmans:
> https://pr-5-brightspace-ui-core.d2l.dev/

Oops, ended up with a hard-coded value in my testing, will fix.

> https://pr-3011-brightspace-ui-core.d2l.dev/

... ☝️ index.html is looking pretty messed up 😄. Seems like it's not being properly decompressed?

.github/actions/publish-to-s3/action.yml
eKoopmans: Okay, I like having it as an option but I'm okay to remove it 👍.

.github/workflows/ci.yml
```diff
@@ -43,3 +43,55 @@ jobs:
...
+
+      - name: Assume role
+        uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
+          role-to-assume: arn:aws:iam::037018655140:role/brightspace-ui-core.d2l.dev-githubactions-access
+          role-duration-seconds: 3600
+          aws-region: ca-central-1
```
eKoopmans, line 72: I originally had this in `publish-to-s3` directly, but it wasn't allowing me to access `secrets`. Does that sound right? Would be nice to have it in the other action, but might not be possible.

.github/actions/publish-to-s3/action.yml
```diff
@@ -83,8 +83,9 @@ runs:
         FILES_COMPRESS_ONLY: ${{ steps.parse-files.outputs.COMPRESS_ONLY }}
         FILES_DEFAULT: ${{ steps.parse-files.outputs.DEFAULT }}
       run: |
-        [[ $FILES_COMPRESS_AND_CACHE ]] && aws s3 sync --delete "$PUBLISH_DIRECTORY" "$BUCKET_PATH" $COMPRESS_ENCODING $CACHE_POLICY --exclude '*.*' $FILES_COMPRESS_AND_CACHE
-        [[ $FILES_COMPRESS_ONLY ]] && aws s3 sync --delete "$PUBLISH_DIRECTORY" "$BUCKET_PATH" $COMPRESS_ENCODING $CACHE_DEFAULT --exclude '*.*' $FILES_COMPRESS_ONLY
-        [[ $FILES_CACHE_ONLY ]] && aws s3 sync --delete "$PUBLISH_DIRECTORY" "$BUCKET_PATH" $CACHE_POLICY --exclude '*.*' $FILES_CACHE_ONLY
-        aws s3 sync --delete "$PUBLISH_DIRECTORY" "$BUCKET_PATH" $CACHE_DEFAULT $FILES_DEFAULT
+        aws s3 sync --delete "./build/" "s3://brightspace-ui-core.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'br'
```
eKoopmans, line 86: This fixes the br encoding issue - so I'll need to track down what's wrong in the logic I commented out.

.github/workflows/ci.yml
svanherk: Yup this is right - the secrets would need to be passed into the actions as inputs, or as `env` variables

.github/actions/publish-to-s3/action.yml
eKoopmans: Removed.

.github/actions/publish-to-s3/action.yml
eKoopmans: Keeping it! Spent all that time finally getting it working 😆. Should be nice and general for the future.

eKoopmans:
Okay, this thing's back up and running and ready for review! The issues with the publish step have been resolved - long story short, Bash is very particular when it comes to quotes 😄.

dlockhart (approved):
Looks awesome!

.github/actions/comment-on-pr/action.yml
```diff
@@ -0,0 +1,41 @@
...
+  MESSAGE:
+    description: The message to display
+    required: true
+  ONLY_POST_ONCE:
+    description: Check whether the current message has already been posted before posting again
+    default: false
+    required: false
+  UNIQUE_KEY:
+    description: A unique key attached invisibly to the comment, for use with the ONLY_POST_ONCE option
+    default: gh-actions--comment-on-pr
```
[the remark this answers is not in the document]
eKoopmans: 😆 Yes, that was deliberate - I was going for something akin to BEM naming. I'm *very* unattached to the name, so if you've got a preference I'm all ears. I just wanted something that was unlikely to turn up in any other comments 🤷.

.github/workflows/ci.yml
```diff
@@ -43,3 +43,54 @@ jobs:
...
+        uses: ./.github/actions/comment-on-pr/
+        with:
+          MESSAGE: |
+            Thanks for the PR! 🎉
+
+            We've deployed an automatic preview for this PR - you can see your changes here:
+
+            | URL |
+            |---|
+            | https://pr-${{ github.event.number }}-brightspace-ui-core.d2l.dev/ |
```
svanherk, line 91: Mostly just curious, was this inspired by somewhere that could have multiple urls listed?  Or was the table just to look nice?

eKoopmans: Yes to both 😄. Blog [uses a table](https://github.com/Brightspace/blog.d2l.dev/pull/699#issuecomment-1276661516) with only one row, for "internal", but with the idea that there'll be a public URL in the future too. I kept the table formatting cause it makes it stand out.

.github/actions/comment-on-pr/action.yml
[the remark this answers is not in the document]
svanherk: I'm also not too worried about the name at all lol, it just triggered this uncontrollable part of my brain that's like "Duplicated character! Potential spelling mistake!" If you wanted something _really_ unlikely you could also append the PR number to the end down below when you write/look for it, but I really don't think it's going to come up in real life regardless.

svanherk (approved):
SO excited for this!  I already thought twice in the last three days "I wish I could just access the demo pages quickly" - although one of those times was for a repo outside `core` lol... maybe someday!

svanherk:
Just thought of this now - should we add the main https://brightspace-ui-core.d2l.dev/ route to the `core`'s README?

eKoopmans:
Sure, sounds good - I'll put up a PR. There's a small thing I forgot to remove in this PR, so I was going to put up a fix anyway 😄.

PR #3015, opened 2022-11-24
feat: add demo page-level language selection

dlockhart:
Builds on Joseph's "language toggle" but moves it up to be a page-level dropdown.

I've added logic here that sets the document's direction based on the language chosen, which removes the need for a separate "RTL" switch. That switch was pretty buggy anyway. This did require a modification to the `RtlMixin` so that it listens for changes to the document language and re-calculates based on that.

English:
Screen Shot 2022-11-25 at 2 11 36 PM

Arabic:
Screen Shot 2022-11-25 at 2 11 47 PM


components/demo/demo-page.js
```diff
@@ -86,6 +109,13 @@ class DemoPage extends LitElement {
 			this._noScroll = true;
 		}
 	}
+
+	_handleLanguageChange(e) {
+		const newLanguageCode = e.target[e.target.selectedIndex].value;
+		document.documentElement.dir = newLanguageCode === 'ar' ? 'rtl' : 'ltr';
```
dlockhart, line 115: If we like this approach, I'll make a separate change to the Intl library so that when you set the language to Arabic, it automatically sets the document's direction.

components/demo/demo-page.js
```diff
@@ -67,9 +80,19 @@ class DemoPage extends LitElement {
 		const classes = {
 			'no-scroll': this._noScroll
 		};
+		const selectedLanguageCode = getDocumentLocaleSettings().language;
+		const languageOptions = supportedBaseLocales.map((code) => {
```
dlockhart, line 84: This outputs the not super friendly language code... I could also add something to the Intl library that exposes a friendly list of locale names. If we like this approach, I can do that.

mixins/rtl-mixin.js
```diff
@@ -13,9 +14,25 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
...
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this._localeSettings.removeChangeListener(this._handleLanguageChange);
+	}
+
+	_handleLanguageChange() {
 		const dir = document.documentElement.getAttribute('dir');
 		// avoid reflecting "ltr" for better performance
-		if (dir && dir !== 'ltr') {
+		if (dir && (dir !== 'ltr' || this.dir === 'rtl')) {
```
dlockhart, line 35: This logic needed slight tweaking to handle a change after the initial value was set. If things go from `rtl` -> `ltr`, we DO need to reflect that change. I could also set it back to `undefined` -- what do folks think?

dbatiste: Is it valid to assign `this.dir = undefined`? When I log `dir` for elements without an explicit `dir` assignment, the browser returns an empty string, though [Mozilla says](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dir) the property's values can be `ltr`, `rtl`, and `auto`.

mixins/rtl-mixin.js
```diff
@@ -13,9 +14,25 @@ export const RtlMixin = dedupeMixin(superclass => class extends superclass {
...
 	constructor() {
 		super();
+		this._localeSettings = getDocumentLocaleSettings();
+		this._handleLanguageChange = this._handleLanguageChange.bind(this);
+		this._handleLanguageChange();
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this._localeSettings.addChangeListener(this._handleLanguageChange);
```
dbatiste, line 24: Would these listeners only get invoked if the language changes after the page loads? i.e. in a normal page where the language doesn't change, would `_handleLanguageChange` only get called once, from the `constructor`?

I'm not sure how concerned we should be about the number of handlers wired up. I can imagine pages easily wiring up hundreds due to the number of components, perhaps even more than a thousand. If the change handler is only needed for pages such as the demo page where the language can change, we could do this wire-up conditionally based on some global flag that the demo page sets. Ideally we can just wire it up and not worry about it though.

dbatiste (commented):
I like it! Just a couple comments/questions.

mixins/rtl-mixin.js
dlockhart: Correct, these only get called if the language changes after the page loads. That currently only happens in the demo page (and maybe a few of our tests), although it's not impossible to foresee a scenario where we may allow the user to switch languages without reloading the whole app. Pretty edge case-y though.

I'm not especially worried about the add/remove of the listeners... it's just [appending to an array and then splicing it](https://github.com/BrightspaceUI/intl/blob/main/lib/common.js#L129. Agreed though that we could do some kind of global flag if it becomes an issue.

What I'm hoping though is that our dependence on `RtlMixin` will actually start to drop off now that we're in the clear to start using CSS logical properties. I'm guessing 95% of our uses are related to that.

mixins/rtl-mixin.js
dlockhart: Hmm, interesting. I just checked, and even before these changes, any component using `RtlMixin` was having its `dir` set to `undefined` by default. I'm guessing Lit's doing that since we're redefining `dir` as a property and aren't initializing it in the constructor.

If we're worried about it being `undefined`, we could _not_ define it as a Lit property (since it's already a native one) and just manually call `requestUpdate` when we set it.

dbatiste: Cool - I'm happy either way as long as the browser doesn't 🤮 .

components/demo/demo-page.js
dlockhart: Actually on second thought, the Intl library doesn't set the `lang` on the `<html>` element so I don't think this makes sense. It just listens for mutations to those attributes and adapts accordingly.

components/demo/demo-page.js
```diff
@@ -3,8 +3,11 @@ import './code-view.js';
 import '../colors/colors.js';
 import '../typography/typography.js';
 import { css, html, LitElement } from 'lit';
+import { getDocumentLocaleSettings, supportedBaseLocales } from '@brightspace-ui/intl/lib/common.js';
```
dbatiste, line 6: Would you want to use the base locales or the culture-specific ones?

dlockhart: I can do either, but I figured the base locales accomplished the task. Is there any advantage to the culture-specific ones?

dbatiste: I think either should be fine. I guess the only thing would be if someone wants to check the verbiage for a culture specific locale.

dlockhart: Yeah. Ok I'll use the larger set -- need to update my PR in Intl then.

dlockhart: Ok -- updated with the culture-specific ones. Screenshots in description updated as well.

dbatiste (approved):

components/demo/demo-page.js
```diff
@@ -86,6 +112,13 @@ class DemoPage extends LitElement {
 			this._noScroll = true;
 		}
 	}
+
+	_handleLanguageChange(e) {
+		const newLanguageCode = e.target[e.target.selectedIndex].value;
+		document.documentElement.dir = newLanguageCode === 'ar-sa' ? 'rtl' : 'ltr';
```
eKoopmans, line 118: So now to test RTL visually, we'll need to explicitly select `ar-sa`. Sounds good, might be worth calling out on our Slack or something though 😄.

eKoopmans (approved):

PR #3022, opened 2022-11-25
feat: Add build version and date to demo index

eKoopmans:
This PR adds the build version and date to `index.html` on both static builds and web-dev-server. This was @geurts 's suggestion, and should make it easier to verify which version we're looking at, now that we can have multiple copies.

I wanted to use an existing plugin for the replacements (i.e. [@rollup/plugin-replace](https://www.npmjs.com/package/@rollup/plugin-replace)), but none of the existing tools worked properly on non-JS files and web-dev-server, so I made my own simple one. I kept its signature the same as a proper rollup plugin, even though it's not used that way directly in either build config.

The main effort was the build infrastructure - if we wanted to change how the info is displayed or put it on other pages, that should now be easy.


eKoopmans:
Okay, the date is working properly locally, but of course in CI it's using a different timezone and format.

Any opinions on the best way to go? Unfortunately I can't easily use `helpers/dateTime.js`, because some of its imports are CJS and `web-dev-server.config.js` is picky about that.

Anyway, signing off for now, have a good weekend!

index.html
```diff
@@ -19,6 +19,11 @@
 
 	<h1 class="d2l-heading-2">Demos</h1>
 
+	<ul>
+		<li><b>Build version:</b> __buildVersion__</li>
+		<li><b>Build date:</b> __buildDate__</li>
+	</ul>
```
dlockhart, line 25: Another option here that should let you use an off-the-shelf plugin would be to build a small web component do to this. That way, the replacement would be a in a JS file. It could also probably be a bit smart and if the replacement didn't occur -- like in dev -- it could just plug in some default values. That would remove the need for the duplicate plugin and the `web-dev-sever.config`. 🤷

web-dev-server.config.js
```diff
@@ -0,0 +1,25 @@
...
+
+const webDevReplace = config => {
+	const replaceInstance = replace(config);
+	return {
+		name: 'replace',
+		transform: context => replaceInstance.transform(context.body, context.path)?.code ?? context.body,
+	};
+};
+
+export default {
```
dlockhart, line 15: If we are going to introduce a `web-dev-server` config file, we could move the ` --node-resolve --watch --open` flags from `package.json` down here.

rollup/rollup.config.js
```diff
@@ -2,7 +2,11 @@ import copy from 'rollup-plugin-copy';
 import del from 'rollup-plugin-delete';
 import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
 import glob from 'glob-all';
+import replace from './rollup-plugin-replace-simple.js';
 import resolve from '@rollup/plugin-node-resolve';
+import { version } from '../package.json';
+
+const buildDate = new Intl.DateTimeFormat().format(new Date());
```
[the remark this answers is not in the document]
eKoopmans: Yep that works, although it includes the time (maybe a good thing?) and uses mm/dd/YYYY (not a good thing IMO): `11/29/2022, 2:55:57 PM`. Using `en-CA` might be more appropriate: `2022-11-29, 2:56:06 p.m.`.

If we want to stick with just YYYY-MM-DD: `Intl.DateTimeFormat('en-CA', { timeZone: 'America/Toronto' }).format(new Date())`.

I do also have access to `helpers/dateTime.js`, not sure if I should be using that or not - I notice `getToday()` has no way to override the timezone, so might just be easier not to use them...

web-dev-server.config.js
eKoopmans: Good call, looks like I'll go with no `web-dev-server` config though.

index.html
dlockhart: Yeah that works, or could even just be defined inline in this file inside a `<script type="module">`.

eKoopmans: I considered that, but inline means it's not in a JS file, which means `@rollup/plugin-replace` won't replace it 😂.

rollup/rollup.config.js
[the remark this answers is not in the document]
dlockhart: The `YYYY-MM-DD` works for me. If you do want to use something else, I'd use [our Intl stuff](https://github.com/BrightspaceUI/intl/#datetime-formatting). It's just expecting a `Date` object in the local timezone though.

index.html
dlockhart: Hahaha, right.

rollup/rollup.config.js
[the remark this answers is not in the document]
eKoopmans: Seems like some (most? all?) of our intl date-time utils could be replaced with `Intl.DateTimeFormat`:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

Was there a decision at some point not to use that?

dlockhart (approved):

rollup/rollup.config.js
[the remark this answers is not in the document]
dlockhart: Oh believe me, I'd love to. The main reason is that we allow our end users and administrators to customize everything about their locales. So an admin can set up the "en-CA" locale, but make it use a 24 hour clock or comma as a decimal separator. Then on top of that, an end-user can override that with whatever settings they want.

If we could ever reign that in and just say "if you choose en-CA, you get these settings and that's it", then we could use all the built-in stuff.

PR #3025, opened 2022-11-28
feat: Collapsible Panel Component

jameswklassen:
## Description

Implementation of the new `d2l-collapsible-panel` component.

## Design
The [collapsible panel design spec](https://www.figma.com/file/5M7vgvmRKOIE7Bp9cidx7K/%F0%9F%8D%B0-Collapsible-Panel-%2F-Cake-Tray)

## Checklist
- [x] Implementation
- [x] Design approval
- [x] Visual diff tests
- [x] Accessibility tests
- [x] Unit tests
- [x] Basic README

## Rally
https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fportfolioitem%2Ffeature%2F641742855359


components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+				</div>
+			</div>
+			<div class="d2l-collapsible-panel-divider"></div>
+		`;
+	}
+
+	_renderHeading() {
+		const titleClasses = {
+			'd2l-collapsible-panel-title': true,
+			[`d2l-heading-${this.headingLevel}`]: true,
```
dbatiste, line 308: Should this be using `headingStyle` instead of `headingLevel`?

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+			e.preventDefault();
+		}
+	}
+
+	_renderHeader() {
+		const icon = 'tier1:arrow-expand';
+		// const icon = `tier1:arrow-${this.expanded ? 'collapse' : 'expand'}`;
+
+		return html`
+			<div class="d2l-collapsible-panel-header" @click="${this._handleHeaderClick}" @keydown="${this._onKeyDown}" tabindex="0">
```
dbatiste, line 289: I think we'll want to be careful of a11y here. This `div` would be focusable but would not have any semantic meaning. But having a `button` element that contains other focusable and interactive elements can be problematic I think. 

So as mentioned in discussion, I may be inclined to have an offscreen button that receives focus and the label. Then you could just use a `click` handler on it... no need for special `keydown` handler, you can manage the focus order, and the focusable elements aren't nested. You'd have to listen for `focus`/`blur` to in turn render a class in this component for the focus state styling (blue ring).

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+			 * @default "default"
+			 */
+			type: { type: String, reflect: true },
+			/**
+			 * Whether or not the content should extend the full width.
+			 * This is only valid when using the "inline" panel style.
+			 * @type {boolean}
+			 */
+			fullWidth: { attribute: 'full-width', type: Boolean, reflect: true },
+			_hasSummary: { type: Boolean, reflect: true },
```
dbatiste, line 68: This isn't a required change per se... but reflecting these private properties is kinda leaky. In effect, it is leaking them as attributes on the host, which is public. We _do_ have several components that do this, especially when the CSS is targeting the host (ex. `:host([some-attr]) { ... }`) but when the target is an internal element then we usually try to avoid leaking by rendering an internal class, which means the CSS changes from:

```css
:host([some-attr]) .some-elem {
    ...
}
```
to the following:
```css
.some-elem.some-class {
    ...
}
```

With such a change, the property definition can become `_hasSummary: { state: true },` since reflection is no longer necessary.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+	static get properties() {
+		return {
+			/**
+			 * REQUIRED: The title of the panel
+			 * @type {string}
+			 */
+			title: { type: String },
+			/**
+			 * The semantic heading level (h1-h5)
+			 * @type {'1'|'2'|'3'|'4'|'5'}
```
dbatiste, line 35: Are these intended to go to `5`? In the header styling method I see code only checking for `1-4`.

components/collapsible-panel/collapsible-panel-summary-item.js
```diff
@@ -0,0 +1,42 @@
...
+		`];
+	}
+
+	constructor() {
+		super();
+		this.text = '';
+	}
+
+	render() {
+		return html`<p class="d2l-body-small">${this.text}</p>`;
```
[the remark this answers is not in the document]
jameswklassen: Good callout, I still haven't tested this part of the component with a screen reader yet 👍

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+		}
+		if (this.headingLevel === 4) {
+			return html`<h4 class="${classMap(titleClasses)}">${this.title}</h4>`;
+		}
+	}
+
+	_toggleExpand() {
+		const event = `d2l-collapsible-panel-${this.expanded ? 'collapse' : 'expand' }`;
+
+		this.dispatchEvent(new CustomEvent(
```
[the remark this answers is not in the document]
jameswklassen: Good point, I didn't consider the transient state of expanding/collapsing as something we'd have to communicate, but that makes sense.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,344 @@
...
+		`;
+	}
+
+	_renderHeading() {
+		const titleClasses = {
+			'd2l-collapsible-panel-title': true,
+			[`d2l-heading-${this.headingLevel}`]: true,
+		};
+
+		if (!this.headingStyle) {
```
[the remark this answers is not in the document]
jameswklassen: Yeah, I'll have to revisit my logic here 😆. I was initially trying to support scenarios where consumers only provide a level or style, but my logic is definitely flawed.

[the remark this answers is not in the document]
dbatiste: I see. Is there supposed to be default level/style values? Might be worth getting Jeff/Abbie's input here. If there are no defaults then one of the values must be provided, and we'd be required to assume the value based on the other if one of the values is not provided, or make both properties required. I think we can entertain any of these scenarios.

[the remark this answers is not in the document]
jameswklassen: Going visually off the designs, I assumed 3 to be the default style (no idea about semantic level). I don't think we've explicitly discussed this, so I'll start a discussion.

components/collapsible-panel/collapsible-panel.js
jameswklassen: Ah, for some reason I thought core supported 1-5, but looks like we only support 1-4.

components/collapsible-panel/collapsible-panel.js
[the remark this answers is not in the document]
jameswklassen: Confirmed that the default heading level is 3

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+ */
+class CollapsiblePanel extends RtlMixin(LitElement) {
+
+	static get properties() {
+		return {
+			/**
+			 * REQUIRED: The title of the panel
+			 * @type {string}
+			 */
+			title: { type: String },
```
dlockhart, line 31: `title` is actually a reserved HTML attribute -- it's how you get those ugly native tooltips to show up, and I'm actually seeing this show up as a tooltip whenever I hover anywhere on the panel. I don't think that's what you intended here? I'd recommend calling this something like `panel-title` instead.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+				--d2l-collapsible-panel-spacing-inline: 0.9rem;
+				--d2l-collapsible-panel-header-spacing: 0.6rem;
+				--d2l-collapsible-panel-transition-time: 0.2s;
+				--d2l-collapsible-panel-arrow-time: 0.5s;
+				display: block;
+			}
+			.d2l-collapsible-panel {
+				border: 1px solid var(--d2l-color-mica);
+				border-radius: 0.3rem;
+				display: block;
```
dlockhart, line 83: `display: block` shouldn't be needed, since `<div>` are block level by default.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+			}
+			:host([padding=large][type=inline]) .d2l-collapsible-panel {
```
dlockhart, line 85: Nit: since these are strings I'd personally prefer `:host([padding="large"][type="inline"])`. And due to the way variables cascade, I think you can actually drop the `.d2l-collapsible-panel` part and just apply the variable value to the host itself.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+				--d2l-collapsible-panel-spacing-inline: 2rem;
+			}
+			:host(:not([expanded])) .d2l-collapsible-panel {
+				cursor: pointer;
```
dlockhart, line 89: I think this rule isn't necessary since the inner `.d2l-collapsible-panel-header` already applies the cursor pointer.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+			}
+			:host([type=subtle]) .d2l-collapsible-panel {
+				background-color: white;
+				border: none;
+				box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.03);
+			}
+			:host([type=inline]) .d2l-collapsible-panel {
```
dlockhart, line 96: Instead of removing the border and re-adding, could this be accomplished by just removing the borders you don't want instead?

```css
border-left: none;
border-right: none;
```

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+				border-top: 1px solid var(--d2l-color-mica);
+			}
+			:host([heading-style="1"]) {
+				--d2l-collapsible-panel-header-spacing: 1.2rem;
+			}
+			:host([heading-style="4"]) {
+				--d2l-collapsible-panel-header-spacing: 0.3rem;
+			}
+			.d2l-collapsible-panel-header {
+				border-radius: 8px;
```
dlockhart, line 109: I think you have a mismatch between this radius and the radius on the host -- the host is `0.3rem` (`6px` @ `20px` root font size) compared to `8px` here.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+				position: sticky;
+				top: 0;
+			}
+			.d2l-collapsible-panel.focused.scrolled .d2l-collapsible-panel-header {
+				top: 2px;
+			}
+			.d2l-collapsible-panel-title {
+				flex: 1;
+				margin: 0.3rem;
+				margin-inline-start: var(--d2l-collapsible-panel-spacing-inline);
```
dlockhart, line 125: Nice, love that we can start using these!

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+			}
+			:host([type=inline]) .d2l-collapsible-panel-summary,
+			:host([type=inline]) .d2l-collapsible-panel-content {
+				padding-top: 0;
+			}
+			:host([type="inline"]) .d2l-collapsible-panel-summary {
+				margin-top: -0.3rem;
+			}
+			.d2l-collapsible-panel:not(.has-summary) .d2l-collapsible-panel-summary {
+				display: none;
```
dlockhart, line 226: Alternate way to do this would be to put the `hidden` attribute on it when you render it.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+		this.type = 'default';
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+
+		if (!this.expandCollapseLabel) {
+			this.expandCollapseLabel = this.title;
+		}
+
```
dlockhart, line 246: I think it would be cleaner to do both of these computations inside `render()` (or I guess `_renderHeading` in this case). For example:

```javascript
render() {

  const expandCollapseLabel = this.expandCollapseLabel || this.title;
  const headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ?
    this.headingLevel : this.headingStyle;
}
```

That way you're not magically changing the value of `expandCollapseLabel` / `headingStyle` on the consumer. Also the act of changing these is going to cause a re-render, which isn't great.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+			>${this.expandCollapseLabel}</button>
+			<div class="d2l-collapsible-panel" @click="${this._handlePanelClick}">
+				<div class="d2l-collapsible-panel-top-sentinel"></div>
+				${this._renderHeader()}
+				<d2l-expand-collapse-content ?expanded="${this.expanded}">
+					<div class="d2l-collapsible-panel-content">
+						<slot name="content"></slot>
+					</div>
+				</d2l-expand-collapse-content>
+				<d2l-expand-collapse-content ?expanded="${!this.expanded}">
```
dlockhart, line 292: Could this entire part not be rendered if there's no summary elements in the slot? You may need to make `this.hasSummary` into a state property instead of just a CSS class.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
+import '../button/button.js';
+import '../colors/colors.js';
+import '../icons/icon.js';
+import '../icons/icon-custom.js';
```
dlockhart, line 4: I don't think you need `button.js` or `icon.js`.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+				</div>
+			</div>
+			<div class="d2l-collapsible-panel-divider"></div>
+		`;
+	}
+
+	_renderHeading() {
+		const titleClasses = {
+			'd2l-collapsible-panel-title': true,
+			[`d2l-heading-${this.headingStyle}`]: true,
```
dlockhart, line 371: Why is this inside an array?

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+		};
+
+		const headingLevel = parseInt(this.headingLevel);
```
dlockhart, line 374: Do you want to normalize it in case they set it to something `< 1` or `> 4`? Probably similar for `headingStyle`.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,405 @@
...
+		}
+		if (headingLevel === 4) {
+			return html`<h4 class="${classMap(titleClasses)}">${this.title}</h4>`;
+		}
+	}
+
+	_toggleExpand() {
+		this.expanded = !this.expanded;
+
+		if (!this.expanded) {
```
dlockhart, line 392: There's a small bug here. Because `_toggleExpand` is only ever called when you interact with the component, if you programatically add/remove `expanded` from the host this code doesn't run, so the `scrolled` class sticks around.

A better place for all this logic is in `updated()`, which will run any time `expanded` is changed no matter how it was changed. You likely _don't_ want to fire the event though from there, so that part could stay here.

components/collapsible-panel/collapsible-panel-summary-item.js
```diff
@@ -0,0 +1,42 @@
...
+			text: { type: String },
+		};
+	}
+
+	static get styles() {
+		return [bodySmallStyles, css`
+			:host {
+				color: var(--d2l-color-galena);
+				display: block;
+			}
```
dlockhart, line 25: If you override the default `display`, you need to also add:

```css
host([hidden]) {
  display: none;
}
```

[More details](https://daylight.d2l.dev/developing/best-practices/styles/#do%3A-style-the-%22display%22-value-of-the-host).

dlockhart (commented):
Really neat! So far just looked at the components themselves, will do another pass in the tests/docs/etc. later.

components/collapsible-panel/collapsible-panel.js
jameswklassen: arg I always forget about the reserved attributes.....

components/collapsible-panel/collapsible-panel.js
jameswklassen: ahh, good catch on the host! missed this when I refactored the inline spacing to a css variable

components/collapsible-panel/collapsible-panel.js
jameswklassen: @dlockhart If the panel is collapsed, the entire panel is clickable (not just the header). When the panel has no summary, the entire panel is the header, but if a summary exists, the header no longer encompasses the panel.

I could add `.has-summary` here, but I think that makes the intent less clear.

image

components/collapsible-panel/collapsible-panel.js
jameswklassen: Yeah, I initially had `_hasSummary` as a state property, but that meant I couldn't use it in css selectors.

That said, the component has changed a lot since then...maybe it's possible to not need `has-summary` in css 🤔

components/collapsible-panel/collapsible-panel.js
jameswklassen: heh... I'm using a [computed property name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names) here so I can programatically set `d2l-heading-___`.

If that's not something we can use, I could replace it with something like this:

```js
d2l-heading-1: this.headingStyle === 1,
d2l-heading-2: this.headingStyle === 2,
d2l-heading-3: this.headingStyle === 3,
d2l-heading-4: this.headingStyle === 4,
```

components/collapsible-panel/collapsible-panel.js
jameswklassen: Yup, remnants of a past implementation

components/collapsible-panel/collapsible-panel.js
dlockhart: Ah yep, you're correct I forgot about the summaries.

components/collapsible-panel/collapsible-panel.js
dlockhart: James and I chatted outside the review -- this is the correct usage of computed property names so I see no reason to change it. I was just confused as I'd never seen this syntax before. TIL!

components/collapsible-panel/collapsible-panel.js
jameswklassen: Hmmm, so looked into this a bit.

I'm not entirely sure it's possible.... `hasSummary` is set on the `@slotchange` event of `summary`. Which, if we don't render it, will never fire (?) 🤔

dlockhart: Hahaha right. Maybe it could at least get tied to a `hidden` on the `d2l-expand-collapse-content`? That would at least ensure it never takes up any space.

jameswklassen: I tried this -- adding `hidden` alongside `expanded` creates weird issues. Expand-collapse-content does handle this for us though:

https://github.com/BrightspaceUI/core/blob/c72043d2b75c01a7fca6f491d168949f5d92688c/components/expand-collapse/expand-collapse-content.js#L56-L58

dlockhart: Ok cool!

components/collapsible-panel/collapsible-panel.js
```diff
@@ -298,6 +287,21 @@ class CollapsiblePanel extends RtlMixin(LitElement) {
 		`;
 	}
 
+	updated(changedProperties) {
+		if (changedProperties.has('expanded')) {
+			if (!this.expanded) {
+				const element = this.shadowRoot.querySelector('.d2l-collapsible-panel');
+				element.classList.remove('scrolled');
```
dlockhart, line 294: Hmm, this is breaking the "render() is a function of state" rule.

Suggestion: make an `isScrolled` as a state-based property. Set it to `true` when scrolling happens and `false` here, and then set the `scrolled` CSS class on the `div` inside of `render()` based on its value.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -298,6 +287,21 @@ class CollapsiblePanel extends RtlMixin(LitElement) {
...
+			}
+
+			const event = `d2l-collapsible-panel-${this.expanded ? 'expand' : 'collapse' }`;
+
+			this.dispatchEvent(new CustomEvent(
```
dlockhart, line 299: Typically we only want to fire events based on user interaction (i.e. the user clicking to expand or collapse), not based on programatic changes. The way this is currently, `elem.expanded = false;` will fire the event. Probably want to move this back into `_toggleExpand()`?

components/collapsible-panel/collapsible-panel.js
```diff
@@ -331,6 +335,13 @@ class CollapsiblePanel extends RtlMixin(LitElement) {
 		element.classList.add('has-summary');
 	}
 
+	_normalizeHeadingNumber(number) {
```
dlockhart, line 338: Nit: could just be a local function defined at the top of this file and not a member function.

components/collapsible-panel/README.md
```diff
@@ -0,0 +1,87 @@
...
+## Collapsible Panel [d2l-collapsible-panel]
+
+The `d2l-collapsible-panel` element is a container that provides specific layout slots such as `header`, `summary`, `actions`, and a default slot for the expanded content.
+
+<!-- docs: demo live name:d2l-collapsible-panel size:large -->
+```html
+<script type="module">
+	import '@brightspace-ui/core/components/button/button-subtle.js';
+  import '@brightspace-ui/core/components/collapsible-panel/collapsible-panel.js';
+  import '@brightspace-ui/core/components/collapsible-panel/collapsible-panel-summary-item.js';
```
dlockhart, line 35: Nit: indentation

components/collapsible-panel/README.md
```diff
@@ -0,0 +1,87 @@
...
+	import '@brightspace-ui/core/components/inputs/input-search.js';
+	import '@brightspace-ui/core/components/filter/filter.js';
+	import '@brightspace-ui/core/components/filter/filter-dimension-set.js';
+</script>
+
+<d2l-collapsible-panel title="Availability Dates and Conditions">
```
dlockhart, line 41:
```suggestion
<d2l-collapsible-panel panel-title="Availability Dates and Conditions">
```

components/collapsible-panel/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Properties:
+
+| Property | Type | Description |
+|--|--|--|
+| `expanded` | Boolean | Whether or not the panel is expanded |
+| `expand-collapse-label` | String | Label describing the contents of the header (used by screen readers) |
+| `full-width` | Boolean | Whether or not the content should extend to the full width (only valid when using "inline" panel type) |
+| `heading-style` | Number | The heading style to use |
+| `heading-level` | Number | Semantic heading level (h1-h4) |
+| `title` | String | The title of the panel |
```
dlockhart, line 84:
```suggestion
| `panel-title` | String | The title of the panel |
```

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+	_handlePanelClick(e) {
+		if (!this.expanded) {
+			this._toggleExpand();
+			e.stopPropagation();
+		}
+	}
+
+	_handleSummarySlotChange() {
+		const element = this.shadowRoot.querySelector('.d2l-collapsible-panel');
+		element.classList.add('has-summary');
```
dlockhart, line 335: Ditto here I think in that this should be a state-based `hasSummary` property that causes the class to be added/removed in `render()`.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+			this._intersectionObserver = new IntersectionObserver(entries => {
+				if (!this.expanded) return;
+
+				const entry = entries[0];
+				const element = this.shadowRoot.querySelector('.d2l-collapsible-panel');
+				if (entry.isIntersecting) {
+					element.classList.remove('scrolled');
+				} else {
+					element.classList.add('scrolled');
+				}
```
dlockhart, line 255: Once there's an `isScrolled` state property, this can obviously be simplified:

```suggestion
				this.isScrolled = !entry.isIntersecting;
```

components/collapsible-panel/collapsible-panel-summary-item.js
dbatiste: This comment was marked as resolved, but there doesn't appear to be any styles for this. Did we decide to exclude them for some reason?

jameswklassen: Oh good catch. I added these styles in 8d86e8d8725e260ce85254e25b7b09d6a83fbdd1, but only to the collapsible panel (not the summary item). Will fix 👍

components/collapsible-panel/collapsible-panel.js
jameswklassen: Ahh, good call. I was avoiding having two equivalent values (css class + state property), but I guess it's fine to have both 😄

components/collapsible-panel/collapsible-panel.js
jameswklassen: Oh sorry, just re-read [your comment](https://github.com/BrightspaceUI/core/pull/3025#discussion_r1053643225) which I initially misread...will fix.

> You likely don't want to fire the event though from there, so that part could stay here.

components/collapsible-panel/README.md
jameswklassen: Yup 😄 demo file is still somewhat volatile

components/collapsible-panel/README.md
jameswklassen: gah...thought I caught em all.

components/collapsible-panel/collapsible-panel.js
dbatiste: +1.  I like to avoid `:not(...)` - whenever I see it I usually look for a better way, since the `:not()` is effectively an extra selector check the browser needs to do, optimization that it might be doing aside.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+
+	_handlePanelClick(e) {
+		if (!this.expanded) {
+			this._toggleExpand();
+			e.stopPropagation();
+		}
+	}
+
+	_handleSummarySlotChange() {
+		this._hasSummary = true;
```
dbatiste, line 336: Is this necessarily true? For instance, if a consumer removes all content from the slot.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	_onBlur() {
```
dbatiste, line 339: I think Dave brought this issue up elsewhere - these `blur` and `focus` handlers are modifying the DOM outside of render, which generally isn't cool. `UI = render(props)` should always be true (and in theory the inverse should also be true). 

To address that, we need an internal reactive state property, `_focused: { state: true }`, and apply the class in `render`.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+			<div class="d2l-collapsible-panel-header" @click="${this._handleHeaderClick}">
+				<div class="d2l-collapsible-panel-header-primary">
+					${this._renderPanelTitle()}
+					<div class="d2l-collapsible-panel-header-actions" @click="${this._handleActionsClick}">
+						<slot name="actions" @slotchange="${this._handleActionsSlotChange}"></slot>
+					</div>
+					<div class="d2l-collapsible-panel-opener">
+						<d2l-icon-custom size="tier1">
+							<svg xmlns="http://www.w3.org/2000/svg" width="10" height="18" fill="none" viewBox="0 0 10 18">
+								<path stroke="#494c4e" stroke-linejoin="round" stroke-width="2" d="m9 9-8 8V1l8 8Z"/>
```
dbatiste, line 360: I think this can be `stroke="var(--d2l-color-tungsten)"`.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -0,0 +1,408 @@
...
+		let headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ? this.headingLevel : this.headingStyle;
+		headingStyle = normalizeHeadingNumber(headingStyle);
+
+		const titleClasses = {
+			'd2l-collapsible-panel-title': true,
+			[`d2l-heading-${headingStyle}`]: true,
+		};
+
+		const headingLevel = normalizeHeadingNumber(this.headingLevel);
+		if (headingLevel === 1) {
```
dbatiste, line 383: A `switch` statement could be a little cleaner here.
```javascript
switch (headingLevel) {
	case 1: return html`<h1 class="${classMap(titleClasses)}">${this.panelTitle}</h1>`;
	case 2: return html`<h2 class="${classMap(titleClasses)}">${this.panelTitle}</h2>`;
	case 3: return html`<h3 class="${classMap(titleClasses)}">${this.panelTitle}</h3>`;
	case 4: return html`<h4 class="${classMap(titleClasses)}">${this.panelTitle}</h4>`;
}
```

components/collapsible-panel/test/collapsible-panel.visual-diff.js
```diff
@@ -0,0 +1,54 @@
...
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await visualDiff.createPage(browser);
+		await page.goto(`${visualDiff.getBaseUrl()}/components/collapsible-panel/test/collapsible-panel.visual-diff.html`, { waitUntil: ['networkidle0', 'load'] });
+		await page.bringToFront();
+	});
+
+	after(async() => await browser.close());
+
+	[
```
dbatiste, line 20: I think we should have RTL versions of some of these.

`d2l-switch` has an example on how to manage this when there are several components in play. (see [html page](https://github.com/BrightspaceUI/core/blob/main/components/switch/test/switch.visual-diff.html#L6), see [javascript](https://github.com/BrightspaceUI/core/blob/main/components/switch/test/switch.visual-diff.js#L78)). The alternative is to setup the fixture with `dir="rtl"`, but that is a limited solution that only targets the host, so the solution I linked is better.

We're not totally consistent in how we organize the tests, but we do try to avoid running unnecessary tests (like check RTL for something that we would never expect to be affected by RTL).

components/collapsible-panel/collapsible-panel.js
jameswklassen: Oh... I could've sworn I tried this and it didn't work. But it does. Thanks 😄

components/collapsible-panel/test/collapsible-panel.visual-diff.js
jameswklassen: Good call. Yeah a RTL test probably makes sense here. I'll take a look.

components/collapsible-panel/README.md
```diff
@@ -0,0 +1,164 @@
...
+	This is the "subtle" collapsible panel style
+</d2l-collapsible-panel>
+
+<d2l-collapsible-panel type="inline" panel-title="Inline panel">
+	This is the "inline" collapsible panel style
+</d2l-collapsible-panel>
+```
+
+## Summary Items [d2l-collapsible-panel-summary-item]
+The summary area takes information from the expanded panel and summarizes it for the collapsed version.This can help the user understand what information is inside the panel without having to click on it.
```
dbatiste, line 96:
```suggestion
The summary area takes information from the expanded panel and summarizes it for the collapsed version. This can help the user understand what information is inside the panel without having to click on it.
```

components/collapsible-panel/README.md
```diff
@@ -0,0 +1,164 @@
...
+		<d2l-collapsible-panel-summary-item text="Always available"></d2l-collapsible-panel-summary-item>
+	</div>
+	Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas odio ligula, aliquam efficitur sollicitudin non, dignissim quis nisl. Nullam rutrum, lectus sed finibus consectetur, dolor leo blandit lorem, vitae consectetur arcu enim ornare tortor. Praesent lobortis libero in libero sagittis consectetur. Maecenas ut velit efficitur, consectetur augue vitae, finibus turpis. In id tempor quam. Integer sed facilisis mi. Interdum et malesuada fames ac ante ipsum primis in faucibus. Ut a volutpat lacus. Suspendisse potenti. Quisque egestas erat urna, et accumsan est accumsan sit amet. Sed luctus vestibulum lacus. Mauris nisi orci, rhoncus sed est sit amet, pretium facilisis felis.
+</d2l-collapsible-panel>
+```
+
+## Accessibility
+
+### Panel label
+By default, the panel is described by screen readers with the `panel-title` attribute. There may be situations where the screen reader should read additional information. In this case, the aria label can be explicitly defined using the `expand-collapse-label` property.
```
dbatiste, line 158: Just update wording slightly to not leak implementation details. 

```suggestion
By default, the panel is described by screen readers with the `panel-title` attribute. There may be situations where the screen reader should read additional information. In this case, a special label can be specified using the `expand-collapse-label` property.
```

dbatiste (approved):

PR #3027, opened 2022-11-29
DE51092 - Correct List Count

svanherk:
This moves the list header and the list's pager out of the actual list, as both were being counted as list items and causing axe tests to fail.  We didn't actually have an axe test for the header so I've added one, fixed a demo that was broken, and removed an unused class.

In a separate PR, I'm going to try to add a description for the list toolbar.

### DIFF
--- a/components/list/demo/list-item-actions.html
+++ b/components/list/demo/list-item-actions.html
@@ -326,11 +326,11 @@
 
 			<d2l-demo-snippet id="grid-button-demo">
 				<template>
-					<d2l-list-header slot="header">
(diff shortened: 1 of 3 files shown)


dbatiste (approved):
Looks good! 

I suspect there was an original intended use for that class but it never ended up getting used.

dlockhart (approved):
Nice!

eKoopmans (approved):
Thanks for fixing this!

> In a separate PR, I'm going to try to add a description for the list toolbar

Do you mean the list header, or is list toolbar something else? There's already some description of the list-header in list's readme (also selection's readme...)

oelhanafey (approved):
Thanks!

PR #3031, opened 2022-11-29
chore: Publish after release for correct version

eKoopmans:
Fixing a small defect from #3022 - currently the stated version on [our demo page](https://brightspace-ui-core.d2l.dev/) is one version behind.

In the release workflow, we need to force `publish` to happen after `release`, so that the version has been correctly updated before publishing.


dlockhart (approved):
Makes sense!

eKoopmans:
Visual-diff flake: `d2l-input-date-time-range skeleton wide-hidden-labels-values`

PR #3037, opened 2022-11-30
DE51092 - Add a section and aria label around the list header to add context

svanherk:
As discussed, we'd like to add a bit more context for a screen reader user now that the list header is no longer inside the list. 
 We've gone with a `section` to give us a `region` role, since:
- `aria-labelledby` and `aria-describedby` need an interactive element to not be ignored
- `aria-description` doesn't appear to work on Safari
- the `toolbar` role implies keyboard behavior we don't have

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -56,6 +56,9 @@ class List extends PageableMixin(SelectionMixin(LocalizeCoreElement(LitElement))
 			slot[name="pager"]::slotted(*) {
 				margin-top: 12px;
 			}
+			section {
+				display: unset;
+			}
 		`;
 	}
 
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -1,6 +1,7 @@
 import { css, html, LitElement } from 'lit';
 import { getNextFocusable, getPreviousFocusable } from '../../helpers/focus.js';
 import { SelectionInfo, SelectionMixin } from '../selection/selection-mixin.js';
+import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { PageableMixin } from '../paging/pageable-mixin.js';
 
 const keyCodes = {
@@ -16,7 +17,7 @@ export const listSelectionStates = SelectionInfo.states;
  * @slot pager - Slot for `d2l-pager-load-more` to be rendered below the list
  * @fires d2l-list-items-move - Dispatched when one or more items are moved. See [Event Details: d2l-list-items-move](#event-details%3A-%40d2l-list-items-move).
  */
-class List extends PageableMixin(SelectionMixin(LitElement)) {
+class List extends PageableMixin(SelectionMixin(LocalizeCoreElement(LitElement))) {
 
 	static get properties() {
 		return {
@@ -111,7 +112,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 	render() {
 		const role = !this.grid ? 'list' : 'application';
 		return html`
-			<slot name="header"></slot>
+			<section aria-label="${this.localize('components.list.toolbar')}"><slot name="header"></slot></section>
 			<div role="${role}">
 				<slot @keydown="${this._handleKeyDown}" @slotchange="${this._handleSlotChange}"></slot>
 			</div>
--- a/lang/en.js
+++ b/lang/en.js
@@ -65,6 +65,7 @@ export default {
 	"components.input-time-range.errorBadInput": "{startLabel} must be before {endLabel}",
 	"components.input-time-range.startTime": "Start Time",
 	"components.interactive.instructions": "Press Enter to interact, Escape to exit",
+	"components.list.toolbar": "List toolbar",
 	"components.list-item-drag-handle.default": "Reorder item action for {name}",
 	"components.list-item-drag-handle.keyboard": "Reorder item, current position {currentPosition} out of {size}. To move this item, press up or down arrows.",
 	"components.list-item-tooltip.title": "Keyboard Navigation for Lists:",
--- a/lang/ar.js
+++ b/lang/ar.js
@@ -65,6 +65,7 @@ export default {
 	"components.input-time-range.errorBadInput": "يجب أن يكون تاريخ {startLabel} قبل {endLabel}",
 	"components.input-time-range.startTime": "وقت البدء",
 	"components.interactive.instructions": "اضغط على Enter للتفاعل، وEscape للخروج",
+	"components.list.toolbar": "List toolbar",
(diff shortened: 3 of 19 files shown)


components/list/list.js
```diff
@@ -111,7 +112,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 	render() {
 		const role = !this.grid ? 'list' : 'application';
 		return html`
-			<slot name="header"></slot>
+			<section aria-label="${this.localize('components.list.toolbar')}"><slot name="header"></slot></section>
```
svanherk, line 115: @eKoopmans @dbatiste I figured it made sense to put this here rather than inside `selection-header`, since that seemed a bit more general.  But depending on how the table header/toolbar is going to work, maybe it makes more sense to be lower?  `d2l-lsit-header` itself doesn't `render` anything, so I'd have to expose a way for it to set the text.

components/list/list.js
```diff
@@ -55,6 +56,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			slot[name="pager"]::slotted(*) {
 				margin-top: 12px;
 			}
+			section {
+				display: unset;
```
dbatiste, line 60: Curious to know more about this - why the default `display` value isn't good...?

svanherk: Seems to be interacting with `sticky` in a way I wasn't expecting, and causes this visual-diff failure with the default `block` display: https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/list/2022-11-30-20.38.52.695-a648e674d047ed0eae6d71f1aaa92c5ccd113b79-report.html
Setting this to `unset` (or `contents` or `inline`) seems to fix it locally, will see if it does in CI as well

components/list/list.js
svanherk: Yeah my thinking was the alternative would be to wrap [this](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-header.js#L132-L145) in a `section` and call `_getSelectionHeaderAriaLabel` or whatever to get the `aria-label`, which `d2l-list-header` would override.  Similar to how it gets the container classes right now.  But it did feel more complicated and I'd want to do more regression testing if I'm placing the section inside like that

components/list/list.js
svanherk: Yeah, that seemed to fix it.  I'm guessing it was somehow messing with the intersection observer, I can play around a bit

components/list/list.js
eKoopmans: Hey sorry, just getting to this all now.

I personally like replacing (not wrapping) the [selection-header-container div](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-header.js#L132-L145) with a `section`, and having the aria-label overrideable.

lang/en.js
```diff
@@ -65,6 +65,7 @@ export default {
 	"components.input-time-range.errorBadInput": "{startLabel} must be before {endLabel}",
 	"components.input-time-range.startTime": "Start Time",
 	"components.interactive.instructions": "Press Enter to interact, Escape to exit",
+	"components.list.toolbar": "List toolbar",
```
oelhanafey, line 68: In the lms, you can create a langterm entry in the `definitions` file and then when translations go through it will create translations for all of the other languages that are missing it. Does it work differently in core where we need to manually create an entry for every language?

components/list/list.js
eKoopmans: I'm going to have a quick look into why section would be messing up sticky...

lang/en.js
svanherk: So I actually used `npx mfv add-missing -s en -p ./lang/` to do this for me, because otherwise I _think_ `npm run test:translations` in CI will fail

components/list/list.js
svanherk: > I'm going to have a quick look into why section would be messing up sticky...

Awesome, that would be great!  I'm not finding anything obvious, there's some posts about `overflow` being a problem or some saying "Change `display` to `initial` or `inline`" but they don't explain _why_

eKoopmans: 😂 This is my life with sticky. Yes, overflow's the huge main problem that always rears its head.

That's not what's going on here though - it's about having *any* parent element (even a div). But there are other cases where that's not a problem - still trying to get to the bottom of it...

eKoopmans: Okay so yes, it 100% makes sense that putting a wrapping div around a sticky element would break it. Here's the Two Basic Rules of Sticky (™️ Erik Koopmans):

1. A sticky element is only free to move within the confines of its nearest block-level ancestor
2. A sticky element sticks relative to its nearest "scrolling container" (e.g. element with any overflow defined)

That's it. There's no breaking of those rules allowed, period (when using CSS sticky). So a wrapping `section` (with block display) applies to item 1 - and since the wrapper is the same size as the sticky element, it has nowhere to move.

I was getting confused because I've been dealing so much with Rule 2, I forgot about Rule 1 😛.

components/list/list.js
eKoopmans: Another option - we could have the selection-header itself be the region, by doing something like:

```js
static get properties() {
    return {
        // ...
       role: { type: String, reflect: true },
       ariaLabel: { type: String, attribute: 'aria-label', reflect: true },
}

constructor() {
    super();
    this.role = 'region';
    this.ariaLabel = this.localize('components.selection.toolbar');
}
}
```

svanherk: Hmm yeah, the WCAG docs recommend `section` over using the `role` directly but I think this is a case where it would make a lot of sense - we don't _want_ another element that's messing with the sticky positioning.  We'd have to recalculate the langterm on language change if we did the above, or we could have the consumer pass it in.

@eKoopmans Do you know yet if `table-wrapper` will have a slot for the table header/toolbar?  As in, will _we_ be rendering the `table-header` and can pass in the langterm for the consumer (like list does), or will every consumer be rendering it themselves? I think that would make the decision on who passes/sets the aria-langterm

eKoopmans: The plan is to use the same pattern as list, with a slot for the table-header.

svanherk: Ok, here's what it looks like placed inside `d2l-selection-header`: https://github.com/BrightspaceUI/core/pull/3040
I did end up opening a second PR to make it easier to compare

dbatiste: Seems good to me, and probably the preferred approach if it avoids issues with sticky position.

svanherk:
Closing this PR in favour of https://github.com/BrightspaceUI/core/pull/3040

PR #3040, opened 2022-12-01
DE51092 - Add section and label to selection header

svanherk:
This is an alternative to https://github.com/BrightspaceUI/core/pull/3037 where we place the `section` right in the `selection-header` instead, so that the sticky positioning is unaffected.  We could also keep the `div` and set `role="region"` but WCAG recommends `section` when possible and I haven't found any regressions with the below.


svanherk:
Visual-Diff Flake: `d2l-input-date-time-range wide-hidden-labels-values`

components/list/list-header.js
```diff
@@ -1,12 +1,13 @@
...
 import { findComposedAncestor } from '../../helpers/dom.js';
+import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { SelectionHeader } from '../selection/selection-header.js';
 
 /**
  * A header for list components containing select-all, etc.
  * @slot - Responsive container using `d2l-overflow-group` for `d2l-selection-action` elements
  */
-class ListHeader extends SelectionHeader {
+class ListHeader extends LocalizeCoreElement(SelectionHeader) {
```
eKoopmans, line 10: We shouldn't need this here I think, since the base `SelectionHeader` class already uses `LocalizeCoreElement`.

svanherk: Yeah I wasn't sure about this - in theory if we ever removed `LocalizeCoreElement` from the `SelectionHeader`, then the `localize` call in here would start silently breaking.  And we dedupe the `LocalizeMixin`, so I don't _think_ having it twice is a problem.  But in practice, it's highly unlikely we would ever refactor away the `LocalizeCoreElement` in `SelectionHeader` without the `localize` call changing too.

It feels similar to how we always want to include an import we use, even if we import another component that already imports it.  But I don't know if we've ever run into that same scenario with mixins 🤔

eKoopmans: Hmm... the `LocalizeCoreElement` and `LocalizeDynamicMixin` layers aren't deduped though. Currently they don't do anything that would cause problems (not referencing super anywhere) - but seems like that's its own risk of silent breakage, if either of those added something self-referencing.

svanherk: Fair enough, seems like both ways are a slight risk but the "repeated localize mixins" one is higher

svanherk: Thinking about it more, I have this super hazy memory of us already dealing with _something_ like this... @dlockhart, did you and Margaree run into something with extending `LocalizeCoreElement` multiple times?

dlockhart: It had to do with combining a mixin that used `LocalizeDynamicMixin` and `LocalizeCoreElement`. We added this note to the documentation because of it:

> Note: If using LocalizeCoreElement or a mixin that utilizes LocalizeCoreElement as well as LocalizeDynamicMixin or a mixin that uses LocalizeDynamicMixin, LocalizeDynamicMixin must appear before LocalizeCoreElement in the chain.

 You might actually be in that situation, but because underneath everything it's all the same localization resources, it should be fine.

svanherk: @eKoopmans We'd mentioned passing around the langterm name instead, which I don't _love_, but I wonder if that's the best way to future-proof and divide responsibility here.  Keeps the `SelectionHeader` in charge of localizing since it knows how to do that, but still allows `ListHeader` to pick the wording

oelhanafey: Could we construct a unit test that ensures the localization is working correctly in `ListHeader`? That way it would catch a regression if we made a breaking change to the mixins.

dbatiste: I _think_ visual-diff could catch that.

svanherk: > I think visual-diff would catch that.

Might be misunderstanding what you're saying - the label is only used for an `aria-label`, not a visual one

eKoopmans: I like a simple unit test, that'll also make sure it's correctly overriding the behaviour from the selection-header Something like:

```js
// Note: The property `element.ariaLabel` doesn't work in Firefox
expect(element.getAttribute('aria-label')).to.equal('Expected value');
```

dbatiste: Sorry, you are correct. If it were visible though, we would want to forgo the unit test.

components/selection/test/selection.test.js
```diff
@@ -86,6 +87,20 @@ describe('d2l-selection-action-menu-item', () => {
 
 });
 
+describe('d2l-selection-header', () => {
```
svanherk, line 90: Added these as well for some nice symmetry, but not tied to them

eKoopmans (approved):

components/selection/selection-header.js
```diff
@@ -127,10 +128,10 @@ export class SelectionHeader extends SelectionObserverMixin(RtlMixin(LocalizeCor
 
 	render() {
 		const classes = this._getSelectionHeaderContainerClasses();
-		const label = this._getSelectionHeaderLabel();
+		const label = (this._hasActions || !this.noSelection) ? this._getSelectionHeaderLabel() : null;
```
eKoopmans, line 131: I think this will have to move into `this._getSelectionHeaderLabel` for the table header to work, because its conditions will be a bit different (since it doesn't display a select-all checkbox). But I can make that change in my PR.

eKoopmans (approved):

components/selection/selection-header.js
svanherk: Ah ok, sounds good! Yeah I wanted to keep it out of `list-header` if it was just going to be the exact same in both places, but once table is different feel free to move it as needed.  Once that happens it probably makes sense to also add tests for it in `list` and `table` accordingly

Issue #3034, opened 2022-11-30
d2l-pager-load-more not accessible/focusable in d2l-list with grid attribute


ChrisLabattD2L:
Currently you cannot access the `d2l-pager-load-more` via keyboard navigation when it's being used in association with a `d2l-list` that has the `grid` attribute.

The only work-around for this is if there are other focusable components after the list & load-more button. In that case, you can tab to that, then reverse tab into the `d2l-pager-load-more` component. This isn't much a workaround though, since you'd need to have a design with this in mind, which isn't always possible.

dbatiste:
Thanks for raising this Chris. We should be able to make a fix for this.

ChrisLabattD2L:
Thanks Dave! I've noticed that there's an instance in which this issue doesn't occur, which is when the last item in the list isn't focusable due to not having a `href`. I'm not sure if that's related to the problem or not, but I figure I'd mention it in case it's relevant.

svanherk:
I've created [DE51137](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F672337163059) if you want to follow along!

ChrisLabattD2L:
Thanks Stacey!

svanherk:
This should be fixed for `core` version `2.67.4`!  @ChrisLabattD2L let us know if you still see any issues!

ChrisLabattD2L:
Just tested it quickly, and it's no longer an issue, thanks 😄

PR #3044, opened 2022-12-01
DE51137 - Fix list to load more pager focusing in grid mode

svanherk:
Fixes https://github.com/BrightspaceUI/core/issues/3034. In `grid` mode, the user navigates a list with the arrow keys, and tabbing forward/backwards will bring them to the first focusable element after/before the list.  This was working properly for backwards-tabbing because we were passing in the list `slot` to `getPreviousFocusable`, to account for the header.  But forward tabbing was skipping over the load-more pager, because we were passing `this` into `getNextFocusable`.  This would have worked fine until the pager was added, so now we just pass the list `slot` into both.

### DIFF
--- a/index.html
+++ b/index.html
@@ -128,6 +128,7 @@
 			<ul>
 				<li><a href="components/list/demo/list.html">d2l-list</a></li>
 				<li><a href="components/list/demo/list-drag-and-drop.html">d2l-list (drag & drop)</a></li>
+				<li><a href="components/list/demo/list-nested.html">d2l-list (nested)</a></li>
 				<li><a href="components/list/demo/list-item-actions.html">d2l-list-item (actions)</a></li>
 				<li><a href="components/list/demo/list-item-layouts.html">d2l-list-item (layouts)</a></li>
 				<li><a href="components/list/demo/list-item-scroll.html">d2l-list-item (scrolling)</a></li>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -128,6 +128,7 @@ <h2 class="d2l-heading-3">Components</h2>
 			<ul>
 				<li><a href="components/list/demo/list.html">d2l-list</a></li>
 				<li><a href="components/list/demo/list-drag-and-drop.html">d2l-list (drag & drop)</a></li>
+				<li><a href="components/list/demo/list-nested.html">d2l-list (nested)</a></li>
```
svanherk, line 131: Adding this secret demo page to the list 😛

eKoopmans (approved):
Nice!

dbatiste (approved):
Awesome. Glad that turned out to be the root cause. 😄

index.html
dbatiste: The drag & drop page also has a nested example. With all the various ways the list can be configured, it's hard to represent each of them alone with a bizzzillion demo pages, but sometimes that what we want to debug an issue. 😄

svanherk: Yeah, would probably be good to consolidate at some point, and then really flesh out those few demos to have all the functionality. In this case I wanted a nested one that already had the header, and I always forget about that nested demo page

svanherk:
> Awesome. Glad that turned out to be the root cause. 😄

@dbatiste Yeah you were dead on about where you thought the issue was, and then I went down that crazy NVDA rabbit hole

PR #3046, opened 2022-12-02
Expand/collapse list feature - US133784

imcdonald-d2l:
See README for additional properties and demos for functional behavior.

**Note:** will merge vdiff changes once code and functionality has been reviewed.

https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fuserstory%2F615923325399


components/list/list.js
```diff
@@ -193,8 +208,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		if (refresh) {
 			this._itemsShowingTotalCount = await this.getItems().reduce(async(count, item) => {
 				await item.updateComplete;
-				if (item._selectionProvider) {
-					return (await count + await item._selectionProvider._getListItemsShowingTotalCount(true));
+				if (item._hasChildren && (!item.expandable || item.expanded)) {
```
imcdonald-d2l, line 211: Had to swap to using the new `_hasChildren` since `_selectionProvider` is only set when `selectable` is set on `ListItemCheckboxMixin`. So load more component wasn't working properly with non-selectable lists.

components/button/button-icon.js
```diff
@@ -62,6 +62,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 					--d2l-button-icon-h-align: calc(((2rem + 2px - 0.9rem) / 2) * -1);
 					--d2l-icon-fill-color: var(--d2l-button-icon-fill-color, var(--d2l-color-tungsten));
 					display: inline-block;
+					line-height: 0;
```
dbatiste, line 65: What was happening that required this change?

components/list/README.md
```diff
@@ -347,6 +408,8 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
 | `drag-target-handle-only` | Boolean | Make the drag target the drag handle only. |
 | `drop-nested` | Boolean | Whether nested items can be dropped on this item |
 | `drop-text` | String | Text to drag and drop |
+| `expandable` | Boolean | Whether or not to show the expand/collapse toggle. |
+| `expanded` | Boolean | Default state for expand collapse toggle - if not set, collapsed will be the default state. Requires `expandable` to be set. |
```
dbatiste, line 412: We're not entirely consistent with how we specify defaults in these tables, partly because we've changed over time, and because we've had a number of different authors, some not all from our team (like the list component). 

I think our current practice is to specify them, and to do it like this, however we would opt for being consistent in the current context (which implies the default rather than being explicit). Honestly, the entire list documentation needs an overhaul, so in this case I would just be consistent with what's there.

```markdown
| `expanded`, default: `false` | Boolean | Whether the item is expanded. Requires `expandable` to be set. |
```

I think it would be good to update the wording as I have above to be clear that this property controls the state, and it's not only for the default/initial state.

components/button/button-icon.js
imcdonald-d2l: The height of the button icon was getting an extra pixel since it was using `line-height` instead of the icon height.

components/list/list-item-expand-collapse-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+		return html`
+			<div class="d2l-list-children-loading">
+				<d2l-loading-spinner size="40"></d2l-loading-spinner>
+			</div>`;
+	}
+
+	_renderExpandCollapse() {
+		return html`
+			<div slot="expand-collapse" class="d2l-list-expand-collapse" @click="${this._toggleExpandCollapse}">
+				${this.expandable ? html`<d2l-button-icon
```
dbatiste, line 132: Maybe this should be (like checkbox):
```javascript
return this.expandable ? html`
	<d2l-button-icon ...></d2l-button-icon>
` : nothing;
```
And the `<div slot="expand-collapse">...</div>` be moved out to the `ListItemMixin` where it controls the layout (what slot of the generic layout this renders into), like the rest of the slots. I think originally, one of the ideas with these lower level mixins was to make them agnostic to the actual layout or slots being used. That may or may not have been done effectively, but the `ListItemMixin` is the thing that tries to integrate these things into the layout.

components/button/button-icon.js
dbatiste: I think this should be fine. Though I get 42x42px with or without this. 🤔

components/list/list-item-mixin.js
```diff
@@ -530,6 +586,17 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 		this._hovering = false;
 	}
 
+	_onNestedSlotChange() {
+		if (this.selectable) {
+			this._onNestedSlotChangeCheckboxMixin();
+		}
+		const nestedList = this._getNestedList();
+		if (this._hasChildren !== !!nestedList) {
```
dbatiste, line 594: Not intending to bike-shed, but I'm wondering if this should be `_hasNestedList`. Thoughts? Would it matter if there was a nested list that's empty? It's been really difficult in the list components, but we've tried to be consistent and accurate with how stuff is named, hopefully to avoid any confusion.

components/button/button-icon.js
imcdonald-d2l: Weird, it definitely was busted without it for me earlier. See thread: https://d2l.slack.com/archives/G03Q166G2/p1669911968210739

components/list/list-item-mixin.js
imcdonald-d2l: I'd be fine renaming to `_hasNestedList`, not a lot of effort to clean that up.

components/list/list-item-expand-collapse-mixin.js
imcdonald-d2l: I could move the `d2l-list-expand-collapse` div to the parent `ListItemMixin` but we would want to move all the corresponding styles as well (which is the majority of them in this mixin). For the sake of code cleanliness/readability I would prefer to avoid this move. Willing to move if there is a strong preference though.

components/list/README.md
imcdonald-d2l: To clarify, do we want to state `default: false` as you have in your example? It would not be consistent with current documentation so I assume we do not want to include that.

components/button/button-icon.js
dbatiste: Ok, yeah. I think the custom dimensions are an important factor there.

components/list/README.md
dbatiste: Correct. I would exclude it for now given it would be inconsistent. Feel free to mention it in the description if you feel it's not implied. But I think we should update the description as mentioned.

components/list/list-item-expand-collapse-mixin.js
dbatiste: My preference is to centralize all the layout in two places - `ListItemMixin` and `d2l-list-item-generic-layout` so that we know where all the layout is managed. But it's not a huge deal... a lot of this stuff is already commingled, and my intention is that these lower level mixins are never used anywhere except `ListItemMixin`.

components/list/list-item-expand-collapse-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+			expanded: { type: Boolean, reflect: true },
+			_siblingHasNestedItems: { state: true },
+			_renderExpandCollapseSlot: { type: Boolean, reflect: true, attribute: '_render-expand-collapse-slot' },
+			_showNestedLoadingSpinner: {  state: true }
+		};
+	}
+
+	static get styles() {
+		const styles = [ css`
+			:host {
```
dbatiste, line 31: The animation should be a progressive enhancement for when the user has no motion preference.
```javascript
	.d2l-list-expand-collapse {
		transition: width var(--d2l-expand-collapse-slot-transition-duration) cubic-bezier(0, 0.7, 0.5, 1);
	}
}
```

components/list/list-item-expand-collapse-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+
+	updated(changedProperties) {
+		if (changedProperties.has('_siblingHasNestedItems') || changedProperties.has('expandable')) {
+			this._renderExpandCollapseSlot = this.expandable || this._siblingHasNestedItems;
+		}
+		if (changedProperties.has('_draggingOver') && this._draggingOver && this.dropNested && !this.expanded && this.expandable) {
+			let elapsedHoverTime = 0;
+			let dragIntervalId = null;
+			const watchDraggingOver = () => {
+				if (elapsedHoverTime === dragHoverDropTime) {
```
dbatiste, line 94: Nit: this relies on the `dragHoverDropTime` being evenly divisible by the interval, which is true as configured, but if either were ever changed there is a reasonable chance it would no longer be, and this condition might never be met. Suggest change this to `>=`.

components/list/list-item-drag-drop-mixin.js
```diff
@@ -559,14 +601,34 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
...
+						break;
+					}
+				}
+			}
 			dragImage.count = selectionInfo.keys.length;
+			dragImage.includePlusSign = includePlus;
+			e.dataTransfer.setDragImage(dragImage, 24, 26);
+		} else if (rootList.dragMultiple && this.expandable) {
+			const flattenedListItems = this._getFlattenedListItems(this);
+			let dragImage = this.shadowRoot.querySelector('d2l-list-item-drag-image');
```
dbatiste, line 625: Looks like we could factor this out into a function to avoid the duplication from above. It could be done inline in this method or at the class level.
```javascript
const getDragImage = (count, includePlusSign) => {
	let dragImage = this.shadowRoot.querySelector('d2l-list-item-drag-image');
	if (!dragImage) {
		dragImage = document.createElement('d2l-list-item-drag-image');
		this.shadowRoot.appendChild(dragImage);
	}
	dragImage.count = count;
	dragImage.includePlusSign = includePlusSign;
	return dragImage;
};
```

components/list/list-item-expand-collapse-mixin.js
imcdonald-d2l: I moved the slot div to `ListItemMixin` but kept the specific styles in the `ListItemExpandCollapseMixin`... not ideal but don't really want all the styles getting lost in the larger parent mixin.

components/list/list-item-expand-collapse-mixin.js
```diff
@@ -0,0 +1,161 @@
...
+				<d2l-loading-spinner size="40"></d2l-loading-spinner>
+			</div>`;
+	}
+
+	_toggleExpandCollapse() {
+		if (!this.expandable) {
+			return;
+		}
+		this.expanded = !this.expanded;
+		this.dispatchEvent(new CustomEvent('d2l-list-item-expand-collapse-toggled', {
```
dbatiste, line 158: I think this event is public, ya? If so, I think in order for it to show up in the documentation on the Daylight site, we will want a comment above this line [such as is done here](https://github.com/BrightspaceUI/core/blob/main/components/dialog/dialog-mixin.js#L297).

components/list/list-item-mixin.js
```diff
@@ -530,6 +586,17 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
...
 	}
 
+	_onNestedSlotChange() {
+		if (this.selectable) {
+			this._onNestedSlotChangeCheckboxMixin();
+		}
+		const nestedList = this._getNestedList();
+		if (this._hasNestedList !== !!nestedList) {
+			this._hasNestedList = !!nestedList;
+			this.dispatchEvent(new CustomEvent('d2l-list-item-nested-change', { bubbles: true, composed: true }));
```
dbatiste, line 596: I this this is consider an internal/private event, ya? If so, this should have `/** @ignore */` on the line preceding the `dispatch` call, [such as here](https://github.com/BrightspaceUI/core/blob/main/mixins/labelled-mixin.js#L184). It wouldn't have a description, but it would still show up in the generated Daylight docs.

components/list/test/list-item-expand-collapse-mixin.test.js
```diff
@@ -0,0 +1,144 @@
...
+				if (test.actionAreaAvailable) {
+					expect(actionArea).to.exist;
+				} else {
+					expect(actionArea).to.not.exist;
+				}
+			});
+		}
+	});
+
+	describe('Render expand/collapse slot and button when appropriate', () => {
```
dbatiste, line 61: Are these already effectively captured by the visual-diff tests? If so, I would be happy to have these pared down.

components/list/test/list.visual-diff.js
```diff
@@ -167,6 +167,13 @@ describe('d2l-list', () => {
 			{ name: 'none-selected', selector: '#nested-none-selected' },
 			{ name: 'some-selected', selector: '#nested-some-selected' },
 			{ name: 'all-selected', selector: '#nested-all-selected', action: () => wait('#nested-all-selected d2l-list-header', 100) }
+		] },
+		{ category: 'expand-collapse', tests: [
```
dbatiste, line 171: I think it would be good to include a focus test here. There are a couple of focus helper methods above that the other tests rely on to apply focus (ex. `focusButton`) - I suspect (hopefully) we could use them here to capture that state. 

There is also a category above, "focus method" which tests the logic of `item.focus()` to ensure those elements get focused when they are the "first" focusable element. To do so, I think those tests target very simple fixtures for those cases. It would be good to include the expandable case up there in that category, alongside the href, button, selectable, and actions cases.

components/selection/selection-summary.js
```diff
@@ -35,25 +36,54 @@ class Summary extends LocalizeCoreElement(SelectionObserverMixin(LitElement)) {
...
-			summary = (this._provider.selectionCountOverride === 0 && this.noSelectionText ?
-				this.noSelectionText : this.localize('components.selection.selected', 'count', count));
+			this._summary = this._provider.selectionCountOverride === 0 && this.noSelectionText ?
+				this.noSelectionText : this.localize('components.selection.selected', 'count', count);
 		} else {
-			count = (this.selectionInfo.state === SelectionInfo.states.allPages ?
-				this._provider.itemCount : this.selectionInfo.keys.length);
+			let includePlus = false;
+			if (this._provider && this._provider.getItems) {
+				const lazyLoadListItems = this._provider.getItems()[0]._getFlattenedListItems().lazyLoadListItems;
```
dbatiste, line 65: Do you know how often this might get called? 

I'm always a little worried about the performance of these recursive methods. I did a bit of testing a year or two ago and found that for very large lists, the user experience degraded significantly, which is why the `d2l-list` component tends to make use of the `_selectionProvider` reference on the items. It still ends up being recursive, but it avoids the work that `getNestedList` does, and I'd found that was somewhat an improvement when working with large lists

dbatiste, line 65: Note that `d2l-selection-summary` is a generic component that could be used by `d2l-list`, and also `d2l-table-wrapper`, but this method relies on a list specific method, `_getFlattenedListItems`. I think we would want to figure out how to represent this generically as to push the list-specific aspect of this to the actual `d2l-list` class.

imcdonald-d2l: It would get called anytime selection info changes within a list. So every time selection changes we have to traverse the list tree to see if there are lazy load items.

imcdonald-d2l: I could add essentially add this to the list item if that's preferred:
```
_getLazyLoadItems() {
 return this.getItems()[0]._getFlattenedListItems().lazyLoadListItems;
}
```

And then this would change to just:
```
if (this._provider && this._provider._getLazyLoadItems) {
  const lazyLoadListItems = this._provider._getLazyLoadItems();
}
```

Thoughts?

dbatiste: Yeah that sounds ok. We'll want to add a default implementation of `_getLazyLoadItems` to [SelectionMixin](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-mixin.js#L38) that returns an empty map since `d2l-table-wrapper` won't have a need to implement this.

dbatiste: I think that's what I found to be janky with huge, deeply nested lists, as a result of all the logic to roll-up the selection to parents, etc. We might need to spend some time looking at ways to optimize this. Not sure it needs to be done in this PR thoug.

imcdonald-d2l: Yeah, doing this in a later PR would be ideal. It's already getting pretty large as-is.

components/selection/selection-summary.js
```diff
@@ -35,25 +36,54 @@ class Summary extends LocalizeCoreElement(SelectionObserverMixin(LitElement)) {
...
 	render() {
+		if (!this._summary) {
+			return nothing;
+		}
+		return html`
+			<p class="d2l-body-compact">${this._summary}</p>
+		`;
+	}
+
+	updated(changedProperties) {
```
dbatiste, line 47: It looks like the purpose of this is to compute `includePlus`, and ultimately the summary text that will get rendered on a subsequent pass of the render life cycle. Using `updated` followed by updating a reactive property (`_summary` in this case) is the classic way to handle this in Lit 1. 

Lit 2 provides a new life cycle callback that enables us to avoid the second iteration of the event loop, [willupdate](https://lit.dev/docs/components/lifecycle/#willupdate). It runs before `render`, and enables us to compute (non-reactive) properties that can be used by `render`, thereby avoiding the second update. I don't think we've used `willupdate` in core yet... but I think it would be an improvement here. It's something I think we could use in a few other places in core as we make revisions.

components/selection/selection-summary.js
imcdonald-d2l: We shouldn't need to add this to the `SelectionMixin`. I check `if (this._provider && this._provider._getLazyLoadItems)` in selection-summary so it won't execute if `_getLazyLoadItems` is undefined.

components/selection/selection-summary.js
imcdonald-d2l: Interesting, didn't know that! That is handy. Seems to work as expected so I'll push the changes. 🎉

devpow112 (approved):
Looks good to me :tada:
I'm really looking forwards to using this and switching away from our old, custom implementation for some outcomes stuff

components/list/list-item-expand-collapse-mixin.js
```diff
@@ -0,0 +1,172 @@
...
+		if (e) {
+			e.stopPropagation();
+		}
+		if (!this.expandable) {
+			return;
+		}
+		this.expanded = !this.expanded;
+		/** Dispatched whenever the list item expand state is toggled. Dispatched with the list item key, previous expanded state, and current expanded state. */
+		this.dispatchEvent(new CustomEvent('d2l-list-item-expand-collapse-toggled', {
+			detail: {
```
dbatiste, line 164: Is this event detail needed? Looks like the consumer can access all this by:
```javascript
key = e.target.key;
oldState = !e.target.expanded;
newState = e.target.expanded;
```

components/selection/selection-summary.js
```diff
@@ -35,25 +35,53 @@ class Summary extends LocalizeCoreElement(SelectionObserverMixin(LitElement)) {
...
 			count = this._provider.selectionCountOverride;
-			summary = (this._provider.selectionCountOverride === 0 && this.noSelectionText ?
-				this.noSelectionText : this.localize('components.selection.selected', 'count', count));
+			this._summary = this._provider.selectionCountOverride === 0 && this.noSelectionText ?
+				this.noSelectionText : this.localize('components.selection.selected', 'count', count);
 		} else {
-			count = (this.selectionInfo.state === SelectionInfo.states.allPages ?
-				this._provider.itemCount : this.selectionInfo.keys.length);
+			let includePlus = false;
+			if (this._provider && this._provider._getLazyLoadItems) {
```
dbatiste, line 62: If we have any more changes to make, I wouldn't mind if there was a comment for other devs who might not be as familiar... something to the effect of "If lazy loading items is supported (ex. d2l-list) then check the keys to determine if the plus sign should be included. If lazy loading is not supported (ex. d2l-table-wrapper), then skip this.". It might not be particularly evident at first glance that `_getLazyLoadItems` is part of the provider API... so a callout might be helpful.

imcdonald-d2l: Good call, I'll add that comment 👍🏻

components/list/list-item-expand-collapse-mixin.js
imcdonald-d2l: Good point, it might not be necessary. I'll test it out!

imcdonald-d2l: Works as expected - I'll push these changes.

devpow112 (approved):

dbatiste (approved):

PR #3051, opened 2022-12-05
feat: Add table-header component

eKoopmans:
This PR adds the `d2l-table-header`, which is equivalent to the existing `d2l-list-header` but for tables. Changes include:

- creating the table-header itself, and adding a slot for it into table-wrapper
- extracting how selection-header renders its selection pieces, so that table-header can omit the `d2l-select-all`
- adding the table-header into the table demos, readme, and tests

Rally: https://rally1.rallydev.com/#/?detail=/userstory/656749634867&fdp=true


components/selection/selection-header.js
```diff
@@ -76,14 +76,13 @@ export class SelectionHeader extends SelectionObserverMixin(RtlMixin(LocalizeCor
...
-			d2l-selection-summary {
-				flex: none;
+			d2l-selection-select-all + d2l-selection-summary {
 				margin-left: 0.9rem;
 			}
-			:host([dir="rtl"]) d2l-selection-summary {
+			:host([dir="rtl"]) d2l-selection-select-all + d2l-selection-summary {
 				margin-left: 0;
 				margin-right: 0.9rem;
 			}
```
eKoopmans, line 88: These styles were tightened so that the spacing is still correct when the select-all is omitted (for the table-header).

components/table/demo/table-test.js
```diff
@@ -1,13 +1,23 @@
...
 import { tableStyles } from '../table-wrapper.js';
 
 const fruits = ['Apples', 'Oranges', 'Bananas'];
 
-const data = [
-	{ name: 'Canada', fruit: { 'apples': 356863, 'oranges': 0, 'bananas': 0 }, selected: false },
-	{ name: 'Australia', fruit: { 'apples': 308298, 'oranges': 398610, 'bananas': 354241 }, selected: false },
+const data = () => [
+	{ name: 'Canada', fruit: { 'apples': 356863, 'oranges': 0, 'bananas': 0 }, selected: true },
+	{ name: 'Australia', fruit: { 'apples': 308298, 'oranges': 398610, 'bananas': 354241 }, selected: true },
```
eKoopmans, line 20: Changing this to a function means that each table on the demo can retrieve a fresh, independent copy of the data, so their selected state isn't linked.

Also changed the default selected value of two items to demo the selected behaviour more easily (borrowed from @dbatiste's draft PR).

components/table/demo/table-test.js
```diff
@@ -98,7 +145,7 @@ class TestTable extends RtlMixin(LitElement) {
 	_renderSortButton(fruit) {
 		const noSort = this._sortField !== fruit.toLowerCase();
 		return html`
-			<th>
+			<th scope="col">
```
eKoopmans, line 148: These `scope` changes aren't strictly necessary, just good hygiene (borrowed from the draft PR that was adding support for sticky + scroll-wrapper).

components/table/README.md
```diff
@@ -299,3 +299,87 @@ If your table supports row selection, apply the `selected` attribute to `<tr>` r
...
+      }
+    }
+
+    static get styles() {
+      return tableStyles;
+    }
+
+    constructor() {
+      super();
+      this._checked = [ true, false ];
```
eKoopmans, line 331: This is fairly basic "checked" logic for the readme - I could instead model the more standard pattern of something like `this._data = [ {selected: true, ...}, ... ]`, but I thought that was maybe overkill here.

components/table/test/table.test.js
```diff
@@ -29,3 +30,17 @@ describe('d2l-table-wrapper', () => {
...
+
+	it('should construct', () => {
+		runConstructor('d2l-table-header');
+	});
+
+	// it('should override default SelectionHeader label', async() => {
+	// 	const el = await fixture(html`<d2l-table-header></d2l-table-header>`);
+	// 	const section = el.shadowRoot.querySelector('section');
+	// 	expect(section.getAttribute('aria-label')).to.equal('Table controls');
+	// });
```
eKoopmans, line 44: @svanherk I've added this in to match the testing in your PR.

My plan is to actually leave this as commented for this PR, and open another PR afterward to uncomment this test and add langterms etc. That way this PR won't get even bigger 😄.

eKoopmans (commented):
This PR will have a merge conflict once #3040 is merged, because we're both tinkering with the internals of `d2l-selection-header`.

components/table/README.md
```diff
@@ -299,3 +299,87 @@ If your table supports row selection, apply the `selected` attribute to `<tr>` r
...
+            <thead>
+              <tr>
+                <th><d2l-selection-select-all></d2l-selection-select-all></th>
+                <th>Column B</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr selected>
+                <td>
+                  <d2l-selection-input selected key="a" label="a" ?checked="${this._checked[0]}" @d2l-selection-change="${() => this._selectRow(0)}"></d2l-selection-input>
```
dbatiste, line 351: I think in practice the linter/analyzer might complain about calling functions in handlers defined in templates. I think it's a perf thing to avoid the recreation of functions when the templates re-render. To avoid it, we'd normally use the same handler for all the `d2l-selection-input`, and use `e.target.key` to do the necessary look-up.

oelhanafey (commented):
The demo page seems to be working locally so I don't think it is anything with your code here, but it looks like the table demo page on the PR site is not working properly:
https://pr-3051-brightspace-ui-core.d2l.dev/components/table/demo/table.html

components/selection/selection-header.js
```diff
@@ -172,6 +163,18 @@ export class SelectionHeader extends SelectionObserverMixin(RtlMixin(LocalizeCor
 		this._hasActions = (e.target.assignedNodes({ flatten: true }).filter(node => node.nodeType === Node.ELEMENT_NODE).length > 0);
 	}
 
+	_renderSelection() {
```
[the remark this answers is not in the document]
eKoopmans: My plan was to keep this as the default implementation. The `d2l-selection-header` currently works as-is, making it empty/abstract would break that.

I know there was some debate about whether it's desirable or not for `d2l-selection-header` to exist standalone, though I can say that it's used directly in our selection demos (on the readme and the internal dev demos). It was a perfect drop-in replacement for the custom-made "selection header" we had in the examples previously.

components/table/test/table.visual-diff.js
```diff
@@ -125,6 +127,25 @@ describe('d2l-table', () => {
 
 					});
 
+					describe('table-header', () => {
+
+						before(async() => {
+							await page.$eval('d2l-test-table-visual-diff', elem => elem.setAttribute('hidden', 'hidden'));
+							await page.$eval('d2l-test-table-sticky-visual-diff', elem => elem.setAttribute('hidden', 'hidden'));
+							await page.$eval('d2l-test-table-header-visual-diff', elem => elem.removeAttribute('hidden'));
+						});
```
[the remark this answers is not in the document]
eKoopmans: This one I'm not sure about - I just followed the pattern that was already there. The sticky tests could potentially cause conflicts, though it doesn't seem likely.

@dlockhart the existing tests have your name on it (though it's just "migrate to Lit"), do you know more about why these are selectively shown/hidden?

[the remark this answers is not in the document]
dlockhart: Hmm unfortunately I really don't recall... yeah possibly an artifact from before the migration to Lit and/or sticky stuff.

oelhanafey (commented):
LGTM! But I'll let someone else have a look and approve it

components/table/table-header.js
```diff
@@ -0,0 +1,21 @@
+import { html, nothing } from 'lit';
+import { SelectionHeader } from '../selection/selection-header.js';
+
+/**
+ * A header for table components containing a selection summary and selection actions.
+ * @slot - Responsive container using `d2l-overflow-group` for `d2l-selection-action` elements
```
dlockhart, line 6: Don't think this slot description should be here? Looks like it was copied from `selection-header.js` to list as well.

components/table/table-header.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+class TableHeader extends SelectionHeader {
+	_renderSelection() {
+		return html`
+			<d2l-selection-summary
+				aria-hidden="true"
+				no-selection-text="${this.localize('components.selection.select-all')}"
+			>
+			</d2l-selection-summary>
+			${this.selectAllPagesAllowed ? html`<d2l-selection-select-all-pages></d2l-selection-select-all-pages>` : nothing}
```
dlockhart, line 16: `d2l-selection-select-all-pages.js` should likely be explicitly imported, even though we're getting it for free from `selection-header.js`.

eKoopmans:
Ping - any additional feedback on this one?

dlockhart (approved):
Looks excellent to me

PR #3059, opened 2022-12-07
US146148 - Remove d2l-filter-tags custom label and empty state

svanherk:
We're not _totally_ sure how these two features made it into filter tags, since there was never a design for them. We think it was because the labs component had them, which was based off an old Wave design.
This PR:
- Removes the ability to set a custom, visible label
  - No one is using this except the LMS test page, which will be [removed shortly](https://github.com/Brightspace/lms/pull/31085)
- Removes the "No Active Filters" text that displayed on empty
  - Now, when there are no active filters, nothing is rendered


components/filter/demo/filter-tags.html
```diff
@@ -73,16 +73,16 @@
...
 			<d2l-demo-snippet>
 				<d2l-filter-tags filter-ids="core-filter core-filter-2"></d2l-filter-tags>
 			</d2l-demo-snippet>
 
-			<h2>Custom label and registered to only the first filter</h2>
+			<h2>Registered to only the first filter</h2>
 
 			<d2l-demo-snippet>
-				<d2l-filter-tags filter-ids="core-filter" label-text="Filters selected:"></d2l-filter-tags>
+				<d2l-filter-tags filter-ids="core-filter"></d2l-filter-tags>
```
svanherk, line 85: This was actually using the wrong attribute name

components/filter/filter-tags.js
```diff
@@ -91,25 +56,17 @@ class FilterTags extends RtlMixin(LocalizeCoreElement(LitElement)) {
...
-					clearable
-					clear-focus-timeout="${CLEAR_TIMEOUT}"
-					@d2l-tag-list-clear="${this._clearFiltersClicked}"
-					description="${this.label || this.localize('components.filter.activeFilters')}">
-					${tagListItems}
-				</d2l-tag-list>
-			`;
-		}
+
+		if (numActiveFilters === 0) return nothing;
```
svanherk, line 60: Already discussed with Jeff whether we want any offscreen text here, but we decided this would be the cleaner experience

components/filter/test/filter-tags.visual-diff.js
```diff
@@ -35,7 +35,7 @@ describe('d2l-filter-tags', () => {
 
 			before(async() => {
 				await page.$eval(selector, async(elem, width) => {
-					elem.parentNode.style.width = `${width + 140}px`; // account for label
+					elem.parentNode.style.width = `${width + 140}px`; // account for filter
```
svanherk, line 38: From what I can tell, this is leaving room for the filter, not the label - this test isn't using the component with the custom label

components/filter/test/filter-tags.visual-diff.js
```diff
@@ -110,7 +110,7 @@ describe('d2l-filter-tags', () => {
 			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false, clip: rect });
 		});
 
-		it('is correct after clicking Clear All with label', async function() {
+		it('is correct after clicking Clear All with two filters', async function() {
```
svanherk, line 113: Even with the label gone, I think this test is a good one to have to make sure both filters were cleared.

dlockhart (approved):
LGTM

components/filter/filter-tags.js
oelhanafey: Would it be worth rendering some whitespace that is the same height as a filter-tag instead of nothing so that when filters are added or removed the page doesn't jitter up or down?

svanherk: Yeah we also talked about this - Jeff specifically didn't want to save space for it.  Ideally he wanted an animation to make that jitter a bit smoother, but after investigating I'm going to throw that in a separate story.  It's not very straightforward because `tag-list` isn't responsible for actually removing the `tag-list-item`s, and the removal happens immediately.  I think we need to store the height, trigger an animation to no height, and _then_ remove the tags from the `slot`.

svanherk: Added this as [US146439](https://rally1.rallydev.com/#/15545167705ud/search?detail=%2Fuserstory%2F673054076481&forbidFormattedIdRedirect=true&keywords=filter%20tag)

PR #3064, opened 2022-12-08
feat: support additional label for input units

dlockhart:
This change is to support a use case for having a denominator-based number input where visually the unit is a shorthand (e.g. `/5` or `px`) but the accessible `aria-label` needs to be longer (e.g. `out of 5` or `pixels`).

It ends up looking like this:
Screen Shot 2022-12-08 at 1 58 33 PM

If we see this pattern repeat itself in more places, we could introduce a `<d2l-input-denominator value="3" max="5">` to simplify things -- for now this does the job though.


components/inputs/demo/input-number.html
```diff
@@ -98,6 +98,18 @@ <h2>Trailing Zeroes</h2>
...
 			</d2l-demo-snippet>
 
+			<h2>Denominator</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-number max="5" min="0" label="Grade" label-hidden required unit="/5" unit-label="out of 5" value="3"></d2l-input-number>
+					<style>
+						d2l-input-number {
+							--d2l-input-text-align: end;
+						}
```
dlockhart, line 108: @tdevos-d2l this is how you can right-align the input value text.

components/inputs/input-text.js
```diff
@@ -454,12 +459,7 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 		super.updated(changedProperties);
 
 		changedProperties.forEach((oldVal, prop) => {
-			if (prop === 'unit') {
-				if (this.unit && this.unit !== '%') {
-					throw new Error('Invalid unit value for d2l-input-text.');
```
dlockhart, line 459: We now need to relax the accepted values for `unit`. I think the reason we were worried about opening this up was that consumers would provide values that made no sense when read with the label. `unit-label` solves that.

components/inputs/input-text.js
```diff
@@ -476,7 +476,10 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 			label = this.getAttribute('aria-label');
 		}
 		if (label) {
-			const unitLabel = this.unit ? ` ${this.unit}` : '';
+			let unitLabel = '';
+			if (this.unit) {
+				unitLabel = this.unitLabel ? ` ${this.unitLabel}` : this.unit;
```
dlockhart, line 481: Slight variation here -- `unit-label` has a space between the label and the unit (e.g. "3 pixels", "3 out of 5") vs. when just `unit` is present there is no space (e.g. "3%"). We may need to make this more flexible in the future.

components/inputs/demo/input-number.html
tdevos-d2l: I think you made *all* the input number demos right-aligned with this 😄

components/inputs/input-text.js
dbatiste: I wonder if we should make the `unit-label` required if `unit` is provided and it's not `%`. I can't really think of any other abbreviations that the browser/screen reader would automatically handle.

components/inputs/demo/input-number.html
dlockhart: 🤦 yes, yes I did.

components/inputs/input-text.js
dlockhart: I like that idea -- will update.

components/inputs/demo/input-number.html
```diff
@@ -98,6 +98,18 @@ <h2>Trailing Zeroes</h2>
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>Denominator</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-number max="5" min="0" label="Grade" label-hidden required unit="/5" unit-label="out of 5" value="3"></d2l-input-number>
```
dbatiste, line 104: I was playing around with the value. Depending on the unit chosen, the consumer may very likely need to use `input-width`... since the default may not be sufficient.
image

It looks like `d2l-input-percent` doesn't actually have enough width when the value is 100. 😞 . That's maybe separate from the issue at hand here.
image

Aside from that, I notice that the font-size for the value and the unit are different - 16px and 19px respectively. Looks a little odd.
image

dbatiste: Changing the default width at this point may or may not be safe.

dlockhart: I noticed the font-size thing too but assumed it was intentional. Could explain why `100` doesn't fit in the percent input. I'll investigate what happened there.

components/inputs/input-text.js
dlockhart: I've added validation to ensure that `unit-label` is set whenever `unit` isn't `%`.

For the space vs. no-space, I think this was a bug actually. The existing code put a space when the label wasn't hidden, but didn't include the space for `label-hidden`. Because it was always just `label%` or `label %` it likely wasn't noticed. I've updated it to always use a space.

components/inputs/demo/input-number.html
dlockhart: Looking at the [original visual diff](https://github.com/BrightspaceUI/core/blob/1688dbcffdc34f91fad3dc6f56349e0ddcc02d94/components/inputs/test/screenshots/ci/golden/input-percent/d2l-input-percent-default-value.png), I think this was always an issue.

Having said that, I'm treating it as a bug since it's way more obvious now that the unit is multiple characters. I've updated the unit to also be `0.8rem`. Looks much better now:

Screen Shot 2022-12-08 at 1 43 47 PM

components/inputs/input-text.js
dlockhart: Updated this to now require `unit-label` unless `unit` is `%`.

components/inputs/input-text.js
```diff
@@ -618,5 +632,19 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 
 	}
 
+	_validateUnit() {
+		clearTimeout(this._validatingUnitTimeout);
+		// don't error immediately in case it doesn't get set immediately
+		this._validatingUnitTimeout = setTimeout(() => {
```
dbatiste, line 638: This turned out to be more complex than I imagined... sorry! 😬 

It looks like `_missingUnitLabelErrorHasBeenThrown` never gets reset once it's been set to `true`. Maybe we want to return early from this method if it's `true`?

```javascript
if (this._missingUnitLabelErrorHasBeenThrown) return;
this._validatingUnitTimeout = setTimeout(() => {
    ...
}, 3000);
```

Not a big deal though, since `_validateUnit()` only gets called when `unit` or `unitLabel` changes, and we probably never expect it to get called again once those properties are set.

dbatiste (approved):
Looks good!

components/inputs/input-text.js
dlockhart: Yeah, this kind of validation is tricky because the calls to `updated` happen in the order that attributes are set.

I can bail early -- that would prevent checking in future calls, but it doesn't save much during component initialization since we still need to check `_missingUnitLabelErrorHasBeenThrown` inside the callback.

A typical scenario will be:
- `updated` gets called with the one attribute, we validate
- `updated` gets called with the other, we validate
- `firstUpdated` gets called, we validate
- 3 seconds elapses, so the first callback runs, then the second, then the third

dbatiste: Wouldn't the first and second callbacks get cancelled due tot the `clearTimeout`?

components/inputs/demo/input-number.html
tdevos-d2l: Going back to the width... would there be any risk to me making the `input-width` based on `ch` units. So if there are 2 digits in the denominator, we want to make room for at least 2 digits in the numerator. 
```
inputWidth = (scoreDenominator.toString().length * 2) + 4 // Add a little extra for padding and the "/"
...
input-width="${inputWidth}ch"
```
Ends up looking like this:
image
image
image
image

components/inputs/input-text.js
dlockhart: Hmm, yep you're right -- moved the check up higher as you suggested.

components/inputs/demo/input-number.html
tdevos-d2l: Hmm yeah. I think our default font's character width for numbers is identical, but I cant guarantee it'll be like that in all cases. I'll have to get creative here or just make it oversized and have it look less good.

Minor nit, but can we have the invalid icon float left when text is right aligned? It defaults to float right and can cut off text when there is a validation error.

dlockhart: That's an excellent idea... but I think that'll require introducing an actual `alignment` attribute instead of relying on the CSS variable. :(

tdevos-d2l (approved):

components/inputs/demo/input-number.html
dlockhart: I'll explore this separately if I have time tomorrow, but I'll merge this for now so you can start using it. It's tricky as-is because the validation icon already aligns itself in a weird way based on a combination of the unit width and the RTL-ness of the page.

PR #3066, opened 2022-12-08
fix: Add offsetParent ponyfill to restore legacy behaviour

eKoopmans:
This PR adds a `offsetParent` ponyfill that restores the old (pre-Chrome-v109) behaviour, and switches `floating-button` to use it. This should unblock core for now, but we'll need further investigation on the broader impact of the `offsetParent` browser change on core and the org at large.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/673051275521&fdp=true


dbatiste (approved):
Looks good. It might be good to add a comment in the ponyfill file that references the original code and also the ticket that Dave opened in case anyone ever wonders why the heck we have that. 😄

dbatiste:
> Looks good. It might be good to add a comment in the ponyfill file that references the original code and also the ticket that Dave opened in case anyone ever wonders why the heck we have that. 😄

Though I see you do have all that detail in the Rally issue, so could find it is we track things back to this PR.

eKoopmans:
> It might be good to add a comment in the ponyfill file that references the original code and also the ticket that Dave opened

Yep was already working on that, thanks 😄.

PR #3070, opened 2022-12-09
feat: introduce value-align for text inputs

dlockhart:
Following up from #3064, this adds a `value-align` attribute to text and number inputs. This removes the need for the `--d2l-input-text-align` CSS variable, and by having it be an attribute we can now move the validation icon to the opposite side of the value text.

Screen Shot 2022-12-09 at 10 00 03 AM


components/inputs/input-number.js
```diff
@@ -331,6 +337,7 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 	}
 
 	render() {
+		const valueAlign = (this.valueAlign === 'end') ? 'end' : 'start';
```
dlockhart, line 340: The validation stuff hates it if you just assign an arbitrary value to an enum field. This seems to help it.

tdevos-d2l (approved):
That was quick, thank you! :shipit:

tdevos-d2l:
Something that I forgot to bring up in the last PR, should the unit have a different colour than the editable text? Makes the user think they should be able to modify the unit.

components/inputs/input-text.js
```diff
@@ -401,8 +410,8 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 		const lastSlotName = (this.dir === 'rtl') ? 'left' : 'right';
 
 		const isValid = ariaInvalid !== 'true' || this.disabled;
-		const invalidIconSide = (this.dir === 'rtl') ? 'left' : 'right';
-		const invalidIconOffset = Math.max((this.dir === 'rtl') ? this._firstSlotWidth : this._lastSlotWidth, 12);
+		const invalidIconSide = ((this.dir === 'rtl' && this.valueAlign === 'start') || (this.dir !== 'rtl' && this.valueAlign === 'end')) ? 'left' : 'right';
+		const invalidIconOffset = Math.max((invalidIconSide === 'left') ? this._firstSlotWidth : this._lastSlotWidth, 12);
```
dlockhart, line 414: All this logic would be a lot simpler I think if we used logical CSS properties for everything, but I wasn't quite ready for that scale of change.

dbatiste (approved):
Looks good!

dbatiste:
I think there are cases (ex. -1002 below) where it's not obvious the number is being hidden until they focus, but it is possible to produce the same sorta thing with the value aligned to the start as well.

tdevos-d2l:
Is the position of the text that's input being recalculated when the `unit` changes? I'm going from `unit='/9'` to `unit='/1112'`
I know it's not common for the unit to be changed, but in our (CE) use case it *can* happen 😬 

https://user-images.githubusercontent.com/90270112/206801572-a3ca34d6-2fd4-47db-a4eb-4f2655dd20dd.mp4

dlockhart:
> Is the position of the text that's input being recalculated when the unit changes?

It isn't, no -- good catch. I'll follow up with a bug fix PR.

PR #3072, opened 2022-12-09
fix: File pattern was excluding table-test.js

eKoopmans:
The table demo was broken on the static build (PR previews) - turns out it's because `table-test.js` was being excluded from the build. I've changed the pattern to only exclude `*.test.js`, instead of `*test.js`.


svanherk (approved):
Ha, that's funny!

PR #3074, opened 2022-12-12
fix: handle unit change in d2l-input-text

dlockhart:
The position of a value within `<d2l-input-text>` is offset by the size of the right/left slot but also the optional unit. We listen for a `slotchange` to update that position, but because the unit isn't inside the slot, if the unit changes we don't update the value position.

That results in ugly things:
Screen Shot 2022-12-12 at 10 58 40 AM

This first introduces a visual-diff test to demonstrate the failure, then I'll add a fix for it.


README.md
```diff
@@ -133,10 +133,10 @@ The golden snapshots in source control must be updated by the [visual-diff GitHu
 To run the tests locally to help troubleshoot or develop new tests, first install these dependencies:
 
 ```shell
-npm install @brightspace-ui/visual-diff@X mocha@Y puppeteer@Z  --no-save
+npm install @brightspace-ui/visual-diff@X  --no-save
```
dlockhart, line 136: These instructions are simpler now that we're using peer dependencies. [update to our npm init library](https://github.com/BrightspaceUI/create/pull/66).

tdevos-d2l (approved):
Thanks! Any word on changing the unit's styling/colour?

dbatiste (approved):

eKoopmans (approved):

dlockhart:
> Any word on changing the unit's styling/colour?

Yep, Jeff has a design for it and I'll put up a PR for that change after this one. Wanted to keep them separate since they'll both affect visual-diff.

svanherk (approved):

PR #3077, opened 2022-12-12
fix: make unit visually smaller, lighter and baseline aligned

dlockhart:
Making a couple visual tweaks to the unit to ensure that it looks distinct from the value.

- `0.7rem` compared to the value's `0.8rem`
- Galena compared to ferrite
- Aligned with the input value's baseline instead of vertically centred

Since the unit is absolutely positioned, `vertical-align: baseline` isn't going to accomplish our desired alignment. So I applied a top margin of `0.05rem` -- which is half the difference in font sizes (`(0.8 - 0.7) / 2`). If anyone has better suggestions, let me know! But that technique seems to get the desired effect in a bunch of different root font sizes:

Normal:
Screen Shot 2022-12-12 at 12 07 28 PM

Small:
Screen Shot 2022-12-12 at 12 08 37 PM

Large:
Screen Shot 2022-12-12 at 12 08 47 PM


tdevos-d2l (approved):
lgtm!

dbatiste (approved):

PR #3080, opened 2022-12-12
Add `clearable` to `d2l-tag-list` clear demo

devpow112:
Without the `clearable` attribute it doesn't allow removal so it doesn't show the clear actions on the daylight site.

### DIFF
--- a/components/tag-list/README.md
+++ b/components/tag-list/README.md
@@ -70,9 +70,17 @@ The `d2l-tag-list-item` provides the appropriate `listitem` semantics and stylin
     e.target.parentNode.removeChild(e.target);
     console.log(`d2l-tag-list-item-clear event dispatched. Value: ${e.detail.value}`);
   });
+
+  document.addEventListener('d2l-tag-list-clear', (e) => {
+    const items = e.target.querySelectorAll('[role="listitem"]');
+    items.forEach((item) => {
+      item.parentNode.removeChild(item);
+    });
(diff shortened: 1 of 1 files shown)


dbatiste (approved):

components/tag-list/README.md
```diff
@@ -70,9 +70,17 @@ The `d2l-tag-list-item` provides the appropriate `listitem` semantics and stylin
...
     console.log(`d2l-tag-list-item-clear event dispatched. Value: ${e.detail.value}`);
   });
+
+  document.addEventListener('d2l-tag-list-clear', (e) => {
+    const items = e.target.querySelectorAll('[role="listitem"]');
+    items.forEach((item) => {
+      item.parentNode.removeChild(item);
+    });
+    console.log('d2l-tag-list-clear event dispatched');
+  });
```
oelhanafey, line 80: The "Clear All" button won't show up in this demo since there is only a single item and it only shows up when there are 4+ items.

I think you may have meant to add an event listener for `d2l-tag-list-item-clear` here instead

devpow112: There was already an event listener for `d2l-tag-list-item-clear` but it was missing the `clearable` attribute. I added the clear all one as well just for completeness.

oelhanafey: Oops, this was added previously higher up in the code. You can disregard.

oelhanafey: Oops missed that!

PR #3082, opened 2022-12-13
feat: animate opacity in `d2l-expand-collapse-content`

jameswklassen:


components/expand-collapse/expand-collapse-content.js
```diff
@@ -35,6 +35,7 @@ class ExpandCollapseContent extends LitElement {
 	static get styles() {
 		return css`
 			:host {
+				--d2l-expand-collapse-content-transition-duration: 0.4s;
```
dlockhart, line 38: I'm sure this is very subjective, but for me I only really notice the opacity change when collapsing... when expanding, by the time the content is tall enough to recognize it, it's pretty much fully visible. So maybe a slightly longer duration? 🤷

jameswklassen: Yeah, I think we're entering subjective area 😄 I'll push something up with a longer opacity duration and see how that looks

dlockhart: Haha I can't tell the difference. But overall I agree that in the videos you provided it _feels_ a lot nicer with the fade. It feels a lot more "real" and less like we're just growing a container's height.

imcdonald-d2l (approved):

components/expand-collapse/expand-collapse-content.js
```diff
@@ -45,17 +46,29 @@ class ExpandCollapseContent extends LitElement {
 			.d2l-expand-collapse-content-container {
 				display: none;
 				overflow: hidden;
-				transition: height 400ms cubic-bezier(0, 0.7, 0.5, 1);
+				transition:
+					height var(--d2l-expand-collapse-content-transition-duration) cubic-bezier(0, 0.7, 0.5, 1),
+					opacity calc(var(--d2l-expand-collapse-content-transition-duration) * 2) cubic-bezier(0, 0.7, 0.5, 1);
```
dbatiste, line 51: I'm not certain this change (to the duration) is working the way we expect on collapse. The state of the component is controlled via the `transitionend` event, and the handler will be dispatched for each property that changes. When the first property transition ends (for `height`), the `display` is set to `none` because it thinks the transition is complete, and I think the user would not see the complete opacity transition. Not sure what happens to the opacity `transitionend` event in that case - it may never get dispatched, or maybe it would. For the expand case, the second event would certainly be dispatched 400ms later. 

I think the code is trying to handle the case of quickly toggling the state while transitioning, but it seems like the extra later event could throw off the timing in such scenarios because the second event (from the first toggle) could cause the state to change early (during the second toggle). I don't know for certain this is a problem, but currently, transitioning more than one property with different durations seems dicey. Are we sure we want that? If so, checking `e.propertyName === 'height'` in the handler may be prudent.

If we do feel we need some adjustment to the opacity transition, perhaps it is the timing function we should be changing rather than the duration.

components/expand-collapse/expand-collapse-content.js
jameswklassen: ok -- so based on [this comment](https://github.com/BrightspaceUI/core/pull/3082#discussion_r1048536206) and through an offline convo with @dbatiste, I think we should keep the durations the same. It can introduce more complexity and race conditions from the `@transitionend`.

https://github.com/BrightspaceUI/core/blob/675b5939dc191df920bbdfc9ad44e6cf0ba565c3/components/expand-collapse/expand-collapse-content.js#L85

If we want to adjust the opacity further, I think tweaking the bezier curve is a simpler path forward.

components/expand-collapse/expand-collapse-content.js
```diff
@@ -44,18 +46,34 @@ class ExpandCollapseContent extends LitElement {
...
 				display: block;
 			}
 
+			.d2l-expand-collapse-content-container[data-state="collapsed"] {
+				opacity: 0;
+			}
+
+			.d2l-expand-collapse-content-container[data-state="collapsing"] {
+				opacity: 0;
+			}
```
dbatiste, line 66: I feel like some of these selectors could be reworked to eliminate the `:not()`, but moving things around obviously carries some risk. I think we can at least collapse some of this duplication though.
```suggestion
			.d2l-expand-collapse-content-container[data-state="collapsed"],
			.d2l-expand-collapse-content-container[data-state="collapsing"] {
				opacity: 0;
			}
```

components/expand-collapse/expand-collapse-content.js
```diff
@@ -44,18 +46,34 @@ class ExpandCollapseContent extends LitElement {
...
+
 			.d2l-expand-collapse-content-container[data-state="expanded"] {
+				opacity: 1;
 				overflow: visible;
 			}
 
+			.d2l-expand-collapse-content-container[data-state="expanding"] {
+				opacity: 1;
+			}
```
dbatiste, line 75: I'm not sure which way I prefer in this case... I probably prefer grouping by property in this case but... 🤷 
```suggestion
			.d2l-expand-collapse-content-container[data-state="expanded"] {
				overflow: visible;
			}

			.d2l-expand-collapse-content-container[data-state="expanded"],
			.d2l-expand-collapse-content-container[data-state="expanding"] {
				opacity: 1;
			}
```

dbatiste (approved):

components/expand-collapse/expand-collapse-content.js
```diff
@@ -56,6 +61,16 @@ class ExpandCollapseContent extends LitElement {
 				overflow: visible;
 			}
 
+			.d2l-expand-collapse-content-container[data-state="collapsed"],
+			.d2l-expand-collapse-content-container[data-state="collapsing"] {
+				opacity: 0;
+			}
```
eKoopmans, line 67: Can you clarify why these are necessary, when the default is already 0?

components/expand-collapse/expand-collapse-content.js
```diff
@@ -44,8 +46,11 @@ class ExpandCollapseContent extends LitElement {
 
 			.d2l-expand-collapse-content-container {
 				display: none;
+				opacity: 0;
 				overflow: hidden;
-				transition: height 400ms cubic-bezier(0, 0.7, 0.5, 1);
+				transition:
+					height var(--d2l-expand-collapse-content-transition-duration) var(--d2l-expand-collapse-content-transition-function),
+					opacity var(--d2l-expand-collapse-content-transition-duration) var(--d2l-expand-collapse-content-transition-function);
```
eKoopmans, line 53: Maybe I missed this in another PR, but where is the `--d2l-expand-collapse-content-transition-function` defined? Doesn't come up on a search of core.

components/expand-collapse/expand-collapse-content.js
jameswklassen: Good catch, I think you're right! I thought I needed this initially...

components/expand-collapse/expand-collapse-content.js
```diff
@@ -44,8 +46,17 @@ class ExpandCollapseContent extends LitElement {
...
+			}
+
+			.d2l-expand-collapse-content-container[data-state="expanding"],
+			.d2l-expand-collapse-content-container[data-state="expanded"],
+			.d2l-expand-collapse-content-container[data-state="collapsing"] {
+				display: block;
 			}
```
eKoopmans, line 60: 😆 I'm going to ask the same question here - aren't these already captured under the `:not([data-state="collapsed"])`? Or is there a state where it can be both `collapsed` and e.g. `collapsing`?

jameswklassen: ughhhh.... did not mean to commit this hunk sorry.

eKoopmans (approved):

components/expand-collapse/expand-collapse-content.js
```diff
@@ -44,8 +46,11 @@ class ExpandCollapseContent extends LitElement {
...
 				display: none;
+				opacity: 0;
 				overflow: hidden;
-				transition: height 400ms cubic-bezier(0, 0.7, 0.5, 1);
+				transition:
+					height var(--d2l-expand-collapse-content-transition-duration) var(--d2l-expand-collapse-content-transition-function),
+					opacity var(--d2l-expand-collapse-content-transition-duration) var(--d2l-expand-collapse-content-transition-function);
 			}
 
 			.d2l-expand-collapse-content-container:not([data-state="collapsed"]) {
```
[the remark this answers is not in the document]
jameswklassen: ohhh...that's what you meant earlier. That seems okay to me 😅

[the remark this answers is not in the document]
dbatiste: Yeah... though earlier when we were chatting I figured it would be more finicky to change.

eKoopmans (approved):

dbatiste (approved):

oelhanafey (approved):

PR #3084, opened 2022-12-14
fix: Wait for aria-valuenow to be defined

eKoopmans:
`primary-secondary` was failing the axe test due to its divider's `aria-valuenow` being undefined when it was focusable. The issue is that the `aria-valuenow` relies on the `_size` property, which is only calculated in response to a resize-observer that triggers after the element is already rendered.

I don't love this fix (which is just to wait longer before doing the axe test), but it should work. I'm unsure whether it may introduce flake though, or whether it should always have size calculated within two frames. The alternatives are:

1. Improve the `_size` calculations to happen more immediately
2. Prevent the divider from being focusable until the size is calculated
    - this option might make sense, though our axe tests would then have a false positive of passing because the resizable version would appear as non-resizable when the test actually fires

### DIFF
--- a/templates/primary-secondary/test/primary-secondary.test.js
+++ b/templates/primary-secondary/test/primary-secondary.test.js
@@ -1,5 +1,5 @@
 import '../primary-secondary.js';
-import { expect, fixture, html } from '@open-wc/testing';
+import { expect, fixture, html, nextFrame } from '@open-wc/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
 describe('d2l-template-primary-secondary', () => {
@@ -10,6 +10,11 @@ describe('d2l-template-primary-secondary', () => {
 	].forEach((testCase) => {
 		it(`${testCase.name} should pass all aXe tests`, async() => {
 			const elem = await fixture(html`<d2l-template-primary-secondary ?resizable="${testCase.resizable}"></d2l-template-primary-secondary>`);
+
+			// Wait for size to be calculated so that aria-valuenow is defined.
+			await nextFrame();
+			await nextFrame();
+
 			await expect(elem).to.be.accessible();
 		});
 	});


dlockhart:
Could we default `__size` to something reasonable, like the "desktopMinSize" value of `320`? It looks like it ultimately gets set to 1/3 of the viewport width roughly once all the fancy calculations are done.

eKoopmans:
Not sure... There's different logic in `_onContentResize` for calculating the initial size, when `_size` is undefined. Defaulting `__size` to any value would prevent that logic from firing.

dlockhart:
Gotcha. What about just defaulting it to `320` inside `render()` so that it writes out _something_ to `aria-valuenow`?

dlockhart:
My worry about just delaying the test: we've had a bunch of 3rd party accessibility audits lately where they run out pages through aXe. So those audits would also find this issue since we're temporarily writing out incorrect ARIA.

dlockhart:
Yep that works for me!

dlockhart (approved):
Nice!

templates/primary-secondary/primary-secondary.js
```diff
@@ -921,12 +921,13 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
 
 	render() {
 		let tabindex;
+		const size = this._size ?? 0;
```
dbatiste, line 924: Can we just init `this._size = 0;` in the constructor?

dlockhart: Haha see my earlier comments...

eKoopmans: Not unless we change the [logic here](https://github.com/BrightspaceUI/core/pull/3084/files#diff-06cfe552d3f9e7117999c057f42fd8b028a379c2a88a4004b6cc8ab40b63285fR1044) too, which expects `this._size` to be undefined to trigger its initial calculations.

eKoopmans: If we prefer that route I can do that, and just add an extra boolean flag like `this._sizeIsInitialized = false` - do we like that better?

dlockhart: I'm good either way.

PR #3087, opened 2022-12-16
Add secondary-first property to template-primary-secondary

oelhanafey:
[US145714](https://rally1.rallydev.com/#/?detail=/userstory/671677824631&fdp=true)

This PR introduces the `secondary-first` property to the `<d2l-template-primary-secondary>` component. `secondary-first` will render the secondary slot on the left side instead of the right side, or on the right side when in RTL. This property only has effect when in Desktop mode (`max-width > 767px`). The mobile view is unaffected.

The resize logic for this component also needed to be updated to account for the secondary container being on the opposite side.


templates/primary-secondary/primary-secondary.js
```diff
@@ -49,6 +49,7 @@ class Resizer {
 		this.isMobile = false;
 		this.panelSize = 0;
 		this.isRtl = false;
+		this.secondaryFirst = false;
```
oelhanafey, line 52: This attribute isn't used for the Mobile Resizers so this could be added directly to the two Desktop Resizers but I found it a bit cleaner up here with `isRtl`

templates/primary-secondary/primary-secondary.js
```diff
@@ -112,15 +113,18 @@ class DesktopKeyboardResizer extends Resizer {
...
-		if (e.keyCode !== leftKeyCode && e.keyCode !== rightKeyCode) {
+		if (e.keyCode !== keyCodes.LEFT && e.keyCode !== keyCodes.RIGHT) {
 			return;
 		}
 		e.preventDefault();
+
+		// The direction of the container is flipped if exactly one of isRtl and secondaryFirst is true
+		const isFlipped = this.isRtl ^ this.secondaryFirst;
+		const direction = e.keyCode === keyCodes.LEFT && !isFlipped
+			|| e.keyCode === keyCodes.RIGHT && isFlipped ? 1 : -1;
```
oelhanafey, line 124: The resizer works by adjusting the width of the secondary container. If the page is in RTL then the placement of the secondary container is flipped but if `secondaryFirst` is also active then it will flip back to the original position. So the direction will be flipped when exactly one of `isRtl` and `secondaryFirst` is set to true.

templates/primary-secondary/primary-secondary.js
```diff
@@ -180,7 +183,8 @@ class DesktopMouseResizer extends Resizer {
 
 	_computeContentX(clientX) {
 		const x = clientX - this.contentRect.left;
-		return this.isRtl ? x : this.contentRect.width - x;
+		// The direction of the container is flipped if exactly one of isRtl and secondaryFirst is true
+		return this.isRtl ^ this.secondaryFirst ? x : this.contentRect.width - x;
```
oelhanafey, line 187: Similar logic to the `DesktopKeyboardResizer`

templates/primary-secondary/primary-secondary.js
```diff
@@ -932,11 +943,19 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
 			'd2l-template-scroll': isWindows
 		};
 		const keyboardHelpText = this._isMobile ? this.localize('templates.primary-secondary.keyboardVertical') : this.localize('templates.primary-secondary.keyboardHorizontal');
+		const primarySection = html`<main class="${classMap(scrollClasses)}"><slot name="primary"></slot></main>`;
+		const secondarySection = html`
+			<div style=${styleMap(secondaryPanelStyles)} class="d2l-template-primary-secondary-secondary-container" @transitionend=${this._onTransitionEnd}>
+				<div class="d2l-template-primary-secondary-divider-shadow"></div>
```
oelhanafey, line 949: This shadow only exists when in Mobile view and `secondary-first` does not have any effect in Mobile view so this would not be shown when placed in the first section.

templates/primary-secondary/test/primary-secondary.axe.js
```diff
@@ -0,0 +1,23 @@
...
+
+			it(`${resizableTest.name} ${secondaryFirstTest.name} should pass all aXe tests`, async() => {
+				const elem = await fixture(html`<d2l-template-primary-secondary ?resizable="${resizableTest.resizable}" ?secondaryFirst="${secondaryFirstTest.secondaryFirst}"></d2l-template-primary-secondary>`);
+				await expect(elem).to.be.accessible();
+			});
+
+		});
+	});
+
+});
```
oelhanafey, line 23: I separated the aXe tests into a separate `axe.js` file so that these tests run in the correct test group.

dlockhart (approved):
Played around with this on the preview site, looks really good to me!

dbatiste:
This seems pretty good to me. One question... what is the expected focus order? Should focus go to the main panel first, or should it go to the secondary panel first? Generally, focus order should match the visual order, but it's not a hard rule. If focus order should go to the main panel first, `flex-direction: row-reverse;` could be used instead of flipping the DOM order. Just figured I'd mention, in case Design expected focus to still go to the main panel first.

oelhanafey:
@dbatiste That's a great question that I hadn't considered. I ran it by Jeff and he said that having it match the visual order is the desired behaviour.

dbatiste (approved):

PR #3097, opened 2022-12-19
Fix More-Less bug where it was not setting base height correctly 

oelhanafey:
[US146805](https://rally1.rallydev.com/#/?detail=/userstory/673763145497&fdp=true)

This PR removes the `__init_setBaseHeight()` call on the first update of `<d2l-more-less>`. This function was being called twice on initialization. Once from `firstUpdated()` and once from `__adjustToContent()`. `__init_setBaseHeight()` calls `this.__init_measureBaseHeight()` asynchronously with `requestAnimationFrame()` and was causing some odd behaviour where the `__baseHeight` was not being set correctly specifically when using an explicit height attribute.

Removing the call from `firstUpdated()` ensures the function is only called once and the baseHeight is set as expected. I removed it from `firstUpdated()` rather than `adjustToContent()` since it ensures that `adjustToContent()` does not run until the base height is set. `adjustToContent()` is responsible for resizing the component and determining whether the more-less button should be rendered.

I also just changed the code demo to use "Lorem ipsum" to be consistent with the other demos on the page.

### DIFF
--- a/components/more-less/demo/more-less.html
+++ b/components/more-less/demo/more-less.html
@@ -30,7 +30,7 @@
 			<d2l-demo-snippet>
 				<template>
 					<d2l-more-less height="8em">
-						<p id="img-target">Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody!Hi everybody! <a href="">Hi!</a></p>
+						<p id="img-target">Lorem ipsum dolor sit amet, consectetur adipiscing elit, magna aliqua quis. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident <a href="">proident.</a></p>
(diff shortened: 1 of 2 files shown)


eKoopmans (approved):
Looks good. I think I reproduced the issue you're talking about - is it something like this?

image

Steps:
1. Shrink the page size so that the more-less needs to have the "More" expander
2. Refresh
3. Expand the page again, and the "More" expander should disappear
4. In the old behaviour, it's still there, but doesn't do much when clicked (makes it a bit smaller)

oelhanafey:
@eKoopmans yeah I was just typing up how to reproduce it. That is exactly the problem. It should not be shown here since the component is shorter than the `8em` height attribute. The reason it is being rendered because it is more than `4em` and it thinks the base height is the default `4em` instead of the `8em` that was set. I'm not sure why that only happens on refresh, some weirdness from when it set twice I suppose.

PR #3099, opened 2022-12-19
Make primary-secondary resizing more responsive

oelhanafey:
[US146328](https://rally1.rallydev.com/#/?detail=/userstory/672770428197&fdp=true)

In `<template-primary-secondary>`, the secondary panel can be difficult or confusing to expand when it's in a fully collapsed state, since dragging has no immediate reaction, and can result in no action if 'dropped' when it is less than half way to open.

**Changes that have been made:**

1. Attempting to resize the secondary panel will now follow the mouse even when the width is less than the minimum secondary panel width or height. It will only auto collapse or expand when you stop resizing (`mouseup` or `touchend`).

2. The thresholds where the panel will expand or collapse have been changed:
**In the desktop view and mobile view with mouse:**
    - When collapsed, the secondary panel will expand when pulled > 10% of the minimum width/height.
    - When expanded, the secondary panel will collapse when pulled less than 75% of the minimum width/height.

    **Mobile touch view:**
There are 3 snap points now, max height, min height and fully collapsed. Thresholds are set at 10% of the distance to the next snap point. 
E.g. If you are at max width. Pulling down > 10% of the distance to min width will snap to the min width, but less than that will snap back up to the max width.

3. In mobile view with a touch screen, you can now use the divider to resize the panel in addition to scrolling on the secondary panel itself.

4. This PR also contains a fix for an issue where the divider arrow when expanded was not facing the right direction when using `secondary-first`


dlockhart:
Interesting! IMO it's better, but now it has a new problem in that I don't know how far I need to enlarge the secondary panel before it "sticks", if that makes sense? It took me a bunch of tries to actually open it enough for it to stay opened.

Before, while it didn't visually show any progress, once you hit the "sticking" point (in both directions) it would either fully expand or fully collapse on its own.

I wonder if there's maybe a middle ground somewhere? Like visually show progress like in this PR, but also have it "click" into place once you open/close it past a certain point?

oelhanafey:
@dlockhart  Yeah Jeff and I were chatting about this problem and we thought the user might be able to figure it out after a few tries but I really like your idea! Maybe we could make it so that if you expand it passed half the minimum width it will expand and if it is less than half the minimum width it will collapse.

templates/primary-secondary/primary-secondary.js
```diff
@@ -672,10 +733,12 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
...
 				display: block;
 			}
-			:host(:not([dir="rtl"])) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-left {
+			:host(:not([dir="rtl"]):not([secondary-first])) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-left,
+			:host([dir="rtl"][secondary-first]) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-left {
 				display: none;
 			}
-			:host([dir="rtl"]) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-right {
+			:host(:not([dir="rtl"])[secondary-first]) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-right,
+			:host([dir="rtl"]:not([secondary-first])) [data-is-expanded] .d2l-template-primary-secondary-divider-handle-right {
```
oelhanafey, line 749: This fixes an issue where the divider arrow when expanded was not facing the right direction when using `secondary-first`

templates/primary-secondary/primary-secondary.js
```diff
@@ -974,7 +1037,7 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
 								</d2l-icon-custom>
 							</div>
 							<div class="d2l-template-primary-secondary-divider-handle-mobile">
-								${size === 0 ? html`<d2l-icon icon="tier1:chevron-up"></d2l-icon>` : html`<d2l-icon icon="tier1:chevron-down"></d2l-icon>`}
+								<d2l-icon icon=${size === 0 ? 'tier1:chevron-up' : 'tier1:chevron-down'}></d2l-icon>
```
oelhanafey, line 1048: Since we added support for scrolling using the divider in mobile view on a touch screen, the resizing was being cancelled due to the `d2l-icon` DOM element being swapped out. This makes sure the icon changes but the DOM element is kept the same.

templates/primary-secondary/primary-secondary.js
```diff
@@ -1003,16 +1074,20 @@ class TemplatePrimarySecondary extends FocusVisiblePolyfillMixin(RtlMixin(Locali
...
-			this._secondary = this.shadowRoot.querySelector('aside');
+		if (!this._divider) {
 			this._divider = this.shadowRoot.querySelector('.d2l-template-primary-secondary-divider');
 		}
+		if (changedProperties.has('_isMobile') && this._isMobile) {
+			this._secondary = this.shadowRoot.querySelector('aside');
+			if (this._divider.isConnected && this._secondary.isConnected) {
+				this._mobileTouchResizer.connect(this._divider, this._secondary);
+			}
+		}
```
oelhanafey, line 1093: This solves an issue where going between mobile and desktop mode would create a new secondary panel DOM element so the resizing would no longer work. Now when switching to mobile view, it will re-grab the secondary panel to send to the resizer. Though this was not an issue previously, so I'm not sure what changed.

dlockhart (approved):
Really liking how this came together and the experience in the demo, nice!

PR #3108, opened 2022-12-22
check for rootView before accessing it

mdulat:
Check for existance of `rootView` before querying it.
Result of [error in activities CI](https://github.com/BrightspaceHypermediaComponents/activities/actions/runs/3754177963/jobs/6389879264#step:7:446).


mdulat:
Actually I think if `rootView` is null then we have bigger problems.

PR #3109, opened 2022-12-22
fix: Lock to rollup 3.7.5 until defect is fixed

eKoopmans:
I've confirmed that rollup@3.8.0 has introduced a defect for the specific combination of settings we use (`preserveModules` + `resolve()` + `lit`), and filed an issue:
https://github.com/rollup/rollup/issues/4765

This defect won't affect BSI because it's locked on rollup@2 currently, and may not be a widespread issue because of the specific combination of settings involved.

In the meantime, locking to 3.7.5 will unblock the PR previews again. I'll open a draft PR to restore rollup's version so we remember to unlock it when the defect is resolved.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "lit-analyzer": "^1",
         "messageformat-validator": "^2",
         "node-sass": "^8",
-        "rollup": "^3",
+        "rollup": "3.7.5",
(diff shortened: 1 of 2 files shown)


svanherk (approved):

oelhanafey (approved):
Thanks for looking into this!

PR #3110, opened 2022-12-22
fix: Restore rollup to latest version

eKoopmans:
Follow-up to #3109.

When the Rollup defect has been resolved, update this to grab the latest version and merge. Can verify it's working when the PR preview builds correctly.


oelhanafey:
This appears to be fixed now in [Rollup v3.8.1](https://github.com/rollup/rollup/releases/tag/v3.8.1)

They even reference you in the PR 😄 I think you'll need to update the package-lock

dlockhart:
Sorry, didn't notice this or your original locking of the version when I merged #3120 this morning!

svanherk:
Looks like the version bump happened in https://github.com/BrightspaceUI/core/pull/3120, but I think we still want to switch it back to `^3` instead of we want the auto-update workflow to get new minors and patches

dlockhart (approved):

dlockhart:
Yeah I normally don't look too closely at those Dependabot PRs assuming they pass, so I didn't notice that the actual change didn't have the `^` in it, my bad.

PR #3122, opened 2023-01-03
Add lang utils - helps lint and format lang term additions and changes

imcdonald-d2l:
Adding a couple commands my team has taken advantage of across our repos. Helps reduce painful copy/paste exercises as well as catches any missing lang terms across files. Let me know if you have any concerns. We have used this for awhile on my team with no issues. 

This PR:
- adds the commands for formatting/linting
- I ran `npm run format:lang` which sorted all the lang terms alphabetically (no changes have actually been made)
- Adds util for checking for unused lang terms - warns user so they can double check and manually delete


imcdonald-d2l:
Still getting unused check working... not ready quite yet.

imcdonald-d2l:
Output of `npm run lint:lang-unused-check`. Util will not actually error out or delete anything, will just warn user so they can double check and manually delete. 

image

cli/check-unused-langterms.js
```diff
@@ -0,0 +1,77 @@
+import { dirname, join, resolve } from 'path';
```
imcdonald-d2l, line 1: File is copy of rubrics usage here: https://github.com/Brightspace/d2l-rubric/blob/main/cli/check-unused-langterms.js

imcdonald-d2l:
I can open a separate PR deleting unused terms if this is approved to be merged.

cli/check-unused-langterms.js
dlockhart: How difficult do you think it would be to add this to the existing `mfv` library? I'm sure @bearfriend would be interested...

eKoopmans (commented):
Looks good to me... only concerns are that the `check-unused-langterms` might be overly noisy (lists all langterms, whether they're found or not), and that it doesn't actually have a functional effect i.e. failing when issues are found, which linting typically would.

cli/check-unused-langterms.js
imcdonald-d2l: Not entirely sure how hard it would be, not super familiar with the `mfv` library. Could potentially be a good inspiration sprint idea?

PR #3129, opened 2023-01-04
fix: Broken docs for table

eKoopmans:
This PR:

- makes several changes to the table readme:
    - live demos now target the underlying components instead of their wrappers (requires [documentation#1175](https://github.com/BrightspaceUI/documentation/pull/1175) to be merged)
    - added `display:block` to table-header demo to improve appearance
    - minor renames to avoid collision with existing `d2l-test-table`
    - removes unused `name:d2l-test-table` on the responsive demo
- adds `DemoPassthroughMixin` to simplify how properties are passed through in `d2l-test-table`
    - this was originally going to be used for documentation too (in a modified form), but is no longer necessary there

Rally: https://rally1.rallydev.com/#/?detail=/userstory/675086171663&fdp=true

### DIFF
--- a/components/table/table-header.js
+++ b/components/table/table-header.js
@@ -6,7 +6,7 @@ import { SelectionHeader } from '../selection/selection-header.js';
 /**
  * A header for table components containing a selection summary and selection actions.
  */
-export class TableHeader extends SelectionHeader {
+class TableHeader extends SelectionHeader {
 	static get properties() {
 		return {
 			/**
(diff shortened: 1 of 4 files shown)


components/table/table-header.js
```diff
@@ -6,7 +6,17 @@ import { SelectionHeader } from '../selection/selection-header.js';
 /**
  * A header for table components containing a selection summary and selection actions.
  */
-class TableHeader extends SelectionHeader {
```
dlockhart, line 9: Did you intend to make these changes with this PR? Don't seem related/needed.

eKoopmans: Ah! Yes mostly, though the `export` isn't necessary.

This doc needed updating for `TableHeader` specifically, so I've overridden the underlying `noSelection` prop to change its description.

dlockhart (approved):

oelhanafey (approved):

PR #3134, opened 2023-01-06
feat: Add status indicator slot to object-property-list

eKoopmans:
This PR adds a slot for a `d2l-status-indicator` that can be rendered at the start of the list. This feature came up while [backporting in consistent-evaluation](https://github.com/Brightspace/consistent-evaluation/pull/1333).

There's an edge case where the separators render incorrectly if the status-indicator is slotted after the items - I have a separate PR ready with a fix, but it requires some changes to how separators are hidden (JS instead of CSS) so I've kept this PR clean of that.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/674102762601&fdp=true

## Screenshots

image

### DIFF
--- a/components/object-property-list/README.md
+++ b/components/object-property-list/README.md
@@ -8,9 +8,11 @@ Object property lists are simple dot-separated lists of text, displayed sequenti
   import '@brightspace-ui/core/components/object-property-list/object-property-list.js';
   import '@brightspace-ui/core/components/object-property-list/object-property-list-item.js';
   import '@brightspace-ui/core/components/object-property-list/object-property-list-item-link.js';
+  import '@brightspace-ui/core/components/status-indicator/status-indicator.js';
 </script>
 
 <d2l-object-property-list>
+  <d2l-status-indicator slot="status" state="default" text="Status"></d2l-status-indicator>
   <d2l-object-property-list-item text="Example item"></d2l-object-property-list-item>
   <d2l-object-property-list-item text="Example item with icon" icon="tier1:grade"></d2l-object-property-list-item>
   <d2l-object-property-list-item-link text="Example link" href="https://www.d2l.com/"></d2l-object-property-list-item-link>
@@ -22,7 +24,7 @@ Object property lists are simple dot-separated lists of text, displayed sequenti
 
 An object property list can be defined using `d2l-object-property-list` and a combination of items (e.g., `d2l-object-property-list-item`, `d2l-object-property-list-item-link`).
 
-<!-- docs: demo code -->
+<!-- docs: demo live name:d2l-object-property-list -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/object-property-list/object-property-list.js';
@@ -36,17 +38,19 @@ An object property list can be defined using `d2l-object-property-list` and a co
 </d2l-object-property-list>
 ```
 
+<!-- docs: start hidden content -->
 ### Properties
 
 | Property | Type | Description |
 |--|--|--|
 | `skeleton-count` | Number | Number of skeleton items to insert if in skeleton mode |
+<!-- docs: end hidden content -->
 
 ### Word wrap
 
 The object property list is designed to wrap in an inline manner if the items are wider than the container.
 
-<!-- docs: demo code -->
+<!-- docs: demo -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/object-property-list/object-property-list.js';
@@ -63,8 +67,7 @@ The object property list is designed to wrap in an inline manner if the items ar
 
 The `d2l-object-property-list-item` component is the basic type of item for an object property list, displaying text and an optional leading icon.
 
-
-<!-- docs: demo code -->
+<!-- docs: demo live name:d2l-object-property-list-item -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/object-property-list/object-property-list.js';
@@ -90,7 +93,7 @@ The `d2l-object-property-list-item` component is the basic type of item for an o
 
 The `d2l-object-property-list-item-link` component is a link item for the object property list. It displays text as a hyperlink, with an optional leading icon.
 
-<!-- docs: demo code -->
+<!-- docs: demo live name:d2l-object-property-list-item-link -->
 ```html
 <script type="module">
   import '@brightspace-ui/core/components/object-property-list/object-property-list.js';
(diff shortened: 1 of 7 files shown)


components/object-property-list/README.md
```diff
@@ -63,8 +79,7 @@ The object property list is designed to wrap in an inline manner if the items ar
 
 The `d2l-object-property-list-item` component is the basic type of item for an object property list, displaying text and an optional leading icon.
 
-
-<!-- docs: demo code -->
+<!-- docs: demo live name:d2l-object-property-list-item -->
```
dlockhart, line 82: Cool that you can do this now!

components/object-property-list/object-property-list-item.js
```diff
@@ -28,22 +27,23 @@ export class ObjectPropertyListItem extends SkeletonMixin(RtlMixin(LitElement))
...
 			.separator d2l-icon {
 				color: var(--d2l-color-galena);
 			}
 			.item-icon {
-				margin: -0.1rem 0.3rem 0 0;
-			}
-			:host([dir="rtl"]) .item-icon {
-				margin: -0.1rem 0 0 0.3rem;
+				margin-top: -0.1rem;
+				margin-inline-end: 0.3rem;
```
dlockhart, line 46: 🎉

components/object-property-list/object-property-list.js
```diff
@@ -30,16 +30,23 @@ class ObjectPropertyList extends LocalizeCoreElement(SkeletonMixin(LitElement))
...
 	render() {
-
 		const slotContents = this.skeleton && this.skeletonCount > 0 ? [...Array(this.skeletonCount)].map(() => html`
 			<d2l-object-property-list-item text="${this.localize('components.object-property-list.item-placeholder-text')}" skeleton></d2l-object-property-list-item>
 		`) : nothing;
 
-		return html`<slot class="d2l-body-small">${slotContents}</slot>`;
+		return html`
+			<div class="d2l-body-small">
+				<slot name="status"></slot>
```
dlockhart, line 46: Wondering about how this new slot should behave when in skeleton mode... should it maybe be included in the `skeletonCount` behaviour? Or were you thinking they'd just put `<d2l-status-indicator skeleton>` inside the slot? (although status indicators don't appear to support skeletons currently)

eKoopmans: Yeah, `skeletonCount` requires there to be nothing slotted at all, and I don't think it makes sense to add an extra attribute on `object-property-list` to indicate that there should be an additional skeletonized status indicator. I was thinking `<d2l-status-indicator skeleton>` but I did notice that it doesn't support skeleton yet.

components/object-property-list/README.md
eKoopmans: Yeah - this one actually should (probably?) have worked even before my fixes, I just better understand how they work now than when I first wrote these docs 😄.

components/object-property-list/object-property-list.js
dbatiste: I am wondering if we should have this new slot or not. The slot very loosely enables us to control where the status indicator is rendered. Could we achieve the same thing by using some CSS like this? Of course, it doesn't explicitly require that the user place it first, but technically, we're not really limiting where they can render this anyway.

```css
::slotted(d2l-status-indicator) {
     margin-inline-end: 0.45rem;
}
```

The other thought that came to mind was to define a `d2l-object-property-list-item`, but that seems like overkill.

eKoopmans: The slot seemed appropriate to me because:

1. It limits it to only one status indicator, and forces it to render first (both are design requirements)
2. It makes the use case/support of status indicators clear and explicit, without having to create a new web component for it
3. It seems conceptually like a good match for the purpose of named slots

Yep we could achieve the margin without the slot, just wondering if there's any downside to using a slot though?

eKoopmans: Ah cool, yep I wasn't sure what happened when putting two items into a single named slot. I guess the good news is, if we ever need multiple status indicators at the start it's sort of supported already?

Putting a status indicator without a slot would be unsupported behaviour, which I called out in the readme "Best practices". But I think your screenshot is still a win - the slotted ones have the right margin and no separator, and the one used incorrectly looks as good as possible, with a separator and no extra right margin.

eKoopmans: One downside I just discovered - there's no screen-reader pause between reading a status and reading the first item. Not sure if that's required - I think the only way to get that behaviour would be to make a custom web-component for `d2l-object-property-list-item-status` 🤦.

dbatiste: Cool. If we do want to lock things down... we've done this sort of thing in one or two places:
```css
::slotted([slot="status"]) {
	display: none;
}
::slotted(d2l-status-indicator[slot="status"]) {
	display: inline-block;
	margin-inline-end: 0.45rem;
}
```
That might be desirable, since the slot isn't intended to be general purpose. Technically we could target `d2l-status-indicator` in the default slot and hide them, if we really do want to enforce not rendering them there.

eKoopmans:
Okay, hi 👋! I think this PR should be all done and ready to go. Jeff and I went over all the spacing/margins and got it the best we could - some inconsistency in the vertical position of the status indicator between Mac and Windows, but just 1 pixel and it's okay.

@dbatiste @dlockhart let me know if you have any more feedback, thanks!

components/object-property-list/object-property-list.js
```diff
@@ -30,16 +31,28 @@ class ObjectPropertyList extends LocalizeCoreElement(SkeletonMixin(LitElement))
 			::slotted(:last-child), slot :last-child {
 				--d2l-object-property-list-item-separator-display: none;
 			}
+			::slotted([slot="status"]) {
+				display: none;
+			}
+			::slotted(d2l-status-indicator[slot="status"]) {
```
dbatiste, line 37: I think in the readme it mentions that the slot is for a **single** status indicator. Is that true, and something we want to enforce? We might be able to lock this down further by using a selector like `::slotted(d2l-status-indicator[slot="status"]:first-of-type)`. Using `:first-child` is also an option, but would require users place the status indicator first, which is perhaps a bit too heavy-handed.

components/object-property-list/screen-reader-pause.js
```diff
@@ -0,0 +1,15 @@
+import { html, LitElement } from 'lit';
+import { getSeparator } from '@brightspace-ui/intl/lib/list.js';
+import { offscreenStyles } from '../offscreen/offscreen.js';
+
+export class ScreenReaderPause extends LitElement {
```
dbatiste, line 5: Did you intend `/object-property-list/` to be the home for this new component?

eKoopmans: Yes I did, mostly because it's only used by the OPL (currently) and I didn't have a better place to put it. Do you have a suggestion? Could go in `helpers`, though those are mostly not web-components...

components/object-property-list/object-property-list.js
eKoopmans: Sure thing, done.

components/object-property-list/screen-reader-pause.js
eKoopmans: Maybe in the `components/offscreen` folder?

dbatiste: That might be an ok place. Not sure where else it would go unless it goes into a dedicated folder. I think it will seem weird for anyone else to reference it in its current location. Maybe others have an opinion. 🤷

eKoopmans: Okay, moved it to `offscreen` and added some documentation 👍. And I've added it to the list of imports on the `documentation` issue.

dbatiste (approved):
Looks good to me!

PR #3136, opened 2023-01-06
fix: OPL separators on hidden items and edge cases

eKoopmans:
This PR is a follow-up to #3134. It changes how separators are rendered in object-property-lists to fix an edge case with status indicators, and also adds `hidden` support! The hidden support should make backporting to Polymer projects a lot easier/more possible in the future.

~Opening as a draft targeting #3134's branch until it is merged.~

Rally: https://rally1.rallydev.com/#/?detail=/userstory/674102762601&fdp=true


dlockhart (approved):
Cool, looks good. It's unfortunate that we need to jump through these kinds of hoops just to handle hidden items, but I can't think of a cleaner way, so 🤷 !

components/object-property-list/object-property-list.js
```diff
@@ -50,10 +53,23 @@ class ObjectPropertyList extends LocalizeCoreElement(SkeletonMixin(LitElement))
...
 			</div>
 		`;
 	}
+
+	_onItemsChanged() {
+		this._setItemSeparatorVisibility();
+	}
+
+	_setItemSeparatorVisibility() {
+		const slottedElements = this._slot.assignedElements();
```
dbatiste, line 66: It looks like we're assuming that the `slot` element will never change. To avoid the assumption, we could query for the slot when needed, but in this particular case, we can have it passed in...
```javascript
_onItemsChanged(e) {
	this._setItemSeparatorVisibility(e.target); //e.target of slotchange is the slot element
}
_setItemSeparatorVisibility(slot) {
	...
}
```

components/object-property-list/object-property-list.js
```diff
@@ -41,6 +38,12 @@ class ObjectPropertyList extends LocalizeCoreElement(SkeletonMixin(LitElement))
 		`];
 	}
 
+	firstUpdated() {
+		this._slot = this.shadowRoot.querySelector('slot:not([name])');
+		this.addEventListener('d2l-object-property-list-item-visibility-change', this._onItemsChanged);
+		if (this._slot.childElementCount) this._onItemsChanged();
```
dbatiste, line 44: Interesting... I never knew the slot had this property. Seems [Mozilla isn't forthcoming in these dets](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot).

dbatiste: Though, are you sure this call is needed in `firstUpdated`? I think the `slotchange` event will be dispatched when the nodes are first distributed to it.

dbatiste:
> Cool, looks good. It's unfortunate that we need to jump through these kinds of hoops just to handle hidden items, but I can't think of a cleaner way, so 🤷 !

I am hopeful that eventually we will be able to do this sort of thing, for example to select an element that has a hidden sibling (or doesn't have a hidden sibling):
```css
:host(:has(+ [hidden])) { ... }
:host(:not(:has(+ [hidden]))) { ... }
```

It works in Safari, but it does not work in Chrome. There appears to be some ongoing discussion regarding what types of selectors should be valid in `:host()`, and whether relative selectors used in `:host()` should operate on light-DOM siblings. https://github.com/w3c/csswg-drafts/issues/7953.

components/object-property-list/object-property-list.js
eKoopmans: This is specifically for the skeleton scenario, when nothing is slotted. `childElementCount` is to detect if there is any default content (e.g. skeleton) being rendered. So it's not anything specific to slots, rather just a default element attribute:
https://developer.mozilla.org/en-US/docs/Web/API/Element/childElementCount

In my testing this was necessary, otherwise the visual-diff for skeleton was rendering a separator on the final item.

eKoopmans:
> select an element that has a hidden sibling

Yeah, it does still get messy here with the facts that:
- an item should render a separator if there is any non-hidden item after it
- but an "item" can be `<d2l-object-property-list-item>` or `<d2l-object-property-list-item-link>`
- and we want to specifically not account for other elements, especially if they're slotted

dbatiste (approved):

PR #3139, opened 2023-01-06
fix: removing overflow on html-block to fix aXe failure

dlockhart:
There's an aXe rule [called `scrollable-region-focusable`](https://dequeuniversity.com/rules/axe/4.6/scrollable-region-focusable?application=axeAPI) that requires that any element that sets `overflow` to `auto` or `hidden` is either focusable itself, or has a focusable element inside. This enables keyboard-only users to scroll the element.

Interestingly, this is only failing aXe in Firefox ,which is why our tests haven't caught it. UCE did instead.

After some discussion and sleuthing through code history, we're reasonably confident that these overflow values are no longer needed. They were [originally introduced](https://github.com/Brightspace/lms/commit/4e74a0321be7d6f3894b85c1c320aecf54b309b6) in 2012 on paragraph tags inside HTML blocks to "only show horizontal scroll bar when content overflows" and then [moved to the whole block](https://github.com/Brightspace/lms/commit/4a4db42339c5096906dbdf03c09d00aafba6a229) in 2013 to "correct issue with non-breaking words getting individual scrollbars".

Back then `overflow-wrap: break-word` didn't exist in CSS so our assumption is that they were trying to prevent really long non-breaking words from introducing scrollbars. Ultimately that `break-word` overflow was [introduced to HTML blocks](https://github.com/Brightspace/lms/commit/193a0550c45d95758782ddd38648a0755016f034) in 2013 (for Firefox and IE at the time!), which we believe made the overflow code redundant.


grant-cleary:
Visual diff failures are curious... Less the minute changes to math layout and more to the size of the block itself. Though maybe that's a visual diff thing. 🤷

dlockhart:
Interesting. Agreed -- I'm not worried about the anit-aliasing changes, but the size changes are weird. Investigating.

dlockhart:
Reason for the visual diffs:

MathJax applies a `margin: 1em 0` on the math, presumably so that there's always a bit of space around it. When the host has `overflow` set, that's one of many things that can prevent [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) with sibling elements. So when `overflow` was removed with this change, suddenly the math's margin gets collapsed with the previous visual diff container.

Visually it looks fine, and there'll always be some space between the math and the thing next to it. I'm OK with this change.

grant-cleary (approved):

dbatiste (approved):

PR #3143, opened 2023-01-09
Fix list-item-mixin setting grid styles to all actions children

awikkerink:
I want to move this component: https://github.com/Brightspace/d2l-outcomes/blob/main/src/components/outcomes-management/outcomes-management-list-item.js to use list-item-mixin. But the actions piece of `this._renderListItem` breaks the actions: https://github.com/Brightspace/d2l-outcomes/blob/main/src/components/outcomes-management/outcomes-management-list-item.js#L88-L110 because the actions selector applies the grid styles to all the children.

Relevant Rally story: [US145754](https://rally1.rallydev.com/#/?detail=/userstory/671769625799&fdp=true)

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -187,7 +187,7 @@ export const ListItemMixin = superclass => class extends LocalizeCoreElement(Lis
 			}
 
 			[slot="content"] ::slotted([slot="illustration"]),
-			[slot="content"] .d2l-list-item-illustration * {
+			[slot="content"] .d2l-list-item-illustration > * {
 				border-radius: 6px;
 				flex-grow: 0;
 				flex-shrink: 0;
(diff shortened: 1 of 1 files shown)


dbatiste (commented):
All looks reasonable.

I'm curious about the icon that gets rendered in the actions slot. What the deal there? I don't expect any issues, but we generally expect actions in the actions slot.

awikkerink:
> All looks reasonable.
> 
> I'm curious about the icon that gets rendered in the actions slot. What the deal there? I don't expect any issues, but we generally expect actions in the actions slot.

The icon indicates the share/locked state of the list item, and the component we're copying over had the icon next to the dropdown. I'm aware we're going to need do some changes to make that icon and tooltop accessible, since the original component had some accessibility issues we'll need to work through.

But the `*` selector affects the dropdown, so this change will be required anyway.

dbatiste (approved):

dbatiste:
It's worth mentioning that list-items should always be direct children of the `d2l-list`. In the original code, this does not appear to be the case. Assuming you switch to the mixin, the component will then become an actual list-item (instead of a component that wraps the a list-item). I presume that is one of the reasons you are wanting to switch to the mixin.

awikkerink:
> Assuming you switch to the mixin, the component will then become an actual list-item (instead of a component that wraps the a list-item). I presume that is one of the reasons you are wanting to switch to the mixin.

Correct

awikkerink:
@dbatiste how do you do a patch release? Looking at the README (I should've looked at it first), I should've prefixed `fix:` to the commit. But since I didn't do that, how should I proceed?

dbatiste:
You can do an empty commit with the `fix:` prefix, or you can looks for some typo to fix, or you can wait for someone else to do a release. Generally there's at least one per day. I have some fixes I am batching and will probably do a PR for them in the morning.

PR #3149, opened 2023-01-11
WCAG 2.2 Focus indicator fixes for core components

dbatiste:
This change includes focus indicator fixes for the following:

See [US145262](https://rally1.rallydev.com/#/?detail=/userstory/668586246427&fdp=true) for more details on focus appearance.

`d2l-button-icon` [translucent] -> [DE51166](https://rally1.rallydev.com/#/?detail=/defect/672626313517&fdp=true)
`d2l-link` -> [DE51167](https://rally1.rallydev.com/#/?detail=/defect/672628648235&fdp=true)
`d2l-menu-item` -> [DE51168](https://rally1.rallydev.com/#/?detail=/defect/672633057207&fdp=true)
`d2l-list-item` -> [DE51189](https://rally1.rallydev.com/#/?detail=/defect/672765848163&fdp=true)


components/link/link.js
```diff
@@ -15,10 +16,15 @@ export const linkStyles = css`
...
 	}
-	.d2l-link:hover, .d2l-link:focus {
+	.d2l-link:hover {
 		color: var(--d2l-color-celestine-minus-1);
 		text-decoration: underline;
 	}
+	.d2l-link:focus {
+		outline: 2px solid var(--d2l-color-celestine);
+		outline-offset: 3px;
+		text-decoration: underline;
```
dlockhart, line 26: Fun, this is a big change! We're good with the text colour staying as celestine on-focus but being celestine-minus-1 on hover?

components/menu/menu-item-styles.js
```diff
@@ -4,34 +4,28 @@ import { css } from 'lit';
...
 	:host(:hover),
 	:host([first]:hover) {
 		background-color: var(--d2l-menu-background-color-hover);
-		border-bottom: 1px solid var(--d2l-menu-border-color-hover);
-		border-top: 1px solid var(--d2l-menu-border-color-hover);
 		color: var(--d2l-menu-foreground-color-hover);
-		z-index: 2;
 	}
 
-	/** separated because Safari <15.4 is having trouble parsing these */
```
dlockhart, line 28: Isn't this comment still applicable for now?

components/link/link.js
dbatiste: I think so, but I will double check with Jeff.

components/menu/menu-item-styles.js
dbatiste: The only overlap between the two is color now... but I will add this back just in case someone thinking to combine them in the future.

components/link/link.js
dbatiste: Ok, confirmed with Jeff that this is expected.

components/list/list-item-content.js
```diff
@@ -12,13 +12,13 @@ class ListItemContent extends LitElement {
 
 	static get styles() {
 		return [ bodySmallStyles, bodyCompactStyles, css`
-			:host {
```
dbatiste, line 15: I think these overflow values were meant to handle crazy long string cases as described in [Maya's PR](https://github.com/BrightspaceUI/core/pull/752/files). I think there might also have been some original code to render ellipsis, but if there was, that is no longer the case.

As described in [these Mozilla notes](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), the `word-break: normal + overflow-wrap: anywhere` achieves essentially the same as `word-break: break-word` (where `break-word` usage on `word-break` also works but is deprecated). Also, `overflow-wrap` is the standardized property that replaced the non-standard `word-wrap` - they map to the exact same thing.

components/menu/menu-item-styles.js
```diff
@@ -4,34 +4,29 @@ import { css } from 'lit';
 export const menuItemStyles = css`
 	:host {
 		background-color: var(--d2l-menu-background-color);
-		border: 1px solid transparent;
```
dbatiste, line 7: Note: this change reduces the height of the menu-items by `1px`. I confirmed with Jeff that this is ok. It does result in a few extra visual-diffs.

components/link/link.js
```diff
@@ -72,6 +72,11 @@ class Link extends FocusMixin(LitElement) {
 			 * @type {boolean}
 			 */
 			main: { type: Boolean, reflect: true },
+			/**
+			 * Whether to truncate the link with ellipsis
+			 * @type {boolean}
+			 */
+			overflowEllipsis: { type: Boolean, attribute: 'overflow-ellipsis' },
```
dlockhart, line 79:
```suggestion
			overflowEllipsis: { type: Boolean, attribute: 'overflow-ellipsis', reflect: true },
```

dlockhart (approved):
This is looking good to me!

PR #3154, opened 2023-01-11
Update token for update-package-lock workflow

svanherk:
We are changing `D2L_GITHUB_TOKEN` to use more limited permissions.  We need to update all places that don't _need_ to be using it, like here.

I can't merge this until I see the new token appear, after the `repo-settings` run is approved.


dlockhart (approved):
🎉

PR #3157, opened 2023-01-12
fix: Collapsible Panel docs update

jameswklassen:
## Description
Some updates to the collapsible panel daylight page.

## Checklist
- [x] Approved by design

### DIFF
--- a/components/collapsible-panel/README.md
+++ b/components/collapsible-panel/README.md
@@ -97,17 +97,16 @@ The `d2l-collapsible-panel` element is a container that provides specific layout
 
 <style>
 	d2l-collapsible-panel {
-		width: 500px;
-		margin-block: 2rem;
+		width: 800px;
 	}
 	/* TODO: remove this when daylight demo resizing is fixed */
 	d2l-collapsible-panel:not([expanded]) {
-		margin-bottom: 8rem;
+		margin-bottom: 14rem;
 	}
 </style>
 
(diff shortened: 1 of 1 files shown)


components/collapsible-panel/README.md
```diff
@@ -204,9 +203,7 @@ Use an inline collapsible panel to progressively disclose sections of a complex
 ## Summary Items [d2l-collapsible-panel-summary-item]
 An optional summary can help the user understand what’s inside the collapsible panel without having to expand it. This can be helpful if the user needs more than the heading to explain what’s inside.
 
-More than one d2l-collapsible-panel-summary-item can be added to the summary slot, and each will appear on its own line.
-
-<!-- docs: demo live -->
+<!-- docs: demo -->
```
oelhanafey, line 206: Currently, the properties for the `d2l-collapsible-panel-summary-item` component won't be shown in the documentation. Would it make sense to have a simple demo initially for it like the demo in `Optional Slots`? That would allow us to display the code snippet and properties of the component and then we could put the more complex demo somewhere else to show off the fully functional component.

jameswklassen: We do have this code snippet for summary items, which shows the maximum functionality of the summary item, which seems sufficient to me. But I'm definitely open to other thoughts here.

image

oelhanafey (approved):

dbatiste (approved):

PR #3168, opened 2023-01-16
fix:announce helper

GZolla:
Prevent messages from being combined in VO
[DE50838](https://rally1.rallydev.com/#/?detail=/defect/669268552007&fdp=true)

### DIFF
--- a/helpers/announce.js
+++ b/helpers/announce.js
@@ -2,6 +2,7 @@ let timeoutId = null;
 let container = null;
 let messageQueueHead = null;
 let messageQueueTail = null;
+let readMessages = [];
 
 export function announce(message) {
 	if (!message) return;
@@ -38,7 +39,12 @@ export function announce(message) {
 function announceFromQueue() {
 	// Clear old messages and announce
 	container.innerHTML = '';
-	container.appendChild(document.createTextNode(messageQueueHead.message));
+
+	let message = messageQueueHead.message;
+	if (readMessages.indexOf(message) > -1) message = message.concat('\u00A0');
+
+	readMessages.push(message);
+	container.appendChild(document.createTextNode(message));
 	messageQueueHead = messageQueueHead.next;
 
 	if (!messageQueueHead) {
@@ -47,6 +53,7 @@ function announceFromQueue() {
 			container.parentNode.removeChild(container);
 			container = null;
 			timeoutId = null;
+			readMessages = [];
 		}, 10000);
 	} else {
 		timeoutId = setTimeout(announceFromQueue, 200);
--- a/helpers/announce.js
+++ b/helpers/announce.js
@@ -1,5 +1,7 @@
 let timeoutId = null;
 let container = null;
+let messageQueueHead = null;
+let messageQueueTail = null;
 
 export function announce(message) {
 	if (!message) return;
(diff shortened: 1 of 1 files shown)


helpers/announce.js
```diff
@@ -22,27 +20,35 @@ export function announce(message) {
...
 
-	/* Need to give browser enough time to create the live region so that it will
-	treat the change as an update. Firefox sometimes ignores changes if the region
-	and update are made too quickly in succession. RequestAnimationFrame is not
-	sufficient here. Also, for some strange reason, sometimes VO will not announce
-	duplicate messages even if we remove the child so we also append a non-breaking space. */
-	const elem = [...container.childNodes].find((c) => c.textContent === message);
-	if (elem) {
-		elem.parentNode.removeChild(elem);
-		message = message.concat('\u00A0');
```
oelhanafey, line 33: I tested out the PR site and it looks like the defect that made this necessary still exists (On Chrome + VoiceOver). If you have identical messages, for example by putting "Hello" in both inputs in the demo, it will only announce the first one.

We added this non-breaking space so that the screen-reader will perceive identical messages differently and announce them.

oelhanafey: @GZolla I don’t think the change solves the problem with the repeated messages.

What the code was doing previously was checking if the last message was equal to the current message and if it was it adds the space. So if you have repeated messages, it would look something like this:
```
Hello // Not equal to last message so prints as usual
Hello\u00A0 // Equal to last message so adds a space
Hello // Not equal to last message so does not add a space
Hello\u00A0 ...
```

In your code, it adds a space if the message is in the list so this will read the first 2 messages but not subsequent messages:
```
Hello
Hello\u00A0
Hello\u00A0 // Identical to previous message, so VO will not read this
Hello\u00A0 // Identical to previous message, so VO will not read this
```

So I think we should be checking the equality of the current and previous messages for this to work.

eKoopmans:
Hi, I'm having a look at this now. I can say that the immediate defect from the story is resolved on my end (Windows + Chrome + NVDA no longer repeats the first message when announcing a second message within 10 seconds).

But I'm not certain that this hasn't introduced any other regressions. Still looking over the code, and also thinking we should set up some standard manual tests for this scenario, to try to consistently measure all the edge cases - something like `announce.test-manual.js` that we can run locally and test the screen-reader behaviour.

Should probably also check in with @dbatiste, his name is on some of the big "watch out!" comments that are being deleted here.

oelhanafey:
After the changes, it looks like it is no longer coalescing messages. Previously in the demo on VO, I would get:
"I like cookies but I also like donuts and many other really yummy things. I also like cake."
Both announce calls were put in a single message.

In this branch, I am getting them in separate messages:
"I like cookies but I also like donuts and many other really yummy things."
"I also like cake."
The result of it not coalescing anymore, is that the second announcement is sometimes interrupting the first one now.

eKoopmans:
Hey @GZolla , I'm just coming back to this - I think there's lots of tricky interactions going on here. I'd probably recommend:

- switching this PR to a draft
- adding a note onto the story on Rally, mentioning this PR and your findings about possible fixes/issues
- returning this to "None" and unassigning yourself, putting it back in the fixer queue 😄 

That said, if you want to keep tackling it that's an option too! Not sure what else you have on your plate, I'm just conscious that we only picked this up as part of the mob programming from dev onboarding, and the goal there wasn't necessary to come to a solution.

PR #3170, opened 2023-01-17
Change html-block text-align to match dir in the html

oelhanafey:
[DE43128](https://rally1.rallydev.com/#/?detail=/defect/519545114348&fdp=true): New Brightspace Editor > 20.21.4 > previews Text direction RTL incorrectly as LTR

The `<d2l-html-block>` component was not correctly aligning text with respect to the `dir` attribute in the block. The existing RTL styles would only pay attention to the `dir` of the `<d2l-html-block>` component itself.

This PR fixes the issue and adds some rtl paragraphs to the vdiff tests.


dlockhart (commented):
This seems perfect and elegant! @grant-cleary and @dbatiste will know more, but [there's a copy of this CSS](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/HtmlBlock/HtmlBlock.css?r=f12771fe#8) in the monolith that'll also likely need updating. I believe it's used in a scenario where the HTML doesn't end up in the web component's shadow root.

dbatiste (approved):

dlockhart (approved):

PR #3176, opened 2023-01-18
fix: disabled list item selectable

GZolla:
Bottom border no longer disappears and checkboxes are not highlighted.

[DE50819](https://rally1.rallydev.com/#/?detail=/defect/668780676319&fdp=true): List-Item > Bottom border disappears when disabled and hovered


oelhanafey (approved):
Looks good! 🎉

oelhanafey:
Ah, you might need to add your fix to these other types of list-items that also seem to have the same issue:
https://github.com/BrightspaceUI/core/blob/6350eb7396077b1495395e05e38f9cfd2cd06547/components/list/list-item-link-mixin.js#L25
https://github.com/BrightspaceUI/core/blob/6350eb7396077b1495395e05e38f9cfd2cd06547/components/list/list-item-button-mixin.js#L18

GZolla:
List-item-button and list-item-link show have a similar bug. However, I think their behavior when disabling needs to be discussed further before adding the fix to them.

PR #3177, opened 2023-01-18
fix: primary secondary resizing defects

oelhanafey:
[DE51555](https://rally1.rallydev.com/#/?detail=/defect/675441395861&fdp=true): Primary-Secondary > Mobile view secondary tray does not open when clicked if slide down hidden

This solves two defects with the primary secondary template in the Mobile view when resizing with the mouse.

### DIFF
--- a/templates/primary-secondary/primary-secondary.js
+++ b/templates/primary-secondary/primary-secondary.js
@@ -334,6 +334,7 @@ class MobileMouseResizer extends Resizer {
 
 	constructor() {
 		super();
+		this._isDragging = false;
 		this._onMouseDown = this._onMouseDown.bind(this);
 		this._onMouseMove = this._onMouseMove.bind(this);
 		this._onMouseUp = this._onMouseUp.bind(this);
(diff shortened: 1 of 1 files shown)


templates/primary-secondary/primary-secondary.js
```diff
@@ -386,12 +389,13 @@ class MobileMouseResizer extends Resizer {
...
-			|| (!this._wasCollapsed && desiredSecondaryHeight < expandedCollapseThreshold)
+			this._isDragging
+			&& ((this._wasCollapsed && desiredSecondaryHeight < collapsedCollapseThreshold)
+			|| (!this._wasCollapsed && desiredSecondaryHeight < expandedCollapseThreshold))
 		) {
-			if (desiredSecondaryHeight > 0) this.dispatchResize(0, true);
+			if (desiredSecondaryHeight !== 0) this.dispatchResize(0, true);
 		}
-		else if (desiredSecondaryHeight < this.contentBounds.minHeight) {
+		else if (this._isDragging && desiredSecondaryHeight < this.contentBounds.minHeight) {
```
oelhanafey, line 398: There appears to be some rounding down in the actual height and `this.contentBounds.minHeight`. This resulted in `desiredSecondaryHeight < this.contentBounds.minHeight` being true when the secondary panel was at the min height. 

Dispatching a resize turns off the click action to open or close the panel and since this check was true, it was disabling the close action when clicking at minimum width. I added a check to not resize if the mouse was not moved, so the close panel click action works in that case.

dlockhart (approved):

PR #3179, opened 2023-01-18
Allow filter values to wrap to 2 lines

GZolla:
Allow filter values to wrap to 2 lines and overflow with an ellipsis at the end.
[US146523](https://rally1.rallydev.com/#/?detail=/userstory/673272772181&fdp=true): Filter > Allow overflowing values to wrap to max 2 lines


components/filter/filter.js
```diff
@@ -118,7 +119,6 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			.d2l-filter-dimension-set-value-text {
 				overflow: hidden;
 				text-overflow: ellipsis;
```
dbatiste, line 121: If we're using the multi-line clamping instead of the single-line ellipsis, I _think_ we might be able to remove this `text-overflow: ellipsis`. I _think_ we still need the `overflow: hidden` however.

components/filter/filter.js
```diff
@@ -128,6 +128,14 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				line-height: unset;
 				overflow: hidden;
 			}
+			
+			.d2l-filter-dimension-set-value-text {
```
dbatiste, line 132: Nit: note that the style linter will give errors for these properties as written. The properties need to be listed alphabetically, and there should be a space after the colon in each case (ex. `hyphens: auto;`).

components/filter/filter.js
svanherk: If that's the case, I think we'd need to move `text-flow: ellipsis` into the `.d2l-filter-dimension-header-text` section so it's still set for the header.  It's still just a single-line ellipsis.

svanherk:
Looking at the visual-diff test results, I think it'd be good to change one of the values to be _even_ longer so that we have a value that wraps to two lines with no ellipsis, and a value that is two lines with the ellipsis.

components/filter/filter.js
dbatiste: Ok cool. If we still need it, no issue keeping it I think.

oelhanafey (commented):
Looks good! One design consideration that I wanted to confirm. When the `dimension-set-value` wraps to two lines, the count vertically aligns with the whole component. Is this what we want to happen?

Screen Shot 2023-01-20 at 11 44 59 AM

Nit: The text on some of the visual diffs that say "Very Long Dimension Title For Testing Text Truncation" don't actually truncate anymore. Could we change that text to say it is wrapping instead for clarity?

GZolla:
I spoke to @geurts and the PR is ready to go and working as intended.

oelhanafey (approved):
LGTM!

PR #3183, opened 2023-01-19
fix: disable button clicking when item is disabled

stevenao:
https://rally1.rallydev.com/#/29180338367ud/custom/486232622040?detail=%2Fdefect%2F673968735311

In addition to removing visual hints like hovering and focus when item is disabled, I feel that it also makes sense to disable the button clicking action.


components/list/list-item-button-mixin.js
```diff
@@ -4,13 +4,32 @@ import { getUniqueId } from '../../helpers/uniqueId.js';
...
 
 		const styles = [ css`
 			:host {
 				--d2l-list-item-content-text-color: var(--d2l-color-celestine);
 			}
+			:host([button-disabled][_hovering-primary-action]) .d2l-list-item-content,
+			:host([_focusing-primary-action]) .d2l-list-item-content {
+				--d2l-list-item-content-text-decoration: none;
+			}
+			:host([button-disabled]) .d2l-list-item-content {
```
stevenao, line 27: @oelhanafey Actually, I am not sure about the greying out. Can you confirm?

components/list/list-item-button-mixin.js
```diff
@@ -59,7 +68,7 @@ export const ListItemButtonMixin = superclass => class extends ListItemMixin(sup
 	}
 
 	_renderPrimaryAction(labelledBy) {
-		return html`<button id="${this._primaryActionId}" aria-labelledby="${labelledBy}" @click="${this._onButtonClick}"></button>`;
+		return html`<button id="${this._primaryActionId}" aria-labelledby="${labelledBy}" @click="${this._onButtonClick}" .disabled="${this.buttonDisabled}"></button>`;
```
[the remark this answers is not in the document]
stevenao: will do it in a separate commit..

components/list/list-item-button-mixin.js
oelhanafey: I don't think we want to grey it out. A list-item-button with the button disabled should just appear like a regular list-item.

I think we also want to unset the color on the button since the current celestine color makes it look like it is clickable.

components/list/list-item-button-mixin.js
```diff
@@ -4,13 +4,30 @@ import { getUniqueId } from '../../helpers/uniqueId.js';
...
 
 	static get styles() {
 
 		const styles = [ css`
-			:host {
+			:host(:not([button-disabled])) {
 				--d2l-list-item-content-text-color: var(--d2l-color-celestine);
 			}
+			:host([button-disabled][_hovering-primary-action]) .d2l-list-item-content,
+			:host([_focusing-primary-action]) .d2l-list-item-content {
```
oelhanafey, line 24:
```suggestion
			:host([button-disabled][_focusing-primary-action]) .d2l-list-item-content {
```
I think we only want to disable the focus styles if the button is disabled

stevenao: Focusing has to do with keyboard navigation. Hovering is the mouse navigation having a focus on the element.  Do we want the color change and underlining style when mouse is over the primary action? I think we need to disable for both.

oelhanafey: Yep, I think we want to keep the focus style for active buttons and disable it only for disabled buttons. That's the current behaviour in both list item links and list item buttons
Current:
Screen Shot 2023-01-26 at 4 29 36 PM
This branch:
Screen Shot 2023-01-26 at 4 29 49 PM

components/list/list-item-button-mixin.js
```diff
@@ -4,13 +4,30 @@ import { getUniqueId } from '../../helpers/uniqueId.js';
...
 	static get styles() {
 
 		const styles = [ css`
-			:host {
+			:host(:not([button-disabled])) {
 				--d2l-list-item-content-text-color: var(--d2l-color-celestine);
 			}
+			:host([button-disabled][_hovering-primary-action]) .d2l-list-item-content,
+			:host([button-disabled][_focusing-primary-action]) .d2l-list-item-content {
+				--d2l-list-item-content-text-color: var(--d2l-color-ferrite);
```
oelhanafey, line 25:
```suggestion
				--d2l-list-item-content-text-color: unset;
```

Current list-items don't have a default color, they inherit it from their parent

oelhanafey (approved):
Thanks for the contribution!

@GZolla was planning on having a closer look at the link and button list item styles next week.

PR #3189, opened 2023-01-20
fix: Collapsible Panel - rename `padding` to `padding-type`

jameswklassen:
## Description
It can be unclear that `padding` is a "type selector" rather than an extension of the css property.

By renaming to `padding-type`, this becomes more clear.


dlockhart (approved):
Nice, thanks!

PR #3192, opened 2023-01-23
Move input-text tooltip location

oelhanafey:
[DE51662](https://rally1.rallydev.com/#/?detail=/defect/676000972225&fdp=true): Input-text > Fix tooltip hover area

This PR moves the tooltip in the `d2l-input-text` component from the `input` element up to the host so that hovering anywhere in the component will open the tooltip. This makes the behaviour consistent with the validation tooltips that are used in the `d2l-input-number` component. There was an edge case where if you opened the tooltip and then turned on skeleton, the tooltip would still appear on hover so I added a check to the tooltip to confirm that the input is not in skeleton.

A separate inconsistency was noticed. Hovering inside of the component but outside of the label and bordered input area, such as to the right of the label, will open the tooltip (if warranted), but it will not enable the hover styles. After discussing with Jeff, we think it would be a good idea to enable hover styles when anywhere inside the input-text component is hovered. I have spun up a separate story to tackle this work: [US148080](https://rally1.rallydev.com/#/?detail=/userstory/684245679355&fdp=true).

### DIFF
--- a/components/inputs/input-label-styles.js
+++ b/components/inputs/input-label-styles.js
@@ -8,8 +8,8 @@ export const inputLabelStyles = css`
 		font-weight: 700;
 		letter-spacing: 0.2px;
 		line-height: 1rem;
-		margin: 0;
-		padding: 0 0 7px 0;
+		margin: 0 0 7px 0;
+		padding: 0;
 	}
 	:host([required]) .d2l-input-label::after,
 	.d2l-input-label-required::after {
@@ -25,10 +25,6 @@ export const inputLabelStyles = css`
 	:host([dir="rtl"]) .d2l-input-label-required::after {
 		left: auto;
 		right: 0.15rem;
-	}
-	:host([skeleton]) .d2l-input-label.d2l-skeletize {
-		margin: 0 0 7px 0;
-		padding: 0;
 	}
 	:host([skeleton]) .d2l-input-label.d2l-skeletize::before {
 		bottom: 0.25rem;
--- a/components/inputs/input-label-styles.js
+++ b/components/inputs/input-label-styles.js
@@ -26,6 +26,10 @@ export const inputLabelStyles = css`
 		left: auto;
 		right: 0.15rem;
 	}
+	:host([skeleton]) .d2l-input-label.d2l-skeletize {
+		margin: 0 0 7px 0;
+		padding: 0;
+	}
 	:host([skeleton]) .d2l-input-label.d2l-skeletize::before {
 		bottom: 0.25rem;
 		top: 0.15rem;
--- a/components/inputs/input-label-styles.js
+++ b/components/inputs/input-label-styles.js
@@ -8,8 +8,8 @@ export const inputLabelStyles = css`
 		font-weight: 700;
 		letter-spacing: 0.2px;
 		line-height: 1rem;
-		margin: 0 0 7px 0;
-		padding: 0;
+		margin: 0;
+		padding: 0 0 7px 0;
 	}
 	:host([required]) .d2l-input-label::after,
 	.d2l-input-label-required::after {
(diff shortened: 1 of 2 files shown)


components/inputs/input-text.js
```diff
@@ -460,14 +460,14 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
...
 					${ (!isValid && !this.hideInvalidIcon && !this._focused) ? html`<div class="d2l-input-text-invalid-icon" style="${styleMap(invalidIconStyles)}" @click="${this._handleInvalidIconClick}"></div>` : null}
-					${ this.validationError ? html`<d2l-tooltip for=${this._inputId} state="error" align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>` : null }
 				</div><div id="after-slot" class="d2l-skeletize" ?hidden="${!this._hasAfterContent}"><slot name="after" @slotchange="${this._handleAfterSlotChange}"></slot></div>
 			</div>
 			${offscreenContainer}
 		`;
 		if (this.label && !this.labelHidden && !this.labelledBy) {
 			const unitLabel = this._getUnitLabel();
 			return html`
+				${ this.validationError && !this.skeleton ? html`<d2l-tooltip state="error" align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>` : null }
```
dlockhart, line 470: Don't we still need to show the validation tooltip if the label is hidden?

components/inputs/input-label-styles.js
```diff
@@ -8,8 +8,8 @@ export const inputLabelStyles = css`
 		font-weight: 700;
 		letter-spacing: 0.2px;
 		line-height: 1rem;
-		margin: 0 0 7px 0;
-		padding: 0;
+		margin: 0;
+		padding: 0 0 7px 0;
```
dlockhart, line 12: Why was this change necessary? A potentially unintended consequence of using padding instead of margins is that it'll no longer "collapse" with the element below the input.

oelhanafey: This was needed to fix an issue where hovering in the middle between the label and the input would not enable the hover styles. 

Though it may not be needed if we add the changes proposed in [US148080](https://rally1.rallydev.com/#/?detail=/userstory/684245679355&fdp=true) to enable hover styles when hovering anywhere in the input-text component. So we could also leave it as-is for now.

components/inputs/input-text.js
```diff
@@ -460,18 +460,21 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
...
-				<label class="d2l-input-label d2l-skeletize" for="${this._inputId}">${this.label}${unitLabel ? html`<span class="d2l-offscreen">${unitLabel}</span>` : ''}</label>
-				${input}`;
+			label = html`<label class="d2l-input-label d2l-skeletize" for="${this._inputId}">${this.label}${unitLabel ? html`<span class="d2l-offscreen">${unitLabel}</span>` : ''}</label>`;
 		}
-		return input;
+
+		return html`
+			${ this.validationError && !this.skeleton ? html`<d2l-tooltip state="error" align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>` : nothing }
+			${label}
+			${input}`;
```
dlockhart, line 477: nit: I like how much clearer everything is with how you've re-arranged it! I'd suggest using the same technique for the validation tooltip just to make it consistent?

```suggestion

		let tooltip = nothing;
		if (this.validationError && !this.skeleton) {
			tooltip = html`<d2l-tooltip state="error" align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>`;
		}

		return html`
			${tooltip}
			${label}
			${input}`;
```

components/inputs/input-label-styles.js
dlockhart: Cool -- yeah as long as it's not too janky as-is, that sounds like a good plan!

dlockhart (approved):
Nice!

PR #3199, opened 2023-01-25
A couple fixes to focus-indicator for menu-item styles for wcag 2.2.

dbatiste:
This is the first of a couple fixes for focus-indicators. This corrects for the reset stylesheet override as well as applies a border radius to the menu-item that is called for in the design (not sure how it got missed).

Missing focus indicator:
image

Fixed:
image


dlockhart (approved):
Looks good, as do the visual diffs.

PR #3203, opened 2023-01-25
Change backdrop to use aria-hidden instead of role=presentation

oelhanafey:
[DE51746](https://rally1.rallydev.com/#/?detail=/defect/681718782515&fdp=true): nicc>nicc>20.22.12.18407>Quizzes>Error when editing quiz

This PR changes the backdrop component to hide the backdrop by using `aria-hidden="true"` instead of `role="presentation"`. Setting role is causing quizzing to crash when dialogs are opened on Chrome 109. After updating Chrome, I can reliably reproduce this defect and can confirm that this change fixes the issue.

`role="presentation"` is not meant for hiding or showing content as it is being used for here. It is meant to ignore certain content that has no semantic meaning on the page like decorative icons. `aria-hidden` makes more semantic sense for these purposes and should also be more efficient as well since we don't need to set the roles on all of the child elements anymore.

Also fixes a small defect where the `slow-transition` property would override reduced-motion.

### DIFF
--- a/components/backdrop/backdrop.js
+++ b/components/backdrop/backdrop.js
@@ -3,7 +3,6 @@ import { css, html, LitElement } from 'lit';
 import { cssEscape, getComposedChildren, getComposedParent, isVisible } from '../../helpers/dom.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 
-export const BACKDROP_ROLE = 'data-d2l-backdrop-role';
 const BACKDROP_HIDDEN = 'data-d2l-backdrop-hidden';
 const BACKDROP_ARIA_HIDDEN = 'data-d2l-backdrop-aria-hidden';
 
@@ -68,7 +67,8 @@ class Backdrop extends LitElement {
 				transition: none;
 			}
 			@media (prefers-reduced-motion: reduce) {
-				:host {
+				:host,
+				:host([slow-transition]) {
 					transition: none;
 				}
 			}
@@ -162,20 +162,12 @@ function hideAccessible(target) {
 			if (path.indexOf(child) !== -1) continue;
 			if (child.hasAttribute(BACKDROP_HIDDEN)) continue;
 
-			const role = child.getAttribute('role');
-			if (role) child.setAttribute(BACKDROP_ROLE, role);
-			child.setAttribute('role', 'presentation');
-
-			if (child.nodeName === 'FORM' || child.nodeName === 'A') {
-				const ariaHidden = child.getAttribute('aria-hidden');
-				if (ariaHidden) child.setAttribute(BACKDROP_ARIA_HIDDEN, ariaHidden);
-				child.setAttribute('aria-hidden', 'true');
-			}
+			const ariaHidden = child.getAttribute('aria-hidden');
+			if (ariaHidden) child.setAttribute(BACKDROP_ARIA_HIDDEN, ariaHidden);
+			child.setAttribute('aria-hidden', 'true');
 
 			child.setAttribute(BACKDROP_HIDDEN, BACKDROP_HIDDEN);
 			hiddenElements.push(child);
-
-			hideAccessibleChildren(child);
 		}
 	};
 
@@ -204,21 +196,12 @@ export function preventBodyScroll() {
 function showAccessible(elems) {
 	for (let i = 0; i < elems.length; i++) {
 		const elem = elems[i];
-		const role = elem.getAttribute(BACKDROP_ROLE);
-		if (role) {
-			elem.setAttribute('role', role);
-			elem.removeAttribute(BACKDROP_ROLE);
+		const ariaHidden = elem.getAttribute(BACKDROP_ARIA_HIDDEN);
+		if (ariaHidden) {
+			elem.setAttribute('aria-hidden', ariaHidden);
+			elem.removeAttribute(BACKDROP_ARIA_HIDDEN);
 		} else {
-			elem.removeAttribute('role');
-		}
-		if (elem.nodeName === 'FORM' || elem.nodeName === 'A') {
-			const ariaHidden = elem.getAttribute(BACKDROP_ARIA_HIDDEN);
-			if (ariaHidden) {
-				elem.setAttribute('aria-hidden', ariaHidden);
-				elem.removeAttribute(BACKDROP_ARIA_HIDDEN);
-			} else {
-				elem.removeAttribute('aria-hidden');
-			}
+			elem.removeAttribute('aria-hidden');
 		}
 		elem.removeAttribute(BACKDROP_HIDDEN);
 	}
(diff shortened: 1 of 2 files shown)


components/backdrop/backdrop.js
```diff
@@ -162,20 +162,12 @@ function hideAccessible(target) {
 			if (path.indexOf(child) !== -1) continue;
 			if (child.hasAttribute(BACKDROP_HIDDEN)) continue;
 
-			const role = child.getAttribute('role');
-			if (role) child.setAttribute(BACKDROP_ROLE, role);
-			child.setAttribute('role', 'presentation');
-
-			if (child.nodeName === 'FORM' || child.nodeName === 'A') {
```
dlockhart, line 169: I vaguely recall discussing this code block before and questions around why exactly this was only targeting forms and links... it seems like it [was introduced in 2013 into our MVC dialog code](https://search.d2l.dev/diff/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Shim/Shim.js?r1=%2Flms%2Flp%2Fframework%2Fweb%2FD2L.LP.Web.UI%2FDesktop%2FControls%2FShim%2FShim.js%404130e589&r2=%2Flms%2Flp%2Fframework%2Fweb%2FD2L.LP.Web.UI%2FDesktop%2FControls%2FShim%2FShim.js%40186e88e1) with no comment about why unfortunately.

I _think_ we should be OK to do this, but it is mildly worrying especially if this is going to be hotfixed.

components/backdrop/backdrop.js
```diff
@@ -162,20 +162,12 @@ function hideAccessible(target) {
...
-				const ariaHidden = child.getAttribute('aria-hidden');
-				if (ariaHidden) child.setAttribute(BACKDROP_ARIA_HIDDEN, ariaHidden);
-				child.setAttribute('aria-hidden', 'true');
-			}
+			const ariaHidden = child.getAttribute('aria-hidden');
+			if (ariaHidden) child.setAttribute(BACKDROP_ARIA_HIDDEN, ariaHidden);
+			child.setAttribute('aria-hidden', 'true');
```
dlockhart, line 167: Note that we have duplicate-ish code [in MVC](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Shim/Shim.js?r=d3403c8c#157) (which legacy actually thankfully uses too). Curious if this bug also happens there? If so, we'll want to apply a similar fix.

components/backdrop/backdrop.js
```diff
@@ -204,21 +196,12 @@ export function preventBodyScroll() {
 function showAccessible(elems) {
 	for (let i = 0; i < elems.length; i++) {
 		const elem = elems[i];
-		const role = elem.getAttribute(BACKDROP_ROLE);
-		if (role) {
-			elem.setAttribute('role', role);
-			elem.removeAttribute(BACKDROP_ROLE);
+		const ariaHidden = elem.getAttribute(BACKDROP_ARIA_HIDDEN);
+		if (ariaHidden) {
+			elem.setAttribute('aria-hidden', ariaHidden);
```
dlockhart, line 201: I personally always prefer `hasAttribute` for this kind of thing, but this matches what the old code was doing so 🤷 .

```suggestion
		if (elem.hasAttribute(BACKDROP_ARIA_HIDDEN)) {
			elem.setAttribute('aria-hidden', elem.getAttribute(BACKDROP_ARIA_HIDDEN));
```

components/backdrop/backdrop.js
oelhanafey: Thanks, I'll have a look into that

components/backdrop/backdrop.js
oelhanafey: I think [this PR](https://github.com/BrightspaceUI/core/pull/2988) is what you're referring to. 

I played around with VO on the old backdrop and I was still able to focus elements that had the `role` but no `aria-hidden` so maybe it was added to prevent those from being focused. But then why wouldn't you just apply it to everything?

dlockhart: Thanks for finding that -- haha I guess this is us circling back!

Ok I sleuthed through the old Bitbucket archive and found where this [was first introduced](https://old-git.d2l.dev/projects/CORE/repos/lp/commits/bf730c5fa89badc28d012ffa1df721bd6256ee8e#framework/web/D2L.LP.Web.UI/Desktop/MasterPages/Dialog/Default/Dialog.js). The commit message is:
> When a dialog shims the main page, ensure that form controls have
aria-hidden=true set, to prevent AT such as JAWS or NVDA from passing control to
shimmed form elements. Closing the dialog properly resets the aria-hidden
attribute.

So that makes sense, and eases my concerns a bit. The code was [then moved into Shim.js here](Reduce number of H1 elements when modal dialogs are open.
This aligns the code in MVC and Legacy dialogs to ensure that main page elements
are set to aria presentation mode when a dialog is shimmed.
Legacy dialogs will now also use an h2 element instead of h1 for the d2l-heading-title element.) and that's where the `<a>` code was added. I'll dig further and see if I can find the old legacy version of this and see if there are more clues.

oelhanafey: That's lucky that they added such a detailed commit message 😄 

Thanks for finding this

dlockhart: Unfortunately I hit a brick wall on the `<a>` stuff. The closest I came was [this change](https://old-git.d2l.dev/projects/CORE/repos/lp/commits/f13f2fe0632cbf1efc558037b3db02383597b3a7#framework/web/D2L.Web/ui/Dialog/03_ModalDialog.js), which switched legacy dialogs to use the MVC shim code in 2013. But that [legacy shim code](https://old-git.d2l.dev/projects/CORE/repos/lp/browse/framework/web/D2L.Web/ui/page/JavaScript/ShimController.js) doesn't do anything with ARIA... so yeah, dead end.

oelhanafey: Yeah, I suspected it was something related to aria-hidden support. Thanks for searching into this, I'll continue to test it to make sure everything is working as expected.

components/backdrop/backdrop.js
oelhanafey: Doesn't appear to affect MVC Dialogs. Confirmed with Nathan that all reports have been related to creating/editing a quiz. I can't even get the LMS to crash in other areas that use the d2l-dialog.

oelhanafey: It might be a good idea anyway to update that code to be consistent with what we have here but we probably don't need to hotfix it

dbatiste (approved):
This change looks good to me, assuming we've done the thorough testing that Dave mentioned in his comments.

dlockhart (approved):

components/backdrop/backdrop.js
mdgbayly: I was able to reproduce the issue on a web component dialog from the top right ellipsis in Quick Eval, but yeh mostly just reproducable on FACE pages, with Android Chrome. And could not get it to fail with legacy dialogs even on FACE pages.

dlockhart: So bizarre! Reminds me of like IE6 memory leak issues from the 2000s haha.

PR #3205, opened 2023-01-26
fix: change box-sizing to border-box for dialogs

GZolla:


GZolla:
> I think this is "ok", but I don't necessarily think it's any better than your original proposal that included the border-width in the `calc(...)`. In either case, we're accounting for the border-width somewhere, and in this case, it ends up introducing it into JavaScript whereas with the other solution, it is isolated to CSS, which some might argue is preferred.

Makes sense, I thought something along those lines when committing the last change, it seems those calculations(allowing children nodes to extend the size of a parent until a reaches a `max-height`) should be possible in css, but they are not.

The two problems with the pure CSS solution are that we would need to add rules to every media query; and that the `width` attribute [this PR](https://github.com/BrightspaceUI/core/pull/3194) is adding would be creating dialogs 2px wider than the set value(unless JS is introduced there, which defeats the purpose of pure CSS).

dbatiste (approved):
Looks good.

PR #3211, opened 2023-01-27
Add localizeHTML method

bearfriend:
Adds basic markup capabilities to off-stack localization.

Uses a restricted version of Intl MessageFormat [Rich Text support](https://formatjs.io/docs/intl-messageformat/#rich-text-support) via a new localize-mixin method `localizeHTML`.

Markup replacements must use the `markup` (name could change) tag function on their templates, and elements are currently restricted to:
`d2l-link`, `d2l-tooltip-help`, `p`, `br`, `strong`, `em`, `b`, and `i`

[US141398](https://rally1.rallydev.com/#/?detail=/userstory/643224646457&fdp=true): Localization - Consumers format lang-terms containing links, and possibly other tags

There were quite a few PRs that all ultimately were combined here. To make it easier to find conversations etc. I've listed them below:

- https://github.com/BrightspaceUI/core/pull/3370
- https://github.com/BrightspaceUI/core/pull/3380
- https://github.com/BrightspaceUI/core/pull/3381
- https://github.com/BrightspaceUI/core/pull/3386
- https://github.com/BrightspaceUI/core/pull/3470


mixins/localize-dynamic-mixin.js
```diff
@@ -27,6 +27,13 @@ export const LocalizeDynamicMixin = superclass => class extends LocalizeMixin(su
 					);
 				}
 
+				this.__resolvedLanguage = {
+					chain: langs,
+					resolved: lang,
+					file: '',
+					resources
+				};
```
bearfriend, line 35: Whoops. This snuck in from another branch.
```suggestion
```

mixins/localize-mixin.js
```diff
@@ -128,6 +136,28 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
...
+	localizeHTML(key, { _links = '', _html = '', _tooltips = '', ...replacements } = {}) {
+		const links = [].concat(_links);
+		const tooltips = [].concat(_tooltips);
+		const html = [].concat(_html);
+		const localized = this.localize(key, replacements);
+		const sanitized = localized.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+		const markedUp = sanitized
+			// replace [a], [a 0] ... [a 9], [/a]
+			.replace(/\[a( [0-9])?\]([^]*?)\[\/a\]/g, (m, n, t) => `<d2l-link ${links[Number(n || 0)] || ''}>${t}</d2l-link>`)
+			// replace [tt], [tt 1] ... [tt 9], [/tt]
```
bearfriend, line 148:
```suggestion
			// replace [tt], [tt 0] ... [tt 9], [/tt]
```

mixins/localize-mixin.js
```diff
@@ -128,6 +136,28 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
...
+			.replace(/\[tt( [0-9])?\]([^]*?)\[\/tt\]/g, (m, n, t) => `<d2l-tooltip-help inherit-font-style text="${t}">${tooltips[Number(n || 0)] || ''}</d2l-tooltip-help>`)
+			// UNDOCUMENTED: replace [html], [html 0] ... [html 9]
+			.replace(/\[html( [0-9])?\]/g, (m, n) => html[Number(n || 0)] || '')
+			// replace good-listed markup
+			.replace(LocalizeMixinClass.markupRegex, k => markupMap[k] || k);
+
+		if (_links.length) import('../components/link/link.js');
+		if (_tooltips.length) import('../components/tooltip/tooltip-help.js');
```
dlockhart, line 156: I was wondering how we'd do this -- cool!

mixins/localize-mixin.md
```diff
@@ -121,6 +121,49 @@ If your localized string contains arguments, pass them as a key-value object as
...
+#### `[br]`
+The `[br]` markup converts to the `<br />` tag
+
+
+### Complex markup
+
+Complex markup is more powerful, but requires passing additional information in the second parameter
+
+#### `[a]Linked[/a]`
+The `[a]` markup converts to the `<d2l-link>` tag, and takes an attributes string in the `_links` property
```
dlockhart, line 149: Wondering if this should be `[link]` in order to align better with SML? 🤔

mixins/localize-mixin.md
```diff
@@ -121,6 +121,49 @@ If your localized string contains arguments, pass them as a key-value object as
...
+
+#### `[tt]Tooltipped[/tt]`
+The `[tt]` markup converts to the `<d2l-tooltip-help>` tag, and takes tooltip text in the `_tooltips` property
```
dlockhart, line 152: `<tt>` is [already an element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt), so I think we'll want to go with something like `[tooltip-help]` here to be explicit.

mixins/localize-mixin.md
```diff
@@ -121,6 +121,49 @@ If your localized string contains arguments, pass them as a key-value object as
...
+
+Complex markup can be used multiple times by including an index and passing an array of its required information
+##### Example
+
+```javascript
+// en.js
+export default {
+  greetingLinks: `Hi {name}, here's [a]a link[/a]. Here's [a 1]another[/a].`
```
dlockhart, line 160: What's the purpose of the `1` here? Just to identify the link in case they need to be flipped around in the translation?

Maybe we could give them a friendly name identifier similar to how replacements work (e.g. `[a home]` and `[a back]`) and then use those identifiers as keys in an object (instead of an array)?

mixins/localize-mixin.md
```diff
@@ -121,6 +121,49 @@ If your localized string contains arguments, pass them as a key-value object as
...
+};
+```
+```javascript
+this.localizeHTML('greetingLinks', { name: 'Mary', _links: ['href="a.html"', 'href="b.html" target="_blank"'] });
```
dlockhart, line 164: This is definitely the most flexible approach. Wondering about pros & cons of it being an object (example `{ 'href': 'a.html', primary: true}`)? Thinking about things like HTML/URL encoding in the string...

dlockhart (commented):
So cool! Haha I guess if we just delay long enough, you're going to implement the entire localization roadmap? Sweet!

mixins/localize-mixin.md
bearfriend: I've gone back and forth on it, because

1. It's not the same as the SML. Customers really shouldn't be using any of these, as they can't know if `localizeHTML` is being called, and even if they try it on one that is, links and tooltips (and html) can't be provided with the data they need from the lang term (nor should they be, I think).
2. I was trying to keep everything as short as possible for readability/translators, but maybe having a useful name adds enough context to actually be more helpful.

mixins/localize-mixin.md
bearfriend: Oof, yeah. I'm leaning toward the more contextually-helpful names now.

mixins/localize-mixin.md
bearfriend: Yep, we can't rely on the English order to fill them in. I like the friendly names/object idea.

mixins/localize-mixin.md
bearfriend: I also thought about using an object, but ultimately was annoyed with the verbosity/html-to-js conversion brain cycles. Also would mean more maintenance if component APIs change, unless we did type checking I guess.

mixins/localize-mixin.md
dlockhart: Agreed that since it's not the same as SML, we shouldn't feel obliged to align with it... and yeah maybe even purposefully distance ourselves from it to avoid confusion. I also agree that the data links/tooltips/etc. need shouldn't come from the lang term at all.

mixins/localize-mixin.js
```diff
@@ -128,6 +136,29 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
...
+	localizeHTML(key, {
+		_link = '', _tooltip = '', _html = '',
+		_links = {}, _tooltips = {},
+		...replacements
+	} = {}) {
+		const localized = this.localize(key, replacements);
+		const sanitized = localized.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+		const markedUp = sanitized
+			// replace [link], [link key], [/link]
+			.replace(/\[a(?: ?([\w\d-]*))?\]([^]*?)\[\/a\]/g, (m, n, t) => `<d2l-link ${_links[n] || _link}>${t}</d2l-link>`)
```
bearfriend, line 148: I stuck with `a` for now. I don't have a strong opinion one way or the other, though.

mixins/localize-mixin.js
```diff
@@ -128,6 +136,29 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
...
+			// replace [tooltip-help], [tooltip-help key], [/tooltip-help]
+			.replace(/\[tooltip-help(?: ?([\w\d-]*))?\]([^]*?)\[\/tooltip-help\]/g, (m, n, t) => `<d2l-tooltip-help inherit-font-style text="${t}">${_tooltips[n] || _tooltip}</d2l-tooltip-help>`)
+			// UNDOCUMENTED: replace [html]
+			.replace('[html]', _html || '')
```
bearfriend, line 152: I've reduced this down to a simple replace. I think this should remain undocumented for now, and I don't want to encourage overcomplicating langterms with lots of injected HTML. We can expand it later if truly needed.

mixins/localize-mixin.js
```diff
@@ -136,24 +136,25 @@ export const LocalizeMixin = dedupeMixin(superclass => class LocalizeMixinClass
...
-			// replace [a], [a 0] ... [a 9], [/a]
-			.replace(/\[a( [0-9])?\]([^]*?)\[\/a\]/g, (m, n, t) => `<d2l-link ${links[Number(n || 0)] || ''}>${t}</d2l-link>`)
-			// replace [tt], [tt 1] ... [tt 9], [/tt]
-			.replace(/\[tt( [0-9])?\]([^]*?)\[\/tt\]/g, (m, n, t) => `<d2l-tooltip-help inherit-font-style text="${t}">${tooltips[Number(n || 0)] || ''}</d2l-tooltip-help>`)
-			// UNDOCUMENTED: replace [html], [html 0] ... [html 9]
-			.replace(/\[html( [0-9])?\]/g, (m, n) => html[Number(n || 0)] || '')
+			// replace [link], [link key], [/link]
+			.replace(/\[a(?: ?([\w-]*))?\]([^]*?)\[\/a\]/g, (m, n, t) => `<d2l-link ${_links[n] || _link}>${t}</d2l-link>`)
+			// replace [tooltip-help], [tooltip-help key], [/tooltip-help]
+			.replace(/\[tooltip-help(?: ?([\w-]*))?\]([^]*?)\[\/tooltip-help\]/g, (m, n, t) => `<d2l-tooltip-help inherit-font-style text="${t}">${_tooltips[n] || _tooltip}</d2l-tooltip-help>`)
```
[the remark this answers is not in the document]
dlockhart: Maybe `href`s only need to be HTML-encoded... 🤔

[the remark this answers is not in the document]
bearfriend: The text from the lang file (i.e. `${t}`) definitely needs to replace double quotes so that `[tooltip-help]text" onmouseover="import('malicious.js'[/tooltip-help]` doesn't work. And only because it gets moved to the attribute value.

The rest of the values are passed from the code or have `<` and `>` replaced, making them unable to produce parseable html, so shouldn't need to be modified further unless I'm totally missing something.

mixins/demo/localize-mixin.html
```diff
@@ -6,21 +6,16 @@
...
 	<body unresolved>
 		<d2l-demo-page page-title="localize-mixin">
 
 			<h2>localize()</h2>
-
 			<d2l-demo-snippet>
-
 				<script type="module">
+import { html, LitElement } from 'lit';
+import { LocalizeDynamicMixin } from '../../mixins/localize-dynamic-mixin.js';
```
bearfriend, line 18: I think this will show up converted to a full webserver path, which is a little odd, but not a huge problem.

mixins/demo/localize-mixin.html
```diff
@@ -55,14 +59,15 @@ <h2>localize()</h2>
...
 			<d2l-demo-snippet>
 				<script type="module">
+import { html, LitElement } from 'lit';
+import { LocalizeDynamicMixin } from '../../mixins/localize-dynamic-mixin.js';
 class Mission extends LocalizeDynamicMixin(LitElement) {
 
 	static get localizeConfig() {
 		const langResources = {
-			'en': { mission: `Transforming the way [a]the world[/a] learns` },
+			'en': { mission: `[em]Transforming[/em] the way [a]the world[/a] learns` },
```
bearfriend, line 70:
```suggestion
			'en': { mission: `[i]Transforming[/i] the way [a]the world[/a] learns` },
```

mixins/demo/localize-mixin.html
```diff
@@ -71,7 +71,7 @@ <h2>localizeHTML()</h2>
 	}
 
 	render() {
-		const _link = `href="https://.wikipedia.org/wiki/Earth" target="_blank"`;
+		const _link = `href="https://wikipedia.org/wiki/Earth" target="_blank""`;
```
bearfriend, line 79: fix typo typo?
```suggestion
		const _link = `href="https://wikipedia.org/wiki/Earth" target="_blank"`;
```

bearfriend:
Much improved markdown docs, thanks. 

Besides comments above, approved ✅

mixins/demo/localize-mixin.html
bearfriend: Correction: Just converts to a relative path into `node_modules`, which might be a bad example of how to import it, but still not a problem.

mixins/demo/localize-mixin.html
dlockhart: I was just testing it out -- forgot to remove, thanks.

mixins/demo/localize-mixin.html
```diff
@@ -6,14 +6,90 @@
...
 		</script>
 	</head>
 	<body unresolved>
 		<d2l-demo-page page-title="localize-mixin">
+
+			<h2>localize()</h2>
 			<d2l-demo-snippet>
-				<d2l-test-localize name="Bill"></d2l-test-localize>
+				<script type="module">
+import { html, LitElement } from 'lit';
```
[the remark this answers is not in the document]
bearfriend: That would appear to be an issue of not having `--node-resolve` on the s3 version. It works fine locally with the web server.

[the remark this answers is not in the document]
bearfriend: I removed that test file because it was being used for two different things and had lost it's original purpose, which was to test synchronous langterm retrieval. So I rewrote the demo page to include the relevant code for using the mixin in the `demo-snippet` (not very helpful without it), and modified the tests to include both sync and async setups internally.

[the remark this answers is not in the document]
dlockhart: I'm going to re-introduce a couple files for these, but we won't use them from the tests. :)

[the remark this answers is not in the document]
bearfriend: Sorry, not "not having `--node-resolve`". I mean, it not functioning the same as the local web server. I haven't looked at the way the s3 version works.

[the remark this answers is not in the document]
bearfriend: It's unfortunate that the mixins can't have more useful demo code displayed, since they are not just element APIs.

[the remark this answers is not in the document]
dlockhart: For sure, yeah.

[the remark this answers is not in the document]
eKoopmans: Yep I chatted briefly with @dlockhart about this - `web-dev-server` is doing some extra magic that `rollup` doesn't (and can't easily?) do when building.

eKoopmans (commented):
Sorry to be coming in so late with this feedback, but I'm not sure about this approach personally. Just wondering if you considered a more explicit approach to attributes, something like:

```js
// en.js:
{ termName: '[a href="{myUrl}"]Some link text[/a]' }

// Usage:
this.localizeHtml('termName', {
  myUrl: 'home.html'
});

// Then localizeHtml has a list of attributes it knows about for each element, parses them, ignores the rest...
```

mixins/localize-mixin.js
```diff
@@ -128,6 +143,29 @@ export const LocalizeMixin = dedupeMixin(superclass => class extends superclass
 
 	}
 
+	localizeHTML(key, {
+		link = '', tooltipHelp = '', _html = '',
+		links = {}, tooltipHelps = {},
```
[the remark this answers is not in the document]
bearfriend: 1. Can't say I love it myself, but I'm less a fan of allowing and maintaining multiple types in one argument. We could simply say it has to be named and provide an object, but since probably 99% of cases are going to use a single link, I wanted to avoid that.
2. Yeah, I'd not be opposed to using `[link]` to match
3. The text should definitely be run through `this.localize` _almost_ every time, but that's the consumer's responsibility. Maybe it's user-provided. Maybe it's server-generated.

As for providing the link attributes, I agree it's not super pretty. It's just super simple. There was previous discussion [here](https://github.com/BrightspaceUI/core/pull/3211#discussion_r1089089517).

mixins/demo/localize-mixin-mission.js
```diff
@@ -0,0 +1,24 @@
...
+		};
+		return {
+			importFunc: async lang => langResources[lang]
+		};
+	}
+
+	render() {
+		const _link = 'href="https://wikipedia.org/wiki/Earth" target="_blank"';
+		return html`
+			<p>${this.localizeHTML('mission', { _link })}</p>
```
eKoopmans, line 19: Not getting a link from this - should it be:

```suggestion
			<p>${this.localizeHTML('mission', { link: _link })}</p>
```

bearfriend, line 19:
```suggestion
		const link = 'href="https://wikipedia.org/wiki/Earth" target="_blank"';
		return html`
			<p>${this.localizeHTML('mission', { link })}</p>
```

mixins/localize-mixin.md
```diff
@@ -121,6 +121,65 @@ If your localized string contains arguments, pass them as a key-value object as
...
+
+The `[b]bold[/b]` markup converts to the `<strong>` tag.
+
+#### Italic
+
+The `[i]italic[/i]` markup converts to the `<em>` tag.
+
+#### Line Breaks
+
+The `[br]` markup converts to the `<br>` tag.
```
dbatiste, line 144: Have paragraphs `<p>` been considered? Paragraph elements support styling in ways that `<br>` does not - spacing is the one that generally comes to mind. I think screen readers can also have short-cuts for moving the virtual cursor between paragraphs. tinyMCE switched to paragraphs 10+ years ago so all of our user-authored HTML generally consists of them.

mixins/localize-mixin.md
```diff
@@ -121,6 +121,65 @@ If your localized string contains arguments, pass them as a key-value object as
...
+  links: {
+    home: 'href="home.html"',
+    back: 'href="back.html"'
+  }
+});
+```
+
+#### Help Tooltips
+
+The `[tooltip-help]help tooltip[/tooltip-help]` markup converts to the `<d2l-tooltip-help>` tag, and takes tooltip text in the `tooltipHelp` property.
```
dbatiste, line 175: Would nesting work? For example: `[tooltip-help][b]help[/b] tooltip[/tooltip-help]`.

Worth noting that tooltips cannot contain complex semantics, and we force text semantics to ensure screen readers don't discard the content entirely.

bearfriend: A simpler nesting is tested [here](https://github.com/BrightspaceUI/core/pull/3211/files#diff-516e261002a7e8e2dd6f3717ecdd945ef45431ce53e2979e666a54cf0cdaf5c9R132), but I think they should all work.

mixins/test/localize-mixin.test.js
```diff
@@ -126,6 +125,23 @@ const Test3LocalizeDynamicMixn = superclass => class extends LocalizeDynamicMixi
 	}
 };
 
+const Test1LocalizeHTML = superclass => class extends LocalizeStaticMixin(superclass) {
+	static get resources() {
+		return {
+			en: {
+				test1: 'This is [b]important[/b], this is [b][i]very important[/i][/b], and this is [span]not[/span]',
```
dbatiste, line 132: Technically this is valid from an HTML perspective... `I <strong>like <em>jelly</strong> donuts</em>.`. Obviously this sort of thing only make sense for certain types of elements. 

Would `I [b]like [i]jelly[/b] donuts[/i]` work or would it explode? I'm not sure if I care that much about this edge case, but if it _does_ work, maybe we should have a test for it?

bearfriend: The simple markup opening and closing tags are replaced separately, ignorant of their meaning or semantics, so it's just going to map to however you format it.

mixins/localize-mixin.md
bearfriend: And by "work", I mean not break. Your example above should output whatever `<d2l-tooltip-help text="<strong>help</strong> tooltip">...</d2l-tooltip-help>` does, which I assume doesn't include actual bolded text.

mixins/localize-mixin.md
```diff
@@ -121,6 +121,65 @@ If your localized string contains arguments, pass them as a key-value object as
...
+
+## `localizeHTML()`
+
+You can include some basic markup in your localization resources and safely convert them to HTML with the `localizeHTML()` method.
+
+### Simple Markup
+
+#### Bold
+
+The `[b]bold[/b]` markup converts to the `<strong>` tag.
```
dbatiste, line 136: Are we using the right token or tag names? I hate to second guess this, but this is probably worth reviewing before we commit one way or another. We seem to be using them interchangeably, where Mozilla seems to suggest we should not. Similar story for italic.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b#usage_notes
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#b_vs._strong

dbatiste: And to be fair... their explanation doesn't make it much clearer to me. 😕 

>Don't they both do the same thing?"
...
>Not exactly. The `<strong>` element is for content that is of greater importance, while the `<b>` element is used to draw attention to text without indicating that it's more important.

bearfriend: I intentionally chose to use `em` and `strong` as they have actual meaning to add to words within a sentence. I figured if you were trying to style something, you should probably be styling the entire thing, which can be done with CSS in your actual code.

eKoopmans:
Hi everyone, I've been investigating an alternative using `intl-messageformat`. In the description:

> Reasons not to use `IntlMessageFormat` [Rich Text support](https://formatjs.io/docs/intl-messageformat/#rich-text-support):
> 
> 1. Does not support attributes on tags
> 2. Does not support self-closing tags (e.g. `<br />`)
> 3. HTML returned from translation files would need to be trusted or parsed in a more complicated way

I believe that points 1 and 3 are actually avoidable! I have a POC and would like to demo it, I think there is a possibly more robust and standard way of doing things with `IntlMessageFormat` directly.

eKoopmans:
Opened a draft PR with the POC, #3370.

mixins/test/localize-mixin.test.js
eKoopmans: Interesting - I think that's actually an error in HTML5, but will still be parsed:
https://stackoverflow.com/a/28840806

(that SO answer quotes step 8, in the spec it looks like that's changed to step 6, but still the same idea)

dbatiste: Haha, I feel like I just dated myself. 😬

helpers/localize.js
```diff
@@ -0,0 +1,30 @@
...
+export function markup(strings, ...expressions) {
+	strings.forEach(str => validateMarkup(str, true));
+	expressions.forEach(exp => validateMarkup(exp, true));
+	return { ...html(strings, ...expressions), _markup: true };
+}
+
+export const linkGenerator = ({ href, target }) => {
+	import('../components/link/link.js');
+	return chunks => markup`<d2l-link href="${ifDefined(href)}" target="${ifDefined(target)}">${chunks}</d2l-link>`;
+};
```
eKoopmans, line 30: We likely want to add a helper for tooltip-help, something like:

```js
export const tooltipHelpGenerator = ({ text }) => {
	import('../components/tooltip/tooltip-help.js');
	return chunks => markup`<d2l-tooltip-help text="${ifDefined(text)}">${chunks}</d2l-tooltip-help>`;
};
```

P.S. I'm surprised that for tooltip-help, the `text` is what's displayed inline, and the slot contents are in the tooltip - would have guessed the other way around!

mixins/localize-mixin.md
eKoopmans, line 1: Updates here and to tests will be needed once all the rest of the approach is locked in and confirmed.

eKoopmans (commented):
So excited for this to be on (hopefully) the home stretch! I've filled in my thoughts on the remaining open questions and work to do, the rest of the team is going to have a look soon.

mixins/localize-mixin.md
dlockhart: Coming back to this... `<strong>` vs. `<b>` and `<em>` vs. `<i>` is a [semantic vs. stylistic](https://stackoverflow.com/questions/271743/whats-the-difference-between-b-and-strong-i-and-em) choice. I agree with Danny that styles can be applied outside of localization, which makes me want to _not_ support `<b>` or `<i>` at all -- only `<strong>` and `<em>`.

helpers/localize.js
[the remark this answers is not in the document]
dlockhart: My vote would actually be for either moving them directly into `localize-mixin.js` or at the very least re-exporting the useful ones from that file. That way folks only need to have 1 import statement:

```javascript
import { linkGenerator, LocalizeMixin, markup } from '@brightspace-ui/core/mixins/localize-mixin.js';`
```

helpers/localize.js
```diff
@@ -0,0 +1,30 @@
...
+	if (content) {
+		if (content.map) return content.map(item => validateMarkup(item));
+		if (content._markup) return content;
+		if (Object.hasOwn(content, '_$litType$')) throw markupError;
+		if (applyRegex && content.constructor === String && disallowedTagsRegex.test(content)) throw markupError;
+	}
+	return content;
+}
+
+export function markup(strings, ...expressions) {
```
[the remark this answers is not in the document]
dlockhart: I like `localizeMarkup`, which aligns with `validateMarkup`.

helpers/localize.js
```diff
@@ -0,0 +1,30 @@
...
+	strings.forEach(str => validateMarkup(str, true));
+	expressions.forEach(exp => validateMarkup(exp, true));
+	return { ...html(strings, ...expressions), _markup: true };
+}
+
+export const linkGenerator = ({ href, target }) => {
```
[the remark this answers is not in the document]
dlockhart: My vote is `generateLink`, `generateTooltipHelp`, etc. That way they're all alphabetically together in import statements and you don't really need to remember their names or which ones exist -- you can just start typing "generat..." and it'll fill in what's available.

mixins/demo/localize-mixin-mission.js
```diff
@@ -0,0 +1,44 @@
...
+			${this.localizeHTML('mission', replacements)}
+			<div>1. ${this.localizeHTML('a', { a })}</div>
+			<div>2. ${this.localizeHTML('a', { a: null })}</div>
+			<div>3. ${this.localizeHTML('a', { a: 1 })}</div>
+			<div>4. ${this.localizeHTML('a', { a: {} })}</div>
+			<div>5. ${this.localizeHTML('a', { a: false, b: new Date() })}</div>
+			<div>6. ${this.localizeHTML('a', { a: true, c: [1, 2, 3] })}</div>
+			<div>7. ${this.localizeHTML('a', { a: true, c: markup`<b>bold</b>` })}</div>
+			<div>8. ${this.localizeHTML('a', { a: true, c: [markup`<br>`] })}</div>
+			<div>9. ${this.localizeHTML('a', { a: true, c: `<test>` })}</div>
```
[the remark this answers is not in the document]
dlockhart: Agreed.

bearfriend:
Getting back to this today. I just cleaned up and committed what I had in progress, but I'll go back through the last comments now.

dlockhart:
Oh wow crazy timing, I had also started working on this today to try out some of the suggestions. Will coordinate.

bearfriend:
> Oh wow crazy timing, I had also started working on this today to try out some of the suggestions. Will coordinate.

Oh, shoot. Ok, let's chat and make sure I didn't blow anything up.

mixins/test/localize-mixin.test.js
```diff
@@ -297,6 +328,50 @@ describe('LocalizeMixin', () => {
...
+			const manual = elem.localizeHTML('test2', { link: chunks => markup`<d2l-link href="http://d2l.com">${chunks}</d2l-link>` });
+			const disallowed = elem.localizeHTML('test3', { link: chunks => markup`<div>${chunks}</div>` });
+			const badTemplate = elem.localizeHTML('test3', { link: chunks => html`${chunks}` });
+			const tooltip = elem.localizeHTML('test4', { tooltip: generateTooltipHelp({ contents: 'Tooltip text' }) });
+
+			const items = ['milk'];
+			const pluralLink = elem.localizeHTML('pluralTest', { itemCount: items.length, item: items[0], link: generateLink({ href: 'checkout' }) });
+
+			items.push('bread', 'eggs');
+			const val8 = elem.localizeHTML('pluralTest', { itemCount: items.length, link: generateLink({ href: 'checkout' }), html: () => items.map(i => markup`<p>${i}</p>`)});
```
bearfriend: Sad little name.
```suggestion
			const pluralMap = elem.localizeHTML('pluralTest', { itemCount: items.length, link: generateLink({ href: 'checkout' }), html: () => items.map(i => markup`<p>${i}</p>`)});
```

mixins/test/localize-mixin.test.js
```diff
@@ -297,6 +328,50 @@ describe('LocalizeMixin', () => {
...
+				const { strings, values } = data;
+				const flatValues = [...values, ''].map((v, idx) => {
+					if (typeof v === 'object') return getRenderString(v);
+					if (v.toString() === 'Symbol(lit-nothing)') {
+						return '{lit-nothing}'
+					}
+					return v;
+				});
+				return strings.reduce((acc, s, idx) => acc + s + flatValues[idx], '').replace(/\s[\w-]+="{lit-nothing}"/g, '');
+			}
```
bearfriend: This is basically just looping through all the template parts and building up the final string. I did this instead of actually rendering each thing out, as that's not really what we're trying to test here, but I'm not tied to it.

mixins/test/localize-mixin.test.js
```diff
@@ -297,6 +329,104 @@ describe('LocalizeMixin', () => {
...
+			expect(getRenderString(disallowed)).to.equal('This is <link>replaceable</link>');
+			expect(getRenderString(badTemplate)).to.equal('This is replaceable');
+			expect(getRenderString(tooltip)).to.equal('This is a <d2l-tooltip-help inherit-font-style text="tooltip-help">Tooltip text</d2l-tooltip-help> within a sentence');
+			expect(getRenderString(pluralLink)).to.equal('You have milk in your cart. <d2l-link href="checkout">Checkout</d2l-link>');
+			expect(getRenderString(pluralMap)).to.equal('Items in your cart:<p>milk</p><p>bread</p><p>eggs</p><d2l-link href="checkout">Checkout</d2l-link>');
+		});
+
+		let a;
+		[{
+			get renderString() { return elem.localizeHTML('typeChecker', { a }); },
```
bearfriend, line 380: I guess this should just be `undefined` but it hurts.
```suggestion
		[{
			get renderString() { return elem.localizeHTML('typeChecker', { a: undefined }); },
```

eKoopmans: Haha yep, I'd say so. I'm curious about your reluctance towards the [`undefined` primitive](https://developer.mozilla.org/en-US/docs/Glossary/Undefined) actually, that's a new one to me.

bearfriend: Basically just the [eslint writeup](https://eslint.org/docs/latest/rules/no-undefined) on it. Just one of those things I've avoided for years because it's fairly easy to do usually.

eKoopmans: Neat! I didn't know undefined could be overridden, I assumed it was equivalent to `null`.

mixins/test/localize-mixin.test.js
```diff
@@ -297,6 +329,89 @@ describe('LocalizeMixin', () => {
 
 	});
 
+	describe('localizeHTML', async() => {
+
+		const elem = await fixture(`<${localizeHTMLTag}></${localizeHTMLTag}>`);
+
+		const getRenderString = data => {
+			render(data, elem);
+			return elem.innerHTML.replace(/<!--[^]*?-->/g, '');
```
bearfriend, line 338: It hadn't occurred to me that `render` was something I could import. We still have to attach it to the DOM instead of just getting a string, but I think this is still better than my previous custom function. It improved the output a bit by showing html entities where we would expect rejected/non-templated markup to fail.

helpers/localize.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function generateLink({ href, target }) {
+	import('../components/link/link.js');
+	return chunks => localizeMarkup`<d2l-link href="${ifDefined(href)}" target="${ifDefined(target)}">${chunks}</d2l-link>`;
+}
+
+export function generateTooltipHelp({ contents }) {
+	import('../components/tooltip/tooltip-help.js');
+	return chunks => localizeMarkup`<d2l-tooltip-help inherit-font-style text="${ifDefined(chunks)}">${contents}</d2l-tooltip-help>`;
```
eKoopmans, line 33: Are we sure we want to invert the signature of tooltip-help like this? A couple thoughts:

- I don't like breaking the pattern of mirroring the component, with chunks being the contents and any attributes being passed into the function
- This could end up pretty broken - chunks can be anything, not just a string, while the `text` attribute needs a string only

mixins/test/localize-mixin.test.js
eKoopmans: I haven't tried it myself, but could you use `expect(el).dom.to.equal('...')` from [`@open-wc/semantic-dom-diff`](https://open-wc.org/docs/testing/testing-package/#chai), instead of getting the innerHTML and stripping comments?

eKoopmans:
BTW the failing list visual-diff should be fixed if you merge or rebase onto main, it's a flakey test that Stacey just fixed.

mixins/test/localize-mixin.test.js
bearfriend: Works wonderfully!

helpers/localize.js
bearfriend: I'm not sure what you mean. This is how the component API works. I get that it's weird in an inline context like this, but without changing the component I don't see what can be done here.

It can only be as broken as when a dev adds the same stuff to `text` without doing so through `localizeHTML`. It'll be pretty obvious that it's broken before it goes off for translation.

bearfriend: Oh, yeah it just can't work this way because of translations. We would be able to construct a regex that would split out the `chunks` into a whole separate term to translate (that would get put back together automatically), but the `text` is part of the original langterm so there's no way to get that from code into the translation file to go out as one whole langterm.

bearfriend: Also, we always want `inherit-font-style`, no? If it's not meant to be inline then it doesn't need to be in the langterm.

eKoopmans: Okay, sounds reasonable 👍. I think we may be good to go then, except for the readme!

dlockhart: I can take a swing at the README.

mixins/localize/README.md
```diff
@@ -121,6 +121,66 @@ If your localized string contains arguments, pass them as a key-value object as
 
 ```javascript
 render() {
-  return html`<p>${this.localize('hello', {firstName: 'Mary'})}</p>`;
+  return html`<p>${this.localize('hello', { firstName: 'Mary' })}</p>`;
 }
 ```
+
+## `localizeHTML()`
```
dlockhart, line 128: I've updated this to reflect the new state of things. I purposefully didn't document advanced usage where consumers could pass anything they want... I figured we could keep that to ourselves for now in the hopes that we could create more helpers for commonly requested things.

mixins/localize/README.md
```diff
@@ -127,59 +127,60 @@ render() {
...
+* `<b>bold</b>`
+* `<strong>strong</strong>`
+* `<i>italic</i>`
+* `<em>emphasis</em>`
 
-#### Italic
+Remember that `<strong>` is for content of greater importance (browsers show this visually using bold), while `<b>` only bolds the text visually without increasing its importance.
 
-The `[i]italic[/i]` markup converts to the `<em>` tag.
+Similarly `<em>` *emphasizes* a particular piece of text (browsers show this visually using italics), whereas `<i>` only italicizes the text visually without emphasis.
```
eKoopmans: Currently our handling makes `<b>` and `<strong>` result in the exact same `<strong>` HTML (and likewise with i/em). Is that fine, or do you want them to actually keep their meaning?

mixins/localize/README.md
```diff
@@ -127,59 +127,60 @@ render() {
...
+Then import `generateTooltipHelp` and pass it the tooltip term value:
 
 ```javascript
-this.localizeHTML('termName', {
-  tooltipHelp: 'Cephalopods are members of the molluscan class Cephalopoda'
+import { generateTooltipHelp } from '@brightspace-ui/core/mixins/localize/localize-mixin.js';
+
+this.localizeHTML('octopus', {
+  tooltip: generateTooltipHelp(this.localize('cephalopodTooltip'))
 });
```
eKoopmans: This doesn't match the signature of `generateTooltipHelp`, should look like:

```suggestion
this.localizeHTML('octopus', {
  tooltip: generateTooltipHelp({ contents: this.localize('cephalopodTooltip') })
});
```

eKoopmans (commented):
Thanks Dave, we're so close!

mixins/localize/README.md
dlockhart: Good catch -- updated!

mixins/localize/README.md
dlockhart: I think they should keep their meaning to avoid confusion. I've updated the replacement code to do this.

bearfriend:
I would approve this if I could ✅

Excellent work, everyone. So glad this landed where it did.

PR #3213, opened 2023-01-27
Override quotes rule for lang/ in eslintrc

bearfriend:


dlockhart (approved):
Thanks, that's much nicer.

PR #3215, opened 2023-01-30
Update focus-visible for several components

dbatiste:
This PR is a dependency of https://github.com/Brightspace/brightspace-integration/pull/8345.

This PR adds a `isFocusVisibleSupported` and updates `link.js` to conditionally render the CSS based on support. This enables us to use `:focus-visible` in browsers that support it, and falling back to `:focus` for focus indicator styles [in browsers that do not](https://caniuse.com/?search=focus-visible). It does not try to polyfill the focus-visible behaviour. It also relies on the `.d2l-focus-visible-not-supported` being set by BSI for the `.d2l-link` styles that are generated from Sass and rendered in the global scope.


components/link/link.js
```diff
@@ -3,6 +3,23 @@ import { css, html, LitElement } from 'lit';
...
+		text-decoration: underline;
+	}
+` : css`
+	.d2l-link:focus {
+		border-radius: 3px;
+		outline: 2px solid var(--d2l-color-celestine);
+		outline-offset: 1px;
+		text-decoration: underline;
+	}
+`;
```
dlockhart, line 22: Wonder if doing something like below might reduce the duplication and chance that we'll make a change to one but not the other one day? Maybe Lit won't even let you though.

```suggestion
const linkFocusPseudoClass = isFocusVisibleSupported() ? 'focus-visible' : 'focus';
const linkFocusStyles =  css`
	.d2l-link:${unsafeCSS(linkFocusPseudoClass)} {
		border-radius: 3px;
		outline: 2px solid var(--d2l-color-celestine);
		outline-offset: 1px;
		text-decoration: underline;
	}
`;
```

components/link/link.scss
```diff
@@ -14,12 +14,17 @@
...
 		text-decoration: underline;
 	}
-	&:focus,
-	&.d2l-link-focus {
+	&:focus-visible {
 		outline: 2px solid $d2l-color-celestine;
 		outline-offset: 1px;
 		text-decoration: underline;
 	}
+	.d2l-focus-visible-not-supported &:focus {
```
dlockhart, line 22: I think this can just be combined with the above rule? I'd want to inspect the CSS that this outputs though since remember at least old-Safari chokes on everything in a list of rules if it contains `:focus-visible`.

dlockhart (commented):
Interesting idea! I do like the compromise, and especially our ability to serve modern CSS to modern browsers without needing the slow polyfill.

components/link/link.js
dbatiste: Yep! That would be good, and I don't see any reason to avoid `unsafeCSS` in this case.

components/link/link.scss
dbatiste: I can look into that. Yeah, that was my worry... that Safari would choke.

components/link/link.js
dbatiste: I made this change with a separate commit since it required adding the `unsafeCSS` import as well.

components/link/link.scss
dbatiste: If I combine them, then this is what it produces:
```css
.d2l-test-link:focus-visible,
.d2l-focus-visible-not-supported .d2l-test-link:focus {
  outline: 2px solid #006fbf;
  outline-offset: 1px;
  text-decoration: underline;
}
```
Which from my understanding, we wouldn't want.

dlockhart: Correct. Ah well, thanks for trying!

components/menu/menu-item-styles.js
```diff
@@ -20,9 +23,9 @@ export const menuItemStyles = css`
 		color: var(--d2l-menu-foreground-color-hover);
 	}
 
-	/** separated from hover selectors because Safari <15.4 is having trouble parsing these */
-	:host(:focus-visible),
-	:host([first]:focus-visible) {
+	:host(.focus-visible),
```
dlockhart, line 26: So could we not just remove menu's use of the `FocusVisiblePolyfill` mixin now and then also remove the `.focus-visible` class target?

dbatiste: The `FocusVisiblePolyfill` isn't actually even used by menus, but I'd included the `.focus-visible` class for the case where focus is set programatically. But it turns out that somehow `:focus-visible` is working without that... I'm not entirely sure how. 🤯   It seems like it's safe to remove it.

I did a quick test that switches button to use `:focus-visible` for a dialog opener, and `:focus-visible` seems to work in that programatic case as well, at least in Chrome.

dbatiste: Well, the visual-diffs at least benefit from the `.focus-visible` class. Maybe it is worth keeping.

components/link/link.js
```diff
@@ -1,8 +1,11 @@
 import '../colors/colors.js';
-import { css, html, LitElement } from 'lit';
+import { css, html, LitElement, unsafeCSS } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { FocusMixin } from '../../mixins/focus-mixin.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { isFocusVisibleSupported } from '../../helpers/focus.js';
+
+const focusPseudoClass = isFocusVisibleSupported() ? 'focus-visible' : 'focus';
```
eKoopmans, line 8: Rather than duplicating the `'focus-visible' : 'focus'` logic in multiple files, what are your thoughts on something like:

```js
import { getFocusPseudoClass } from '../../helpers/focus.js';

// This could probably be embedded directly wherever it's used, but might be
// a small perf benefit to do it once at the top instead of e.g. in a render function...
const focusPseudoClass = getFocusPseudoClass();
```

components/link/link.scss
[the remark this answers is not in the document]
dbatiste: Yeah, this would be added in BSI. I have the change ready, just didn't put up a PR since I wasn't sure how palatable this change would be. It would import the helper and just do:
```javascript
if (!isFocusVisibleSupported()) {
	document.documentElement.classList.add('d2l-focus-visible-not-supported');
}

```

components/link/link.js
dbatiste: Yep. That's valid. I can't think of any reason why we'd need to do it in each place that needs it.

components/menu/menu-item-styles.js
dlockhart: Hmm probably because they're calling `forceFocusVisible` which sets that class? I bet if they switched to just programatically call `.focus()` that could work.

dbatiste: The tests are currently using standard `focus()`. I think they were new tests I'd added for my wcag 2.2 changes, and they effectively relied on `:focus`. I think there is some value in keeping them, but not sure we want to include `.focus-visible` to make them work.

components/link/link.js
dbatiste: I've updated with your suggestion. Verified that the `isFocusVisibleSupported` only gets called once per page, and `getFocusPseudoClass` only gets called once per module, no matter how many instances of the component there are.

dlockhart (approved):
LGTM!

PR #3219, opened 2023-01-31
fix: missing reflect keyword for button-disabled

stevenao:


oelhanafey:
I don't think this is necessary, is it? We would only need `reflect` if the `button-disabled` property was being changed in the JS and we wanted that reflected back to the attribute.

stevenao:
We are actually changing it in JS. 

https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/28e53a821165a42768e6e3b537b1d9b5c2b2fcc2/components/activity/editor/collection/custom/quiz/d2l-activity-collection-item-quiz.js#L306

oelhanafey:
@stevenao Could you just use `setAttribute` instead?

stevenao:
sure.. i will close this PR.

oelhanafey:
Hey @stevenao, sorry I was mistaken about this. We do want this change since we should reflect properties that are used in [CSS attribute selectors](https://daylight.d2l.dev/developing/best-practices/attributes-properties/#do%3A-reflect-properties-used-in-css-attribute-selectors) in order for changes to reflect properly.

oelhanafey (approved):

PR #3220, opened 2023-01-31
feat: add selection-disabled property

GZolla:
Rally: [US147774](https://rally1.rallydev.com/#/?detail=/userstory/681832541101&fdp=true): list-item-button > Rename disabled property to selection-disabled


components/list/list-item-checkbox-mixin.js
```diff
@@ -12,8 +12,10 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 			/**
 			 * **Selection:** Disables the input
 			 * @type {boolean}
+			 * @ignore
 			 */
-			disabled: { type: Boolean },
+			disabled: { type: Boolean }, // deprecated
+			selectionDisabled: { type: Boolean, attribute: 'selection-disabled' },
```
oelhanafey, line 18: The Daylight documentation pulls the component properties, slots and events directly from the component itself. That's why we have marked up the documentation for the other properties and why the `@ignore` was necessary on the deprecated component so it is no longer shown on the docs. We should add similar markup for this new property.

components/list/list-item-checkbox-mixin.js
```diff
@@ -109,7 +112,7 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 
 	_onCheckboxActionClick(event) {
 		event.preventDefault();
-		if (this.disabled) return;
+		if (this.disabled || this.selectionDisabled) return;
```
dbatiste, line 115: Maybe we can leverage the `willUpdate` life cycle callback to set `selectionDisabled` if `disabled` changes?  This would enable us to clean up a lot of the code now, so that the future PR to remove `disabled` is simpler. It assumes that the consumer would only either be using `disabled` or `selection-disabled`, which I think is fine.
```javascript
	if (changedProperties.has('disabled')) this.selectionDisabled = this.disabled;
}
```
Note however, that in order for this to work properly, `this.disabled = false;` needs to be removed from the `constructor`, so that `selection-disabled` only gets calculated if the consumer has explicitly set `disabled`.

components/list/test/list.visual-diff.html
```diff
@@ -238,7 +238,7 @@
 			<div class="visual-diff" id="selectable">
 				<d2l-list>
 					<d2l-list-item label="Item 1" selectable key="1">Item 1</d2l-list-item>
-					<d2l-list-item label="Item 2" disabled selectable key="2">Item 2</d2l-list-item>
+					<d2l-list-item label="Item 2" selection-disabled selectable key="2">Item 2</d2l-list-item>
```
oelhanafey, line 241: It would also be a good idea to add unit tests for `selection-disabled` in [list-item-checkbox-mixin.test.js](https://github.com/BrightspaceUI/core/blob/main/components/list/test/list-item-checkbox-mixin.test.js)

components/list/list-item-mixin.js
```diff
@@ -343,9 +343,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				transition: border-color 400ms linear;
 			}
 			:host([_highlighting]) [slot="outside-control-container"],
-			:host(:not([disabled]):not([skeleton])[_hovering][_highlighting]) [slot="outside-control-container"],
-			:host(:not([disabled]):not([skeleton])[_focusing][_highlighting]) [slot="outside-control-container"],
-			:host(:not([disabled]):not([skeleton])[selected][_highlighting]) [slot="outside-control-container"] {
+			:host(:not([disabled]):not([selection-disabled]):not([skeleton])[_hovering][_highlighting]) [slot="outside-control-container"],
+			:host(:not([disabled]):not([selection-disabled]):not([skeleton])[_focusing][_highlighting]) [slot="outside-control-container"],
+			:host(:not([disabled]):not([selection-disabled]):not([skeleton])[selected][_highlighting]) [slot="outside-control-container"] {
```
oelhanafey, line 348: The disabled property is also used for some styling for [list-item-link-mixin.js](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-link-mixin.js) and [list-item-button-mixin.js](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-button-mixin.js).

components/list/list-item-checkbox-mixin.js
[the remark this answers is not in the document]
oelhanafey: I made this mistake yesterday. For reference: https://daylight.d2l.dev/developing/best-practices/attributes-properties/#do%3A-reflect-properties-used-in-css-attribute-selectors

components/list/list-item-mixin.js
GZolla: > The disabled property is also used for some styling for [list-item-link-mixin.js](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-link-mixin.js) and [list-item-button-mixin.js](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-button-mixin.js).

For links, the disabled property seems to refer to different things than `selection-disabled` so I left them there just in case consumers still exist. It disables glow/ring when the primary action is disabled(I'm not sure if disabling links should exist, just remove the href-property).
For buttons, I think it will be a separate PR that also tackles [this thread](https://github.com/BrightspaceUI/core/pull/3194)

dbatiste: > I'm not sure if disabling links should exist, just remove the href-property

Disabled links is not a thing. To "disable" a link, the consumer simply does not specify an href/action-href, and no link is rendered.

oelhanafey: The list-item-links and list-item-buttons inherit from the `ListItemMixin`, so your changes to the `disabled` property will also be reflected in places where those mixins are used.

components/list/test/list.visual-diff.html
GZolla: Should I leave the checks for disabled even though it is a deprecated property?

oelhanafey: I think that would be a good idea, so we can ensure that the property is still working as intended. Those tests can be removed when the property is being completely removed after the existing uses are switched over.

GZolla: Ok, I added a test to check that `selection-disabled` is added even when only `disabled` is set. I think is sufficient and cleaner, hence easier to remove when the property is removed.

components/list/README.md
```diff
@@ -414,6 +413,7 @@ The `d2l-list-item` provides the appropriate `listitem` semantics for children w
 | `key` | String | Value to identify item if selectable or draggable |
 | `label` | String | Explicitly defined label for the element |
 | `labelled-by` | String | The id of element that provides the label for this element |
+| `selection-disabled` | Boolean | Disables selection |
```
oelhanafey, line 416: Nit: This property should be placed alphabetically

oelhanafey, line 416: Because the `ListItemButtonMixin` extends `ListItemMixin`, `selection-disabled` is also added to the `list-item-button` so we should update that part of the README as well

components/list/list-item-checkbox-mixin.js
```diff
@@ -55,7 +62,7 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 
 	constructor() {
 		super();
-		this.disabled = false;
+		this.selectionDisabled = false;
```
oelhanafey, line 65: Nit: This should also be placed alphabetically with respect to the other properties

oelhanafey (approved):
Looks good!

PR #3222, opened 2023-01-31
fix: github-script usage after upgrade to v6

eKoopmans:
This PR should fix the PR previews! They've started failing because `third-party-actions@actions/github-script` was upgraded yesterday to v6 (from v3), and the Octokit API moved a lot of commands e.g. `github.issues` to `github.rest.issues`.


dlockhart (approved):
🎉

PR #3224, opened 2023-02-01
feat: Rename selection headers to controls

eKoopmans:
This PR begins the process of renaming `d2l-selection-header`, `d2l-list-header`, and `d2l-table-header` to `d2l-*-controls`. This is primarily being done because for tables, the "header" language is confusing.

There are three pieces left in that will need to be cleaned up:

1. `list-header.js`, which now just duplicates `list-controls.js` to create `<d2l-list-header>`
2. `--d2l-list-header-background-color`
3. `slot="header"` on `<d2l-list>`

The next steps will be:

1. Update imports on documentation for Daylight
2. Switch all consumers of `d2l-list-header` to `d2l-list-controls` (fortunately, `selection-header` and `table-header` aren't used anywhere externally yet)
3. Remove cleanup code above

Rally: https://rally1.rallydev.com/#/?detail=/userstory/673800144323&fdp=true


dlockhart (approved):
Nice, this looks good. One option for the visual diffs that I've done in the past is to actually rename the goldens locally to their new names and push that change as part of the commit. That way visual-diff actually passes right away and you don't have files being deleted and re-added, preserving the Git history if that's important.

eKoopmans:
@dlockhart Yeah I should have done that - the visual-diff PR showed mostly "moved" files, but two showed up as delete + new. I spent a while comparing them to make sure there was no change.

Do you think I should go back and just rename the files, and force-push over this? I'm happy to if you think that would be better.

dlockhart:
I don't feel strongly about it. If the diffs were older and had a lot of history that might be important then it might be worth it, but that's not the case so "meh".

PR #3227, opened 2023-02-01
fix: undefined errors in tabs component

oelhanafey:
This fixes two issues in the tabs component that were causing crashes in the d2l-my-courses-ui repo's tests but could potentially cause issues elsewhere as well.

The first change in `_calculateScrollPosition()` checks if the selected tab exists. It's possible, due to timing issues, for the selected tab to be removed in between when the tab is selected and the tab measures are updated. This would probably happen very rarely but in these situations, the component's translation value is reset to the default position.

The second change in `handlePanelsSlotChange()` checks that there is a selected tab. This was previously crashing because if all of the tabs were removed from the component, it would pass in a null `selectedTabInfo` to `_updateScrollPosition()`


dbatiste (approved):
Looks good! Thanks for digging into that!

PR #3234, opened 2023-02-03
Update d2l-list-item-content to enable overflow ellipsis.

dbatiste:
[DE51906](https://rally1.rallydev.com/#/?detail=/defect/684848898275&fdp=true)

This fixes a regression from [this wcag 2.2 change](https://github.com/BrightspaceUI/core/commit/82fc48fcd70615fee566f5c09783925cfc311dbc#diff-bccc585941c4e778feb72e6c2af5e7becca1e157f48db84de916a0062bdea883). 

This PR updates `d2l-list-item-content` to enable consumers to truncate the primary text with ellipsis when required. Previous to the wcag 2.2 change, overflow: hidden was preventing the contents from overflowing. It was removed to allow for the new focus indicator outline, however this caused the contents of the flex container to potentially overflow.

**Overflowing (regression):**
image

**Fixed (wrapping):**
image

**Fixed (overflow ellipsis):**
image

**Fixed (short text):**
`inline-grid` required so that the outlines doesn't go 100% width.
image


components/list/list-item-content.js
```diff
@@ -19,7 +19,7 @@ class ListItemContent extends LitElement {
 			.d2l-list-item-content-text > div {
 				border-radius: var(--d2l-list-item-content-text-border-radius);
 				color: var(--d2l-list-item-content-text-color);
-				display: inline-block;
+				display: inline-grid;
```
dbatiste, line 22: I found some old StackOverflow posts suggesting that `min-width: 0;` is a for this issue of elements overflowing their containers, but it didn't work for me. The `grid` and `inline-grid` where the only things I found that worked.

bearfriend: Applied this to the Work to Do widget and it's truncating it, but still allowing `d2l-list-item-content` to take up 100% width unless I apply `inline-grid` to _it_ as well.

`inline-grid` on `d2l-list-item-content > div`
Screen Shot 2023-02-03 at 3 25 04 PM

`inline-grid` on `d2l-list-item-content, d2l-list-item-content > div`
Screen Shot 2023-02-03 at 3 25 18 PM

dbatiste: Interesting. Note that it is the inner-most `div` (`.d2l-list-item-content-text > div`) that gets this, not the first `div`. This is what the Quick Eval widget looks like with the fix. I haven't tested Work To Do widget, but I can.
image

bearfriend: Yeah sorry, typed that out wrong. Edited.

dbatiste: Wonder if we are doing something different, or the case I am looking at is slightly different. This is what I get:
image

bearfriend: Ah, it's the secondary text that's keeping it long in my first screenshot above. Applying `inline-flex` to the `d2l-list-item-content-text-secondary` _and_ w2d's own `d2l-w2d-attribute-list` fixes that but that seems not ideal.

dbatiste: Ok, I've made a couple more tweaks. I think it should be working for that case now too. I've added a bunch more visual-diff tests to capture this case as well as a couple others.

bearfriend (approved):

dlockhart (approved):

PR #3240, opened 2023-02-06
chore: Cleanup formatting in list readme and selection readme/visual-diff

eKoopmans:
This PR is just cleanup - it should have no functional effect on the docs or visual-diffs. I've made three changes:

1. **list readme:** Switch from `style="width: 100%"` to `display: block`
    - the 100% width was only required because the demo blocks use `display: flex` by default
    - having that inline style in there could have been confusing to consumers
2. **selection readme:** Switches tabs to double-spaces
3. **selection visual-diff:** Fixes some incorrectly closed html tags


dlockhart (approved):
Nice, thanks for the cleanup!

PR #3241, opened 2023-02-06
chore: Add basic vdiffs for selection-controls

eKoopmans:
Setting up some basic visual-diffs for `d2l-selection-controls`. There's already a lot of coverage on `d2l-list-controls`, so I don't want to duplicate too much of that here - maybe in the future it would make sense to consolidate the `d2l-list-controls` tests here.


dlockhart (approved):
LGTM now that the import is fixed :)

dlockhart (approved):
Yep that's even better -- less chance or weird flake!

PR #3243, opened 2023-02-06
fix: Sticky behaviour in selection header demos

eKoopmans:
On our `list`, `selection`, and `table` demos, the sticky behaviour of the "controls" component wasn't being properly demonstrated. It was the worst of both worlds: the controls don't stick to the top of the page, but it still gets styled as if it's stuck, adding a shadow underline that looks out of place.

This PR fixes those by using `no-sticky` in most demos, and making the demo container scrollable in the demo of the controls itself.

I've also included two other fixes for the table demo specifically:

1. Fixed the logic of the `<d2l-sample-table-with-controls>` itself, which wasn't handling selection correctly
2. Added a `z-index` to the `d2l-table-controls`, to make it float properly over the table
    - This is sort of a placeholder, I have a whole other story to improve `d2l-table-controls` sticky behaviour, which has a complicated interaction with the table's own sticky header

### Broken

d2l-list-header sticky broken

### Fixed

d2l-list-header sticky fixed

### DIFF
--- a/components/table/table-controls.js
+++ b/components/table/table-controls.js
@@ -1,6 +1,6 @@
 import '../selection/selection-select-all-pages.js';
 import '../selection/selection-summary.js';
-import { html, nothing } from 'lit';
+import { css, html, nothing } from 'lit';
 import { SelectionControls } from '../selection/selection-controls.js';
 
 /**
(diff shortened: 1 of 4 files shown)


components/table/table-controls.js
```diff
@@ -17,6 +17,18 @@ class TableControls extends SelectionControls {
...
+		return [super.styles, css`
+			:host {
+				--d2l-selection-controls-background-color: var(--d2l-table-controls-background-color);
+				z-index: 2; /* must be greater than d2l-table rows */
+			}
+			:host([no-sticky]) {
+				z-index: auto;
+			}
+		`];
+	}
```
eKoopmans, line 30: I called this out in the description, but this code is "sort of" a placeholder - it's safe to merge it as is and will improve the demos, or we can wait for #3283 to merge first. Note that in #3283 I ended up using `z-index: 4` instead of `z-index: 2`.

dbatiste (approved):

dlockhart (approved):

PR #3254, opened 2023-02-09
feat: add list-item key to d2l-list-item-expand-collapse-toggled event

mdulat:
so that we know which item was expanded/collapsed


dbatiste:
> so that we know which item was expanded/collapsed

I thought the key would be accessible via `e.target.key`?

mdulat:
> I thought the key would be accessible via `e.target.key`?

Oh you're right. I didn't see that, when I printed out `e` it was listing `target` as a top-level component (i.e. d2l-activity-discussion-topic-editor) instead of the d2l-list-item. But looks like `e.target.key` does have the key. Thanks.

dbatiste:
The `currentTarget` is another option. That might give you the actual list-item.

PR #3258, opened 2023-02-09
Add introductory-text to filter-dimension-set

oelhanafey:
[US148379](https://rally1.rallydev.com/#/?detail=/userstory/685199365675&fdp=true)

[Style specs](https://www.figma.com/proto/4sRGPOeVbaXHb77zh17KCv/Filter---Enhancements-for-Wave?page-id=0%3A1&node-id=714%3A1653&starting-point-node-id=2%3A48&scaling=min-zoom)

Adds the introductory-text attribute to the filter-dimension-set component. There are different margins for the single and multi-dimensional cases. Did not add ability to listen to data change events for the property to keep the filter component as simple as possible, but it can easily be added if needed.


components/filter/filter.js
```diff
@@ -322,6 +336,14 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 
 		const dimension = this._getActiveDimension();
 
+		const introductoryTextClasses = {
+			'd2l-body-compact': true,
+			'd2l-filter-dimension-intro-text': true,
+			'd2l-filter-dimension-intro-text-multi-dimension': !singleDimension
```
svanherk, line 342: As discussed in the team channel, sounds like we could change this to `multi-dimension` and the one below to `with-intro` and get things a little shorter and cleaner 🎉

svanherk:
Looks great! On small-but-not-mobile screens, the header takes up a lot of the old scroll space (you can see about two less list items now), but that's more of a design remark.  And maybe not a common use case of Wave anyways
image

PR #3260, opened 2023-02-09
Revert overflow changes from d2l-html-block since they can cause regressions.

dbatiste:
This PR reverts the `overflow` changes to `d2l-html-block` from https://github.com/BrightspaceUI/core/commit/1e224f58f0d91a2b03a045abb03630a4ef6ee6fd. We discovered the regression in one of our custom homepage widgets, and it's possible that clients could have similar cases in a variety of contexts. We'll need to get this in before tomorrow.

@JordanRS-D2L : FYI. I think we will need a similar usage of `ignoredRules` (see changes to our test here) in the UCE code to address the failure that you were encountering. Are there other changes that will be needed there?

We may find a common place to share rules to be ignored. This appears to be the only way (at least that I've found so far) to ignore rules using the open-wc axe test helpers. Aside from `ignoredRules`, we have the option of specifying `ignoredTags` in that object passed to `accessible(...)`.


components/tooltip/README.md
```diff
@@ -36,7 +36,7 @@ Note: there is a [known defect](https://rally1.rallydev.com/#/?detail=/defect/64
 
 **Interactive Target Elements:**
 
-If the tooltip's target is an interactive element then it will automatically be accessibile. [A list of interactive elements can be found here.](https://github.com/BrightspaceUI/core/blob/main/components/tooltip/tooltip.js#L38)
```
dlockhart, line 39: 🤮 haha

dlockhart (approved):
Looks good. UCE should just need to add a similar `ignoredRules` entry.

PR #3263, opened 2023-02-09
Change filter component to use empty-state-simple

oelhanafey:
[US148589](https://rally1.rallydev.com/#/?detail=/userstory/685338409787&fdp=true)

[Style specs](https://www.figma.com/proto/4sRGPOeVbaXHb77zh17KCv/Filter---Enhancements-for-Wave?page-id=0%3A1&node-id=714%3A1653&starting-point-node-id=2%3A48&scaling=min-zoom)

This switches the filter component to use the `<d2l-empty-state-simple>` component for its empty states instead of creating the empty states itself. 

In the design specs, the border between the header and the empty state is removed. That border is added by dropdown-content, not the filter component so in order to align with the spec we would need to add a new property to the dropdown-content component. Jeff said that it was fine either way, with or without the border, so I think we can merge it with the border for now and I will follow-up on whether we want to add that property to dropdown-content as a separate PR, if needed.


components/filter/filter.js
```diff
@@ -389,9 +390,10 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 
 		if (this._isDimensionEmpty(dimension)) {
 			return html`
-				<p class="d2l-filter-dimension-info-message d2l-body-small" role="alert">
```
svanherk, line 392: Before we had `role="alert"` here so that after the user opened an empty dimension it would immediately tell them it was empty, or below after a search finished loading it would say "5 search results" or whatever without having to navigate.  Does `d2l-empty-state-simple` have any announcing built in?  I'm not sure if adding `role="alert"` onto the component would also "just work", or if maybe wrapping it in a div with that would have the same results 🤔

oelhanafey: There's nothing built in, no. I will do some testing on this to see what is possible, good observation!

oelhanafey: Adding `role="alert"` directly to the empty state component does work but it was previously being placed within an `aria-live` region and was sometimes being read out twice, at least on VO. I fixed that and moved the `aria-live` region down to the inner list layer which fixes that.

But it led me to finding more screenreader issues. I'm getting duplicate screenreader calls when it reads all items in the list
e.g. `Fall Fall Winter Winter Spring Spring`

Also, when you are in a multi-dimensional filter and drilled into a dimension it lists the number of items in the list as the number of dimension-sets in the menu rather than the dimension-set-values in the list.
e.g. 'Menu 3 items' but there are 4 items in the list. This one might not have an easy solution since we are using a list within a menu.

Can you repro any of those on NVDA? I'll spin up a defect for these.

svanherk: You're seeing these defects on `main`, or after your changes? On `main`, I'm seeing:

> But it led me to finding more screenreader issues. I'm getting duplicate screenreader calls when it reads all items in the list
> e.g. Fall Fall Winter Winter Spring Spring
> 

So most of the time, I open it and get "Application. Art checkbox not checked." But you're right that sometimes, it starts reading the list items instead and does say "Art Art (Twelve). Astronomy Astronomy (two)." It's hard to get it to do it again though lol, is VoiceOver more consistent?

> Also, when you are in a multi-dimensional filter and drilled into a dimension it lists the number of items in the list as the number of dimension-sets in the menu rather than the dimension-set-values in the list.
> e.g. 'Menu 3 items' but there are 4 items in the list. This one might not have an easy solution since we are using a list within a menu.

Definitely not seeing that with NVDA. NVDA will properly say "Course, sub-menu 1 of 3. 1 role applied." and then when you drill in it doesn't talk about the menu again. It doesn't talk about counts at all, because our list items don't have that in `grid` mode (since they are no longer using the list roles). So seems like VO is talking about the parent when it shouldn't be 😟

svanherk: > Adding role="alert" directly to the empty state component does work but it was previously being placed within an aria-live region and was sometimes being read out twice, at least on VO. I fixed that and moved the aria-live region down to the inner list layer which fixes that.

I wonder if that's causing the duplicate list item reading as well. It _seems_ like I needed to add that as part of the empty state work: https://github.com/BrightspaceUI/core/pull/1432/files.  So if we don't actually need it to announce empty state for VO and NVDA, we could try removing it

oelhanafey: I see them on both `main` and this branch. 

I can consistently get the double announcing. If I open it, it will say "Application. Art checkbox not checked" and then start reading them afterwards. Also if you search something and then clear it, it will start reading all of the items again and they'll be duplicated.


> So seems like VO is talking about the parent when it shouldn't be 😟

Yeah.. it's not good :( And it's a bit inconsistent but sometimes I'm able to navigate the menu while i'm drilled into a dimension set.

oelhanafey: > So if we don't actually need it to announce empty state for VO and NVDA, we could try removing it

Yeah, using `role="alert"` creates an assertive `aria-live` region so you don't need to set one yourself. Removing it would also fix the double announcing defect

svanherk: > If I open it, it will say "Application. Art checkbox not checked" and then start reading them afterwards

Interesting, NVDA just stops reading them until you arrow to the next one, and then just reads that one.

> Yeah.. it's not good :( And it's a bit inconsistent but sometimes I'm able to navigate the menu while I'm drilled into a dimension set.

This might be a `hierarchical-view` issue. It sounds a bit like this: https://github.com/BrightspaceUI/core/pull/1272.  But worse, if it's an only VO issue 😟

oelhanafey: Thanks for helping me to investigate. I created a defect for this: [DE52188](https://rally1.rallydev.com/#/?detail=/defect/686309407849&fdp=true).

components/filter/filter.js
```diff
@@ -142,8 +143,11 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				color: var(--d2l-color-chromite);
 			}
 
+			.d2l-empty-state-container {
+				padding: 0.9rem;
```
oelhanafey, line 147: Needed to use padding in a container rather than margin to avoid margin collapse

svanherk:
Looking at the demo page, I'm seeing the empty state text turn dark blue on hover. It only happens for the multi-dimension case, so I think it's a `hierarchical-view` style bleeding through.  Haven't found the exact source yet

oelhanafey:
> Looking at the demo page, I'm seeing the empty state text turn dark blue on hover. It only happens for the multi-dimension case, so I think it's a hierarchical-view style bleeding through. Haven't found the exact source yet


Thanks, it's coming from the `d2l-menu-item` hover style.

oelhanafey:
So the empty state was previously using `d2l-body-small` which has an explicit colour style. I switched it to use `d2l-body-compact` which does not have a colour style. That's why the styles weren't bleeding before.

components/filter/filter.js
svanherk: So currently, NVDA + Chrome works fine, but Firefox just says nothing 😞 now. If I move the `role="alert"` to the container `d2l-empty-state-container` `div`s instead, it works in both.  Does that work in VoiceOver?

oelhanafey: Ah, it works in VO. Moving the role up to the container doesn't appear to break anything in VO so it should be fine to move that role up to the div container

components/filter/filter.js
```diff
@@ -443,6 +453,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		return html`
 			${searchResults}
 			<d2l-list
+				aria-live="polite"
```
svanherk, line 456: Did we still want this, or did this cause the double reading bug in VoiceOver?

svanherk (approved):

components/filter/filter.js
oelhanafey: Nope, I was removing that, sorry for the messy logs. Rebased instead of merging and I ended up messing up the branch 🤦

PR #3274, opened 2023-02-13
fix: menu-item-selectable-mixin aria-checked

johngwilkinson:
I have just run npm update on a repo and am now getting a failing axe test:
> Required ARIA attribute not present: aria-checked
on default unselected `<d2l-menu-item-radio>`s.

Adding aria-checked in my `render()` doesn't seem to work (it vanishes), and the Daylight page doesn't mention it, so I suppose it is managed by the component. I see a change handler here that sets the property here. Nothing in our code seems to have changed recently, but the MDN docs seem to suggest that the property should be present with a value of "false" for unselected items, whereas our code removes it in that case - so I think it's just that the axe lib just got more picky.

- [x] try axe tests with this change
- [x] check behaviour with this change

### DIFF
--- a/components/menu/test/menu-item-selectable-mixin.test.js
+++ b/components/menu/test/menu-item-selectable-mixin.test.js
@@ -10,7 +10,7 @@ describe('MenuItemSelectableMixin', () => {
 
 	describe('accessibility', () => {
 
-		it('has no "aria-checked" false by default', async() => {
+		it('has "aria-checked" false by default', async() => {
 			const elem = await fixture(`<${tag}></${tag}>`);
 			expect(elem.getAttribute('aria-checked')).to.equal('false');
 		});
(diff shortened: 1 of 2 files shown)


dbatiste (commented):
Looks good John. This must have been an oversight when this mixin was originally created. Looks like there's just a couple tests that need to be fixed up. 

I did a quick check and our `d2l-selection-input` (used in lists and tables) appears to render `aria-checked="true|false"`, consistent with your change.

components/menu/test/menu-item-selectable-mixin.test.js
```diff
@@ -10,12 +10,12 @@ describe('MenuItemSelectableMixin', () => {
 
 	describe('accessibility', () => {
 
-		it('has no "aria-checked by default', async() => {
+		it('has no "aria-checked" false by default', async() => {
```
eKoopmans, line 13: I'd probably remove the "no" here: `'has "aria-checked" false by default'`.

eKoopmans (approved):

PR #3276, opened 2023-02-13
fix: force visual-diff update

dlockhart:
Looks like Puppeteer updated and caused a lot of visual diffs. This is an empty commit so we can deal with those separately.


dlockhart:
Closing in favour of #3269.

PR #3280, opened 2023-02-13
re-apply sass link focus indicator

dlockhart:
These are the new WCAG 2.2 focus styles for links created using Sass, [which Dave temporarily reverted](https://github.com/BrightspaceUI/core/pull/3265) last week because they were causing weird outlines in the LMS that we didn't want in the March release.

In addition to re-applying them, this switches the border-radius from `3px` to `2px` as per a discussion with Jeff.


components/link/link.scss
```diff
@@ -9,13 +9,23 @@
...
-	&:hover,
-	&:focus,
-	&.d2l-link-focus {
+	&:hover {
 		color: $d2l-color-celestine-minus-1;
 		text-decoration: underline;
-		outline-width: 0;
 	}
+	&:focus-visible {
+		border-radius: 2px;
```
dlockhart, line 17: I put the border-radius down here vs. Dave's original change had it at the root. This matches what the web component is doing, and isn't applying a radius when focus isn't present, which I felt was a bit cleaner.

eKoopmans (approved):
LGTM 👍 - though FYI, the link to #3265 in the PR description is broken...

dbatiste (approved):

PR #3283, opened 2023-02-13
feat: Add sticky support for table-controls

eKoopmans:
This PR adds sticky support for `d2l-table-controls`. Specifically, logic has been added so that the table's `sticky-headers` will always position themselves below the controls when the controls are sticky.

This logic already existed (for when there are multiple sticky header-rows), but it was broken when there were more than 2 header-rows, so I've also fixed that behaviour. That logic is all separated now in the `_updateStickyTops` method.

Keeping this as a draft while I add some visual-diffs for the sticky behaviour.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/656750407555&fdp=true


components/table/demo/table.html
```diff
@@ -16,28 +16,28 @@
 			<h2>Default</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-test-table type="default"></d2l-test-table>
+					<d2l-test-table type="default" sticky-controls="true"></d2l-test-table>
```
dlockhart, line 19: These shouldn't need a value since they're boolean attributes?

```suggestion
					<d2l-test-table type="default" sticky-controls></d2l-test-table>
```

components/table/table-wrapper.js
```diff
@@ -359,19 +361,54 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
...
+	}
+
+	async _handleTableChange() {
+		await new Promise(resolve => requestAnimationFrame(resolve));
+
+		this._applyClassNames();
+		this._updateStickyTops();
+	}
 
+	_registerMutationObserver(observerName, callback, target, options) {
```
dlockhart, line 384: Nice refactor. 👍

components/table/table-wrapper.js
```diff
@@ -359,19 +361,54 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
...
+			this[observerName] = new MutationObserver(callback);
+		}
+		if (target) this[observerName].observe(target, options);
 	}
+
+	_updateStickyTops() {
+		if (!this._table || !this.stickyHeaders) return;
+
+		const controlsHeight = this._controls && !this._controls.noSticky ? this._controls.offsetHeight + 4 : 0; // +4 for the shadow's `bottom: -4px`.
+		const borderRadius = this.type === 'default' ? 5 : 0;
```
dlockhart, line 397: Wonder if it's worth defining this as a `const` above and then using it here and in the CSS? Weirdly it looks like there's a mix of `5px` and `0.3rem` (`6px` @ `20px` font size)...

components/table/table-wrapper.js
```diff
@@ -359,19 +361,54 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
...
+		let rowTop = controlsHeight - borderRadius;
+
+		const stickyRows = Array.from(this._table.querySelectorAll('tr.d2l-table-header, tr[header], thead tr'));
+		stickyRows.forEach(r => {
+			const ths = Array.from(r.querySelectorAll('th'));
+			ths.forEach((th) => {
+				th.style.top = `${rowTop}px`;
```
dlockhart, line 404: Would expressing this using the existing radius CSS variable work?

```suggestion
				th.style.top = `calc(${rowTop} - var(--d2l-table-border-radius))`;
```

dlockhart (approved):
This is looking good to me, lots of great refactoring that puts this component in a better place despite the additional complexity!

components/table/demo/table.html
eKoopmans: Good catch, will fix!

components/table/table-wrapper.js
eKoopmans: Definitely an option - this would mean removing `const borderRadius`. Not sure if there's any performance or readability benefit of doing it this way vs accounting for it in js, what do you think?

components/table/table-wrapper.js
eKoopmans: Haha true, I'll investigate that inconsistency...

components/table/table-wrapper.js
dlockhart: The big pro to just using `--d2l-table-border-radius` is not needing to duplicate its definition in two places. But yeah, we'd be pushing the calculation from JavaScript down to the browser's CSS engine. I would expect the browser to have clever logic such that it only recalculates when the variable value changes.

components/table/table-wrapper.js
dlockhart: If we need to switch that from `rem`-based to `px`-based in the variable definition, that's likely fine. We've been doing a bit of that for other components lately for the WCAG 2.2 work.

components/table/table-wrapper.js
```diff
@@ -225,12 +220,31 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
...
+				--d2l-table-controls-shadow-display: none;
+			}
+			.d2l-sticky-headers-backdrop {
+				position: sticky;
+				top: calc(var(--d2l-table-sticky-top) + var(--d2l-table-border-radius));
+				width: 100%;
+				z-index: 2; /* Must sit under d2l-table sticky-headers but over sticky columns and regular cells */
+			}
+			.d2l-sticky-headers-backdrop::after {
+				background-color: var(--d2l-table-controls-background-color, white);
```
eKoopmans, line 241: This needs a solid colour - I've used `--d2l-table-controls-background-color` with a fallback to `white`, but that's not bulletproof (e.g. `--d2l-table-controls-background-color` might be defined explicitly in the `d2l-table-controls` scope, which we wouldn't have access to here).

components/table/table-controls.js
```diff
@@ -21,7 +21,8 @@ class TableControls extends SelectionControls {
 		return [super.styles, css`
 			:host {
 				--d2l-selection-controls-background-color: var(--d2l-table-controls-background-color);
-				z-index: 2; /* must be greater than d2l-table rows */
+				--d2l-selection-controls-shadow-display: var(--d2l-table-controls-shadow-display);
+				z-index: 5; /* Must be greater than d2l-table-wrapper and d2l-scroll-wrapper */
```
eKoopmans, line 25: There's a small `z-index` collision here where the scroll-wrapper's controls (left/right scroll buttons) can get hidden behind the table controls. But on the flip side, the controls need to be on top or else things like the overflow dropdown get hidden.

I think the only solution would be to have those scroll-wrapper buttons stick below the controls, instead of the top of the screen...

eKoopmans (commented):
Okay! Lots of updates yesterday, but this is once again ready to review. I've had to change things to match Jeff's designs - the sticky controls now sit higher above the table's sticky header rows, with no shadow:

image

The main challenge here was making sure the non-sticky part of the table isn't visible between the controls and the sticky headers, which I've done with a div called `.d2l-sticky-headers-backdrop`. Also `z-index` was a pain!

There's two small concerns that I intend to leave as-is, I've called them out on this review.

components/table/table-wrapper.js
```diff
@@ -202,7 +195,8 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
 			type: {
 				reflect: true,
 				type: String
-			}
+			},
+			_controlsScrolled: { state: true }
```
dlockhart, line 199: Probably want to default to `false` in the constructor.

components/table/table-wrapper.js
```diff
@@ -225,12 +220,31 @@ export class TableWrapper extends RtlMixin(SelectionMixin(LitElement)) {
 				display: none;
 			}
 			:host([type="light"]) {
+				--d2l-table-border-radius: 0rem; /* stylelint-disable-line length-zero-no-unit */
+				--d2l-table-border-radius-sticky-offset: 0rem; /* stylelint-disable-line length-zero-no-unit */
```
dlockhart, line 224: Did the `calc`s not work with `0`?

dlockhart (approved):
This is looking great -- some super complex problems here with nice solutions. No notes.

components/table/table-wrapper.js
eKoopmans: Haha frustratingly no - `calc()` seems to need units, even for 0.

components/table/table-controls.js
eKoopmans: I found a real simple fix for this actually!

image

Going to add it in a subsequent PR though, don't want to complicate this PR any further.

components/table/table-wrapper.js
eKoopmans: A bit more context on this btw:
https://stackoverflow.com/a/32517653

`calc()` needs all values to be of the same type. 0 is an `integer` type, which can't be added/subtracted with a `length` type (e.g. `6px`).

dlockhart: Interesting! Feels like the browsers still could have been like "I know what you're trying to do here, I'll help you out", but no.

PR #3284, opened 2023-02-14
Remove z-index from d2l-input-textarea.

dbatiste:
This PR removes the `z-index` from `d2l-input-textarea` to avoid some stacking context issues. It looks like they were introduced when the component was initially converted to Lit, but it appears they aren't actually needed.

There's really two contributing factors to this issue:
1. `d2l-input-textarea` creating a stacking context that is higher than other things on the page when we don't want it to
2. `d2l-card` creating a stacking context for nested `d2l-tooltip`s, effectively overriding the effectiveness of it's `z-index`. It would be good to eliminate this as well, but if it's possible, it will be more involved and can be dealt with separately.

**Before:**
image

**After:**
image

### DIFF
--- a/components/inputs/input-textarea.js
+++ b/components/inputs/input-textarea.js
@@ -114,7 +114,6 @@ class InputTextArea extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMi
 				position: absolute;
 				resize: none;
 				top: 0;
-				z-index: 2;
 			}
 			:host([no-border]) textarea.d2l-input {
 				border-color: transparent;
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Every time a `z-index` gets introduced to our code, some kind of loud ⏰ should go off!

dbatiste:
> Every time a `z-index` gets introduced to our code, some kind of loud ⏰ should go off!

Haha yeah... I was so happy to get rid of this... but my happiness was short-lived because I had to introduce a new one for the collapsible panel's sticky header in my next PR. 😢

PR #3285, opened 2023-02-14
Assign d2l-collapsible-panel's sticky header a z-index.

dbatiste:
[DE51932](https://rally1.rallydev.com/#/?detail=/defect/684982721887&fdp=true)

This PR adds a `z-index` to `d2l-collapsible-panel`'s sticky header so that it floats on top of `d2l-card`, or `d2l-list-item` with dropdown or tooltips open. This is similar to what is required for the `d2l-list-controls`.

**Before:**
image

**After:**
image


dbatiste:
> Haha yeah... that _was_ short lived.

Yeah, sorry to disappoint. I think to avoid this we need a rethink of a `d2l-card`, `d2l-list-item`, and maybe one or two others. I'm not 100% certain they can avoid it, but I have a couple ideas on trying.

PR #3298, opened 2023-02-15
Update the role when the parent lists grid attribute changes

MitchelPaulin:
I need to be able to disable `grid` mode when the list has some components in editing state, for example I want to be able to do something like 

```js
?grid=${!this._editing}
```

This wasn't working however, I tracked down the issue to a role that only gets updated on `connectedCallback`, this PR adds a mutation observer so the role is updated whenever the grid attribute is also updated. 

See [this slack message](https://d2l.slack.com/archives/C036FJMD4P7/p1676400509284509) for the motivation behind this change.

### DIFF
--- a/components/list/list-item-role-mixin.js
+++ b/components/list/list-item-role-mixin.js
@@ -43,7 +43,7 @@ export const ListItemRoleMixin = superclass => class extends superclass {
 
 	disconnectedCallback() {
 		super.disconnectedCallback();
-		this.observer.disconnect();
+		if (this.observer) this.observer.disconnect();
 	}
 
 };
--- a/components/list/list-item-role-mixin.js
+++ b/components/list/list-item-role-mixin.js
@@ -24,12 +24,26 @@ export const ListItemRoleMixin = superclass => class extends superclass {
 		const parent = findComposedAncestor(this.parentNode, node => node && node.tagName === 'D2L-LIST');
 		if (!parent) return;
 
+		this.observer = new MutationObserver(() => { this.role = parent.hasAttribute('grid') ? 'rowgroup' : 'listitem'; });
+		this.observer.observe(parent, {
+			attributes: true, /* required for legacy-Edge, otherwise attributeFilter throws a syntax error */
+			attributeFilter: ['grid'],
+			childList: false,
+			subtree: false,
+			characterData: false
+		});
+
 		const separators = parent.getAttribute('separators');
 
 		this.role = parent.hasAttribute('grid') ? 'rowgroup' : 'listitem';
 		this._nested = (parent.slot === 'nested');
 		this._separators = separators || undefined;
 		this._extendSeparators = parent.hasAttribute('extend-separators');
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.observer.disconnect();
 	}
 
 };


devpow112 (approved):

dbatiste:
Interesting - when does `this._editing` get set to true, and once the user is done editing, what triggers it to be set back so that normal grid navigation resumes? Does the consumer manage all this?

Ideally we avoid changing the roles back and forth for the list and all the items.

The W3C provides some guidance about the interaction pattern for this, and we'd like to implement it in a consistent way. It's what motivated introducing `InteractiveMixin` and I suspect that we need to expand on it. [Editing and Navigating Inside a Cell](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#gridNav_inside). They have a few different examples, but this one shows the editable interaction pattern in a grid: [Example](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/dataGrids/#ex2_label)

As for `d2l-list-item-generic-layout`'s handling of `[Space]/[Enter]` - I see no mention of it in the general W3C guidance that would provide extra insight into why we added that, so I'm not convinced it is needed. Unfortunately Zina is gone so we can't ask her.  I can only assume she had a good reason for it, but it doesn't seem to serve much purpose. Maybe if it's removed we will discover it with some testing.

MitchelPaulin:
> Interesting - when does `this._editing` get set to true, and once the user is done editing, what triggers it to be set back so that normal grid navigation resumes? Does the consumer manage all this?

All done by the consumer, you can see in the PR that is linked exactly how I am using it 

> Ideally we avoid changing the roles back and forth for the list and all the items.
> 
> The W3C provides some guidance about the interaction pattern for this, and we'd like to implement it in a consistent way. It's what motivated introducing `InteractiveMixin` and I suspect that we need to expand on it. [Editing and Navigating Inside a Cell](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#gridNav_inside). They have a few different examples, but this one shows the editable interaction pattern in a grid: [Example](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/dataGrids/#ex2_label)

Even in interactive mode unfortunately the keyboard events were getting eaten, there may be a way to expand it to disable allow those event through and disable the keyboard navigation though using the mixin, but this runs into another issue where the user needs to "know" they have to enter interactive mode, which for someone who doesn't use keyboard navigation normally could be jarring. I may just be misunderstanding the user of InteractiveMixin here though 

> As for `d2l-list-item-generic-layout`'s handling of `[Space]/[Enter]` - I see no mention of it in the general W3C guidance that would provide extra insight into why we added that, so I'm not convinced it is needed. Unfortunately Zina is gone so we can't ask her. I can only assume she had a good reason for it, but it doesn't seem to serve much purpose. Maybe if it's removed we will discover it with some testing.

It wouldn't be just SPACE. HOME, END, LEFT and RIGHT can also be used to edit text to move the caret around, and those navigation inputs would not work with grid mode either

dbatiste:
> Even in interactive mode unfortunately the keyboard events were getting eaten, there may be a way to expand it to disable all keyboard navigation when in interactive mode, 

Thats why I was thinking that we'd have to expand on the `InteractiveMixin`, since it was originally implemented with only one use-case in mind, knowing that other use-cases might come up.

> but this runs into another issue where the user needs to "know" they have to enter interactive mode, which for someone who doesn't use keyboard navigation normally could be jarring.

That's the suggested pattern from the W3C on how to handle this case. We can certainly get designers involved again to discuss the pattern though.

PR #3299, opened 2023-02-15
Fix title overflow in d2l-collapsible-panel.

dbatiste:
[DE52098](https://rally1.rallydev.com/#/?detail=/defect/685751354391&fdp=true)

This PR fixes the overflow behaviour of the `d2l-collapsible-panel` title so that it will not push the actions and opener icon out of the container. It configures the title to wrap anywhere (as per [overflow-wrap: anywhere](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)), and truncate the title with ellipsis after 2 lines as per Design request. It also forces the `background-color` to white as required.

**Before:**
image

**After:**
image


components/collapsible-panel/collapsible-panel.js
```diff
@@ -97,6 +97,7 @@ class CollapsiblePanel extends RtlMixin(LitElement) {
 				--d2l-collapsible-panel-spacing-inline: 2rem;
 			}
 			.d2l-collapsible-panel {
+				background-color: #ffffff;
```
dlockhart, line 100: I've seen cases where the background colour behind the panel bleeds through:

image

This would likely impact those? [Slack thread](https://d2l.slack.com/archives/C03NDJRSE1Y/p1675710396572409).

dbatiste: The background-color case is a bit more complicated and deserves its own PR so I am going to remove it from here. I'll open a separate PR to address it.

dlockhart (approved):

PR #3307, opened 2023-02-17
fix: Scroll-wrapper buttons and sticky controls

eKoopmans:
Small follow-up to #3283, to fix an edge case with z-indexes and scroll-wrapper - the scroll-wrapper buttons now sit below any sticky controls:

image

Rally: https://rally1.rallydev.com/#/?detail=/userstory/656750407555&fdp=true


dlockhart (approved):
Nice!

PR #3310, opened 2023-02-17
Fix d2l-tooltip so that only one tooltip is displayed at a time.

dbatiste:
[DE51993](https://rally1.rallydev.com/#/?detail=/defect/685233504963&fdp=true)

This PR fixes the `d2l-tooltip` component so that only one tooltip is displayed at once. Previous to this change, more than one tooltip could be displayed but it was never really supported - as such, tooltip collisions could occur. In addition, logic in `ListItemMixin` assumes only one tooltip is displayed at a time when it adds its `_tooltip-showing` attribute to elevate `z-index` for list-items.

The following screenshots show the results when a tooltip is shown by clicking followed by showing a second tooltip by subsequently hovering.

**Before:**
image

image

image

**After:**
image

image

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -430,6 +430,8 @@ class Tooltip extends RtlMixin(LitElement) {
 		this._onTargetTouchStart = this._onTargetTouchStart.bind(this);
 		this._onTargetTouchEnd = this._onTargetTouchEnd.bind(this);
 
+		this._onTooltipShowOther = this._onTooltipShowOther.bind(this);
+
 		this.announced = false;
 		this.closeOnClick = false;
 		this.delay = 300;
@@ -474,6 +476,7 @@ class Tooltip extends RtlMixin(LitElement) {
 		super.disconnectedCallback();
 		this._removeListeners();
 		window.removeEventListener('resize', this._onTargetResize);
+		document.body.removeEventListener('d2l-tooltip-show', this._onTooltipShowOther);
 		clearDismissible(this._dismissibleId);
 		delayTimeoutId = null;
 		this._dismissibleId = null;
@@ -859,6 +862,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		}, 500);
 	}
 
+	_onTooltipShowOther() {
+		// only allow one tooltip showing at a time
+		this.hide();
+	}
+
 	_removeListeners() {
 		if (!this._target) {
 			return;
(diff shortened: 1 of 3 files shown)


components/tooltip/tooltip.js
```diff
@@ -474,6 +476,7 @@ class Tooltip extends RtlMixin(LitElement) {
 		super.disconnectedCallback();
 		this._removeListeners();
 		window.removeEventListener('resize', this._onTargetResize);
+		document.body.removeEventListener('d2l-tooltip-show', this._onTooltipShowOther);
```
dlockhart, line 479: I think this is already removed in `_removeListeners()`?

components/tooltip/tooltip.js
```diff
@@ -859,6 +862,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		}, 500);
 	}
 
+	_onTooltipShowOther() {
+		// only allow one tooltip showing at a time
+		this.hide();
```
dlockhart, line 867: It would complicate things a lot, but one scenario this doesn't handle is your first screenshot in the PR description: you have a tooltip that shows on-focus or on-hover, you focus on it, the tooltip shows, then you hover over another element, the tooltip hides, then you stop hovering... despite focus still being on the first element, the tooltip won't come back.

It's a simplified solution, but we could: also listen for `d2l-tooltip-hide` and if we were hidden show ourselves again. That tracking around "if we were hidden" would have to reset when the tooltip was shown/hidden through normal means. It would also only work for one tooltip -- if multiple were hidden (that'd be weird right?) then only one could actually show itself again.

Anyway, not a big deal at all -- not sure if the complexity of the solutions is worth it.

components/tooltip/tooltip.js
dbatiste: Na... this one is different, for the new listener that I'm adding when the tooltip is shown. Normally it would be removed when the tooltip is hidden, but by off-chance the tooltip was disconnected while still being shown, this just makes sure we're not leaking memory. If the tooltip happens to be reconnected, it is initialized in the hidden state, and so upon showing again the handler would be once again wired up.

components/tooltip/tooltip.js
dbatiste: I see, yeah, currently they would have to hover over it again, or move focus away and then back again as if they had used `Escape` to dismiss it. Would it be weird to show the original tooltip again after moving the mouse away from the second target - the user would think it's been dismissed, but then it reappears?

components/tooltip/tooltip.js
dlockhart: Ah sorry my bad. The diff shown in GitHub made it look like you'd added a remove there as well but it was hiding a bunch of code so it wasn't _actually_ there.

Screen Shot 2023-02-17 at 5 46 01 PM

components/tooltip/tooltip.js
dlockhart: Yeah it might actually make it look even more buggy and confusing eh. Meh, good with whatever.

dlockhart (approved):

PR #3311, opened 2023-02-17
fix: list item styles

GZolla:
A couple of bugs exist in list-item related to button-disabled and selection-disabled:

- Selectable list items with `button-disabled`  are not selected when clicking on them as the disabled button covers most of the element
- ~~`list-item-button` when used in the same list as `list-item` has a top border because of `:first-of-type`~~
    - No use cases exist as of now of combined d2l-list-item and d2l-list-item-button, solutions would need to be programmatic and therefore expensive

- Selected items with selection-disabled have no bottom border.
- Hovering primary action makes checkbox glow when clicking won't change selection.

Rally: https://rally1.rallydev.com/#/?detail=/defect/684867182819&fdp=true


components/list/list-item-button-mixin.js
```diff
@@ -20,38 +20,15 @@ export const ListItemButtonMixin = superclass => class extends ListItemMixin(sup
...
-			:host(:not([selectable]):not([button-disabled]):not([skeleton])[_focusing]) [slot="outside-control-container"],
-			:host([selectable]:not([selection-disabled][button-disabled]):not([skeleton])[_hovering]) [slot="outside-control-container"],
-			:host([selectable]:not([selection-disabled][button-disabled]):not([skeleton])[_focusing]) [slot="outside-control-container"] {
-				background-color: white;
-				border-color: #b6cbe8; /* celestine alpha 0.3 */
-			}
-			:host([selectable]:not([selection-disabled][button-disabled]):not([skeleton])[_hovering]) [slot="outside-control-container"],
-			:host(:not([selectable]):not([button-disabled]):not([skeleton])[_hovering]) [slot="outside-control-container"] {
-				box-shadow: 0 0 10px rgba(0, 0, 0, 0.1);
-			}
```
GZolla, line 54: All selection/focusing is now handled by styles in `list-item-mixin`

components/list/list-item-link-mixin.js
```diff
@@ -21,20 +21,7 @@ export const ListItemLinkMixin = superclass => class extends ListItemMixin(super
...
-				border-top-color: transparent;
-			}
-			:host(:not([skeleton])[action-href]:not([action-href=""])[_hovering]) [slot="outside-control-container"],
-			:host(:not([skeleton])[action-href]:not([action-href=""])[_focusing]) [slot="outside-control-container"] {
-				background-color: white;
-				border-color: #b6cbe8; /* celestine alpha 0.3 */
-			}
-			:host(:not([skeleton])[action-href]:not([action-href=""])[_hovering]) [slot="outside-control-container"] {
-				box-shadow: 0 0 10px rgba(0, 0, 0, 0.1);
-			}
```
GZolla, line 37: All selection/focusing is now handled by styles in list-item-mixin

components/list/list-item-mixin.js
```diff
@@ -317,17 +321,20 @@ export const ListItemMixin = superclass => class extends composeMixins(
...
 			}
 
-			:host(:not([selection-disabled]):not([skeleton])[selected]) [slot="outside-control-container"],
-			:host(:not([selection-disabled]):not([skeleton])[selectable][_hovering]) [slot="outside-control-container"],
+			:host([_hovering-primary-action]) [slot="outside-control-container"],
+			:host([_hovering-selection]) [slot="outside-control-container"],
+			:host([_focusing-primary-action]) [slot="outside-control-container"],
 			:host(:not([selection-disabled]):not([skeleton])[selectable][_focusing]) [slot="outside-control-container"] {
-				background-color: white;
+				background-color: white !important;
```
GZolla, line 328: `!important` used to overwrite line 336 for selected list items.

components/list/test/list.visual-diff.js
```diff
@@ -147,6 +147,18 @@ describe('d2l-list', () => {
...
 		] },
+		{ category: 'selectableHref', tests: [
+			{ name: 'hover href', selector: '#selectableHref', action: () => hover('#selectableHref [selectable]') },
+			{ name: 'hover selection', selector: '#selectableHref', action: async() => {
+				const [x, y, height] = await page.$eval('#selectableHref [selectable]', elem => {
+					const rect = elem.getBoundingClientRect();
+					return [rect.x, rect.y, rect.height];
+				});
+				page.mouse.move(x + 5, y + height / 2);
+			} },
```
GZolla, line 158: **hover selection** test hovers over the area of the `list-item` that is not primary action, hence activating selection styles.

components/list/list-item-button-mixin.js
```diff
@@ -20,38 +20,15 @@ export const ListItemButtonMixin = superclass => class extends ListItemMixin(sup
...
-			:host([button-disabled][_focusing-primary-action]) .d2l-list-item-content {
-				--d2l-list-item-content-text-color: unset;
-				--d2l-list-item-content-text-decoration: none;
-			}
-			:host([button-disabled]) button {
-				cursor: default;
+			:host([skeleton]) [slot="content-action"],
+			:host([button-disabled]) [slot="content-action"] {
+				pointer-events: none;
 			}
```
oelhanafey, line 26: This doesn't appear to be a regression but it looks like the selection-inputs still work in skeleton mode. 

This can be a separate ticket since it was an issue before as well but I think it would be a good idea to spin up a defect for that.

components/list/test/list.visual-diff.js
oelhanafey: Is it possible to grab a specific element here instead of manually moving the mouse? I'm a bit worried this could be finicky and might break if we change the component.

components/list/list-item-mixin.js
oelhanafey: Similarly, it's not clear to me why these styles need to be changed. Couldn't we still use `_hovering` as we are with `_focusing`?

components/list/list-item-mixin.js
```diff
@@ -147,12 +147,16 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			:host(:last-of-type[_separators="between"]) [slot="control-container"]::after,
 			:host([_separators="none"]) [slot="control-container"]::before,
 			:host([_separators="none"]) [slot="control-container"]::after,
-			:host([selectable][_hovering]:not([selection-disabled])) [slot="control-container"]::before,
-			:host([selectable][_hovering]:not([selection-disabled])) [slot="control-container"]::after,
```
oelhanafey, line 151: The `_hovering` property still exists right? Why do we need to change all the uses of this property. Or was the intention to remove it altogether?

GZolla: It still exists but it is set to true even if hovering over a secondary action, this means styles where being applied to both the checkbox and the list item even though neither selection nor primary action where being hovered.

components/list/test/list.visual-diff.js
GZolla: I tried that, shadow dom seems to prevent selecting items that are generated and not directly declared as as child. Ideally `[slot="control"]` inside the element would be selected.

components/list/list-item-checkbox-mixin.js
```diff
@@ -156,6 +165,8 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 		};
 		return this.selectable ? html`
 			<div @click="${this._onCheckboxActionClick}"
+				@mouseenter="${this._onMouseEnterSelection}"
+				@mouseleave="${this._onMouseLeaveSelection}"
```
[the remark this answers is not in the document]
oelhanafey: Sorry, I didn't notice. That's actually consistent with this component so this is fine as is

components/list/list-item-mixin.js
oelhanafey: Ah so the checks for selection-disabled and skeleton are not needed anymore because they are being checked before adding those properties? It's still probably a good idea to avoid using `!important` if we can.

Maybe instead, we could conditionally add a class to the container to give these styles higher specificity

components/list/test/list.visual-diff.js
oelhanafey: Could you instead just manually add the `_hovering-selection` attribute to the list-item?

components/list/test/list.visual-diff.js
```diff
@@ -147,6 +147,16 @@ describe('d2l-list', () => {
...
+		{ category: 'selectableHref', tests: [
+			{ name: 'hover href', selector: '#selectableHref', action: () => hover('#selectableHref [selectable]') },
+			{ name: 'hover selection', selector: '#selectableHref', action: async() => {
+				const item = await page.$('#selectableHref [selectable]');
+				const control = await page.evaluateHandle(item => item.shadowRoot.querySelector('[slot="control"]'), item);
+				return control.hover();
+			} },
+			{ name: 'hover secondary action', selector: '#selectableHref', action: () => hover('#selectableHref [selectable] d2l-button-icon') },
+		] },
+
```
oelhanafey, line 159:
```suggestion
```

components/list/test/list.visual-diff.html
```diff
@@ -341,6 +341,19 @@
 				</d2l-list>
 			</div>
 		</div>
+		<div style="width: 400px;">
```
oelhanafey, line 344:
```suggestion

		<div style="width: 400px;">
```

components/list/list-item-checkbox-mixin.js
```diff
@@ -123,6 +125,13 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 			}
 		}
 	}
+	_onMouseEnterSelection() {
+		this._hoveringSelection = !this.selectionDisabled;
```
oelhanafey, line 129: Since we are checking the `selectionDisabled` property here instead of in the CSS, if selection is disabled when the mouse enters the selection area, the `_hoveringSelection` property will not be set. Then, if selection is enabled, the selection would be hovered but the correct styles wouldn't be used. Or would the `mouse-enter` event be dispatched again when the component re-renders? Is that a case we need to cover?

oelhanafey (commented):
I've been playing around with it and I think these styles look better! 

I added some code spacing nits to keep the document spacing consistent.

components/list/list-item-checkbox-mixin.js
oelhanafey: I don't think turning off pointer events fixes the problem. It introduces a similar issue, where we if `_hovering-selection` is set and then the selection is disabled, the `_hovering-selection` styles will never be removed.

It might be better to keep these `selection-disabled` checks in the CSS styles.

GZolla: I looked into that and saw that if hovering over the element when selection is disabled, the styles will only update after moving the mouse. This can become instant with a call to `willUpdate` to update `_hovering-selection` after changing `selection-disabled`, but that maybe too expensive just to handle that edge case.

I also saw that the same problem with `_hovering-primary-action`(even before these PR's changes) and I could apply the same changes there.

The main issue with setting the checks in the style is that they would need to be overridden in `d2l-list-item-button-mixin` and `d2l-list-item-link-mixin` which means a lot of repeated code and complex selections that initially caused the defects this PR tries to solve.

oelhanafey: I think using `willUpdate` to set `_hoveringSelection` to false sounds reasonable here then with your initial solution. It shouldn't be expensive as it would occur before the DOM updates so it should still only update once.

oelhanafey (approved):
Looks good. From my testing, this seems to work as expected. It would be a good idea to get Mark to have a look as well because there are a lot of edge cases.

PR #3315, opened 2023-02-21
Support anchor tabs for single-page app routing

bearfriend:
In order to support deep linking to tabs within single-page apps, `d2l-tabs` needs to support using anchors for tabs.

- Moves the actual tab inside tab-internal shadow DOM
- If an `href` is provided use an anchor for the tab
- Do not select the tab or dispatch an event when clicked
- Adds basic support for SPA routing to all demos, if applied

I have another branch that uses the host itself as its `renderRoot` (so attaches the template to the light dom) that would allow us to use `aria-controls` correctly, but since we don't use it at all now (and `ARIAMixin` is gaining support for element references soon-ish) I left this in the shadow DOM.

[US148994](https://rally1.rallydev.com/#/?detail=/userstory/686279846799&fdp=true): Add support for anchor tabs to d2l-tabs

### DIFF
--- a/components/tabs/tab-internal.js
+++ b/components/tabs/tab-internal.js
@@ -5,86 +5,8 @@ import { RtlMixin } from '../../mixins/rtl-mixin.js';
 
 const keyCodes = {
 	ENTER: 13,
-	SPACE: 32,
-	ARROWS: [35, 36, 37, 38, 39, 40]
+	SPACE: 32
 };
-
-export const tabInternalStyles = css`
-	d2l-tab-internal {
-		box-sizing: border-box;
-		display: inline-block;
-		max-width: 200px;
-	}
-	[role="tab"] {
-		color: inherit;
-		display: inline-block;
-		outline: none;
-		position: relative;
-		text-decoration: unset;
-		vertical-align: middle;
-		max-width: 100%;
-	}
-	.d2l-tab-text {
-		margin: 0.5rem;
-		overflow: hidden;
-		padding: 0.1rem;
-		text-overflow: ellipsis;
-		white-space: nowrap;
-	}
-	d2l-tab-internal:first-child .d2l-tab-text {
-		margin-left: 0;
-	}
-	d2l-tab-internal[dir="rtl"]:first-child .d2l-tab-text {
-		margin-left: 0.6rem;
-		margin-right: 0;
-	}
-	.d2l-tab-selected-indicator {
-		border-top: 4px solid var(--d2l-color-celestine);
-		border-top-left-radius: 4px;
-		border-top-right-radius: 4px;
-		bottom: 0;
-		display: none;
-		margin: 1px 0.6rem 0 0.6rem;
-		position: absolute;
-		-webkit-transition: box-shadow 0.2s;
-		transition: box-shadow 0.2s;
-		width: calc(100% - 1.2rem);
-	}
-	d2l-tab-internal:first-child .d2l-tab-selected-indicator {
-		margin-left: 0;
-		width: calc(100% - 0.6rem);
-	}
-	d2l-tab-internal[dir="rtl"]:first-child .d2l-tab-selected-indicator {
-		margin-left: 0.6rem;
-		margin-right: 0;
-	}
-	[role="tab"].focus-visible > .d2l-tab-text {
-		border-radius: 0.3rem;
-		box-shadow: 0 0 0 2px var(--d2l-color-celestine);
-		color: var(--d2l-color-celestine);
-	}
-	d2l-tab-internal[selected] [role="tab"]:focus {
-		text-decoration: none;
-	}
-	[role="tab"]:hover {
-		color: var(--d2l-color-celestine);
-		cursor: pointer;
-	}
-	d2l-tab-internal[selected] [role="tab"]:hover {
-		color: inherit;
-		cursor: default;
-	}
-	d2l-tab-internal[selected] .d2l-tab-selected-indicator {
-		display: block;
-	}
-
-	@media (prefers-reduced-motion: reduce) {
-		.d2l-tab-selected-indicator {
-			-webkit-transition: none;
-			transition: none;
-		}
-	}
-	`;
 
 class Tab extends RtlMixin(FocusVisiblePolyfillMixin(LitElement)) {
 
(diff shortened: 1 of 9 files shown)


components/tabs/tab-internal.js
```diff
@@ -90,50 +98,71 @@ class Tab extends RtlMixin(LitElement) {
...
-			if (e.keyCode !== keyCodes.SPACE) return;
-			e.stopPropagation();
-			e.preventDefault();
-		});
-		this.addEventListener('keyup', (e) => {
-			if (e.keyCode !== keyCodes.ENTER && e.keyCode !== keyCodes.SPACE) return;
-			this.ariaSelected = 'true';
-		});
+		this.selected = false;
+		this.role = 'none';
```
bearfriend, line 105: Required to keep the `tablist > tab` role relationship intact.

components/tabs/tab-internal.js
```diff
@@ -90,50 +98,71 @@ class Tab extends RtlMixin(LitElement) {
...
 
 	render() {
-		return html`
+		return this.href ? html`
+		<a
+		 href="${this.href}"
+		 role="tab"
+		 tabindex="${this.activeFocusable ? 0 : -1}"
+		 aria-selected="${this.selected}"
+		 aria-controls="${this.controlsPanel}"
```
bearfriend, line 115: This currently doesn't work because of shadow DOM, but `aria-controls` isn't used currently for `d2l-tabs` anyway. I have another branch that doesn't use shadow DOM where this would work.

components/tabs/tab-internal.js
```diff
@@ -90,50 +98,71 @@ class Tab extends RtlMixin(LitElement) {
...
+		 title="${this.text}"
+		 @keydown="${this._handleKeyDown}"
+		 @keyup="${this._handleKeyUp}">
```
bearfriend, line 118: There is no click handler because the anchor should ultimately push a new state into the history and the app should then select the tab based on that.

components/tabs/tab-panel-mixin.js
```diff
@@ -40,6 +45,11 @@ export const TabPanelMixin = superclass => class extends superclass {
 			:host([selected]) {
 				display: block;
 			}
+			:host(.focus-visible) {
+				outline: none;
+				border-radius: 0.3rem;
+				box-shadow: 0 0 0 2px var(--d2l-color-celestine);
+			}
```
bearfriend, line 52: Tab panels [should](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) be focusable if the first element within the panel is not focusable.
>When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

components/tabs/tabs.js
```diff
@@ -93,6 +93,7 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 				display: flex;
 			}
 			.d2l-tabs-layout-shown {
+				align-items: center;
```
bearfriend, line 96: Snuck this in to vertically align my `ext` slot contents with the tabs.

components/tabs/tabs.js
```diff
@@ -291,10 +292,8 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(FocusVisiblePolyf
 				}
 
 				const expanded = await this._tryExpandTabsContainer(measures);
-				if (expanded) {
-					return;
-				} else {
-					return this._updateScrollPosition(tabInfo);
+				if (!expanded) {
+					await this._updateScrollPosition(tabInfo);
```
bearfriend, line 296: This is from some experimentation that I only partially reverted. I can revert this too.

index.html
```diff
@@ -1,11 +1,12 @@
...
-	<link rel="stylesheet" href="components/demo/styles.css" type="text/css">
+	<link rel="stylesheet" href="/components/demo/styles.css" type="text/css">
 	<script type="module">
-		import './components/colors/colors.js';
-		import './components/typography/typography.js';
-		import './index.js';
+		/* eslint-disable import/no-absolute-path, indent */
+		import '/components/colors/colors.js';
+		import '/components/typography/typography.js';
+		import '/index.js';
```
bearfriend, line 9: Required for SPA routing.

mixins/arrow-keys-mixin.js
```diff
@@ -44,7 +44,7 @@ export const ArrowKeysMixin = superclass => class extends superclass {
 
 	async _focus(elem) {
 		if (elem) {
-			if (this.arrowKeysOnBeforeFocus) await this.arrowKeysOnBeforeFocus(elem);
+			this.arrowKeysOnBeforeFocus && await this.arrowKeysOnBeforeFocus(elem);
```
bearfriend, line 47: More experimentation only partially reverted.

package.json
```diff
@@ -18,7 +18,7 @@
     "lint:eslint": "eslint . --ext .js,.html --ignore-path .gitignore",
     "lint:lit": "lit-analyzer \"{components,controllers,directives,helpers,mixins,templates,test,tools}/**/*.js\" --strict --rules.no-unknown-tag-name off",
     "lint:style": "stylelint \"**/*.{js,html}\" --ignore-path .gitignore",
-    "start": "web-dev-server --node-resolve --watch --open",
+    "start": "web-dev-server --app-index index.html --node-resolve --watch --open",
```
bearfriend, line 21: This supports SPA routing

components/tabs/tabs.js
bearfriend: ... but maybe we don't want this.

bearfriend:
One small issue is that if normal tabs and anchor tabs are mixed, you can get into a state where you can't get back to an anchor tab that matches the current path without first navigating to a different path. This is a bad pattern I would want to discourage anyway. You could either enable the click handler for anchor tabs (would encourage/enable poor routing implementation), or put a boolean on `d2l-tabs` to enable the `href`s to be used, and then require them.

bearfriend:
If we want the static site to be able to refresh correctly, we should be able to add a redirect rule to the bucket.

PR #3317, opened 2023-02-21
Add custom empty states to filter

oelhanafey:
Keeping this as a draft until the previous empty state PR (#3263) is merged in. 

[US148588](https://rally1.rallydev.com/#/?detail=/userstory/685338133643&fdp=true)

This PR introduces 2 new slots to the `d2l-filter-dimension-set` and adds the `d2l-filter-dimension-set-empty-state` component. If no valid empty state is given, then the default empty state will be rendered.

Since the `d2l-filter` handles the rendering, it needs to be notified of data changes to the empty state. There were 3 separate cases that needed to be handled for empty state changes:
1. The `d2l-filter-dimension-set-empty-state` component attributes change -> Send the changed attributes up to d2l-filter through an event and apply them
2. The `search-empty-state` or `set-empty-state` slot changes -> Re-grab the changed empty state from the dimension set
3. The `d2l-filter-dimenion-set` changes -> Grab both empty states from the dimension set


components/filter/filter-dimension-set.js
```diff
@@ -113,15 +119,62 @@ class FilterDimensionSet extends LitElement {
...
+
 	_handleDimensionSetValueDataChange(e) {
 		e.stopPropagation();
-		this._dispatchDataChangeEvent({ dimensionKey: this.key, valueKey: e.detail.valueKey, changes: e.detail.changes });
+		this._dispatchEvent('d2l-filter-dimension-data-change', { dimensionKey: this.key, valueKey: e.detail.valueKey, changes: e.detail.changes });
+	}
+
+	_handleSearchEmptyStateSlotChange(e) {
+		if (!this._searchEmptyStateSlot) this._searchEmptyStateSlot = e.target;
+		this._dispatchEvent('d2l-filter-dimension-empty-state-slot-change', { dimensionKey: this.key, type: EmptyStateType.Search });
```
oelhanafey, line 166: The approach I took here creates a separate slot change event for the empty state that the d2l-filter handles differently from when the dimension-set-empty-state attributes change. This is slightly different from how the filter handles dimension-set-value changes where the dimension-set will dispatch the same event for both attribute changes and slot changes.

I originally tried to do it the same way with a single `d2l-filter-dimension-empty-state-change` event but I thought the resulting handler in `d2l-filter` was doing too much and it was cleaner to separate them into separate handlers.

components/filter/filter-dimension-set-empty-state.js
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		const changes = new Map();
+		changedProperties.forEach((oldValue, prop) => {
+			if (oldValue === undefined) return;
+
+			if (prop === 'actionHref' || prop === 'actionText' || prop === 'description') {
```
svanherk, line 43: So one thing you might have noticed about `d2l-filter-dimension-set` is that it doesn't list all properties here for simplicity.  We made the decision that, until we had a use-case for it, you don't get to switch your search type or whether the set is single-dimension-only after the fact. Do you know if the same would apply here?  Or does their use case call for different empty state text/action based on the search entered?

components/filter/filter-dimension-set-empty-state.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+class FilterDimensionSetEmptyState extends LitElement {
+
+	static get properties() {
+		return {
+			/**
+			 * The href that will be used for the empty state action. When set with action-text, d2l-filter will render a link action.
+			 * @type {string}
+			 */
+			actionHref: { type: String, attribute: 'action-href' },
```
svanherk, line 15: Do you know if Wave plans to use both the action button and the action href, or just one? Just wondering if we can delay adding the second type until someone really needed it, if we thought one was more suited than the other.

Following that same thought - would the href be a bit jarring, to click something inside a filter and have the page navigate?  Maybe a @geurts question

svanherk (commented):
Couple early questions just about `d2l-filter-dimension-set-empty-state` itself

components/filter/filter-dimension-set-empty-state.js
oelhanafey: That's a good question that I'll forward along. We could start with just one but one of the big things we wanted to emphasize when creating the empty state components was that we wanted people to use the semantically correct action for their use case. We didn't want people to end up using buttons as links.

If we introduce buttons first, then I'm a bit worried people will just use the buttons as links rather than come to us and request that we add functionality for the link action.

components/filter/filter-dimension-set-empty-state.js
oelhanafey: Ah okay. I'll check, but as far as I know that is not a requirement and we could probably do it this way as well. In the case of the empty state though, these properties are just text values so it doesn't add much complexity to handle the change events so personally I see it as a nice-to-have.

But yeah, if we don't need that, then we could potentially remove the `d2l-filter-empty-state-change` events and handlers altogether.

svanherk: Yeah, up to you. Doesn't save a _whole_ lot, just trying to think of ways to keep filter simpler and the component wrapping less awkward.

svanherk: Although, yeah, if we could _also_ clean up the `slot` change events for it, and just say that you need to have it there from the get-go, I think that cleans up _a lot_.

Thinking out loud, since I'm still not all the way through the filter code - without the data events, would that make tying it back to the same `d2l-filter-diemnsion-set-data-change` event and `_handleDimensionDataChange` simpler?  Like when the slot changes for the search empty set, you'd:
```
this._dispatchEvent('d2l-filter-dimension-data-change', { dimensionKey: this.key, changes: new Map([['searchEmptyState', this.getSearchEmptyState()]]) });
```
Which would trigger `_handleDimensionDataChange` to set it, without any changes needed to that function because it doesn't need to cause a recount or a search or anything fancy?

oelhanafey: Yeah, I was doing something like this initially but I thought it was cleaner and easier to understand as a separate event and thought the data change handler was already doing a lot. I don't mind going back to that though if it's preferred and it would be a lot simpler without the empty state data changes.

I'll get back to you on whether those data change events are needed.

components/filter/filter-dimension-set-empty-state.js
oelhanafey: Chatted with Jeff about this. There are no concerns with navigating with an href inside of a filter. I think it would be a good idea to keep both actions for the above reason.

components/filter/filter-dimension-set-empty-state.js
svanherk: > I don't mind going back to that though if it's preferred and it would be a lot simpler without the empty state data changes.

Yeah I'd be interested in seeing what it looks like, maybe as a separate PR/branch to compare easily

oelhanafey: So design doesn't have an explicit rule on search descriptions and we leave it up to the consumer on if they want to use something like `x returned no results` but there isn't any use case that requires it at the moment. It still could be nice to have for the future if it was needed at some point though. 

I'll create a separate branch and try your suggestion

dlockhart (approved):

components/filter/filter-dimension-set.js
```diff
@@ -1,9 +1,16 @@
 import { html, LitElement } from 'lit';
 
+export const EmptyStateType = {
```
svanherk, line 3: Minor: Since this isn't used in here anymore, we could maybe move all the logic to one file or the other. For example, we could clean things up a bit by just sending the type as part of the `_getEmptyState` response.  Something like:
```
getSearchEmptyState() {
	return this._getEmptyState(this._searchEmptyStateSlot, 'search');
}

getSetEmptyState() {
	return this._getEmptyState(this._setEmptyStateSlot, 'set');
}

_getEmptyState(emptyStateSlot, emptyStateType) {
...
		return {
			actionHref: emptyState.actionHref,
			actionText: emptyState.actionText,
			description: emptyState.description,
			type: emptyStateType
		};
}
```
And then in filter have:
```
_createEmptyState(emptyState, dimensionKey) {
  <d2l-empty-state-action-button
	  @d2l-empty-state-action="${this._handleEmptyStateAction}"
	  data-dimension-key="${dimensionKey}"
	  data-type="${emptyState.type}"
```

Or even keep the const and pass `EmptyStateType.Search` and `EmptyStateType.Set` into `_getEmptyState`, but not have to export it anymore. Not a big deal either way, but it doesn't seem like `d2l-filter` really _cares_ what the string is, just that it needs to pass something to the consumer for differentiating.

oelhanafey: I missed that, that sounds reasonable. I like the idea of moving the type into the `_getEmptyState()` response

PR #3319, opened 2023-02-21
Update solution for only showing one tooltip at a time.

dbatiste:
This PR updates `d2l-tooltip` code that enforces only one tooltip being displayed at once. This is an alternate, simpler solution than what I previously implemented in this PR https://github.com/BrightspaceUI/core/pull/3310. It tracks the active tooltip so that it can be hidden if/when another tooltip is shown.

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -886,8 +886,10 @@ class Tooltip extends RtlMixin(LitElement) {
 		clearTimeout(this._hoverTimeout);
 		clearTimeout(this._longPressTimeout);
 		if (newValue) {
-			if (activeTooltip) activeTooltip.hide();
-			activeTooltip = this;
+			if (!this.forceShow) {
+				if (activeTooltip) activeTooltip.hide();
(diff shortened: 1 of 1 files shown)


components/tooltip/tooltip.js
```diff
@@ -890,6 +886,11 @@ class Tooltip extends RtlMixin(LitElement) {
 		clearTimeout(this._hoverTimeout);
 		clearTimeout(this._longPressTimeout);
 		if (newValue) {
+			if (!this.forceShow) {
```
dbatiste, line 889: The code works fine without this condition, but without it, `activeTooltip` can be assigned to a tooltip with `force-show` even though those tooltips cannot be hidden. So this just ensures that we don't have that weird state, and it makes it an explicit consideration in the code.

It would be theoretically possible to have tooltip collisions when using `force-show`. That attribute was (and still is) incompatible with the idea of only showing one tooltip at a time.

dlockhart (approved):

PR #3326, opened 2023-02-22
fix: add `outline-offset` to inline collapsible panel

jameswklassen:
## Description

Inline collapsible panels were designed to take up the full width of their container. This causes an issue with the focus ring, which extends beyond the container bounds.

image

A negative `outline-offset` seems to be the best solution here, but open to thoughts.

## Rally
https://rally1.rallydev.com/#/378960033600d/dashboard?detail=%2Fdefect%2F685783970591


dlockhart:
Nit: for the inline visual-diffs, wondering if it would be useful to surround them in a background color and/or border such that the bug would have been obvious and the change here would have clearly shown it being fixed?

dlockhart (approved):

PR #3333, opened 2023-02-23
Remove special enter/space handling from d2l-list-item-generic-layout.

dbatiste:
This PR removes special `enter`/`space` handling from the `d2l-list-item-generic-layout` because it interferes with interactive elements in list-items while in grid mode. The guidelines for grid mode don't mention any special handling for these keys. This code was being used to toggle the selection checkbox and invoke the primary action link on both `enter` and `space`, which is not native behaviour for links for checkboxes.


components/list/list-item-generic-layout.js
```diff
@@ -445,16 +445,10 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
+
 		let node = null;
 		let preventDefault = true;
 		switch (event.keyCode) {
-			case keyCodes.ENTER:
-			case keyCodes.SPACE:
-				node = getComposedActiveElement();
-				node.click();
-				// mouseup event is necessary for components like d2l-dropdown-more that control state by listening to this event
-				node.dispatchEvent(new MouseEvent('mouseup', { composed: true }));
```
dbatiste, line 456: Entering this `ENTER`/`SPACE` case meant that we were preventing default and stopping propagation even when these keys were pressed for dropdowns/menus in the secondary actions slot, which previously required this fake `mouseup` event hack so dropdowns would open as expected. So eliminating this case logic means we don't need this hack anymore.

components/list/list-item-checkbox-mixin.js
```diff
@@ -150,24 +156,25 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
 		return this.selectable ? html`
 			<d2l-selection-input
 				@d2l-selection-change="${this._onCheckboxChange}"
-				?selected="${this.selected}"
```
dbatiste, line 153: Just the new `keydown` handler here. The rest of the changes are just ordering.

components/list/list-item-checkbox-mixin.js
```diff
@@ -150,24 +156,25 @@ export const ListItemCheckboxMixin = superclass => class extends SkeletonMixin(s
...
-				.hovering="${this._hoveringSelection}">
+				?selected="${this.selected}"
+				?skeleton="${this.skeleton}">
 			</d2l-selection-input>
 		` : nothing;
 	}
 
 	_renderCheckboxAction(inner) {
 		return this.selectable ? html`
-			<div @click="${this._onCheckboxActionClick}"
```
dbatiste, line 167: Just ordering these.

dlockhart (approved):
LGTM.

PR #3334, opened 2023-02-23
Alternative custom empty state approach

oelhanafey:
This is a branch off of #3317 to try out a different approach to adding custom empty states.

The code needs to be cleaned up but the main things to have a look at in this PR: 
`d2l-filter-dimension-set-empty-state`:
    - Removed the change events to clean up code. We could add this back to listen to changes to the component properties in the future if a consumer needed it
`d2l-filter-dimension-set`:
    - Removed the `d2l-filter-empty-state-change` and `d2l-filter-empty-state-slot-change` events. When the empty state slots change, it regrabs the EmptyState for that slot and dispatches the existing `d2l-filter-dimension-change` event.
`d2l-filter`:
    - Since we removed the empty state change events we can remove the handlers here and the `d2l-filter-dimension-data-change` handler handles the empty state changes without any updates to it.
 
It worked pretty seamlessly. I really like this approach and it cleans it up a lot. I think it would be a good idea to pivot to this.

### DIFF
--- a/components/filter/README.md
+++ b/components/filter/README.md
@@ -154,6 +154,7 @@ The filter will announce changes to filter selections, search results, and when 
 
 ### Events
 * `d2l-filter-change`: dispatched when any filter value has changed (may contain info about multiple dimensions and multiple changes in each)
+* `d2l-filter-dimension-empty-state-action`: dispatched when an empty state action button is clicked
 * `d2l-filter-dimension-first-open`: dispatched when a dimension is opened for the first time (if there is only one dimension, this will be dispatched when the dropdown is first opened)
 * `d2l-filter-dimension-search`: dispatched when a dimension that supports searching and has the "manual" search-type is searched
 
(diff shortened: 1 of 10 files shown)


svanherk:
Yeah, this cleans up a lot of the complicated bits. I like it too!

> - Removed the change events to clean up code. We could add this back to listen to changes to the component properties in the future if a consumer needed it

And if we did need to add this back, we could just run `_handleSearchEmptyStateSlotChange` whenever a component property changes.  Something like:
```
<slot name="search-empty-state" @d2l-filter-dimension-set-empty-state-change="${this._handleSearchEmptyStateSlotChange}" @slotchange="${this._handleSearchEmptyStateSlotChange}"></slot>
```
(Could just rename it to `_handleSearchEmptyStateChange` at that point.)

Arguably slightly less efficient to grab all three properties again, but much simpler logic. Not nearly the same perf hit as having to grab and recalculate all `d2l-filter-dimension-set-value`s each time one changes.

PR #3336, opened 2023-02-23
feat: DE51630 allow no scrollbar

mdulat:
[DE51630](https://rally1.rallydev.com/#/?detail=/defect/675773914365&fdp=true): [New Discussion Creation] > New forum dialog flickers upon opening it 

Allows client to specify no scrollbar so that we don't get a scrollbar flicker when opening a dialog.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -42,6 +42,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			 * The optional title for the dialog
 			 */
 			titleText: { type: String, attribute: 'title-text' },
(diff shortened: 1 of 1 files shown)


dlockhart:
Hi @mdulat! Is it still needed after the fullscreen dialog stuff went in? I see the underlying defect is closed. It’s pretty stale at this point, so hoping we can close it out.

mdulat:
It's a separate issue which is still an issue, but there is a [Gaudi defect](https://rally1.rallydev.com/#/?detail=/defect/687701331483&fdp=true) to address it properly, as this solution isn't great. So I'll close this out.

PR #3340, opened 2023-02-24
feat: switch button from using FocusVisiblePolyfill to the new prefix method

dlockhart:
With the colour input work (which uses `buttonStyles`), I was getting really annoyed that I'd get a focus outline every time I clicked on the button. It's even worse because clicking the button opens a dropdown and focus moves into there, so the outline would flicker.

This switches our shared `buttonStyles` and `ButtonMixin` from using the `:focus-visible` polyfill to the new prefix utility. I _think_ this is safe to do to consumers of the mixin? A worry would be consumers that are using the mixin AND specifying styles for `.focus-visible` like ButtonIcon is, but [a code search](https://search.d2l.dev/search?project=Brightspace&project=Brightspace-Configs&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&full=%22ButtonMixin%22+%22.focus-visible%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5) yields no results.

Another issue I ran into was places like dialogs and focus-trap that were using `forceFocusVisible()` to programatically set visible focus. The workaround here is to also specify styles for the `.focus-visible` CSS class, but now only `forceFocusVisible()` will be adding/removing that and not the polyfill.


components/button/button-icon.js
```diff
@@ -116,7 +117,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 					--d2l-icon-fill-color: var(--d2l-button-icon-fill-color-hover, var(--d2l-color-tungsten));
 					background-color: var(--d2l-button-icon-background-color-hover);
 				}
-				button.focus-visible {
+				button:${unsafeCSS(getFocusPseudoClass())} {
```
dlockhart, line 120: Button-icon is overriding the default `:focus-visible` styles from `buttonStyles`, which is why it needs to do this.

components/button/test/button-icon.visual-diff.js
```diff
@@ -35,7 +34,7 @@ describe('d2l-button-icon', () => {
 					const selector = `#${entry.category}`;
 
 					if (name === 'hover') await page.hover(selector);
-					else if (name === 'focus') await page.$eval(selector, (elem) => forceFocusVisible(elem));
+					else if (name === 'focus') await focus(page, selector);
```
dlockhart, line 37: We'll see if this works... object-property-list does this and it works there. 🤞

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -15,7 +16,7 @@ describe('d2l-list-item-drag-handle', () => {
 	let browser, page;
 
 	const focusMethod = (selector) => {
-		return page.$eval(selector, (item) => { item.focus(); });
+		return page.$eval(selector, (item) => { forceFocusVisible(item); });
```
dlockhart, line 19: This was using `buttonStyles` which now doesn't have an outline for `:focus` so it needed to switch to this.

dbatiste (approved):

components/list/test/list-item-drag-handle.visual-diff.js
```diff
@@ -42,7 +43,6 @@ describe('d2l-list-item-drag-handle', () => {
 		});
 
 		itWithReload('keyboard-mode', async function() {
-			await focusMethod('d2l-list-item-drag-handle');
```
dlockhart, line 45: This wasn't working but was also redundant with the earlier tests for focus, so I just removed focusing on the drag handle.

PR #3347, opened 2023-02-27
feat: Shrink margins when dialog is tall

eKoopmans:
This PR makes three sets of changes, as described in the story:

1. Changes the dialog default top margin to `75px`
2. Switches `ease-in` animations to `ease-out`
3. Adds a "full-height" mode to dialogs that automatically switches to 1.5rem top/bottom margins when the dialog is too tall for the viewport

I confirmed these design details with Jeff:

- ✔️ We want the new `75px` to apply to all margins, i.e. including `d2l-dialog-confirm`
- ✔️ The switch should happen when the "preferred height" exceeds the available height
- ✔️ We're okay with this being a full switch, not a gradual change between the two modes
- ✔️ We're okay with this being invoked any time consumers call `.resize()`

Rally: https://rally1.rallydev.com/#/?detail=/defect/685414266339&fdp=true


components/dialog/dialog-fullscreen.js
```diff
@@ -127,10 +127,6 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 					animation: d2l-dialog-fullscreen-open 400ms ease-out;
 				}
 
-				dialog::backdrop {
-					transition: opacity 200ms ease-out;
-				}
```
eKoopmans, line 132: This is removed because it's already captured in `dialog-styles.js`, now that both are `ease-out`.

oelhanafey (commented):
I like the changes!

I noticed a weird case with nested dialogs and unsure of the expected behaviour here. The switch to `full-size` is based on the height of that specific dialog. When you have a child dialog that is a different size from the parent,  then you could end up in this middle state where the dialogs have different placements.

Screen Shot 2023-02-27 at 3 50 18 PM

eKoopmans:
Very good catch, thanks @oelhanafey! I'll ask Jeff what the desired behaviour is there.

PR #3357, opened 2023-03-01
fix: Move dialog full-height to an internal class

eKoopmans:
This PR moves the dialog full-height behaviour to an internal class, instead of sprouting the `full-height` attribute on the component; recommended by @dbatiste. There should be no functional changes.

Rally: https://rally1.rallydev.com/#/?detail=/defect/685414266339&fdp=true


dbatiste (approved):
Looks great! (linting aside)

oelhanafey (approved):

oelhanafey (approved):

PR #3358, opened 2023-03-01
feat: color input that relies on MVC

dlockhart:
This introduces a `<d2l-input-color>` component which currently relies on the MVC framework to render the actual palette for choosing a colour.

It operates in two "launch types" that consumers don't need to worry about:
1. `launch-type` = `dialog`: this is the default mode intended for use from other web components. When clicked, the colour input launches the existing MVC dialog containing the palette.
2. `launch-type` = `dropdown`: this will be used from the existing MVC control where it will render the palette into the `<slot>` and the input will render that inside a dropdown.


components/inputs/demo/input-color-palette.js
```diff
@@ -0,0 +1,67 @@
+import '../../button/button.js';
+import '../input-text.js';
+import { css, html, LitElement } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+class InputColorPalette extends LitElement {
```
dlockhart, line 6: This is just a demo helper component that simulates what the MVC control will do by rendering stuff inside the `<slot>`:

Screen Shot 2023-03-01 at 1 52 43 PM

components/inputs/docs/input-color.md
```diff
@@ -0,0 +1,60 @@
...
+  import '@brightspace-ui/core/components/inputs/input-color.js';
+</script>
+<d2l-input-color type="foreground" value="#ff0000"></d2l-input-color>
+```
+
+## Color Input [d2l-input-color]
+
+The `<d2l-input-color>` will open a dialog to allow the user to select a color from a palette.
+
+<!-- docs: demo live name:d2l-input-color -->
```
dlockhart, line 19: I'm tempted to not add this component to the Daylight Site just yet seeing as it's so reliant on the MVC framework. Curious what others think.

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+			type: { reflect: true, type: String },
+			/**
+			 * Value of the input as a HEX color
+			 * @type {string}
+			 */
+			value: { type: String },
+			/**
+			 * @ignore
+			 */
+			launchType: { attribute: 'launch-type', type: String },
```
dlockhart, line 132: Consumers should never be setting this (other than the MVC framework).

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this._validateLabel();
+	}
+
+	render() {
+
+		const label = !this.labelHidden ? html`<div class="d2l-input-label">${this._getLabel()}</div>` : nothing;
```
dlockhart, line 271: Unlike with other inputs this "label" is just "for show", since technically this is a button and not an input and the button uses a combination of the tooltip and `aria-label` as its label.

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+	_getLabel() {
+		if (this.label !== undefined) return this.label;
+		if (this.type === 'background') return this.localize('components.input-color.backgroundColor');
+		if (this.type === 'foreground') return this.localize('components.input-color.foregroundColor');
+		return '';
+	}
+
+	_getOpener() {
+		if (this.readonly) {
+			return html`<div class="readonly-wrapper" id="opener" role="img" tabindex="0">${this._getSwatch()}</div>`;
```
dlockhart, line 297: When in `readonly` mode, I had a good accessibility experience with using an `img` role labelled by the tooltip.

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+		}
+		const buttonClass = {
+			'd2l-button': true,
+			'd2l-dropdown-opener': this.launchType === 'dropdown',
+			'opened': this._opened
+		};
+		const ariaLabel = this._opened ? this._getTooltipLabel() : undefined;
```
dlockhart, line 304: When the dropdown is open we hide the tooltip, so we need an explicit `aria-label`.

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+		const button = html`<button id="opener" class="${classMap(buttonClass)}" aria-disabled="${ifDefined(this.disabled ? 'true' : undefined)}" aria-label="${ifDefined(ariaLabel)}" @click="${this._handleOpenDialog}">${this._getSwatch()}<d2l-icon icon="tier1:chevron-down-small"></d2l-icon></button>`;
```
dlockhart, line 305: Similar to other components, we don't actually `disable` the button so that it can still receive focus and show the tooltip.

components/inputs/input-color.js
```diff
@@ -0,0 +1,411 @@
...
+		if (this.launchType === 'dialog') {
+			return button;
+		}
+		return html`
+			<d2l-dropdown @d2l-dropdown-close="${this._handleClose}" @d2l-dropdown-open="${this._handleOpenDropdown}">
+				${button}
+				<d2l-dropdown-content max-width="350" mobile-tray="left" no-padding ?opened="${this._opened}">
+					<slot @d2l-input-color-dropdown-close="${this._handleCloseRequest}"><div style="padding: 10px">Currently, color inputs only support a "launch-type" of "dropdown" from within the MVC framework. Use "launch-type" of "dialog" instead.</div></slot>
```
dlockhart, line 313: This is just a helpful message to folks if they try to use this in `dropdown` mode outside of a MVC page.

Screen Shot 2023-03-01 at 10 19 40 AM

lang/en.js
```diff
@@ -32,6 +32,9 @@ export default {
 	"components.form-element.input.url.typeMismatch": "URL is not valid",
 	"components.form-element.valueMissing": "{label} is required.",
 	"components.form-error-summary.errorSummary": "{count, plural, one {There was {count} error found in the information you submitted} other {There were {count} errors found in the information you submitted}}",
+	"components.input-color.backgroundColor": "Background Color",
+	"components.input-color.foregroundColor": "Foreground Color",
```
dlockhart, line 36: This uncovered the fact that core doesn't support the `en-CA` or `en-GB` language packs. Will deal with that separately.

components/inputs/docs/input-color.md
dbatiste: I am also on the fence, but I think I lean towards including it so that designers/devs know what we have and where it stands.

components/inputs/docs/input-color.md
```diff
@@ -0,0 +1,60 @@
...
+
+| Property | Type | Description |
+|---|---|---|
+| `associated-value` | String | Optional value of an associated color as a HEX which will be used for color contrast analysis. |
+| `disabled` | Boolean, default: `false` | Disables the input. |
+| `disallow-none` | Boolean, default: `false` | Disallows the user from selecting "None" as a color value. |
+| `label` | String, required | Label for the input. Comes with a default value for background & foreground types. |
+| `label-hidden` | Boolean, default: `false` | Hides the label visually. |
+| `readonly` | Boolean, default: `false` | Puts the input into a read-only state. |
+| `type` | String, default: `custom` | Type of color being chosen. Can be one of: `custom`, `background`, `foreground`. |
```
dbatiste, line 38: If we don't like "custom", "other" might be an alternative. I don't feel strongly one way or another.

dlockhart: Cool. I went with "custom" because it aligns with the convention used in the MVC framework already -- `ColourSelectorForeground`, `ColourSelectorBackground` and `ColourSelectorCustom`.

components/inputs/input-color.js
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	_handleClose() {
+		this._opened = false;
+		this.dispatchEvent(new CustomEvent(
+			'd2l-input-color-close', { bubbles: false, composed: false }
+		));
+	}
+
+	_handleCloseRequest(e) {
```
dbatiste, line 355: It looks like this handler only runs if the dropdown is in play. Same for `_handleClose`. When the dialog is in play, when does `this._opened` get set to `false`?

components/inputs/docs/input-color.md
oelhanafey: Would designers reading this page understand what the MVC framework is and whether they are able to use the component? I do agree that I think it would probably be better to have the documentation there than not, but I can see that causing confusion.

components/inputs/input-color.js
dlockhart: Good catch. `_opened` was really just for the dropdown and to hide the tooltip when it was open -- not needed for the dialog case since it's technically never "open".

components/inputs/docs/input-color.md
dlockhart: They might have a general understanding, but yeah it's mostly a dev-focused message. The important thing I think is that it can be used everywhere except FRAs (and we could technically make it work there if we really wanted to).

components/inputs/input-color.js
```diff
@@ -0,0 +1,415 @@
...
+		this._updateValueAndDispatchEvent(e.detail.newValue);
+	}
+
+	_handleOpenDialog() {
+		if (this.launchType !== 'dialog') return;
+		if (window?.D2L?.LP?.Web?.UI?.Desktop?.MasterPages?.Dialog?.Open === undefined) {
+			console.warn('<d2l-color-input>: when "launch-type" is "dialog", component must be hosted in a LMS page.');
+			return;
+		}
+		this._opened = true;
```
dbatiste, line 368: Based on the above comments, this line isn't necessary?

dlockhart: Ah forgot to push, fixed now.

dbatiste (approved):

PR #3362, opened 2023-03-02
en-GB support

dlockhart:
This adds `en-GB` support to core and to the dynamic mixin. I'm still in discussions with the localization team about `en-CA`, but for now it seems like it's not necessary to support it and isn't supported by any of the infrastructure anyway.


components/demo/demo-page.js
```diff
@@ -80,7 +80,8 @@ class DemoPage extends LitElement {
 		const classes = {
 			'no-scroll': this._noScroll
 		};
-		const selectedLanguageCode = getDocumentLocaleSettings().language;
+		let selectedLanguageCode = getDocumentLocaleSettings().language;
+		if (selectedLanguageCode === 'en') selectedLanguageCode = 'en-us';
```
dlockhart, line 84: The way the language selector in the demo page works, it goes through all the options and finds the first partial match. But because the page's language is `en` it was matching `en-ca` even though the components will interpret that as `en-us`.

lang/en-gb.js
```diff
@@ -0,0 +1,129 @@
...
+	"components.form-element.input.number.rangeOverflow": "{maxExclusive, select, true {Number must be less than {max}.} other {Number must be less than or equal to {max}.}}",
+	"components.form-element.input.number.rangeUnderflow": "{minExclusive, select, true {Number must be greater than {min}.} other {Number must be greater than or equal to {min}.}}",
+	"components.form-element.input.text.tooShort": "{label} must be at least {minlength} characters",
+	"components.form-element.input.url.typeMismatch": "URL is not valid",
+	"components.form-element.valueMissing": "{label} is required.",
+	"components.form-error-summary.errorSummary": "{count, plural, one {There was {count} error found in the information you submitted} other {There were {count} errors found in the information you submitted}}",
+	"components.input-color.backgroundColor": "Background Colour",
+	"components.input-color.foregroundColor": "Foreground Colour",
```
dlockhart, line 36: The reason we need this.

dbatiste (approved):

eKoopmans (approved):

PR #3376, opened 2023-03-08
feat: Use getFocusPseudoClass for focus-visible

eKoopmans:
This PR cleans up our `:focus-visible` approach in core. All [non-core consumers of `focus-visible-polyfill-mixin`](https://search.d2l.dev/search?full=%22focus-visible-polyfill-mixin%22&si=full&searchall=true) have already been migrated off of it, so it's safe to remove it and the `focus-visible` dependency.

I had to add one new helper, `isFocusVisibleApplied()`, to detect when the `:focus-visible` pseudo-selector is applied to an element - this was easier when it was only the class `.focus-visible` that we were checking for.

In all usages of `getFocusPseudoClass()` we also keep the selector `.focus-visible`, which is still used by `forceFocusVisible()` so it can't be removed.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/685236909367&fdp=true


package.json
```diff
@@ -71,7 +71,6 @@
     "@brightspace-ui/intl": "^3",
     "@formatjs/intl-pluralrules": "^1",
     "@open-wc/dedupe-mixin": "^1",
-    "focus-visible": "^5",
```
dbatiste, line 74: 🎉

helpers/focus.js
```diff
@@ -186,6 +186,12 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
 
 }
 
+export function isFocusVisibleApplied(node) {
```
dbatiste, line 189: Might be good to have a couple of tests for this.

dlockhart (approved):
Yay, awesome!

helpers/focus.js
```diff
@@ -188,6 +188,7 @@ export function isFocusable(node, includeHidden, includeTabbablesOnly, includeDi
 
 export function isFocusVisibleApplied(node) {
 	if (!node) return false;
+	// return node.classList.contains('focus-visible');
```
dlockhart, line 191: ^

test/test-helpers.js
```diff
@@ -0,0 +1,11 @@
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
+
+export const testRunnerKeyboardFocus = async(element) => {
+	await sendKeys({ press: 'Tab' });
+	element.focus();
+};
+export const testRunnerMouseFocus = async(element) => {
+	const { x, y } = element.getBoundingClientRect();
+	await sendMouse({ type: 'click', position: [x, y] });
+};
```
dlockhart, line 10: Nice!

test/test-helpers.js
```diff
@@ -0,0 +1,11 @@
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
```
dbatiste, line 1: I think there may already be a place to put these.

https://github.com/BrightspaceUI/core/blob/main/tools/dom-test-helpers.js

There's the `tools` and the `test` folder. The original intention was that stuff in `test` wasn't thought to be useful to other projects so it's not published, whereas the stuff in `tools` is/was, so it is published. That could probably stand to be reviewed, but I would probably combine these with the existing ones that are published. Assuming they are moved, I'd probably rename them since the other methods don't follow this format... probably `focusWithKeyboard` and `focusWithMouse`

eKoopmans: Thanks Dave, done - I kept my helpers in a separate file though, because they're specifically for `@web/test-runner`, not the DOM in general (there's no way to force "keyboard mode"/"mouse mode" in the DOM directly). We may end up wanting similar helpers for `visual-diff` in the future.

dbatiste: Sounds good!

dbatiste (approved):

PR #3380, opened 2023-03-09
Localize rich text

bearfriend:
Iteration of #3370

POC using IntlMessageFormat directly for rich HTML with langfiles, with validation of allowed html elements.

### DIFF
--- a/helpers/localize.js
+++ b/helpers/localize.js
@@ -1,0 +1,40 @@
+import { html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+export const acceptedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
+
+const markupError = `localize() rich-text replacements must use markup templates with only the following accepted elements: ${acceptedTags}. [link to docs]`;
+const regex = new RegExp(`<(?!\/?(${acceptedTags.join('|')}))`);
+
+export function validateMarkup(parts) {
+	return parts.map(p => {
+		debugger;
+		if (p._markup) return p;
+		else if (p.constructor !== String) throw markupError;
+		return p;
+	});
+}
+
+export function markup(str, ...parts) {
+	let reject = str.some(str => regex.test(str));
+	reject = reject || parts.some(subparts => { // subparts === chunks?
+		if (subparts.constructor === String) {
+			return regex.test(subparts);
+		}
+		else if (Array.isArray(subparts)) {
+			 return !subparts.some(subpart => subpart.constructor === String || subpart._markup);
+		}
+		else {
+			return !subparts._markup;
+		}
+	});
+	if (reject) throw markupError;
+	const _html = html(str, ...parts);
+	_html._markup = true;
+	return _html;
+}
+
+export const linkGenerator = ({ href, target }) => {
+	import('../components/link/link.js');
+	return chunks => markup`<d2l-link href="${ifDefined(href)}" target="${ifDefined(target)}">${chunks}</d2l-link>`;
+};
--- a/helpers/localize.js
+++ b/helpers/localize.js
@@ -4,7 +4,7 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 export const allowedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
 
 const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following allowed elements: ${allowedTags}. [link to docs]`;
-const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')}|>|\s)+).*?>`);
+const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')})?[\\s>])`);
 
 export function validateMarkup(content, applyRegex) {
 	if (content) {
--- a/helpers/localize.js
+++ b/helpers/localize.js
@@ -1,23 +1,24 @@
 import { html } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 
-export const acceptedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
+export const allowedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
 
-const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following accepted elements: ${acceptedTags}. [link to docs]`;
-const acceptedTagsRegex = new RegExp(`<(?!/?(${acceptedTags.join('|')}))`);
+const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following allowed elements: ${allowedTags}. [link to docs]`;
+const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')}|>|\s)+).*?>`);
 
 export function validateMarkup(content, applyRegex) {
 	if (content) {
-		if (content.map) return content.map(item => validateMarkup(item, applyRegex));
+		if (content.map) return content.map(item => validateMarkup(item));
 		if (content._markup) return content;
 		if (Object.hasOwn(content, '_$litType$')) throw markupError;
-		if (applyRegex && content.constructor === String && acceptedTagsRegex.test(content)) throw markupError;
+		if (applyRegex && content.constructor === String && disallowedTagsRegex.test(content)) throw markupError;
 	}
 	return content;
 }
 
 export function markup(strings, ...expressions) {
-	validateMarkup(strings, true) && validateMarkup(expressions, true);
+	strings.forEach(str => validateMarkup(str, true));
+	expressions.forEach(exp => validateMarkup(exp, true));
 	return { ...html(strings, ...expressions), _markup: true };
 }
 
--- a/helpers/localize.js
+++ b/helpers/localize.js
@@ -4,7 +4,9 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 export const allowedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
 
 const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following allowed elements: ${allowedTags}. [link to docs]`;
-const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')})?[\\s>])`);
+const validTerminators = '([>\\s/]|$)';
+const allowedAfterTriangleBracket = `/?(${allowedTags.join('|')})?${validTerminators}`;
+const disallowedTagsRegex = new RegExp(`<(?!${allowedAfterTriangleBracket})`);
 
 export function validateMarkup(content, applyRegex) {
 	if (content) {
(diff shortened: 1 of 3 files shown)


eKoopmans (commented):
Haha okay. There's a bit of cleaning up to do but I'm satisfied with this approach - using the native support but maintaining our iron grip. A few suggestions:

1. I really think we get rid of `localizeHTML` and have this directly as part of `localize`, especially since we'll want to `validateMarkup` in `localize` anyway
2. We should add a helper for `d2l-tooltip-help`
3. I think we could clean up the "reject" logic in `markup` by calling `validateMarkup` (possibly with some recursion in `validateMarkup` for arrays?)
4. I'd personally keep the localize helpers in the same spot as the localize mixin - ideally create a folder under `mixins` for all the `localize` stuff.

eKoopmans:
Also, nice work 😄

bearfriend:
Yeah, definitely still messy. Just wanted to get something up before EOD.

> A few suggestions...

I think I agree with all of this, except I _think_ the reject logic is different in all three cases so can't be consolidated/recursive. Probably can be cleaner, though.

helpers/localize.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function markup(str, ...parts) {
+	let reject = str.some(str => regex.test(str));
+	reject = reject || parts.some(subparts => { // subparts === chunks?
+		if (subparts.constructor === String) {
+			return regex.test(subparts);
+		}
+		else if (Array.isArray(subparts)) {
+			 return !subparts.some(subpart => subpart.constructor === String || subpart._markup);
```
eKoopmans, line 25: These arrays (in the `html` literal) can be infinitely nested, so some recursion is probably necessary. I've got a suggestion I can put up soon (let's see if we can get infinite recursion on PRs! 😉)

bearfriend: Ah, I see. I think we are thinking of two different approaches. Mine (obviously not properly implemented here), would be to disallow even strings, so every level would be independently calling `markup` and validating. This is more like what lit itself does, but I could be convinced to allow strings and just do the recursion inside the validation logic.

bearfriend: And now that I'm thinking it through, I guess lit is more forgiving. Writing that logic might be overcomplicating things, so now I'm leaning toward the recursion.

eKoopmans: Yeah I guess what I'm saying is this should probably be valid (it would be valid in an html literal): `markup`${['a', ['b', ['c', 'd']]]}`` - and you probably want to do all those checks all the way down.

eKoopmans: Okay - a draft with some additional suggestions:
https://github.com/BrightspaceUI/core/pull/3381

eKoopmans:
> ideally create a folder under mixins for all the localize stuff

When I suggested this I forgot that we can't easily move the existing `localize-mixin.js` because it would break imports everywhere. Also moving the `.md` could break the documentation repo, except I think we don't document mixins there for some reason?

bearfriend:
> > ideally create a folder under mixins for all the localize stuff
> 
> When I suggested this I forgot that we can't easily move the existing `localize-mixin.js` because it would break imports everywhere. Also moving the `.md` could break the documentation repo, except I think we don't document mixins there for some reason?

We can't, but we could rethink the structure here, and just make the existing mixin import the new structure and map it to the old structure.

eKoopmans:
So, where are we at with this? Are we okay to finish collapsing the PRs?

bearfriend:
> So, where are we at with this? Are we okay to finish collapsing the PRs?

Yeah, this just needs to be approved, I guess. I did make a change to how the validation happens that should be reviewed.

helpers/localize.js
```diff
@@ -1,23 +1,24 @@
 import { html } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 
-export const acceptedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
+export const allowedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
 
-const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following accepted elements: ${acceptedTags}. [link to docs]`;
-const acceptedTagsRegex = new RegExp(`<(?!/?(${acceptedTags.join('|')}))`);
+const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following allowed elements: ${allowedTags}. [link to docs]`;
+const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')}|>|\s)+).*?>`);
```
eKoopmans, line 7: Okay so parsing out this regex:

- match anything that starts with `<`
- exclude things that optionally have a `/` (closing tag), and have one or more of:
  - the allowed tags OR `>` OR one whitespace
- and then the matching can finish with any number of characters followed by a `>`

I'm wondering: why? Some specific questions:
- why would it be allowed to have more than one of the accepted tags? That means `<pstrongbr>` would be accepted?
- why is `>` in there? That means `<>` and `<p>>` would be accepted?
- why is `\s` in there? That means `<p br> d2l-link>> b>` would be accepted?
- do we still run into excluding things we don't want to? For instance `<d2l-tooltip-help text="< a > are symbols">`? (This will just look at that `< and >` in the text and say it's invalid)

I'm getting more and more reluctant about this approach again 😅. I think it's a pretty good rule to never try to parse XML with regex...

eKoopmans: Also putting it out there - `<button>` gets past these filters, because `<b.*>` is valid. I think I might understand what you were going for with `>` and `\s`, something like:

```
const disallowedTagsRegex = new RegExp(`<(?!/?((${allowedTags.join('|')})(\\s+|>))).*?>`, 'g');
```

Essentially, saying "any allowed tag, always followed by either `\s+` or `>`"? (Also the double `\\` seems to be required).

bearfriend: Ok, made the opening tag optional to exclude things like `<>`, `< >`, and `< s>`:

```javascript
const disallowedTagsRegex = new RegExp(`<(?!/?((${allowedTags.join('|')})?[>\\s])).*?>`);
```

Match anything that:

- Begins with `<`
- Then, is _not_ followed by...
  - Optionally, `/`, then...
  - Optionally, any of the allowed tags
  - Then whitespace or `>`
 - Then anything, any number of times
 - ...until and including `>`
 
Screen Shot 2023-03-14 at 2 13 55 PM

And it shouldn't need to be global, since it only needs to match once to be `true`.

eKoopmans: Because [you can't parse HTML with regex](https://stackoverflow.com/a/1732454), how about a parser-based approach modeled on [this](https://stackoverflow.com/a/47140708)?

```js
const parser = new DOMParser();
const allowedTagsLookup = allowedTags.reduce((acc, v) => { acc[v] = true; return acc; }, {});

function hasDisallowedTags(html) {
	const elements = [ ...parser.parseFromString(html, 'text/html').body.querySelectorAll('*') ];
	return elements.some(v => !allowedTagsLookup[v.tagName.toLowerCase()]);
}
```

I don't like the performance impact of any of this (including the regex approach) given it'll be invoked on each render though. I really wish we could just trust our internal consumers...

eKoopmans: Scratch that - since this check is being run on the pieces of a template literal, which can be split up like `<a text="${'b'}">` => `['<a text="', '">']`, any actual dom-parser won't work. And neither will a regex that requires a closing `>`.

😭

bearfriend: That's fine if we really want to do that, but I would point out that this is not trying to parse HTML with regex. It's simply looking for actually a relatively simple _negative_ pattern. It doesn't care about attributes or child nodes, including text nodes with html entities. It only cares that the tags being used are allowed. Any false negative matches by definition wouldn't create a serious problem (beyond requiring further restriction), and any false positive matches could be worked around by escaping/encoding, which probably should be done anyway.

bearfriend: Splitting up html like ```[markup`<a href="`, markup`d2l.com">`]``` doesn't actually render as html because individually they do not parse.

eKoopmans: Agreed... But ```markup`<a href="${'b'}">link</a>` ```, when parsed by `markup()`, will have an array of strings `['<a href="', '">link</a>']`. And then each of those strings is passed to `validateMarkup()`.

bearfriend: Well, in that case the second part contains `</a>` which will fail. I guess someone could do `</${'a'}>` but at that point they're knowingly trying to get around it, which I don't think we want to do anything about.

eKoopmans: Alright. So we just want to catch any time a consumer uses `<some-tag-name`, unless it's one of our allowed tag names. With the understanding that there will be rare false positives if a consumer does like `text="<3 is a heart"`.

I can bring myself to accept a regex for that. I think it would look like:

```js
const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')})?([>\\s]|$))`)
```

eKoopmans: And for `</a>` - okay, but that doesn't apply to a situation like `<input type="${myType}">`, which has no closing tag...

bearfriend: Here's what looks like it works best to me:

```javascript
const disallowedTagsRegex = new RegExp(`<(?!/?((${allowedTags.join('|')})?)[\\s>]).+?[>\\s]`)
```

eKoopmans: Breaking it down:

```
<
(?!
	/?
	(	<- Redundant brackets
		(${allowedTags.join('|')})?
	)
	[\\s>]	<- I think we should include $ to account for string termination
)
.+?[>\\s]	<- I don't think this is necessary, we already account for "<>" and "< " in allowed usage above
```

bearfriend: 1. Yes, that is redundant
2. `$` doesn't hurt, but isn't necessary since it will just search until the end of the string anyway, and we don't explicitly need it to be the end
3. Yep, this was just helpful for positively matching during testing

eKoopmans: For 2, this is a false positive without the `$`:

```js
disallowedTagsRegex.test('<d2l-link') // true
disallowedTagsRegex.test('<d2l-link ') // false (expected)
```

eKoopmans: Okay... maybe `'<d2l-link'` without anything after it won't ever be possible in an `html` template literal.

One other to add to the list of valid terminators though: `/` (e.g. to account for `<br/>`).

helpers/localize.js
```diff
@@ -0,0 +1,28 @@
+import { html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+export const allowedTags = Object.freeze(['d2l-link', 'd2l-tooltip-help', 'p', 'br', 'b', 'strong', 'i', 'em']);
+
+const markupError = `localizeHTML() rich-text replacements must use markup templates with only the following allowed elements: ${allowedTags}. [link to docs]`;
+const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')})?[\\s>])`);
```
eKoopmans, line 7:
```suggestion
const validTerminators = '([>\\s/]|$)';
const disallowedTagsRegex = new RegExp(`<(?!/?(${allowedTags.join('|')})?${validTerminators})`);
```

helpers/localize.js
bearfriend: Ahh, I see what you mean with `$`. It doesn't hurt to include it, so we can.

helpers/localize.js
bearfriend: Kind of hurts my head more to break these out and try to piece them back together. Also, `allowedAfterTriangleBracket` is allowed in the code, but _not_ allowed in order to match the RegEx 🤯.

(`allowedTags` isn't a piece of the RegEx, and in fact is manipulated inline to be useful to it)

eKoopmans: I'm not attached to it, though I think it improves the readability. `allowedAfterTriangleBracket` is what's allowed to be there - `disallowedTagsRegex` negates that, as the name implies. Regex can be hard to read and even harder to document, so I like to self-document with variable names.

Anyway, I think separating the `validTerminators` at least is beneficial.

eKoopmans (approved):

PR #3385, opened 2023-03-10
replace forceFocusVisible() with focusWithKeyboard()

dlockhart:
In our visual-diff tests, using `forceFocusVisible` to test focus state isn't actually necessary now that we have the `focusWithKeyboard` visual-diff helper. In fact, it's dangerous because that only exercises the `.focus-visible` CSS class, and not the actual `:focus-visible` pseudo-class.

In a follow-up PR, I'll be updating visual-diff tests that are using `focus()` to test either `:focus-visible` or "focus with a mouse" to use `focusWithKeyboard` or `focusWithMouse`.

### DIFF
--- a/components/button/test/button-icon.visual-diff.html
+++ b/components/button/test/button-icon.visual-diff.html
@@ -5,8 +5,6 @@
 	<script type="module">
 		import '../../typography/typography.js';
 		import '../button-icon.js';
-		import { forceFocusVisible } from '../../../helpers/focus.js';
-		window.forceFocusVisible = forceFocusVisible;
 	</script>
 	<title>d2l-button-icon</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
--- a/components/button/test/button-icon.visual-diff.js
+++ b/components/button/test/button-icon.visual-diff.js
@@ -1,6 +1,5 @@
-/*global forceFocusVisible */
+import { focus, focusWithKeyboard, VisualDiff } from '@brightspace-ui/visual-diff';
 import puppeteer from 'puppeteer';
-import VisualDiff from '@brightspace-ui/visual-diff';
 
 describe('d2l-button-icon', () => {
 
@@ -35,8 +34,8 @@ describe('d2l-button-icon', () => {
 					const selector = `#${entry.category}`;
 
 					if (name === 'hover') await page.hover(selector);
-					else if (name === 'focus') await page.$eval(selector, (elem) => forceFocusVisible(elem));
-					else if (name === 'click') await page.$eval(selector, (elem) => elem.focus());
+					else if (name === 'focus') await focusWithKeyboard(page, selector);
+					else if (name === 'click') await focus(page, selector);
 
 					const rectId = (name.indexOf('disabled') !== -1) ? name : entry.category;
 					const rect = await visualDiff.getRect(page, `#${rectId}`);
--- a/components/button/test/button-icon.visual-diff.js
+++ b/components/button/test/button-icon.visual-diff.js
@@ -1,4 +1,4 @@
-import { focus, focusWithKeyboard, VisualDiff } from '@brightspace-ui/visual-diff';
+import { focusWithKeyboard, focusWithMouse, VisualDiff } from '@brightspace-ui/visual-diff';
 import puppeteer from 'puppeteer';
 
 describe('d2l-button-icon', () => {
@@ -35,7 +35,7 @@ describe('d2l-button-icon', () => {
 
 					if (name === 'hover') await page.hover(selector);
 					else if (name === 'focus') await focusWithKeyboard(page, selector);
-					else if (name === 'click') await focus(page, selector);
+					else if (name === 'click') await focusWithMouse(page, selector);
 
 					const rectId = (name.indexOf('disabled') !== -1) ? name : entry.category;
 					const rect = await visualDiff.getRect(page, `#${rectId}`);
--- a/components/card/card-footer-link.js
+++ b/components/card/card-footer-link.js
@@ -2,6 +2,7 @@ import '../colors/colors.js';
 import '../count-badge/count-badge-icon.js';
 import '../icons/icon.js';
 import { css, html, LitElement } from 'lit';
+import { FocusMixin } from '../../mixins/focus-mixin.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { offscreenStyles } from '../offscreen/offscreen.js';
 import { RtlMixin } from '../../mixins/rtl-mixin.js';
@@ -10,7 +11,7 @@ import { RtlMixin } from '../../mixins/rtl-mixin.js';
  * An icon link that can be placed in the `footer` slot.
  * @slot tooltip - slot for the link tooltip
  */
-class CardFooterLink extends RtlMixin(LitElement) {
+class CardFooterLink extends FocusMixin(RtlMixin(LitElement)) {
 
 	static get properties() {
 		return {
@@ -98,6 +99,10 @@ class CardFooterLink extends RtlMixin(LitElement) {
 		super();
 		this.download = false;
 		this.secondaryCountType = 'notification';
+	}
+
+	static get focusElementSelector() {
+		return 'a';
 	}
 
 	render() {
--- a/components/card/test/card.visual-diff.js
+++ b/components/card/test/card.visual-diff.js
@@ -1,7 +1,6 @@
-/*global forceFocusVisible */
+import { focusWithKeyboard, focusWithMouse, VisualDiff } from '@brightspace-ui/visual-diff';
 import { open } from '../../dropdown/test/dropdown-helper.js';
 import puppeteer from 'puppeteer';
-import VisualDiff from '@brightspace-ui/visual-diff';
 
 describe('d2l-card', () => {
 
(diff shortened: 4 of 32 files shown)


components/button/test/button-icon.visual-diff.js
```diff
@@ -35,8 +34,8 @@ describe('d2l-button-icon', () => {
 					const selector = `#${entry.category}`;
 
 					if (name === 'hover') await page.hover(selector);
-					else if (name === 'focus') await page.$eval(selector, (elem) => forceFocusVisible(elem));
-					else if (name === 'click') await page.$eval(selector, (elem) => elem.focus());
+					else if (name === 'focus') await focusWithKeyboard(page, selector);
+					else if (name === 'click') await focus(page, selector);
```
dlockhart, line 38: I don't love how we're using `focus()` for "click"... but at the same time maybe it's just the test name that's incorrect. Technically this is testing the "focus styles when focus is placed using the mouse".

components/button/test/button-icon.visual-diff.js
```diff
@@ -35,8 +34,8 @@ describe('d2l-button-icon', () => {
 					const selector = `#${entry.category}`;
 
 					if (name === 'hover') await page.hover(selector);
-					else if (name === 'focus') await page.$eval(selector, (elem) => forceFocusVisible(elem));
-					else if (name === 'click') await page.$eval(selector, (elem) => elem.focus());
+					else if (name === 'focus') await focusWithKeyboard(page, selector);
+					else if (name === 'click') await focusWithMouse(page, selector);
```
dbatiste, line 38: In hindsight, the "click" tests aren't named super well for what they were capturing. I know you probably don't want to cause a bunch of diff changes though.

dbatiste (commented):
Looks good.

components/card/card-footer-link.js
```diff
@@ -10,7 +11,7 @@ import { RtlMixin } from '../../mixins/rtl-mixin.js';
  * An icon link that can be placed in the `footer` slot.
  * @slot tooltip - slot for the link tooltip
  */
-class CardFooterLink extends RtlMixin(LitElement) {
+class CardFooterLink extends FocusMixin(RtlMixin(LitElement)) {
```
dlockhart, line 14: This and collapsible panel were in a situation where previously they were relying on the fact that `forceFocusVisible` finds the first focusable descendent to focus on, whereas now we simply call `focus()` on the element itself. By implementing `FocusMixin`, the components can delegate focus, which solves the problem.

components/inputs/test/input-search.visual-diff.js
```diff
@@ -57,38 +42,4 @@ describe('d2l-input-search', () => {
 		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 	});
```
dlockhart, line 44: I ended up deleting these tests. First, they were relying on querying the element's shadow root in order to focus -- which `focusWithKeyboard()` doesn't support. Second, they're essentially redundant because they're testing the focus/hover styles of `<d2l-button-icon>`, which we already have tests for.

components/tabs/test/tabs.visual-diff.js
```diff
@@ -118,19 +117,21 @@ describe('d2l-tabs', () => {
 				});
 
 				it('focus next', async function() {
+					await page.keyboard.press('Tab');
```
dlockhart, line 120: This is the first of several cases where we want to focus on something inside any element's shadow root, which `focusWithKeyboard()` and `focusWithMouse()` don't support.

We have a couple options:
1. This hackery, which essentially duplicates `focusWithKeyboard` but splits it into two. I don't love it.
2. We extend `focusWithKeyboard` to take an optional `shadowRootSelector`. When present, it would add an extra step before focusing to query the first selector's shadow root.

Thoughts?

components/tooltip/test/tooltip-help.visual-diff.js
```diff
@@ -50,7 +49,7 @@ describe('d2l-tooltip-help', () => {
 				await page.click(tooltipHelpSelector);
 			}
 
-			if (testCase.focus || testCase.hover || testCase.click) {
```
dlockhart, line 53: `focusWithKeyboard` _will_ open the tooltip, so we don't need to wait for it.

components/button/test/button-icon.visual-diff.html
```diff
@@ -5,8 +5,6 @@
 	<script type="module">
 		import '../../typography/typography.js';
 		import '../button-icon.js';
-		import { forceFocusVisible } from '../../../helpers/focus.js';
-		window.forceFocusVisible = forceFocusVisible;
```
eKoopmans, line 9: 🎉

components/inputs/test/input-search.visual-diff.js
```diff
@@ -58,21 +57,13 @@ describe('d2l-input-search', () => {
...
 
 	it('focus-search-button', async function() {
-		await page.evaluateHandle(() => {
-			forceFocusVisible(
-				document.querySelector('#no-value').shadowRoot.querySelector('d2l-button-icon[icon="tier1:search"]')
-			);
-		});
+		await focusWithKeyboard(page, ['#no-value', 'd2l-button-icon[icon="tier1:search"]']);
```
eKoopmans, line 60: Ah btw - after you've gone and added this array selector functionality for shadow roots, I remembered that we already have a solution using the `pierce/` selector in Puppeteer:
https://github.com/BrightspaceUI/core/blob/main/components/table/test/table.visual-diff.js#L153

Oops!

components/inputs/test/input-search.visual-diff.js
eKoopmans: Is this comment out of date? Looks like the tests are restored.

components/card/test/card.visual-diff.js
```diff
@@ -28,22 +27,22 @@ describe('d2l-card', () => {
 		{ name: 'align-center', selector: '#align-center' },
 		{ name: 'badge', selector: '#badge' },
 		{ name: 'actions', selector: '#actions' },
-		{ name: 'actions-focus', selector: '#actions', action: (selector) => page.$eval(`${selector} > d2l-button-icon`, (elem) => forceFocusVisible(elem)) },
+		{ name: 'actions-focus', selector: '#actions', action: (selector) => focusWithKeyboard(page, `${selector} > d2l-button-icon`) },
 		{ name: 'actions-rtl', selector: '#actions-rtl' },
-		{ name: 'no-link-focus', selector: '#header-content', action: (selector) => page.$eval(selector, (elem) => elem.focus()) },
+		{ name: 'no-link-focus', selector: '#header-content', action: (selector) => focusWithMouse(page, selector) },
```
eKoopmans, line 32: I'm surprised that this test was working as-is (and doesn't prompt a changed visual-diff), but looking closer it makes sense. It's entirely coincidental - `visualDiff.resetFocus()` was entering mouse mode, and `forceFocusVisible` wasn't entering keyboard mode, so it just worked out that this test would have been in mouse mode.

It's nice to be making these much more explicit!

components/collapsible-panel/test/collapsible-panel.visual-diff.js
```diff
@@ -20,8 +19,8 @@ describe('d2l-collapsible-panel', () => {
 
 	after(async() => await browser.close());
 
-	function focusElement(selector) {
-		return page.$eval(selector, (elem) => forceFocusVisible(elem));
+	async function focusElement(selector) {
+		return focusWithKeyboard(page, `${selector} > d2l-collapsible-panel`);
```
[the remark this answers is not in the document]
eKoopmans: Does this function need to be `async`?

eKoopmans (commented):
Looks great! So excited for this risk-of-flake getting eliminated from the tests.

components/inputs/test/input-search.visual-diff.js
dlockhart: Oh yup -- after we added the ability to pierce shadow root's I ended up restoring them. While it's true that they're testing button-icon things, they're also somewhat testing the position of that focus-ring'd button-icon inside the search box. So kind of useful still.

components/inputs/test/input-search.visual-diff.js
dlockhart: Oh neat -- I'll try that out and remove the other solution.

components/collapsible-panel/test/collapsible-panel.visual-diff.js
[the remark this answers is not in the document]
dlockhart: Hmm, maybe not. I had it `await focusWithKeyboard`-ing at one point while I was debugging, which would have required the `async`. But I think as it is now, it can be removed.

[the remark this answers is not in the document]
eKoopmans: Yep returning an async function will pass along the promise, `async` is only necessary when directly using `await`.

components/inputs/test/input-search.visual-diff.js
eKoopmans: Anyone else looking here - it turns out that `pierce/` doesn't "partially pierce" the shadow DOM, meaning you can't have different parts of the selector in different shadow DOMs - so it's less useful than I was thinking. [Official doc](https://pptr.dev/guides/query-selectors#pierce-selectors-pierce)

eKoopmans (approved):

eKoopmans:
BTW this PR didn't trigger a build, which means the two functional changes adding `FocusMixin` won't be picked up by consumers right away. Thought that might be a possible explanation for https://github.com/BrightspaceUI/htmleditor/pull/1298, but I don't think so.

PR #3386, opened 2023-03-10
Separate localize and localizeHTML completely; Allow all non-lit html…

bearfriend:
_One_ more for the stack.

- Separates localize and localizeHTML completely to avoid messiness/allowing default elements in `localize`
- Allows all non-lit types to be passes as replacement params
- Catches all `localize`/`localizeHTML` errors so as not to block `render`
- Fixes some naming

### DIFF
--- a/mixins/demo/localize-mixin-mission.js
+++ b/mixins/demo/localize-mixin-mission.js
@@ -6,8 +6,11 @@ class Mission extends LocalizeDynamicMixin(LitElement) {
 
 	static get localizeConfig() {
 		const langResources = {
-			'en': { mission: '<p><link1>Transforming</link1> the way</p><link2> <b>{name}</b></link2> learns. \'<div></div>\'' },
-			'fr': { mission: '<p><link1>Transformer</link1> la façon dont</p><link2> \'<br>\'\'</br>\' <b>{name}</b></link2> apprend' }
+			'en': {
+				mission: '<p><link1>Transforming</link1> the way</p><link2> <b>{name}</b></link2> learns. \'<div></div>\'',
+				a: '{a, select, true {T <i>{c}</i>} false {F - {b, date, medium}} other {O - {a}}}'
+			},
+			'fr': { mission: '<p><link1>Transformer</link1> la façon dont</p><link2> <br></br> <b>{name}</b></link2> apprend' }
 		};
 		return {
 			importFunc: async lang => langResources[lang]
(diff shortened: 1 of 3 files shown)


mixins/localize-mixin.js
```diff
@@ -97,8 +97,11 @@ export const LocalizeMixin = dedupeMixin(superclass => class LocalizeMixinClass
 
 	localize(key) {
 
+		const { language, value } = this.__resources?.[key] ?? {};
+		if (!value) return '';
+
 		let params = {};
-		if (arguments.length > 1 && typeof arguments[1] === 'object') {
+		if (arguments.length > 1 && arguments[1].constructor === Object) {
```
bearfriend, line 104: This is matching `new String()` objects, which technically should work as a param name.

mixins/demo/localize-mixin-mission.js
```diff
@@ -22,8 +25,16 @@ class Mission extends LocalizeDynamicMixin(LitElement) {
 			link1: linkGenerator({ href: 'https://wikipedia.org/wiki/Culture_change', target: '_blank' }),
 			link2: chunks => markup`<d2l-link href="https://wikipedia.org/wiki/Earth" target="_blank"><em>${chunks}</em> ${surnameMarkup}</d2l-link>`
 		};
+		let a;
```
eKoopmans, line 28: `const a = undefined`? I've got this thing against `let` statements where they're not absolutely needed 😆.

mixins/demo/localize-mixin-mission.js
```diff
@@ -22,8 +25,16 @@ class Mission extends LocalizeDynamicMixin(LitElement) {
...
 		return html`
 			${this.localizeHTML('mission', replacements)}
+			<div>1. ${this.localizeHTML('a', { a })}</div>
+			<div>2. ${this.localizeHTML('a', { a: null })}</div>
+			<div>3. ${this.localizeHTML('a', { a: 1 })}</div>
+			<div>4. ${this.localizeHTML('a', { a: {} })}</div>
+			<div>5. ${this.localizeHTML('a', { a: false, b: new Date() })}</div>
+			<div>6. ${this.localizeHTML('a', { a: true, c: [1, 2, 3] })}</div>
+			<div>7. ${this.localizeHTML('a', { a: true, c: markup`<b>test</b>` })}</div>
```
eKoopmans, line 37: I assume these can be moved into a unit test once everything's collapsed...

eKoopmans (approved):
Looks great! 🎉

mixins/demo/localize-mixin-mission.js
bearfriend: Uh oh. I agree, but I've also got a thing against literal `undefined` where they aren't absolutely needed. I think my brain is doing a kernel panic.

PR #3390, opened 2023-03-10
Clean up unnecessary EmptyStateType

oelhanafey:
Remove exported EmptyStateType constants that aren't needed anymore after some code cleanup previously.


svanherk (approved):
Nice!

PR #3396, opened 2023-03-13
Replace focus() with focusWithKeyboard()

dlockhart:
This builds on [the previous change](https://github.com/BrightspaceUI/core/pull/3385), this time switching visual-diff uses of `focus()` to `focusWithKeyboard()`.

I've left 8 occurrences of `focus()` from the date/time input components, simply due to them having some complex relationships between the focus code and events. [Here's an example](https://github.com/BrightspaceUI/core/blob/9f91e8b53818282a04e241f4886de6a1e1a8a8cf/components/inputs/test/input-date-range.visual-diff.js#L185), where a nested shadow root query is done to wire up an event listener that resolves a Promise and the `focus()` call initiates that. If anyone has ideas about how to solve something like this using `focusWithKeyboard`, I'm all 👂s!


components/list/test/screenshots/ci/golden/list/d2l-list-expand-collapse-button-focus.png
dlockhart, line 1: These two changes are expected -- we're now getting proper keyboard focus styles compared to before. 🎉

components/tabs/test/screenshots/ci/golden/tabs/d2l-tabs-basic-selected-tab-focus.png
dlockhart, line 1: Also expecting these two changes now that focus is happening with the keyboard properly.

eKoopmans:
Still looking at this PR 👀, got split focus (pun intended?) right now with axe-con.

> I've left 8 occurrences of focus() from the date/time input components, simply due to them having some complex relationships between the focus code and events. [Here's an example](https://github.com/BrightspaceUI/core/blob/9f91e8b53818282a04e241f4886de6a1e1a8a8cf/components/inputs/test/input-date-range.visual-diff.js#L185), where a nested shadow root query is done to wire up an event listener that resolves a Promise and the focus() call initiates that. If anyone has ideas about how to solve something like this using focusWithKeyboard, I'm all 👂s!

For that example, here's my suggestion:

```js
		async function focusOnInput(page, selector, inputSelector) {
			await page.$eval(selector, (elem, inputSelector) => {
				elem.blur();
				const input = elem.shadowRoot.querySelector(inputSelector);
					elem.addEventListener('d2l-tooltip-show', resolve, { once: true });
				});
			}, inputSelector);
			await focusWithKeyboard(page, [selector, inputSelector]);
		}
```

dlockhart:
No worries! Unfortunately that won't work since focusing is the thing that fires the `d2l-tooltip-show` event, so the Promise from the first async call won't ever resolve.

I think I can accomplish this by re-ordering them and doing the focus in a different thread:
```javascript
setTimeout(() => focusWithKeyboard(page, [selector, inputSelector]));
await page.$eval(selector, (elem, inputSelector) => {
	elem.blur();
	const input = elem.shadowRoot.querySelector(inputSelector);
		elem.addEventListener('d2l-tooltip-show', resolve, { once: true });
	});
}, inputSelector);
```

That approach worked elsewhere, but I'm not sure if it's introducing a timing-related flake.

eKoopmans:
Ah interesting, I missed that. How about:

```js
function promiseWithResolve() {
	let promiseResolve;
	const promise = new Promise(resolve => promiseResolve = resolve);
	return Object.assign(promise, { resolve: promiseResolve });
}

		async function focusOnInput(page, selector, inputSelector) {
			const tooltipShown = promiseWithResolve();
			await page.$eval(selector, (elem, inputSelector) => {
				elem.blur();
				const input = elem.shadowRoot.querySelector(inputSelector);
				elem.addEventListener('d2l-tooltip-show', tooltipShown.resolve, { once: true });
			}, inputSelector);
			await focusWithKeyboard(page, [selector, inputSelector]);
			await tooltipShown;
		}
```

dlockhart:
Won't work either since `tooltipShown` would need to be passed into `$eval` as a parameter, and only JSON serializable object types or JSHandles are supported.

components/tabs/test/tabs.visual-diff.js
```diff
@@ -7,12 +7,10 @@ describe('d2l-tabs', () => {
...
-	const focusTabs = (selector) => {
-		return page.$eval(selector, (elem) => {
-			return new Promise((resolve) => {
-				elem.focus();
-				requestAnimationFrame(resolve);
-			});
+	const focusTabs = async(selector) => {
+		await focusWithKeyboard(page, selector);
+		await page.evaluate(() => {
+			new Promise(resolve => requestAnimationFrame(resolve));
```
eKoopmans, line 13: So, some cases of needing `requestAnimationFrame`...

eKoopmans (approved):

components/tabs/test/tabs.visual-diff.js
dlockhart: Yeah, not a new case but another case still the same.

dbatiste: As you might know, tabs does a lot of stuff when moving focus around, and those things are `async`, which is probably why rAF is needed.

dbatiste (approved):

PR #3399, opened 2023-03-14
US146334 allow client to specify the width of a fullscreen dialog tha…

mdulat:
…t is smaller than the default

Part of [US146334](https://rally1.rallydev.com/#/?detail=/userstory/672777856045&fdp=true): [ui] update restriction type

We need a dialog that is full height but smaller width.

### DIFF
--- a/components/dialog/dialog-fullscreen.js
+++ b/components/dialog/dialog-fullscreen.js
@@ -202,7 +202,7 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 	}
 
 	render() {
-		this._width = Math.max(1170, this.width);
+		this._width = this.width;
 		const heightOverride = {} ;
 		let topOverride = null;
 		if (this._ifrauContextInfo) {


dbatiste:
This effectively overrides the intended behaviour of `d2l-dialog-fullscreen`. @geurts , do you have ideas on where we want to go with this? Should we regroup and review?

eKoopmans:
Yes, this has been [discussed earlier](https://d2l.slack.com/archives/C0PHG3QB0/p1676399301730559?thread_ts=1676051612.628289&cid=C0PHG3QB0) and resulted in [DE52039](https://rally1.rallydev.com/#/?detail=/defect/685414266339&fdp=true), where we improved the standard dialog to be able to take up more of the vertical space. Jeff was reluctant to remove this minimum-width restriction on the fullscreen dialog, so using a standard dialog was the proposed alternative.

I believe the issue here is that the standard dialog vertically resizes itself, while the desired behaviour is a dialog that remains full-height always, but is also narrower than the minimum `1170px`.

geurts:
Right -- I'll comment in the Slack thread to continue the conversation there. Looks like we dropped the ball and didn't respond to Margaret's latest comment there, which I guess lead to this pull request :(

dbatiste:
@mdulat , ok if we close this PR given https://github.com/BrightspaceUI/core/pull/3512 is merged?

PR #3405, opened 2023-03-15
replace focus calls that relied on events

dlockhart:
Draft of replacing the remaining `focus()` calls in our visual-diff that rely on a complex dance between focusing and waiting for a tooltip to show up before taking the screenshot.

These all had copies of a `focusOnInput` which I pulled out into a helper.


components/inputs/test/input-date-range.visual-diff.js
```diff
@@ -253,13 +242,13 @@ describe('d2l-input-date-range', () => {
 					it('focus start', async function() {
 						await focusOnInput(page, '#min-max', startDateSelector);
 						const rect = await getRectTooltip(page, '#min-max');
-						await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+						await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect, captureBeyondViewport: false });
```
dlockhart, line 245: These particular tests needed either `captureBeyondViewPort: false` or an extra `requestAnimationFrame`. Otherwise, the tooltip was only half displayed when the screenshot was taken.

components/inputs/test/input-color.visual-diff.js
```diff
@@ -32,14 +32,14 @@ describe('d2l-input-color', () => {
 	].forEach((name) => {
 		it(name, async function() {
 			const rect = await visualDiff.getRect(page, `#${name} > d2l-input-color`);
-			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect, captureBeyondViewPort: false });
```
dlockhart, line 35: These tests just started suddenly timing out in this particular PR... not locally, not in any of the others. No idea what's going on, but trying the magical fix.

components/inputs/test/input-date-range.visual-diff.js
eKoopmans: My best bet is that this is just because `captureBeyondViewport: false` has a few extra `await`s in it, so the browser has more time to finish painting:
https://github.com/puppeteer/puppeteer/blob/16b39de5ca5bdf70d8dbfe8168057991c6164bac/packages/puppeteer-core/src/common/Page.ts#L1400

If that's the case then throwing in an extra `requestAnimationFrame` may be more transparent. Maybe that could be added to the `focusOnInput` helper?

eKoopmans (approved):

components/inputs/test/input-color.visual-diff.js
```diff
@@ -13,6 +13,10 @@ describe('d2l-input-color', () => {
 		await page.bringToFront();
 	});
 
+	beforeEach(async() => {
+		await resetFocus(page);
+	});
```
dlockhart, line 18: Was getting a lot of flake in these input-color tests... not totally sure why it's cropping up now, but resetting the focus in between tests and waiting a bit longer for the tooltip (like we do in other tests) seemed to fix it consistently.

eKoopmans (approved):

svanherk:
Did `captureBeyondViewport: false` not help?  In the past, that's helped a lot with tests involving a `d2l-tooltip`, because `d2l-tooltip` listens to `resize` events on the `window`, which are thrown by the actual taking of the screenshot when you don't have `captureBeyondViewport: false`.

dlockhart:
> Did captureBeyondViewport: false not help?

It helped with the date/time tests, but weirdly it was only needed in some of them so we opted for a `rAF` instead.

It didn't help with colour inputs, but resetting focus between tests did (so far).

.github/workflows/visual-diff.yml
```diff
@@ -3,7 +3,7 @@ on: pull_request
 jobs:
   build:
     name: Visual Diff
-    timeout-minutes: 20
+    timeout-minutes: 30
```
eKoopmans, line 6: 😭

dlockhart: I know. The focus reset in input color tests caused them all the shift ever so slightly and get re-generated, which caused the tests to timeout... I don't think I'll ever _not_ be working on this.

eKoopmans (approved):
🎉 Let's go, big money no flakies! 🤞

.github/workflows/visual-diff.yml
dlockhart: I'm considering leaving this at 30 minutes, since it seems like any more than 5 diffs and the runs actually timeout instead of creating the PR with the changes.

PR #3408, opened 2023-03-16
Upgrade to Node 18

dlockhart:
This should be mergable while folks are still on Node 16, since the v3 lock-file is backwards compatible with v2. Core can also be a good testing bed for actions like update-package-lock, visual-diff and semantic-release.


svanherk (approved):
🎉

PR #3413, opened 2023-03-16
fix: remove forceFocusVisible from focus-trap

dlockhart:
This shouldn't have ever required `forceFocusVisible`, since the purpose of the focus-trap is for keyboard users and when you're tabbing around with the keyboard standard `:focus-visible` styles get used.

The only concern would be that `forceFocusVisible` searched for nested items to focus on, but in this case it's always being used after a call to `getNextFocusable` or `getPreviousFocusable` -- which will always return focusable things.


eKoopmans (approved):

dbatiste (approved):
Agree. I think this was originally used with the thinking that `:focus-visible` would not work when programatically applying focus, and that focus woud effectively be lost when moving focus into the trap.

components/dialog/test/dialog-helper.js
```diff
@@ -63,7 +63,7 @@ export async function reset(page, selector) {
 			}
 		});
 	});
-	return page.click('#open');
+	return focusWithKeyboard(page, '#open');
```
dlockhart, line 66: Part of dialog's "reset" code in between tests involved clicking on this button. But that put the browser back into mouse-mode instead of keyboard-mode, so the first test in each batch was now not getting the focus outline.

eKoopmans: This is also confusing because the `#open` button actually does nothing, the dialog is opened instead with `dialog.opened = true`. I was running into this when working on dialog and trying to debug some visual-diffs...

dlockhart: Haha yeah. The reason this came up now is that dialogs "focus first" logic looks in the content for something to focus on, but then falls back to the focus trap and focuses on it, which would then focus on the "X" button.

dlockhart: lol oh no... now there's a bunch of diffs because the open button is in them and it now has focus.

dlockhart:
This now has more visual-diff updates than I was expecting, but I think these make things more consistent overall. Let me know if anyone disagrees.

components/dialog/test/dialog-helper.js
eKoopmans: I'd be tempted to say this would be a bit cleaner as `page.$eval(selector, (dialog) => dialog.focus());` - that way it doesn't change the state either way. But it should have the same result as long as nothing else is switching to mouse mode.

dbatiste: >This is also confusing because the #open button actually does nothing, the dialog is opened instead with dialog.opened = true. 

I think there are at least a couple tests that open and close the dialog, showing that the opener receives the focus upon closing. That might be why some of this is done with an actual opener element rather than just setting the property.

dbatiste (approved):

components/dialog/test/dialog-helper.js
dlockhart: I'm. not sure I follow Erik... this code is run in the `beforeEach` of every test to make sure any dialogs opened in the previous test are closed and the page is back and ready to open another dialog.

eKoopmans: Yep - it's not an issue, I'm just saying that `focus()` would be less opinionated than using `focusWithKeyboard()` which explicitly forces keyboard mode.

That said, the whole keyboard/mouse mode would ideally be reset to the default (keyboard) before every single visual-diff test anyway, to reduce side effects. So ignore me, it's just an observation 😄

eKoopmans (approved):

PR #3416, opened 2023-03-17
fix: remove forceFocusVisible

dlockhart:
This removes `forceFocusVisible` from the last usage in dialogs, and removes it entirely as something core supports. `tryApplyFocus` also goes away as it was using `forceFocusVisible` and was only used in 1 place in dialogs.

For dialogs specifically, we'd still like a focus ring when a dialog opens and focus gets placed on one of the buttons. To accomplish this, I've created a dialog-specific implementation in the dialog mixin, and used a `.force-dialog-focus-visible` CSS class to be very explicit about what we're doing.

Once this merges, I'll remove the `.focus-visible` CSS class from any other repos that were using it.

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -3,7 +3,7 @@ import '../../helpers/viewport-size.js';
 import { allowBodyScroll, preventBodyScroll } from '../backdrop/backdrop.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
-import { getComposedActiveElement, getFirstFocusableDescendant, getNextFocusable, isFocusable, isFocusVisibleApplied } from '../../helpers/focus.js';
+import { getComposedActiveElement, getFirstFocusableDescendant, getNextFocusable, isFocusable } from '../../helpers/focus.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { html } from 'lit';
@@ -417,8 +417,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			// edge case: no children were focused, try again after one redraw
 			const activeElement = getComposedActiveElement();
 			if (!activeElement
-			|| !isComposedAncestor(dialog, activeElement)
-			|| !isFocusVisibleApplied(activeElement)) {
+			|| !isComposedAncestor(dialog, activeElement)) {
 				// wait till the dialog is visible for Safari
 				requestAnimationFrame(() => this._focusInitial());
 			}
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -423,7 +423,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			const activeElement = getComposedActiveElement();
 			if (!activeElement
 			|| !isComposedAncestor(dialog, activeElement)
-			|| !isFocusVisibleApplied(activeElement)) {
+			|| (!isFocusVisibleApplied(activeElement) && !activeElement.classList.contains('force-dialog-focus-visible'))) {
 				// wait till the dialog is visible for Safari
 				requestAnimationFrame(() => this._focusInitial());
 			}
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -3,7 +3,7 @@ import '../../helpers/viewport-size.js';
 import { allowBodyScroll, preventBodyScroll } from '../backdrop/backdrop.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
-import { forceFocusVisible, getComposedActiveElement, getNextFocusable, isFocusVisibleApplied, tryApplyFocus } from '../../helpers/focus.js';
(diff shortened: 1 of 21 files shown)


components/dialog/dialog-mixin.js
```diff
@@ -231,12 +231,32 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 		if (this._opener && this._opener.focus) {
 			// wait for inactive focus trap
 			requestAnimationFrame(() => {
-				tryApplyFocus(this._opener);
+				let focusable = this._opener;
 				this._opener = null;
+				if (!isFocusable(focusable)) {
```
dlockhart, line 236: This essentially re-creates `tryApplyFocus`.

components/dialog/dialog-mixin.js
```diff
@@ -231,12 +231,32 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
 			});
 		}
 	}
 
+	_forceDialogFocusVisible(elem) {
+		if (!isFocusable(elem, false, false)) {
+			elem = getFirstFocusableDescendant(elem);
+		}
+		if (!elem) return;
+		if (elem.tagName === 'BUTTON') {
```
dlockhart, line 253: Theoretically this could end up putting the CSS class on random buttons that have no idea about it, but I think that's OK? I originally had it looking for `<d2l-button>`, but sometimes this got called with its inner `<button>` so it was too late to check for it.

dlockhart:
I am SHOCKED that there weren't visual-diff failures. What's even going on. 🤯

components/dialog/dialog-mixin.js
```diff
@@ -231,12 +231,32 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
...
+			elem.addEventListener('blur', () => elem.classList.remove('force-dialog-focus-visible'), { once: true });
+			elem.classList.add('force-dialog-focus-visible');
+		}
+		if (elem) elem.focus();
```
eKoopmans, line 257: Is this check necessary, given `if (!elem) return;`?

components/dialog/dialog-mixin.js
eKoopmans: I think this would be more readable to keep it as a separate method, `this._tryApplyFocus()`.

components/tag-list/demo/tag-list.html
```diff
@@ -82,7 +82,7 @@ <h2>Clearable Tag List</h2>
 							e.target.removeChild(tag);
 						});
 						// consumer must move focus someplace logical
-						forceFocusVisible(nextFocusable);
+						if (nextFocusable) nextFocusable.focus();
```
eKoopmans, line 85: Could just do `nextFocusable?.focus()`? Or are we hesitant to use fancy techniques in demos lest they get applied with bad `.browerslist`s? 😅

helpers/focus.js
eKoopmans, line 1: Should be able to remove the `.focus-visible` class check from `isFocusVisibleApplied` too.

eKoopmans (commented):
Incredible! I never dreamed it possible. Nice work Dave!

helpers/focus.js
eKoopmans: Ah, but we may want/need to add a check for the new `.force-dialog-focus-visible` class - I'm not sure exactly the interaction there in terms of how `isFocusVisibleApplied` is used in dialogs.

dlockhart: Good catch. I've removed it from `isFocusVisibleApplied` since it's really dialog-specific now, but since dialog does use this function I've moved the check for `.force-dialog-focus-visible` into that dialog code.

dbatiste:
I'm not entirely sure why we still want to force focus visible upon opening the dialog. Wouldn't we just want to let `:focus-visible` do its thing, and if someone is using a mouse then they won't see focus indicator, just like everything else?

dbatiste:
I see. I'll chime in there with my vote. 😄

dbatiste (approved):

dlockhart:
After some further discussion we've decided to not even support this in dialogs, which I think is the best outcome. I've removed it there as well.

components/dialog/dialog-mixin.js
```diff
@@ -3,7 +3,7 @@ import '../../helpers/viewport-size.js';
 import { allowBodyScroll, preventBodyScroll } from '../backdrop/backdrop.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
-import { forceFocusVisible, getComposedActiveElement, getNextFocusable, isFocusVisibleApplied, tryApplyFocus } from '../../helpers/focus.js';
+import { getComposedActiveElement, getFirstFocusableDescendant, getNextFocusable, isFocusable, isFocusVisibleApplied } from '../../helpers/focus.js';
```
dbatiste, line 6: Do you know if the `isFocusVisibleApplied` usage still needed?

dlockhart: Hmm. So when the dialog opens, it's still using `isFocusVisibleApplied` to check if the active element has focus-visible applied and if not, it makes another attempt at finding and focusing. That [was introduced here](https://github.com/BrightspaceUI/core/commit/a1eb5d3d1d162d27d0b161a8908eff30166c4fb8).

This is a little weird now though, since clicking with the mouse is going to result in that always being `false`, compared to using the keyboard.

IMO, the check on the line above that ensures the active element is _inside_ the dialog should be enough to convince us it's OK to focus on it?

dbatiste: That's what I'ma thinking too.

eKoopmans: Yeah I was hoping `isFocusVisibleApplied` could be removed with these changes, but I didn't fully understand the motivation behind it. It's also used in `animate.js` btw.

PR #3419, opened 2023-03-20
remove ResizeObserver from date and time inputs

dlockhart:
Draft for now while I do some more testing and thinking.

Attempting to fix some `ResizeObserver` loop limit errors that folks are seeing in their tests in Safari. There's a delicate balance between the resize observers in the date and time inputs that are watching their "hidden content" and the observer in the `<d2l-date-time-range-to>` that's observing its parent. They can easily trigger each other and cause an infinite loop.

Proposed fix: from initial testing, it doesn't look like the "hidden content" observer needs to be an observer at all. It's used to try to guess how wide the widest possible input could be based on the user's locale settings. After the first calculation, I can't see scenarios (other than the user switching languages, which is handled) where we'd need to recalculate.


components/inputs/input-time.js
```diff
@@ -270,19 +264,28 @@ class InputTime extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixin(
...
 			this.requestUpdate();
-			this.updateComplete.then(() => this._onResize(hiddenContent));
+			this.updateComplete.then(() => updateHiddenContentWidth());
 		});
 
 		await (document.fonts ? document.fonts.ready : Promise.resolve());
 
-		this._hiddenContentResizeObserver = new ResizeObserver(() => this._onResize(hiddenContent));
-		this._hiddenContentResizeObserver.observe(hiddenContent);
+		updateHiddenContentWidth();
```
svanherk, line 288: I'm wondering if this is calculating too early?  And the ResizeObserver was added so there'd be a delay, and it would run _next_ time we resize?

dlockhart: Thanks, the history on why it was introduced was something I wanted to investigate and test. Turns out that's exactly why -- when the components are `display: none` initially, the width returns as "auto" which can't be parsed. The observer fixes that.

I've updated it slightly so that if it can't get a width right away, THEN it adds the observer. This still fixes the tests in Insights but if things were initially hidden there could still be issues so I'm going to separately look into the other observers and see if they can be done a different way.

components/inputs/input-date.js
```diff
@@ -14,7 +14,6 @@ import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { LabelledMixin } from '../../mixins/labelled-mixin.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
-import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
```
svanherk, line 17: Do you remember why this was using the polyfill? Are we at the point where we can kill the polyfill everywhere?

components/inputs/input-date.js
```diff
@@ -195,26 +194,36 @@ class InputDate extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixin(
 
 		this._textInput = this.shadowRoot.querySelector('d2l-input-text');
 
+		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-date-hidden-text');
+		const updateHiddenContentWidth = () => {
```
svanherk, line 198: Minor: `tryUpdateHiddenContentWidth` might be a bit clearer, since it returns whether it did or not

components/inputs/input-date.js
svanherk: What happens if you try to use it Safari pre-13.1 without feature detection?  Does it blow up?

svanherk (approved):

components/inputs/input-date.js
dlockhart: Blows up with:
> ReferenceError: can't find variable: ResizeObserver

components/inputs/input-date.js
```diff
@@ -195,26 +194,36 @@ class InputDate extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixin(
 
 		this._textInput = this.shadowRoot.querySelector('d2l-input-text');
 
+		const hiddenContent = this.shadowRoot.querySelector('.d2l-input-date-hidden-text');
```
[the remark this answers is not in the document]
dlockhart: Yeah, just optimized like the time input was doing -- seems all good.

PR #3425, opened 2023-03-22
fix: adding polyfill back

dlockhart:
Forgot I'd removed this before restoring the `ResizeObserver` uses, and then didn't get Stacey's subtle hints about it in the review. Adding back!


svanherk (approved):
Thanks! 😄

PR #3426, opened 2023-03-22
[DE52324] Update CSS to allow _focusing attribute to be elevated to a…

jesseag:
… z-index of 10. Changed _tooltip-showing attribute to be elevated to a z-index of 20.

Rally ticket: https://rally1.rallydev.com/#/10153049633d/dashboard?detail=%2Fdefect%2F689632043385

Allow a `d2l-list-item` with a `_focusing` attribute to be elevated to a higher relative `z-index`. This ensures that the list item has a higher `z-index` than other list items. This means that opened dialogs (emoji editor, attributes and Delete Criterion Row) will be higher `z-index` than other elements and the grayed out background correctly grays out other list items and the rest of the page.

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -108,9 +108,12 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				display: none;
 			}
 
-			:host([_tooltip-showing]),
-			:host([_dropdown-open]) {
+			:host([_dropdown-open]),
+			:host([_focusing]) {
 				z-index: 10; /* must be greater than adjacent selected items (if this is increased, d2l-collapsible-panel must be updated too) */
+			}
+			:host([_tooltip-showing]) {
+				z-index: 20; /* must be greater than adjacent selected items (if this is increased, d2l-collapsible-panel must be updated too) */
 			}
 			:host([_fullscreen-within]) {
 				position: fixed; /* required for Safari */


components/list/list-item-mixin.js
```diff
@@ -108,10 +108,13 @@ export const ListItemMixin = superclass => class extends composeMixins(
...
 
-			:host([_tooltip-showing]),
-			:host([_dropdown-open]) {
+			:host([_dropdown-open]),
+			:host([_focusing]) {
 				z-index: 10; /* must be greater than adjacent selected items (if this is increased, d2l-collapsible-panel must be updated too) */
 			}
+			:host([_tooltip-showing]) {
+				z-index: 20; /* must be greater than adjacent selected items (if this is increased, d2l-collapsible-panel must be updated too) */
+			}
```
jesseag, line 117: I choose `z-index: 20` here but this value is relative. It could really be anything but it just needs to be `> 10` which is the `_focusing` `z-index` above.

Here is a video of the bug showing what happens if the `_tooltip-showing` index is not higher than 10. You can see that if a lower list item is focused then the previous list item's tooltip will get hidden behind the focused list item.

tooltip_bug

dbatiste: @jesseag I see. The tooltip case makes sense. We don't have the same issue with dropdowns or dialogs opened from the list-item because adjacent items cannot become focused while they are open. I think I would be inclined to tighten the range. If anything comes along that needs to insert itself in the range, I think we/I want to know about it.

@semibran yeah, that's a good suggestion. We've briefly considered introducing some type of `z-index` manager a couple years ago and I generated a list of our usage - it is like the wild wild west. I think a set of CSS variables might help align on, and manage the values, but I think we'd want to take a broader fresh look at `z-index` usage. Probably not a mixin though, since we'd want to be able to use those values in non-component code (i.e. legacy/MVC LMS) as well. I think if we do that, we can do it separate from this PR.

jesseag:
@dbatiste 

I also noticed another bug where the 1st litem is selected and then the `_tooltip-showing` attribute is set incorrectly for the 2nd list item that has not been yet been selected by keyboard actions. 

I noticed that this issue does not happen if we reproduce it when selecting the 2nd list-item and displaying it's tooltips. The 3rd list item does not set it's `_tooltip-showing` attribute to true and the bug is not reproducible.

I reverted all my code and tested again to ensure that this was not an issue caused by my changes. The issue is pre-existing prior to my changes. Here is a video of the issue.

list_item_2_tooltip_getting_set_incorrectly

dbatiste (approved):
LGTM!

PR #3429, opened 2023-03-23
Fix d2l-tabs extension slot stacking.

dbatiste:
This PR fixes the stacking of elements in `d2l-tabs` ext `slot` due to the stacking context that is/was created by the `transform` that is used for animation when showing/hiding the tabs bar (going from 1 to many or many to 1 tabs).

**Before:**
image

**After:**
image

**Extra background...** we got bitten by [this transform behaviour again](https://developer.mozilla.org/en-US/docs/Web/CSS/transform):
> If the property has a value different than none, a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) will be created. In that case, the element will act as a [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block) for any position: fixed; or position: absolute; elements that it contains.

### DIFF
--- a/components/tabs/demo/tabs.html
+++ b/components/tabs/demo/tabs.html
@@ -21,7 +21,7 @@
 
 			<h2>Tabs</h2>
 
-			<d2l-demo-snippet id="first">
+			<d2l-demo-snippet>
 				<template>
 					<d2l-tabs>
 						<d2l-tab-panel text="Biology">Tab content for Biology</d2l-tab-panel>
--- a/components/tabs/demo/tabs.html
+++ b/components/tabs/demo/tabs.html
@@ -7,6 +7,10 @@
 		<script type="module">
 			import '../../demo/demo-page.js';
 			import '../../button/button-subtle.js';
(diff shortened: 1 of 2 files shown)


components/tabs/demo/tabs.html
dbatiste, line 1: The add/remove buttons were just moved down to a different demo-snippet to make it easier to test this case. The JavaScript was updated to be a little more flexible to no longer assume it is interacting with the first `d2l-tabs` on the demo page.

components/tabs/tabs.js
```diff
@@ -96,7 +96,7 @@ class Tabs extends LocalizeCoreElement(ArrowKeysMixin(RtlMixin(LitElement))) {
 				display: flex;
 				max-height: 60px;
 				opacity: 1;
-				transform: translateY(0);
+				transform: none;
```
dbatiste, line 99: The animation appears to be exactly the same using either of these values.

bearfriend (approved):

dlockhart (approved):

PR #3434, opened 2023-03-24
fix: Flaky list-controls visual diff

eKoopmans:
This should fix the common flaky test `d2l-list-controls-sticky-top` (e.g. [here](https://github.com/BrightspaceUI/core/pull/3433/files)).

The cause of the flake is that an underline appears whenever the controls are offscreen, e.g. if the tests have scrolled down for some other test. This fix always scrolls it into view before capturing.

I also had to move the mouse because tests that use `hover()` move the mouse on the screen and then leave it there, which was causing hover styles to appear on this list!

### DIFF
--- a/components/list/test/list.visual-diff.js
+++ b/components/list/test/list.visual-diff.js
@@ -46,11 +46,9 @@ describe('d2l-list', () => {
 	};
 
 	const scrollTo = (selector, y) => {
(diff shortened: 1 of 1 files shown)


oelhanafey (approved):
Nice! This flake was annoying 😆

PR #3436, opened 2023-03-25
Fix d2l-input-checkbox and input-radio-styles label layout

dbatiste:
[DE52712](https://rally1.rallydev.com/#/?detail=/defect/694394016671&fdp=true)

[Related Slack thread](https://d2l.slack.com/archives/C0PHG3QB0/p1679745600486689)

This PR fixes the `d2l-input-checkbox` and `input-radio-styles` layout so that label contents do not overflow their containers. It also updates the visual-diff to capture these cases. The existing demo has a multi-line example, but it was not obvious that it was overflowing - inspecting the DOM reveals that it was. This is problematic in cases where these elements are contained in elements that hide overflow, like dropdowns.

All of the following examples are in a `200px` wide containing block.

**Before checkbox:**
image

**After checkbox:**
image

**Before radio:**
image

**After radio:**
image


bearfriend (approved):

components/inputs/test/input-checkbox.visual-diff.html
```diff
@@ -32,17 +32,27 @@
 			<d2l-input-checkbox id="wc-disabled-indeterminate" indeterminate disabled>Checkbox (indeterminate, disabled)</d2l-input-checkbox>
 		</div>
 		<div class="visual-diff">
-			<d2l-input-checkbox id="wc-multiline" style="width: 200px;">
+			<d2l-input-checkbox id="wc-multiline" style="overflow: hidden; width: 200px;">
```
dbatiste, line 35: Added these `overflow: hidden;` to make it more obvious if there is a regression. Without this, it's not particular obvious when there is a problem because our visual diff captures a slightly larger area.

dlockhart (approved):
👍

PR #3453, opened 2023-03-30
fix: Safari list drag-and-drop

eKoopmans:
This PR fixes a defect where list drag-and-drop was broken in Safari. This was a regression introduced [here](https://github.com/BrightspaceUI/core/commit/e24344da7f99b89619d4b1ec32a71f4b4118679d) as part of #3046 (expand/collapse list). Turns out Safari is a bit different in how it triggers `'dragleave'`, which was causing the drop-target to continually be rendered and derendered.

The solution is to clean up where the `_draggingOver` state is being updated - we've moved it inside of the `DragState` class, and also now trigger a cleanup whenever the drop-target becomes invalid (e.g. when an item is hovering over itself).

Joint effort with @svanherk @oelhanafey @dbatiste. Tested on Chrome (Windows + Mac) and Safari.

Rally: https://rally1.rallydev.com/#/?detail=/defect/694750082045&fdp=true

### DIFF
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -738,7 +738,6 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 
 	_onHostDragEnter(e) {
 		const dragState = getDragState();
-		if (this === dragState.dragTarget) return;
 
 		// check if any of the drag targets are ancestors of the drop target
 		const invalidDropTarget = dragState.dragTargets.find(dragTarget => {
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -105,6 +105,7 @@ class DragState {
 
 	_cleanUpOnLeave() {
 		if (!this._activeDropTarget) return;
+		this._activeDropTarget._draggingOver = false;
 		this._activeDropTarget._dropLocation = dropLocation.void;
 		this._activeDropTarget._inTopArea = false;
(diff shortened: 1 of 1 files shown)


components/list/list-item-drag-drop-mixin.js
```diff
@@ -105,6 +105,7 @@ class DragState {
 
 	_cleanUpOnLeave() {
 		if (!this._activeDropTarget) return;
+		this._activeDropTarget._draggingOver = false;
```
eKoopmans, line 108: I moved this a bit from where we had it on our call (in `setActiveDropTarget`). Seemed more appropriate here!

svanherk: Yeah this spot seems perfect!

oelhanafey:
We could probably remove [this line](https://github.com/BrightspaceUI/core/blob/83d75c1fb3b1259e2021dc73907762a6b420f95e/components/list/list-item-drag-drop-mixin.js#L741) here as well that isn't doing anything anymore since dragTarget was removed [here](https://github.com/BrightspaceUI/core/commit/5246f337e5b966a3b2da095a65e175229efc0ce3#diff-d9fcf498e8eeb8f354d093d76b0845f7a0eec01ca8e45aa3ed597cf64d8e0f29L49).

The case where `this == dragTarget` should still be being checked by the lines below that iterates through the list of `dragTargets` and checks if it is a descendent of the `dropTarget`

svanherk:
I did a quick sanity check on Firefox and it's working as expected there

svanherk (approved):
This looks like what I expected!

PR #3456, opened 2023-03-30
fix: Backport Safari list drag-and-drop fix for LCS version

svanherk:
This backports the fix to the `core` version [LCS needs for the customization](https://github.com/Brightspace/lcs-profession-map-admin/blob/1ece2aff17bf1679ef76513bf5bba704de64328f/package-lock.json#L1267) of the client that raised the issue.

We don't need to do anything with BSI or the LMS for this one - just update the thread when this is merged and released so they can rebuild and manually deploy `lcs-profession-map-admin`.

They'll be getting [a few other patches](https://github.com/BrightspaceUI/core/compare/v2.83.0..v2.83.3) as part of this, but there's not much we can do about that.  Nothing jumped out at me as particularly worrisome. The updates to `filter` are self-contained, and the updates to `list` are a bug fix: https://github.com/BrightspaceUI/core/pull/3234 (FYI @dbatiste)


dlockhart (approved):
Visual-diff free, nice!

eKoopmans (approved):

PR #3459, opened 2023-03-30
feat: Support subscribers on pageable-mixin

eKoopmans:
This PR converts PageableMixin to use SubscriberControllers internally. This will support being able to subscribe to pageable info from a non-pager (e.g. for the "selection summary" component in list/table controls).

There are **several** breaking changes here in terms of PageableMixin, anything that uses it (i.e. `d2l-list`), and `pager-load-more`:

- `item-count` needs to be moved from `d2l-pager-load-more` to its parent pageable
- `_getItemsShowingCount` has been renamed to `_getItemShowingCount` (removed an s) and is no longer async
- `_updatePagerCount` has been changed and renamed to `_updateItemShowingCount`
- `_getLastItemIndex` no longer needs to be defined by the consumer

I'll track what changes need to be made outside of core on Rally, and get PRs ready for all of them so we can synchronize them as closely as possible.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/694042692959&fdp=true


components/list/list.js
```diff
@@ -77,7 +75,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 
 	connectedCallback() {
 		super.connectedCallback();
-		this.addEventListener('d2l-list-items-showing-count-change', this._handleListItemsShowingCountChange);
+		this.addEventListener('d2l-list-item-showing-count-change', this._handleListItemShowingCountChange);
```
eKoopmans, line 78: There are no external consumers of this event or method, so renaming should be safe 😄.

components/list/list.js
```diff
@@ -64,8 +64,6 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		this.dragMultiple = false;
 		this.extendSeparators = false;
 		this.grid = false;
-		this._itemsShowingCount = 0;
-		this._itemsShowingTotalCount = 0;
```
eKoopmans, line 68: I talked with @dbatiste about the difference between these two variables, and we agreed that the distinction is no longer necessary. `_itemsShowingTotalCount` originally had a more expensive calculation that recursed through nested lists, but more recently it was decided that we didn't want to show that detail, and only want the "showing count" to represent top-level items.

The expensive calcs were [removed as part of expand/collapse](https://github.com/BrightspaceUI/core/pull/3046/files#diff-afabf94fa43a55055dff6b1ca27abd24b629f017ad9dee5d4c48f123e80a0533L194), so following that change the two variables were essentially identical.

components/paging/pageable-mixin.js
```diff
@@ -1,35 +1,67 @@
...
 
-	_updatePagerCount(count, slot) {
-		if (!slot) slot = this.shadowRoot.querySelector('slot[name="pager"]');
-		const elements = slot.assignedElements({ flatten: true });
-		if (elements.length > 0) {
-			elements[0].itemShowingCount = count;
-		}
+	_updateItemShowingCount() {
+		this._itemShowingCount = this._getItemShowingCount();
+	}
```
eKoopmans, line 57: This essentially replaces `_updatePagerCount`, and is what consumers need to call any time the count changes. It's a bit different in that the subscribers will be updated in the next update cycle (via the `updated` check), rather than immediately.

If we were very concerned about that immediacy I could change `_itemShowingCount` to an internal property and explicitly call `updateSubscribers` from here - I'm okay with either approach, but the current way protects us a bit more in case of consumers explicitly setting `_itemShowingCount` for any reason.

components/paging/pageable-subscriber-mixin.js
```diff
@@ -0,0 +1,33 @@
...
+
+		this._pageableInfo = { itemCount: -1, itemShowingCount: 0 };
+		this._pageableEventSubscriber = new EventSubscriberController(this, 'pageable');
+		this._pageableIdSubscriber = new IdSubscriberController(this, 'pageable', { idPropertyName: 'pageableFor' });
+	}
+
+	async getUpdateComplete() {
+		await super.getUpdateComplete();
+		await (this.pageableFor ? this._pageableIdSubscriber._updateComplete : this._pageableEventSubscriber._updateComplete);
+	}
```
eKoopmans, line 27: A bit of magic to make the tests work nicely without having to add an explicit `requestAnimationFrame` in them. Should also benefit anything else that's waiting on the subscriber.

dlockhart (approved):

components/paging/pageable-subscriber-mixin.js
dbatiste: What does `_updateComplete` represent here?

controllers/subscriber/subscriberControllers.js
```diff
@@ -111,7 +118,11 @@ export class EventSubscriberController extends BaseSubscriber {
 
 	hostConnected() {
 		// delay subscription otherwise import/upgrade order can cause selection mixin to miss event
-		requestAnimationFrame(() => this._subscribe());
+		this._updateComplete = promiseWithResolve();
```
dbatiste, line 121: What's the rationale for `promiseWithResolve`? Can we just do this?
```javascript
this._updateComplete = new Promise(resolve => {
    this._subscribe();
    resolve();
});
```

eKoopmans: Umm... yes 😅. Not sure why I thought that wasn't an option earlier, thanks!

components/paging/pageable-subscriber-mixin.js
dbatiste: The pageable subscriber would also have `updateComplete`. I am wondering if we should name this differently to avoid any confusion. Thoughts?

eKoopmans: Are you thinking renaming the subscriber controllers, like:

```js
await (this.pageableFor ? this._pageableIdSubscriberController._updateComplete : this._pageableEventSubscriberController._updateComplete);
```

eKoopmans: Another option to consider is renaming this whole mixin as `PageableObserverMixin`, which would be consistent with `SelectionObserverMixin`. Then "subscriber" refers to the controller and "observer" refers to the mixin/component?

PR #3460, opened 2023-03-31
Fix list-item drag handle target size for WCAG 2.2.

dbatiste:
[DE52760](https://rally1.rallydev.com/#/?detail=/defect/694748658255&fdp=true)

This PR fixes the list-item's drag handle target size for click. The visual dimensions and the button remain the same, but the click target size is updated to meet the WCAG 2.2 criteria. This is accomplished by rendering clickable targets over top of the icons that extend slightly into the margins.

**Before (focused):**
image

**Before (focused, showing click targets sizes - 18px by 18px):**
image

**After (focused):**
image

**After (focused, showing click target sizes - 26px by 22px):**
with the 2px gap and 22px tall target, we achieve the required 24px distance from furthest edge to closest edge
image


dlockhart (approved):
LGTM

dbatiste:
Talked to Jeff and we decided to widen the click target (updated the screenshot in the description).

PR #3463, opened 2023-03-31
feat: upgrade intl-messageformat

dlockhart:
We'd like to use the latest `intl-messageformat` library in order to leverage version 8's ability to do rich formatting. The release of this PR into BSI will need to be coordinated with [a similar change in `@brightspace-ui/localize-behavior`](https://github.com/BrightspaceUI/localize-behavior/pull/38). No [other BSI things](https://github.com/search?type=code&q=saved%3A%22D2L+Code%22+%22intl-messageformat%22+path%3A%2F%5Epackage.json%2F&saved_searches=%5B%7B%22name%22%3A%22D2L+Code%22%2C%22query%22%3A%22org%3ABrightspace+OR+org%3ABrightspaceUI+OR+org%3ABrightspaceHypermediaComponents+OR+org%3ABrightspaceUILabs%22%7D%5D&expanded_query=org%3ABrightspace+org%3ABrightspaceUI+org%3ABrightspaceHypermediaComponents+org%3ABrightspaceUILabs+%22intl-messageformat%22+path%3A%2F%5Epackage.json%2F) depend on it directly.

Looking through [the CHANGELOG](https://github.com/formatjs/formatjs/blob/main/packages/intl-messageformat/CHANGELOG.md), I don't see any breaking changes that are particularly concerning in versions 8, 9 or 10.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,4 +6,3 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "@formatjs/intl-pluralrules"
-      - dependency-name: "intl-messageformat"
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@formatjs/intl-pluralrules": "^1",
         "@open-wc/dedupe-mixin": "^1",
         "ifrau": "^0.40",
-        "intl-messageformat": "^7",
+        "intl-messageformat": "^10",
         "lit": "^2",
         "prismjs": "^1",
         "resize-observer-polyfill": "^1"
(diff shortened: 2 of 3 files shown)


.github/dependabot.yml
```diff
@@ -6,4 +6,3 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "@formatjs/intl-pluralrules"
```
dlockhart, line 8: I tried to upgrade this as well but ran into all kinds of breaking changes and ESM issues. It's not a required change, so delaying that for now.

eKoopmans (approved):
🎉

PR #3468, opened 2023-04-03
US150791 - Fix flaky list visual-diff

svanherk:
Ok, I think this is stable.  I've run just the list tests over 20 times on CI, but the real test will be if we see it in the wild again.

This PR has two pieces:
1. Setting `captureBeyondViewport` to `false` for the sticky tests. This alone was enough to fix it for me locally - I can see in `headless` mode that this stops the `screenshotAndCompare` call from causing an adjustment to the scroll that may or may not be fixed in time for the screenshot.  In CI though, it just seems to reduce the flake, which makes fix 2 really important.
2. Scrolling to a different location, requesting an animation frame, and then scrolling to the desired location. I'm seeing that without this, retries are currently broken - calling `scrollTo(..., 0)` actually does nothing if the browser _thinks_ it's already there, even if it's showing the sticky top.

I think retries being broken is why this test is particularly flaky, so hoping this PR _at least_ helps, if not removes the flake entirely.


dbatiste (approved):
LGTM!

dlockhart (approved):
Nice!

oelhanafey (approved):

components/list/test/list.visual-diff.js
```diff
@@ -147,8 +148,14 @@ describe('d2l-list', () => {
 			{ name: 'some selected', selector: '#selectableSomeSelectedControls' },
 			{ name: 'all selected', selector: '#selectableAllSelectedControls' },
 			{ name: 'all selected pages', selector: '#selectableAllSelectedControlsPages' },
-			{ name: 'sticky top', selector: '#stickyControls', action: () => scrollTo('#stickyControls > div', 0) },
-			{ name: 'sticky scrolled', selector: '#stickyControls', action: () => scrollTo('#stickyControls > div', 45) }
+			{ name: 'sticky top', selector: '#stickyControls', action: async() => {
+				await scrollTo('#stickyControls > div', 45);
+				await scrollTo('#stickyControls > div', 0);
```
eKoopmans, line 153: I'm pretty confused about why this would help... are you sure this isn't essentially just increasing the timeout so that the `scrollIntoView` has properly triggered? The number only changes the vertical scrolling within the div, and I really can't see the browser getting confused about that part.

Did you try this same fix with `0` both times?

eKoopmans (commented):
I'm super excited about this flake being fixed, thanks Stacey!

components/list/test/list.visual-diff.js
eKoopmans: Thanks Stacey! We just caught up, Stacey showed me these failures in action. Scrolling the container down and back up definitely fixes them - which tells me there might be something wrong in the sticky logic itself! I'll have a look and see if I can find out why.

PR #3470, opened 2023-04-03
move into localize directory

dlockhart:
Not sure how readable the diff is going to be here, but this shows what moving all the various localize things into their own `localize` directory under `mixins` would look like. It also moves the helpers right inline as well.

I didn't see a point in keeping separate files for the helpers, static mixin and dynamic mixin. The static mixin is going away once the one usage of it is removed. The dynamic mixin is what everyone _should_ be using, and it needs the normal mixin anyway. Having the `generateLink`, etc. in there also seems handy so folks can just import everything they need from a single entry point, but curious what others think.

Since we actually just want people to be using the dynamic mixin only, this also renames `LocalizeDynamicMixin` -> `LocalizeMixin` and `LocalizeMixin` -> `_LocalizeMixinBase` in the new export structure. The old one keeps the old names.

Lastly, the dynamic mixin has been updated to handle an `importFunc` that's synchronous.


bearfriend:
In my original stab at this, I took the opportunity of the new directory to make `LocalizeDynamicMixin` export as `LocalizeMixin` and [modified](https://github.com/BrightspaceUI/core/compare/l10n-markup...reorganize-localize?expand=1#diff-51f475848540a16111c8770a6cd3e929439fd2a1cbbf08f4e2496bf4504ec4dcR18-R19) it to handle synchronous fetching.

Not sure I care much either way, but do we want to consider that or some similar reworking while we have the opportunity here?

dlockhart:
I really like the idea of renaming it, since it should really be the only thing folks need to use. I'll make that change.

Sorry, what would modifying it to do synchronous fetching do? Oh like because of the optional chaining in the `catch` call the provided function doesn't have to return a Promise?

bearfriend:
> I really like the idea of renaming it, since it should really be the only thing folks need to use. I'll make that change.
> 
> Sorry, what would modifying it to do synchronous fetching do? Oh like because of the optional chaining in the `catch` call the provided function doesn't have to return a Promise?

Right, `await {...}` will just give you the static resources object, but it obviously doesn't have a `catch` method. So if we separate that out, people can use this regardless of how they get their lang resources and benefit from the OSLO simplification etc.

bearfriend:
> So if we separate that out...

Wait, that implementation won't work. Nothing will have catch after it's awaited.

mixins/localize/localize-mixin.js
```diff
@@ -0,0 +1,288 @@
...
+	static async getLocalizeResources(langs, { importFunc, osloCollection, useBrowserLangs }) {
+
+		// in dev, don't request unsupported langpacks
+		if (!importFunc.toString().includes('switch') && !useBrowserLangs) {
+			langs = langs.filter(lang => supportedLangpacks.includes(lang));
+		}
+
+		for (const lang of [...langs, fallbackLang]) {
+
+			const resources = await importFunc(lang).catch(() => {});
```
bearfriend, line 207: I think this is what we want
```suggestion
			const resources = await Promise.resolve(importFunc(lang)).catch(() => {})
```

mixins/localize-mixin.js
```diff
@@ -1,189 +1 @@
...
-		return this.constructor['getLocalizeResources'] !== undefined;
-	}
-
-	_onResourcesChange() {
-		/** @ignore */
-		this.dispatchEvent(new CustomEvent('d2l-localize-resources-change'));
-	}
-
-});
+export { _LocalizeMixinBase as LocalizeMixin } from './localize/localize-mixin.js';
```
dlockhart, line 1: Ideally [all the usages of this](https://github.com/search?type=code&q=saved%3A%22D2L+Code%22+%22import+%7B+LocalizeMixin+%7D+from+%27%40brightspace-ui%2Fcore%2Fmixins%2Flocalize-mixin.js%27%22&saved_searches=%5B%7B%22name%22%3A%22D2L+Code%22%2C%22query%22%3A%22org%3ABrightspace+OR+org%3ABrightspaceUI+OR+org%3ABrightspaceHypermediaComponents+OR+org%3ABrightspaceUILabs%22%7D%5D&expanded_query=org%3ABrightspace+org%3ABrightspaceUI+org%3ABrightspaceHypermediaComponents+org%3ABrightspaceUILabs+%22import+%7B+LocalizeMixin+%7D+from+%27%40brightspace-ui%2Fcore%2Fmixins%2Flocalize-mixin.js%27%22) would switch over to the dynamic mixin.

mixins/localize-static-mixin.js
```diff
@@ -1,28 +1 @@
...
-			resources: resolvedResources
-		};
-	}
-
-	static get resources() {
-		return { 'en': {} };
-	}
-
-};
+export { LocalizeStaticMixin } from './localize/localize-mixin.js';
```
dlockhart, line 1: There's only 1 remaining usage of this that I'm hoping to get rid of shortly, at which point the static mixin can go away.

mixins/localize-dynamic-mixin.js
```diff
@@ -1,42 +1 @@
...
-			}
-		}
-	}
-
-	static get localizeConfig() {
-		return {};
-	}
-
-};
+export { LocalizeMixin as LocalizeDynamicMixin } from './localize/localize-mixin.js';
```
eKoopmans, line 1: Is `LocalizeDynamicMixin` now obsolete? It looks like `getLocalizeResources` has now been moved into `LocalizeMixin` itself? This is making my head hurt a bit... 😅

eKoopmans: Oops, looks like you answered this question in the description, my bad. I guess it's been a long day! 🙃 

My main concern is this is getting to be a lot of changes all wrapped up under the banner of "add localizeHTML method", it'll be pretty hard to see what actually changed...

bearfriend: Normally I'd agree with that concern, but in this case I think it's all just very intertwined, and while we could do a bunch to make the history very clean I'm not sure it's worth it. The history will still be here via all these branches/PRs if we need to comb through it.

No objection if we do want to do that though.

dlockhart: Yeah I agree, I'm not very comfortable with it but also at this point they are so closely related that doing them one at a time is going to add a lot of effort to re-sort through all the conflicts. When/if this does merge to `main` I do want to go through the other mixins and apply this same treatment.

eKoopmans: Haha okay... yolo!

eKoopmans (approved):

PR #3478, opened 2023-04-04
Dispatch d2l-list-item-selected when Enter is pressed

oelhanafey:
[DE52824](https://rally1.rallydev.com/#/?detail=/defect/695920026421&fdp=true)

This PR fixes an issue in the `list-item-checkbox-mixin` where the pressing the ENTER key on the action area was changing the selection value but not dispatching the `d2l-list-item-selected` event. Added a unit test to cover this case.


dbatiste:
Hey Oscar, do you know what causes this issue? In the Rally ticket, it sounds like this was working in a previous release.

oelhanafey:
@dbatiste it was from this PR: https://github.com/BrightspaceUI/core/pull/3333

PR #3480, opened 2023-04-04
Fix z-indices of upper layer components

oelhanafey:
[DE52725](https://rally1.rallydev.com/#/?detail=/defect/694583527535&fdp=true)

This PR fixes a defect where dropdowns with `no-auto-close` would stay open when a dialog opened and since their z-indices were higher than `backdrop` they would not be faded with the rest of the background.

`d2l-backdrop` has `z-index: 999` which is shared by many components. I'm proposing to change the upper layer of our web components to `z-index: 998` and then have the backdrop layer 1 higher. If we need to put something higher than the upper level in the future, then we can push those components down a layer and still use `z-index: 998` as the upper layer.

Proposed layers:
```
...
upperLayer - 2: 996
upperLayer - 1: 997
upperLayer: 998
backdropLayer: 999
dialogLayer: 1000
tooltipLayer: 1001
```

Alternatively, if it is more understandable, we could push everything above and including the backdrop up a layer, so the upper layer is `z-index: 999` but I'm worried that that might put some components outside of core in the backdrop that previously weren't.

### DIFF
--- a/components/button/floating-buttons.js
+++ b/components/button/floating-buttons.js
@@ -42,7 +42,7 @@ class FloatingButtons extends RtlMixin(LitElement) {
 				position: -webkit-sticky;
 				position: sticky;
 				right: 0;
-				z-index: 999;
+				z-index: 997;
 			}
 
 			:host([_floating][always-float]) {
--- a/components/dropdown/dropdown-content-styles.js
+++ b/components/dropdown/dropdown-content-styles.js
@@ -244,7 +244,7 @@ export const dropdownContentStyles = css`
 		height: 100vh;
 		position: fixed;
 		top: 0;
-		z-index: 998;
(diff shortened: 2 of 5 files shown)


oelhanafey:
I've tested the changes on the html-editor and I haven't found any regressions or issues with stacking.

Though I did come across this weird case that was present even before my change:
Screen Shot 2023-04-06 at 1 08 42 PM
This happens in the "Inside Dialog" demo in fullscreen mode when you open the accessibility checker. Moving the dialog fixes it. FYI @grant-cleary

grant-cleary:
Yeah, I think there are already some known issues with the z-index of fullscreen mode when rendered in a dialog. There's like three layers of stacking contexts in there and it's just a mess all around.

PR #3485, opened 2023-04-05
Refactor move button

dbatiste:
This PR factors out the keyboard button from `d2l-list-item-drag-handle` into a new `d2l-button-move` component. The new component will also be used by the html-editor's resize button in order to meet WCAG 2.2 - 2.5.7 Drag Movements criteria. The original behaviour is maintained - for now, only up/down arrows are rendered in vertical layout while events will be dispatched for moving in any direction, but visual layout may be updated in the future. The new component looks like this:

image


components/button/button-icon.js
```diff
@@ -167,7 +167,6 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				aria-haspopup="${ifDefined(this.ariaHaspopup)}"
 				aria-label="${this.ariaLabel ? this.ariaLabel : ifDefined(this.text)}"
 				?autofocus="${this.autofocus}"
-				class="d2l-label-text"
```
dbatiste, line 170: Clean-up - It appears this class was copy and pasted, but serves no purpose for `d2l-button-icon`.

components/button/button-move.js
```diff
@@ -0,0 +1,211 @@
...
+				?autofocus="${this.autofocus}"
+				?disabled="${this.disabled && !this.disabledTooltip}"
+				id="${this._buttonId}"
+				@keydown="${this._handleKeydown}"
+				name="${ifDefined(this.name)}"
+				title="${ifDefined(this.text)}"
+				type="${this._getType()}">
+				<d2l-icon icon="tier1:arrow-toggle-up" class="up-icon"></d2l-icon>
+				<d2l-icon icon="tier1:arrow-toggle-down" class="down-icon"></d2l-icon>
+				<div class="up-layer" @click="${this._handleUpClick}"></div>
```
dbatiste, line 154: As mentioned in previous PR, these `div` layers are rendered on top of the icons and extend into the margin where the focus outline is displayed, so that it meets the minimum WCAG 2.2 - 2.5.8 Target Size criteria.

components/list/list-item-drag-handle.js
```diff
@@ -266,49 +272,19 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
-			return;
-		}
-		this._keyboardActive = false;
-		this._dispatchAction(dragActions.save);
+		e.preventDefault();
 		e.stopPropagation();
-	}
+		this._doAction(action);
 
-	_onInactiveKeyboard(e) {
```
dbatiste, line 298: This just got split up into `_onDraggerButtonClick` and `_onDraggerButtonKeydown` when `_onInactiveKeyboard` was renamed for clarity.

components/list/list-item-drag-handle.js
```diff
@@ -212,52 +167,103 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 				break;
+			default:
+				return;
+		}
+
+		this._doAction(action);
+
+	}
+
+	async _onMoveButtonKeydown(e) {
```
dbatiste, line 253: While the above `_onMoveButtonAction` handler maps the move event `moveActions` to `dragActions` that list and list consumers understand, the drag-handle still needs to handle key events for tab, escape, enter, and space as it previously did.

components/list/list-item-drag-handle.js
```diff
@@ -192,9 +142,14 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
 		this._keyboardActive = true;
 	}
 
-	focus() {
-		const node = getFirstFocusableDescendant(this);
-		if (node) node.focus();
+	async focus() {
```
dlockhart, line 145: You could implement `FocusMixin` here I think and it'll handle a bunch of this for you.

components/list/list-item-drag-handle.js
```diff
@@ -212,52 +167,103 @@ class ListItemDragHandle extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
 				break;
-			case keyCodes.DOWN:
+			case moveActions.down:
 				this._movingElement = true;
 				action = dragActions.down;
 				break;
-			case keyCodes.HOME:
+			case moveActions.right:
+				this._movingElement = true;
+				action = (this.dir === 'rtl' ? dragActions.unnest : dragActions.nest);
```
dlockhart, line 223: Would it make sense for `button-move` to handle the RTL flip for consumers? If so, maybe we use directional agnostic terms like start/end or back/forward.

dbatiste: I don't _think_ so... the `d2l-button-move`'s action detail reflects the direction the user selected exactly, and the consumer may or may not want to invert the operation. In the case of list-item, we do want it inverted. However if this were being used to move a floating container on the page, we'd want the direction it's moved to match the arrow keys being pressed, so we wouldn't want it inverted in that case.

dlockhart: Right. I guess if we ended up with a lot more consumers it could end up being an option they pass in.

components/button/button-move.js
```diff
@@ -0,0 +1,211 @@
...
+	rootHome: 'root-home',
+	home: 'home',
+	rootEnd: 'root-end',
+	end: 'end'
+});
+
+/**
+ * A button component that provides a move action via a single button.
+ */
+class ButtonMove extends ButtonMixin(RtlMixin(LitElement)) {
```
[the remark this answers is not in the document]
dbatiste: Yeah I can probably remove that. There might be a little bit of click/disabled logic I'd need to copy over.

PR #3489, opened 2023-04-05
Put mixins in their place

dlockhart:
This moves the remaining mixins into their own directories within the `mixins` directory. It leaves the original files there with "redirects" so existing places importing them won't break.


mixins/arrow-keys-mixin.js
dlockhart, line 1: Grr... despite doing this in 2 commits -- one to move, then one to make changes -- it appears as though git isn't preserving history. Any ideas?

svanherk: Maybe you'll have to just do the move, merge (but don't release), then put up a second PR to add back the "new" files with the redirects?

eKoopmans: I think this is more a matter of how Github (and/or git) visualizes the diffs between two arbitrary commits. The commit-by-commit history was still there - but separate PRs works too 😄.

PR #3492, opened 2023-04-05
Put mixins in their place (part 2)

dlockhart:
Looks like we need to do this in two parts in order to preserve history. I'll merge part 1 but not release, then merge part 2 shortly after.

Part 1: move the mixin files into their own directories and adjust all paths within core
Part 2: restore the original mixin files but just have them export from the new location

### DIFF
--- a/mixins/arrow-keys-mixin.js
+++ b/mixins/arrow-keys-mixin.js
@@ -1,0 +1,1 @@
+export { ArrowKeysMixin } from './arrow-keys/arrow-keys-mixin.js';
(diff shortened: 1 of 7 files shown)


eKoopmans (approved):
No pointer file for `redirect-mixin`, but looks like that's because there's [no external consumers](https://search.d2l.dev/search?full=%22interactivemixin%22&si=full&searchall=true) 🎉.

eKoopmans:
FYI we get this sort of history in GH now:

image

Which seems reasonable - following the link shows the rest of the history.

And blames look perfect, e.g. for [`focus-mixin`](https://github.com/BrightspaceUI/core/blame/main/mixins/focus/focus-mixin.js)!

PR #3493, opened 2023-04-06
Fix localize() param handling

bearfriend:


dlockhart (approved):

mixins/localize/localize-mixin.js
```diff
@@ -106,7 +106,7 @@ export const _LocalizeMixinBase = dedupeMixin(superclass => class LocalizeMixinC
 		if (!value) return '';
 
 		let params = {};
-		if (arguments.length > 1 && arguments[1].constructor === Object) {
+		if (arguments.length > 1 && arguments[1]?.constructor === Object) {
```
omsmith, line 109: Looks like there's a few tests for this already, add a quick one about passing null/undefined?

PR #3495, opened 2023-04-06
d2l-button-move updates

dbatiste:
This PR adds two things to `d2l-button-move`:

* fine-grained disabled options to control whether moving in different directions is allowed (since we would like to disable the appropriate interaction in the `d2l-html-editor`'s resizer when we reach min/max dimensions)
* support for dark theme (since `d2l-html-editor`'s code sample editor requires it)

Expanded on the visual-diffs which resulted in renaming the existing ones.

**Fine-grained disabled (`disabled-up`):**
image

**Dark theme:**
image


components/button/button-move.js
```diff
@@ -49,11 +51,41 @@ class ButtonMove extends FocusMixin(RtlMixin(LitElement)) {
 			 * @type {boolean}
 			 */
 			disabled: { type: Boolean, reflect: true },
```
dlockhart, line 53: Is there a purpose in still having `disabled` with all the fine-grained control down below? How do these new attributes interact with it and `disabled-tooltip`?

components/button/button-move.js
```diff
@@ -49,11 +51,41 @@ class ButtonMove extends FocusMixin(RtlMixin(LitElement)) {
...
 			/**
 			 * Tooltip text when disabled
 			 * @type {string}
 			 */
 			disabledTooltip: { type: String, attribute: 'disabled-tooltip' },
+			/**
+			 * Disables the up interaction
+			 * @type {boolean}
+			 */
+			disabledUp: { type: Boolean, attribute: 'disabled-up' },
```
dlockhart, line 88: These will all need to `reflect` since you're using CSS attribute selectors down below.

components/button/button-move.js
```diff
@@ -49,11 +51,41 @@ class ButtonMove extends FocusMixin(RtlMixin(LitElement)) {
 			 * @type {boolean}
 			 */
 			disabled: { type: Boolean, reflect: true },
+			/**
+			 * Disables the down interaction
+			 * @type {boolean}
+			 */
+			disabledDown: { type: Boolean, attribute: 'disabled-down' },
```
dlockhart, line 58: Wondering if you could just get away with `disabled-down` and `disabled-up` since those are the only things that are visually conveyed? Then the consumer could just ignore home/end/left/right if they consider them not possible options.

dbatiste: Yeah, that's certainly an option. That's what I originally had, but I didn't want to leak the up/down events since the event handler would have to do some work to re-evaluate whether the operation is valid. It followed that I wanted to treat them all the same way. I don't have a current need to disable left/right/home/end, but as soon as that need arises, I would want to add the properties to stop the events. Thouhts?

components/button/button-move.js
dbatiste: Maybe I should compute `disabled` for the `button` based on these `disabled-*`, and ditch the `disabled-tooltip` until we have a concrete need for it.

components/button/button-move.js
dlockhart: Could we make the assumption at least that "home" is tied to the "left" button and "end" is tied to the "right" one? And then should we be adding "page up" and "page down" support for the up/down equivalent?

If so, then you could suppress the "home" event when the right button is disabled, and vice versa.

dbatiste: Hmmm... in the list's case, the home/end are more closely aligned with up/down. It's hard to say what home/end really mean in this context. But page-up/page-down could certainly come into play as well. 🤷

dlockhart: Interesting, that's not what I would have expected as page up/down are what I'd normally use to "scroll all the way up" or down. Anyway that's getting into changing how drag & drop on lists works, which is out of scope here for sure!

dlockhart (approved):

components/button/button-move.js
dbatiste: Ok, got this all cleaned up!

components/button/button-move.js
dbatiste: Yeah I don't recall how they arrived at that. There is some overlap with the grid mode of lists, where Home/End moves focus to the first/last items, and PageUp/PageDown moves the focus up/down by 5 items. I seem to recall Jeff also asking for PageUp/PageDown to do similar with moving items up/down by 5, but I don't think it made it into the code.

PR #3497, opened 2023-04-06
fix: remove unused exports

dlockhart:
I've adjusted the places that were using these to use the new locations. The rest have far too many references to realistically update for the time being.


mixins/localize-static-mixin.js
dlockhart, line 1: This whole mixin is going away in a follow-up PR now that the only use has been removed.

dbatiste (approved):

PR #3501, opened 2023-04-08
Fix localizeHTML validation

bearfriend:
Setting `formattedMessage` before top-level validation allowed `html` template errors to be logged but the result was still returned.

Anything inside a `localizeMarkup` template fails validation while formatting, meaning they are caught before `formattedMessage` was set.


dlockhart (approved):
👍

PR #3502, opened 2023-04-09
Fix d2l-input-text value update

bearfriend:
Currently, if a `d2l-input-text` element's properties are changed, the `value` is updated outside of the lit update cycle, beating other properties, like `type`. If the previous (yet un-updated) `type` and the new `value` are not compatible (e.g. `number`, `"Text"`, respectively), the `input` will not only fail to update its `value` but will remove its current `value`.

Waiting for `updateComplete` to update the `value` ensures compatibility.

You can see this on the [S3 version](https://pr-3503-brightspace-ui-core.d2l.dev/mixins/localize/demo/localize-mixin.html) of #3503 in the Localize Mixin Sandbox demo by loading the Nested template, and then immediately loading the Ordinal template.


dlockhart (approved):
Wow that's a crazy bug!

PR #3503, opened 2023-04-09
Add localize-mixin sandbox demo

bearfriend:
Adds a `localize-mixin` sandbox demo to interactively show how the mixin and `messageformat` itself works. It also includes generator templates and various message templates to demonstrate proper usage and formatting.


.eslintrc.json
```diff
@@ -14,18 +14,10 @@
...
       "quotes": 0
     }
-  },
-  {
-    "files": "mixins/demo/**.html",
-    "rules": {
-      "quotes": 0,
-      "no-undef": 0,
-      "indent": 0
-    }
```
bearfriend, line 29: This was added [here](https://github.com/BrightspaceUI/core/commit/12fbea8546ea840e54ae0ced2d6cf72955206504), unnecessarily because we reverted the related demo changes.

mixins/localize/demo/localize-mixin-sandbox.js
```diff
@@ -0,0 +1,348 @@
...
+					${k}: ${val.constructor === String && !this.tags[k] ? `'${val.toString()}'` : val.toString()}`).join(',')}
+				}` : ''});
+			</d2l-code-view>
+		</div>
+
+		<h2 id="result" class="d2l-heading-4">
+			<span>Result</span>
+			<div>
+				<d2l-dropdown open-on-hover>
+					<d2l-icon id="results-help" icon="tier1:help" class="d2l-dropdown-opener"></d2l-icon>
```
dlockhart, line 211: Not a huge deal since this isn't product code, but this likely isn't super accessible since the icon isn't really meant to be an opener/target for things like dropdowns as it's unlabelled. Probably a button-icon would be more appropriate.

mixins/localize/demo/localize-mixin-sandbox.js
```diff
@@ -0,0 +1,348 @@
...
+		<div ?hidden="${this._error}">
+			<h2 class="d2l-heading-4">Code</h2>
+			<d2l-code-view id="code" language="javascript">
+				this.${localizeMethod}('${this._selectedTemplate.key}'${renderedArgs.length ? `, {${renderedArgs.map(([k, val]) => `
+					${k}: ${val.constructor === String && !this.tags[k] ? `'${val.toString()}'` : val.toString()}`).join(',')}
+				}` : ''});
+			</d2l-code-view>
+		</div>
+
+		<h2 id="result" class="d2l-heading-4">
```
dlockhart, line 207: Should the heading just be around the "result" text?

dlockhart (commented):
This is very cool! 😎

mixins/localize/demo/localize-mixin-sandbox.js
bearfriend: Ah, right, originally I had a tooltip on it, but you can't get to the link in a tooltip so I switched it out but hadn't thought of this aspect. It opens on hover, it that weird for a button?

mixins/localize/demo/localize-mixin-sandbox.js
bearfriend: Yes. I don't think I meant to do this, hah.

mixins/localize/demo/localize-mixin-sandbox.js
dlockhart: It's a little bit strange, but at least a button would have a `tabindex` so it would also open when it received focus.

mixins/localize/demo/localize-mixin-sandbox.js
```diff
@@ -0,0 +1,348 @@
...
+		return {
+			selectedTemplate: { type: String, attribute: 'selected-template', reflect: true },
+			_selectedTemplate: { state: true },
+			_error: { state: true }
+		};
+	}
+
+	static get styles() {
+		return [heading4Styles, bodyCompactStyles, css`
+			[hidden] {
```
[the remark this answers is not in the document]
bearfriend: Apparently, It wasn't. Six months later I have no idea why I added it to begin with.

dlockhart (commented):
Really cool! As mentioned in standup, this might get even more visibility and use if it became part of our (currently sparse but hopefully better one day) localization documentation on the Daylight Site.

mixins/localize/demo/lang/localize-sandbox.js
```diff
@@ -0,0 +1,63 @@
+export const langResources = {
+	basic: `The octopus has eight legs`,
+	arguments: `The {animalType} has {animalLegCount} legs`,
```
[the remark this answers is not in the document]
bearfriend: Yep, it confused me during my own demo

bearfriend:
> Really cool! As mentioned in standup, this might get even more visibility and use if it became part of our (currently sparse but hopefully better one day) localization documentation on the Daylight Site.

I'm going to look into this now

PR #3506, opened 2023-04-10
Fix localize-mixin mission demo

bearfriend:
Looks like we left a bunch of testing junk in the `mission` demo 😬


dlockhart (approved):
Oops, thanks!

PR #3508, opened 2023-04-10
Remove slot API from html-block control when deferring rendering

grant-cleary:
The `html` attribute was introduced a while ago to avoid double rendering (once in the liteDOM, once in the shadowDOM) certain expensive elements (e.g. iframes) within the the `html-block` component when using the standard deferred rendering workflow. Once all usages of the original slot workflow have been migrated to use the new attribute, we can strip out the slot and stamping logic that we previously used.


components/html-block/html-block.js
```diff
@@ -210,51 +209,34 @@ class HtmlBlock extends LitElement {
...
-	}
-
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 		this._updateContextKeys();
 	}
 
 	render() {
+		if (this.html && this.noDeferredRendering) {
+			throw new Error('Cannot use html attribute with no-deferred-rendering.');
```
grant-cleary, line 219: I'm not 100% sold on this. This is for sure an invalid use case. The previous behaviour here was to ignore the `html` attribute and prioritize the slot in this case, but that was mostly because while both the `html` and `slot` APIs were valid, the likelihood of mixing these by accident was significantly higher. I'm not sure that's a concern anymore.

components/html-block/test/html-block.visual-diff.html
```diff
@@ -26,64 +26,60 @@
 	</div>
 
 	<div class="visual-diff">
-		<d2l-html-block id="typography">
```
grant-cleary, line 29: These were all subtly different from the demo case and so I can almost guarantee some of these will be changed slightly. I'll take a peek if the visual diff tests fail and either update the diffs if we've lost useful test info, or merge the new goldens if it's just slightly different content/styling.

components/html-block/html-block.js
```diff
@@ -210,51 +209,34 @@ class HtmlBlock extends LitElement {
...
+
 		const renderContainerClasses = {
 			'd2l-html-block-rendered': true,
 			'd2l-html-block-compact': this.compact
 		};
 
 		return html`
 			<div class="${classMap(renderContainerClasses)}"></div>
-			<slot @slotchange="${this._handleSlotChange}"></slot>
+			${this.noDeferredRendering ? html`<slot @slotchange="${this._handleSlotChange}"></slot>` : ''}
```
grant-cleary, line 229: There's no need to render the slot at all anymore if we're deferring rendering.

components/html-block/html-block.js
dlockhart: You may want to include the web component name in the exception. For example:

> \<d2l-html-block\>: "html" attribute is not supported with "no-deferred-rendering"

One thing to watch out for is that properties can get set out-of-order, so even if someone's using this in a valid way, `render()` can still be called such that this is in a temporary invalid state.

To get around this in other components, we call a `_validateProperties()` method whenever `html` or `noDeferredRendering` change in `updated()`. It then sets a 3s timeout and _then_ does the check and throws. [Example from color input](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-color.js#L402).

grant-cleary: Ah yes, good points. I'll make these changes.

grant-cleary: I also stole your error message because I like it better. 😆

grant-cleary:
All consumers have now been updated, so this should be good to go. :)

dbatiste (approved):

PR #3512, opened 2023-04-11
Add full-height to dialog

oelhanafey:
[US150875](https://rally1.rallydev.com/#/?detail=/userstory/696122119573&fdp=true)

This PR adds the `full-height` property to the Dialog components which, when set, will set the Dialog to the maximum height. The maximum height is the minimum of the available height on the screen and `2 * width` to ensure that the dialog doesn't end up too unevenly tall.


components/dialog/dialog.js
```diff
@@ -33,6 +33,11 @@ class Dialog extends LocalizeCoreElement(AsyncContainerMixin(DialogMixin(LitElem
 			 */
 			describeContent: { type: Boolean, attribute: 'describe-content' },
 
+			/**
+			 * Whether to render the dialog at the maximum height
+			 */
+			fullHeight: { type: Boolean, attribute: 'full-height' },
```
dlockhart, line 39: This needs to be defaulted to `false` in the constructor.

eKoopmans (approved):
LGTM, I'm glad you were able to work it in nicely with `this._isFullHeight` - thanks Oscar!

PR #3514, opened 2023-04-11
feat: Replace "Select all" with item count in controls

eKoopmans:
This PR switches the default text for list-/table-controls from "Select all" to state the number of items, e.g. "5 items" or "3 of 7 items". A few other side effects of this PR:

- changed `_pageableInfo` to default to `null` on pageable-subscribers when they're not subscribed to anything
- renamed existing pager langterm and added a new one
- added pageable support to tables!
- simplified how table-controls inherits/diverges from selection-controls

Rally: https://rally1.rallydev.com/#/?detail=/userstory/673795655747&fdp=true

## Screenshot

image


components/selection/selection-controls.js
```diff
@@ -166,12 +181,8 @@ export class SelectionControls extends SelectionObserverMixin(RtlMixin(LocalizeC
 
 	_renderSelection() {
 		return html`
-			<d2l-selection-select-all></d2l-selection-select-all>
-			<d2l-selection-summary
-				aria-hidden="true"
```
eKoopmans, line 171: I removed `aria-hidden="true"` - I think this was being hidden because "Select all" was being repeated (aria-label on the select all checkbox, then again here). Since we're now displaying unique info here, I don't think we want it hidden.

components/list/list.js
```diff
@@ -183,8 +183,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 	}
 
 	_getItemByIndex(index) {
-		const items = this.getItems();
-		if (index > items.length - 1) return;
+		const items = this.getItems() || [];
```
dlockhart, line 186: Are we worried that the behaviour here changed a bit? Like previously if you asked for an item index beyond the bounds it would give you `undefined`. Not it'll throw an exception. It looks like the only place calling this is handling it though, so probably ok.

components/selection/selection-controls.js
dlockhart: Makes sense!

dlockhart (approved):
Awesome! Really nice seeing it all come together in such a simple way.

components/list/list.js
dlockhart: Oh interesting, TIL!

PR #3518, opened 2023-04-12
fix: Edge cases in selection-summary

eKoopmans:
A few tweaks here to selection-summary:

- fix inconsistent behaviour with single-selection where the summary sometimes rendered, based on a race condition (now never renders)
- ensure the summary still displays for pageables with no selection provider
    - added a visual-diff and demos for this case
- don't render select-all when there's no selection provider

I chatted with @dbatiste about this today, we agreed that the summary shouldn't ever be rendered for single-selection (radio buttons). There's one weird edge case that we don't think needs supporting:

- a single-selection list with list-controls that is **pageable**
- with these changes, the selection-summary would display nothing
- maybe we'd want the selection summary to say e.g. "3 of 7 items"...
    - but why would you ever have a list of radio items that was pageable, where you can't select them all?

Rally: https://rally1.rallydev.com/#/?detail=/userstory/673795655747&fdp=true

## Screenshots

This used to sometimes display summary text with "0 items":

image

Summary on a pageable with no selection:

image


components/selection/selection-summary.js
```diff
@@ -50,10 +50,13 @@ class Summary extends LocalizeCoreElement(SelectionObserverMixin(LitElement)) {
 	}
 
 	_updateSelectSummary() {
-		if (this._provider && this._provider.selectionSingle) return;
+		if (this._provider?.selectionSingle) {
+			this._summary = null;
+			return;
+		}
```
eKoopmans, line 56: This is the main change here - making sure `_summary` is always `null` for single selection.

components/selection/selection-select-all.js
```diff
@@ -44,7 +44,7 @@ class SelectAll extends FocusMixin(LocalizeCoreElement(SelectionObserverMixin(Li
 	}
 
 	render() {
-		if (this._provider && this._provider.selectionSingle) return;
+		if (!this._provider || this._provider.selectionSingle) return;
```
eKoopmans, line 47: Dave B suggested leaving this unchanged (still render when there's no selection provider), but I investigated a bit further and found:

- we definitely don't want it to render in the selection-controls
- [`selection-input` doesn't render with no provider](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-input.js#L93)

So it seems reasonable to not render this when there's no selection provider detected.

components/selection/selection-summary.js
```diff
@@ -50,10 +50,13 @@ class Summary extends LocalizeCoreElement(SelectionObserverMixin(LitElement)) {
...
 
 		let count;
-		if (this._provider && this._provider.selectionCountOverride !== undefined) {
+		if (this._provider?.selectionCountOverride !== undefined) {
```
dbatiste, line 59: Is the `_provider` check necessary here? Given the change above, it looks like we should never reach this code if  `_provider` is `null`/`undefined`.

components/selection/selection-summary.js
dbatiste: I think there is a slight behaviour change here. Previously we would have proceeded when `_provider` was `null` or `undefined`, whereas now we would not. I _think_ that's ok, and should allow us to simplify the conditional statements below.

components/selection/selection-summary.js
dbatiste: Maybe I am missing something. It looks like if the `_provider` isn't set, we _will_ wipe out the summary and return above.

dbatiste: Sorry, I was reading it as "provider is undefined or selection single". Is that what you meant it to be @eKoopmans ?

eKoopmans: Oops, I mostly thought this was cleaning things up, but you're right that this isn't the same - e.g. if `this._provider` is `null`, this check will now pass. I'll revert these optional chaining bits.

@dbatiste I ended up changing the logic from what we discussed yesterday. Yesterday we had decided that `_summary` should always be `null` if there's no provider, but that was breaking for scenarios that are pageable but not selectable.

dbatiste: Cool... yeah... easy to get tripped up with optional-chaining on statements like this one. 😄

dbatiste (approved):

PR #3527, opened 2023-04-17
Add skeleton to list-item-content

oelhanafey:
[DE52725](https://rally1.rallydev.com/#/?detail=/defect/694583527535&fdp=true)

This PR proposes the addition of the skeleton-mixin to list-item-content. This would allow consumers to skeletize all the contents of list items.

It's not an ideal skeleton style, since it creates a big grey box over all of the content. But a consumer has tried to do this [here](https://github.com/BrightspaceHypermediaComponents/activities/blob/bdb49a2011acf877108b7f70a58d9d72a3bfee34/components/d2l-activity-editor/d2l-activity-discussion-editor/d2l-activity-discussion-topic-manage-restrictions-dialog.js#L63), by copying and pasting the skeleton styles from core. This means it will not be responsive to changes to the mixin and will have to be updated in both places.

I don't love this. My thoughts are that this is not the expected use case of skeleton and we should encourage consumers to create their own skeleton content for situations like this which will result in a nicer looking skeleton.

In standup, we brought up the idea of expand/collapse being a potential use case. I think this fits better as part of the `d2l-list-item` skeleton. Currently, the skeleton for `d2l-list-item` will disable expand/collapse but does not skeletize the expand/collapse icon. We could expand those styles to also skeletize the icon.

Screen Shot 2023-04-17 at 11 51 35 AM

### DIFF
--- a/components/list/list-item-content.js
+++ b/components/list/list-item-content.js
@@ -1,17 +1,17 @@
 import '../colors/colors.js';
 import { bodyCompactStyles, bodySmallStyles } from '../typography/styles.js';
 import { css, html, LitElement } from 'lit';
-
+import { SkeletonMixin, skeletonStyles } from '../skeleton/skeleton-mixin.js';
 /**
  * A component for consistent layout of primary and secondary text in a list item.
  * @slot - Primary text of the list item
  * @slot secondary - Secondary text of the list item
  * @slot supporting-info - Information that supports the list item
  */
-class ListItemContent extends LitElement {
+class ListItemContent extends SkeletonMixin(LitElement) {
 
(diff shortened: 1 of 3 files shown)


dlockhart:
I agree with your analysis here -- that it makes more sense to improve our the list item's `skeleton` behaviour works by including the expand/collapse handle (and maybe the "..." actions button?), and then leave skeletonizing the content to the consumer.

oelhanafey:
The actions button is a slot, similar to the content so maybe that should also be the consumer's responsibility? Unless we just skeletize the entire slot content similar to what I'm doing here.

PR #3530, opened 2023-04-18
Skeletize expand/collapse icon in list-item skeleton

oelhanafey:
[US151504](https://rally1.rallydev.com/#/?detail=/userstory/698070048939&fdp=true)

This PR skeletizes the expand/collapse icon when the list-item is in skeleton. This brings it in line with the `list-item-checkbox-mixin` which does the same thing for selection.


components/list/list-item-expand-collapse-mixin.js
```diff
@@ -127,6 +128,7 @@ export const ListItemExpandCollapseMixin = superclass => class extends superclas
 		}
 		return html`
 			<d2l-button-icon
+				class="${this.skeleton ? 'd2l-skeletize' : ''}"
```
dlockhart, line 131: I'd recommend using `classMap` here as it should just not output an empty `class` attribute in the case where you don't have a skeleton.

dlockhart (approved):
Looks good! Thanks for adding this.

dbatiste (approved):

PR #3533, opened 2023-04-19
fix: Event subscription before registry is ready

eKoopmans:
While backporting the table-load-more to some consumers, I found a strange race condition that was preventing the load-more pager from subscribing to the table. This PR fixes that race condition.

In short:

- in the backport, the pager is being rendered **before** the table-wrapper
- this is because it's being passed as a slot to a component that internally renders the table-wrapper
    - so the outer component renders the pager first, then the inner component renders table-wrapper and slots the pager in place
- that means when the subscription event fires, there's nothing to catch it!

I chatted with @dbatiste and we landed on this fix: keep firing the subscription event until it's caught, for some fixed amount of time. In my testing 20ms was usually enough, I've set the retry window to 100ms but we can make it larger if we want.

Also, this fix was surprisingly similar to one already in place for the ID subscriber, so I combined the approaches and added unit tests for both.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/656749937623&fdp=true

### DIFF
--- a/controllers/subscriber/subscriberControllers.js
+++ b/controllers/subscriber/subscriberControllers.js
@@ -137,7 +137,7 @@ export class EventSubscriberController extends BaseSubscriber {
 	}
 
 	hostConnected() {
-		this._subscriptionComplete = this._keepTrying(() => this._subscribe(), 20, 100);
+		this._subscriptionComplete = this._keepTrying(() => this._subscribe(), 40, 400);
 	}
 
 	hostDisconnected() {
--- a/controllers/subscriber/subscriberControllers.js
+++ b/controllers/subscriber/subscriberControllers.js
@@ -9,11 +9,39 @@ class BaseController {
 		this._name = name;
 		this._options = options;
 		this._eventName = `d2l-subscribe-${this._name}`;
-		this._subscriptionComplete = Promise.resolve();
 	}
 }
 
 class BaseSubscriber extends BaseController {
+	constructor() {
+		super(...arguments);
+
+		this._subscriptionComplete = Promise.resolve();
+		this._timeouts = new Set();
+	}
+
+	hostDisconnected() {
+		this._timeouts.forEach(timeoutId => clearTimeout(timeoutId));
+	}
+
+	_keepTrying(callback, interval, maxTime, errorPayload, elapsedTime = 0) {
+		const response = callback();
+
+		if (response) return response;
+		if (elapsedTime >= maxTime) {
+			if (this._options.onError) this._options.onError(errorPayload);
+			return response;
+		}
+
+		return new Promise(resolve => {
+			const timeoutId = setTimeout(async() => {
+				this._timeouts.delete(timeoutId);
+				resolve(await this._keepTrying(callback, interval, maxTime, errorPayload, elapsedTime + interval));
+			}, interval);
+			this._timeouts.add(timeoutId);
+		});
+	}
+
 	_subscribe(target = this._host, targetLabel) {
 		const isBroadcast = target === this._host;
 
@@ -25,10 +53,7 @@ class BaseSubscriber extends BaseController {
 		target.dispatchEvent(evt);
 
 		const { registry, registryController } = evt.detail;
-		if (!registry) {
-			if (this._options.onError) this._options.onError();
-			return;
-		}
+		if (!registry) return false;
 
 		if (targetLabel) {
 			this._registries.set(targetLabel, registry);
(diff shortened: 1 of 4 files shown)


controllers/subscriber/subscriberControllers.js
```diff
@@ -111,16 +137,11 @@ export class EventSubscriberController extends BaseSubscriber {
...
 
 	hostConnected() {
-		// delay subscription otherwise import/upgrade order can cause selection mixin to miss event
-		this._subscriptionComplete = new Promise(resolve => {
-			requestAnimationFrame(() => {
-				this._subscribe();
-				resolve();
-			});
-		});
+		this._subscriptionComplete = this._keepTrying(() => this._subscribe(), 20, 100);
```
eKoopmans, line 140: One reason to keep these retry numbers different from the ID subscriber:

We *expect* the ID subscriber to fail, in certain conditions. For example pageable-observer uses two subscribers (event + ID), and when it's in ID mode we want the event subscriber to silently fail. So we don't want those expected failures to be too expensive.

The ID subscriber, on the other hand, would never get to its "keep-trying" stage if it's expecting to fail, because it only triggers when there's an explicit ID specified.

controllers/subscriber/subscriberControllers.js
```diff
@@ -9,11 +9,39 @@ class BaseController {
...
+		this._timeouts.forEach(timeoutId => clearTimeout(timeoutId));
+	}
+
+	_keepTrying(callback, interval, maxTime, errorPayload, elapsedTime = 0) {
+		const response = callback();
+
+		if (response) return response;
+		if (elapsedTime >= maxTime) {
+			if (this._options.onError) this._options.onError(errorPayload);
+			return response;
```
dlockhart, line 33: Is returning `response` especially helpful here since we already know it's `false`-y? Would `return;` (`return undefined`) suffice?

controllers/subscriber/subscriberControllers.js
```diff
@@ -9,11 +9,39 @@ class BaseController {
...
+		}
+
+		return new Promise(resolve => {
+			const timeoutId = setTimeout(async() => {
+				this._timeouts.delete(timeoutId);
+				resolve(await this._keepTrying(callback, interval, maxTime, errorPayload, elapsedTime + interval));
```
dlockhart, line 39: It probably doesn't matter for the purpose of what you're doing here, but "elapsed time" probably doesn't just equal the sum of all the intervals, since the browser will do its best to call your timeout after interval, but there's no guarantee. If accuracy is needed (probably not), then the [user timing API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) may be more appropriate.

controllers/subscriber/test/subscriberControllers.test.js
```diff
@@ -80,6 +81,16 @@ const eventSubscriberCombined = defineCE(generateSubscriber('event', 'test-subsc
...
 
+const indirectSlotRegistries = defineCE(class extends LitElement {
+	render() {
+		return html`${unsafeHTML(`
+			<${registries} id="registry-shadow">
+				<slot></slot>
+			</${registries}>
+		`)}`;
+	}
+});
```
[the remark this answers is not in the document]
dlockhart: Cool!

dlockhart (approved):
Nice, I really like this solution and it's great that we were able to reuse it from elsewhere and centralize it!

controllers/subscriber/subscriberControllers.js
eKoopmans: For the current two consumers of this function it doesn't matter, but I've tried to keep the return value unchanged in case the consumer wanted to know exactly what falsey response came out. This way the response (after awaiting `_keepTrying`) is always a direct response from the callback they invoked.

controllers/subscriber/subscriberControllers.js
eKoopmans: Thanks, yeah I think this is sufficient for here.

controllers/subscriber/subscriberControllers.js
eKoopmans: I'm going to increase the window, I'm nervous about users with slow machines missing it. I'm thinking 400ms with a 40ms interval - 10 tries doesn't seem excessive.

controllers/subscriber/subscriberControllers.js
dlockhart: Cool, thanks for the context!

PR #3537, opened 2023-04-19
fix: Initializing itemShowingCount for pageables

eKoopmans:
Another small change - as part of #3533 I made a fix to how pageable initializes `itemShowingCount`, I'd like to switch it with a slightly different fix here.

The previous fix (reprocessing `_getItemShowingCount` on each subscription) could have a performance concern if there's a lot of subscribers, removing it again and switching to `connectedCallback` should be sufficient. Not a big deal, but I wanted to make the change while it's on my mind.

### DIFF
--- a/components/paging/pageable-mixin.js
+++ b/components/paging/pageable-mixin.js
@@ -21,8 +21,8 @@ export const PageableMixin = superclass => class extends CollectionMixin(supercl
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


dlockhart (approved):

eKoopmans:
Haha okay, I'm going to leave well enough alone, the visual diffs definitely aren't happy with this change 😅.

PR #3542, opened 2023-04-20
US151362 - Fix offscreen RTL scrollbar issue for browsers that support CSS Logical Properties

svanherk:
The first commit fixes the visual-diff tests - applying `dir="rtl"` to the `div` was only really testing the `sass` RTL styles, the `d2l-offscreen` and `offscreenStyles` tests were still using the `left: -10000px;` style (which doesn't cause the scrollbar issue).  Fixing the tests expectedly caused the RTL goldens to[ start showing up blank](https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/offscreen/2023-04-20-14.58.13.997-8f075706b837fc58e7fe90b13891178843f1f95f-report.html).

Then the second commit fixes this issue, so the `visual-diff` tests pass again.  Based on my testing, adding `inset-inline-start` does exactly what we want - a browser that supports it will use it and ignore `left` and `right`. A browser that doesn't will happily ignore it, and will still have the scrollbar issue in RTL.  This will be Safari <14.1, Chrome/Edge < 87, and Firefox < 63: https://caniuse.com/mdn-css_properties_inset-inline-start


dlockhart (approved):
Nice!

components/offscreen/offscreen.js
```diff
@@ -25,6 +26,7 @@ class Offscreen extends RtlMixin(LitElement) {
 		return css`
 			:host {
 				height: 1px;
+				inset-inline-start: -10000px;
```
svanherk, line 29: Technically, adding it here is not needed to solve the issue, because `d2l-offscreen` imports the `RTLMixin`.  Same with in the `sass` file, which doesn't need to worry about `shadowRoot`s. But I figured I'd put it in all three places now, for consistency - to highlight any issues early and to remind us when we go to remove the `left` and `right` solution.

PR #3546, opened 2023-04-21
feat: Support combining sticky and scroll-wrapper

eKoopmans:
This PR allows tables with sticky behaviour to use the scroll-wrapper. Details:

- It's relatively seamless - just add the `scroll-wrapper` attribute, e.g. `<d2l-table-wrapper sticky-headers scroll-wrapper>`
- It supports sticky columns (via the `sticky` attribute on individual cells)
- It **does not** support sticky headers that are outside of `thead` (e.g. `tr.d2l-table-header` and `tr[header]`)

To facilitate all this, a new `.customScrollers` property has been added to `d2l-scroll-wrapper` that allows it to register primary and secondary scrollers and link their scrolling together.

This should be mostly good to go, but if someone could help with the two items in "remaining work" I'd really appreciate it!

Rally: https://rally1.rallydev.com/#/?detail=/userstory/673801352295&fdp=true

Closes #2867 

## Remaining work

- [x] Resolve issue where the scroll-wrapper doesn't always recognize it's on the right edge
    - Seems to be caused by the contents resizing by one pixel at the end, and/or half-pixels
    - I was able to force a fix with `Math.ceil(Math.abs(this._container.scrollLeft))` on Line 220 of `scroll-wrapper.js`, but I'm not sure if that's robust
- [x] Unit and visual-diff tests (including one that scrolls a secondary scroller by focusing something in it that's offscreen)

## Screenshots

image

image


components/table/table-wrapper.js
```diff
@@ -190,6 +214,15 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				reflect: true,
 				type: Boolean
 			},
+			/**
+			 * Whether to wrap the table in a scroll-wrapper (enabled by default when sticky-headers is false).
+			 * @type {boolean}
+			 */
+			scrollWrapper: {
```
dlockhart, line 221: Wondering if this should be called something like `sticky-headers-scroll-wrapper`? Just due to the confusion that by default tables get a scroll wrapper _unless_ they have sticky headers. And even that was only because we didn't initially support both, but now that we do... yeah it's confusing. :)

components/table/table-wrapper.js
```diff
@@ -266,6 +299,7 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		this._table = null;
 		this._tableIntersectionObserver = null;
 		this._tableMutationObserver = null;
+		this._tableScrollers = {};
```
dlockhart, line 302: Init `scroll-wrapper` to `false`.

components/table/table-wrapper.js
```diff
@@ -388,6 +424,10 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		this._table = e.target.assignedNodes({ flatten: true }).find(
 			node => (node.nodeType === Node.ELEMENT_NODE && node.tagName === 'TABLE' && node.classList.contains('d2l-table'))
 		);
+		this._tableScrollers = (this.scrollWrapper && this.stickyHeaders) ? {
+			primary: this._table?.querySelector('tbody'),
+			secondary: this._table?.querySelector('thead'),
```
dlockhart, line 429: Not sure if we care about this, but there's a weird edge case here where if they don't use `tbody` and `thead` but do set `scroll-wrapper`, then it'll try to use it but fail and result in just a normal scroll-wrapper... which isn't going to work very well.

components/table/table-wrapper.js
eKoopmans: Yeah, I debated this. Options:

1. ❌ No extra flag at all, change `sticky-headers` to always use this new mode
    - doesn't work because the new mode doesn't support all existing sticky behaviour (header rows that aren't in thead), and because some use cases might specifically not want scroll-wrappers
2. ❌ No extra flag at all, change the default (non-`sticky-headers`) to always use the new mode
    - pretty unintuitive, because sticky behaviour would be enabled without the `sticky-headers` flag
    - also there might use cases that specifically don't want any sticky behaviour
3. ❓ A single `sticky-headers-scroll-wrapper` flag, used *instead of* `sticky-headers`
    - definitely an option; I wondered if this would be more confusing though
    - also the CSS is a bit messier this way, need to double most checks for all the shared logic like `d2l-table-wrapper[sticky-headers] ..., d2l-table-wrapper[sticky-headers-scroll-wrapper] ...`
4. ⭐ Add an extra `scroll-wrapper` flag used in combo with `sticky-headers`
    - this is what I opted for...
    - could name this more verbosely, I thought it would be overkill to use e.g. `<d2l-table-wrapper sticky-headers sticky-headers-scroll-wrapper>`
    - but yes, it's confusing to people not using sticky-headers, because in that case specifying `scroll-wrapper` makes no difference (it's always enabled)

components/table/table-wrapper.js
eKoopmans: So I've just double-checked on our earliest supported browsers on Saucelabs, and they all automatically create a `tbody` when the table has any contents. The thead needs to be explicit, even on modern browsers, which makes sense.

So it'll always treat the tbody as the primary scroller, it just might not have any linked scroller if there's no thead. That sounds okay to me, since the thead is the only thing that can be sticky anyway...

I've updated the jsdoc to call out more explicitly that `sticky-headers-scroll-wrapper` requires the sticky headers to be in `thead`.

dlockhart: Nice -- sounds good!

dlockhart (approved):
LGTM, thanks for seeing this through Erik!

eKoopmans:
Thanks Dave! Still have those two "Remaining work" items, plus the failing visual-diff, could use a hand getting those across the finish line.

I've taken a look at the [failing visual-diff](https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/scroll-wrapper/2023-04-21-20.53.00.580-4723695972019cccb0e015c318ea0e42b3821026-report.html), and TBH I don't understand why the golden looks the way it does - do you have any context?

Anyway, I'll return to this on Monday, have a good weekend!

dlockhart:
> I've taken a look at the [failing visual-diff](https://visual-diff.d2l.dev/screenshots/BrightspaceUI/core/scroll-wrapper/2023-04-21-20.53.00.580-4723695972019cccb0e015c318ea0e42b3821026-report.html), and TBH I don't understand why the golden looks the way it does - do you have any context?

Looked into this a bit. I think it's because [of this change](https://github.com/BrightspaceUI/core/pull/1386), where we purposefully disable the scroll wrapper when you're printing. The CSS that it was using to do this is no longer "hitting" with your changes. I can try to figure out a way to fix this up.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -56,7 +65,6 @@ class ScrollWrapper extends RtlMixin(LitElement) {
 			.d2l-scroll-wrapper-container {
 				box-sizing: border-box;
 				outline: none;
-				overflow-x: auto;
 				overflow-y: var(--d2l-scroll-wrapper-overflow-y, visible);
 			}
 			.d2l-scroll-wrapper-container:${unsafeCSS(getFocusPseudoClass())} {
```
eKoopmans, line 70: @dlockhart with your fix of the print styling, this is the only remaining style that isn't right - and I'm not sure a great way to fix it.

We want the `focus-visible` styles to be changed on the actual container, which with split-scrollers is some arbitrary element, rather than `.d2l-scroll-wrapper-container`. Unfortunately we can't add pseudo-selector styles via JS, and the element isn't guaranteed to be in the same DOM scope, so we can't fix it by e.g. just adding a class to the target.

dlockhart: I've pushed some changes that address this in exactly the way you suggested.

components/scroll-wrapper/scroll-wrapper.js
```diff
@@ -219,6 +283,58 @@ class ScrollWrapper extends RtlMixin(LitElement) {
...
+		this._baseContainer = this.shadowRoot.querySelector('.d2l-scroll-wrapper-container');
+		this._container = this.customScrollers?.primary || this._baseContainer;
+		this._secondaryScrollers = this.customScrollers?.secondary || [];
+		if (this._secondaryScrollers.length === undefined) this._secondaryScrollers = [ this._secondaryScrollers ];
+		this._allScrollers = [ this._container, ...this._secondaryScrollers ];
+
+		if (this._container) {
+			this._container.classList.add('d2l-scroll-wrapper-focus');
+			const styleRoot = getStyleSheetInsertionPoint(this._container);
+			if (styleRoot && 'adoptedStyleSheets' in styleRoot) {
```
dlockhart, line 311: This is what we need to do to apply the focus styles into the right DOM scope for the container. The approach I took was to essentially use `querySelector` as I travel up the DOM to find the correct document fragment in which to insert the CSS.

Browser support for constructible stylesheets is OK, and I've tested this in older browsers and it doesn't blow up. Instead, it ends up using the browser's default focus outline -- which looks very similar to ours.

Just using the built-in focus ring is an option here for us... it would certainly be a lot easier! @dbatiste FYI in case you have opinions as the existing focus styles were added for WCAG 2.2.

eKoopmans: Awesome, thanks Dave!

I just remembered I wanted to add a visual diff for the split scrollers, with something focusable in a "secondary scroller" that causes it to scroll, to make sure the primary scroller properly matches it. I'll try to add that later today.

dlockhart:
My understanding of what we have now:

- `<d2l-table-wrapper>`: no sticky headers, scroll wrapper
- `<d2l-table-wrapper sticky-headers>`: sticky headers, no scroll wrapper
- `<d2l-table-wrapper sticky-headers sticky-headers-scroll-wrapper>`: sticky headers, scroll wrapper

If we were starting from scratch, I think the ideal scenario would be have something like a `no-scroll-wrapper` that could be applied irregardless of whether the headers are sticky. But we can't do that because currently `sticky-headers` forces no scroll wrapper (due to it not working for years) and I'm guessing we're not interested in going through all those places and updating them.

I'm not sure I follow your question though... how would a single `sticky-headers-scroll-wrapper` work?

eKoopmans:
Sorry, got pulled away this afternoon. I meant, as an alternative:

- `<d2l-table-wrapper>`: no sticky headers, scroll wrapper
- `<d2l-table-wrapper sticky-headers>`: sticky headers, no scroll wrapper
- `<d2l-table-wrapper sticky-headers-scroll-wrapper>`: sticky headers, scroll wrapper

dlockhart:
Ok gotcha. But then they could add both `sticky-headers` and `sticky-headers-scroll-wrapper` and it would be a little odd. Not that they can't use `sticky-headers-scroll-wrapper` as the code is now without `sticky-headers` and it's also strange. I think past-us has made this very difficult for now-us, no matter what we try to do here.

I'm fine with where the code is at now. It's not how we'd do it starting from scratch, but that's not a luxury we have.

PR #3558, opened 2023-04-25
DE52736 - Add demo and visual diffs for nested lists

svanherk:
When nested list items have no controls, there's currently no nesting style - it appears as a flat list.  You can see this in the new demo added here: https://pr-3558-brightspace-ui-core.d2l.dev/components/list/demo/list-nested.html.

The nesting logic is complicated, depending on what controls the parent items and and children items have. While Jeff is working on how much indent we want for each scenario, I've added new visual-diff tests to capture the differing iterations. The image is quite large but seems to work fine like that. Each test takes around five seconds, but there's only four of them. If we see issues I can break it out - I've set things up with the helper component to make the transition to `fixture`-based visual-diffs pretty smooth.

The different things that will impact the nesting level are:
- What type of parent item the list has
- Whether the list has an expandable children (because then space is reserved on _all_ children)
- Whether the list has different types of parents (captured in the "one list" scenario)


components/list/test/list-nested.visual-diff.js
```diff
@@ -0,0 +1,45 @@
...
+	let browser, page;
+
+	before(async() => {
+		browser = await puppeteer.launch();
+		page = await visualDiff.createPage(browser, { viewport: { width: 1000, height: 8500 } });
+	});
+
+	beforeEach(async() => {
+		await visualDiff.resetFocus(page);
+		await page.mouse.move(-1, -1);
```
dlockhart, line 17: Curious what the mouse move was needed for?

components/list/test/list-nested.visual-diff.html
```diff
@@ -0,0 +1,33 @@
...
+			import '../../typography/typography.js';
+			import './nested-list-iterations-helper.js';
+		</script>
+		<title>d2l-list-nested</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+	</head>
+	<body class="d2l-typography">
+
+		<div style="width: 700px;">
```
dlockhart, line 20: Could the viewport have just been set to 700px wide? Or maybe the test element itself (`:host { width: 700px; }`)? Just hoping to avoid extra stuff in these HTML files for when we have to migrate them. :)

svanherk: So I added this here instead of using `viewport` to make the pages look better when you go to them directly outside of running the tests, which won't be a thing once we migrate 😅. I'll add it to the component for now at least

components/list/test/list-nested.visual-diff.js
dbatiste: Looks like it is from here: https://github.com/BrightspaceUI/core/pull/3434

From Erik:
>I also had to move the mouse because tests that use hover() move the mouse on the screen and then leave it there, which was causing hover styles to appear on this list!

components/list/test/list-nested.visual-diff.js
```diff
@@ -0,0 +1,45 @@
...
+				await page.goto(`${visualDiff.getBaseUrl()}/components/list/test/list-nested.visual-diff.html?dir=${dir}`, { waitUntil: ['networkidle0', 'load'] });
+				await page.bringToFront();
+			});
+
+			[
+				{ name: 'separate lists all possible combos', selector: '#separate-lists-all-possible-combos' },
+				{ name: 'one list all possible combos', selector: '#one-list-all-possible-combos' }
+			].forEach((info) => {
+				it(info.name, async function() {
+					await page.evaluate(() => {
```
dbatiste, line 35: Is waiting a task necessary for these tests?

dbatiste (approved):
Other than those couple minor questions, looks good/useful to me!

components/list/test/list-nested.visual-diff.js
dlockhart: Seeing as these particular tests don't hover then it should be safe to remove.

This is good though -- I've added resetting the mouse position to my reset logic in the new visual diff stuff.

dlockhart (approved):

components/list/test/list-nested.visual-diff.js
svanherk: Ran it a bunch of times locally without it and still seems stable - I'll remove and see how it goes!

PR #3563, opened 2023-04-25
Focus the select-all checkbox when select-all-pages is clicked so focus isn't lost

dbatiste:
[DE52842](https://rally1.rallydev.com/#/?detail=/defect/696182478911&fdp=true)

This PR applies focus to the `d2l-selection-select-all` checkbox when the `d2l-selection-select-all-pages` button is clicked so that focus isn't lost when the button is hidden/removed from view.

### DIFF
--- a/components/selection/selection-select-all-pages.js
+++ b/components/selection/selection-select-all-pages.js
@@ -40,7 +40,10 @@ class SelectAllPages extends FocusMixin(LocalizeCoreElement(SelectionObserverMix
 	}
 
 	_handleClick() {
-		if (this._provider) this._provider.setSelectionForAll(true, true);
+		if (!this._provider) return;
+
+		this._provider.setSelectionForAll(true, true);
+		this._provider._focusSelectAll();
 	}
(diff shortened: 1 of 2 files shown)


components/selection/selection-select-all-pages.js
```diff
@@ -40,7 +40,10 @@ class SelectAllPages extends FocusMixin(LocalizeCoreElement(SelectionObserverMix
 	}
 
 	_handleClick() {
-		if (this._provider) this._provider.setSelectionForAll(true, true);
+		if (!this._provider) return;
+
+		this._provider.setSelectionForAll(true, true);
+		this._provider._focusSelectAll();
```
dbatiste, line 46: We delegate this task to the provider to locate and focus on the `d2l-selection-select-all`. We don't rely on an event because there may not be an ancestor in the event path that knows how to locate the select-all. The selection provider is in the best position to identify the select-all.

dlockhart (approved):

PR #3564, opened 2023-04-25
Announce instructions for keyboard users when focusing on tag-list

GZolla:
`d2l-tag-list` was only announcing the instructions for navigating and deleting tag list items the first time the user focused into the tag list. These change makes it so its announce whenever the first tag item in the list gains focus, with the exception of the first time this happens where the tooltip is displayed and announced.

Rally: [DE52883](https://rally1.rallydev.com/#/?detail=/defect/697327210647&fdp=true)


components/tag-list/tag-list.js
```diff
@@ -354,8 +354,13 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		setTimeout(() => focusableElem.focus(), this.clearFocusTimeout);
 	}
 
-	_handleKeyboardTooltipShown() {
-		this._hasShownKeyboardTooltip = true;
+	_handleKeyboardTooltipShown(e) {
+		if (!this._hasShownKeyboardTooltip) this._hasShownKeyboardTooltip = true;
+		else if (this.clearable && !this.contains(e.detail?.relatedTarget)) {
```
dbatiste, line 359: What is the `relatedTarget` meant to indicate?

GZolla: Related target is the blurred element from the first `tag-list-item` focus event. So this is meant to run only if the focus is coming from outside the `tag-list`, preventing the announcement when navigating to the first item using arrow keys.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -140,17 +142,19 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		this.addEventListener('focus', async(e) => {
 			// ignore focus events coming from inside the tag content
 			if (e.composedPath()[0] !== this) return;
+			if (this.keyboardTooltipItem && this.keyboardTooltipShown) {
```
dbatiste, line 145: This is kinda of leaky in the sense that the `relatedTarget` means something in the context of the `focus` event handler, but it doesn't really mean anything in the context of the tag-list's `d2l-tag-list-item-tooltip-show` event handler. It would be better if we specified a property on `detail` that was more meaningful, but I wonder if we can just handle the announcement here with something like:

```javascript
const tagList = findComposedAncestor(this, elem => elem.tagName === 'D2L-TAG-LIST');
if (!tagList.contains(e.relatedTarget)) {
    // focus moved from outside the tag-list so do the announcement
    announce(...);
}
```
However, I think that `contains(...)` will not identify elements across shadow boundaries, so something like this is probably more resilient.
```javascript
if(!isComposedAncestor(tagList, e.relatedTarget)) {
    // focus moved from outside the tag-list so do the announcement
    announce(...);
}
```
See: [findComposedAncestor](https://github.com/BrightspaceUI/core/blob/main/helpers/dom.js#L57), [isComposedAncestor](https://github.com/BrightspaceUI/core/blob/main/helpers/dom.js#L201).

Then we don't need this extra event dispatch, and the tag-list code is a bit simpler. Thoughts?

GZolla: This is much cleaner, I'll commit the changes.

dbatiste (approved):
Did some testing and it looks/sounds good.

Note: I don't think this was NVDA specific. I found that on MacOS with VO - the announcement was good on Chrome, but was not repeated on Safari or Firefox. I did not do a test in Windows. 

It would probably be good to update the PR as well as the title and description in the Rally ticket to more accurately reflect what's being fixed, since right now it makes it sounds like this was NVDA specific.

PR #3565, opened 2023-04-25
Add no-selection-text property to d2l-selection-controls so consumers can override it.

dbatiste:
[US151811](https://rally1.rallydev.com/#/?detail=/userstory/698665144937&fdp=true)

Consumers require the ability to override the `no-selection-text` property for `d2l-list-controls` internal `d2l-selection-summary`. To do this, we can expose a new `no-selection-text` property on `d2l-selection-controls`/`SelectionControls` and pass this value down. If provided, it will override the normal pageable counts that are displayed.


components/selection/selection-controls.js
```diff
@@ -150,8 +155,11 @@ export class SelectionControls extends PageableSubscriberMixin(SelectionObserver
 		if (changedProperties.has('noSticky')) {
 			this._stickyObserverUpdate();
 		}
-		if (changedProperties.has('_pageableInfo')) {
-			this._noSelectionText = this._getNoSelectionText();
+	}
+
+	willUpdate(changedProperties) {
+		if (changedProperties.has('noSelectionText') || changedProperties.has('_pageableInfo')) {
+			this._noSelectionText = this.noSelectionText ? this.noSelectionText : this._getNoSelectionText();
```
eKoopmans, line 162: Could probably simplify with `this.noSelectionText || this._getNoSelectionText()`.

eKoopmans (approved):
Looks great, thanks Dave!

components/filter/test/filter.visual-diff.js
```diff
@@ -228,7 +228,7 @@ describe('d2l-filter', () => {
 
 		describe('opener', () => {
 			after(async() => {
-				await page.evaluate(() => document.querySelector('html').setAttribute('lang', 'en'));
+				await page.reload({ waitUntil: ['networkidle0', 'load'] });
```
dbatiste, line 231: Moving the resolution of `_noSelectionText` from `updated` to `willUpdate` seemingly caused the filter tests that followed `multiple-over-99` to show Arabic. I tried adding a delay but it didn't help. The `reload` did however. Really looking forward to having better test isolation.

eKoopmans (approved):

dlockhart (approved):

PR #3573, opened 2023-04-28
Update the d2l-meter-* demos and tests.

dbatiste:
This PR makes a few updates for the meter demos and tests:

* Update the demo page to include more cases to make it easier to test the different ways it can be used
* Revamp the `d2l-meter-*` visual-diff tests to include more cases and remove diffs not adding value
* Update the Daylight demo for `d2l-meter-linear` to include three different ways it can be used per Jeff's request
* Fix the Daylight demo for `d2l-meter-linear` to properly render when `text-inline` is specified


svanherk (approved):
Looks awesome! A lot easier to tell what's going on

PR #3581, opened 2023-05-01
feat: add skeleton to `d2l-collapsible-panel`

GZolla:
Implementing `SkeletonMixn` in `CollapsiblePanel`. The change will skeletonize the panel title and the collapse action icon as well as prevent users from collapsing or expanding the panel when the `skeleton` property is set. 

All the components inside of slots are not skeletonized, this responsibility is still left to the consumers of the component. Only the `d2l-collapsible-panel-summary-item` inside the summary slot will be skeletonized automatically.

Rally: [US151933](https://rally1.rallydev.com/#/?detail=/userstory/699017553365&fdp=true)


components/collapsible-panel/test/screenshots/ci/golden/collapsible-panel/d2l-collapsible-panel-ltr-subtle-large-padding.png
dbatiste, line 1: Looks like this one was a 1px flake. I suspect we'll see this one randomly fail the other way in future tests.

GZolla: I tested it out and it fails because the html page has shifted slightly when adding the new skeleton panel.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -97,6 +98,9 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 			:host([padding-type="large"][type="inline"]) {
 				--d2l-collapsible-panel-spacing-inline: 2rem;
 			}
+			:host([skeleton]) {
+				pointer-events: none;
+			}
```
oelhanafey, line 103: Why is this necessary?

components/collapsible-panel/test/collapsible-panel.test.js
```diff
@@ -98,4 +98,31 @@ describe('d2l-collapsible-panel', () => {
 		});
 	});
 
+	describe('skeleton', () => {
+
+		it('should not expand on click', async() => {
+			const elem = await fixture(html`
+				<d2l-collapsible-panel panel-title="Panel Title" skeleton>
+					<div slot=header>Header</div>
```
oelhanafey, line 106:
```suggestion
					<div slot="header">Header</div>
```

components/collapsible-panel/collapsible-panel.js
```diff
@@ -393,7 +398,7 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 						<slot name="actions"></slot>
 					</div>
 					<div class="d2l-collapsible-panel-opener">
-						<d2l-icon-custom size="tier1">
+						<d2l-icon-custom size="tier1" class='d2l-skeletize'>
```
oelhanafey, line 401:
```suggestion
						<d2l-icon-custom size="tier1" class="d2l-skeletize">
```

components/collapsible-panel/collapsible-panel.js
```diff
@@ -385,6 +389,7 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 	}
 
 	_renderHeader() {
+
 		return html`
```
oelhanafey, line 393:
```suggestion
	_renderHeader() {
		return html`
```

components/collapsible-panel/collapsible-panel.js
```diff
@@ -26,7 +27,7 @@ const defaultHeading = 3;
  * @fires d2l-collapsible-panel-expand - Dispatched when the panel is expanded
  * @fires d2l-collapsible-panel-collapse - Dispatched when the panel is collapsed
  */
-class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
+class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
```
oelhanafey, line 30: It would be nice to also remove the `cursor: pointer` when in skeleton so the panel doesn't appear clickable

oelhanafey, line 30: Do we need to add skeleton to the `d2l-collapsible-panel-summary-item` as well or just the main component?

oelhanafey, line 30: We could update the `README.md` to add the new `skeleton` attribute which is inherited from the `SkeletonMixin`

dbatiste: I think the summary item should support it as well.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -374,6 +376,11 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
 
 	_handleSummarySlotChange(e) {
 		const content = e.target.assignedNodes({ flatten: true });
+		for (const element of content) {
+			if (this.skeleton) element.setAttribute('skeleton', true);
+			else element.removeAttribute('skeleton');
+		}
```
oelhanafey, line 382: Do we want the panel to be auto assigning skeleton to the summary items? What we have done for other components with slots like lists, is give the responsibility of adding skeleton to the slot content to the consumer. That would also give the consumer the ability to separately assign skeleton to the different summary items if it was needed.

In this case though, the slot only accepts one type of component that is tightly coupled with the panel data so maybe there is a case for this. Interested to hear if others think differently.

GZolla: Yeah that was my thought process too. The only problem is we are not actually locking down consumers from adding other components so I was thinking we could either impose that restraint or set skeleton automatically but only for summary items.

GZolla: The use of `li` in `d2l-collapsible-panel` is in an unused FACE related repo, they are still using `li` in `d2l-labs-accordion` but we can change that when backporting.

oelhanafey: We should probably either lock the slot down first or make sure we're only setting the skeleton on summary items similar to what we are doing in [filter](https://github.com/BrightspaceUI/core/blob/4b95ca8e8fb4c3406bce07cf7e61b088ef8dc5d2/components/filter/filter.js#L565).

Also, I think you should be able to use the JS property to set the `skeleton`:
e.g. `element.skeleton = true`

components/collapsible-panel/collapsible-panel.js
GZolla: I was planning to make a single PR for the documentation changes of this PR and [the new slot](https://github.com/BrightspaceUI/core/pull/3599).

components/collapsible-panel/collapsible-panel.js
GZolla: I spoke with @geurts and we will set skeleton only on the summary items slotted, other components are left for the consumers to handle.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -373,6 +376,12 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 
 	_handleSummarySlotChange(e) {
 		const content = e.target.assignedNodes({ flatten: true });
+		for (const element of content) {
+			if (element.tagName !== 'D2L-COLLAPSIBLE-PANEL-SUMMARY-ITEM') continue;
+			if (this.skeleton && !element.skeleton) element.skeleton = true;
+			else if (!this.skeleton && element.skeleton) element.skeleton = false;
```
oelhanafey, line 382: Nit: simplification
```suggestion
			if (this.skeleton) element.skeleton = true;
			else if (!this.skeleton) element.skeleton = false;
```

components/collapsible-panel/test/collapsible-panel.visual-diff.js
```diff
@@ -57,6 +57,7 @@ describe('d2l-collapsible-panel', () => {
 			{ name: 'default-large-padding-summary', selector: '#default-large-padding-summary' },
 			{ name: 'default-large-padding-expanded', selector: '#default-large-padding-expanded' },
 			{ name: 'default-long', selector: '#default-long' },
+			{ name: 'skeleton', selector: '#skeleton' },
```
oelhanafey, line 60: Can we also add a vdiff for the summary item skeleton?

oelhanafey (commented):
Looks good!

components/collapsible-panel/collapsible-panel.js
dbatiste: I am late to reading this thread, but I don't think we should do this. The light-DOM belongs to the consumer, so I think it should be up to them to set the skeleton attribute. In general I don't think we should be updating DOM outside our scope of ownership... though I know we do from time to time.

oelhanafey (approved):

PR #3583, opened 2023-05-01
Add skeletons to d2l-tabs

dbatiste:
[US151813](https://rally1.rallydev.com/#/?detail=/userstory/698666424603&fdp=true)

This PR adds skeletons to `d2l-tabs` and its internally rendered `d2l-tab-internal` component. It does not skeletize anything in the light-DOM (panel content or extension slot), since that is intentionally left to the consumer to skeletize.

image


components/tabs/test/tabs.visual-diff.html
```diff
@@ -48,6 +48,23 @@
...
+		<d2l-tabs id="skeleton" skeleton>
+			<d2l-tab-panel text="All">Tab content for All</d2l-tab-panel>
+			<d2l-tab-panel text="Biology">Tab content for Biology</d2l-tab-panel>
+			<d2l-tab-panel text="Physics">Tab content for Physics</d2l-tab-panel>
+			<d2l-tab-panel text="Chemistry">Tab content for Chemistry</d2l-tab-panel>
+		</d2l-tabs>
+	</div>
+
+	<div class="visual-diff">
+		<d2l-tabs id="ellipsis">
```
dbatiste, line 61: This is a new extra test I'm including here. A scenario we previously didn't have covered

svanherk:
What does it look like if they don't have the tabs/tab names yet? I'm thinking of the HM case where they need to wait on a call to get all the department names, for example.

svanherk:
> What does it look like if they don't have the tabs/tab names yet? I'm thinking of the HM case where they need to wait on a call to get all the department names, for example.

Not sure if this is an issue for the use case that brought this up, I'm specifically thinking of My Courses if we were to switch to using a skeleton (versus now, where it just shows a loading spinner until the tab names are returned).

dbatiste:
> > What does it look like if they don't have the tabs/tab names yet? I'm thinking of the HM case where they need to wait on a call to get all the department names, for example.
> 
> Not sure if this is an issue for the use case that brought this up, I'm specifically thinking of My Courses if we were to switch to using a skeleton (versus now, where it just shows a loading spinner until the tab names are returned).

This is what it currently looks like without any text... basically just a dot, which obviously isn't what they want. I could maybe try applying a min-width when in skeleton mode, but if the consumer wants something specific, I think they need to put some placeholder/temp text in. Do you know if they do that elsewhere?

image

svanherk:
> > > What does it look like if they don't have the tabs/tab names yet? I'm thinking of the HM case where they need to wait on a call to get all the department names, for example.
> > 
> > 
> > Not sure if this is an issue for the use case that brought this up, I'm specifically thinking of My Courses if we were to switch to using a skeleton (versus now, where it just shows a loading spinner until the tab names are returned).
> 
> This is what it currently looks like without any text... basically just a dot, which obviously isn't what they want. I could maybe try applying a min-width when in skeleton mode, but if the consumer wants something specific, I think they need to put some placeholder/temp text in. Do you know if they do that elsewhere?
> 
> image

The places I could think of were My Courses, and the old (no longer used) labs Filter would've run into something similar. But there's no plan to switch to a skeleton, just that if they did it wouldn't "just work". Looking at Tristan's user story, it seems the tabs are users, which _may_ have the same HM loading issue, may not.

tdevos-d2l:
Something else to consider: In consistent-evaluation part of the reason we wanted a skeleton for tabs was to not show which tab was `selected` until after the skeleton is cleared. This is because part of us loading the page is deciding which tab we will automatically select for the user initially. Our designer's original design for tab skeletons did not show the tab selector in the skeleton state. The way you have it should be better than it was before, but something for you to consider if it's easy to add

tdevos-d2l:
> Looking at Tristan's user story, it seems the tabs are users, which _may_ have the same HM loading issue, may not.

FWIW, in _our_ implementation we don't render the tabs until we have the names loaded so this won't be the case for us

dlockhart (approved):

dbatiste:
Looks like I can apply a `min-width` and use a `&nbsp;` in skeleton mode to achieve the desired effect if the consumer doesn't provide any text. That's probably a worthwhile addition.

> In consistent-evaluation part of the reason we wanted a skeleton for tabs was to not show which tab was selected until after the skeleton is cleared.

Tabs doesn't have the concept of not having a tab selected. If the consumer doesn't specify a selected tab, the first tab will be selected automatically. Having said that, @geurts , what are your thoughts on hiding the selected tab indicator in skeleton mode? And do you think that the tabs should be in-operable while in skeleton mode? Currently they are operable, and for some usages, that could actually be desirable if consumers set up their panel skeletons to provide sufficient hint of the content being loaded. It starts to get muddy with disabled behaviour creeping in (which tabs do not support). If we leave them operable, I think we should leave the selected indicator as-is.

PR #3586, opened 2023-05-02
Announce introductory text when opening single dimension

oelhanafey:
[DE52350](https://rally1.rallydev.com/#/?detail=/defect/690468457635&fdp=true)

Previously, the introductory text in filter was not very accessible since it can appear before any other focusable elements which makes it easy to miss for screen reader users. This PR changes the filter component to announce the introductory text on dimension open. 

Carin and Jeff tested this out and are happy with where it ended up.


svanherk (approved):
Gave the demo page a spin with JAWS and NVDA, seems to be working well!

PR #3594, opened 2023-05-03
Increase CSS specificity of list extend-separator styles

oelhanafey:
[DE53209](https://rally1.rallydev.com/#/?detail=/defect/699523809885&fdp=true)

This extend-separators style was being overridden by a separate RTL style due to CSS specificity. Added a check for RTL to increase the specificity. This was preventing non-selectable list-items in lists with `extend-separators` from being tabbed forward in RTL.

Before:
Screen Shot 2023-05-03 at 3 41 50 PM

After:
Screen Shot 2023-05-03 at 3 45 53 PM

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -169,7 +169,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			.d2l-list-item-content-extend-separators > [slot="control"] {
 				width: 3rem;
 			}
-			.d2l-list-item-content-extend-separators > [slot="content"] {
+			.d2l-list-item-content-extend-separators > [slot="content"],
+			:host([dir="rtl"]) .d2l-list-item-content-extend-separators > [slot="content"] {
 				padding-left: 0.9rem;
 				padding-right: 0.9rem;
 			}


components/list/list-item-mixin.js
```diff
@@ -169,7 +169,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			.d2l-list-item-content-extend-separators > [slot="control"] {
 				width: 3rem;
 			}
-			.d2l-list-item-content-extend-separators > [slot="content"] {
+			.d2l-list-item-content-extend-separators > [slot="content"],
+			:host([dir="rtl"]) .d2l-list-item-content-extend-separators > [slot="content"] {
```
oelhanafey, line 173: This case was missed because we don't run RTL versions of our list visual diff tests. It might be beneficial to run our tests in both directions similar to our other components to catch defects like this in the future. 

There are a lot of tests for the component though which is likely why it wasn't done in the first place. Maybe this could be done as we move these tests to the new visual diff framework since they run much faster.

dbatiste: Yeah, I think RTL tests were created for only a select few of the list tests. I find them difficult to maintain so I've long wanted to just re-do all of the tests (inc the unit tests) for the list, but it's a pretty big task given the scope of what can be done with it.

dbatiste (approved):

PR #3597, opened 2023-05-04
Add header-text to filter-dimension-set

oelhanafey:
[US148609](https://rally1.rallydev.com/#/?detail=/userstory/685391079311&fdp=true)
[Style specs](https://www.figma.com/proto/4sRGPOeVbaXHb77zh17KCv/Filter-%3E-Enhancements-for-Wave?type=design&page-id=0%3A1&node-id=714-1384&starting-point-node-id=2%3A48&scaling=min-zoom)

This PR adds the header-text property to filter-dimension-set. This will display a header for the list when the search is empty to add additional list context. Searching will remove the list header.


components/filter/filter.js
```diff
@@ -487,6 +492,15 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			if (count === 0) return searchResults;
 		}
 
+		let listHeader = nothing;
+		if (dimension.headerText && dimension.searchValue === '') {
+			listHeader = html`
+				<d2l-list-item>
+					<h4 class="d2l-heading-4 list-header-text">${dimension.headerText}</h4>
+				</d2l-list-item>
```
oelhanafey, line 500: This is rendered within the list because of the future work required to add the `selected-first` property in [US148608](https://rally1.rallydev.com/#/?detail=/userstory/685391076375&fdp=true). That will require us to render list items above and below the header.

components/filter/filter.js
```diff
@@ -487,15 +494,30 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			if (count === 0) return searchResults;
 		}
 
+		let listHeader = nothing;
+		let offscreenHeader = nothing;
+		if (dimension.headerText && dimension.searchValue === '') {
+			listHeader = html`
+				<d2l-list-item>
+					<h4 class="d2l-heading-4 list-header-text" aria-hidden="true">${dimension.headerText}</h4>
```
oelhanafey, line 502: After discussion, to make the header text more readable, I added an offscreen header that is rendered outside of the list. I added an `aria-hidden` here when `header-text` is set and `selected-first` is not since the header will be at the top and does not need to be listed twice in the a11y DOM. I think we do still want to render the header in the list in this case so the appearance is consistent.

When selected-first is added, the plan is to make the offscreen header read something like `${headerText}. Selected items at top.` and remove the `aria-hidden` on the list header in that case since we want the header in the a11y DOM to break up the list into 2 parts. The list header may not be read depending on how you are navigating, so the offscreen header will work as a backup which is read outside of the list.

components/filter/filter.js
```diff
@@ -164,6 +164,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				padding: 0.9rem;
 			}
 
+			.list-header-text {
+				color: var(--d2l-color-ferrite);
+				margin: 0;
+				padding-bottom: 0.05rem;
+				padding-top: 0.65rem;
```
oelhanafey, line 171: Design spec requires `1.2rem` top and `0.6rem` bot padding. By default the list item is giving `0.55rem` padding at the top and bottom.

As of writing this, the spec in Figma has not yet been updated to reflect this.

components/filter/filter.js
```diff
@@ -736,8 +758,10 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 
 			switch (type) {
 				case 'd2l-filter-dimension-set': {
+					info.headerText = dimension.headerText;
 					info.introductoryText = dimension.introductoryText;
 					info.searchType = dimension.searchType;
+					info.searchValue = '';
```
oelhanafey, line 764: This was previously not being initialized

components/filter/filter-dimension-set.js
```diff
@@ -11,6 +11,11 @@ class FilterDimensionSet extends LitElement {
 
 	static get properties() {
 		return {
+			/**
+			 * The header text to display as the list header
```
svanherk, line 15: Maybe more of a Jeff question, but do we want to give a bit of insight here on when to use this? Similar to the introductory test, it takes up space and shouldn't really be needed unless you need to provide that extra context because searching will provide different options?

oelhanafey: Yeah, it would be good to add some additional context. The only use case of the header text at this point is for when the filter displays different lists of content when searching and not searching. Though I'm unsure if this is the only case this should be used. I'll check with Jeff.

components/filter/filter.js
svanherk: Gave it a try with NVDA and JAWS - they won't hit the offscreen `h4` at all if navigating with tab. If they navigate with arrow keys only, they'll hear it in single-dimension mode - it's a bit awkward though because then the list won't work properly unless they know to turn on Forms/Focus mode. In multi-dimension mode they won't hear it, I _think_ because of [DE52188](https://rally1.rallydev.com/#/15545167705ud/search?detail=%2Fdefect%2F686309407849).

If they try to find it based on the headings shortcut that works, but it's maybe a tad confusing because the other headings on the page aren't hidden when the filter is opened. Did Carin also play around with this stuff? Maybe she's navigating in a way I don't know about

oelhanafey: Right, if you navigate with tabbing then the issue will be the same. Even in our ideal case where we have 2 lists with labels, I don't think those labels would get read if you are tabbing. That being said, I'm not sure if screen reader users normally navigate with tabs, it seems like that would end up skipping information in most situations. 

The main idea with adding the header above the list was so that users would encounter the header text before entering the list, where they may begin navigating with arrow keys which would skip over the header. Carin hasn't had a look at this, it may be good to get her opinion and see how she would navigate.

svanherk: > Even in our ideal case where we have 2 lists with labels, I don't think those labels would get read if you are tabbing

Hmm, I wonder if that's a VoiceOver thing? If I add a `label` attribute to `d2l-list` and pass that to `aria-label` on the [`div` that sets the `role`](https://github.com/BrightspaceUI/core/blob/main/components/list/list.js#L123), and pass the `header-text` in there instead, then NVDA and JAWS will read that header text instead of "application" each time I tab to the list.

svanherk: >  That being said, I'm not sure if screen reader users normally navigate with tabs, it seems like that would end up skipping information in most situations.
> Carin hasn't had a look at this, it may be good to get her opinion and see how she would navigate.

Yeah, this might be one of those cases where how I'm testing doesn't match how a real user would behave

components/filter/filter-dimension-set.js
oelhanafey: Jeff and I came up with an update for the description. It was difficult to generalize but we didn't want to specify it as only being for the case that Wave is using it for.

We could have added some additional insight on when to use it to the documentation page for the component but we thought that may encourage more users to use the property, which we don't want, so opted for the bulky and slightly vague description.

components/filter/filter.js
oelhanafey: This is now using the header text as a list label which seems to work nicely with a screenreader. The header itself is just used for presentation and is given `aria-hidden="true"`. 

I don't think there is a need for this to be navigable by header in the screenreader so this seems okay.

For selected first we will say: `${headerText}. Selected filters appear first.`

components/filter/filter.js
```diff
@@ -487,15 +494,26 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
+		}
+
 		return html`
 			${searchResults}
 			<d2l-list
 				id="${SET_DIMENSION_ID_PREFIX}${dimension.key}"
 				@d2l-list-selection-change="${this._handleChangeSetDimension}"
 				extend-separators
 				grid
+				label="${ifDefined(dimension.headerText)}"
```
svanherk, line 513: Kinda weird, but to be accurate to Wave's case, should we _not_ set this if a search has been done (like we do for the visual header)?  Since the search results might _not_ be promoted by the company?

oelhanafey: Yeah, you are correct. It's a bit weird but the label should not be there when searching

svanherk (approved):
🎉

PR #3599, opened 2023-05-04
feat: Add slot `before` to Collapsible Panel

GZolla:
Adding a slot in the header of collapsible panel. The slot currently aligns with both the panel title and the existing header slot.

Rally: [US151931](https://rally1.rallydev.com/#/?detail=/userstory/699017182311&fdp=true)


components/collapsible-panel/collapsible-panel.js
```diff
@@ -171,6 +185,9 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 				display: flex;
 				margin-inline-start: var(--d2l-collapsible-panel-spacing-inline);
```
dbatiste, line 186: Any chance you can sneak in this extra little layout fix since you're working on the layout? Currently the text is allowed to flow right to the edge.

I think it might just be a matter of adding `margin-inline-end: var(--d2l-collapsible-panel-spacing-inline);` for this `.d2l-collapsible-panel-header-secondary` element. It would end up being `0.9rem`/`18px;`. Then it will match up wit the content.

image

components/collapsible-panel/collapsible-panel.js
```diff
@@ -364,6 +382,12 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 		}
 	}
 
+	_handleHeaderStartSlotChange(e) {
```
dbatiste, line 385: I'm tempted to suggest naming this `_handleHeaderBeforeSlotChange` to be consistent with the "before" slot name. It almost makes it sounds like it's for a fictitious event that is dispatched before the slot change happens. 🤷‍♂️ I'm content either way.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -77,6 +77,7 @@ class CollapsiblePanel extends FocusMixin(RtlMixin(LitElement)) {
 			noSticky: { attribute: 'no-sticky', type: Boolean },
 			_focused: { state: true },
 			_hasSummary: { state: true },
+			_hasBefore: { state: true },
```
dbatiste, line 80: Nit: move this before `_hasSummary`.

dbatiste (approved):

PR #3605, opened 2023-05-04
Add check that next focusable list item belongs to the same list

oelhanafey:
[DE53222](https://rally1.rallydev.com/#/?detail=/defect/699633940503&fdp=true)

When navigating a list with arrow keys, the list item will try to focus on the next/previous list item sibling. If that list item is not focusable, it will skip the list item and focus the next focusable element.

The list does not check that the next focusable belongs to the same list. If that non-focusable element is at the start or end of the list, you are able to arrow outside of the list and not able to get back into it.

This PR adds a check to confirm that the next focusable element belongs to the same list as the currently focused element.

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -302,7 +302,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			if (!listItem._tryFocus()) {
 				// ultimate fallback to generic method for getting next/previous focusable
 				const nextFocusable = previous ? getPreviousFocusable(listItem) : getNextFocusable(listItem);
-				if (nextFocusable && this._isContainedInSameList(curListItem, nextFocusable)) {
+				if (nextFocusable && this._isContainedInSameRootList(curListItem, nextFocusable)) {
 					nextFocusable.focus();
 				}
 			}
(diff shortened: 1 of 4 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -442,6 +444,11 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			this.shadowRoot.querySelector(`.d2l-cell[data-cell-num="${this._cellNum}"]`);
 	}
 
+	_isContainedInList(node1, node2) {
+		const list = findComposedAncestor(node1, node => node.tagName === 'D2L-LIST');
```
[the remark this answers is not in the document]
oelhanafey: Implemented your suggestion and tested in various nested list cases. It seems to be working as expected

components/list/list-item-generic-layout.js
```diff
@@ -301,7 +302,9 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			if (!listItem._tryFocus()) {
 				// ultimate fallback to generic method for getting next/previous focusable
 				const nextFocusable = previous ? getPreviousFocusable(listItem) : getNextFocusable(listItem);
-				if (nextFocusable) nextFocusable.focus();
+				if (nextFocusable && this._isContainedInSameList(curListItem, nextFocusable)) {
```
dbatiste, line 305: Maybe we should name this `_isContainedInSameRootList` to make it clearer the scope of he check?

dbatiste (approved):
Looks good!... maybe just a different name for that internal function, but not a huge deal.

components/list/list-item-generic-layout.js
oelhanafey: Yep, I like that 👍

dbatiste (approved):

PR #3610, opened 2023-05-05
Spike into visual-diffs

dlockhart:
This isn't intended to ever merge, but goes alongside [this spike results document](https://desire2learn.atlassian.net/wiki/spaces/GAUD/pages/3587080546/Visual-Diff+Spike+Results) to show how things could look in practice.

Ultimately, a lot of this will get broken out into a new `@brightspace-ui/testing-helpers` library.


components/button/test/button-icon.vdiff.js
```diff
@@ -0,0 +1,33 @@
+import '../button-icon.js';
+import { fixture, focusWithKeyboard, focusWithMouse, hoverWithMouse, html, screenshotAndCompare } from '../../../tools/web-test-runner-helpers.js';
```
dlockhart, line 2: Imagine these being imported from `@brightspace-ui/testing-helpers`.

components/icons/icon.js
```diff
@@ -6,8 +6,9 @@ import { loadSvg } from '../../generated/icons/presetIconLoader.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 import { runAsync } from '../../directives/run-async/run-async.js';
 import { unsafeSVG } from 'lit/directives/unsafe-svg.js';
+import { WaitForMeMixin } from '../../tools/web-test-runner-helpers.js';
```
dlockhart, line 9: Example of a component that does asynchronous work (dynamically importing the icon) that we want to wait for before visual-diffing.

tools/visual-diff-plugin.js
dlockhart, line 1: This is the `@web/test-runner` plugin that consumers don't need to know about and will likely be buried in `@brightspace-ui/testing-helpers`.

tools/visual-diff-reporter.js
dlockhart, line 1: This doesn't do anything really, but is an example of a custom reporter. This is where we'd implement the generation of our HTML report files.

tools/web-test-runner-helpers.js
dlockhart, line 1: Most if not all of this will live in `@brightspace-ui/testing-helpers`.

tools/web-test-runner-helpers.js
```diff
@@ -1,12 +1,172 @@
...
+	mouseResetNeeded = false;
+
+let awaitedFonts = false;
+async function waitForFonts() {
+	if (awaitedFonts) return Promise.resolve();
+	awaitedFonts = true;
+	return await document.fonts.ready;
+}
+
+export const WaitForMeMixin = dedupeMixin(superclass => class extends superclass {
```
dlockhart, line 26: This will likely be defined in `@brightspace-ui/core`.

web-test-runner.config.js
```diff
@@ -20,8 +38,47 @@ export default {
 					}
 				})
 			]
+		},
```
dlockhart, line 41: I think we'll want to take all of this and turn it into a shared config that consumers can just import from `@brightspcae-ui/testing-helpers`.

svanherk:
Rerunning the jobs here to test the new `configure-aws-credentials` version

tools/web-test-runner-helpers.js
```diff
@@ -1,12 +1,172 @@
...
+
+	if (awaitNextFrame) {
+		await nextFrame();
+	}
+
+}
+
+export const screenshotAndCompare = async(elem, name, opts) => {
+
+	const rect = elem.getBoundingClientRect();
```
svanherk, line 168: We currently do some fancy logic to handle getting the proper `rect` for dropdowns (and filter, etc.): https://github.com/BrightspaceUI/core/blob/main/components/dropdown/test/dropdown-helper.js#L28-L44
Do we see that being added here, or providing users a way to override? Or should they create a "large enough" wrapper element?

dbatiste: Making a wrapper element that is "large enough" is the simplest, although adjusting the hard-coded dimensions any time the contents change is a bit of an annoyance. It can provide some degree of buffer when contents change that would normally leave us without a diff due to the size changing, though my understanding is that we're going to improve that aspect. I tend to like being able to override with something more automatic like this.

dbatiste: Having said that, the consumer could always have some script that resizes the wrapper before the screenshot.

dlockhart: What if they could do something like this in `fixture`:

```javascript
const elem = await fixture(html`<d2l-my-elem>...</d2l-my-elem>`, {
  margin: {
    blockEnd: 200,
    inlineStart: 50,
    inlineEnd: 75
  }
});
```

Then we'd reserve `200px` at the bottom, `50px` to the left and `75px` to the right in the form of padding on the body (we default to `30px` today). Alternately we could just put a huge amount of padding on the body -- like `500px` -- but then the viewport size would need to change.

So that would reserve the space initially as it renders. Then when they take the screenshot maybe it can reuse that value when it captures the rectangle? `screenshotAndCompare` already supports a single-value "margin" option (defaults to `10px` all around).

dbatiste: I think that would work if the consumer hard-codes it.

Seems like that wouldn't work if the consumer wants to avoid hard-coding - if they wanted to measure the opener + the content to figure out the size needed (like the helper Stacey linked), it needs to have access to the element, but it won't exist yet. 🐔 & 🥚 ... 😄 . It would need a way to adjust margin after it's been rendered. Or, for the rare cases where a consumer might want this, they just include a wrapper div in their fixture and they set the dimensions of it based on measurement just before doing the screenshot.

svanherk:
@dlockhart Should be good to close now!

PR #3618, opened 2023-05-09
Add selected-first to filter-dimension-set

oelhanafey:
[US148608](https://rally1.rallydev.com/#/?detail=/userstory/685391076375&fdp=true)

This PR adds the `selected-first` property which render selected items at the top of the filter dimension.
Screen Shot 2023-05-09 at 12 33 22 PM

The property renders the filter dimension as a single list with both the selected and unselected items. The selected items will be rendered above the header text if it is present. Selected items will be at the top both when searching and when the search is empty and they are recalculated on dropdown open (or dimension open in the multi-dim case), and when searching.

**For Testing**
On the filter test page, the `Single Set Dimension - Header Text and Selected First` demo and the `Multiple Dimensions` demo in the course dimension are using the new `selected-first` property.


svanherk:
> In this approach, we use 2 lists, one for the selected items and one for the non-selected items. This lets us have the header be rendered outside of the list so this may be more semantically correct. One consideration here is the select-all needs to connect to a single list. After chatting with Jeff, to make this work we would need to refactor this to have 2 separate select-all components in the filter content area which may not be worth the effort.

I'm surprised `d2l-selection-select-all` and `d2l-selection-summary` can't hook up to multiple lists, given they accept a `selection-for` property.  But looks like that currently expects only one subscriber's id, so we'd probably need to move that to the `SubscriberController`s to get the multiple subscriber stuff for free.

svanherk:
> Question: Could someone on Windows check if the list item counts are being read out on JAWS and NVDA when the search value is empty?

Yup!

So both NVDA and JAWS are not reading the list count anyways! Same thing as VoiceOver - I verified with the list demo that they do when grid mode is off, but not when grid mode is on.

One issue though - are you currently expecting the list header to be read out? Both Jaws and NVDA are not able to access it. That might be one place where the separate lists could help, with an `aria-decribedby` of the header for the second one and something like "selected items" for the first.  I'll go play around with the other PR in JAWS and NVDA.

I also noticed that the first time you open the single dimension list, the focus goes to the first unselected item.  But on subsequent opens, it goes to the first selected item. Not sure which we want, but I'm assuming it's a timing issue the first time we calculate the selected items.

oelhanafey:
> I'm surprised d2l-selection-select-all and d2l-selection-summary can't hook up to multiple lists, given they accept a selection-for property. But looks like that currently expects only one subscriber's id, so we'd probably need to move that to the SubscriberControllers to get the multiple subscriber stuff for free.

Ah interesting, yeah if we use `SubscriberController` then we might be able to refactor to get the behaviour we want in the double lists.

> One issue though - are you currently expecting the list header to be read out? Both Jaws and NVDA are not able to access it. That might be one place where the separate lists could help, with an aria-describedby of the header for the second one and something like "selected items" for the first. I'll go play around with the other PR in JAWS and NVDA.

Interesting, as in you can't navigate to it with your cursor? Or just when you're arrowing/tabbing? On VO, it appears as a header in the rotor menu and if you navigate with the cursor sequentially you will run into the header. But that's a good point about having separate labels for the 2 lists. Surprised we don't have any aria labels or descriptions in d2l-list but those might be nice to add and would be a clear benefit to the 2 list approach.

>I also noticed that the first time you open the single dimension list, the focus goes to the first unselected item. But on subsequent opens, it goes to the first selected item. Not sure which we want, but I'm assuming it's a timing issue the first time we calculate the selected items.

Yeah, this is a separate defect [DE53191](https://rally1.rallydev.com/#/?detail=/defect/699354295129&fdp=true). The filter should be focusing the first focusable. But I will see if I can at least make the focus consistent

svanherk:
> Interesting, as in you can't navigate to it with your cursor? Or just when you're arrowing/tabbing? On VO, it appears as a header in the rotor menu and if you navigate with the cursor sequentially you will run into the header. But that's a good point about having separate labels for the 2 lists. Surprised we don't have any aria labels or descriptions in d2l-list but those might be nice to add and would be a clear benefit to the 2 list approach.

Yeah if you want we can Zoom and I can show you, but I find VoiceOver works totally different than NVDA and JAWS. Once we're in the list in "forms/focus" mode, none of the shortcuts work. It's only what you can get to with tabbing and the arrow keys. Technically they could turn off forms/focus mode and find it, but not sure if that is an expected workflow or not. Maybe, because that's also the only way to see the bracket counts 🤔

(I have no idea what you mean by "rotor menu" lol)

components/filter/filter.js
```diff
@@ -505,6 +517,56 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
+					?hidden="${item.hidden}"
+					key="${item.key}"
+					label="${item.text}"
+					selectable
+					?selected="${item.selected}">
+					<div class="d2l-filter-dimension-set-value d2l-body-compact">
+						<div class="d2l-filter-dimension-set-value-text">${item.text}</div>
+						${item.count !== undefined ? html`<div class="d2l-body-small">(${formatNumber(item.count)})</div>` : nothing}
+					</div>
+				</d2l-list-item>
```
svanherk, line 536: Looks like all three of these are the same and only needs access to the `item`, so would be good to have a helper function like `_createSetDimensionItem(item)` or something.

components/filter/filter-dimension-set.js
```diff
@@ -120,6 +126,7 @@ class FilterDimensionSet extends LitElement {
 				disabled: value.disabled,
 				key: value.key,
 				selected: value.selected,
+				selectedOnRender: false,
```
svanherk, line 129: Is this needed? Seems like this is only set and controlled by the filter itself, and the value isn't used before it's initialized to the proper thing

components/filter/filter.js
```diff
@@ -882,6 +938,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		});
 	}
 
+	_setSelectedOnRender(dimension) {
```
svanherk, line 941: I'm wondering if `setInitiallySelected` and `initiallySelected` is a little clearer? If that's not accurate, then I still don't know what `selectedOnRender` means lol.  Or maybe `saveCurrentlySelected` and `currentlySelected`?

oelhanafey: Maybe `currentlySelected` could work but it's not exactly accurate since it saves the state it was selected. It used to be `selectedOnOpen` but after changes that was no longer accurate. The property will save whether a value was selected when first opening the dimension and when searching. 

We didn't want the selected items to bubble to the top when you click on them so it only recalculates when you re-render the full list due to searching.

svanherk: `shouldBubble` and `_updateSetDimensionShouldBubble`? Or maybe that's weird lol

oelhanafey: I like that better! Decouples the logic from the name so if we change up when we bubble in the future, the variable name is still accurate. Maybe just `_updateDimensionShouldBubble` to shorten it, similar to `_isDimensionEmpty()` or `_buildDimension()`

components/filter/filter.js
```diff
@@ -917,6 +951,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		subscribers.forEach(subscriber => subscriber.updateActiveFilters(this.id, this._activeFilters));
 	}
 
+	_updateDimensionShouldBubble(dimension) {
+		for (const value of dimension.values) {
+			if (value.selected) value.shouldBubble = true;
+			else value.shouldBubble = false;
```
svanherk, line 957: Minor: This could technically be something like `value.shouldBubble = !!value.selected;`, but I don't mind it like this either!

components/filter/filter.js
```diff
@@ -505,6 +517,56 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			listLabel = dimension.headerText;
 		}
 
+		let selectedListItems = nothing;
+		let listItems = nothing;
+		if (dimension.selectedFirst) {
+			if (listLabel) listLabel = this.localize('components.filter.selectedFirstListLabel', { headerText: dimension.headerText });
```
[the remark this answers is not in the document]
svanherk: If there's no header text, do we need to tell users the selected items are first?

svanherk (approved):
Looks good!

components/filter/filter.js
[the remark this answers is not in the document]
oelhanafey: I don't think so. The reason we are adding that information to the label is specifically to differentiate the selected items from the items below the header text since the selected items are displayed above the header text and may be unrelated.

If there is no separation of the list, then I don't think that clarification needs to be made and the user would be able to find out which items are selected when navigating the list.

components/filter/filter.js
oelhanafey: Yep, I don't think it needs to use `!!` since `selected` is already a boolean.

PR #3619, opened 2023-05-09
Selected-first double list for filter-dimension-set

oelhanafey:
This PR shows how selected-first would look as two lists.

**For Testing**
On the filter test page, the single set dimension demo and the multiple dimension demo in the course dimension are using the new `selected-first` property


svanherk:
Yeah, if we go with this one, I think we'd need to add labels to the list - something like "List: Selected Filters" and "List: Skills Promoted at Your Company" (basically "List: <whatever their header is>").

I haven't taken a look at the code yet to see which is more elegant, but the keyboard interaction is very different between the two.  When it's all one list, you arrow key around as if it's one list, and tab takes you back to the top.  But when it's two lists, you need to tab to get from the first list to the second list.  So if you have one selected item, pressing the arrow keys does nothing and you tab to get to the next list. That could be super unintuitive, or super helpful if you want to go between selected and unselected options quickly.  This is probably a @geurts question, but I think that'll be a big deciding factor

oelhanafey:
We have decided to go with the single list approach in #3618.

We weren't very with the keyboard navigation in the double list approach which requires you to tab to navigate between the two lists rather than being able to use arrow keys.

We also are not sure at this time if we would like to add an aria label to the d2l-list component which was the main benefit of going with the two list approach.

PR #3626, opened 2023-05-11
Focus first focusable in dropdown on initial focus with focus-trap

oelhanafey:
[DE53191](https://rally1.rallydev.com/#/?detail=/defect/699354295129&fdp=true)

Filter had a defect where focus was being put in the list rather than on the first focusable element. This was actually stemming from dropdown where the focus trap was focusing the first focusable element in the default slot on initial focus rather than the first focusable element in the entire dropdown. This caused any focusable elements in the header slot to not receive the initial focus.

This PR fixes it so that it will instead focus the first focusable element from the entire dropdown. This issue is only present with the focus-trap, dropdown content without focus-trap already has this behaviour.


dbatiste (approved):
I'm kinda surprised that we didn't target a higher level element from the beginning. Almost makes me wonder if it was intentional. Could also have been an oversight, since focus-trap and the header were added at different times.

svanherk:
The visual-diffs here are great, you can clearly see the fix for filter in all scenarios!

PR #3630, opened 2023-05-12
US152487 - Fix flaky sticky tests for d2l-collapsible-panel

svanherk:
Applying a similar fix to what we did for the `list` `sticky` tests


dlockhart (approved):

components/collapsible-panel/test/collapsible-panel.visual-diff.js
```diff
@@ -47,7 +55,6 @@ describe('d2l-collapsible-panel', () => {
 			{ name: 'default', selector: '#default' },
 			{ name: 'default-focus', selector: '#default', action: focusElement },
 			{ name: 'default-expanded', selector: '#default-expanded' },
-			{ name: 'default-expanded-sticky', selector: '#default-expanded-sticky' },
```
svanherk, line 50: The `sticky` tests need `captureBeyondViewport: false`, so just moving them to their own test block

components/collapsible-panel/test/collapsible-panel.visual-diff.html
```diff
@@ -25,9 +25,11 @@
 			padding: 1rem;
 			width: 600px;
 		}
+		.visual-diff.inline {
+			padding: 0;
+		}
 		.visual-diff.no-overflow {
 			overflow: hidden;
-			padding: 0;
```
svanherk, line 30: `overflow: hidden` doesn't work with `sticky`, so just applying it to the `focus` tests that need it to catch a regression from https://github.com/BrightspaceUI/core/pull/3326

oelhanafey (approved):
Thanks for fixing this Stacey, and the scrolled tests are a nice addition!

PR #3632, opened 2023-05-12
Update ListItemCheckboxMixin so that Enter key does not toggle the checkbox.

dbatiste:
[DE53334](https://rally1.rallydev.com/#/?detail=/defect/700493301447&fdp=true)

This PR updates the list-item checkboxes so that `Enter` key does not toggle the checkbox. This is to make them behave the same as native checkboxes.

### DIFF
--- a/components/list/test/list-item-checkbox-mixin.test.js
+++ b/components/list/test/list-item-checkbox-mixin.test.js
@@ -88,7 +88,6 @@ describe('ListItemCheckboxMixin', () => {
 	});
 
 	describe('Dispatches custom event when action area is selected', () => {
-		const actions = [ 'click' ];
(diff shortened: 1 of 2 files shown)


components/list/test/list-item-checkbox-mixin.test.js
```diff
@@ -88,7 +88,7 @@ describe('ListItemCheckboxMixin', () => {
 	});
 
 	describe('Dispatches custom event when action area is selected', () => {
-		const actions = [ 'click', 'enter' ];
+		const actions = [ 'click' ];
```
dlockhart, line 91: Looks like there's some code down below that can be cleaned up now.

PR #3640, opened 2023-05-16
DE53314 divider handle mobile on enter and space

mdulat:
[DE53314](https://rally1.rallydev.com/#/?detail=/defect/700334191725&fdp=true): Magnifying the page causes the adjustable split view separator element to be unable to be triggered by enter/spacebar

We can get keyboard input on mobile when using a desktop with small screensize (or zoomed in which gives mobile view). In this case, we make ENTER or SPACE on the divider handle work the same as a click (i.e. collapse entirely, or restore to original position).

### DIFF
--- a/templates/primary-secondary/primary-secondary.js
+++ b/templates/primary-secondary/primary-secondary.js
@@ -24,7 +24,9 @@ const keyCodes = Object.freeze({
 	LEFT: 37,
 	UP: 38,
 	RIGHT: 39,
-	DOWN: 40
+	DOWN: 40,
+	ENTER: 13,
+	SPACE: 32
 });
 
 function isMobile() {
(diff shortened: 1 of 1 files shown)


templates/primary-secondary/primary-secondary.js
```diff
@@ -303,27 +305,36 @@ class MobileKeyboardResizer extends Resizer {
...
+		if (e.keyCode !== keyCodes.UP && e.keyCode !== keyCodes.DOWN && e.keyCode !== keyCodes.ENTER && e.keyCode !== keyCodes.SPACE) {
 			return;
 		}
 		let secondaryHeight;
-		if (this.panelSize === 0) {
-			if (e.keyCode === keyCodes.UP) {
-				secondaryHeight = this.contentBounds.minHeight;
+		if (e.keyCode === keyCodes.ENTER || e.keyCode === keyCodes.SPACE) {
+			if (this.panelSize === 0) {
+				secondaryHeight = this.restoreSize || this.contentBounds.minHeight;
```
dlockhart, line 314: I think this should be `this._restoreSize`? (same in the `else`)

dlockhart (approved):
This looks good to me and it makes sense that it does the same as clicking. Thanks!

PR #3642, opened 2023-05-17
Remove offscreen header from filter

oelhanafey:
Removes offscreen header to test accessibility. PR for spinning up test site

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -555,7 +555,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		if (dimension.headerText && dimension.searchValue === '') {
 			listHeader = html`
 				<d2l-list-item>
-					<h4 class="d2l-heading-4 list-header-text" aria-hidden="true">${dimension.headerText}</h4>
+					<h4 class="d2l-heading-4 list-header-text">${dimension.headerText}</h4>
 				</d2l-list-item>
 			`;
 		}


components/list/list-item-generic-layout.js
```diff
@@ -542,7 +542,6 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 	}
 
 	_setFocusInfo(e) {
-		e.stopPropagation();
```
dbatiste, line 545: Looks like this was added in [this PR](https://github.com/BrightspaceUI/core/pull/1671/files) when nested lists were introduced. Seems like a couple of `stopPropagation()` calls are being made to prevent outer list items from appearing focused when nested list item elements have the focus. If you haven't check that, I think it would be worth doing some unit testing involving nested lists, though hopefully a visual-diff test would catch any regression for that.

oelhanafey: Yeah, I was testing to see if we liked the behaviour of this before diving too deep but you are correct. When a child list item is selected, it adds `_focusing` on all the parent list items as well. Interesting that it wasn't caught by visual-diffs.

So it seems like if we wanted to do something like this we would need to dispatch our own list-specific `focusin` event.

The announcing of the filter header on list focus works decently well, but not sure if this is a good practice for us to get in the habit of.

oelhanafey: Ah, good idea! Jeff is going to do some investigation on if there are other use cases where a label on the list may be appropriate, so if we can go with that then that is the ideal solution, but this can be a backup

oelhanafey:
Added the label property to list so this is not needed

PR #3648, opened 2023-05-18
reset scroll position between table visual diffs

dlockhart:
By adding tests for what grades is doing, those new grades tests were affecting existing tests because they scrolled the page around. This resets the window's scroll position before each test to try and make them more isolated.

This also sets a baseline for the grades tests (which will look somewhat broken) that can be fixed as part of #3635.

### DIFF
--- a/components/table/test/table-test-sticky-visual-diff.js
+++ b/components/table/test/table-test-sticky-visual-diff.js
@@ -477,6 +477,108 @@ class TestTableStickyVisualDiff extends LitElement {
 					</table>
 				</d2l-table-wrapper>
 			</div>
+			<div class="d2l-visual-diff">
+				<d2l-table-wrapper type="${type}" sticky-headers id="grades-row-header">
+					<table class="d2l-table">
+						<thead>
+							<tr>
+								<th rowspan="2" class="top" sticky>Name</th>
+								<th colspan="5" style="z-index: 3">
+									Category
+									<span style="position: relative;">
+										&nbsp;&#8964;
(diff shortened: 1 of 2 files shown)


components/table/test/table-test-sticky-visual-diff.js
```diff
@@ -477,6 +477,108 @@ class TestTableStickyVisualDiff extends LitElement {
 					</table>
 				</d2l-table-wrapper>
 			</div>
+			<div class="d2l-visual-diff">
```
dlockhart, line 480: This test simulates the modifications grades has had to do to make dropdowns within the row header appear "on top" of the second header row:

Screenshot 2023-05-18 at 9 25 08 AM

Note that it's a bit glitchy -- that's expected and is what #3635 will resolve. This sets the golden baseline.

components/table/test/table-test-sticky-visual-diff.js
```diff
@@ -477,6 +477,108 @@ class TestTableStickyVisualDiff extends LitElement {
...
+								<td>6/10</td>
+								<td>1/5</td>
+								<td>6/10</td>
+								<td>6/10</td>
+							</tr>
+						</tbody>
+					</table>
+				</d2l-table-wrapper>
+			</div>
+			<div class="d2l-visual-diff">
```
dlockhart, line 531: This test simulates the modifications grades has had to do to make dropdowns within the column header appear "on top" of the other rows below it:

Screenshot 2023-05-18 at 9 25 17 AM

dbatiste (approved):

PR #3650, opened 2023-05-18
Update `d2l-overflow-group` to not wrap items while resizing unless necessary.

dbatiste:
[DE53422](https://rally1.rallydev.com/#/?detail=/defect/701287303215&fdp=true)

This PR fixes an issue where items briefly wrap within the `d2l-overflow-group` during resizing, resulting in some FOUC. This happens because the browser is wrapping the items before the overflow-group chomping results are applied. 

The solution is to use use `flex-wrap: nowrap` except when we know items will wrap instead of being chomped. In practice, items are almost never expected to wrap because `min-to-show="0|1"` in the wild, and the `auto-show` attribute/`d2l-button-group-show` class is only applied to force show the item in the LMS for the primary button. Aside: we don't really support the wrapping edge case very well, but that's a separate issue.

### DIFF
--- a/components/overflow-group/overflow-group-mixin.js
+++ b/components/overflow-group/overflow-group-mixin.js
@@ -61,12 +61,9 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 				type: String,
 				attribute: 'opener-type'
 			},
-			_chompIndex: {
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Neat!

PR #3652, opened 2023-05-18
DE52736 - v2 of the Nested List Iterations Helper

svanherk:
The previous version combined all types of parents and children together, but this isn't very realistic and we wouldn't advise anyone to use our lists this way.  This new version is based off a mockup Jeff made to try to wrap his head around all the combinations we actually expect people to use.


components/list/demo/demo-list-nested-iterations-helper.js
```diff
@@ -0,0 +1,126 @@
...
+		border: solid 1px black;
+		margin: 1rem;
+		padding: 1rem;
+	}
+	d2l-list:not([slot="nested"]) > d2l-list-item > d2l-list-item-content {
+		color: blue;
+	}
+	d2l-list-item {
+		box-shadow: -1px 0 0 0 red;
+	}
```
svanherk, line 21: Sent to Jeff to see if he wants to keep this styling or not.

components/list/demo/demo-list-nested-iterations-helper.js
```diff
@@ -0,0 +1,126 @@
...
+			</tr>
+			<tr class="d2l-table-header">
+				<th scope="row">Non-Exp/Collapsible Children</th>
+				<td>${this._createList([option.parent, true], [option.child, false])}</td>
+				<td>${this._createList([option.parent, false], [option.child, false])}</td>
+			</tr>
+		`);
+
+		return html`
+			<d2l-table-wrapper>
```
svanherk, line 66: I'm thinking it might be better to _not_ use our table, to avoid these visual-diffs changing when we make table changes. Used them to get things started since they looked nice, but I could try to copy a subset of the styles. Thoughts?

svanherk: I could also offer a non-table version that just returns them all with spacing for visual-diff only

dbatiste: Yeah, I think my vote would be to isolate from other components.

dbatiste: Specifically for visual-diff.

dbatiste (approved):

components/list/demo/demo-list-nested-iterations-helper.js
dlockhart: +1. Seems like you could accomplish what you need with a native `<table>` and some simple CSS.

dlockhart (approved):

PR #3654, opened 2023-05-18
Fix d2l-overflow-group chomping in RTL.

dbatiste:
[DE50816](https://rally1.rallydev.com/#/?detail=/defect/668749858657&fdp=true)

This PR fixes the `d2l-overflow-group` chomping when in RTL. The root cause of this issue was that the calculation of the width of the items was incorrect - it was not correctly factoring in the size of the margin in RTL.

Due to the timing of when measurements are taken, it is necessary to test this by setting `dir="rtl"` before loading the page. If one tries to test simply by toggling to the Arabic language in the demo page, it will be using LTR measurements and not a valid test.

**Before:**
image

**After:**
image

Note: before my [previous change to how the items wrap](https://github.com/BrightspaceUI/core/pull/3650), this would have incorrectly wrapped below...

**Before:**
image

**After:**
image

### DIFF
--- a/components/overflow-group/overflow-group-mixin.js
+++ b/components/overflow-group/overflow-group-mixin.js
@@ -61,12 +61,9 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 				type: String,
 				attribute: 'opener-type'
 			},
-			_chompIndex: {
-				state: true
-			},
-			_mini: {
-				state: true
-			}
+			_chompIndex: { state: true },
+			_mini: { state: true },
+			_wrapping: { state: true }
 		};
 	}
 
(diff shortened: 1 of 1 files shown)


components/overflow-group/overflow-group-mixin.js
```diff
@@ -297,8 +296,8 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 				isChomped: false,
 				isHidden: itemHidden,
 				width: Math.ceil(parseFloat(computedStyles.width) || 0)
-					+ parseInt(computedStyles.marginRight) || 0
```
dbatiste, line 300: We need brackets here to properly calculate the width, otherwise the left margin (which is what is used in RTL) is not added.

PR #3656, opened 2023-05-19
Check that next focusable is a list item

oelhanafey:
[DE53222](https://rally1.rallydev.com/#/?detail=/defect/699633940503&fdp=true)

This is a follow-up to #3605. I added a check that it will only focus the next item if the next focusable belongs to the same list, but this doesn't account for other non list-item components that can be placed in the d2l-list like list-controls.

This PR fixes the check so it confirms that the next focusable is a list item and that list item belongs to the same list as the currently focused list item.

### DIFF
--- a/components/list/test/list.test.js
+++ b/components/list/test/list.test.js
@@ -3,7 +3,7 @@ import '../list-controls.js';
 import '../list-item.js';
 import '../list-item-button.js';
 import '../list-item-content.js';
-import { expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { expect, fixture, html, oneEvent, waitUntil } from '@open-wc/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
 const keyCodes = Object.freeze({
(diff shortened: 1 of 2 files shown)


components/list/test/list.test.js
```diff
@@ -136,6 +140,32 @@ describe('d2l-list', () => {
...
 			});
 		});
+
+		it('Focus stays in list with list controls when in grid mode when up arrow is pressed', async() => {
+			const elem = await fixture(html`
+				 <d2l-list grid>
+					<d2l-list-controls slot="controls"></d2l-list-controls>
+					<d2l-list-item key="L1-1" label="L1-1"></d2l-list-item>
+					<d2l-list-item selectable key="L1-2" label="L1-2"></d2l-list-item>
+				</d2l-list>
```
oelhanafey, line 150: Previously, pressing UP when focused on `L1-2` would focus the list controls and you would be unable to arrow back down into the list items.

dbatiste (approved):

PR #3657, opened 2023-05-19
Optimize d2l-overflow-group item MutationObservers

dbatiste:
[DE53433](https://rally1.rallydev.com/#/?detail=/defect/701335833197&fdp=true)

Previously, `slotchange` events would wire up new `MutationObserver`s to items, but it would not disconnect the old ones. This would result in duplicate observers observing the same element when slot changes occur after the initial rendering, and consequently, the handler being called multiple times for each item mutation.

This PR updates to use a single `MutationObserver` for all items within the `d2l-overflow-group`. The browser automatically ignores calls to observe elements that are already being observed by a given `MutationObserver`.

Note: unlike `ResizeObserver` and `IntersectionObserver` that had/have a leak, the browser apparently does properly clean up elements that are no longer referenced but are being observed by a `MutationObserver`. Some discussion on this can be found [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1220041#c12). So, I think we're ok just letting the browser handle removed items, which is what we were relying on previously when the page would unload.


components/overflow-group/overflow-group-mixin.js
```diff
@@ -90,10 +90,11 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 
 		this._handleItemMutation = this._handleItemMutation.bind(this);
 		this._handleResize = this._handleResize.bind(this);
+		this._itemObserver = new MutationObserver(this._handleItemMutation);
 		this._resizeObserver = new ResizeObserver((entries) => requestAnimationFrame(() => this._handleResize(entries)));
 
 		this._hasResized = false;
-		this._isObserving = false;
+		this._isObservingResize = false;
```
dbatiste, line 97: I decided to rename this to avoid any confusing, since now we have two different observers being used.

components/overflow-group/overflow-group-mixin.js
```diff
@@ -370,8 +371,7 @@ export const OverflowGroupMixin = superclass => class extends LocalizeCoreElemen
 			await this._getItems();
 
 			this._slotItems.forEach(item => {
-				const observer = new MutationObserver(this._handleItemMutation);
```
dbatiste, line 373: In this code, we would create a new `MutationObserver` (for every item in the slot, which wasn't necessary), but we would never actually disconnect them on future `slotchange` events.

dlockhart (approved):
Looks good!

PR #3658, opened 2023-05-19
feat: suppress duplicate ENTER presses

dlockhart:
As described in DE53398, there's nothing preventing a user from holding down the ENTER key in a search input and having hundreds/thousands of change events fire. In certain scenarios, those change events could cause downstream work to be done over and over, which we don't want.

This rate limits the handling of the ENTER key such that it only fires the event once per second OR if the value has changed since the last time it was fired.

### DIFF
--- a/components/inputs/input-search.js
+++ b/components/inputs/input-search.js
@@ -8,7 +8,8 @@ import { inputStyles } from './input-styles.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
-const INPUT_TIMEOUT_MS = 400;
+export const INPUT_TIMEOUT_MS = 400;
+export const SUPPRESS_ENTER_TIMEOUT_MS = 1000;
 
 /**
  * This component wraps the native "<input type="search">"" element and is for text searching.
@@ -90,6 +91,7 @@ class InputSearch extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) 
 		super();
 		this._inputTimeout = undefined;
 		this._lastSearchValue = '';
+		this._suppressEnter = false;
 		this.disabled = false;
 		this.noClear = false;
 		this.searchOnInput = false;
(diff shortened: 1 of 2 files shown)


components/inputs/test/input-search.test.js
```diff
@@ -98,27 +99,64 @@ describe('d2l-input-search', () => {
...
-		it('should fire "search" event when search input changes in search-on-input mode', async() => {
-			const elem = await fixture(searchOnInputFixture);
-			elem.value = 'foobar';
-			setTimeout(() => getTextInput(elem).dispatchEvent(new Event('input')));
-			const { detail } = await oneEvent(elem, 'd2l-input-search-searched');
-			expect(detail.value).to.equal('foobar');
-		});
+		describe('fake timer', () => {
 
-		it('should fire "search" event only once after two consecutive input events', async() => {
```
dlockhart, line 109: I moved these two tests inside the suite where fake timers are used. That way we can advance the clock programatically instead of waiting `50ms` in the test and slowing everything down.

dbatiste (approved):
LGTM!

components/inputs/input-search.js
```diff
@@ -8,7 +8,8 @@ import { inputStyles } from './input-styles.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
-const INPUT_TIMEOUT_MS = 400;
+export const INPUT_TIMEOUT_MS = 400;
+export const SUPPRESS_ENTER_TIMEOUT_MS = 1000;
```
oelhanafey, line 12: I wonder if this could be too long and feel sluggish in certain scenarios. Seems like we could make this shorter while still suppressing the majority of the events.

For example, maybe we're searching for something that is stored client-side, press ENTER and immediately notice a typo, fix it and press ENTER again. I think that 1 second could be noticeable in that situation.

oelhanafey (approved):

ca2hill (approved):
Looks great.  Thanks for tackling this.

PR #3659, opened 2023-05-19
Add label to d2l-list

oelhanafey:
[US152804](https://rally1.rallydev.com/#/?detail=/userstory/701306059915&fdp=true)

This PR adds the `label` property to `d2l-list` which adds an accessible label that can be used to provide additional context when it is not obvious.

Adding an `aria-label` to nested lists was causing the label to replace the child content and skip over all of the nested content. Since there doesn't seem to be a use case for labels on nested lists, we decided to only permit the property on top-level lists. Adding the property to a nested list should have no effect.

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -70,7 +70,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		this.dragMultiple = false;
 		this.extendSeparators = false;
 		this.grid = false;
-		this.label = '';
+		this.label = undefined;
 		this._listItemChanges = [];
 		this._childHasExpandCollapseToggle = false;
 
@@ -124,7 +124,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 
 	render() {
 		const role = !this.grid ? 'list' : 'application';
-		const ariaLabel = this.label && this.slot !== 'nested' ? this.label : undefined;
+		const ariaLabel = this.slot !== 'nested' ? this.label : undefined;
 		return html`
 			<slot name="controls"></slot>
 			<slot name="header"></slot>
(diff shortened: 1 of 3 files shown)


components/list/list.js
```diff
@@ -117,10 +124,11 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 
 	render() {
 		const role = !this.grid ? 'list' : 'application';
+		const ariaLabel = this.label && this.slot !== 'nested' ? this.label : undefined;
```
oelhanafey, line 127: There is a bit of an edge case here where the component is just checking if it is within a slot named "nested" but not that it is placed within a d2l-list. So it's possible that web component could have a "nested" slot that the component would interpret as a nested list.

But this is how the list and list-item have been checking for this throughout the components.

dlockhart: When no label is set and the constructor defaults it to the empty string, I _think_ this logic accomplishes what you want (an `undefined` value for `ariaLabel`) but worth double-checking (your unit tests indicate that it's working fine). `('' && true)` is falsy I think. But might be clearer if `label` defaulted to `undefined` in the constructor?

components/list/test/list.test.js
```diff
@@ -51,6 +51,37 @@ describe('d2l-list', () => {
...
+		it('should set aria-label on list when label is defined', async() => {
+			const elem = await fixture(html`<d2l-list label="Test Label"></d2l-list>`);
+			const list = elem.shadowRoot.querySelector('[role="list"]');
+			expect(list.getAttribute('aria-label')).to.equal('Test Label');
+		});
+
+		it('should not set aria-label on list when label is not defined', async() => {
+			const elem = await fixture(html`<d2l-list></d2l-list>`);
+			const list = elem.shadowRoot.querySelector('[role="list"]');
+			expect(list.getAttribute('aria-label')).to.be.null;
```
dlockhart, line 65: Nit:
```suggestion
			expect(list.hasAttribute('aria-label')).to.be.false;
```

dlockhart (approved):
Nice!

components/list/list.js
oelhanafey: Yeah, empty strings are falsy but defaulting to `undefined` would be more clear.

dlockhart (approved):

PR #3663, opened 2023-05-23
fix: use padding in place to icon to calculate maximum date input size

dlockhart:
**Defect:** DE53025

This is a bizarre bug that despite my best efforts I haven't been able to uncover the underlying reason behind it. I'm calling it *The Mysterious Case of the Missing Icon Width*.

### Background

The date input wants to size its width such that the largest possible date can always fit. Since this can vary based on date value and the locale, it renders some invisible "wide" versions of itself and uses those to set its width. Because the date input renders a calendar icon before the value, `<d2l-icon>` is included in the invisible versions as well.

In case the entire component is initially hidden (resulting in the invisible version having no width), it wires up a `ResizeObserver` to redo the calculation. It used to always use the `ResizeObserver`, but that code was simplified in #3419 which caused This Bug.

### This Bug

In Very Specific Situations (which of course happen in legacy LMS pages), the width of the icon in that initial calculation is ignored. That causes the input width to be `40px` (`0.6rem` margin + `0.8rem` icon + `0.6rem` margin) too small:

Screenshot 2023-05-23 at 2 16 51 PM

Why does it ignore the width of the icon? For the life of my, I do not know. The Very Specific Situations involve an inline `<fieldset>`, a label and a relative `<div>` parent. The following HTML reproduces it:

```html
<fieldset style="display: inline">
  <label>Date</label>
  <div style="position: relative;">
    <d2l-input-date-time label="Date" label-hidden></d2l-input-date-time>
  </div>
</fieldset>
```

Any of the following fixes it:
- Make the `<fieldset>` not inline
- Remove the label
- Make the `<div>` not relatively positioned (or remove it entirely)

Seems like a very primitive Webkit bug, since only Firefox behaves correctly.

### Solution

The solution shown here replaces the `<d2l-icon>` and its margins in the invisible versions with `2rem` of padding.

I tried a bunch of other things like an inline SVG or an inline-block `<span>` sized the same as the icon and none of them worked.


svanherk:
Since you determined the minimum HTML to reproduce anyways, should we have a visual-diff test of it to catch a possible regression if we ever change the padding size?

dlockhart:
> Since you determined the minimum HTML to reproduce anyways, should we have a visual-diff test of it to catch a possible regression if we ever change the padding size?

Yeah good call -- added.

svanherk (approved):

components/inputs/input-date.js
dbatiste, line 1: Was just doing a bit of exploring for this issue...

I notice I can reproduce the issue in both Chrome and Safari when the dev console is closed, but not when it's open.

I also notice that if I add a `setTimeout` to this method, it seems to resolve the issue. Maybe it is a timing issue - like Lit has updated it's stuff but the browser hasn't done layout yet so it's still too early to take a measurement. Though, I would think if measurements were stale, it would force layout at that point.

```javascript
const tryUpdateHiddenContentWidth = () => {
	setTimeout(() => {
		const width = Math.ceil(parseFloat(getComputedStyle(hiddenContent).getPropertyValue('width')));
		if (isNaN(width)) return false; // hidden elements will have "auto" width
		this._hiddenContentWidth = `${width}px`;
		return true;
	}, 0);
};
```

dlockhart: Yeah it's definitely a timing issue, I just wasn't excited about adding timeouts to this code.

I can reproduce regardless of the dev console, so not sure what's going on there.

PR #3666, opened 2023-05-23
fix: simplify collapsible panel summary item example

GZolla:
Simple change to reduce the amount of code displayed in the demo page for the `d2l-collapsible-panel-summary-items`.


dlockhart (approved):

dbatiste:
Note: this reduces the code behind the scenes, but this won't reduce the amount of code displayed in the Daylight demo. I know this still conveys the general idea of changes to the panel contents being reflected in the summary, but we may want to check with Jeff which way he prefers. If we do want these changes, we probably also want to update the margin in the styles to reduce the overall height of the demo container, given that the "add-ons" are removed and there would be excess whitespace.

GZolla:
Clos

PR #3672, opened 2023-05-25
Update dialogs and dropdowns to stop bubbling of touchstart.

dbatiste:
[DE51920](https://rally1.rallydev.com/#/?detail=/defect/684873597487&fdp=true)

This PR updates `DialogMixin` and `DropdownContentMixin` to stop bubbling of the `touchstart` event so that external elements such as `d2l-template-primary-secondary` (aside panel) do not react to it when they shouldn't. Without this change, users cause the mobile secondary/aside panel to resize when they touch-start within an open dialog or dropdown nested within the panel. 

I confirmed with Jeff that we still want the `d2l-template-primary-secondary` secondary panel behaviour that resizes the panel, just not when dialogs or dropdowns are open (removing that behaviour and just targeting the divider would have been an alternative solution). 

background-resize

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -142,6 +142,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 
 	_addHandlers() {
 		window.addEventListener('resize', this._updateSize);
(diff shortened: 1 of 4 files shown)


dlockhart (approved):
LGTM

PR #3675, opened 2023-05-25
Add no-padding property to d2l-dialog-fullscreen.

dbatiste:
[US153003](https://rally1.rallydev.com/#/?detail=/userstory/702058561511&fdp=true)

This PR adds the `no-padding` property to the `d2l-dialog-fullscreen` component to facilitate cases where consumers want to apply their own padding internally within the content components, for example if using `d2l-template-primary-secondary`.


templates/primary-secondary/demo/integration.html
```diff
@@ -342,19 +341,35 @@
...
-					<p>Trysail Sail ho Corsair red ensign hulk smartly boom jib rum gangway. Case shot Shiver me timbers gangplank crack Jennys tea cup ballast Blimey lee snow crow's nest rutters. Fluke jib scourge of the seven seas boatswain schooner gaff booty Jack Tar transom spirits.</p>
-					<p>Deadlights jack lad schooner scallywag dance the hempen jig carouser broadside cable strike colors. Bring a spring upon her cable holystone blow the man down spanker</p>
-					<p>Shiver me timbers to go on account lookout wherry doubloon chase. Belay yo-ho-ho keelhaul squiffy black spot yardarm spyglass sheet transom heave to.</p>
-					<p>Trysail Sail ho Corsair red ensign hulk smartly boom jib rum gangway. Case shot Shiver me timbers gangplank crack Jennys tea cup ballast Blimey lee snow crow's nest rutters. Fluke jib scourge of the seven seas boatswain schooner gaff booty Jack Tar transom spirits.</p>
+				<d2l-dialog-fullscreen id="dialogMain" title-text="Dialog Title" no-padding>
+					<div style="position: relative; height: 100%; width: 100%;">
```
dbatiste, line 345: It looks like there was an assumption made that `d2l-template-primary-secondary` would be applied as a page template since it is absolute positioned. If a consumer uses the component within a fullscreen dialog, it needs to be placed within a positioned element that fills the entire dialog content container, or whatever area the consumer wants it to fill.

In this demo case, I've achieved that by applying `position: relative` and `height: 100%` to the immediate child of the dialog's content container. This may be ok, as the consumer is choosing to use this element in a different context and so is responsible for providing an appropriate container to render within. 

However, an alternative solution would be to make the dialog's `.d2l-dialog-content` element `position: relative`. That might be ok too, and would eliminate the need for consumers to have to worry about this, but it would be introducing a new positioned container element to dialog contents. I would find it also a little strange that dialog would do this, when it's not the dialog that needs it. The outer dialog container itself already acts as a positioned parent, which gives me some confidence that we could introduce a lower level positioned element without causing any issues.

Interested in thoughts.

components/icons/catalogue.md
dbatiste, line 1: Can probably ignore this change. I think my previous PR included an extra icon that I had in the folder when I generated my local catalogue, so this is basically correcting that.

components/dialog/dialog-fullscreen.js
```diff
@@ -27,7 +27,11 @@ class DialogFullscreen extends LocalizeCoreElement(AsyncContainerMixin(DialogMix
 			 * @type {boolean}
 			 */
 			async: { type: Boolean },
-
+			/**
+			 * Render with no content padding
+			 * @type {boolean}
+			 */
+			noPadding: { type: Boolean, reflect: true, attribute: 'no-padding' },
```
dlockhart, line 34: Probably want to default to `false` in the constructor.

dlockhart (approved):

templates/primary-secondary/demo/integration.html
dbatiste: Ok, just to close the loop on this - we've decided to update the dialog content container to be `position: relative`. We'd like this scenario to "just work" without requiring the consumer to apply positioning and dimensions. Part of the rationale also being that the dialog is "resetting" the context and therefore reasonable to expect it to define the positioned container for containment.

dlockhart (commented):
Looks like the relative positioning messed up some v-diffs!

dlockhart (approved):

PR #3678, opened 2023-05-26
Change filter manual search-type to dispatch events when search is empty

oelhanafey:
[US148611](https://rally1.rallydev.com/#/?detail=/userstory/685391459987&fdp=true)

Currently the manual search type in the filter will fire events when searching but not when clearing the search and not when you first open the search. Events will be needed in those circumstances for Wave's filter work since in their case the initial list does not show all items in the filter, and likely also for the Load More implementation.

This PR changes the manual search-type to trigger an event when searching as well as when the dimension is first opened and when the search is cleared.

All current consumers of the manual search-type have been changed to be able to handle empty search values.


components/filter/demo/filter-search-demo.js
```diff
@@ -96,5 +106,24 @@ class FilterSearchDemo extends LitElement {
...
+			});
+		} else {
+			this._fullData.forEach(value => {
+				if (value.text.toLowerCase().indexOf(searchValue.toLowerCase()) > -1) {
+					keysToDisplay.push(value.key);
+				}
+			});
+		}
+		return keysToDisplay;
+	}
```
oelhanafey, line 126: This implements a search similar to what Wave needs. When search is empty, it will display the selected items as well as the items that are related to the header, in this case `admin` and `instructor`. When searching, it will do a regular search on the entire list

components/filter/filter.js
```diff
@@ -658,7 +658,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		this._dimensions.forEach(dimension => {
 			if (dimension.searchType !== 'none' && dimension.searchValue !== '') {
 				dimension.searchValue = '';
-				this._performDimensionSearch(dimension);
+				this._search(dimension);
```
oelhanafey, line 661: In the full-manual case, it will trigger a search event on clear

components/filter/filter.js
```diff
@@ -697,7 +697,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				}
 				this._activeFiltersSubscribers.updateSubscribers();
 			} else if (prop === 'values') {
-				if (dimension.searchValue) shouldSearch = true;
+				if (dimension.searchValue || dimension.searchType === 'full-manual') shouldSearch = true;
```
oelhanafey, line 700: Previously it was making the assumption that if search is empty, then all values will be shown but we don't want this in `full-manual`. This doesn't trigger a search event but will re-filter the items based on the results of the previous search.

svanherk:
I know this is still in draft but just wondering - could we turn the existing `manual` into a full manual search?  Would that just require consumers to start sending back their full list of ids when they get an event with a search value of "", or would there more to it? I guess we could even provide a shortcut for them to sand back something like "all"...

Not saying that's necessarily cleaner than this, just thinking out loud 🤔

oelhanafey:
@svanherk Yeah I considered it, I think that's all that would need to be done. I did it this way to be non-disruptive but if we want to go in and fix the consumers of the manual search-type then this is a possibility.

components/filter/filter.js
```diff
@@ -590,10 +590,13 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
-	_dispatchDimensionFirstOpenEvent(key) {
-		if (!this._openedDimensions.includes(key)) {
-			this.dispatchEvent(new CustomEvent('d2l-filter-dimension-first-open', { bubbles: true, composed: false, detail: { key: key } }));
-			this._openedDimensions.push(key);
+	_dispatchDimensionFirstOpenEvent(dimension) {
+		if (!this._openedDimensions.includes(dimension.key)) {
+			if (dimension.searchType === 'manual') {
+				this._search(dimension);
+			}
+			this.dispatchEvent(new CustomEvent('d2l-filter-dimension-first-open', { bubbles: true, composed: false, detail: { key: dimension.key } }));
```
svanherk, line 598: I wonder if the order matters here 🤔. I think the main point of this event is for consumers to wait to load the options until they get this event. Would they rather get the opened event and _then_ the search event, since searching before loading the options doesn't really make sense?

components/filter/demo/filter-search-demo.js
```diff
@@ -96,5 +104,24 @@ class FilterSearchDemo extends LitElement {
 
 	}
 
+	_performTwoListsSearch(searchValue) {
+		const keysToDisplay = [];
+		if (searchValue === '') {
+			keysToDisplay.push('admin', 'instructor');
```
svanherk, line 110:
```suggestion
	_performInitialSubsetSearch(searchValue) {
		const initialSubset= ['admin', 'instructor'];
		const keysToDisplay = [];
		if (searchValue === '') {
			keysToDisplay = initialSubset;
```

Minor: Probably overkill, but might make it a little clearer how this works. And maybe we go with some naming like "Event on Search - Initial Subset" or something? "Two Lists" makes me think of the header list versus the selected list, even though we _didn't_ go with two lists 😛

components/filter/test/filter.test.js
```diff
@@ -311,7 +311,7 @@ describe('d2l-filter', () => {
 
 		it('clear all - clears all dimensions and searches', async() => {
 			const elem = await fixture(`<d2l-filter>
-				<d2l-filter-dimension-set key="1" text="Dim 1" search-type="manual"><d2l-filter-dimension-set-value key="test" text="test" selected></d2l-filter-dimension-set-value></d2l-filter-dimension-set>
+				<d2l-filter-dimension-set key="1" text="Dim 1"><d2l-filter-dimension-set-value key="test" text="test" selected></d2l-filter-dimension-set-value></d2l-filter-dimension-set>
```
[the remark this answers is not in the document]
oelhanafey: I'm removing it to simplify the test, yeah. Because now in manual search it will dispatch a search event that would need to be handled. I can clean up that `searchKeysToDisplay` part which looks like it should've been `elem._dimensions[0].searchKeysToDisplay`.

I added a test below that is testing the event when search is set to the empty string, but I think I should add an additional one for `clearAll` and can keep this one testing the automatic search.

components/filter/filter.js
oelhanafey: Got it, yeah that is logical. I didn't know that was the purpose of the event

svanherk: I still think they'll need to handle it properly on their side if they're using both, since loading the options is usually `async`.  But I _think_ that order will be more intuitive to do so

svanherk (approved):

components/filter/demo/filter-search-demo.js
```diff
@@ -40,6 +40,11 @@ class FilterSearchDemo extends LitElement {
 						<d2l-filter-dimension-set-value key="${value.key}" text="${value.text}" ?selected="${value.selected}"></d2l-filter-dimension-set-value>
 					`)}
 				</d2l-filter-dimension-set>
+				<d2l-filter-dimension-set key="event-initial-subset" text="Event on Search - Initial Subset" search-type="manual" header-text="Related Roles at D2L" selected-first>
+					${this._fullDataSingle.map(value => html`
```
svanherk, line 44: Missed this the first time - does this need it's own data array? The demo seems fine, but I'm surprised setting things in this demo isn't affecting the `event-single` one, based on the code in `_handleFilterChange` 🤷

oelhanafey: Fixed the test but unsure if the demos are setup as we expect them.

The way the demos are setup it will correctly keep track of the data from those filters but changes will not be reflected on the filter. If we want that to be the case, then we would need to change the arrays to state properties and assign new arrays to those properties on filter changes, as opposed to altering the existing array.

That's why we couldn't actually see that bug affecting the demo.

svanherk: That's annoying, I guess I tried to setup the demo as an example of what a consumer needs to do, but only half set it up 😒. Does `this.requestUpdate()` work instead of having to set to new arrays?

components/filter/demo/filter-search-demo.js
```diff
@@ -60,7 +66,13 @@ class FilterSearchDemo extends LitElement {
 		e.detail.dimensions.forEach(dimension => {
 			if (!dimension.dimensionKey.includes('event')) return;
 
-			const dataToUpdate = dimension.dimensionKey === 'event-single' ? this._fullDataSingle : this._fullData;
+			const dataToUpdateMap = {
+				'event': this._fullData,
+				'event-single': this._fullDataSingle,
+				'event-initial-subset': this._fullDataInitialSubset,
+			};
+			if (dimension.dimensionKey !== 'event-initial-subset') return;
```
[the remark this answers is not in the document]
oelhanafey: Oops, nope that accidentally made it through when I was doing some experimenting

components/filter/demo/filter-search-demo.js
oelhanafey: Yep, so `requestUpdate()` does work. I also needed to deep copy the `initialData` since it was previously using the same objects even after copying the array.

svanherk: Right, kinda messy but I real consumer shouldn't run into that using actual data from the source.  Thanks for making this demo actually work! 😅

PR #3697, opened 2023-06-05
fix: d2l-list > prevent separators causing unexpected scrollbar

GZolla:
Before `d2l-list-item` would render the separator between itself and the NEXT item, overflowing so that the style changes on the next item (selection/hover/focus effects) would cover it. This changes it so it renders the separator between itself and the item coming BEFORE and applying the same logic to hide it. 

The benefit of this change is that before, the overflowing separator at the bottom of the list was not increasing the size of the container (since it is not part of the layout) but would cause a scrollbar to appear so that users could scroll the border into view. This is not the case for overflow at the start of the list.

I'm expecting a whole lot of v-diffs to fail since borders are moving up a couple of pixels to change the logic as mentioned above.

Rally: [DE52887](https://rally1.rallydev.com/#/?detail=/defect/697344876657&fdp=true)


components/list/test/screenshots/ci/golden/list-nested/d2l-list-nested-ltr-all-iterations-draggable.png
dbatiste, line 1: Are you expecting these diffs:
image

There are a few in there that are like this... this is just the first one.

components/list/test/screenshots/ci/golden/list/d2l-list-expand-collapse-default-expanded.png
dbatiste, line 1: Another missing separator here. 
image

components/list/list-item-mixin.js
```diff
@@ -107,6 +107,13 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			:host[hidden] {
 				display: none;
 			}
+			:host([_hovering-selection]),
+			:host([_hovering-primary-action]),
+			:host([selectable][_focusing]),
+			:host([_focusing-primary-action]),
+			:host([selected]:not([selection-disabled]):not([skeleton])) {
+				z-index: 1;
```
dbatiste, line 115: Welcome to the game of wack-a-mole! I think this can result in stacking issues when things like tooltips, dropdowns, and dialogs are rendered within list-items. 

I this case, I _think_ the `:host([selectable][_focusing])` selector trumps the `:host([_dropdown-open])` selector immediately below, resulting in both list-items having `z-index: 1`, and because DOM elements later in the DOM will stack on top of earlier ones (given the same `z-index` and stacking context), the selected adjacent list-item that follows will stack on top of the open dropdown (or tooltip, etc).

Aside: it would be really nice to have a visual-diff test that catches this. We can probably dream up a few such tests.

image

components/list/test/screenshots/ci/golden/list-nested/d2l-list-nested-ltr-all-iterations-draggable.png
GZolla: I talked to @geurts about the change to the bottom separator and it has been a defect for some time. I just committed changes to revert the other diffs.

components/list/list-item-mixin.js
dbatiste: It's unfortunate that we need to add `z-index` here. We're sort of desperate to remove the use of `z-index` from the list components due to all the issues it causes us. Is there any chance you tried to reserve some space at the bottom of the list, for example by applying a couple pixels of padding to the bottom of the list?

GZolla: I looked into this, it merits a discussion but my take is that this solution is also solving the defect on the separator between a nested list and the next parent item(see the other thread) and I can't think of an easy alternative to solve that defect. 
Initially, I thought that `d2l-list-item` could have separators outside of `d2l-list`, specially since exporting the mixin means there is custom list items out there, and hence why I did not try the padding on list. It does seem that separators are set only by `d2l-list` and hence could be the responsibility of the list to render which would remove the need to do any overlapping and hence solve the need for `z-index` or padding magic.

dbatiste: I think we can probably accept the `z-index` for now if @geurts really wants those separator side-effects. We already have the hacky work-arounds for the stacking contexts created by the generic layout (although they don't handle everything). We **really** want to walk back the use of `z-index` in list-items though. This might be ok... time will tell as stacking issues are often discovered later in use cases our demos and tests don't completely cover. Rubrics tends to be a case where issues surface.

If we don't want those separator side-effects, then I am in favour of investigating alternative solution such as padding the bottom of the list.

There is a good chance that if/when we update the rendering for list-items to remove all the stacking contexts, the separators will revert back the way they were, but we can cross that bridge when we come to it.

PR #3699, opened 2023-06-05
Add web-test-runner generateConfig()

bearfriend:
First pass at a shared config for all things `web-test-runner`

Note that this is targetting @dlockhart's [dlockhart/visual-diff-spike](https://github.com/BrightspaceUI/core/tree/dlockhart/visual-diff-spike) branch

[US152244](https://rally1.rallydev.com/#/?detail=/userstory/699786552875&fdp=true): Visual-diff: create a shared WTR config


tools/web-test-runner-config.js
```diff
@@ -0,0 +1,114 @@
...
+		browsers.push('firefox');
+	}
+	if (argv.indexOf('webkit') > -1) {
+		browsers.push('webkit');
+	}
+	return browsers.map((b) => playwrightLauncher({
+		product: b,
+		createBrowserContext: ({ browser }) => browser.newContext({ deviceScaleFactor: 2, reducedMotion: 'reduce' })
+	}));
+}
```
bearfriend, line 29: I think we need to nail down exactly how we want this to work. Should we always be using `playwrightLauncher` (for unit, axe, and vdiff)? Should we always include `chromium` if only `firefox` was passed in?

tools/web-test-runner-config.js
```diff
@@ -0,0 +1,114 @@
...
+					</html>`
+			};
+	}
+}
+
+const supportedGroups = ['aXe', 'visual-diff'];
+
+export function generateConfig({
+	pattern = type => `./**/*.${type}.js`,
+	groups = ['aXe', 'visual-diff'],
```
dlockhart, line 87: In my head I was thinking we do visual-diff with just a `visualDiff: true` flag in the config? And then setting that would automatically add the group.

Core is unique I think in how it splits out aXe (and it likely doesn't even need to anymore as it was done for performance reasons that likely don't exist now), so we could even leave that part out for now. Core could just take this generated config and add aXe itself.

tools/web-test-runner-config.js
dlockhart: I do think we always want reduced motion enabled... so yeah if we can always go through something like this that would be ideal.

tools/web-test-runner-config.js
dlockhart, line 1: This is going to live here, so if you'd like to move it there right away I'm good with that!
https://github.com/BrightspaceUI/testing

PR #3700, opened 2023-06-06
fix: `d2l-calendar` > Use more semantic elements and roles

jameswklassen:
## The issue 🐛 
The combination of Chromium and NVDA was causing `d2l-calendar` to not read dates when navigating with the calendar with arrow keys.

## The changes
* remove `role="presentation"` from the `table` element
  * this was forcing us to set our own functionality, and in some situations was preventing the screen reader from reading anything out.

## Related PR
- https://github.com/BrightspaceUI/core/pull/602

## Functionality details

<details>
<summary>Current screen reader behavior</summary>

* Windows
    * NVDA
        * Chromium 
            * tabbing in:  does not read anything 
            * moving:  does not read anything 
        * Firefox
            * tabbing in: Saturday 9. Selected. May 2020  row 2  column 7
            * moving: Friday 8. Not Selected. May 2020  not selected  column 6
    * JAWS
        * Chromium
            * tabbing in: Saturday 9. Selected. May 2020 selected contains controls
            * moving: Friday 8. Not Selected. May 2020 contains controls
        * Firefox
            * tabbing in: May 2020 table. column 6 row 1. Saturday 9. Selected. May 2020 selected contains controls
            * moving: Friday 8. Not Selected. May 2020 contains controls
* Mac
    * Voiceover
        * Chromium
            * tabbing in: Saturday 9. Selected. May 2020
            * moving: Friday 8. Not Selected. May 2020, button
        * Firefox
            * tabbing in: Entering May 2020 table. Saturday 9. Selected. May 2020, Ma 2020, table 7 columns, 6 rows
            * moving: Friday 8. Not Selected. May 2020, button, column 6 of 7
        * Safari
            * tabbing in: Entering Click on a day to select it. May 2020 table. Saturday 9. Selected. May 2020, Click on a day to select it., May 2020, table 7 columns, 7 rows
            * moving: Friday 8. Not Selected. May 2020, button, column 6 of 7
</details>

<details>
<summary>New screen reader behavior</summary>

* Windows
    * NVDA
        * Chromium
            * tabbing in: May 2020  table. Saturday 9. May 2020  row 2  column 7
            * moving around: Friday 8. May 2020  not selected  column 6
        * Firefox
            * tabbing in: May 2020  table Saturday 9 May 2020
            * moving around: Friday 8 May 2020  not selected  column 6
    * JAWS
        * Chromium
            * tabbing in: May 2020  table. column 7 row 2  Saturday 9 May 2020 selected contains controls
            * moving around: Friday 8 May 2020 contains controls
        * Firefox
            * tabbing in: May 2020  table. column 7 row 2  Saturday 9 May 2020 selected contains controls
            * moving around: Friday 8 May 2020 contains controls
* Mac
    * Voiceover
        * Chromium
            * tabbing in: Entering September 2018 table. Saturday 8 September 2018, September 2018, table 7 columns, 6 rows
            * moving around: Friday 7 September 2018, button, column 6 of 7
        * Firefox
            * tabbing in: Entering September 2018 table. Saturday 8 September 2018, September 2018, table 7 columns, 6 rows
            * moving around: Friday 7 September 2018, button, column 6 of 7
        * Safari
            * tabbing in: Entering September 2018 table. Saturday 8 September 2018, September 2018, table 7 columns, 7 rows
            * moving around:Friday 7 September 2018, button, column 6 of 7
</details>

## Rally
[DE47890](https://rally1.rallydev.com/#/15545167705d/defects?detail=%2Fdefect%2F630216629055&view=17f0e440-d291-4296-ac11-349b05e5aa0d)

### DIFF
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -468,13 +468,13 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				return html`
 					<td
 						aria-selected="${selected}"
+						aria-label="${description}"
 						data-date=${date}
 						data-month=${month}
 						data-year=${year}
 						@keydown="${this._onKeyDown}"
 						tabindex=${focused ? '0' : '-1'}>
 						<button
-							aria-label="${description}"
 							class="${classMap(classes)}"
 							@click="${this._onDateSelected}"
 							?disabled="${disabled}"
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -474,12 +474,13 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 						data-year=${year}
 						@keydown="${this._onKeyDown}"
 						tabindex=${focused ? '0' : '-1'}
-						type="button">
+					>
 						<button
 							class="${classMap(classes)}"
 							@click="${this._onDateSelected}"
 							?disabled="${disabled}"
 							tabindex="-1"
+							type="button"
 						>
 							${date}
 						</button>
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -473,15 +473,13 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 						data-year=${year}
 						@keydown="${this._onKeyDown}"
 						role="gridcell"
-						tabindex=${focused ? '0' : '-1'}
-					>
+						tabindex=${focused ? '0' : '-1'}>
 						<button
 							class="${classMap(classes)}"
 							@click="${this._onDateSelected}"
 							?disabled="${disabled}"
 							tabindex="-1"
-							type="button"
-						>
+							type="button">
 							${date}
 						</button>
 					</td>`;
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -463,7 +463,7 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				const month = day.getMonth();
 				const date = day.getDate();
 				const weekday = calendarData.descriptor.calendar.days.long[calendarData.daysOfWeekIndex[index]];
-				const description = `${weekday} ${date}. ${selected ? this.localize(`${this._namespace}.selected`) : this.localize(`${this._namespace}.notSelected`)} ${formatDate(day, { format: 'monthYear' })}`;
+				const description = `${weekday} ${date} ${formatDate(day, { format: 'monthYear' })}`;
 				return html`
 					<td
 						aria-label="${description}"
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -467,12 +467,11 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				// role="gridcell" used for screen reader (e.g., JAWS and VoiceOver) behavior to work properly
 				return html`
 					<td
-						aria-selected="${selected ? 'true' : 'false'}"
+						aria-selected="${selected}"
 						data-date=${date}
 						data-month=${month}
 						data-year=${year}
 						@keydown="${this._onKeyDown}"
-						role="gridcell"
 						tabindex=${focused ? '0' : '-1'}>
 						<button
 							aria-label="${description}"
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -466,7 +466,6 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				const description = `${weekday} ${date} ${formatDate(day, { format: 'monthYear' })}`;
 				return html`
 					<td
-						aria-label="${description}"
 						aria-selected="${selected ? 'true' : 'false'}"
 						data-date=${date}
 						data-month=${month}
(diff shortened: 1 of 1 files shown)


jameswklassen:
I'm not _entirely_ sure why axe tests are failing...

[`aria-selected`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) is valid for elements with `role="gridcell"` (which was removed in this PR), but I would expect a `td` to also have this role?... 

Unsure why this is now invalid 😕

components/calendar/calendar.js
```diff
@@ -467,20 +467,21 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 				// role="gridcell" used for screen reader (e.g., JAWS and VoiceOver) behavior to work properly
 				return html`
 					<td
+						aria-label="${description}"
 						aria-selected="${selected ? 'true' : 'false'}"
 						data-date=${date}
 						data-month=${month}
 						data-year=${year}
 						@keydown="${this._onKeyDown}"
-						role="gridcell"
```
dbatiste, line 475: I see that the `table` is being marked up with `role="presentation"`. When I inspect the "rows" and the "cells" in the Accessibility Object Model viewer in Chrome, I notice that the `tr`s don't have a role at all, and the `td`s have a generic role unless one is explicitly specified. Whereas when `role="presentation"` is removed (normal table semantics), the `tr` and `td` elements get their default `row` and `gridcell` roles applied automatically as one would expect. 

So, with your current changes, the `td` is getting a generic role assigned to it (at least in Chrome on MacOS), which I think makes sense since we're basically telling the browser to ignore table semantics. If on the other hand we actually want grid semantics, we either need to be using the table elements with their default roles, or we need to be assigning the appropriate `row` and `gridcell` roles. Also worth noting that Mozilla states that `gridcell` must be contained within a `row`, so what we were specifying previously was technically invalid.

Not sure if that clears anything up with respect to the roles. Currently we are using the table element simply for layout. We're not leveraging any of it's semantics, at least not properly, and the cell announcement comes entirely from the label. The header cells are purely for sighted users. I can't help but wonder if some of the crazy inconsistency could be avoided if we were just using Aria (if we actually want grid semantics) + CSS grid (for layout) instead of table elements.

dbatiste:
> I'm not _entirely_ sure why axe tests are failing...
> 
> [`aria-selected`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) is valid for elements with `role="gridcell"` (which was removed in this PR), but I would expect a `td` to also have this role?...
> 
> Unsure why this is now invalid 😕

From Mozilla:
>The aria-selected attribute indicates the current "selected" state for [gridcell](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/gridcell_role), [option](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role), [row](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/row_role) and [tab](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) roles.

And based on my previous comment where the `td` is getting a generic role assigned to it when within a `<table role="presentation">` and no explicit valid role is specified, I guess this error makes sense.

components/calendar/calendar.js
```diff
@@ -467,20 +467,21 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
...
-						tabindex=${focused ? '0' : '-1'}>
+						tabindex=${focused ? '0' : '-1'}
+					>
 						<button
-							aria-label="${description}"
 							class="${classMap(classes)}"
 							@click="${this._onDateSelected}"
 							?disabled="${disabled}"
```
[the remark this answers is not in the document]
jameswklassen: Not entirely sure what you mean by "factored into the keyboard interaction".

It seems like it worked before, and still works now?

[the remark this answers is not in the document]
dbatiste: I see, the component doesn't allow focus to move to those invalid dates to begin with.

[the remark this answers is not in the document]
jameswklassen: Yeah, I think this line is the important one?

https://github.com/BrightspaceUI/core/blob/79aa90eccd8534f3e9956c56120e6da86a1a82e1/components/calendar/calendar.js#L829

components/calendar/calendar.js
```diff
@@ -519,7 +518,7 @@ class Calendar extends LocalizeCoreElement(RtlMixin(LitElement)) {
 								icon="tier1:chevron-right">
 							</d2l-button-icon>
 						</div>
-						<table aria-labelledby="${labelId}" role="presentation">
+						<table aria-labelledby="${labelId}">
```
dbatiste, line 521: Is the role="gridcell" still needed on the `td` if we're removing the `role="presentation"`? On Mac+Chrome at least, that's the default role when native table semantics are in play.

jameswklassen: I don't think it _should_ be needed. But remove it, and Firefox doesn't know how to read the elements 🤦. (At least, that's what I think is happening)

dbatiste: Just a quick follow-up note for anyone reading this later. Chrome and Safari default `td` to `gridcell` role, whereas Firefox defaults it to `cell`. As per `aria-selected` specs, the role must be `gridcell`, `option`, `row`, or `tab`.

PR #3712, opened 2023-06-08
ignore: test out locales

dlockhart:
Just want to test out these locales to see if the fonts are now installed on the runners. We previously had to remove these because installing the font packs was too slow.


dlockhart:
Looks like without manually installing these languages (which is costly), the fonts still aren't there. Closing.

PR #3715, opened 2023-06-09
feat: add `d2l-collapsible-panel-group`

jameswklassen:
## Rally
[US153600](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fuserstory%2F703942502753)


components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -0,0 +1,36 @@
...
+				flex-direction: column;
+				row-gap: 0.5rem;
+			}
+		`];
+	}
+
+	render() {
+		const classes = {
+			padding: this.children?.[0].type !== 'inline'
+		};
```
jameswklassen, line 30: Not crazy about this. An alternative would be to have `d2l-collapsible-panel` handle spacing itself with a `margin-bottom`

components/collapsible-panel/collapsible-panel.js
```diff
@@ -315,6 +321,7 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
 			'has-summary': this._hasSummary,
 			'has-before': this._hasBefore,
 			'scrolled': this._scrolled,
+			'group-last': this.parentElement.nodeName === 'D2L-COLLAPSIBLE-PANEL-GROUP' && Array.from(this.parentElement.children).indexOf(this) === this.parentElement.children.length - 1,
```
jameswklassen, line 324: I _think_ we can use `children` in this way...from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/children#value)

> An [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection) which is a **live, ordered collection of the DOM elements** which are children of node.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -309,12 +315,14 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
...
 
 	render() {
+		console.log();
 		const classes = {
 			'd2l-collapsible-panel': true,
 			'focused': this._focused,
 			'has-summary': this._hasSummary,
 			'has-before': this._hasBefore,
 			'scrolled': this._scrolled,
+			'group-last': this.parentElement.nodeName === 'D2L-COLLAPSIBLE-PANEL-GROUP' && this === this.parentElement.lastElementChild,
```
dbatiste, line 325: One of the things that this doesn't work well for is - this panel won't be aware when panels are added or removed from the group. For that, it is better that the group listens for `slotchange` and then applies a class to the children. I think that's what our menu component does.

jameswklassen: Sounds good, I'll look at the implemenation for `menu`

components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -0,0 +1,26 @@
...
+				display: none;
+			}
+			:host ::slotted(d2l-collapsible-panel) {
+				display: unset;
+			}
+		`];
+	}
+
+	render() {
+		return html`<slot></slot>`;
```
dbatiste, line 22: The group component can perhaps handle the panel spacing automatically using flex:

```html
<div style="display: flex; flex-direction: column; gap: 0.5rem;">
    <slot></slot>
</div>
```

Some inspection of the contents (via `slotchange` or other means) would still be necessary to know if the children are inline.

I know you're also trying to do some stuff with the border though.

components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -0,0 +1,69 @@
...
+	static get styles() {
+		return [super.styles, css`
+			:host ::slotted(*) {
+				display: none;
+			}
+			:host ::slotted(d2l-collapsible-panel) {
+				display: unset;
+			}
+			.spaced {
+				display: flex;
```
dbatiste, line 26: I'm not sure if there are any side-effects of overriding the slot's `display: contents`. If not, this is great (and good to know!). If there is, you might need a wrapper div to apply the flex layout.

components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -0,0 +1,69 @@
...
+	async _handleSlotChange() {
+		this._panels = await this._getPanels();
+		this._updatePanelAttributes();
+	}
+
+	_updatePanelAttributes() {
+		if (!this._panels || this._panels.length === 0) return;
+
+		for (const panel of this._panels) {
+			panel._noBottomBorder = true;
```
dbatiste, line 62: We want to conditionally do this based on whether the panels are inline, yeah?

components/collapsible-panel/collapsible-panel-group.js
jameswklassen: I'm not sure if there is....it looked fine to me in the demo page...but I can wrap in a div for safety

components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -58,6 +58,8 @@ class CollapsiblePanelGroup extends SkeletonMixin(LitElement) {
 	_updatePanelAttributes() {
 		if (!this._panels || this._panels.length === 0) return;
 
+		if (this._panels?.[0]?.type !== 'inline') return;
```
jameswklassen, line 61: If we ever want to update panel attributes for the other cases, this will need to be reworked a bit. But I think the performance bump of short circuiting iteration is worth it.

jameswklassen: That's true, the optional chain is redundant.

I think I'd rather remove it and leave them both as explicit checks, since they're logically doing different things, and it's _somewhat_ likely that we could remove/rework the `inline` check.

components/collapsible-panel/collapsible-panel-group.js
jameswklassen: Did a bit of research on this -- it _seems_ safe to override the `display: contents`.

From the [CSS spec](https://drafts.csswg.org/css-display/#box-generation):

> The element itself does not generate any boxes, but its children and pseudo-elements still generate [boxes](https://drafts.csswg.org/css-display/#box) and [text sequences](https://drafts.csswg.org/css-display/#css-text-sequence) as normal. For the purposes of box generation and layout, the element must be treated as if it had been replaced in the [element tree](https://drafts.csswg.org/css-display/#element-tree) by its contents


I'm _guessing_ we're safe to change this to `flex`? Since the `slot` is now acting as an element that controls layout (vs. an "invisible" element).

components/collapsible-panel/collapsible-panel-group.js
```diff
@@ -0,0 +1,71 @@
+import { css, html, LitElement } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
+
+/**
+ * A component that renders collapsible panels
```
dbatiste, line 6: I think "A component that renders a container and layout for collapsible panels" would be a more accurate description.

dbatiste:
Looking good.

I think we'll want a couple visual-diffs to capture the rendering behaviour it provides.

components/collapsible-panel/collapsible-panel-group.js
jameswklassen: Oh yeah, words...I forgot about updating this. Yeah I like your suggestion.

dbatiste:
We should add the new component to [this test](https://github.com/BrightspaceUI/core/blob/main/components/collapsible-panel/test/collapsible-panel.test.js#L8).

dbatiste:
We should add something to the readme as well. While not necessary, I think we want consumers using the group to lay out their panels.

components/collapsible-panel/README.md
```diff
@@ -75,6 +75,7 @@ The collapsible panel is a container that can show or hide additional content. I
 * Use to progressively disclose information or settings to simplify a complex workflow
 * Keep the header consistent across the collapsed and expanded states of the panel
 * Use the subtle collapsible panel for backgrounds that aren't white
+* Use [`d2l-collapsible-panel-group`](#d2l-collapsible-panel-group) to group multiple panels for better visual consistency
```
jameswklassen, line 78: @geurts has signed off on this point.

Don't think I did the link correctly....?

should it be `daylight.d2l.dev/components/collapsible-panel#d2l-collapsible-panel-group`?

components/collapsible-panel/README.md
```diff
@@ -358,6 +359,36 @@ Collapsible panels have two optional slots, `header` and `actions` that can be u
...
+	</d2l-collapsible-panel>
+	<d2l-collapsible-panel panel-title="Submission 2">
+		<p>Sweet roll candy dessert caramels shortbread gummies toffee oat cake cookie. Wafer gummies shortbread sweet halvah jujubes sweet. Cake chocolate chocolate bar carrot cake marzipan. Icing chupa chups jujubes macaroon toffee chocolate bar wafer croissant.</p>
+	</d2l-collapsible-panel>
+	<d2l-collapsible-panel panel-title="Submission 3">
+		<p>Sweet roll candy dessert caramels shortbread gummies toffee oat cake cookie. Wafer gummies shortbread sweet halvah jujubes sweet. Cake chocolate chocolate bar carrot cake marzipan. Icing chupa chups jujubes macaroon toffee chocolate bar wafer croissant.</p>
+	</d2l-collapsible-panel>
+</d2l-collapsible-panel-group>
+```
+
```
jameswklassen, line 391: This doesn't have official approval yet -- we can either omit it, or leave it in and iterate. 

I'm fine with either.

components/collapsible-panel/README.md
dbatiste: I'm not 100% sure, but I think it would need to be:
```
https://daylight.d2l.dev/components/collapsible-panel/#collapsible-panel-d2l-collapsible-panel
```

jameswklassen: Do you mean: 

```
https://daylight.d2l.dev/components/collapsible-panel/#collapsible-panel-group-d2l-collapsible-panel-group
```

components/collapsible-panel/README.md
dbatiste: I think it's ok. If Jeff wants to tweak it, we can certainly do that later.

dbatiste (approved):
I think that link in the readme probably needs to get updated one way or another, but otherwise I think this looks good.

PR #3716, opened 2023-06-09
Fix a race condition in d2l-icon that can cause the wrong icon to be shown

mpharoah-d2l:
Simplifies `d2l-icon` by using safe built-in Lit Element directives instead of the custom build `runAsync`.

This fixes a race condition in which, if an icon is set to two different new values quickly, sometimes the older of the two new values will permanently (or until the icon changes again) be rendered instead of the newer one.

### DIFF
--- a/components/icons/icon.js
+++ b/components/icons/icon.js
@@ -1,11 +1,12 @@
 import '../colors/colors.js';
-import { css, html, LitElement } from 'lit';
+import { css, html, LitElement, noChange } from 'lit';
 import { fixSvg } from './fix-svg.js';
+import { guard } from 'lit/directives/guard.js';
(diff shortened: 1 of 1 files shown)


devpow112 (approved):

dbatiste (approved):
Thanks Matt. Looks good!

mpharoah-d2l:
Forgot to prefix the commit message with `fix:`. Can a version bump be manually triggered?

devpow112:
> Forgot to prefix the commit message with `fix:`. Can a version bump be manually triggered?

Usually I've just seen people make an empty commit and merge another PR with the right prefix to trigger release.

PR #3719, opened 2023-06-09
fix: reserve space for last separator

GZolla:
This change adds some padding to the end of the list in order to prevent the last separator on the list to force a scrollbar to appear.

Rally: [DE52887](https://rally1.rallydev.com/#/?detail=/defect/697344876657&fdp=true)

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -56,6 +56,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			:host {
 				display: block;
 			}
+			:host > div {
+				padding-bottom: 2px;
+			}
 			:host([hidden]) {
 				display: none;
 			}


components/list/list.js
```diff
@@ -56,6 +56,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			:host {
 				display: block;
 			}
+			:host > div {
```
dbatiste, line 59: Targeting the `div` works right now, but the list component delegates rendering of the paging container to the pageable mixin. If someone modifies that rendering method to include a wrapper `div` for some reason without knowing that consumers have selectors targeting them, they would get `2px` of whitespace when they shouldn't. It would be safer to add a class to the `div` and target that (or the `role` attribute) to avoid the risk of that happening.

dbatiste (approved):

PR #3711, opened 2023-06-08
web-test-runner shared config

bearfriend:
Basic setup changes to use `@brightspace-ui/testing` and its incoming `createConfig` function.

To test locally `npm link` the [testing repo](https://github.com/BrightspaceUI/testing).

PR #3721, opened 2023-06-09
Create wtr config with @brightspace-ui/testing

bearfriend:
Recreated from https://github.com/BrightspaceUI/core/pull/3711 to target `main`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Accessibility Tests
         run: npm run test:axe
       - name: Unit Tests
-        run: npx web-test-runner --config web-test-runner.config.js --group default --playwright --browsers chromium firefox webkit
+        run: npm run test:headless
   publish:
     name: Publish static site
     if: ${{ github.triggering_actor != 'dependabot[bot]' }}
--- a/package.json
+++ b/package.json
@@ -20,9 +20,10 @@
     "start": "web-dev-server --node-resolve --watch --open",
     "test": "npm run lint && npm run test:translations && npm run test:headless && npm run test:axe",
     "test:axe": "web-test-runner --group aXe",
-    "test:headless": "web-test-runner --group default",
-    "test:headless:watch": "web-test-runner --group default --watch",
-    "test:translations": "mfv -e -s en -p ./lang/ -i untranslated"
+    "test:headless": "web-test-runner --group unit",
+    "test:headless:watch": "web-test-runner --group unit --watch",
+    "test:translations": "mfv -e -s en -p ./lang/ -i untranslated",
+    "test:vdiff": "web-test-runner --group vdiff"
   },
   "files": [
     "custom-elements.json",
@@ -45,6 +46,7 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7",
     "@brightspace-ui/stylelint-config": "^0.8",
+    "@brightspace-ui/testing": "^0.1",
     "@open-wc/semantic-dom-diff": "^0.19.9",
     "@open-wc/testing": "^3",
     "@rollup/plugin-dynamic-import-vars": "^2",
--- a/package.json
+++ b/package.json
@@ -52,8 +52,6 @@
     "@rollup/plugin-node-resolve": "^15",
     "@rollup/plugin-replace": "^5",
     "@web/dev-server": "^0.2",
-    "@web/test-runner": "^0.16",
-    "@web/test-runner-playwright": "^0.10",
     "axe-core": "^4",
     "chalk": "^5",
     "eslint": "^8",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -20,9 +20,10 @@
     "start": "web-dev-server --node-resolve --watch --open",
     "test": "npm run lint && npm run test:translations && npm run test:headless && npm run test:axe",
     "test:axe": "web-test-runner --group aXe",
-    "test:headless": "web-test-runner --group default",
-    "test:headless:watch": "web-test-runner --group default --watch",
-    "test:translations": "mfv -e -s en -p ./lang/ -i untranslated"
+    "test:headless": "web-test-runner --group unit",
+    "test:headless:watch": "web-test-runner --group unit --watch",
+    "test:translations": "mfv -e -s en -p ./lang/ -i untranslated",
+    "test:vdiff": "web-test-runner --group vdiff"
```
dlockhart, line 26: We don't need this line yet though right? Since it won't do anything and isn't called by anything.

.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Accessibility Tests
         run: npm run test:axe
       - name: Unit Tests
-        run: npx web-test-runner --config web-test-runner.config.js --group default --playwright --browsers chromium firefox webkit
+        run: npm run test:headless
```
dlockhart, line 43: Isn't this going to change it from running in all browsers to running only in Chrome?

web-test-runner.config.js
```diff
@@ -1,38 +1,13 @@
...
-		`<html lang="en">
-			<body>
-				<script src="./tools/resize-observer-test-error-handler.js"></script>
-				<script type="module" src="${testFramework}"></script>
-			</body>
-		</html>`
-};
+export default createConfig({
+	pattern,
+	vdiff: true,
```
dlockhart, line 7: If we're planning to merge this right away, I think we want this off.

package.json
bearfriend: Yeah, we can do testing with `npx` for now.

.github/workflows/ci.yml
bearfriend: Nope. The `vdiff` group explicitly passes `['chromium']` but the `unit` group doesn't pass anything so it'll default to all `ALLOWED_BROWSERS`

web-test-runner.config.js
bearfriend: All it does is add the group, which you'd have to explicitly target, and if we remove the `test:vdiff` script, which I think we should, it'll allow us to run it with `npx`. But we can turn it off here and just flip it on when we want to test it instead.

dlockhart: Ok. I was figuring we'd still be maintaining some kind of WIP branch like my PoC one where we could toggle it on and do whatever else we wanted to, but we can sort that out later once we know more about how we're going to use/test this in core while still developing the vdiff library.

.github/workflows/ci.yml
dlockhart: Gotcha. So I guess a decision for us around whether we actually want that by default locally vs. how it's behaving now (Chrome only).

svanherk: Hmm, so you're thinking maybe we want this to be:
```
web-test-runner --group unit
```
And in `package.json`, we'd have `"test:headless": "web-test-runner --group unit --browsers chromium",`?

bearfriend: I feel like defaulting to more will make sure people don't accidentally not test everything, but we can flesh this out more later.

dlockhart: That would keep the behaviour the same as it is now, but yeah running everything always seems fine too. Let's go with that and we can always add the `--browsers chromium` flag if we end up hating it!

web-test-runner.config.js
bearfriend: Yeah we can do either or both. I'll turn these things off.

web-test-runner.config.js
```diff
@@ -1,38 +1,13 @@
-import { playwrightLauncher } from '@web/test-runner-playwright';
+import { createConfig, getBrowsers } from '@brightspace-ui/testing';
```
dlockhart, line 1: May want to hold off on merging until my changes go in... at which point this will be:

```suggestion
import { createConfig, getBrowsers } from '@brightspace-ui/testing/wtr-config.js';
```

package.json
```diff
@@ -45,6 +46,7 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7",
     "@brightspace-ui/stylelint-config": "^0.8",
+    "@brightspace-ui/testing": "^0.1",
```
svanherk, line 49: We should be able to delete these below:
```
    "@web/test-runner": "^0.16",
    "@web/test-runner-playwright": "^0.10",
```

svanherk (approved):
22

svanherk:
> 22

What the...? 😕 👻 😨 😱

dlockhart (approved):
🎉

PR #3737, opened 2023-06-13
include wildcard dependency name

dlockhart:
Thanks @eKoopmans for pointing out that this still requires a dependency name.


dbatiste:
FWIW, I think validity can be checked here:
image

dlockhart:
Yeah it also failed [on my merge commit](https://github.com/BrightspaceUI/core/runs/14194674745) but I didn't notice.

eKoopmans (approved):

PR #3746, opened 2023-06-15
fix: `d2l-expand-collapse-content` > set expanded state in another spot

jameswklassen:
## Description

If the component is expanding but is empty, the height of the contents will be `0`, and therefore there is nothing to animate, and the `transitionend` event will not be dispatched.

This led to an issue where the component was stuck in an "expanding" state with `height: 0`.

If we detect this scenario, we simply move to the end "expanded" state with `height: auto`. Any subsequent addition of elements to the panel would then be displayed immediately.

## Rally
[DE53776](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?Iteration=%2Fiteration%2F668139821167&Release=u&cpoid=15545167705&detail=%2Fdefect%2F704665648499&rankScope=%2Fiteration%2F668139821167&typeDef=7362609627)

### DIFF
--- a/components/expand-collapse/expand-collapse-content.js
+++ b/components/expand-collapse/expand-collapse-content.js
@@ -119,9 +119,7 @@ class ExpandCollapseContent extends LitElement {
 				));
 			}
 			if (reduceMotion || firstUpdate) {
-				this._state = states.EXPANDED;
-				this._height = 'auto';
-				this._eventPromiseResolve();
+				this._setExpanded();
 			} else {
 				this._state = states.PREEXPANDING;
 				await this.updateComplete;
@@ -130,6 +128,7 @@ class ExpandCollapseContent extends LitElement {
 					this._state = states.EXPANDING;
 					const content = this.shadowRoot && this.shadowRoot.querySelector('.d2l-expand-collapse-content-inner');
 					if (content) this._height = `${content.scrollHeight}px`;
+					if (content?.scrollHeight === 0) this._setExpanded();
 				}
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


dbatiste:
For anyone else who comes across this change - the jist of the issue is that if the panel is expanding but it's empty, the height of the contents will be 0 and therefore there is nothing to animate and the `transitionend` event will not be dispatched, leaving the component in an intermediate "expanding" state with `height: 0`. If we detect this scenario, we simply move to the end "expanded" state with `height: auto`. Any subsequent addition of elements to the panel would then be displayed immediately.

jameswklassen:
> For anyone else who comes across this change - the jist of the issue is that if the panel is expanding but it's empty, the height of the contents will be 0 and therefore there is nothing to animate and the `transitionend` event will not be dispatched, leaving the component in an intermediate "expanding" state with `height: 0`. If we detect this scenario, we simply move to the end "expanded" state with `height: auto`. Any subsequent addition of elements to the panel would then be displayed immediately.

Thanks @dbatiste. Updated PR description with your words for more clarity.

components/expand-collapse/expand-collapse-content.js
```diff
@@ -130,6 +128,7 @@ class ExpandCollapseContent extends LitElement {
 					this._state = states.EXPANDING;
 					const content = this.shadowRoot && this.shadowRoot.querySelector('.d2l-expand-collapse-content-inner');
 					if (content) this._height = `${content.scrollHeight}px`;
```
dbatiste, line 130: This change is "ok" because the `this._height` will not be applied until Lit's async lifecycle runs. If this synchronously updated the DOM, the line that immediately follows would force the browser to recompute layout in order to obtain a fresh `scrollHeight`, even if the property that invalidated the layout wasn't height related. Just something to be cautious of... so in general when obtaining DOM measurements that are used more than once, I usually obtain them once and assign them to a constant or variable.

dbatiste (approved):

PR #3750, opened 2023-06-15
switch to use focusWithKeyboard/mouse from testing library

dlockhart:
`@brightspace-ui/testing` now exports `focusWithKeyboard` and `focusWithMouse`. This switches our tests that were using either `triggerFocusFor`/`triggerBlurFor` (which don't exist in our testing library) or core's own copy of those methods to use them directly.


components/tooltip/test/tooltip.test.js
```diff
@@ -112,7 +113,7 @@ describe('d2l-tooltip', () => {
 	describe('show-hide', () => {
 
 		it('should show when target is focused', async() => {
-			await triggerFocusFor(tooltipFixture.querySelector('#explicit-target'));
+			await focusWithKeyboard(tooltipFixture.querySelector('#explicit-target'));
```
dbatiste, line 116: Do you want the `setTimeout` here as well?

dlockhart: Tooltip does a lot of internal `await`-ing to update its position and stuff before firing its event, so technically no we don't need anything here.

dbatiste (approved):

PR #3751, opened 2023-06-15
adding keys/labels to reduce logs in test results

dlockhart:
In some scenarios, if tests are slow enough the missing labels actually end up throwing while tests are still executing and because the exception doesn't get thrown by `LabelledMixin` for 3 seconds it can actually fail a totally unrelated test, which is very confusing! Don't ask me how I know.

### DIFF
--- a/components/list/test/list-item-expand-collapse-mixin.test.js
+++ b/components/list/test/list-item-expand-collapse-mixin.test.js
@@ -18,7 +18,7 @@ const tag = defineCE(
 
 describe('ListItemExpandCollapseMixin', () => {
 	it('Sets expandable to false when no key is given', async() => {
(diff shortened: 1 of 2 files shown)


dbatiste (approved):

components/list/test/list-item-expand-collapse-mixin.test.js
```diff
@@ -18,7 +18,7 @@ const tag = defineCE(
 
 describe('ListItemExpandCollapseMixin', () => {
 	it('Sets expandable to false when no key is given', async() => {
```
dbatiste, line 20: Oh, looks like this test is actually testing behaviour for the absence of a key. Lol.

PR #3754, opened 2023-06-16
switch to keyboard helpers

dlockhart:
I'm noticing some super rare flake in these tests in Firefox. I'm hoping that switching them to use the actual keyboard to trigger things instead of faking it out like it was doing will improve things.

### DIFF
--- a/components/list/test/list-item-drag-handle.test.js
+++ b/components/list/test/list-item-drag-handle.test.js
@@ -1,19 +1,7 @@
 import { expect, fixture, html, oneEvent } from '@open-wc/testing';
 import { dragActions } from '../list-item-drag-handle.js';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
-
-const keyCodes = Object.freeze({
-	END: { key: 'end', code: 35 },
-	HOME: { key: 'home', code: 36 },
-	UP: { key: 'up arrow', code: 38 },
-	DOWN: { key: 'down arrow', code: 40 },
-	SPACE: { key: 'space', code: 32 },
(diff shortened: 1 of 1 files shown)


components/list/test/list-item-drag-handle.test.js
```diff
@@ -1,19 +1,7 @@
...
-	UP: { key: 'up arrow', code: 38 },
-	DOWN: { key: 'down arrow', code: 40 },
-	SPACE: { key: 'space', code: 32 },
-	ENTER: { key: 'enter', code: 13 },
-	ESC: { key: 'escape', code: 27 },
-	TAB: { key: 'tab', code: 9 },
-	LEFT: { key: 'left arrow', code: 37 },
-	RIGHT: { key: 'right arrow', code: 39 }
-});
+import { sendKeys } from '@web/test-runner-commands';
```
dlockhart, line 4: We probably want to export this from `@brightspace-ui/testing` at some point so folks don't need to deal with `test-runner-commands`.

bearfriend: Would we just want to `export '@brightspace-ui/testing/commands';`?

bearfriend (approved):

components/list/test/list-item-drag-handle.test.js
dlockhart: I think so. I'm not sure if it needs a separate `/commands` export or not. 🤔 

But yeah, we could selectively re-export only the useful stuff and it also gives us an opportunity to adjust their APIs if they're not clear.

PR #3755, opened 2023-06-16
stub console to suppress expected warnings

dlockhart:
Cleaning up some of the console warnings/errors that get spit out while our unit tests are running by stubbing them out. 

There's still a single log entry that's expected for an error being thrown in a Promise -- haven't figured out a way to catch that yet.


components/list/test/list-item-checkbox-mixin.test.js
```diff
@@ -14,7 +15,14 @@ const tag = defineCE(
 );
 
 describe('ListItemCheckboxMixin', () => {
-	// will log "ListItemCheckboxMixin requires a key."
```
dlockhart, line 17: Not anymore!

svanherk (approved):

PR #3756, opened 2023-06-16
fix flakey tests with key utilities

dlockhart:
More relatively rare flake in Firefox with these list tests. 🤞 that switching them to use the real keyboard makes them more reliable.

### DIFF
--- a/components/list/test/list.test.js
+++ b/components/list/test/list.test.js
@@ -4,12 +4,9 @@ import '../list-item.js';
 import '../list-item-button.js';
 import '../list-item-content.js';
 import { expect, fixture, html, oneEvent, waitUntil } from '@open-wc/testing';
+import { focusWithKeyboard } from '@brightspace-ui/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
-
-const keyCodes = Object.freeze({
-	UP: { name: 'up arrow', key: 38 },
-	DOWN: { name: 'down arrow', key: 40 }
-});
+import { sendKeys } from '@web/test-runner-commands';
 
 const awaitListElementUpdates = async(rootElement, queries) => {
 	await rootElement.updateComplete;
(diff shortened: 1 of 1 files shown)


components/list/test/list.test.js
```diff
@@ -85,10 +82,10 @@ describe('d2l-list', () => {
 	describe('navigation', () => {
 
 		[
-			{ keyPress: keyCodes.UP },
-			{ keyPress: keyCodes.DOWN }
+			{ keyPress: 'ArrowUp' },
+			{ keyPress: 'ArrowDown' }
```
jameswklassen, line 86: How come we can change this from a complex object to a string?

components/list/test/list.test.js
```diff
@@ -117,22 +114,22 @@ describe('d2l-list', () => {
 
 				const listItem = elem.querySelector('[key="L2-2"]');
 				const listItemLayout = elem.querySelector('[key="L2-2"]').shadowRoot.querySelector('d2l-list-item-generic-layout');
-				listItem.focus();
+				await focusWithKeyboard(listItem);
 				await waitUntil(() => listItem.hasAttribute('_focusing'), 'List item should be focused', { timeout: 3000 });
 
-				const event = new KeyboardEvent('keydown', { keyCode: testCase.keyPress.key });
-				setTimeout(() => listItemLayout.dispatchEvent(event));
+				await focusWithKeyboard(listItemLayout);
```
jameswklassen, line 120: Why do we need to focus `listItem` and `listItemLayout` separately now?

dlockhart:
The menu test that flaked is on my hit-list for a separate fix, so re-running.

components/list/test/list.test.js
dlockhart: Because previously we were simulating a key press using `KeyboardEvent` which requires a key "code" but we wanted a friendly name to put in the test name, so we needed two bits of information for each test. Now we're using the built-in Playwright stuff that uses the [standard string-based](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) keys, so we only need the friendlier value.

components/list/test/list.test.js
dlockhart: It has to do with how `sendKeys` works compared to our simulation previously. `sendKeys` itself doesn't take an element, so it'll send those key presses to whatever element currently has focus. Previously we were focusing on the list item to put it into "focus" mode, and they simulating a key press on the layout element.

svanherk (approved):

PR #3759, opened 2023-06-20
switch to @brightspace-ui/testing mouse and keyboard commands

dlockhart:
Now that all the keyboard & mouse commands are surfaced from `@brightspace-ui/testing`, this switches core over to use them instead of `@web/test-runner-commands`. We also renamed `focusWithKeyboard` to `focusElem` and removed `focusWithMouse`.

I ended up uncovering a some flake unrelated to these changes in Firefox & Webkit, so there are a couple fixes here. More to come later though as these seem to mostly just happen locally and I'm hoping the switch to the new `fixture` will resolve some of them.

### DIFF
--- a/components/filter/test/filter.test.js
+++ b/components/filter/test/filter.test.js
@@ -495,10 +495,11 @@ describe('d2l-filter', () => {
 				expect(dimensions[0].dimensionKey).to.equal('dim');
 				expect(dimensions[0].cleared).to.be.false;
 				expect(dimensions[0].changes.length).to.equal(2);
-				expect(dimensions[0].changes[0].valueKey).to.equal('2');
-				expect(dimensions[0].changes[0].selected).to.be.true;
-				expect(dimensions[0].changes[1].valueKey).to.equal('1');
-				expect(dimensions[0].changes[1].selected).to.be.false;
+
+				const change1 = dimensions[0].changes.find(e => e.valueKey === '1');
+				const change2 = dimensions[0].changes.find(e => e.valueKey === '2');
+				expect(change2.selected).to.be.true;
+				expect(change1.selected).to.be.false;
 				expect(elem._dimensions[0].values[0].selected).to.be.false;
 				expect(elem._dimensions[0].values[1].selected).to.be.true;
 
--- a/components/list/test/list.test.js
+++ b/components/list/test/list.test.js
@@ -4,9 +4,8 @@ import '../list-item.js';
 import '../list-item-button.js';
 import '../list-item-content.js';
 import { expect, fixture, html, oneEvent, waitUntil } from '@open-wc/testing';
-import { focusWithKeyboard } from '@brightspace-ui/testing';
+import { focusElem, sendKeysElem } from '@brightspace-ui/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
-import { sendKeys } from '@web/test-runner-commands';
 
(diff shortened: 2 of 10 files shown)


components/filter/test/filter.test.js
```diff
@@ -495,10 +495,11 @@ describe('d2l-filter', () => {
 				expect(dimensions[0].dimensionKey).to.equal('dim');
 				expect(dimensions[0].cleared).to.be.false;
 				expect(dimensions[0].changes.length).to.equal(2);
-				expect(dimensions[0].changes[0].valueKey).to.equal('2');
-				expect(dimensions[0].changes[0].selected).to.be.true;
-				expect(dimensions[0].changes[1].valueKey).to.equal('1');
-				expect(dimensions[0].changes[1].selected).to.be.false;
```
dlockhart, line 501: @svanherk this was the Safari flake I was chatting with you about. Because it's the underlying list that un-selects the one item when a new one is selected, I don't know that there's any guarantee the changes will be in a particular order. So I updated this to not care about the order.

components/list/test/list.test.js
```diff
@@ -157,15 +155,14 @@ describe('d2l-list', () => {
...
+				await waitUntil(() => listItem.hasAttribute('_focusing'), 'Initial item should be focused', { timeout: 3000 });
 
-				await focusWithKeyboard(listItemLayout);
-				setTimeout(() => sendKeys({ down: testCase.keyPress }));
+				setTimeout(() => sendKeysElem('down', testCase.keyPress, listItemLayout));
 				await oneEvent(listItemLayout, 'keydown');
 
 				const focusedElement = elem.querySelector(`[key=${testCase.expectedFocus}`);
-				expect(focusedElement.hasAttribute('_focusing')).to.be.true;
+				await waitUntil(() => focusedElement.hasAttribute('_focusing'), 'Next item should be focused', { timeout: 3000 });
```
dlockhart, line 165: This was flaking a bit in Firefox, so made it a bit more resilient.

components/menu/test/menu.test.js
```diff
@@ -176,51 +161,47 @@ describe('d2l-menu', () => {
...
 			await oneEvent(elem, 'd2l-hierarchical-view-show-complete');
-			let focused = (document.activeElement.tagName === 'D2L-MENU-ITEM-RETURN');
-			if (!focused) {
-				focused = (document.activeElement === nestedMenu);
-			}
-			expect(focused).to.be.true;
+			await waitUntil(() => {
+				return (document.activeElement.tagName === 'D2L-MENU-ITEM-RETURN') ||
+					(document.activeElement === nestedMenu);
+			}, 'Focus on return');
```
dlockhart, line 175: Also flaky, and this is one that I'm still seeing occasional flake locally so will be spending more time on it.

components/switch/test/switch.test.js
```diff
@@ -41,9 +41,7 @@ describe('d2l-switch', () => {
 	['Space', 'Enter'].forEach((key) => {
 		it(`should toggle when ${key} is pressed`, async() => {
 			const elem = await fixture(switchFixture);
-			setTimeout(() => elem.focus());
-			await oneEvent(elem, 'focus');
-			setTimeout(() => sendKeys({ press: key }));
+			setTimeout(() => sendKeysElem('press', key, elem));
```
dlockhart, line 44: `sendKeysElem` focuses first, so this test can be simplified.

PR #3760, opened 2023-06-20
feat: add load more pager to filters

GZolla:
Adding load more pager to filters. Added the `pager-has-more` property to `d2l-filter-dimension-set` that when set will render a `d2l-pager-load-more` button (rendering done in `d2l-filter`). 
The handling of pagination and adding/removing the property from the dimension is still the responsibility of the consumer (see the added demo as an example). 


components/filter/filter.js
```diff
@@ -710,6 +715,17 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		this._activeDimensionKey = null;
 	}
 
+	_handleDimensionLoadMore(e) {
+		const dimensionKey = e.target.parentNode.id.slice(SET_DIMENSION_ID_PREFIX.length);
+
+		/** @ignore */
+		this.dispatchEvent(new CustomEvent('d2l-filter-dimension-load-more', {
```
dbatiste, line 722: Would a consumer ever try to do this?
```html
<d2l-filter-dimension-set @d2l-filter-dimension-load-more=${...}>...</d2l-filter-dimension-set>
```
Are we sure the `<filter>` is the correct not to dispatch the event?

GZolla: Consumers would need to check for the key in the event. I replicated the behavior for the `d2l-filter-dimension-empty-state-action`, since rendering happens on filter it is also responsible for dispatching events.

svanherk: Yeah having the `filter` dispatch all the events is a side-effect of our choice to have filter render everything, which we recently met about and decided not to change, even with the limitations. I think we could _technically_ dispatch the event off the corresponding data-only dimension component, but like @GZolla mentioned the other dimension-specific events currently don't work this way.  I'm also not sure if the analyzer for the Daylight site would understand where to put the event in that case?

components/filter/filter.js
```diff
@@ -541,6 +542,10 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				${selectedListItems}
 				${listHeader}
 				${listItems}
+				<d2l-pager-load-more slot="pager"
```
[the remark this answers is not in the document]
svanherk: Sounds like we don't want this! Marking as resolved

components/filter/filter.js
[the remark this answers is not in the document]
svanherk: Sounds like the answer to this is "no", because it's tied to `list` which already extends `PageableMixin`

components/filter/demo/filter-load-more-demo.js
```diff
@@ -0,0 +1,74 @@
...
+		e.detail.complete();
+	}
+
+	_renderDimensionSet(dimension) {
+		const items = dimension.items;
+		const loadedItems = dimension.lastLoadedIndex >=  items - 1 ? items : items.slice(0, dimension.lastLoadedIndex);
+
+		return html`<d2l-filter-dimension-set key="${dimension.key}" text="${dimension.text}" ?pager-has-more="${loadedItems.length < items.length}">
+			${repeat(loadedItems, item => item.key, item => html`
+				<d2l-filter-dimension-set-value key="${item.key}" text="${item.text}"></d2l-filter-dimension-set-value>
```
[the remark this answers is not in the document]
GZolla: The last commit modifies the demo to handle this

components/filter/demo/filter-load-more-demo.js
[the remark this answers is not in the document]
GZolla: The demo now uses the same logic we use for search: we add all items but hide them and then show only those that are "loaded". This means the count will be correct even when some values are hidden by the search or the load more.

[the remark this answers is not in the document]
GZolla: I'm not sure what the best approach is here, but looking at other components we could just leave it as is. If manual search is not set the load more will work but the search would just filter the added items which sort of breaks its use. To me this is similar to failing to add keys to list items, we don't generate a key but just let the component break, meaning that the consumer is solely responsible for properly setting up the component.

components/filter/filter.js
```diff
@@ -710,6 +732,24 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
+	_handleDimensionLoadMore(e) {
+		const dimensionKey = e.target.parentNode.id.slice(SET_DIMENSION_ID_PREFIX.length);
+		const dimension = this._getDimensionByKey(dimensionKey);
+		const searchCompleteCallback = this._getSearchCallback(dimension);
+
+		this.dispatchEvent(new CustomEvent('d2l-filter-dimension-load-more', {
+			detail: {
+				dimensionKey,
+				complete: (keysToDisplay) => {
+					searchCompleteCallback({ keysToDisplay });
```
svanherk, line 744: Still trying to wrap my head around how this works if the consumer can't give us all the values up front and will need to go fetch them when they get this event, which I _expect_ is what everyone using `load-more` will need for performance reasons.

I think this will happen, which won't work:
1. We fire this event, and the consumer loads the new values on their side and adds new `d2l-filter-dimension-set-value`s for them.
2. Then they call `complete` and tell us what keys to display (they will probably need to be given the search value with the load event to know this?)
3. But step 1 triggers `_handleSlotChange`, which I think will clear everything related to searching

Maybe we book a meeting for when @geurts is back, because I think we need answers to these to figure out how to set this up: 
- What happens when someone clicks load more when a search is applied - does it give _only_ the new values that match that search? Does it load 20 more values but only show ones that match, meaning you could get no new items added?
- When you remove that search, does it keep the new values in the list, but then there's "missing" values in between?

It would be easiest to just disable load more if you've searched something, but I think that's the exact use case nova needs.

components/filter/filter.js
```diff
@@ -710,6 +732,24 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		this._activeDimensionKey = null;
 	}
 
+	_handleDimensionLoadMore(e) {
+		const dimensionKey = e.target.parentNode.id.slice(SET_DIMENSION_ID_PREFIX.length);
+		const dimension = this._getDimensionByKey(dimensionKey);
+		const searchCompleteCallback = this._getSearchCallback(dimension);
```
svanherk, line 738:
```suggestion
		const applySearch = this._getSearchCallback(dimension);
```
This might make it a bit clearer how it's being used here? Basically after load more is complete, we trigger a search to update visibility.  It's kinda an odd workflow - it works because we _know_ they're using manual search.  But I do think it's cleaner than sending a load-more event and search event separately.

components/filter/filter-dimension-set.js
```diff
@@ -45,7 +50,7 @@ class FilterDimensionSet extends LitElement {
 			 * Whether to render the selected items at the top of the filter
 			 * @type {boolean}
 			 */
-			selectedFirst: { type: Boolean, attribute: 'selected-first' },
+			selectedFirst: { type: Boolean, attribute: 'selected-first', reflect : true },
```
svanherk, line 53: Is the `reflect` needed to work, or is it just to tell the consumer we changed the value on them?

components/filter/filter-dimension-set.js
```diff
@@ -132,6 +138,16 @@ class FilterDimensionSet extends LitElement {
...
 
+	willUpdate(changedProperties) {
+		if (changedProperties.has('hasMore') && this.hasMore) {
+			if (this.searchType !== 'manual') {
+				console.warn('Paging requires search type set to manual.');
+				this.hasMore = false;
+			}
+			else this.selectedFirst = true;
+		}
+	}
```
svanherk, line 149: Interested in some other opinions here.  Currently, if you've set `has-more` but not `search-type="manual"`, we turn off `has-more` and don't render the pager.  We can't just turn on manual search for them, because they won't have events setup to handle it. Then, if they _have_ correctly configured search, we override their setting of `selected-first` because we also require that to be on (and we can force that on for them, there's nothing they need to do).

This fails pretty gracefully, but just wondering if we should actually be more destructive to make it obvious things are misconfiguring?  Should we log a console error in one or both scenarios, and/or be less graceful and more obviously broken?

I'm good with this, but wanted to check since we've used a few different strategies across `core` before

components/filter/filter.js
```diff
@@ -607,7 +613,26 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
+	_getSearchCallback(dimension) {
+		return function({ keysToDisplay = [], displayAllKeys = false } = {}) {
+			requestAnimationFrame(() => {
+				dimension.displayAllKeys = displayAllKeys;
+				dimension.searchKeysToDisplay = keysToDisplay;
+				this._performDimensionSearch(dimension);
+				dimension.loading = false;
+				if (dimension.selectedFirst) {
+					this._updateDimensionShouldBubble(dimension);
+				}
```
svanherk, line 632: Are we now running this twice for the search scenario (once [here](https://github.com/BrightspaceUI/core/blob/518f5c4264aa3825ff02e06f35c438579f572a7c/components/filter/filter.js#L800-L802), and then again above)?

If so, could we move it into `_performDimensionSearch`? Seems like that hits all paths - automatic search, manual search, and search after load more?

components/filter/filter.js
```diff
@@ -710,6 +735,25 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
 
+	_handleDimensionLoadMore(e) {
+		const dimensionKey = e.target.parentNode.id.slice(SET_DIMENSION_ID_PREFIX.length);
+		const dimension = this._getDimensionByKey(dimensionKey);
+		const applySearch = this._getSearchCallback(dimension);
+
+		this.dispatchEvent(new CustomEvent('d2l-filter-dimension-load-more', {
+			detail: {
+				key: dimensionKey,
+				value: dimension.searchValue,
```
svanherk, line 746: Maybe `searchValue` to make it clear, or `currentlyAppliedSearch`?

components/list/test/screenshots/ci/golden/list/d2l-list-nested-all-selected.png
svanherk, line 1: Do you know why this visual diff is here? I wonder if merging `main` in would remove this, doesn't seem related to your changes

components/filter/filter.js
GZolla: My thought process was to have the same attribute names between this and the manual search event. Since we also decided to document those together, I think it makes it more intuitive to the consumer.

components/filter/filter-dimension-set.js
```diff
@@ -31,6 +31,11 @@ class FilterDimensionSet extends LitElement {
 			 * @type {boolean}
 			 */
 			loading: { type: Boolean },
+			/**
+			 * Wether the dimension has more values to load
```
svanherk, line 35:
```suggestion
			 * Whether the dimension has more values to load. Must be used with selected-first and search-type="manual"
```
Could also be done as part of the documentation story if you're trying to get this merged before branching!

svanherk (approved):
Just realized we should probably add a visual-diff test that shows the load-more pager, and has a couple values selected-first but `selected-first` isn't applied. But I'm fine with that being a follow-up PR!

svanherk:
Also, just realized the focus is being lost after hitting "load More".  @dbatiste, is this something the list is supposed to handle automatically, or the consumer has to?

dbatiste:
> Also, just realized the focus is being lost after hitting "Load More". @dbatiste, is this something the list is supposed to handle automatically, or the consumer has to?

I don't think focus should be lost. There were some previous changes that would disable the button while loading, and that _did_ cause focus to be lost, but that was removed a long time ago. Focus should stay on the button while loading, and when the consumer calls `e.detail.complete()` we focus the first newly loaded item ([based on index](https://github.com/BrightspaceUI/core/blob/main/components/paging/pager-load-more.js#L127)). We do ignore subsequent clicks while loading.

svanherk:
> > Also, just realized the focus is being lost after hitting "Load More". @dbatiste, is this something the list is supposed to handle automatically, or the consumer has to?
> 
> I don't think focus should be lost. There were some previous changes that would disable the button while loading, and that _did_ cause focus to be lost, but that was removed a long time ago. Focus should stay on the button while loading, and when the consumer calls `e.detail.complete()` we focus the first newly loaded item ([based on index](https://github.com/BrightspaceUI/core/blob/main/components/paging/pager-load-more.js#L127)). We do ignore subsequent clicks while loading.

@GZolla Is there a defect or something to track this? Just don't want to forget about it

PR #3762, opened 2023-06-20
feat: `d2l-collapsible-panel` > support h5 and h6 semantic levels

### LINKED ISSUE #3761: collapsible-panel: Support heading level 5 and 6
Within CE we are working on making our headings more easily navigable for screen readers. Due to the d2l-rubrics component using the collapsible-panel component and our need to make its heading an `h5`, it is necessary to support this.

image

jameswklassen:
## Description
There are use cases that require semantic `h5` and `h6` elements for the panel title. There are no heading styles beyond h4, but the semantic elements can still be supported.

If a `h5` or `h6` is selected, it will default to use the `h4` styles. Consumers can still override styles, like before.

Closes #3761


components/collapsible-panel/collapsible-panel.js
```diff
@@ -452,21 +459,23 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
...
-		let headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ? this.headingLevel : this.headingStyle;
-		headingStyle = normalizeHeadingNumber(headingStyle);
+		const headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ? this.headingLevel : this.headingStyle;
+		this.headingStyle = normalizeHeadingStyle(headingStyle);
 
 		const titleClasses = {
 			'd2l-collapsible-panel-title': true,
 			'd2l-skeletize': true,
-			[`d2l-heading-${headingStyle}`]: true,
+			[`d2l-heading-${this.headingStyle}`]: true,
```
jameswklassen, line 468: I noticed that `heading-style` was appearing in the DOM as the default "3", regardless of what the component did, since we never update `this.headingStyle`.

Figured it made sense to set the property here as well...shouldn't change any functionality...

components/collapsible-panel/collapsible-panel.js
```diff
@@ -452,21 +459,23 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
 	}
 
 	_renderPanelTitle() {
-		let headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ? this.headingLevel : this.headingStyle;
-		headingStyle = normalizeHeadingNumber(headingStyle);
+		const headingStyle = (this.headingStyle === defaultHeading && this.headingLevel !== this.headingStyle) ? this.headingLevel : this.headingStyle;
+		this.headingStyle = normalizeHeadingStyle(headingStyle);
```
dlockhart, line 463: I don't love mutating the property here, since if the normalization did something it's going to cause a re-render (and we're already in `render()`).

jameswklassen: Yeah that's a good point...I forgot why we did it this way.

That said, I'm not crazy about having an incorrect `heading-style` attribute on the DOM element...maybe the normalization would make more sense in `updated`? Then we just use `this.headingStyle` and `this.headingLevel` directly in this render method.

dlockhart: I'm not too worried about it, but if you want to ensure that it's always valid I'd suggest creating a getter/setter and doing the validation in the setter. [Here's an example from color input](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-color.js#L248).

dlockhart (approved):
Looks good!

components/collapsible-panel/collapsible-panel.js
dbatiste: Completely agree.

Also, for cases where we need to compute an internal reactive property in advance of `render()`, we have [willUpdate(...)](https://lit.dev/docs/components/lifecycle/#willupdate). It's rarely used in core, partly because a lot of the components were written before it existed. There's probably a few cases where we could avoid double rendering by using it. Anyway, just something to keep in mind.

dbatiste (approved):

PR #3764, opened 2023-06-20
switch to @brightspace-ui/testing fixture

dlockhart:
The vast majority of the changes here are just adjusting the import location, however having `fixture` now wait for everything to be ready did require adjusting a few tests that were dependent on it _not_ doing that.

Once this is merged I'll do a separate pull request to remove all the test code that's manually waiting for nested element `updateComplete`s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         run: npm run lint:style
   playwright:
     name: Accessibility and Unit Tests
-    timeout-minutes: 10
+    timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
       - uses: Brightspace/third-party-actions@actions/checkout
--- a/components/dialog/test/dialog.test.js
+++ b/components/dialog/test/dialog.test.js
@@ -1,5 +1,5 @@
 import '../dialog.js';
-import { expect, fixture, oneEvent } from '@open-wc/testing';
+import { expect, fixture, oneEvent } from '@brightspace-ui/testing';
 import { getComposedActiveElement } from '../../../helpers/focus.js';
 import { html } from 'lit';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
@@ -17,7 +17,8 @@ describe('d2l-dialog', () => {
 	describe('focus management', () => {
 
 		it('should focus on close button if no focusable elements inside', async() => {
-			const el = await fixture(html`<d2l-dialog opened>not focusable</d2l-dialog>`);
+			const el = await fixture(html`<d2l-dialog>not focusable</d2l-dialog>`);
+			el.opened = true;
 			await oneEvent(el, 'd2l-dialog-open');
 			expect(getComposedActiveElement().getAttribute('aria-label')).to.equal('Close this dialog');
 		});
@@ -52,16 +53,18 @@ describe('d2l-dialog', () => {
 		});
 
 		it('should not focus on an autofocus element that had not been made focusable', async() => {
-			const el = await fixture(html`<d2l-dialog opened><p autofocus>focus</p><button>focus</button></d2l-dialog>`);
-			const paragraph = el.querySelector('p');
+			const el = await fixture(html`<d2l-dialog><p autofocus>focus</p><button>focus</button></d2l-dialog>`);
+			el.opened = true;
 			await oneEvent(el, 'd2l-dialog-open');
+			const paragraph = el.querySelector('p');
 			expect(getComposedActiveElement()).to.not.equal(paragraph);
 		});
 
 		it('should not focus on a descendant autofocus element that had not been made focusable', async() => {
-			const el = await fixture(html`<d2l-dialog opened><div><p autofocus>focus</p></div><button>focus</button></d2l-dialog>`);
-			const paragraph = el.querySelector('p');
+			const el = await fixture(html`<d2l-dialog><div><p autofocus>focus</p></div><button>focus</button></d2l-dialog>`);
+			el.opened = true;
 			await oneEvent(el, 'd2l-dialog-open');
+			const paragraph = el.querySelector('p');
 			expect(getComposedActiveElement()).to.not.equal(paragraph);
 		});
 
--- a/components/dropdown/test/dropdown-content.test.js
+++ b/components/dropdown/test/dropdown-content.test.js
@@ -1,7 +1,6 @@
 import '../dropdown.js';
 import '../dropdown-content.js';
-import { aTimeout, expect, fixture, html, nextFrame, oneEvent } from '@open-wc/testing';
-import { focusElem } from '@brightspace-ui/testing';
+import { aTimeout, expect, fixture, focusElem, html, nextFrame, oneEvent } from '@brightspace-ui/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
 const normalFixture = html`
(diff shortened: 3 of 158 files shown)


components/dialog/test/dialog.test.js
```diff
@@ -17,7 +17,8 @@ describe('d2l-dialog', () => {
 	describe('focus management', () => {
 
 		it('should focus on close button if no focusable elements inside', async() => {
-			const el = await fixture(html`<d2l-dialog opened>not focusable</d2l-dialog>`);
+			const el = await fixture(html`<d2l-dialog>not focusable</d2l-dialog>`);
+			el.opened = true;
```
dlockhart, line 21: As discussed previously, these types of tests where an event is fired as part of boot-up need to be adjusted to control when the event gets fired themselves.

components/dropdown/test/dropdown-content.test.js
```diff
@@ -62,7 +61,6 @@ describe('d2l-dropdown', () => {
 	beforeEach(async() => {
 		dropdown = await fixture(normalFixture);
 		content = dropdown.querySelector('d2l-dropdown-content');
-		await content.updateComplete;
```
dlockhart, line 65: Will do more of this kind of thing in a follow-up PR.

components/hierarchical-view/test/hierarchical-view.test.js
```diff
@@ -24,7 +24,6 @@ describe('d2l-hierarchical-view', () => {
 		const el = await fixture(viewsFixture);
 		view1 = el.querySelector('#view1');
 		view2 = el.querySelector('#view2');
-		await oneEvent(view1, 'd2l-hierarchical-view-resize');
```
dlockhart, line 27: This event has already been fired by the time `fixture` resolves.

components/inputs/test/input-search.test.js
```diff
@@ -103,7 +103,7 @@ describe('d2l-input-search', () => {
 
 			let clock;
 			beforeEach(() => {
-				clock = useFakeTimers();
+				clock = useFakeTimers({ toFake: ['clearTimeout', 'setTimeout'] });
```
dlockhart, line 106: The new `fixture` relies on `await nextFrame()` when it resets stuff, so we need to be more targeted with our mocking otherwise `requestAnimationFrame` gets mocked.

components/list/test/list-item-drag-handle.test.js
dlockhart, line 1: These tests were VERY flaky in Firefox. I spent the entire day trying to figure out why occasionally the handle loses focus before the keys can be sent to it but got nowhere. So I ended up implementing a "retry".

components/list/test/list-item-drag-handle.test.js
```diff
@@ -37,15 +36,12 @@ describe('ListItemDragHandle', () => {
...
 		let element;
 		beforeEach(async() => {
 			element = await fixture(html`<d2l-list-item-drag-handle></d2l-list-item-drag-handle>`);
-			element.focus();
-			const actionArea = element.shadowRoot.querySelector('button');
-			setTimeout(() => {
-				actionArea.dispatchEvent(new Event('click'));
-			});
-			await oneEvent(actionArea, 'click');
+			element.activateKeyboardMode();
```
dlockhart, line 43: Activating keyboard mode does everything that clicking the button does.

components/menu/test/menu.test.js
```diff
@@ -256,22 +254,19 @@ describe('d2l-menu', () => {
...
 		);
 
-		const elem = await fixture(`
+		const elemPromise = fixture(`
 			<d2l-menu>
 				<d2l-menu-item text="fast"></d2l-menu-item>
 				<${delayedUpdateMenuItem}></${delayedUpdateMenuItem}>
 				<span>not a menu item</span>
 			</d2l-menu>
 		`);
```
dlockhart, line 266: Had to rework how this "slow element" works since `fixture` now won't resolve until everything is ready anyway.

components/more-less/more-less.js
```diff
@@ -147,12 +146,6 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
 	}
 
 	render() {
-		requestAnimationFrame(
-			() => this.dispatchEvent(new CustomEvent('d2l-more-less-render', {
-				bubbles: false,
-				composed: false
-			}))
-		);
```
dlockhart, line 155: No longer needed now that `fixture` waits for this for us.

mixins/labelled/test/label-mixin.test.js
```diff
@@ -92,7 +91,7 @@ describe('LabelledMixin', () => {
 			const labelledElem = elem.querySelector('[labelled-by="label1"]');
 			const labelElem = elem.querySelector('#label1');
 			labelElem.textContent = 'new label value';
-			await nextFrame();
```
dlockhart, line 95: This was really flaky so I switched it to use an event instead of rely on rAF.

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         run: npm run lint:style
   playwright:
     name: Accessibility and Unit Tests
-    timeout-minutes: 10
+    timeout-minutes: 15
```
dlockhart, line 24: I've re-run this many times locally and in CI.
- Locally, these changes are 1.1x (aXe) to 1.5x (unit) slower, which is barely noticeable
- In CI however, these changes are 1.3x (aXe) to 1.8x (unit) slower, which is adding ~3.5 minutes to this job. It's still quicker than the visual-diff tests though! 😅 

It's admittedly not great, but also not shocking since we're now `await`-ing all nested Lit components for ~1650 tests, and doing a lot more resetting of the browser state between tests. If the slowness becomes a problem, there's a lot of tests using timeouts that we could look into fixing up.

components/list/test/list-item-drag-handle.test.js
```diff
@@ -74,7 +77,14 @@ describe('ListItemDragHandle', () => {
 			it(`Dispatch drag handle action event for "${testCase.result}" when "${testCase.keyPress}" is pressed.`, async() => {
 				const actionArea = element.shadowRoot.querySelector('d2l-button-move');
 				setTimeout(() => dispatchKeyEvent(actionArea, testCase.keyPress, !!testCase.shift));
-				const e = await oneEvent(element, 'd2l-list-item-drag-handle-action');
+				let e = await oneEvent(element, 'd2l-list-item-drag-handle-action');
+				// Firefox will trigger loss of focus randomly, so retry
+				if (e.detail.action !== testCase.result && e.detail.action === dragActions.save) {
+					element.activateKeyboardMode();
+					await element.updateComplete;
+					setTimeout(() => dispatchKeyEvent(actionArea, testCase.keyPress));
```
svanherk, line 85:
```suggestion
					setTimeout(() => dispatchKeyEvent(actionArea, testCase.keyPress, !!testCase.shift));
```
?

components/list/test/list-item-drag-handle.test.js
svanherk: We'll have to keep an eye on these, to see if one retry tends to be enough. Longer tests make flake that much more painful

svanherk (approved):

components/inputs/test/input-search.test.js
dlockhart: Yep I've already included it in the [draft upgrade guide](https://desire2learn.atlassian.net/wiki/spaces/GAUD/pages/3667394686/Upgrade+Guide+draft), but I could definitely see it going into a README as well.

PR #3769, opened 2023-06-21
remove retries

dlockhart:
Removing these retries now that we've uncovered the source of the problem -- that Firefox + Playwright in a multithreaded environment causes focus to be randomly lost. The testing library [has been updated](https://github.com/BrightspaceUI/testing/pull/35) such that Firefox always runs single-threaded.


dbatiste (approved):

dlockhart:
Interestingly this didn't affect the time it took in CI to run the tests *at all*, which likely means they're always single-threaded... which doesn't bode well for our visual-diff tests being faster.

svanherk:
> Interestingly this didn't affect the time it took in CI to run the tests at all, which likely means they're always single-threaded... which doesn't bode well for our visual-diff tests being faster.

That's definitely worrisome 😟

PR #3770, opened 2023-06-21
fix: update constructor helper

dlockhart:
@ChrisLabattD2L noticed that since switching from `@open-wc/testing` to `@brightspace-ui/testing`, repos using the [constructor test helper](https://github.com/BrightspaceUI/core/blob/main/tools/constructor-test-helper.js#L1) now fail to import `@brightspace-ui/testing` because it's only a `devDependency`. This would have also been a problem previously, but all repos using it must have also depended on `@open-wc/testing`.

Longer term, I think this helper should just move over to `@brightspace-ui/testing`, but that will require switching all consumers over to it... of which there are >450.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@brightspace-ui/intl": "^3",
+        "@brightspace-ui/testing": "^0.10",
         "@formatjs/intl-pluralrules": "^1",
         "@open-wc/dedupe-mixin": "^1",
         "ifrau": "^0.40",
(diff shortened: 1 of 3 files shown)


dlockhart:
Ugh, license checker.

svanherk:
Hmm, this is gonna a mess in BSI I think - maybe we just keep this one place using `@open-wc`'s export for now, until everyone's switched to using it from `testing`?

dlockhart:
Yeah agreed -- already pushed that change, which seems to be the least bad of all the bad options.

svanherk (approved):

ChrisLabattD2L (approved):

PR #3772, opened 2023-06-21
Fix tag-list-item clearable functionality

bearfriend:
[DE53055](https://rally1.rallydev.com/#/?detail=/defect/698395690329&fdp=true): Fixer > User Tag List Items > Fix clearable functionality

Many consumers of `d2l-tag-list-item` and `TagListItemMixin` use custom (usually data) attributes on tags to identify them. This PR adds a `key` property to the `tag-list-item-mixin` to act as a unique identifier for each tag, and includes it in the `d2l-tag-list-item-clear` event (though I'm not sure how valuable it really is there since `e.target.key` is already available).

Sometimes there are additional attributes added to denote relationships to other objects/entities. This seems like a reasonable and appropriate practice to me that does not require facilitation through the component/mixin.

Additionally, and probably more importantly, I've added a `displayText` option to `_renderTag` which allows consumers to give the tag a simple label for use within various langterms and screenreader announcements. This is useful when the contents of the tag are [not a simple string](https://github.com/BrightspaceHypermediaComponents/users/blob/8b256864f75e52f1e491aa24aad8eabe149a11e5/components/user-tag-list-item.js#L97-L105).

The only place that uses `_renderTag` with non-string contents already doesn't announce or localize properly, so this should be updated to pass `displayText`, but it can be done at any time:
[BrightspaceHypermediaComponents/users](https://github.com/BrightspaceHypermediaComponents/users/blob/8b256864f75e52f1e491aa24aad8eabe149a11e5/components/user-tag-list-item.js#L113)

Changing the event `detail` is technically a breaking change, but nothing is actually using the existing `detail.value`, so these uses of `d2l-tag-list-item-clear` _can_ be updated to use `key` but are not required:
[Brightspace/insights-adoption-dashboard](https://github.com/Brightspace/insights-adoption-dashboard/blob/1fd2584624207e3b099fde9e91a7cbe3f264d8fc/components/tags-list.js#L173)
[Brightspace/d2l-outcomes](https://github.com/Brightspace/d2l-outcomes/blob/8fc3523234a7432e6b11a30aac453249903df88e/src/components/tags/activity-alignment-tag-list.js#L306)
[Brightspace/cds-reflection-tool](https://github.com/Brightspace/cds-reflection-tool/blob/0e99db37084fc5aad2fa24da17c3db97634121d1/src/reflection-tool-admin/components/admin-email-draft.js#L541)

`🤷` Weird things I kinda wish I didn't see:
[Brightspace/nova](https://github.com/Brightspace/nova/blob/c85904d0af7f9336279c84ab65f3f9d1b65158bd/src/app/shared/components/report/report-applied-filter-options/report-applied-filter-options.js#L37)
[Brightspace/nova](https://github.com/Brightspace/nova/blob/c85904d0af7f9336279c84ab65f3f9d1b65158bd/src/app/admin/views/tenants/manage-career-explorer.js#L274)
[Brightspace/cds-custom-pacing-tool-ui](https://github.com/Brightspace/cds-custom-pacing-tool-ui/blob/92d630ede90c18cf79ad1133fef140730283028e/src/components/dates-manager/dates-manager.js#L137)


components/tag-list/tag-list-item-mixin.js
```diff
@@ -27,6 +27,11 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			 * @type {boolean}
 			 */
 			clearable: { type: Boolean },
+			/**
+			 * Acts as a unique identifier for the tag in dispatched events
```
bearfriend, line 31:
```suggestion
			 * Acts as a unique identifier for the tag
```

components/tag-list/test/tag-list.test.js
bearfriend, line 1: I'll swap to `@brightspace-ui/testing` when there is consensus on the approach.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -169,12 +175,12 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
...
 		if (!this.clearable) return;
 
-		announce(this.localize('components.tag-list.cleared-item', { value: this.text }));
+		announce(this.localize('components.tag-list.cleared-item', { value: this._displayText }));
 
 		/** Dispatched when a user selects to delete an individual tag list item. The consumer must handle the actual element deletion and focus behaviour if there are no remaining list items. */
 		this.dispatchEvent(new CustomEvent(
 			'd2l-tag-list-item-clear',
-			{ bubbles: true, composed: true, detail: { value: this.text } }
+			{ bubbles: true, composed: true, detail: { key: this.key } }
```
bearfriend, line 183: I don't think there's much value in this.
```suggestion
			{ bubbles: true, composed: true }
```

components/tag-list/tag-list-item-mixin.js
```diff
@@ -27,6 +27,11 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			 * @type {boolean}
 			 */
 			clearable: { type: Boolean },
+			/**
+			 * Acts as a unique identifier for the tag in dispatched events
+			 * @type {string}
+			 */
+			key: { type: String },
```
bearfriend, line 34: Should this be moved to the `tag-list-item` component, and then mixin consumers can set up their own if needed?

components/tag-list/tag-list-item-mixin.js
svanherk: Yeah I agree, but do you think it'd help people "do the right thing"? Encourages using the event detail to get the data from the data array, rather than hooking up an anonymous function to each item passing in data like a few places are doing?

components/tag-list/tag-list-item-mixin.js
```diff
@@ -212,12 +218,10 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		`;
 	}
 
-	_renderTag(tagContent, options) {
-		if (!options) options = {};
+	_renderTag(tagContent, options = {}) {
+		this._displayText = options.displayText || (tagContent?.constructor === String ? tagContent : '');
```
svanherk, line 222: Maybe we just make `displayText` (or maybe `summaryText` to really differentiate it from the tag's text) a required second parameter? Would be a breaking change but feels a little cleaner, since we do _need_ something to announce and use as the accessible text for the icon button

components/tag-list/tag-list-item-mixin.js
bearfriend: Yeah, I guess it could, and it doesn't hurt any to leave it here.

components/tag-list/tag-list-item-mixin.js
bearfriend: Based on the conversation above, leaving this here might help encourage proper data retrieval for mixin consumers too.

components/tag-list/tag-list-item-mixin.js
dbatiste: I'm not sure it's going to prevent that. If they're not inclined to use `e.target.key` then I suspect they won't use `e.detail.key` either. I usually opt for not duplicating the data that just as easily accessed, but as you said, it doesn't really hurt.

components/tag-list/tag-list-item-mixin.js
dbatiste: Agreed. And per offline conversation, if we are going to prove this as part of the event detail in the mixin, then I think it makes sense for this `key` to be defined in the mixin.

dbatiste (approved):
This is looking good to be. @svanherk , do you have anything further for this PR?

components/tag-list/demo/tag-list.html
```diff
@@ -71,8 +71,8 @@ <h2>Clearable Tag List</h2>
 					import { getNextFocusable } from '../../../helpers/focus.js';
 
 					document.addEventListener('d2l-tag-list-item-clear', (e) => {
+						console.log(`d2l-tag-list-item-clear event dispatched. Value: ${e.target.text}`);
```
svanherk, line 74: Should we add keys to the demo, as a good example? And then we can add it hear alongside the value, to show how to get both

components/tag-list/tag-list-item-mixin.js
```diff
@@ -27,6 +27,11 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			 * @type {boolean}
 			 */
 			clearable: { type: Boolean },
+			/**
+			 * Acts as a unique identifier for the tag
+			 * @type {string}
+			 */
+			key: { type: String },
```
svanherk, line 34: Do we want to make this as "Required" in the docs? 🤔 It doesn't really break anything if you don't have it, but the event _does_ expect it.

components/tag-list/test/tag-list.test.js
```diff
@@ -91,6 +91,24 @@ describe('d2l-tag-list-item', () => {
 		});
 	});
 
+	describe('display text', () => {
+		it('should default to tagContent', async() => {
```
svanherk, line 95:
```suggestion
		it('should use text for tag-list-item', async() => {
```
Looks like this isn't really accurate anymore, maybe something like this?

svanherk (approved):

components/tag-list/tag-list-item-mixin.js
```diff
@@ -209,12 +215,11 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		`;
 	}
 
-	_renderTag(tagContent, options) {
-		if (!options) options = {};
+	_renderTag(tagContent, options = {}) {
+		if (options.labelText?.constructor !== String) throw new TypeError('options.labelText must be a string');
```
svanherk, line 219: Oh, I guess we'll want to have the `users` PR ready to go to merge right after this? Or the user tag items will start throwing?

bearfriend: Oh yikes! Yep.

PR #3775, opened 2023-06-22
Remove `z-index` and stacking contexts from list-items

bearfriend:
[DE52324](https://rally1.rallydev.com/#/?detail=/defect/689632043385&fdp=true): Html emoji editor can get cut off by the rubric
[US151482](https://rally1.rallydev.com/#/?detail=/userstory/698053793761&fdp=true): Spike > Update d2l-list item rendering to avoid z-index

Removes all usage of `z-index` (and any internal stacking contexts) in list items to allow all descendants, and the list item itself, to stack organically within the document. This means no more special handling is needed to support open dropdowns, tooltips, or other overflowing descendants.

CSS `order` achieves essentially the same result as `z-index`, but it is scoped to the grid items and simply re-orders them instead of continually increasing their priority compared to everything else, so there is no arms race.

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -597,7 +597,6 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				align-nested="${ifDefined(this.draggable && this.selectable ? 'control' : undefined)}"
 				@focusin="${this._onFocusIn}"
 				@focusout="${this._onFocusOut}"
-				@d2l-fullscreen-within="${this._onFullscreenWithin}"
 				class="${classMap(classes)}"
 				data-breakpoint="${this._breakpoint}"
 				data-separators="${ifDefined(this._separators)}"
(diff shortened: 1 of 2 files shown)


components/list/list-item-mixin.js
```diff
@@ -126,9 +114,10 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			}
 
 			[slot="control-container"] {
+				pointer-events: none;
```
bearfriend, line 117: `position: relative;` seems to bump this item out of control of `order`, so I left it on top but disallowed `pointer-events`, which seems to work fine.

dbatiste: No objection to changing this, just calling this out if we run into any issues with it. I don't _think_ this `z-index` value was causing us any grief. It was just being used for rendering the separators in its `::before`/`::after` pseudo-elements. The element itself is not actually a container for item content, and so would not be setting a stacking context for anything else.

dbatiste (approved):
Looks great!

components/list/list-item-mixin.js
bearfriend: The problem was that without the parent stacking context, `-1` puts it behind the list itself (or whatever it eventually hits) so the borders were no longer visible.

PR #3781, opened 2023-06-23
Fix misaligned list borders

bearfriend:
Currently, the borders between list items are not evenly spaced between the contents of those list items and also do not line up with the borders around a list item when it is hovered, causing a jitter/jumping effect.

Before and after:
bad-borders good-borders

Once visual diff finishes I'll run this by Jeff before proceding.


bearfriend:
#3782 is the vdiff PR to review the report. I split it into two because the second one just changes the height of everything, producing a useless report.

dbatiste (approved):

PR #3791, opened 2023-06-28
Move tag list tests to proper browsers interactions

bearfriend:
Uses proper browser mouse/keyboard interactions from the testing library instead of javascript methods/events to simulate them.

### DIFF
--- a/components/tag-list/test/tag-list.test.js
+++ b/components/tag-list/test/tag-list.test.js
@@ -1,7 +1,7 @@
 import './tag-list-item-mixin-consumer.js';
 import '../tag-list.js';
 import '../tag-list-item.js';
-import { expect, fixture, html, nextFrame, oneEvent, waitUntil } from '@brightspace-ui/testing';
+import { clickElem, expect, fixture, html, oneEvent, sendKeysElem } from '@brightspace-ui/testing';
 import { getComposedActiveElement } from '../../../helpers/focus.js';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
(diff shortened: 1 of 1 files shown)


components/tag-list/test/tag-list.test.js
```diff
@@ -73,10 +51,9 @@ describe('d2l-tag-list', () => {
 	describe('clearable', () => {
 		it('should dispatch expected events when Clear All clicked', async() => {
 			const elem = await fixture(clearableFixture);
-			await waitUntil(() => elem._items, 'List items did not become ready');
 			const button = elem.shadowRoot.querySelector('d2l-button-subtle.d2l-tag-list-clear-button');
 
-			setTimeout(() => button.click());
+			clickElem(button);
```
dlockhart, line 56: I think we still want to do this kind of thing in a timeout in case the event gets fired synchronously from the click handler.

bearfriend: I wasn't getting any flake locally without it, and I _think_ it might always be safe to do this without, but it'd be good to know definitively either way before we convert/write more of these. I'll look into it.

bearfriend: Also, if it's not safe, maybe we wrap them so it is always safe?

dlockhart: Yeah maybe it makes sense for us to put the `setTimeout` inside all of these utilities eh? The Promises they return can still resolve after the action has complete if that's what they're going for, but if you don't care you can just call it and then `await` an event on the next line.

bearfriend: Looked into this and discussed offline. There are multiple steps in other processes that make these interactions inherently significantly delayed (in the event loop, not actually time) so for now we're thinking these should be safe to run in this order without intervention, but we can always build in a delay later if needed.

svanherk (approved):

PR #3792, opened 2023-06-28
Add lines property to d2l-collapsible-panel-summary-item.

dbatiste:
[US154543](https://rally1.rallydev.com/#/?detail=/userstory/706029989793&fdp=true)

This PR adds an optional `lines` property to `d2l-collapsible-panel-summary-item` that enables consumers to specify the number of lines to display before truncating with ellipsis. If no value, or a value less than 1 is specified, it will not truncate as consistent with previous behaviour.

**Example: `lines="2"`**
image


components/collapsible-panel/collapsible-panel-summary-item.js
```diff
@@ -39,7 +52,13 @@ class CollapsiblePanelSummaryItem extends SkeletonMixin(LitElement) {
 	}
 
 	render() {
-		return html`<p class="d2l-body-small d2l-skeletize">${this.text}</p>`;
+		const classes = {
+			'd2l-body-small': true,
+			'd2l-skeletize': true,
+			'truncate': this.lines > 0
+		};
+		const styles = (this.lines > 0) ? { '-webkit-line-clamp': this.lines } : {};
```
jameswklassen, line 60: Why aren't we adding this rule in the `truncate` definition?

dbatiste: Just because the value is coming from the consumer-specified attribute. We _could_ set a CSS variable in one of the life cycle methods to use in CSS, but seems unnecessary complexity for this one property. It would be amazing if we could just use [attr](https://developer.mozilla.org/en-US/docs/Web/CSS/attr), but it would not work in this context... maybe one day.

jameswklassen (approved):

PR #3796, opened 2023-06-29
feat: apply hover effects anywhere on the component

GZolla:
Apply the input hover effects when hovering over anywhere on the component, not just the input or label.
Rally: [US148080](https://rally1.rallydev.com/#/?detail=/userstory/684245679355&fdp=true)

### DIFF
--- a/components/inputs/input-styles.js
+++ b/components/inputs/input-styles.js
@@ -25,7 +25,7 @@ export const inputStyles = css`
 		width: 100%;
 	}
 	.d2l-input,
-	:host(:hover) .d2l-input:disabled,
+	.d2l-input:hover:disabled,
 	.d2l-input:focus:disabled,
 	[aria-invalid="true"].d2l-input:disabled {
 		border-color: var(--d2l-input-border-color, var(--d2l-color-galena));
@@ -43,18 +43,14 @@ export const inputStyles = css`
 		font-size: 0.8rem;
 		font-weight: 400;
 	}
-	:host(:hover) .d2l-input,
+	.d2l-input:hover,
 	.d2l-input:focus,
 	.d2l-input-focus {
+		border-color: var(--d2l-color-celestine);
 		border-width: 2px;
 		outline-style: none;
 		outline-width: 0;
 		padding: var(--d2l-input-padding-focus, calc(0.4rem - 1px) calc(0.75rem - 1px));
-	}
-	:host(:hover) .d2l-input:not([aria-invalid="true"]),
-	.d2l-input:focus,
-	.d2l-input-focus {
-		border-color: var(--d2l-color-celestine);
 	}
 	[aria-invalid="true"].d2l-input {
 		border-color: var(--d2l-color-cinnabar);
@@ -75,13 +71,13 @@ export const inputStyles = css`
 		line-height: normal;
 	}
 	textarea.d2l-input,
-	:host(:hover) textarea.d2l-input:disabled,
+	textarea.d2l-input:hover:disabled,
 	textarea.d2l-input:focus:disabled,
 	textarea[aria-invalid="true"].d2l-input:disabled {
 		padding-bottom: 0.5rem;
 		padding-top: 0.5rem;
 	}
-	:host(:hover) textarea.d2l-input,
+	textarea.d2l-input:hover,
 	textarea.d2l-input:focus {
 		padding: var(--d2l-input-padding-focus, calc(0.75rem - 1px));
 		padding-bottom: calc(0.5rem - 1px);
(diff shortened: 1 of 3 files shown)


components/inputs/input-styles.js
dbatiste, line 1: I think this changes the `disabled` hover styles too, which I don't think is expected.

components/inputs/input-styles.js
```diff
@@ -43,15 +43,19 @@ export const inputStyles = css`
...
+	:host(:hover) .d2l-input,
 	.d2l-input:focus,
 	.d2l-input-focus {
-		border-color: var(--d2l-color-celestine);
 		border-width: 2px;
 		outline-style: none;
 		outline-width: 0;
 		padding: var(--d2l-input-padding-focus, calc(0.4rem - 1px) calc(0.75rem - 1px));
 	}
+	:host(:hover) .d2l-input:not([aria-invalid="true"]),
```
dbatiste, line 54: Targeting `:host(:hover)` assumes that the consumer of these styles is a simple input component wrapper, not a component composed of multiple elements. I didn't dig into them but it does appear that [app components are also consuming `inputStyles` directly](https://search.d2l.dev/search?project=Brightspace&project=Brightspace-Configs&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22inputStyles%22&defs=&refs=&path=&hist=&type=&xrd=&nn=6&si=full&si=full). I think we need to assume they could be pulling those styles into a huge component composed of lots of different things, possibly even multiple inputs. Even if those cases are "ok" with this change, this does still make me feel a bit uneasy. I think it would be better if there was an element that we could target that would isolate this behaviour, like one of the wrapper `div`s. 🤔

dlockhart: Agreed. These input styles are intended to be used standalone on a `<input class="d2l-input">` and as Dave mentioned the "host" may be a much larger component composed of many different things.

I think if we want to do this it'll have to be something constrained to `<d2l-input-text>`, which is the thing that knows about the label. We may need to programatically add/remove a `.d2l-input-hover` CSS class when the host is hovered/blurred, and then add CSS here to handle it -- similar to how this already understands `.d2l-input-focus`.

components/inputs/input-textarea.js
```diff
@@ -194,6 +197,20 @@ class InputTextArea extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMi
...
 	}
 
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.removeEventListener('mouseover', this._handleMouseEnter);
+		this.removeEventListener('mouseout', this._handleMouseLeave);
+		this.removeEventListener('click', this._handleClick);
+	}
+
+	firstUpdated(changedProperties) {
```
dlockhart, line 207: I know `d2l-input-text` does it this way, but I think these should be in `connectedCallback`. Otherwise, disconnecting will remove them and re-connecting won't add them again (since `firstUpdated` only runs the first time).

components/inputs/input-textarea.js
```diff
@@ -86,7 +87,8 @@ class InputTextArea extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMi
 			 * Value of the input
 			 * @type {string}
 			 */
-			value: { type: String }
+			value: { type: String },
+			_hovered: { type: Boolean }
```
dlockhart, line 91: We can now use reactive state properties for things we don't want to make public like this.

```suggestion
			_hovered: { state: true }
```

components/inputs/input-text.js
```diff
@@ -338,8 +338,9 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 		if (!container) return;
 		container.removeEventListener('blur', this._handleBlur, true);
 		container.removeEventListener('focus', this._handleFocus, true);
-		container.removeEventListener('mouseover', this._handleMouseEnter);
-		container.removeEventListener('mouseout', this._handleMouseLeave);
+		this.removeEventListener('mouseover', this._handleMouseEnter);
+		this.removeEventListener('mouseout', this._handleMouseLeave);
+		this.removeEventListener('click', this._handleClick);
```
dlockhart, line 343: Since these don't rely on `container`, they should likely move up above `if (!container) return`.

components/inputs/input-text.js
```diff
@@ -542,6 +544,12 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 		));
 	}
 
+	_handleClick(e) {
+		const input = this.shadowRoot && this.shadowRoot.querySelector('.d2l-input');
```
dlockhart, line 548: We can use optional chaining now for this kind of thing:
```suggestion
		const input = this.shadowRoot?.querySelector('.d2l-input');
```

dlockhart (approved):

PR #3798, opened 2023-06-29
Recursively check for last item in nested lists

bearfriend:
[DE53272](https://rally1.rallydev.com/#/?detail=/defect/700010022767&fdp=true): List > Grid navigation not navigating properly to deeply nested list items

When looking for the previous item to focus in grid mode, the last item should be checked recursively for nested lists.

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -407,12 +407,17 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			if (previousElement.role === 'rowgroup') {
 
 				// this check needs to account for standard list-items as well as custom
-				const nestedList = previousElement.querySelector('[slot="nested"]') || previousElement.shadowRoot.querySelector('d2l-list');
+				let nestedList = previousElement.querySelector('[slot="nested"]') || previousElement.shadowRoot.querySelector('d2l-list');
 				if (nestedList) {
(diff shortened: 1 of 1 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -407,12 +407,17 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
-						return nestedListItems[nestedListItems.length - 1];
+					let nestedListItems = [...nestedList.children].filter(node => node.role === 'rowgroup');
+
+					let lastItem;
+					while (nestedListItems?.length) {
+						lastItem = nestedListItems[nestedListItems.length - 1];
+						nestedList = lastItem.querySelector('[slot="nested"]') || lastItem.shadowRoot.querySelector('d2l-list');
+						nestedListItems = nestedList && [...nestedList.children].filter(node => node.role === 'rowgroup');
 					}
+					return lastItem;
```
bearfriend, line 420: This should probably use `do...while`. I always forget about that.

dlockhart, line 420: Previously, this would have only returned something if the nested list had items -- otherwise it would have fallen out of this block and returned `previousElement`. I think now if `nestedItems` initially has nothing in it, `lastItem` won't get set to anything and the whole thing will return `undefined`.

dlockhart (approved):
Looks good!

PR #3800, opened 2023-06-29
Switch keydown uses to sendKeysElem

svanherk:
Switch all uses of `keydown` to `switchKeysElem`.  We can't remove `keydown` yet - it's still used in `copy-role` (removed with https://github.com/Brightspace/copy-role/pull/98) and [`d2l-polymer-behaviors-ui`](https://github.com/Brightspace/d2l-pdf-viewer/blob/6cf0a246f150c8167c7bf9beef305c1cb8e9222d/d2l-pdf-viewer-toolbar.js#L4).

### DIFF
--- a/components/button/test/button-move.test.js
+++ b/components/button/test/button-move.test.js
@@ -1,4 +1,4 @@
-import { expect, fixture, html, oneEvent, sendKeysElem } from '@brightspace-ui/testing';
+import { clickElem, expect, fixture, html, oneEvent, sendKeysElem } from '@brightspace-ui/testing';
 import { moveActions } from '../button-move.js';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
@@ -15,8 +15,8 @@ describe('d2l-button-move', () => {
 	describe('events', () => {
 
 		[
-			{ name: 'up icon clicked', action: el => setTimeout(() => el.shadowRoot.querySelector('.up-layer').click()), expectedAction: moveActions.up },
-			{ name: 'down icon clicked', action: el => setTimeout(() => el.shadowRoot.querySelector('.down-layer').click()), expectedAction: moveActions.down },
+			{ name: 'up icon clicked', action: el => clickElem(el.shadowRoot.querySelector('.up-layer')), expectedAction: moveActions.up },
+			{ name: 'down icon clicked', action: el => clickElem(el.shadowRoot.querySelector('.down-layer')), expectedAction: moveActions.down },
 			{ name: 'up key pressed', action: el => sendKeysElem(el.shadowRoot.querySelector('button'), 'press', 'ArrowUp'), expectedAction: moveActions.up },
 			{ name: 'down key pressed', action: el => sendKeysElem(el.shadowRoot.querySelector('button'), 'press', 'ArrowDown'), expectedAction: moveActions.down },
 			{ name: 'left key pressed', action: el => sendKeysElem(el.shadowRoot.querySelector('button'), 'press', 'ArrowLeft'), expectedAction: moveActions.left },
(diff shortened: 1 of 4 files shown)


tools/dom-test-helpers.js
```diff
@@ -1,6 +1,6 @@
 import { getComposedChildren } from '../helpers/dom.js';
 
-export function keyDown(element, keyCode, ctrlKey) {
+export function keyDown(element, keyCode) {
```
svanherk, line 3: None of the remaining uses set this, so figured we could at least clean that up now

mixins/interactive/test/interactive-mixin.test.js
```diff
@@ -57,8 +56,8 @@ describe('InteractiveMixin', () => {
 			toggle.click();
 			await new Promise(resolve => setTimeout(resolve, 0));
 			expect(getComposedActiveElement()).to.equal(elem.shadowRoot.querySelector('.content-button'));
-			keyDown(toggle, 27);
-			await new Promise(resolve => setTimeout(resolve, 0));
+			sendKeysElem(toggle, 'press', 'Escape');
+			await waitUntil(() => getComposedActiveElement().classList.contains('interactive-toggle'));
```
svanherk, line 60: `sendKeysElem` seems to be a lot slower, so waiting for something more concrete.

dlockhart (approved):
Nice! Side benefit: "ArrowRight" makes way more sense than "38". 😆

mixins/interactive/test/interactive-mixin.test.js
svanherk: Looks like yes 🤦. Totally forgot this function was `async` 😅, every other place using it needs to wait for a certain event anyways

components/button/test/button-move.test.js
```diff
@@ -16,20 +15,20 @@ describe('d2l-button-move', () => {
...
-			{ name: 'left key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 37), expectedAction: moveActions.left },
-			{ name: 'right key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 39), expectedAction: moveActions.right },
-			{ name: 'home key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 36), expectedAction: moveActions.home },
-			{ name: 'ctrl-home key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 36, true), expectedAction: moveActions.rootHome },
-			{ name: 'end key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 35), expectedAction: moveActions.end },
-			{ name: 'ctrl-end key pressed', action: el => keyDown(el.shadowRoot.querySelector('button'), 35, true), expectedAction: moveActions.rootEnd }
+			{ name: 'up icon clicked', action: el => setTimeout(() => el.shadowRoot.querySelector('.up-layer').click()), expectedAction: moveActions.up },
+			{ name: 'down icon clicked', action: el => setTimeout(() => el.shadowRoot.querySelector('.down-layer').click()), expectedAction: moveActions.down },
```
dlockhart, line 19: Can these two just use `clickElem` and then not require the `setTimeout`?

svanherk: Yeah already updated lol, realized that when reviewing Danny's PR

mixins/interactive/test/interactive-mixin.test.js
bearfriend: Yeah it's a little weird to remember at first. There will definitely be uses of both, like:

```javascript
sendKeysElem(elem, 'press', 'Tab');
await oneEvent(elem, 'blur');
```
and:
```javascript
await sendKeysElem(elem, 'press', 'Tab');
expect(elem.classList).to.not.include(['active']);
```

svanherk: Yeah exactly. It's so nice though, so far everything "just works" (as long as you remember the above 😝)

bearfriend (approved):

PR #3807, opened 2023-07-05
Remove direct stylelint dependency and rely on our config's dependencies.

dbatiste:
This PR removes the direct `stylelint` dependency so that we rely only on the dependency specified by our `@brightspace-ui/stylelint-config`. Removing this dependency helps to avoid conflicts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,7 @@
     "rollup": "^3",
     "rollup-plugin-copy": "^3",
     "rollup-plugin-delete": "^2",
-    "sinon": "^15",
-    "stylelint": "^15"
+    "sinon": "^15"
   },
   "dependencies": {
     "@brightspace-ui/intl": "^3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,8 +38,7 @@
         "rollup": "^3",
         "rollup-plugin-copy": "^3",
         "rollup-plugin-delete": "^2",
-        "sinon": "^15",
-        "stylelint": "^15"
+        "sinon": "^15"
       }
     },
     "node_modules/@75lb/deep-merge": {


package.json
```diff
@@ -62,8 +62,7 @@
     "rollup": "^3",
     "rollup-plugin-copy": "^3",
     "rollup-plugin-delete": "^2",
-    "sinon": "^15",
-    "stylelint": "^15"
+    "sinon": "^15"
```
dlockhart, line 65: Is this wise given that we call it directly from a NPM script? I think typically we try to avoid this kind of thing in case the other dependency _stopped_ including it one day. Obviously we own that other dependency in this case, so the concern is lessesned.

svanherk: I think the plan is we're going to move this from a direct dependency to a peer dependency in `@brightspace-ui/stylelint-config` so `core` will still install it, but the version we want.

Still a similar problem, if we eventually remove it there.  But otherwise we'll need to add an override to every place we install it until it's fixed.  We could technically wrap it like we plan to do in `testing`, and then even if we switch out `stylelint` to something else one day, we can keep the same command to do the linting 🤔

dlockhart: Right. Yeah I was thinking about what wrapping it would look like as well... it's an interesting option, but not one that we can quickly rush out.

dbatiste: Yeah that's a good point. I think that's why this was kept around. Unfortunately this currently conflicts with the pinned versions in our config, so I think we'd need to pin this also if we keep it. I see we kept `eslint` around, likely with the same rationale. 🤔

dbatiste: Was just thinking about the wrapping idea more... the `stylelint \"**/*.{js,html}\" ...` isn't the only place where `stylelint` is surfaced to the consumer. There 's also the `.stylelintrc.json` file where the consumer can specify the config (they should be using ours) and also overrides. Technically the consumer could put something into the `.stylelintrc.json` that requires a certain version of `stylelint` - a reason to have the consumer specify the dependency, although in practice it probably isn't an issue.

I wonder if I can get around this issue of both versions being installed by using `"stylelint": ">=15"` instead of `"stylelint": "^15"` in the consumer? Maybe then it would resolve to 15.9.0 specified in the config's package.

dbatiste: That didn't work, so I think we're either faced with pinning in consumers, or removing it from consumers.

dbatiste:
Closing since the stylistic rules plugin is now fixed. No immediate need to remove this dependency.

PR #3808, opened 2023-07-06
fix: Safari list item slot order

bearfriend:
[DE54029](https://rally1.rallydev.com/#/?detail=/defect/706646564733&fdp=true): Safari list-item slot order

After https://github.com/BrightspaceUI/core/pull/3775, list items were broken in Safari because it renders the slots out of order in a very strange way. After giving each slot an explicit `order` CSS property or rearranging the DOM, Safari renders the slots in the correct order _visually_, but attempting to interact with the slots that were on top both in the DOM and visually still did not work, with pointer events being sent to the slots behind for no apparent reason.

To fix this, I needed to both re-order the slots in the DOM _and_ strategically disable pointer events or change the columns on slots that should not be interactive in each scenario. This allows the order to be purely DOM-based, without `order` or `z-index`.

I also did some pretty minor cleanup of things that seemed unnecessary or poorly named.

### DIFF
--- a/components/list/list-item-generic-layout.js
+++ b/components/list/list-item-generic-layout.js
@@ -68,71 +68,58 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 					[content-end actions-start] minmax(0, min-content)
 					[end actions-end];
 				grid-template-rows: [main-start] [main-end nested-start] [nested-end];
-				//position: relative;
 			}
 
-			::slotted(*) {
-				//order: 0;
-				//position: relative;
-			}
-
-			::slotted([slot="nested"]) {
-				grid-column: content-start / end;
-				grid-row: nested-start / nested-end;
-			}
 			:host([align-nested="control"]) ::slotted([slot="nested"]) {
 				grid-column: control-start / end;
 			}
 			:host(.d2l-dragging-over) ::slotted([slot="nested"]) {
-				//order: 11; /* must be greater than item's drop-target to allow dropping onto items within nested list  */
-				z-index: 2; /* fix for webkit order limitation */
+				z-index: 2; /* webkit dom order fix */
 			}
 
 			::slotted([slot="drop-target"]) {
-				//height: 100%;
-				//order: 10;
-				//position: absolute;
-				//top: 0;
-				//width: 100%;
 				grid-column: 1 / -1;
-				grid-row: 1 / 2;
 			}
 
 			:host(.d2l-dragging-over) ::slotted([slot="drop-target"]) {
-				z-index: 1; /* fix for webkit order limitation */
+				z-index: 1; /* webkit dom order fix */
 			}
 
 			::slotted([slot="outside-control"]),
 			::slotted([slot="expand-collapse"]),
 			::slotted([slot="control"]),
 			::slotted([slot="content"]),
-			::slotted([slot="actions"]) {
+			::slotted([slot="actions"]),
+			::slotted([slot="outside-control-action"]),
+			::slotted([slot="control-action"]),
+			::slotted([slot="content-action"]),
+			::slotted([slot="outside-control-container"]),
+			::slotted([slot="control-container"]),
+			::slotted([slot="drop-target"]) {
 				grid-row: 1 / 2;
 			}
 			::slotted([slot="outside-control"]) {
 				grid-column: outside-control-start / outside-control-end;
-				//order: 2;
 			}
 
 			::slotted([slot="expand-collapse"]) {
 				cursor: pointer;
 				grid-column: expand-collapse-start / expand-collapse-end;
-				//order: 5;
 			}
 
 			::slotted([slot="control"]) {
 				grid-column: control-start / control-end;
-				//order: 6;
 				width: 2.2rem;
 			}
 
 			::slotted([slot="content"]) {
 				grid-column: content-start / content-end;
-				//order: 4;
 			}
 
-			:host([no-primary-action]) ::slotted([slot="content"]) {
-				pointer-events: none;
+			:host(:not([no-primary-action])) ::slotted([slot="content"]),
+			::slotted([slot="control"]),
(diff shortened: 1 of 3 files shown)


components/list/list-item-generic-layout.js
```diff
@@ -225,20 +237,33 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 		}
 	}
 
+	_focusFirstCell() {
```
bearfriend, line 240: In a previous attempt, I changed the `data-cell-num` from 1 through 8 to -5 through 2, which meant I had to make other changes to handle that. It shouldn't be necessary anymore, but I don't hate starting at 0 to indicate where focus starts so I left it in for now.

components/list/list-item-generic-layout.js
```diff
@@ -68,33 +68,34 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
+				z-index: 2; /* webkit dom order fix */
 			}
 
 			::slotted([slot="drop-target"]) {
-				height: 100%;
-				order: 6;
-				position: absolute;
-				top: 0;
-				width: 100%;
+				grid-column: 1 / -1;
```
dbatiste, line 81: Is it possible to use the grid line names?

components/list/list-item-generic-layout.js
```diff
@@ -225,20 +237,33 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
 		if (this.dir === 'rtl') {
-			row._focusLastItem();
+			row._focusLastCell();
 		} else {
-			row._focusNextCell(1);
+			row._focusFirstCell();
 		}
 	}
 
-	_focusLastItem() {
```
dbatiste, line 238: I don't feel strongly about it, but I _think_ the "Item" term was being used because it is focusing the last item in the cell.

components/list/list-item-generic-layout.js
bearfriend: I had previously moved things around so `content-actions` was not the first cell in the DOM order and I was retargetting focus to first cell (which was `6`) manually on `focusin`, and at the point neither the DOM order or the cell numbers really made any sense, so I wanted to be able to say `_focusCellItem(0)` and have it go to our starting position.

However, I agree with reverting to the original.

components/list/list-item-generic-layout.js
```diff
@@ -68,33 +68,34 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 					[content-end actions-start] minmax(0, min-content)
 					[end actions-end];
 				grid-template-rows: [main-start] [main-end nested-start] [nested-end];
-				position: relative;
```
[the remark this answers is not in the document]
bearfriend: I think that was being used for the absolutely positioned `drop-target`, which was changed to just fill the row.

components/list/list-item-generic-layout.js
bearfriend: I wanted to avoid names for this one since it should likely fill the row regardless of future column changes.

components/list/list-item-generic-layout.js
bearfriend: That makes sense. Yeah it was certainly inconsistent. I can go back through and try to add that context everywhere if it makes sense.

components/list/list-item-generic-layout.js
```diff
@@ -197,20 +205,24 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
...
-			<slot name="content-action" class="d2l-cell" data-cell-num="6"></slot>
-			<slot name="outside-control-action" class="d2l-cell" data-cell-num="1"></slot>
-			<slot name="outside-control" class="d2l-cell" data-cell-num="2"></slot>
-			<slot name="control-action" class="d2l-cell" data-cell-num="3"></slot>
-			<slot name="expand-collapse" class="d2l-cell" data-cell-num="4"></slot>
-			<slot name="control" class="d2l-cell" data-cell-num="5"></slot>
-			<slot name="actions" class="d2l-cell" data-cell-num="7"></slot>
-			<slot name="content" class="d2l-cell" data-cell-num="8" @focus="${!this.noPrimaryAction ? this._preventFocus : null}"></slot>
+
+			<slot name="content-action" class="d2l-cell" data-cell-num="0"></slot>
```
dbatiste, line 211: Original order...
```
1 ::slotted([slot="outside-control-action"]) {
2 ::slotted([slot="expand-collapse"]) {
3 ::slotted([slot="control-action"]) {
4 ::slotted([slot="content-action"]) {
5 ::slotted([slot="actions"]) {
6 ::slotted([slot="drop-target"]) ➡️  now using `z-index: 1`
7 :host(.d2l-dragging-over) ::slotted([slot="nested"]) ➡️ now using `z-index: 2`
```
With these changes, it looks like `content-action` is being moved to be underneath everything, where is was previously on top of `control-action`. Does that cause issues?

bearfriend: Also the `z-index` for drag-n-drop probably aren't necessary anymore either.

components/list/list-item-generic-layout.js
dbatiste: Would that correspond to the `start` and `end` grid names that are defined above? Not a big deal though.

components/list/list-item-generic-layout.js
dbatiste: Awesome. Thanks for the explanation. 😄

dbatiste (approved):
I did a little bit of manual testing with it and all seems to be "ok".

Just the follow-up about the start/end grid lines, but nbd.

Looks good.

components/list/list-item-generic-layout.js
bearfriend: Hah, yep. I forgot they were named twice like the others. That's totally reasonable. I'll switch to those.

PR #3809, opened 2023-07-06
feat: Add `SkeletonGroupMixin`

jameswklassen:
## Description
Add a new `SkeletonGroupMixin` which allows consumers to wrap multiple skeleton components and ensure they all appear as "loaded" at the same time.

## Rally
[US154180](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F705086286295&parentRef=%2Fportfolioitem%2Finitiative%2F672628034037)


components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,19 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
```
dlockhart, line 2: Why does it need `RtlMixin`?

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,19 @@
...
+	constructor() {
+		super();
+
+		this._skeletonSubscribers = new SubscriberRegistryController(this, 'skeleton', {
+			updateSubscribers: this._checkSubscribersReadyToDisplay.bind(this)
+		});
+	}
+
+	_checkSubscribersReadyToDisplay(subscribers) {
+		const readyToDisplay = ![...subscribers.values()].some(subscriber => subscriber.skeleton);
```
dlockhart, line 16: Would it work to keep track of the previous `readyToDisplay` value and only update the subscribers if it's changed? Just thinking of a case where there's like 100 items and they each become ready independently but the whole thing isn't ready until they all are.

components/collapsible-panel/collapsible-panel.js
```diff
@@ -112,7 +112,7 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
 				border: 1px solid var(--d2l-color-mica);
 				border-radius: 0.4rem;
 			}
-			:host(:not([expanded]):not([skeleton])) .d2l-collapsible-panel {
+			:host(:not([expanded]):not([skeleton-display])) .d2l-collapsible-panel {
```
dlockhart, line 115: There are a lot of components using `SkeletonMixin` with custom CSS like this outside of core that will need the same update (if they want to work with groups). I can't think of a way around it, but just wanted to point it out.

We'll want to update the documentation accordingly.

components/skeleton/skeleton-mixin.js
```diff
@@ -155,7 +156,11 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 			 * Renders the input as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton)
 			 * @type {boolean}
 			 */
-			skeleton: { reflect: true, type: Boolean  }
+			skeleton: { reflect: true, type: Boolean  },
+			skeletonDisplay: { reflect: true, type: Boolean, attribute: 'skeleton-display' },
+			_parentSkeleton: { state: true },
```
dlockhart, line 161: Does `_parentSkeleton` need to be a state property, or could it just be a regular private variable? The only advantage to it being state would be that changing it will trigger a re-render, but it seems like we only care about re-rendering when `skeleton` or `skeletonDisplay` change.

Similarly for `_readyToDisplay` and `_waitToDisplay` -- I don't think they need to reflect to attributes and can also just be private variables.

components/skeleton/skeleton-mixin.js
```diff
@@ -168,6 +173,40 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+	}
+
+	changeDisplay(renderContent) {
+		this._readyToDisplay = renderContent;
+		this._updateSkeletonDisplay();
 	}
 
+	_onSubscribe() {
+		this._waitToDisplay = true;
+		this._readyToDisplay = false;
```
dlockhart, line 201: Likely need to call `_updateSkeletonDisplay` on-subscribe and on-unsubscribe, since its value depends on these values.

dlockhart (commented):
I like this approach! And it ended up being pretty simple, which is nice.

components/skeleton/skeleton-group-mixin.js
jameswklassen: it doesn't! just haven't cleaned that part up yet.

components/skeleton/skeleton-group-mixin.js
jameswklassen: Oh yeah, that's a good optimization. I'll play with that.

components/skeleton/skeleton-mixin.js
jameswklassen: Ah yeah, I forgot to take out `reflect` stuff...I originally was playing around directly with these variables.

When you say private variable, do you mean setting them as `this._x`, but not declaring in properties?

dlockhart: Exactly yeah -- just set it in the constructor.

components/skeleton/skeleton-mixin.js
```diff
@@ -155,7 +156,8 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 			 * Renders the input as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton)
 			 * @type {boolean}
 			 */
-			skeleton: { reflect: true, type: Boolean  }
+			skeleton: { reflect: true, type: Boolean  },
```
dlockhart, line 159: So it doesn't get forgotten once this goes out of Draft: this needs documentation.

components/collapsible-panel/collapsible-panel.js
dlockhart: Cool. Alternate name that comes to mind: `skeleton-active`

svanherk: And at least changing this can be updated gradually (worst case, you un-skeletize early and ignore the rest of the group) versus changing the property name and having to update every consumer to avoid breaking things. Could we reflect until we've updated everyone, then stop reflecting so it's clear no one should rely on the `skeleton` attribute going forward? 🤔 

Can't think of any good names 😬

components/skeleton/skeleton-mixin.js
```diff
@@ -155,7 +156,8 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+			skeletonDisplay: { reflect: true, type: Boolean, attribute: 'skeleton-display' },
```
svanherk, line 160: Or, maybe crazy idea... does this work, or totally break things?
```
skeleton: { type: Boolean, attribute: none },
_skeletonDisplay: { reflect: true, type: Boolean, attribute: 'skeleton' },
```
🤔 Like can we keep accept `skeleton` as the property, but then control the setting of the attribute? Lit might not like it, it might mean we'd need to update every consumer to target the property?

dlockhart: That's an interesting idea!

Another crazy one that I haven't fully thought through: can we create a getter/setter for the `skeleton` property and use that? Like the getter would return what you have for `skeleton-display` and the setter would save the value for later.

Then we wouldn't need to update any consumers.

components/collapsible-panel/collapsible-panel.js
jameswklassen: `skeleton-active` it is!

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,22 @@
...
+		super();
+
+		this._readyToDisplayCached = false;
+		this._skeletonSubscribers = new SubscriberRegistryController(this, 'skeleton', {
+			updateSubscribers: this._checkSubscribersReadyToDisplay.bind(this)
+		});
+	}
+
+	_checkSubscribersReadyToDisplay(subscribers) {
+		const readyToDisplay = ![...subscribers.values()].some(subscriber => subscriber.skeleton);
```
dlockhart, line 16: If there don't end up being any subscribers, I think this will remain in a skeleton state? But maybe that doesn't matter since the group itself doesn't have any skeleton display?

components/skeleton/skeleton-mixin.js
```diff
@@ -168,6 +170,55 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+			this._skeleton = val;
+			this.requestUpdate('skeleton', oldVal);
+			this.requestUpdate('_skeletonActive', oldVal);
+		}
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('skeleton')) {
+			this._updateDisplayState();
```
dlockhart, line 199: I don't think this call to `_updateDisplayState` is necessary since at this point `_waitToDisplay` won't have been changed (yet). When the below call happens to `updateSubscribers`, the controller _could_ call `changeDisplay` at which point it's going to update the display state.

components/skeleton/skeleton-mixin.js
```diff
@@ -168,6 +170,55 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+
+	_onUnsubscribe() {
+		this._waitToDisplay = false;
+		this._readyToDisplay = true;
+		this._updateDisplayState();
 	}
 
+	_updateDisplayState() {
+		if (this._waitToDisplay) { this.skeleton = !this._readyToDisplay; }
+	}
```
dlockhart, line 223: I would have expected this to be something more like:

```suggestion
	_updateDisplayState() {
		if (this._waitToDisplay) {
			this._skeletonActive = !this._readyToDisplay;
		} else {
			this._skeletonActive = this._skeleton;
		}
	}
```

Essentially, if there's no skeleton group in play (`_waitToDisplay` is `false`), then `_skeletonActive` mirrors whatever the consumer sets `skeleton` to. If there is a group (`_waitToDisplay` is `true`), then it mirrors `_readyToDisplay`.

components/skeleton/skeleton-group-mixin.js
jameswklassen: Yeah, I _think_ that's okay? `readyToDisplay` only gets set on the subscribers, so the case where there are no subscirbers seems to be a moot point.

We could add a short circuit `if (!subscribers) { return }` to be very clear.

svanherk:
So one question I kinda touched on in my slack brain dump that'll I'll put here too to get Dave's thoughts - can we have skeleton groups in skeleton groups?  Can something extend both `SkeletonGroupMixin` and `SkeletonMixin`, and will that just work as is (now that we are able to control `skeleton` without needing a new attribute)?

If it is something we want to support and does just work, a demo of that would be good. If it doesn't just work that could be a separate PR, but thought I'd raise it now

components/skeleton/skeleton-mixin.js
```diff
@@ -167,7 +168,54 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+
+	set skeleton(val) {
+		const oldVal = this._skeleton;
+		if (oldVal === val) return;
+		this._skeleton = val;
+		if (!this._skeletonWait) {
+			this._skeletonActive = val;
+			this.requestUpdate('skeleton', oldVal);
+		} else {
+			this._parentSkeleton._registryController?.updateSubscribers();
```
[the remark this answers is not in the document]
svanherk: Not sure I love `notifySkeletonChange`, even just `handleSubscriberChange` made public is fine?

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,28 @@
...
+
+		this._skeletonActiveCached = true;
+		this._skeletonSubscribers = new SubscriberRegistryController(this, 'skeleton', {
+			onSubscribe: this._handleSubscriberChange.bind(this),
+			onUnsubscribe: this._handleSubscriberChange.bind(this),
+			updateSubscribers: this._checkSubscribersReadyToDisplay.bind(this)
+		});
+	}
+
+	_checkSubscribersReadyToDisplay(subscribers) {
```
svanherk, line 17: This seems to be using the old terminology, maybe something like `_checkSubscribersSkeletonState`

controllers/subscriber/subscriberControllers.js
```diff
@@ -81,6 +81,13 @@ export class SubscriberRegistryController extends BaseController {
 		return this._subscribers;
 	}
 
+	handleSubscriberChange() {
+		if (!this._subscribers || this._subscribers.size === 0) return;
+		if (!this._options.handleSubscriberChange) return;
+
+		this._options.handleSubscriberChange(this._subscribers);
+	}
+
```
svanherk, line 90: It's been a while so maybe what I'm imagining doesn't work how I'd expect.  But what I was thinking was what you have now but with this deleted, and `handleSubscriberChange: this.handleSubscriberChange.bind(this)` deleted in `skeleton-group-mixin`.

components/skeleton/skeleton-group-mixin.js
jameswklassen: @dlockhart looks like the controller will already short circuit if there are no subscribers...so this seems fine as is.

https://github.com/BrightspaceUI/core/blob/c438ce8e4151f4846cbda0082ffb9d50986938a0/controllers/subscriber/subscriberControllers.js#L103-L115

dlockhart: Right, so it's not even going to call `updateSubscribers` on the group if there aren't any. Which I think will still keep the group in its initial skeleton state indefinitely?

components/skeleton/test/skeleton-group.visual-diff.html
```diff
@@ -0,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<link rel="stylesheet" href="../../../test/styles.css" type="text/css">
+		<script>
+			const rtl = (window.location.search.indexOf('dir=rtl') !== -1);
+			if (rtl) document.documentElement.setAttribute('dir', 'rtl');
```
dlockhart, line 7: You shouldn't need this stuff if you don't have any RTL tests.

components/skeleton/test/skeleton-group.visual-diff.js
```diff
@@ -0,0 +1,78 @@
...
+
+	[
+		{ name: 'make-one-skeleton', skeleton: true },
+		{ name: 'make-one-not-skeleton', skeleton: false },
+	].forEach((info) => {
+		it(info.name, async function() {
+			await page.$eval(`#${info.name}`, async(element, skeleton) => {
+				const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
+				const panel = panelGroup.querySelector('d2l-collapsible-panel');
+				panel.skeleton = skeleton;
```
dlockhart, line 44: You may want to `await panel.skeleton.updateComplete` here just to ensure that modifying the property has gone through Lit's update cycle.

components/skeleton/test/skeleton-group.visual-diff.js
```diff
@@ -0,0 +1,78 @@
...
+		it(name, async function() {
+			await page.$eval(`#${name}`, async(element) => {
+				const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
+				const panel = panelGroup.querySelector('#to-remove');
+				panel?.remove();
+			});
+			const rect = await visualDiff.getRect(page, `#${name}`);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+	});
```
dlockhart, line 76: Don't need the `forEach` on these since they only have one variation.
```suggestion
	it('add-element', async function() {
		await page.$eval(`#${name}`, async(element) => {
			const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
			panelGroup.innerHTML += '<d2l-collapsible-panel skeleton panel-title="blah"></d2l-collapsible-panel>';
		});
		const rect = await visualDiff.getRect(page, `#${name}`);
		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
	});

	it('remove-element', async function() {
		await page.$eval(`#${name}`, async(element) => {
			const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
			const panel = panelGroup.querySelector('#to-remove');
			panel?.remove();
		});
		const rect = await visualDiff.getRect(page, `#${name}`);
		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
	});
```

components/skeleton/test/skeleton-group.visual-diff.html
```diff
@@ -0,0 +1,90 @@
...
+				width: 300px;
+			}
+		</style>
+		<title>d2l-skeleton-group</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta charset="UTF-8">
+	</head>
+	<body class="d2l-typography">
+		<div class="visual-diff" id="all-skeleton">
+			<d2l-skeleton-group-test-wrapper>
```
dlockhart, line 26: Do you know what the plan is in terms of initially adopting skeleton groups in our existing components? Like do we plan to bake it into collapsible-panel, collapsible-panel-group, list, etc? Or yes, but in a separate PR?

components/skeleton/test/skeleton-group.visual-diff.js
jameswklassen: Yeah, I considered that...it is kind of nice to have the `name` variable rather than copy-pasting a string. I guess it's fine to have some duplication there. 🤷

components/skeleton/test/skeleton-group.visual-diff.js
jameswklassen: Is there benefit to awaiting the single property rather than the whole component? I'm used to doing stuff like this

```js
await panel.updateComplete
```

dlockhart: Sorry my bad, that's what I meant -- there's no awaiting properties. :)

components/skeleton/test/skeleton-group.visual-diff.html
jameswklassen: As discussed in standup, I don't think there are any _immediate_ plans to implement this into core components. And I think any developments on that front would need design considerations.

components/skeleton/test/skeleton-group.visual-diff.js
```diff
@@ -0,0 +1,71 @@
...
+			}, info.skeleton);
+			const rect = await visualDiff.getRect(page, `#${info.name}`);
+			await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+		});
+	});
+
+	it('add-element', async function() {
+		await page.$eval('#add-element', async(element) => {
+			const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
+			panelGroup.innerHTML += '<d2l-collapsible-panel skeleton panel-title="blah"></d2l-collapsible-panel>';
```
bearfriend, line 55: Effectively the same thing but feels like bad
```suggestion
			panelGroup.insertAdjacentHTML('beforeend', '<d2l-collapsible-panel skeleton panel-title="blah"></d2l-collapsible-panel>');
```

components/skeleton/test/skeleton-group.visual-diff.js
```diff
@@ -0,0 +1,71 @@
...
+		});
+		const rect = await visualDiff.getRect(page, '#add-element');
+		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+	});
+
+	it('remove-element', async function() {
+		await page.$eval('#remove-element', async(element) => {
+			const panelGroup = element.querySelector('d2l-skeleton-group-test-wrapper').querySelector('d2l-collapsible-panel-group');
+			const panel = panelGroup.querySelector('#to-remove');
+			panel?.remove();
```
bearfriend, line 65: I did a quick search but didn't turn anything up. Is chaining better/worse than doing one selector like this?
```suggestion
		await page.$eval('#remove-element', el => {
			el.querySelector('d2l-skeleton-group-test-wrapper d2l-collapsible-panel-group #to-remove')?.remove();
```

bearfriend (commented):
Basically a bunch of nits that shouldn't hold back a merge.

components/skeleton/test/skeleton-group.visual-diff.js
jameswklassen: I actually had this.... 😆 then removed it....

components/skeleton/test/skeleton-group.visual-diff.js
jameswklassen: 🤷 I have no idea...for debugging I think it's easier if a particular line fails, but I don't feel strongly enough about it one way or another...

components/skeleton/demo/skeleton-group-test.js
```diff
@@ -0,0 +1,95 @@
...
+
+	_addItem() {
+		this._items.push((this._items.length + 1));
+		this.requestUpdate();
+	}
+
+	_loadItems() {
+		const ids = this._items;
+		ids.forEach(id => this.shadowRoot.getElementById(id).skeleton = true);
+		ids.forEach(id =>  setTimeout(() => { this.shadowRoot.getElementById(id).skeleton = false; }, Math.random() * 2000));
```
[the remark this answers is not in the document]
jameswklassen: Ah, yup good call. The old implementation was a holdover, since I had the skeleton "resetting" happening in its own function.

components/skeleton/demo/skeleton-mixin.html
```diff
@@ -5,61 +5,49 @@
...
+			<d2l-demo-snippet>
+				<d2l-list>
+					<d2l-list-item>
+						<d2l-list-item-contents>
+							<d2l-collapsible-panel skeleton panel-title="Title 1" heading-level="3" heading-style="3" padding-type="default" type="default"></d2l-collapsible-panel>
+							<d2l-collapsible-panel panel-title="Title 1" heading-level="3" heading-style="3" padding-type="default" type="default"></d2l-collapsible-panel>
+							<d2l-input-text label="Label"></d2l-input-text>
+							<d2l-input-checkbox>Item</d2l-input-checkbox>
+							<d2l-input-checkbox>Item</d2l-input-checkbox>
+						</d2l-list-item-contents>
```
svanherk, line 40:
```suggestion
						<d2l-list-item-content>
							<d2l-collapsible-panel skeleton panel-title="Title 1" heading-level="3" heading-style="3" padding-type="default" type="default"></d2l-collapsible-panel>
							<d2l-collapsible-panel panel-title="Title 1" heading-level="3" heading-style="3" padding-type="default" type="default"></d2l-collapsible-panel>
							<d2l-input-text label="Label"></d2l-input-text>
							<d2l-input-checkbox>Item</d2l-input-checkbox>
							<d2l-input-checkbox>Item</d2l-input-checkbox>
						</d2l-list-item-content>
```

components/skeleton/skeleton-mixin.js
```diff
@@ -167,7 +169,46 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+
+	get skeleton() {
+		return this._skeletonActive;
+	}
+
+	set skeleton(val) {
+		const oldVal = this._skeleton;
+		if (oldVal === val) return;
+		this._skeleton = val;
+		this._skeletonActive = val;
```
jameswklassen, line 190: this seems wrong, but after we set this, we notify the parent that something changed, and they'll recalculate this value for us.

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,51 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+import { SkeletonMixin } from './skeleton-mixin.js';
+import { SubscriberRegistryController } from '../../controllers/subscriber/subscriberControllers.js';
+
+export const SkeletonGroupMixin = dedupeMixin(superclass => class extends SkeletonMixin(superclass) {
+
+	static get properties() {
+		// TODO: remove reflect
```
dlockhart, line 8: Reminder about this TODO. This can hopefully be reactive state properties instead?

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+	onSubscriberChange() {
+		this._skeletonSubscribers.updateSubscribers();
+	}
+
+	_checkSubscribersSkeletonState(subscribers) {
+		this._anySubscribersInSkeleton = [...subscribers.values()].some(subscriber => (
+			subscriber._skeleton || subscriber._anySubscribersInSkeleton
+		));
+
+		this._skeletonActive = this._skeleton || this._anySubscribersInSkeleton || this._setByParent;
```
dlockhart, line 41: It's a _little_ leaky for the group to set something that's "internal" to `SkeletonMixin`. Could it just call `setSkeletonActive()` instead?

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,51 @@
...
+
+		subscribers.forEach(subscriber => {
+			subscriber.setSkeletonActive(this._skeletonActive);
+			subscriber._setByParent = this._skeletonActive && !subscriber._skeleton;
```
dlockhart, line 45: This also feels a little weird since `_setByParent` is a property of `SkeletonGroup` but all the subscribers are `Skeleton`s? I'd suggest creating a `setSetByParent()` and then checking to see if it exists before calling it.

components/skeleton/skeleton-group-mixin.js
jameswklassen: yup, I just did this for debugging so I could see properties in the dom

components/skeleton/skeleton-group-mixin.js
jameswklassen: Addressed in f1707173d0a0b5e5d468cd60010f5ab67e3b09c7

components/skeleton/skeleton-group-mixin.js
jameswklassen: Addressed in c7e2d4993f7713322058d199c09e2d0ab0d04d21

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,48 @@
...
+
+		this.setSkeletonActive(this._skeleton || this._anySubscribersInSkeleton || this._setByParent);
+
+		subscribers.forEach(subscriber => {
+			subscriber.setSkeletonActive(this._skeletonActive);
+			subscriber.setSetByParent(this._skeletonActive && !subscriber._skeleton);
+		});
+
+		this._parentSkeleton?.registry?.onSubscriberChange();
+	}
```
svanherk, line 46: Not sure I'm suggesting actually changing all the names, but is this accurate?
```suggestion
	_checkSubscribersSkeletonState(subscribers) {
		this._anySubscribersWithSkeletonActive = [...subscribers.values()].some(subscriber => (
			subscriber._skeletonSetExplicitly || subscriber._anySubscribersWithSkeletonActive
		));

		this.setSkeletonActive(this._skeletonSetExplicitly || this._anySubscribersWithSkeletonActive || this._skeletonSetByParent);

		subscribers.forEach(subscriber => {
			subscriber.setSkeletonActive(this._skeletonActive);
			subscriber.setSkeletonSetByParent(this._skeletonActive && !subscriber._skeletonSetExplicitly);
		});

		this._parentSkeleton?.registry?.onSubscriberChange();
	}
```

`_skeletonSetExplicitly` is probably overkill, but I always forget what `_skeleton` is. I do think updating `_setByParent` to include skeleton improves the readability a bit, even if it's skeleton overload

components/skeleton/skeleton-mixin.js
svanherk: I think a comment would be good here.  Does setting it to `undefined` do the same thing, forcing an update or does that break something?

components/skeleton/demo/skeleton-group-nested-test.js
```diff
@@ -0,0 +1,76 @@
...
+			<d2l-collapsible-panel-group>
+				<d2l-collapsible-panel panel-title="Title"></d2l-collapsible-panel>
+				<d2l-collapsible-panel panel-title="Title"></d2l-collapsible-panel>
+			</d2l-collapsible-panel-group>
+		`;
+	}
+
+	_loadGroup() {
+		this.skeleton = !this.skeleton;
+	}
```
svanherk, line 65: I think I may have found one broken edge case from clicking around wildly. 🙈 Turn on the "`d2l-list` skeleton", then turn on the "group skeleton", then turn off the "`d2l-list` skeleton".  It un-skeletonizes when it shouldn't, because the "group skeleton" switch saw it's `skeleton` was already on and told it to turn itself off when it was toggled.

I actually think this is just a demo problem though, because the switch tries to mess with the value of `skeleton` directly while `SkeletonGroupMixin` is trying to mess with it.  I think this might work?
```suggestion
	_loadGroup() {
		this._skeletonGroup = !this._skeletonGroup;
		this.skeleton = this._skeletonGroup;
	}
```
And add `_skeletonGroup` as another property above, just like `_skeletonList` and `_skeletonInput`

components/skeleton/demo/skeleton-group-wrapper.js
```diff
@@ -0,0 +1,12 @@
...
+import '../../collapsible-panel/collapsible-panel-group.js';
+import { html, LitElement } from 'lit';
+import { SkeletonGroupMixin } from '../skeleton-group-mixin.js';
+
+class SkeletonGroupTestWrapper extends SkeletonGroupMixin(LitElement) {
+	render() {
+		return html`<slot></slot>`;
+	}
+}
+customElements.define('d2l-skeleton-group-test-wrapper', SkeletonGroupTestWrapper);
```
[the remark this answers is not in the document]
jameswklassen: Yeah, I'm not too familiar with the test repo yet, so I'm not entirely sure where it'll go but happy to move it wherever.

jameswklassen:
> I think a comment would be good here. Does setting it to undefined do the same thing, forcing an update or does that break something?

@svanherk yeah a comment probably makes sense here. Setting `undefined` does get us to an _eventual_ correct state, but does cause a brief visual flicker in some scenarios. For that reason I'd like to keep it as-is.

components/skeleton/demo/skeleton-group-wrapper.js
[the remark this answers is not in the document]
dlockhart: I think Stacey meant to say "directory" instead of "repo". Like this file is just used in visual-diffs, but it's located in the "demo" directory and likely should be in the "test" directory with the visual-diffs. You could even just move it to be an inline `<script type="module">` in the visual-diff HTML files if you like.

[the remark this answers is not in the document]
jameswklassen: ohhh, I see! Yeah might go the inline approach then -- I didn't think that was a possibility 🤯.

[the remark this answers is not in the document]
svanherk: Yup sorry about the confusion, definitely meant to say test _folder_, not repo! 🤦 Got testing repos on the brain

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,48 @@
...
+		super();
+		this._anySubscribersWithSkeletonActive  = false;
+		this._skeletonSubscribers = new SubscriberRegistryController(this, 'skeleton', {
+			onSubscribe: this.onSubscriberChange.bind(this),
+			onUnsubscribe: this.onSubscriberChange.bind(this),
+			updateSubscribers: this._checkSubscribersSkeletonState.bind(this),
+		});
+	}
+
+	updated(changedProperties) {
```
dlockhart, line 23:
```suggestion
	updated(changedProperties) {
		super.updated(changedProperties);
```

components/skeleton/skeleton-mixin.js
```diff
@@ -152,10 +153,12 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
 		return {
 			/**
-			 * Renders the input as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton)
+			 * Render the component as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton).
 			 * @type {boolean}
 			 */
-			skeleton: { reflect: true, type: Boolean  }
+			skeleton: { reflect: true, type: Boolean  },
+			_skeletonSetByParent: { state: true },
+			_skeletonSetExplicitly: { reflect: true },
```
dlockhart, line 161: I don't think either of these need to be reactive state properties and can just be private class properties instead. Reactive state is only useful if you'd like changing their values to trigger a re-render, but it looks like they're just used to calculate the skeleton value.

`SkeletonTestNestedGroup` _does_ use `_skeletonSetExplicitly` during its render to pass the value to its switch... but hopefully it still works.

components/skeleton/test/skeleton-group.visual-diff.html
```diff
@@ -0,0 +1,228 @@
...
+				<d2l-collapsible-panel-group>
+					<d2l-collapsible-panel skeleton panel-title="Title 1"></d2l-collapsible-panel>
+					<d2l-collapsible-panel skeleton panel-title="Title 2"></d2l-collapsible-panel>
+					<d2l-collapsible-panel skeleton panel-title="Title 3"></d2l-collapsible-panel>
+				</d2l-collapsible-panel-group>
```
dlockhart, line 44: Thoughts about testing the "skeleton group" portion of collapsible panel, list, table, etc. within the existing visual-diff tests for those components? It's essentially just another state they can be in.

And then most of these tests could be removed and could test skeleton-group-specific things like the adding/removing stuff, and they wouldn't necessarily need to reference other core components at all to do so.

My main concern is that we'll change collapsible panel in a minor way and these tests will "fail" unnecessarily.

jameswklassen: Hmm yeah, I could see lots of these living with collapsible panel visual diffs. Not exactly sure where that line would be -- a lot of these tests are specific to what happens in specific "skeleton group" situations. (1 skeleton element, removing the last skeleton element, etc). Would those all go with collapsible panel?

dlockhart: Hmm yeah I'm flip-flopping a bit on it to be honest. I don't think skeleton groups needs tests for _every_ component that we make a group, in the same way it doesn't have tests for all (or any) of the components that can be skeletons. 

But also list/collapsible-panel-group shouldn't really need tests for skeleton group either since it's not really something they can "mess up" -- they just extend the mixin and they're done.

So yeah, I guess maybe leave this as-is for now? I think ideally skeleton group would have a few "test components" for testing all the things it needs to do and wouldn't be relying on any other core components... that way if those core components change, the skeleton group tests don't fail. But I also don't think it's worth it spending time re-writing these in that way!

components/skeleton/skeleton-mixin.js
jameswklassen: Ah, good point. Yeah Will play around with this.

components/skeleton/skeleton-mixin.js
```diff
@@ -7,7 +7,7 @@ import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 // DE50056: starting in Safari 16, the pulsing animation causes FACE
 // (and possibly elsewhere) to render a blank page
 const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
-const animation = isSafari ? css`none` : css`loadingPulse 1.8s linear infinite`;
+const animation = isSafari ? css`` : css``;
```
dlockhart, line 10: Did you mean to remove this?

jameswklassen: nope...I removed it since it breaks the syntax highlighting.

🤦‍♂️

dlockhart (approved):
Cool! This seems to be in a state where it could be released and folks could start experimenting with it!

components/skeleton/skeleton-group-mixin.js
```diff
@@ -0,0 +1,49 @@
...
+	onSubscriberChange() {
+		this._skeletonSubscribers.updateSubscribers();
+	}
+
+	_checkSubscribersSkeletonState(subscribers) {
+		this._anySubscribersWithSkeletonActive  = [...subscribers.values()].some(subscriber => (
+			subscriber._skeletonSetExplicitly  || subscriber._anySubscribersWithSkeletonActive
+		));
+
+		this.setSkeletonActive(this._skeletonSetExplicitly  || this._anySubscribersWithSkeletonActive  || this._skeletonSetByParent);
```
svanherk, line 39:
```suggestion
		this._anySubscribersWithSkeletonActive = [...subscribers.values()].some(subscriber => (
			subscriber._skeletonSetExplicitly || subscriber._anySubscribersWithSkeletonActive
		));

		this.setSkeletonActive(this._skeletonSetExplicitly || this._anySubscribersWithSkeletonActive || this._skeletonSetByParent);
```

components/skeleton/skeleton-mixin.js
```diff
@@ -152,10 +153,10 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 	static get properties() {
 		return {
 			/**
-			 * Renders the input as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton)
+			 * Render the component as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton).
 			 * @type {boolean}
 			 */
-			skeleton: { reflect: true, type: Boolean  }
+			skeleton: { reflect: true, type: Boolean  },
```
svanherk, line 159:
```suggestion
			skeleton: { reflect: true, type: Boolean },
```

svanherk (approved):
Yay! 🎉 🎉 🎉

components/skeleton/test/skeleton-group.visual-diff.html
```diff
@@ -0,0 +1,172 @@
...
+			}
+			d2l-skeleton-group-test-wrapper {
+				display: flex;
+				flex-direction: column;
+				gap: 0.3rem;
+			}
+			d2l-input-text,
+			d2l-list {
+				margin-bottom: 0.6rem;
+			}
```
svanherk, line 24: Oh also, I assume this no longer does anything and is just leftover

PR #3813, opened 2023-07-07
Sort selectables by DOM order

bearfriend:
[DE54052](https://rally1.rallydev.com/#/?detail=/defect/707093503341&fdp=true): selection-mixin does not order selectables

Selectables order is currently not guaranteed, which means using arrow keys to move selection between radio buttons can jump around randomly.

This successfully sorts the selectables based on DOM order, but it does so on every `keyup` event. It likely can be done much more infrequently and the sorted array stored.

### DIFF
--- a/components/selection/selection-mixin.js
+++ b/components/selection/selection-mixin.js
@@ -57,7 +57,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		super();
 		this.selectionSingle = false;
 		this._selectAllPages = false;
-		this._selectionObservers = new Map();
+		this._selectionObservers = new Set();
 		this._selectionSelectables = new Map();
 	}
 
@@ -91,7 +91,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		if (this._selectAllPages) {
 			state = SelectionInfo.states.allPages;
 		} else {
-			this._selectionSelectables.forEach(selectable => {
+			this._selectionSelectables.forEach((path, selectable) => {
 				if (selectable.selected) keys.push(selectable.key);
 				if (selectable._indeterminate) state = SelectionInfo.states.some;
 			});
@@ -110,7 +110,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 
 		this._selectAllPages = (selected && selectAllPages);
 
-		this._selectionSelectables.forEach(selectable => {
+		this._selectionSelectables.forEach((path, selectable) => {
 			if (!selectable.disabled && !!selectable.selected !== selected) {
 				selectable.selected = selected;
 			}
@@ -120,7 +120,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 
 	subscribeObserver(target) {
 		if (this._selectionObservers.has(target)) return;
-		this._selectionObservers.set(target, target);
+		this._selectionObservers.add(target);
 		this._updateSelectionObservers();
 	}
 
@@ -134,7 +134,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 	}
 
 	_focusSelectAll() {
-		for (const observer of this._selectionObservers.values()) {
+		for (const observer of this._selectionObservers) {
 			if (observer.tagName === 'D2L-SELECTION-SELECT-ALL') {
 				observer.focus();
 				break;
@@ -157,7 +157,12 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		if (e.keyCode < keyCodes.LEFT || e.keyCode > keyCodes.DOWN) return;
 
 		const selectables = Array.from(this._selectionSelectables.values())
-			.filter(item => !item.disabled);
+			.filter(item => !item.disabled)
+			.sort((a, b) => {
+				const idx = a.findIndex((el, idx) => el !== b[idx]);
+				return a[idx].compareDocumentPosition(b[idx]) & Node.DOCUMENT_POSITION_FOLLOWING ? -1 : 1;
+			})
+			.map(p => p.slice(-1)[0]);
 		let currentIndex = selectables.findIndex(selectable => selectable.selected);
 		if (currentIndex === -1) currentIndex = 0;
 		let newIndex;
@@ -181,7 +186,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		if (!e.detail.selected) this._selectAllPages = false;
 		if (this.selectionSingle && e.detail.selected) {
 			const target = e.composedPath().find(elem => elem.tagName === 'D2L-SELECTION-INPUT');
-			this._selectionSelectables.forEach(selectable => {
+			this._selectionSelectables.forEach((path, selectable) => {
 				if (selectable.selected && selectable !== target) selectable.selected = false;
 			});
 		}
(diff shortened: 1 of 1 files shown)


components/selection/selection-mixin.js
```diff
@@ -57,7 +57,7 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		super();
 		this.selectionSingle = false;
 		this._selectAllPages = false;
-		this._selectionObservers = new Map();
+		this._selectionObservers = new Set();
```
bearfriend, line 60: Changed this to a set because there was no need for a key and sets are faster.

components/selection/selection-mixin.js
```diff
@@ -157,7 +157,12 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 		if (e.keyCode < keyCodes.LEFT || e.keyCode > keyCodes.DOWN) return;
 
 		const selectables = Array.from(this._selectionSelectables.values())
-			.filter(item => !item.disabled);
+			.filter(item => !item.disabled)
+			.sort((a, b) => {
+				const idx = a.findIndex((el, idx) => el !== b[idx]);
+				return a[idx].compareDocumentPosition(b[idx]) & Node.DOCUMENT_POSITION_FOLLOWING ? -1 : 1;
+			})
+			.map(p => p.slice(-1)[0]);
```
bearfriend, line 165: Given two reversed selectable composed paths, `[a,d,e]` and `[a,b,c]`, compare the DOM order of the first selectable ancestors that do not match, so `d.compareDocumentPosition(b)` results in the properly sorted selectables `[c, e]`.

components/selection/selection-mixin.js
bearfriend: Actually, this inadvertently fixes another bug. where clicking `d2l-selection-select-all-pages` in `d2l-selection-controls` is fails to focus the `d2l-selection-select-all`.

components/selection/selection-mixin.js
```diff
@@ -223,7 +226,24 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
...
+	_updateSortedSelectables() {
+
+		if (this._updateSortedSelectablesRequested) return;
+
+		this._updateSortedSelectablesRequested = true;
+		setTimeout(() => {
+			this._sortedSelectables = Array.from(this._selectionSelectables.values())
+				.sort((a, b) => {
+					const idx = a.findIndex((el, idx) => el !== b[idx]);
+					return a[idx].compareDocumentPosition(b[idx]) & Node.DOCUMENT_POSITION_FOLLOWING ? -1 : 1;
```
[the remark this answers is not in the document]
bearfriend: Yep, it just fails (or, more accurately, returns inconsistent results) across shadowDOM boundaries so you can only compare things in the same root.

components/selection/selection-mixin.js
```diff
@@ -223,7 +226,24 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 			const info = this.getSelectionInfo(true);
 			this._selectionObservers.forEach(observer => observer.selectionInfo = info);
 			this._updateObserversRequested = false;
-		}, 0);
+		});
+	}
+
+	_updateSortedSelectables() {
```
dbatiste, line 232: This list could have a lot of items and would be called for each selection input that is subscribed or unsubscribed. The potential perf issue (mainly during initial load or load more) is partly mitigated by debouncing using the `setTimeout` here. If items are drag & dropped, the old item is removed from the DOM/unsubscribed, and a new item is created /subscribed so I _think_ DOM order mutations are covered by this. Having said that, this runs whether it's needed or not. Pondering options... 

We could wait to initialize this until the first time `_handleRadioKeyUp()` is invoked, the downside being that there could be a noticeable delay, which wouldn't be great. Or, we could do the sort upon subscription in a [requestIdleCallback](https://github.com/BrightspaceUI/core/blob/main/helpers/requestIdleCallback.js)(Safari still doesn't support it, hence our helper) if we wanted to be more zealous about deferring the work. I think it might be hard gauge the usefulness of that, because we're already deferring using `setTimeout`, and it's highly dependent on other work that's queued. I'm not certain the cost of this is impactful, but from previous experiments I did find that perf of large lists already degrades quite a bit (ex. quizzing 1000+ items with deep nesting), so being cautious here.

I was originally thinking we could avoid any upfront cost by creating `getNextElement(node, predicate, options)` and `getPreviousElement(node, predicate, options)` in our [dom.js](https://github.com/BrightspaceUI/core/blob/main/helpers/dom.js) helpers similar to [getNextFocusable](https://github.com/BrightspaceUI/core/blob/main/helpers/focus.js#L93) and [getPreviousFocusable](https://github.com/BrightspaceUI/core/blob/main/helpers/focus.js#L60) where the provided predicate would check for `d2l-selection-input`, that it belongs to the selection mixin's set, and is not `hidden` or `disabled`. Those focus helpers could perhaps be refactored to use them if appropriate, by moving the focusable check into a predicate. It would involve some "minimal" DOM walking, but only when a keyboard user is changing the selection. The DOM walking would be directly proportional to the amount of DOM the consumer is stuffing into the list-item, because it would typically be walking past those elements until it walks to the sibling list-item, and then into it to eventually land on the adjacent `d2l-selection-input`. The logic to do this pretty much already exists - I think we'd just need to make it more generic. 

I think it's jus a matter of what we prefer, and whether there are any gotcha's that we haven't thought of with these approaches.

components/selection/selection-mixin.js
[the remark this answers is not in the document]
dbatiste: Does that shadowDOM limitation not cause us issues in list, since the `d2l-selection-input` elements are baked into separate DOM sub-trees of each list-item?

components/selection/selection-mixin.js
svanherk: Ok, took way longer than I wanted it to (a lot has changed in almost two years - go figure) but here's an updated POC: https://github.com/BrightspaceUI/core/pull/3819

A lot of registries won't care what order their subscribers are in, so if there's even a slight perf cost I think it makes sense to keep it outside the subscriber controller code.  But just wanted to show what that would look like - it's currently an array of elements, there's no path info stored with it.

components/selection/selection-mixin.js
svanherk: > I was originally thinking we could avoid any upfront cost by creating getNextElement(node, predicate, options) and getPreviousElement(node, predicate, options) in our [dom.js](https://github.com/BrightspaceUI/core/blob/main/helpers/dom.js) helpers...

One benefit of this would be then it definitely doesn't care about the subscriber stuff

dbatiste: Another option might be to just call `getNextFocusable()`/`getPreviousFocusable()`. It could get a focusable that we aren't interested in, such as a secondary action or some other focusable element in the list-item content, so we could just call it in a loop until we locate a `d2l-selection-input` that meets our criteria. No refactoring of those helpers required. The generic helpers I mentioned above might be valuable in the future though.

components/selection/selection-mixin.js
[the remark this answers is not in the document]
bearfriend: Using `compareDocumentPosition` on the two selectables directly does cause issues, but this is comparing their common-root ancestors instead, using the composedPath we already have from the the subscribe event handler.

components/selection/selection-mixin.js
bearfriend: If I had to guess I'd think calling `compareDocumentPosition` on a series of direct siblings would be pretty fast to do up front, and less risky than doing almost any amount of tree walking in real-time on every key event. But maybe not. Shouldn't be too hard to test both out on a giant list.

It can definitely be limited to only the radio inputs, and maybe delayed to the first focus?

dbatiste: With the helpers, the cost (both processing and memory, not that were talking _that_ big of a list) is the same regardless of the size of the list, whereas if we're sorting, the cost increases as the list gets bigger. I _think_ the sorting solution covers the mutation cases, but maybe we're missing something - waiting till they hit the arrow keys feels safer because it treats the DOM as the single source of truth and avoids mutation issues altogether, albeit at the risk of doing the tree-walking incorrectly. Having said that, those helpers are used a lot so I don't think there's a tonne of risk there (famous last words!). If it turns out that using the helpers is janky, I might change my preference though.

dbatiste:
Closing since the other PR referenced addresses this.

PR #3815, opened 2023-07-10
Move list tests to proper browser interactions

bearfriend:
Now that #3808 and #3798 have been merged, opening this to move list tests to proper browser interactions, which is what caught the Safari bug to begin with.

### DIFF
--- a/components/list/test/list.test.js
+++ b/components/list/test/list.test.js
@@ -3,28 +3,14 @@ import '../list-controls.js';
 import '../list-item.js';
 import '../list-item-button.js';
 import '../list-item-content.js';
-import { expect, fixture, focusElem, html, oneEvent, sendKeysElem, waitUntil } from '@brightspace-ui/testing';
+import { clickElem, expect, fixture, html, oneEvent, sendKeysElem } from '@brightspace-ui/testing';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
-const selectionInputRendering = async item => {
-	return new Promise(resolve => {
-		const intervalId = setInterval(() => {
-			const selectionInput = item.shadowRoot.querySelector('d2l-selection-input');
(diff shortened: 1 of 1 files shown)


components/list/test/list.test.js
```diff
@@ -92,14 +78,10 @@ describe('d2l-list', () => {
 				`);
 
 				const listItem = elem.querySelector('[key="L2-2"]');
-				const listItemLayout = elem.querySelector('[key="L2-2"]').shadowRoot.querySelector('d2l-list-item-generic-layout');
-				await focusElem(listItem);
-				await waitUntil(() => listItem.hasAttribute('_focusing'), 'List item should be focused', { timeout: 3000 });
 
-				setTimeout(() => sendKeysElem(listItemLayout, 'down', testCase.keyPress));
-				await oneEvent(listItemLayout, 'keydown');
+				await sendKeysElem(listItem, 'press', testCase.keyPress);
```
bearfriend, line 82: Just realized I switched this to `press` for no particular reason, but I assume using `down` was to test that it happens specifically on `keydown`. If we use `down` with real browser interactions, do we need to explicitly let the key `up` later?

components/list/test/list.test.js
```diff
@@ -3,28 +3,14 @@ import '../list-controls.js';
...
-			const selectionInput = item.shadowRoot.querySelector('d2l-selection-input');
-			const inputCheckbox = selectionInput.shadowRoot.querySelector('d2l-input-checkbox');
-			if (inputCheckbox) {
-				clearInterval(intervalId);
-				resolve();
-			}
-		}, 10);
-	});
-};
-
```
svanherk, line 21: 🎉

svanherk (approved):
So mush awesome cleanup here!  It makes the tests much more readable when you remove all the code that's just waiting

PR #3816, opened 2023-07-10
Attempt to simplify

dlockhart:
I wanted to play around with a couple ideas to simplify things:
- It should remove the need for the `_readyToDisplay` tracker
- `_skeletonActive` doesn't need to be a state property and cause re-renders
- I've renamed `updateDisplay` / `waitForDisplay` to `updateSkeletonActive` / `skeletonWait` just to align with the new terminology

### DIFF
--- a/components/skeleton/skeleton-group-mixin.js
+++ b/components/skeleton/skeleton-group-mixin.js
@@ -6,17 +6,17 @@ export const SkeletonGroupMixin = dedupeMixin(superclass => class extends superc
 	constructor() {
 		super();
 
-		this._readyToDisplayCached = false;
+		this._skeletonActiveCached = true;
 		this._skeletonSubscribers = new SubscriberRegistryController(this, 'skeleton', {
 			updateSubscribers: this._checkSubscribersReadyToDisplay.bind(this)
 		});
 	}
 
 	_checkSubscribersReadyToDisplay(subscribers) {
-		const readyToDisplay = ![...subscribers.values()].some(subscriber => subscriber.skeleton);
-		if (readyToDisplay !== this._readyToDisplayCached) {
-			this._readyToDisplayCached = readyToDisplay;
-			subscribers.forEach(subscriber => subscriber.changeDisplay(readyToDisplay));
+		const skeletonActive = [...subscribers.values()].some(subscriber => subscriber._skeleton);
+		if (skeletonActive !== this._skeletonActiveCached) {
+			this._skeletonActiveCached = skeletonActive;
+			subscribers.forEach(subscriber => subscriber.setSkeletonActive(skeletonActive));
 		}
 	}
 });
--- a/components/skeleton/skeleton-mixin.js
+++ b/components/skeleton/skeleton-mixin.js
@@ -170,7 +170,7 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 		super();
 		this._skeleton = false;
 		this._skeletonActive = false;
-		this._waitToDisplay = false;
+		this._skeletonWait = false;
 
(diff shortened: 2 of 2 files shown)


components/skeleton/skeleton-group-mixin.js
```diff
@@ -6,17 +6,17 @@ export const SkeletonGroupMixin = dedupeMixin(superclass => class extends superc
...
 			updateSubscribers: this._checkSubscribersReadyToDisplay.bind(this)
 		});
 	}
 
 	_checkSubscribersReadyToDisplay(subscribers) {
-		const readyToDisplay = ![...subscribers.values()].some(subscriber => subscriber.skeleton);
-		if (readyToDisplay !== this._readyToDisplayCached) {
-			this._readyToDisplayCached = readyToDisplay;
-			subscribers.forEach(subscriber => subscriber.changeDisplay(readyToDisplay));
+		const skeletonActive = [...subscribers.values()].some(subscriber => subscriber._skeleton);
```
dlockhart, line 16: Check against the underlying `_skeleton` instead of the `getter` since we want to know the component's actual state.

components/skeleton/skeleton-mixin.js
```diff
@@ -156,8 +156,7 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 			 * Render the component as a [skeleton loader](https://github.com/BrightspaceUI/core/tree/main/components/skeleton).
 			 * @type {boolean}
 			 */
-			skeleton: { reflect: true, type: Boolean  },
-			_skeletonActive: { state: true },
```
dlockhart, line 160: Since we're manually triggering an update whenever we know `skeleton`'s state has changed, we don't need modifying `_skeletonActive` to also trigger an update.

components/skeleton/skeleton-mixin.js
```diff
@@ -187,42 +184,38 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
 
 	set skeleton(val) {
 		const oldVal = this._skeleton;
-		if (oldVal !== val) {
-			this._skeleton = val;
```
dlockhart, line 191: I think we can do all the "work" in the setter and not rely on updated? This is now essentially saying: if we're not in a "wait to display" mode, just mirror _skeletonActive to _skeleton. Otherwise, tell the registry that something has changed and it'll call back to us if necessary.

components/skeleton/skeleton-mixin.js
```diff
@@ -187,42 +184,38 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+			this._skeletonActive = skeletonActive;
+			this.requestUpdate('skeleton', oldVal);
+		}
 	}
 
 	_onSubscribe() {
-		this._waitToDisplay = true;
-		this._readyToDisplay = false;
-		this._updateDisplayState();
+		this._skeletonWait = true;
```
dlockhart, line 206: When we subscribe, `_skeletonActive` becomes the source of truth.

components/skeleton/skeleton-mixin.js
```diff
@@ -187,42 +184,38 @@ export const SkeletonMixin = dedupeMixin(superclass => class extends RtlMixin(su
...
+		if (this._skeleton !== this._skeletonActive) {
+			this._skeletonActive = true;
+			this.requestUpdate('skeleton', this._skeleton);
+		}
 	}
 
 	_onUnsubscribe() {
```
dlockhart, line 213: When we unsubscribe, we revert back to `_skeleton` being the source of truth.

jameswklassen (approved):
I like these changes a lot! Much simpler, now that we're not tracking using the same variable.

PR #3822, opened 2023-07-12
DE53956 Replace optional chaining syntax for compatibility with Polymer

mostlyfabulous:
[DE53956](https://rally1.rallydev.com/#/?detail=/defect/706049515503&fdp=true): [WCAG2.2] Activity Feed > Several buttons with enclosed focus not meeting Focus appearance requirements

We haven't managed to move away from Polymer yet: https://github.com/Brightspace/assignments-ui/pull/323#issuecomment-1631896201

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -479,7 +479,7 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
 					tooltip.updatePosition();
 				}
 			} else if (prop === 'type') {
-				const input = this.shadowRoot?.querySelector('.d2l-input');
+				const input = this.shadowRoot && this.shadowRoot.querySelector('.d2l-input');
 				setTimeout(() => {
 					if (input && this.value !== input.value) {
 						this._setValue(input.value, false);


svanherk:
We've definitely started using optional chaining in a number of places across `core`, this might not be the only spot that's an issue. It's actually not a Polymer problem, just a `polymer-cli` problem because it's so old and not updated/maintained. Is it _only_ `polymer lint` that's failing for you?  Or does the [`polymer test` command](https://github.com/Brightspace/assignments-ui/blob/master/package.json#L42C12-L42C12) fail too?

dlockhart:
Yeah recently when we've been running into this, we've started switching repos off `polymer lint` entirely and converting `polymer test` uses over to the more modern `@web/test-runner`. You're going to run into a lot more than core using modern JS these days.

mostlyfabulous:
It fails on `polymer lint`. I can't run `polymer test` locally anymore 😅- getting errors now when I try.
We've got the De-FRA on the backlog, but I'm not sure when we'll fully be able to start on it.

mostlyfabulous:
Would you recommend we just remove `polymer lint` for now as optional chaining will work otherwise?

dlockhart:
Yep, just remove it.

PR #3830, opened 2023-07-14
Update SelectionMixin to focus radios based on DOM order instead of subscription order.

dbatiste:
[DE54118](https://rally1.rallydev.com/#/?detail=/defect/709376015737&fdp=true)

This PR fixes an issue where `SelectionMixin` manages the focus order of `d2l-selection-input` (radios) incorrectly. The current logic relies on the `Map` of selectables to determine the next/previous item to focus, and it incorrectly assumes that the order within the map (subscription order) is the DOM order.

This fix changes the logic to find the next/previous `d2l-selection-input` (radio) based tree-walking when the key is pressed. An alternate solution that sorts the subscribed selectables based on DOM order is: https://github.com/BrightspaceUI/core/pull/3813

### DIFF
--- a/components/selection/selection-mixin.js
+++ b/components/selection/selection-mixin.js
@@ -1,4 +1,6 @@
+import { getNextFocusable, getPreviousFocusable } from '../../helpers/focus.js';
 import { CollectionMixin } from '../../mixins/collection/collection-mixin.js';
+import { isComposedAncestor } from '../../helpers/dom.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
 const keyCodes = {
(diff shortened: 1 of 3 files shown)


components/selection/selection-mixin.js
```diff
@@ -152,29 +154,41 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
...
-			|| (this.dir === 'rtl' && e.keyCode === keyCodes.LEFT)
-			|| e.keyCode === keyCodes.DOWN) {
-			if (currentIndex === selectables.length - 1) newIndex = 0;
-			else newIndex = currentIndex + 1;
-		} else if ((this.dir !== 'rtl' && e.keyCode === keyCodes.LEFT)
-			|| (this.dir === 'rtl' && e.keyCode === keyCodes.RIGHT)
-			|| e.keyCode === keyCodes.UP) {
-			if (currentIndex === 0) newIndex = selectables.length - 1;
-			else newIndex = currentIndex - 1;
+		const getSelectionInput = (focusable, forward) => {
```
dbatiste, line 162: The next focusable must be:
1. A `d2l-selection-input` (radio)
2. Enabled
3. Visible
4. Subscribed with the SelectionMixin (don't move focus to a nested list)

If we don't find one within the selection component's sub-tree, then we need to cycle focus first/last

bearfriend:
As long as this performs well when there are lots of focusables in between then this looks great.

dbatiste:
> As long as this performs well when there are lots of focusables in between then this looks great.

Ok, I've mocked up some tests that include items with large menu structures in their content to do some performance testing. I found that overall processing takes between 16-30ms in both cases, and that the determining the next/previous focusable on the fly (this solution) costs an additional 1-3ms. Almost all the time is spent with applying focus, Lit life cycle, and browser layout, all of which is done regardless. I can live with that in order to treat the DOM as the single source of truth, and avoid any upfront/one time ordering cost for all the list items.

Determining focusable based on DOM during key event:
image

Determining focusable based on cached DOM order:
image

bearfriend (approved):
Absolutely! That's awesome :shipit:

PR #3833, opened 2023-07-17
fix: update filter docs with paging

GZolla:
Include a section in the filter documentation to emphasize the expected use of paging and manual search.

Rally: [US151801](https://rally1.rallydev.com/#/?detail=/userstory/698648089311&fdp=true)

### DIFF
--- a/components/filter/README.md
+++ b/components/filter/README.md
@@ -235,9 +235,9 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
 
 ## Search and Paging
 
-Searching inside a filter dimension has two types: `automatic` and `manual`.
+Most filters will not need search or paging features since filter value lists are generally short. For longer lists of filter values when Search is necessary, it can be enabled by setting search-type to `automatic` or `manual`.
 
-`automatic` search runs a basic case-insensitive text comparison on the dimension values. 
+`automatic` search runs a basic case-insensitive text comparison on the dimension values that are loaded in the browser, having no awareness of server-side values that are not yet loaded. 
 
 `manual` search dispatches a `d2l-filter-dimension-search` event delegating the search to the component's consumer. The event's detail will contain the key of the dimension from where the event was dispatched (`key`), the text value used for the search (`value`) and a callback (`searchCompleteCallback`). This callback gives the consumer control of which keys to display, either by setting `displayAllKeys` to `true` or passing a list of the keys to display as `keysToDisplay` (all other keys will be hidden). The dimension will be in a loading state until the callback is called.
 ```js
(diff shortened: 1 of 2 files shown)


components/filter/README.md
```diff
@@ -232,6 +233,36 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
 | `selected` | Boolean, default: `false` | Whether the value in the filter is selected or not |
 <!-- docs: end hidden content -->
 
+## Search and Paging
+
+Searching inside a filter dimension has two types: `automatic` and `manual`.
+
+`automatic` search runs a basic case-insensitive text comparison on the dimension values. 
+
+`manual` search dispatches a `d2l-filter-dimension-search` delegating the search to the component's consumer. The event's detail will contain the key of the dimension from where the event was dispatched(`key`), the text value used for the search(`value`) and a callback(`searchCallback`) that gives the consumer control of which keys to display either setting `displayAllKeys` to true or passing a list of the keys to display as `keysToDisplay`, all other keys will be hidden.
```
svanherk, line 242:
```suggestion
`manual` search dispatches a `d2l-filter-dimension-search` event delegating the search to the component's consumer. The event's detail will contain the key of the dimension from where the event was dispatched (`key`), the text value used for the search (`value`) and a callback (`searchCompleteCallback`). This callback gives the consumer control of which keys to display, either by setting `displayAllKeys` to `true` or passing a list of the keys to display as `keysToDisplay` (all other keys will be hidden). The dimension will be in a loading state until the callback is called.
```

components/filter/README.md
```diff
@@ -232,6 +233,36 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
...
+```js
+e.detail.searchCompleteCallback({ keysToDisplay: keysToDisplay });
+e.detail.searchCompleteCallback({ displayAllKeys: true });
+```
+
+Paging works by setting the property `has-more` on a dimension set will display a `d2l-pager-load-more` button at the end of the values. Note however that paging requires the search type to be set to `manual`. Clicking the button replaces its text with a loading spinner dispatches a `d2l-filter-dimension-load-more` event whose detail, like the search event, contains the dimension key(`key`), active search value(`value`) and a callback(`loadMoreCompleteCallback`) that works just like `searchCallback` described above, but will also signal the `d2l-pager-load-more` component to replace its loading spinner with the initial text.
```
svanherk, line 248:
```suggestion
Paging works by setting the property `has-more` on a dimension set, which will display a `d2l-pager-load-more` button at the end of the values. Note however that paging requires the search type to be set to `manual`. Clicking the button replaces its text with a loading spinner and dispatches a `d2l-filter-dimension-load-more` event whose detail, like the search event, contains the dimension key (`key`), active search value (`value`) and a callback (`loadMoreCompleteCallback`) that works just like `searchCompleteCallback` described above. The pager will also be in a loading state until the callback is called.
```

components/filter/README.md
```diff
@@ -232,6 +233,36 @@ This component is built to be used alongside the [d2l-filter-dimension-set](#d2l
...
+```js
+e.detail.loadMoreCompleteCallback({ keysToDisplay: keysToDisplay });
+e.detail.loadMoreCompleteCallback({ displayAllKeys: true });
+```
+
+### Selection and manual search/paging
+
+The filter component depends entirely on the consumer to include the selected filter values in order for the selected counts and `d2l-filter-tags` to display the correct values. Ideally, all values should be loaded into the dimensions and the event callbacks should be leveraged to set the visibility on those values. However, in the cases where this is not possible and new values are being added/removed manually from the dimension, then selection should be persisted, meaning that selected items should always be loaded and included in the dimension and they should not be removed in order to maintain the functionality of counts and filter tags.
```
svanherk, line 256:
```suggestion
The filter component depends entirely on the consumer to include the selected filter values in order for the selected counts and `d2l-filter-tags` to display the correct values. Ideally, all values should be loaded into the dimensions and the event callbacks should be leveraged to set the visibility on those values. However, in the cases where this is not possible and new values are being added/removed manually from the dimension, then selection should be persisted. This means that selected items should always be loaded and included in the dimension and they should not be removed in order to maintain the functionality of counts and filter tags.
```

svanherk: Optionally, maybe this at the end?
```
Use `keysToDisplay` from the search and load more callbacks to hide them accordingly, based on what's been searched/loaded from the user's POV.
```

svanherk (commented):
Just a few minor things below, overall I find this very clear! But I'm probably biased 😄, we'll see what Jeff thinks!

svanherk (approved):

PR #3834, opened 2023-07-17
US155185 - Remove Enter instructions from list navigation tooltip

svanherk:
This line was confusing because you press enter for links and buttons, space for checkboxes, and either for interactive content.  Confirmed with Jeff that we just want to remove this line altogether.


dlockhart (approved):
Good thing OSLO isn't hooked up, or these would be more orphaned lang terms! 😆

PR #3842, opened 2023-07-19
US155333 - Convert overflow-ellipsis property of d2l-link to lines and d support inheriting the display value

svanherk:
**The issue:**

Since we have the `a` in our `d2l-link` as `inline`, long wrapping links look like this:

image

And applying multiline clamping looks like this:

image

**Potential Solution:**

Originally, I just had this to allow consumers to set the `a` as `block` when needed:
```
a {
	display: inherit;
}
```
This fixes the issue, but then the consumer needs to add `padding` in the multi-line clamp case to avoid the outline being cut off.  If they add too little, it's clipped, and if they add too much, you see the overflow text:
image

Rather than have them guess, or create the perfect value that breaks if we change our outline style, my idea here is to replace the single-line clamping `d2l-link` already has with multiline clamping.  I'm using the same technique as [`collapsible-panel-summary-item`](https://github.com/BrightspaceUI/core/pull/3792).

From what I can tell, `overflow-ellipsis` is not currently used outside of `core`.


components/link/link.js
```diff
@@ -74,10 +75,10 @@ class Link extends FocusMixin(LitElement) {
 			 */
 			main: { type: Boolean, reflect: true },
 			/**
-			 * Whether to truncate the link with ellipsis
-			 * @type {boolean}
+			 * The number of lines to display before truncating text with an ellipsis. The text will not be truncated unless a value is specified.
+			 * @type {number}
 			 */
-			overflowEllipsis: { type: Boolean, attribute: 'overflow-ellipsis', reflect: true },
+			lines: { type: Number },
```
dlockhart, line 81: Should default it to `0` in the constructor. (collapsible-panel-summary-item should probably be doing that too)

components/tooltip/tooltip.js
```diff
@@ -976,6 +976,7 @@ class Tooltip extends RtlMixin(LitElement) {
 		const clone = target.cloneNode(true);
 		clone.removeAttribute('id');
 		clone.style.maxWidth = 'none';
+		clone.style.display = 'inline-block';
```
svanherk, line 979: This is a bit worrying, but without this the `clone.scrollWidth` is not properly calculated. `show-truncated-only` is only used in a couple places outside `core` so I can take a look. If this is an issue, we could also only apply it if the target is a link.

svanherk: Ok, consistent-eval and outcomes are both still working fine with this.  Also reproduced the setup Nova has and it works as well.  Opened an optional PR to add that use case here, if we want: https://github.com/BrightspaceUI/core/pull/3848

components/link/test/link.visual-diff.js
```diff
@@ -20,35 +20,45 @@ describe('d2l-link', () => {
...
-	['print', 'screen'].forEach((mediaType) => {
+	[
+		{ name: 'wc-standard' },
+		{ name: 'wc-main' },
+		{ name: 'wc-small' },
+		{ name: 'wc-inline', selector: '#wc-inline d2l-link' },
+		{ name: 'wc-inline-paragraph', selector: '#wc-inline-paragraph d2l-link' },
+		{ name: 'wc-block' },
+		{ name: 'wc-clamp-one-line' },
+		{ name: 'wc-clamp-two-lines' },
```
dbatiste, line 31: Sometimes I've included an "unbreakable" case for the overflow clamping scenarios. Might be worthwhile here as well.

components/link/test/screenshots/ci/golden/link/d2l-link-screen-wc-clamp-unbreakable-focused.png
svanherk, line 1: @dbatiste Is this what you had in mind for the unbreakable test?

dbatiste: Yep, though I think the collapsible panel item tests it with multiple line clamping such as [like this](https://github.com/BrightspaceUI/core/blob/main/components/collapsible-panel/test/screenshots/ci/golden/collapsible-panel/d2l-collapsible-panel-ltr-summary-item-text-lines-unbreakable.png). Either one works though.

dbatiste (approved):

PR #3848, opened 2023-07-24
US155333 - Add demo and visual diff test for d2l-status-indicator truncating tooltip

svanherk:
I reproduced the `show-truncating-only` use case in Nova to make sure it still worked after [setting the clone to `display: inline-block;`](https://github.com/BrightspaceUI/core/pull/3842/files#diff-05b45b190bb322a49005a6c1c6929f7412edc47455639b9d8578ffb6ed40cd77R979).

If we want, we can add it - but we can also not, if we feel it's not something we really support currently. As is, `d2l-status-indicator` doesn't have a focus state and adding `tabindex="0"` uses the browser's styles.  But without that, keyboard-only users can't trigger the tooltip.


svanherk:
As it says in the PR description - not sure if we want this or not.  Thoughts?

dlockhart:
My opinion is that this tooltips + status indicators aren't currently a supported use case. I'm guessing when you do this, tooltip is going to put a warning in the console about being used in an inaccessible manner?

If this _is_ something we want to support, then the best approach would probably be to bake the tooltip into status indicators (e.g. `<d2l-status-indicator tooltip="stuff here">` which would possibly give it a more interactive role and definitely give it a `tabindex`.

svanherk:
> My opinion is that this tooltips + status indicators aren't currently a supported use case. I'm guessing when you do this, tooltip is going to put a warning in the console about being used in an inaccessible manner?

Yup it does - although to be fair, so does hooking it up to a `d2l-link` and `d2l-button`.  Every example on our demo pages causes the warning.  I wonder if we could expand the list [here ](https://github.com/BrightspaceUI/core/blob/main/components/tooltip/tooltip.js#L42) to include components we know are good.  Otherwise the warning is just so noisy I feel like everyone ignores it.

> If this is something we want to support, then the best approach would probably be to bake the tooltip into status indicators (e.g. <d2l-status-indicator tooltip="stuff here"> which would possibly give it a more interactive role and definitely give it a tabindex.

Agreed, that sounds like the better approach than supporting a focus style but having them configure when it's focusable.  Will close this and mention it in standup

svanherk:
Closing - we don't currently support this. Sent the use case to Jeff to determine if we should.

PR #3856, opened 2023-07-26
Add description to d2l-selection-input.

dbatiste:
[US153742](https://rally1.rallydev.com/#/?detail=/userstory/704221701387&fdp=true)

This PR adds the description property to `d2l-selection-input`. This is requires so that Outcomes can update their selectable list to be more usable.

### DIFF
--- a/components/selection/selection-input.js
+++ b/components/selection/selection-input.js
@@ -1,6 +1,7 @@
 import '../inputs/input-checkbox.js';
 import { css, html, LitElement } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { LabelledMixin } from '../../mixins/labelled/labelled-mixin.js';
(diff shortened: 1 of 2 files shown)


components/selection/selection-input.js
```diff
@@ -92,15 +99,20 @@ class Input extends SkeletonMixin(LabelledMixin(LitElement)) {
...
+
 			const radioClasses = {
 				'd2l-input-radio': true,
 				'd2l-selection-input-radio': true,
 				'd2l-skeletize': true,
 				'd2l-hovering': this.hovering,
 				'd2l-disabled': this.disabled
 			};
+
+			const description = this.description ? html`<div id="${this._descriptionId}" hidden>${this.description}</div>` : null;
```
dlockhart, line 111: In general I think we prefer `nothing` in these scenarios?

```suggestion
			const description = this.description ? html`<div id="${this._descriptionId}" hidden>${this.description}</div>` : nothing;
```

dlockhart (approved):

dbatiste:
Closing this since the consumer if going to try and use native `<input type="radio" aria-describedby="... ...">`. No reason for them to use `d2l-selection-input`.

PR #3861, opened 2023-07-27
fix: move pager inside d2l-list-content

GZolla:
Moving the pager inside the d2l-list-content removes a defect where the user could seemingly focus on both the `d2l-pager-load-more` component and the search button (or the pager and the back button), which means the space bar/enter would trigger both actions.
Margin was also added at the bottom the load more button, so the focus ring's bottom border is not cut off.


components/list/list.js
```diff
@@ -136,8 +137,8 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			<slot name="header"></slot>
 			<div role="${role}" aria-label="${ifDefined(ariaLabel)}" class="d2l-list-content">
 				<slot @keydown="${this._handleKeyDown}" @slotchange="${this._handleSlotChange}"></slot>
+				${this._renderPagerContainer()}
```
dbatiste, line 140: I think we previously moved non-list-item elements out of the the `role="list"` element to avoid incorrect screen reader announcement for list-item counts.

dbatiste: Yeah I think that is true. I'm not super keen on changing where the pager renders based on the role though. I feel like there should be another way to address the issue (which sounds really strange in the description).

dbatiste:
Closing this PR since we don't want to move forward with this change.

PR #3864, opened 2023-07-28
Add instructions to d2l-input-text.

dbatiste:
[DE53316](https://rally1.rallydev.com/#/?detail=/defect/700336570793&fdp=true)

This PR adds the `control-instructions` advanced property to the `d2l-input-text` component which wires the tooltip directly up to the input via `aria-describedby`. It also updates `d2l-input-date` to provide the `control-instructions` property instead of rendering a disconnected flakey tooltip for improved accessibility.

This does not remove the `description` property from `d2l-input-text` - that is a separate set of work as mentioned in the Rally ticket.

### DIFF
--- a/components/inputs/demo/input-text.html
+++ b/components/inputs/demo/input-text.html
@@ -69,6 +69,13 @@
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>Control Instructions</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-text label="Name" control-instructions="Use the keyboard to enter some sweet stuff... :)"></d2l-input-text>
+				</template>
+			</d2l-demo-snippet>
+
 			<h3>Invalid</h3>
 			<d2l-demo-snippet>
 				<template>
--- a/components/inputs/demo/input-text.html
+++ b/components/inputs/demo/input-text.html
@@ -69,10 +69,10 @@
 				</template>
 			</d2l-demo-snippet>
 
-			<h2>Control Instructions</h2>
+			<h2>Instructions</h2>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-input-text label="Name" control-instructions="Use the keyboard to enter some sweet stuff... :)"></d2l-input-text>
+					<d2l-input-text label="Name" instructions="Use the keyboard to enter some sweet stuff... :)"></d2l-input-text>
 				</template>
 			</d2l-demo-snippet>
 
--- a/components/inputs/docs/input-text.md
+++ b/components/inputs/docs/input-text.md
@@ -77,10 +77,10 @@ The `<d2l-input-text>` element is a simple wrapper around the native `<input typ
 | `aria-invalid` | String | Indicates that the input value is invalid |
 | `autocomplete` | String | Specifies which types of values [can be autofilled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) by the browser |
 | `autofocus` | Boolean | When set, will automatically place focus on the input |
-| `control-instructions` | String | Additional information relating to how to use the control |
 | `description` | String | A description to be added to the `input` for accessibility |
 | `disabled` | Boolean | Disables the input |
 | `input-width` | String, default: `100%` | Restricts the maximum width of the input box without impacting the width of the label |
+| `instructions` | String | Additional information relating to how to use the component |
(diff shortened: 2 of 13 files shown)


components/inputs/input-text.js
```diff
@@ -461,8 +466,13 @@ class InputText extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMixin(
...
 
 		let tooltip = nothing;
-		if (this.validationError && !this.skeleton && !this.noValidate) {
-			tooltip = html`<d2l-tooltip state="error" announced align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>`;
+		if (!this.skeleton) {
+			if (this.validationError && !this.noValidate) {
+				// this tooltip is using "announced" since we don't want aria-describedby wire-up which would bury the message in VoiceOver's More Content Available menu
+				tooltip = html`<d2l-tooltip state="error" announced align="start">${this.validationError} <span class="d2l-offscreen">${this.description}</span></d2l-tooltip>`;
+			} else if (this.controlInstructions) {
+				tooltip = html`<d2l-tooltip align="start" for="${this._inputId}" delay="1000">${this.controlInstructions}</d2l-tooltip>`;
```
dbatiste, line 474: Unlike the error `d2l-tooltip` above, the offscreen `description` is not included in this tooltip since our intention is that it `description` will be removed from `d2l-input-text`. I've left that `description` in place because removing `description` will be a separate set of work. 

Also, we do _not_ use `announced` here because we want direct tooltip wire-up using `aria-describedby` (we're "ok" with these control instructions being buried in VoiceOver's "More Content Available" menu.

components/inputs/demo/input-text.html
```diff
@@ -69,6 +69,13 @@ <h2>Placeholder</h2>
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>Control Instructions</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-text label="Name" control-instructions="Use the keyboard to enter some sweet stuff... :)"></d2l-input-text>
```
dlockhart, line 75: 😆

components/inputs/docs/input-text.md
```diff
@@ -77,6 +77,7 @@ The `<d2l-input-text>` element is a simple wrapper around the native `<input typ
 | `aria-invalid` | String | Indicates that the input value is invalid |
 | `autocomplete` | String | Specifies which types of values [can be autofilled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) by the browser |
 | `autofocus` | Boolean | When set, will automatically place focus on the input |
+| `control-instructions` | String | Additional information relating to how to use the control |
```
dlockhart, line 80: I'm good with the name, but we don't _typically_ refer to these as "controls" anymore, but "components". Would `instructions` or `usage-instructions` work?

dbatiste: Yep... agreed. Will tweak.

dlockhart (approved):
Nice!

components/inputs/test/input-date-range.visual-diff.js
```diff
@@ -63,11 +63,14 @@ describe('d2l-input-date-range', () => {
 
 	async function getRectInnerTooltip(page, selector, inputDateSelector) {
 		return page.$eval(selector, (elem, inputDateSelector) => {
+			let content = window.querySelectorComposed(elem.shadowRoot, 'd2l-tooltip[showing]');
```
dbatiste, line 66: If this works I will clean this up and use the same solution in the other two places that do the same thing. Unless of course there is a better approach. My goal here is to update the test so that it doesn't need to know which component is responsible for the tooltip. This method just needs to locate the tooltip being displayed so that it can calculate the correct dimensions for the screenshot.

components/inputs/test/input-helper.js
```diff
@@ -65,9 +65,9 @@ export function getRect(page, selector) {
 	});
 }
 
-export function getRectTooltip(page, selector, tooltipIndex) {
-	return page.$eval(selector, (elem, tooltipIndex) => {
-		const content = elem.shadowRoot.querySelectorAll('d2l-tooltip')[tooltipIndex ? tooltipIndex : 0];
+export function getRectTooltip(page, selector) {
```
dbatiste, line 68: This more generic version replaces the other two versions that are being removed.

PR #3870, opened 2023-08-02
fix: wait for positioning before scrolling in filter load more

GZolla:
Wait for dropdown menu to finish its positioning after loading the items before focusing on the first newly loaded item.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -743,7 +743,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				value: dimension.searchValue,
 				loadMoreCompleteCallback: (options) => {
 					applySearch(options);
-					e.detail.complete();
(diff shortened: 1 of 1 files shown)


dbatiste (approved):

components/filter/filter.js
```diff
@@ -743,7 +743,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				value: dimension.searchValue,
 				loadMoreCompleteCallback: (options) => {
 					applySearch(options);
-					e.detail.complete();
+					this.shadowRoot.querySelector('d2l-dropdown-menu').addEventListener('d2l-dropdown-position', e.detail.complete, { once: true });
```
svanherk, line 746: Should this be something like this?
```
const menu = this.shadowRoot.querySelector('d2l-dropdown-menu');
menu ? menu.addEventListener('d2l-dropdown-position', e.detail.complete, { once: true }) : e.detail.complete();
```
Since `d2l-dropdown-menu` will only exist in the multi-dimension case, but not the single case?

dbatiste: This seems like a good check to make.

PR #3876, opened 2023-08-03
fix: reflect input-color's value to formValue

dlockhart:
Jake noticed that when `<d2l-input-color>` is inside a `<d2l-form>`, its value isn't properly reflected in the value that gets submitted with the form. For inputs that don't have an explicit native `<input>`, we need to manually keep that in sync.

### DIFF
--- a/components/inputs/input-color.js
+++ b/components/inputs/input-color.js
@@ -285,6 +285,10 @@ class InputColor extends FocusMixin(FormElementMixin(LocalizeCoreElement(LitElem
 
 		if (changedProperties.has('label') || changedProperties.has('type')) this._validateLabel();
 
+		if (changedProperties.has('value') || changedProperties.has('type') || changedProperties.has('disallowNone')) {
+			this.setFormValue(this.value);
+		}
+
 	}
 
 	_getLabel() {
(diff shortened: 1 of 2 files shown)


components/inputs/input-color.js
```diff
@@ -285,6 +285,10 @@ class InputColor extends FocusMixin(FormElementMixin(LocalizeCoreElement(LitElem
 
 		if (changedProperties.has('label') || changedProperties.has('type')) this._validateLabel();
 
+		if (changedProperties.has('value') || changedProperties.has('type') || changedProperties.has('disallowNone')) {
```
dlockhart, line 288: The value is a function of these 3 properties, so anytime they change we know to update the form value.

JakeProuse (approved):

dbatiste:
> For inputs that don't have an explicit native `<input>`, we need to manually keep that in sync.

Do you mean "_forms_ that don't have explicit native `<input>`? The reason I ask is that it looks like many of the other input components call this method but _do_ have native `input` elements internally.

Besides that, makes sense to me.

dlockhart:
Yeah sorry, I wasn't very clear. What I meant was that if a web component extends `FormElementMixin`, it needs to call `setFormValue`. I think it could get away with _not_ extending `FormElementMixin` if it contained a native `<input>` and reflected its own `name` property down to that `input` -- essentially turning itself into a native input.

dbatiste (approved):

PR #3881, opened 2023-08-08
DE54475 - Add test and demos to show sticky d2l-list-controls scrolled + list hover/focus issue

svanherk:
Setting up the visual-diff list tests and dialog demo to show the issue, which will be fixed in a follow-up PR.

I was having a horrible time trying to get hover working properly for lists in dialog with the old `visual-diff`. I'm going to put up a PR moving the dialog tests to the new `vdiff`.


components/dialog/test/dialog-fullscreen.visual-diff.html
```diff
@@ -216,5 +220,66 @@
 		<div>Bottom</div>
 	</d2l-dialog-fullscreen>
 
+	<d2l-dialog-fullscreen id="dialogWithList" title-text="Dialog Title">
```
svanherk, line 223: Lots of duplication here that can go away with the new vdiff

PR #3892, opened 2023-08-14
US152268 - Hookup new vdiff tests and migrate dialog

svanherk:
This PR:
- Hooks up a new `vdiff` workflow
- Migrates the existing `dialog`, `dialog-fullscreen`, `dialog-mixin` and `dialog-with-mobile-dropdown` tests over
  - The remaining `dialog-confirm` and `dialog-ifrau` are different enough that there wasn't much benefit lumping them with these four


components/dialog/test/dialog-fullscreen.vdiff.js
```diff
@@ -0,0 +1,79 @@
...
+
+function dispatchFullscreenWithinEvent(elem, state) {
+	elem.dispatchEvent(new CustomEvent(
+		'd2l-fullscreen-within', { bubbles: true, composed: true, detail: { state: state } }
+	));
+}
+
+describe('dialog-fullscreen', () => {
+
+	['native', 'custom'].forEach((type) => {
```
svanherk, line 24: To note: We currently have `native` dialogs off because of a number of defects.  So we actually have duplicates of every vdiff test, with both `native` and `custom` using `custom` dialogs.  We can leave this running both (I guess to verify that native is properly turned off), or we could just comment out the `'native'` option until it does something 🤷

dbatiste: Yeah I was thinking about this the other day when you mentioned migrating these tests. I'd be all for commenting out the native tests. We can reenable the tests to generate goldens before we reenable the native implementation so we have useful diffs for that change.

components/dialog/test/dialog-fullscreen.vdiff.js
```diff
@@ -0,0 +1,79 @@
...
+
+			[
+				{ screen: 'wider', viewport: { width: 1400, height: 700 } },
+				{ screen: 'wide', viewport: { width: 800, height: 500 } },
+				{ screen: 'narrow', viewport: { width: 600, height: 500 } },
+				{ screen: 'landscape', viewport: { width: 600, height: 320 } }
+			].forEach(({ screen, viewport }) => {
+				describe(screen, () => {
+					[
+						{ name: 'opened', f: dialog() },
```
dbatiste, line 39: Nit: The single letter property gives me vibes from the olden days when I was doing Fortran limited to 640K. Lol. Unless it's ubiquitous, I think I would always prefer writing it out for ease of readability. Not a big deal though.

svanherk (commented):
Ok, this is ready for final review! The image diffs here were done in stages:
1. Migrate existing images from `/screenshots/ci` -> `golden`
2. Run tests against old images and create useful report to compare
3. Merge in changes

But what that ends up looking like all combined is a diff that deletes the old images and adds slightly different new ones.  The easiest way to compare the diffs will be to look at the final report in https://github.com/BrightspaceUI/core/pull/3893

components/dialog/test/dialog-fullscreen.vdiff.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+describe('dialog-fullscreen', () => {
+
+	[/*'native',*/ 'custom'].forEach((type) => {
+
+		describe(type, () => {
+			before(async() => {
+				window.D2L.DialogMixin.preferNative = type === 'native';
+			});
```
dlockhart, line 29: Nit: shouldn't need to be async

```suggestion
			before(() => window.D2L.DialogMixin.preferNative = type === 'native');
```

components/dialog/test/dialog-fullscreen.vdiff.js
```diff
@@ -0,0 +1,79 @@
...
+
+			describe('internal', () => {
+
+				[
+					{ name: 'no-footer-content', template: createDialog({ content: long }) },
+					{ name: 'no-padding', template: createDialog({ content: html`<div style="background-color: var(--d2l-color-citrine); height: 100%; width: 100%;">No padding!</div>${footer}`, noPadding: true }) },
+					{ name: 'horizontal-overflow', template: createDialog({ content: html`${tabs}${general}` }) },
+					{ name: 'scroll-bottom-shadow', template: createDialog({ content: html`${long}${footer}` }) },
+					{ name: 'scroll-top-shadow', template: createDialog({ content: html`${long}${footer}` }), action: elem => elem.querySelector('#bottom').scrollIntoView() },
```
dlockhart, line 60: Was wondering why this resulted in a diff... so previously `#dialogLong` went to "Line 12" but your `${long}` has 13 lines. Same thing happened in the `dialog.vdiff.js` tests.

dlockhart (approved):
Looks great! Just a couple notes. It's fantastic how readable the report is.

components/dialog/test/dialog-fullscreen.vdiff.js
svanherk: Yeah I should've called that out, I figured it wasn't worth adding another option in `dialog-shared-contents.js` to export or creating a one-off `template` just to have one less line

dbatiste (approved):

components/dialog/test/dialog-fullscreen.vdiff.js
dlockhart: Cool yeah, just wanted to make sure you did it consciously.

PR #3898, opened 2023-08-17
ci: Move static builds to d2l.dev bucket

eKoopmans:
This moves core's static builds to the `d2l.dev` S3 bucket, as part of standardizing the PR Preview process. This will let us delete the custom `brightspace-ui-core.d2l.dev` bucket.

This is step 2 of a multi-step process for UI core:

1. `repo-settings`: Get d2l.dev S3 access (https://github.com/Brightspace/repo-settings/pull/1250)
2. `BrightspaceUI/core`: Switch core's CI and Release to publish to d2l.dev S3
3. `d2l.dev`: Swap the `d2l.dev` routes to point to the new S3 locations
4. `repo-settings`: Remove the assumable role `brightspace-ui-core.d2l.dev-githubactions-access`
5. AWS console: Delete the `brightspace-ui-core.d2l.dev` bucket and IAM roles
    - I'll need help with this last step, I don't have access to the relevant account anymore

[US156216](https://rally1.rallydev.com/#/?detail=/userstory/719304972319&fdp=true)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,14 +66,14 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
-          role-to-assume: arn:aws:iam::037018655140:role/brightspace-ui-core.d2l.dev-githubactions-access
+          role-to-assume: arn:aws:iam::022062736489:role/r+BrightspaceUI+core+repo
           role-duration-seconds: 3600
           aws-region: ca-central-1
 
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://brightspace-ui-core.d2l.dev/branches/pr-${{ github.event.number }}
+          bucket-path: s3://d2l.dev/pr-previews/brightspace-ui-core/pr-${{ github.event.number }}
           publish-directory: ./build/
 
       - name: Notify
@@ -84,7 +84,7 @@ jobs:
 
             We've deployed an automatic preview for this PR - you can see your changes here:
 
-            | URL | https://pr-${{ github.event.number }}-brightspace-ui-core.d2l.dev/ |
+            | URL | https://pr-previews.d2l.dev/brightspace-ui-core/pr-${{ github.event.number }}/ |
             |---|---|
 
             > **Note**
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://d2l.dev/pr-previews/brightspace-ui-core/pr-${{ github.event.number }}
+          bucket-path: s3://d2l.dev/live/prs/BrightspaceUI/core/pr-${{ github.event.number }}
           publish-directory: ./build/
 
       - name: Notify
@@ -84,7 +84,7 @@ jobs:
 
             We've deployed an automatic preview for this PR - you can see your changes here:
 
-            | URL | https://pr-previews.d2l.dev/brightspace-ui-core/pr-${{ github.event.number }}/ |
+            | URL | https://live.d2l.dev/prs/BrightspaceUI/core/pr-${{ github.event.number }}/ |
             |---|---|
 
             > **Note**
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://d2l.dev/live/prs/BrightspaceUI/core/pr-${{ github.event.number }}
+          bucket-path: s3://d2l.dev/live/prs/${{ github.repository }}/pr-${{ github.event.number }}
           publish-directory: ./build/
 
       - name: Notify
@@ -84,7 +84,7 @@ jobs:
 
             We've deployed an automatic preview for this PR - you can see your changes here:
 
-            | URL | https://live.d2l.dev/prs/BrightspaceUI/core/pr-${{ github.event.number }}/ |
+            | URL | https://live.d2l.dev/prs/${{ github.repository }}/pr-${{ github.event.number }}/ |
             |---|---|
 
             > **Note**
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,5 +67,5 @@ jobs:
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://d2l.dev/live/BrightspaceUI/core
+          bucket-path: s3://d2l.dev/live/${{ github.repository }}
           publish-directory: ./build/
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,5 +67,5 @@ jobs:
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://d2l.dev/brightspace-ui-core/main
+          bucket-path: s3://d2l.dev/live/BrightspaceUI/core
           publish-directory: ./build/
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -66,14 +66,14 @@ jobs:
...
-          role-to-assume: arn:aws:iam::037018655140:role/brightspace-ui-core.d2l.dev-githubactions-access
+          role-to-assume: arn:aws:iam::022062736489:role/r+BrightspaceUI+core+repo
           role-duration-seconds: 3600
           aws-region: ca-central-1
 
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://brightspace-ui-core.d2l.dev/branches/pr-${{ github.event.number }}
+          bucket-path: s3://d2l.dev/pr-previews/brightspace-ui-core/pr-${{ github.event.number }}
```
svanherk, line 76: I think I pictured this as:
`bucket-path: s3://d2l.dev/pr-previews/BrightspaceUI/core/pr-${{ github.event.number }}`

But I guess it doesn't _really_ matter, as long as there are no collisions in what people choose 🤔

svanherk: If we did do `org/repo`, we could actually add a special thing to `repo_settings` so someone could do:
```
pr_previews: true
```
and under the hood it would setup the `d2ldev` piece.  But that doesn't work with the "main" site piece.

Just spit-balling here - what are your thoughts on this:
- All previews go into `d2l.dev/pr-previews`
  - `pr-previews/main/<org>/<repo>/` for the main site that doesn't expire
  - `pr-previews/prs/<org>/<repo>/pr-<#>` for the PR sites
- The expiry prefix is changed to `pr-previews/prs`
- For now, someone sets this in `repo-settings` (need to confirm this doesn't blow up):
  ```
  d2ldev:
    s3:
      - subdomain: pr-previews
        path: main/<org>/<repo>
      - subdomain: pr-previews
        path: prs/<org>/<repo>
  ```
  - But we eventually switch this to `pr_previews: true` and set it for them
- If someone wants a fancy path like `brightspace-ui-core.d2l.dev`, they can set it up in `d2l.dev`.  But they don't actually _have_ to set up anything in `d2l.dev` if they're fine with `https://pr-previews.d2l.dev/prs/BrightspaceUI/core/pr-${{ github.event.number }}/` and `https://pr-previews.d2l.dev/main/BrightspaceUI/core/`

eKoopmans: I'm open to that, except that I think the phrase "PR preview" doesn't apply any more to the "main" static builds.

What about calling them all `repo-demos` or `code-demos`?

svanherk: Yeah good with either of those!  Or even `demos` could work, if we want to try to make the default urls as easy to remember as possible (maybe not useful with url autocomplete lol)

eKoopmans: Okay I might be back to being against this, mostly because of how the URLs would work:

- Suppose the generic prefix was `demos` (I also like the prefix `static`)
- So we have PR previews at e.g. https://demos.d2l.dev/BrightspaceUI/core/prs/pr-1234/, which sounds fine
    - or maybe we could remove the `prs/` piece: https://demos.d2l.dev/BrightspaceUI/core/pr-1234/
- But then the main demo would either be at:
    - https://demos.d2l.dev/BrightspaceUI/core/main/ (ugly), or
    - https://demos.d2l.dev/BrightspaceUI/core/ (which is clean but means the repo is _forced_ to avoid the special folder name `prs` or `pr-.+`)

I don't like that folder restriction, but maybe it's fine...

svanherk: Wouldn't the `pr` and `main` need to be at the beginning to be able to set the retention properly, since we can't use wildcards in the prefix?  Like https://demos.d2l.dev/main/BrightspaceUI/core/ (or I think we could get away with https://demos.d2l.dev/BrightspaceUI/core/) and https://demos.d2l.dev/prs/BrightspaceUI/core/

I guess we could map it in some way with the routes, but this way maybe alleviates your concern cause there's no folder name they need to avoid. They publish their main site to `d2ldev/demos/<org>/<repo>` and their PR sites to `d2ldev/demos/prs/<org>/<repo>/<whatever-they-want-really>-<pr_num>`.  And the lifecycle prefix is "/demos/prs/"?

eKoopmans: Yeah the prefix would need to be there in the bucket, but like you said the routing doesn't need to match the bucket.

I like keeping the main demo link clean as https://demos.d2l.dev/BrightspaceUI/core/. You're right that sticking the `prs` prefix into the URL works, though it's a bit buried there...

We could also just do a completely different route and serve the `prs/` bucket up at https://pr-previews.d2l.dev/BrightspaceUI/core/. That might be an anti-pattern though, `d2l.dev` is generally set up so the bucket prefix matches the subdomain.

So then - why not just two separate bucket prefixes? Keep `pr-previews` as it currently is, and just add one for `demos`? Is there any benefit to trying to merge them into one, apart from their content being similar? Keeping `pr-previews` clearly separate might also make the expiration rule more obvious.

eKoopmans: Alright I thought on it a bit, I can get behind https://demos.d2l.dev/prs/BrightspaceUI/core/ - the URL for these aren't too important anyway, since you're only going to be visiting them as links from PRs. And for that same reason, having a whole subdomain dedicated to them isn't too useful.

Any other thoughts before I go configure that in d2l.dev?

svanherk: I'm good with either of your last two comments - going with https://demos.d2l.dev/prs/BrightspaceUI/core/, or just making two bucket prefixes to keep things clear, since we could still wrap that in `repo-settings` behind a single command.  I was trying to think of two buckets prefixes that worked nicely together lol... `demo` and `demo-prs`? `pr-preview` and `repo-preview`? `preview` and `pr-preview`?

eKoopmans:
Good news: I've updated this to use `live.d2l.dev`!

Bad news: It's failing... 😞

https://github.com/BrightspaceUI/core/actions/runs/5965826665/job/16184501505?pr=3898#step:8:93

> upload failed: build/components/alert/alert.js to s3://d2l.dev/live/prs/BrightspaceUI/core/pr-3898/components/alert/alert.js An error occurred (AccessDenied) when calling the PutObject operation: Access Denied

I don't know why... the role should have access to `s3://d2l.dev` with the prefix `live/prs/BrightspaceUI/core`, from https://github.com/Brightspace/repo-settings/pull/1261. My only thoughts right now are:

- maybe the permissions take longer to update? (though I definitely waited until after CD on `repo-settings` had finished)
- maybe something to do with capital letters in the prefix/S3 route?

svanherk:
> maybe the permissions take longer to update? (though I definitely waited until after CD on repo-settings had finished)

You can check when the tokens last rotated here: https://tokens.d2l.dev/status
So _could_ be that, I didn't do the timezone math lol

eKoopmans:
Very cool, thanks Stacey! I think that's it, it last updated at 11:10 eastern, and the job finished 11:21.

eKoopmans:
Okay we're in business! 🚀

svanherk (approved):
Yay! Thanks for all the work this ended up being @eKoopmans! 😅

.github/workflows/ci.yml
```diff
@@ -66,14 +66,14 @@ jobs:
...
-          role-to-assume: arn:aws:iam::037018655140:role/brightspace-ui-core.d2l.dev-githubactions-access
+          role-to-assume: arn:aws:iam::022062736489:role/r+BrightspaceUI+core+repo
           role-duration-seconds: 3600
           aws-region: ca-central-1
 
       - name: Publish
         uses: BrightspaceUI/actions/publish-to-s3@main
         with:
-          bucket-path: s3://brightspace-ui-core.d2l.dev/branches/pr-${{ github.event.number }}
+          bucket-path: s3://d2l.dev/live/prs/BrightspaceUI/core/pr-${{ github.event.number }}
```
omsmith, line 76: Consider making this a little more copy-pastable

```suggestion
          bucket-path: s3://d2l.dev/live/prs/${{ github.repository }}/pr-${{ github.event.number }}
```

eKoopmans: Good call Owen, I've updated [the docs](https://desire2learn.atlassian.net/wiki/spaces/DEVCENTRAL/pages/3493560351/How+To+Host+a+Static+Site+on+d2l.dev) to use `{{ github.repository }}` too!

eKoopmans:
🎉

image

PR #3899, opened 2023-08-17
DE54475 - Fix sticky list controls in dialog issue

svanherk:
I tried to set this up with `vdiff`s (both the old style which were super flaky and https://github.com/BrightspaceUI/core/pull/3892), but I'm running into issues and I don't want to merge flaky tests.  So for now, I've updated the demo pages for list and dialog and the list visual-diffs. dialog vdiffs will come later.

The problem:
Lists jump outside of their own width 12px for their focus ring, and hover adds even a bit more with a shadow effect.  We need to tell `selection-controls` to jump out as well to match, or else you can see the effects around it when scrolled:
image

We are exposing this as a custom css property, with a default of `0` for `d2l-selection-controls`. Then `list` uses and exposes its own property, with a default of `18px` padding.  Dialog overrides this property so any `d2l-list-controls` inside will match its padding exactly (`30px` default, `20px` mobile, `0px` for `no-padding` mode).


svanherk:
If this looks ok, I'm still trying to get this in before branching.  But if there's concerns, we can discuss pushing it or certfixing.  It definitely felt like _everything_ was against me trying to get this in 😩

components/list/list-controls.js
```diff
@@ -37,6 +38,7 @@ export class ListControls extends SelectionControls {
 
 		const parent = findComposedAncestor(this.parentNode, node => node && node.tagName === 'D2L-LIST');
 		if (parent) this._extendSeparator = parent.hasAttribute('extend-separators');
+		if (this._extendSeparator) this.style.setProperty('--d2l-selection-controls-padding', '0px');
```
dbatiste, line 41: Could this be put in the CSS inside of the `.d2l-list-controls-extend-separator` block that's already defined?

svanherk: That only gets added to the `section`, but I also need it on the `div` that acts as the shadow:
image

But if we like that general pattern better, I could create a new class that gets added to both spots?

dbatiste: I think that's my preference, to avoid setting CSS variables through JavaScript. But is that gets messy, this is fine.

svanherk: Ok, pushed that change.  It's not quite as minimal as I'd like, I played around with different ways to get the names shorter but it always exposed a bit too much underlying knowledge of the selection controls

svanherk: Chatted with @dbatiste offline about this. We're going to go with the setting the variable in JavaScript for now. The other simple options felt a bit worse (reflecting a state property or passing classes in that exposed too much of how the underlying `selection-controls` worked). We could go with a larger change to refactor the rendering of the controls, something like this:
```
render() {
    return html`
        <div class="extend-separators">this._render()</div>
    `;
}
```
If we decide that's worth doing, that'll be a follow-up PR, keeping this one easier to backport.

dbatiste (approved):

PR #3904, opened 2023-08-21
feat: Add skeleton grouping to `d2l-collapsible-panel-group`

jameswklassen:
## Description

Make `d2l-collapsible-panel-group` a "skeleton group". Any time a collapsible panel is in a `skeleton` state, all panels will also be in a skeleton state.

## Rally
[US156732](https://rally1.rallydev.com/#/15545167705d/dashboard?detail=%2Fuserstory%2F720188082449&parentRef=%2Fportfolioitem%2Ffeature%2F705085592923)


dlockhart (approved):
That was easy! Much easier now before these get used 100s of times too.

PR #3905, opened 2023-08-21
input-number: incorrect error message when required and invalid

margaree:
Rally: [DE54446](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F718773543971)

This fixes the problem being seen, but I think the problem is actually at a deeper level:

**Scenario**:
When an `input-number` is `required` and also invalid on initial load (e.g., `value` is negative when `min` is 0), the error message "{label} is required" would appear instead of the actual error message (e.g., "Number must be greater than or equal to 0"). Since it does have a value the "is required" message is not a valid error.

**Issue**:
We have an `await this.updateComplete` before the input gets its value (https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-text.js#L620), which was needed to fix another defect, so when validation is happening the `input` doesn't actually have its value and so it's returning a `valueMissing` validity state at that moment, and then `input-number` doesn't check again because from its perspective nothing has changed.

At a deeper level, what's happening is `input-number` digs in and gets `input-text`'s input error message. Then when `input-text` eventually finishes loading it's thinking "I never had any errors so I'm not going to dispatch the `invalid-change` event (in FormElementMixin) because nothing was ever wrong" even though `input-text` did have something wrong, so `input-number` doesn't know to update the error message.

The console logs below happen in `requestValidate` in `FormElementMixin` and in `_handleInvalidChange` in `input-number`.

Screenshot 2023-08-04 at 2 58 21 PM

**Fix Summary**:
Similarly to what's being done in `input-number`'s `validate` (https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-number.js#L412) this just awaits `input-text`'s `updateComplete`.

There are a few options to fix this, but other options could be more fragile because they'd involve modifications to `FormElementMixin`. I think the one here does make sense since we would want `input-text` to finish its processing of the updated value before the more outer `input-number` handles validation.

Let me know if you have any other ideas.


components/inputs/input-number.js
```diff
@@ -390,6 +390,9 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 		});
 
 		if (checkValidity) {
+			const inputTextElem = this.shadowRoot.querySelector('d2l-input-text');
```
dbatiste, line 393: fwiw, this is exactly what is done further below inside of the `validate()` method.

components/inputs/test/input-number.visual-diff.html
```diff
@@ -7,6 +7,9 @@
 			import '../../typography/typography.js';
 			import '../input-number.js';
 			import '../../button/button-icon.js';
+			import { querySelectorComposed } from '../../../helpers/dom.js';
+
+			window.querySelectorComposed = querySelectorComposed;
```
dbatiste, line 12: Apologies if `getRectTooltip`'s dependency on `querySelectorComposed` tripped you up. Ideally we'd never be poking into the internals. If this caused you any headaches, maybe we would add a log to help if the helper is `undefined`, or update the helper to dynamically import this method so we're not relying on the fixture file loading it. 🤷‍♂️

components/inputs/test/input-number.test.js
```diff
@@ -455,6 +455,7 @@ describe('d2l-input-number', () => {
 			const elem = await fixture(minMaxFixture);
 			setTimeout(() => setInnerInputValue(elem, '0'));
 			await oneEvent(elem, 'change');
```
dlockhart, line 457: Could we `await` an `invalid-change` event here instead of the `change` and `updateComplete`?

components/inputs/test/input-number.test.js
```diff
@@ -455,6 +455,7 @@ describe('d2l-input-number', () => {
...
+			await elem.updateComplete;
 			expect(elem.invalid).to.be.true;
```
dlockhart, line 459: Would it be possible to add a test that would have originally failed before this fix but now passes? Just so we don't accidentally re-introduce it.

components/inputs/test/input-number.visual-diff.html
dlockhart: The new vdiff stuff should hopefully resolve all of this, since the HTML & JS get merged.

dbatiste (approved):
I took a brief look at the code and didn't see any other input components that need this. 

I did notice that unlike this component, `d2l-input-date`'s `validate()` method does not `await inputTextElem.updateComplete` of the underlying `d2l-input-text`. Presumably it's not an issue, but I find it interesting/intriguing that it's different.

components/inputs/test/input-number.visual-diff.html
margaree: No worries! The change happened between when I'd written the code initially and when I put up the PR, so I was a bit confused why it broke, but it was a quick fix.

components/inputs/test/input-number.test.js
```diff
@@ -399,7 +399,7 @@ describe('d2l-input-number', () => {
 		].forEach((test) => {
 			it(test.name, async() => {
 				const elem = await fixture(test.fixture);
-				if (test.value !== null) elem.value = test.value;
+				if (test.value !== null && test.value !== undefined) elem.value = test.value;
```
margaree, line 402: When I was working on adding a test I noticed that if I had no value in my test object then it was setting the elem value to undefined (but I had actually defined the value on the fixture), though maybe it was done this way on purpose? Otherwise having the positive case (`if (test.value) elem.value = test.value`) might be more clear

components/inputs/test/input-number.test.js
```diff
@@ -411,6 +411,13 @@ describe('d2l-input-number', () => {
 				expect(elem.invalid).to.equal(!!test.expectedError);
 			});
 		});
+
+		it('should be invalid when required and value is less than min', async() => {
+			const elem = await fixture(html`<d2l-input-number label="label" min="5" required value="0"></d2l-input-number>`);
+			await elem.updateComplete;
```
dlockhart, line 417: Is this line needed? Our wrapper around `fixture` should be waiting for that already.

dlockhart (approved):
Awesome!

PR #3909, opened 2023-08-22
[ignore] list-item: colour indicator

margaree:


components/list/demo/demo-list-nested.js
```diff
@@ -197,6 +197,7 @@ class ListDemoNested extends LitElement {
 		return html`
 			<d2l-list-item
 				action-href="${this.includeActionHref ? 'http://www.d2l.com' : ''}"
+				color="#ffaa47"
```
margaree, line 200: Thoughts on if the # should be included by the consumer or if we should be adding that?

components/list/list-item-mixin.js
```diff
@@ -415,6 +440,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		if (changedProperties.has('breakpoints')) {
 			this.resizedCallback(this.offsetWidth);
 		}
+		if (changedProperties.has('color')) {
+			this.shadowRoot.querySelector('.d2l-list-item-color-inner').style.backgroundColor = this.color;
```
margaree, line 444: Should we do any verification that this is a hex code or rely on the consumer to handle that?

margaree (commented):
@dbatiste this covers the most basic implementation, but just wanted to see if you had any thoughts at this point before I get into the more complicated cases. 

On the demo site the changes can be seen on the [list demo](https://pr-3909-brightspace-ui-core.d2l.dev/components/list/demo/list.html) and the [list-expand-collapse demo](https://pr-3909-brightspace-ui-core.d2l.dev/components/list/demo/list-expand-collapse.html)

components/list/demo/demo-list-nested.js
dbatiste: My first thought is that they should be required to provide the `#` and we shouldn't assume anything. Although we might decide to limit this to hex colors, technically we could allow `rgb`, `rgba`, named colors, CSS properties, etc. and we probably shouldn't assume they'll never be accepted.

components/list/list-item-mixin.js
```diff
@@ -67,6 +67,11 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			 * @type {array}
 			 */
 			breakpoints: { type: Array },
+			/**
+			 * A color indicator to appear at the beginning of a list item. Expected value is a valid color hex code (e.g., '#006fbf').
```
dbatiste, line 71: Hex values could also include the alpha channel. ex. `#RRGGBBAA` for semi-transparent colors. I'm not sure if they will try to use semi-transparent colors, or whether we want to actively stop them from doing so. That's probably a follow-up for the designers, and depending on what they say we might want to update this doc comment.

Related: Although it is more work for the list/list-items to do, I think we need the list components to automatically adjust spacing based on whether other items in the same list have color indicators rendered. Allowing consumers to have any influence over this multiplies the nuttiness of all the spacing permutations that Jeff is trying to sort out. So, assuming that, the consumer shouldn't ever have a need for a fully transparent color (`transparent`, `rgba(R,G,B,0)`, `#RRGGBB00`).

components/list/list-item-mixin.js
dbatiste: That's a good question. The browser will ignore anything that is completely invalid. Having said that, if an invalid color is being set here, then it would be good to know so that we can correct the source. If we throw an error then I _think_ it will get logged.  

Aside from that, I do tend to lean towards limiting this to hex values to start, since I think the source of the colors will always produce hex colors. I would also be interested in Jeff's input too though.

margaree:
Thanks for the feedback! Closing this PR, and will be opening new PRs with the different pieces of this functionality.

PR #3911, opened 2023-08-23
Detect and fix floating buttons obscuring focused items

awikkerink:
Potential fix for [DE54223](https://rally1.rallydev.com/#/?detail=/defect/713450094077&fdp=true)

Since this would affect multiple places, I'm going to look at other options first. Just putting this here as a possible solution.


components/button/floating-buttons.js
```diff
@@ -270,6 +276,27 @@ class FloatingButtons extends RtlMixin(LitElement) {
 		});
 	}
 
+	_scrollIfFloatObsuringFocus() {
+
+		if (!this._floating) return;
```
dlockhart, line 281: Bail super early if we're not even floating.

components/button/floating-buttons.js
```diff
@@ -270,6 +276,27 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
+
+		const currentFocusedItem = getComposedActiveElement();
```
dlockhart, line 283: @awikkerink I found that `document.activeElement` can return parent custom elements in certain scenarios and that was messing with the height calculations. This ensures that it's always the actual element with focus.

components/button/floating-buttons.js
```diff
@@ -270,6 +276,27 @@ class FloatingButtons extends RtlMixin(LitElement) {
...
+		if (isComposedAncestor(this, currentFocusedItem)) return
+
+		const { y: focusedY, height: focusedHeight } = currentFocusedItem.getBoundingClientRect();
+		const { y: floatingY, height: floatingHeight } = this.shadowRoot.querySelector('.d2l-floating-buttons-container').getBoundingClientRect();
+		if (focusedY === 0 || floatingY === 0) return;
+
+		const isObscuring = (floatingY - focusedY) < Math.max(MINIMUM_TARGET_SIZE, focusedHeight);
```
dlockhart, line 290: I switched this around to `Math.max` from `Math.min` -- @awikkerink does that make sense? This ensures that the entire height of the focused element isn't being obscured. Otherwise if say a button that's `42px` height was showing at least `24px`, it wouldn't scroll.

I'm actually tempted to remove the `MINIMUM_TARGET_SIZE` bit altogether since anything that's focusable should already be at least `24px` high.

dlockhart (commented):
Thanks Anthony for the PR! After testing it out, I've made some tweaks and refactored things a little bit. Once @bearfriend's PR to switch floating-button's visual diff tests to the new library merges, I'll add a new vdiff test for this scenario.

components/button/floating-buttons.js
awikkerink: The reason I added this logic was if there was a focusable element (usually with tabIndex=-1) that takes up the entire height of the page, clicking it would result in the page being scrolled to the bottom.

dlockhart: Hmm... I wonder if that was caused by using `document.activeElement` instead of the composed active element? We could put an upper bound on it maybe... like if the focusable thing is taller than `100px`.

awikkerink: Screenshot 2023-08-30 143217

awikkerink: For reference, the page is `/d2l/le/lo/6609/outcomes-gradebook/settings`.

awikkerink: > I'm actually tempted to remove the MINIMUM_TARGET_SIZE bit altogether since anything that's focusable should already be at least 24px high.

I'm not checking if the element is at least 24px high, but only scrolling if enough of the element is obscured to avoid scrolling down the entire page for a large element.

dlockhart: Ok I'm convinced, I'll flip this back.

awikkerink: I'm fine with increasing the height from 24px to 100px or something. I just didn't want the whole page to scroll

dlockhart:
Alright, I've pulled in the new vdiff test and it's now showing as fixed. This is ready for wider review.

awikkerink:
LGTM ✔️. I can't approve since I opened the PR

bearfriend (approved):

PR #3915, opened 2023-08-24
list-item: Colour Indicator Part 1 - Basic Changes

margaree:
Very similar to the other PR with some visual diffs.

I have other pieces of this done, but wanted to put this up first, as it has the least amount of changes to the visual diffs so it's easier to see what's changed.

Done in follow up PRs:
- Leaving space for the color slot when sibling item has color (similar to expand/collapse) - this is pending design approval but the change is fairly simple
- `extend-separators` - what is the expected behaviour/spacing?
- Width should be `3px` at small parent list widths


components/list/demo/list-color-indicator.html
```diff
@@ -0,0 +1,336 @@
+<!DOCTYPE html>
```
margaree, line 1: I created a demo page just to make it easier to see changes during development. If this is too much clutter in the demo pages I can remove this one post-dev and just add color indicators throughout the other pages.

components/list/list-item-mixin.js
```diff
@@ -311,6 +317,14 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				margin: 0;
 			}
 
+			:host([color]) [slot="outside-control-container"] {
+				margin-left: -6px;
```
margaree, line 321: This changes things a bit from current behaviour, which is `-12px` margin-left, which shows up on hover. The last 2 items in the list below really show the difference. If we add the ability for siblings to leave space when a sibling has a color then the difference will no longer exist, but it is a change in behaviour. Any thoughts on this? I'll be checking with design as well.

Screenshot 2023-08-24 at 1 11 13 PM

components/list/list-item-mixin.js
```diff
@@ -596,6 +630,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			'd2l-list-item-content-extend-separators': this._extendSeparators,
 			'd2l-dragging-over': this._draggingOver
 		};
+		const colorStyles = {
+			backgroundColor: this.color || undefined
```
margaree, line 634: Jeff seems to be leaning towards not having validation and just not having the color render if it's invalid.

If we do want to validate but also want to allow any possible valid color value, after color is set we could then check the `background-color` value and see if it got wiped out and throw if so.

components/list/demo/list-color-indicator.html
```diff
@@ -0,0 +1,336 @@
...
+			import '../list.js';
+		</script>
+	</head>
+	<body unresolved>
+
+		<d2l-demo-page page-title="d2l-list Color Indicator">
+
+			<h2>Basic List</h2>
+
+			<d2l-demo-snippet code-view-hidden>
```
dbatiste, line 25: Is there a reason to hide the code-view in these demos? I'm not totally sure but I think we might just hide the code view when it contains test components that aren't relevant to the consumer, or when the code-view might interfere with the demo (floating workflow buttons?).

components/list/list-item-generic-layout.js
```diff
@@ -198,6 +204,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			<slot name="content-action" class="d2l-cell" data-cell-num="6"></slot>
 			<slot name="outside-control" class="d2l-cell" data-cell-num="2"></slot>
 			<slot name="outside-control-action" class="d2l-cell" data-cell-num="1"></slot>
+			<slot name="color"></slot>
```
dbatiste, line 207: `color-indicator` would also be a good slot name. 🤷‍♂️

components/list/demo/list-color-indicator.html
margaree: It was mostly just for development purposes since there was a lot of code, but I'll remove this property.

components/list/list-item-mixin.js
```diff
@@ -18,6 +18,9 @@ import { ListItemRoleMixin } from './list-item-role-mixin.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 import ResizeObserver from 'resize-observer-polyfill';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
+import { styleMap } from 'lit-html/directives/style-map.js';
+
+const HEX_REGEX = /#([0-9a-fA-F]){6}(?:[0-9a-fA-f]{2})?/; // 6 or 8 character hex code
```
dbatiste, line 23: We could create a `color.js` helper that contains a method `isHexColor(alpha)`. This Regex could be moved there, as well as the one that `input-color.js` uses. We could have a few unit tests to cover these in isolation from the components.

Related: when I did my inFusion Dark Mode work, I think I ended up creating a similar color.js that had a lot of color functions, including color parsing, validation, and other things. I see that we have a `contrast.js` helper that I think was a contribution from outside Gaudi - it's functionality could also be represented in `color.js` but I would want to refactor it, and it's probably used by someone outside of core.

components/list/list-item-mixin.js
dbatiste: Are you sure you'd want to override the CSS immediately above (`margin: 0`) for the draggable case? Still refreshing my memory as to what this margin is for. 😬

components/list/test/list.test.js
```diff
@@ -258,6 +258,30 @@ describe('d2l-list-item', () => {
 
 	});
 
+	describe('property validation', () => {
+
+		it('should set color as expected when valid 6 character hex color', async() => {
```
dbatiste, line 263: My hope is that the color validation can be tested as a helper, and then we only just need some visual-diffs to test that the color indicators shows the color in the correct place.

components/list/demo/list-color-indicator.html
dbatiste: If you think it's worthwhile, especially for the huge nested demos, that's totally 😎 .

components/list/list-item-mixin.js
margaree: The design specifies 6 px between the color and left border on hover. If we don't have this then it looks like this:
Screenshot 2023-08-28 at 10 59 12 AM

dbatiste: Gotcha. What about the draggable case?

margaree: Ah right this does add extra to the left in that case. Will fix

helpers/color.js
```diff
@@ -0,0 +1,16 @@
+const HEX_REGEX = /^#([0-9a-fA-F]){6}$/;
```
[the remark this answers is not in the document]
dbatiste: This seems good to me. I think `input-color`'s validation enforced the `#` as well, and I think the component limits the user to using opaque colors. Of course, maybe we will enhance it in the future when we have web component color palette. If someone ever needs to validate a color without throwing or mutating the value, we can always factor that part out into a separate `isHexColor` method.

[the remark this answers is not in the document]
dbatiste: Actually, it looks like it didn't require the starting `#`.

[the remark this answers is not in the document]
dbatiste: Do you know what those 3 extra characters are for? The comment in the code makes no mention of it, but I see it in the Regex as you mentioned.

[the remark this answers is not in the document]
margaree: I do not - I haven't been able to find any additional context on why 3 instead of 2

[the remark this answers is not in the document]
dbatiste: Oh I see, I think it is actually allowing for either 6 or 3 character hex codes, which would technically be valid. `#RGB` === `#RRGGBB`. In actuality, 6, 8, 3, and 4 character versions are technically valid [per the spec](https://www.w3.org/TR/css-color-4/#hex-notation).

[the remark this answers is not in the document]
bearfriend: It's probably good to be specific about how many characters are valid for each type and not using a range including 5 and 7.

Something like:
```javascript
const HEX_CHAR = '[0-9a-f]';
const HEX_REGEX = new RegExp(`^#(${HEX_CHAR}{3}|${HEX_CHAR}{6})$`, 'i');
const HEX_REGEX_ALPHA = new RegExp(`^#(${HEX_CHAR}{4}|${HEX_CHAR}{8})$`, 'i');
```

dbatiste (approved):

PR #3917, opened 2023-08-24
list-item: Color Indicator Part 2 - Sibling Spacing

margaree:
This follows the same pattern expand-collapse uses in order to add spacing when a sibling has a color but the current list item does not.

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -96,7 +96,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			_focusingPrimaryAction: { type: Boolean, attribute: '_focusing-primary-action', reflect: true },
 			_highlight: { type: Boolean, reflect: true },
 			_highlighting: { type: Boolean, reflect: true },
-			_hasNestedList: { state: true }
+			_hasNestedList: { state: true },
+			_renderColorSlot: { type: Boolean, reflect: true, attribute: '_render-color-slot' },
+			_siblingHasColor: { state: true }
 		};
 	}
 
@@ -317,10 +319,10 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				margin: 0;
 			}
 
-			:host([color]) [slot="outside-control-container"] {
+			:host([_render-color-slot]) [slot="outside-control-container"] {
 				margin-left: -6px;
 			}
-			:host([dir="rtl"][color]) [slot="outside-control-container"] {
+			:host([dir="rtl"][_render-color-slot]) [slot="outside-control-container"] {
 				margin-left: 0;
 				margin-right: -6px;
 			}
@@ -415,6 +417,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		this._contentId = getUniqueId();
 		this._displayKeyboardTooltip = false;
 		this._hasNestedList = false;
+		this._renderColorSlot = false;
+		this._siblingHasColor = false;
 	}
 
 	get breakpoints() {
@@ -448,6 +452,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		super.updated(changedProperties);
 		if (changedProperties.has('breakpoints')) {
 			this.resizedCallback(this.offsetWidth);
+		}
+		if (changedProperties.has('_siblingHasColor') || changedProperties.has('color')) {
+			this._renderColorSlot = this.color || this._siblingHasColor;
 		}
 	}
 
@@ -501,6 +508,10 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 		if (reduceMotion) this.scrollIntoView(alignToTop);
 		else this.scrollIntoView({ behavior: 'smooth', block: alignToTop ? 'start' : 'end' });
+	}
+
+	updateSiblingHasColor(siblingHasColor) {
+		this._siblingHasColor = siblingHasColor;
 	}
 
 	_getFlattenedListItems(listItem) {
@@ -631,7 +642,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			'd2l-dragging-over': this._draggingOver
 		};
 		const colorStyles = {
-			backgroundColor: this.color || undefined
+			backgroundColor: this._renderColorSlot ? this.color : undefined
 		};
 
 		const primaryAction = ((!this.noPrimaryAction && this._renderPrimaryAction) ? this._renderPrimaryAction(this._contentId) : null);
@@ -651,7 +662,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				${this._renderDragHandle(this._renderOutsideControl)}
 				${this._renderDragTarget(this.dragTargetHandleOnly ? this._renderOutsideControlHandleOnly : this._renderOutsideControlAction)}
 				<div slot="control-container"></div>
-				${this.color ? html`
+				${this._renderColorSlot ? html`
 				<div slot="color" class="d2l-list-item-color-outer">
 					<div class="d2l-list-item-color-inner" style="${styleMap(colorStyles)}"></div>
 				</div>` : nothing}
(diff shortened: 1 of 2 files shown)


components/list/list-item-mixin.js
```diff
@@ -460,6 +465,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		if (changedProperties.has('breakpoints')) {
 			this.resizedCallback(this.offsetWidth);
 		}
+		if (changedProperties.has('_siblingHasColor') || changedProperties.has('color')) {
```
dbatiste, line 468: I _think_ this computed reactive property will result in a subsequent iteration of the render life cycle. One way to avoid that would be to compute `_renderColorSlot` before `render` instead of after by using the `willChange` callback. 

```javascript
  // only need to check changed properties for expensive computation (according to Lit docs)
  this._renderColorSlot = this.color || this._siblingHasColor;
}
```

An alternative would be to sprout `_siblingHasColor` and update the CSS to check for both `_sibling-has-color` and `color`, thereby avoiding computing and sprouting `_renderColorSlot` altogether. Seems like `_renderColorSlot` really just buys us slightly simpler CSS.

Nit: on naming. If we keep `_renderColorSlot`, I'd probably rename it to something like `_hasColorSlot` or something along those lines, since anything that starts with `_render` sounds like a method that renders something and is how most of our private render helpers are named. I'd probably also rename `_siblingHasColor` to `_hasSiblingColor`. I think Ian introduced the XHasY naming here... I probably let it go because he had a lot of other more important feedback to address.

components/list/list.js
```diff
@@ -224,13 +225,14 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			e.stopPropagation();
 		}
 		const items = this.getItems();
+		let aChildHasColor = false;
 		let aChildHasToggleEnabled = false;
 		for (const item of items) {
```
dbatiste, line 230: I wish we didn't have to do this looping. There's some discussion of adding a CSS API to detect when slots contains things.

components/list/list-item-mixin.js
dbatiste: Btw, I see Ian. introduced the `_render` as well. So what you have for naming is consistent with that. I'd almost be torn about being inconsistent with that. Anyway, main thing above is to be careful about the re-renders that computed properties can introduce.

components/list/list.js
dbatiste: This handles updating the siblings whenever items are added or removed from the slot. Do we have something for handling the scenario where the list loads, and then later a color is added or removed from an item? For example in response to something the user has done.

components/list/list-controls.js
```diff
@@ -19,31 +20,47 @@ export class ListControls extends SelectionControls {
...
+				margin: 0 -0.9rem;
+			}
 			:host([no-sticky]) {
 				z-index: auto;
 			}
-			.d2l-list-controls-extend-separator {
-				padding: 0 0.9rem;
+			.d2l-list-controls-color,
+			.d2l-list-controls-extend-separator  {
+				padding: 6px 1.8rem;
```
margaree, line 31: I found updating the spacing to this caused less issues with the hover state appearing beyond the sticky header

components/list/list-controls.js
```diff
@@ -19,31 +20,47 @@ export class ListControls extends SelectionControls {
...
 		const parent = findComposedAncestor(this.parentNode, node => node && node.tagName === 'D2L-LIST');
 		if (parent) this._extendSeparator = parent.hasAttribute('extend-separators');
 		if (this._extendSeparator) this.style.setProperty('--d2l-selection-controls-padding', '0px');
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('childHasColor') && !this._extendSeparator) {
+			if (this.childHasColor) this.style.setProperty('--d2l-selection-controls-padding', '0px');
+			else this.style.setProperty('--d2l-selection-controls-padding', this._selectionControlsInitialPadding);
```
margaree, line 56: This resets to initial padding value if the color is removed and there is not `extend-separators`. Not sure if there is a better way to do this?

components/list/list-item-mixin.js
```diff
@@ -394,12 +396,12 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				padding-left: 12px;
 				padding-right: 0;
 			}
-			:host([expandable]) .d2l-list-item-color-outer {
-				padding-right: 6px;
+			.d2l-list-item-color-outer + .d2l-list-expand-collapse {
+				margin-left: -6px;
```
margaree, line 400: This wasn't behaving as desired when expanded and a sibling had a color indicator

components/list/list.js
```diff
@@ -210,6 +212,20 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		return items.length > 0 ?  items[0]._getFlattenedListItems().lazyLoadListItems : new Map();
 	}
 
+	_handleColorChange(value, updateSubscribers) {
+		this._childHasColor = value;
+		if (this._controls) this._controls.childHasColor = this._childHasColor;
+		if (updateSubscribers) this._listChildrenUpdatedSubscribers.updateSubscribers();
+	}
+
+	_handleControlsSlotChange(e) {
```
dbatiste, line 221: The [ListControls](https://github.com/BrightspaceUI/core/blob/main/components/list/list-controls.js#L8) extends [SelectionControls](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-controls.js#L18)  which supports [selectionFor](https://github.com/BrightspaceUI/core/blob/main/components/selection/selection-observer-mixin.js#L12) which enables the `d2l-list-controls` to be wired up outside of list's `controls` slot, so I don't think `slotchange` will be enough to cover our bases here. Instead, I _think_ the observer (`d2l-list-controls` in this case) should be updated when the `onSubscribe` callback is invoked, but I could be missing part of how that's supposed to work.

components/list/list-controls.js
```diff
@@ -19,31 +20,48 @@ export class ListControls extends SelectionControls {
 				--d2l-selection-controls-padding: var(--d2l-list-controls-padding, 18px);
 				z-index: 6; /* must be greater than d2l-list-item-active-border */
 			}
+			:host([child-has-color]),
+			:host([_extend-separator]) {
```
dbatiste, line 24: Does this need to be applied to the host, or could it be applied to the main container that `SelectionControls` renders. We didn't really want to sprout private properties on the host if we could help it, which is part of the reason why `_getSelectionControlsContainerClasses` was added - it enables `List`/`TableControls` to specify CSS classes to be rendered on the main internal container element even though they aren't doing any of the actual HTML rendering.

margaree: I'll fiddle with it. There's just some weirdness in the bottom corners (I think it's to do with the drop shadow) on hover when the margin is on the section vs the host:
Screenshot 2023-09-01 at 8 47 03 AM

components/list/list-controls.js
```diff
@@ -22,15 +24,19 @@ export class ListControls extends SelectionControls {
 			:host([no-sticky]) {
 				z-index: auto;
 			}
+			.d2l-list-controls-color,
 			.d2l-list-controls-extend-separator {
-				padding: 0 0.9rem;
+				padding: 0 1.8rem;
```
margaree, line 29: I'll fix this to update only for color

components/list/list.js
```diff
@@ -210,6 +212,12 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		return items.length > 0 ?  items[0]._getFlattenedListItems().lazyLoadListItems : new Map();
 	}
 
+	_handleColorChange(value, updateSubscribers) {
+		this._childHasColor = value;
+		if (this._controls) this._controls.childHasColor = this._childHasColor;
```
margaree, line 217: Will remove this

components/list/list-item-mixin.js
dbatiste: So many scenarios!

dbatiste (approved):
Looks great! I looked through the visual-diffs and commented on a couple which sounds like you have handled. 😄

PR #3922, opened 2023-08-25
List: Move ResizeObserver from list item to parent list

margaree:
We want to tap into this for the color indicator work so this is moving this ResizeObserver for performance.

This seemed relatively straightforward but let me know if there's something I completely missed.

Notes/Questions:
- I considered seeing if we could move the breakpoints logic up to `list` as well but `breakpoints` are set per item so this seemed most straightforward
- Are there cases where the list item would resize but the parent list would not?


dlockhart (approved):
Yeah agreed that this seems straightforward... almost _too_ easy which is a little eery. 😨

components/list/list.js
```diff
@@ -196,6 +208,13 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		return new SelectionInfo(keys, selectionInfo.state);
 	}
 
+	resizedCallback(width) {
```
dbatiste, line 211: How often does this get called when the user is resizing the browser or when the container itself is being resized (like in the primary-secondary template)?

Hopefully the items would only be re-rendered when the break-point gets hit, but we'd be incurring the cost of `getItems()` each time it gets called. I vaguely recall seeing this method coming up in my preliminary perf testing with 1000 items and lots of nesting. Things got pretty janky. Anyway, it would be good to know to what extent this improves perf, and we might consider debouncing this if think it's necessary. Debouncing would affect what the user sees _as_ they are resizing.

margaree: That's a good point. It does run quite often - seems to be when the container is resized.

Could we instead just do something like `this._items = this.getItems();` in `_handleSlotChange` and then just loop through `this._items` in the `resizedCallback`?

margaree: So I perf tested with 47 list items with many nested lists up to 4 deep (can do more than this too if that would define a large list better). The list overall load time is definitely slow and noticeable but I don't know that it's `getItems` causing it. It looks like for the code in `getItems` to run it consistently takes up to 0.1 ms, no matter how many list items. To test this I updated `getItems` to be the following:
```
    getItems(slot) {
        **const start = window.performance.now();**
        if (!this.shadowRoot) return;
        if (!slot) slot = this.shadowRoot.querySelector('slot:not([name])');
        if (!slot) return [];
        const items = slot.assignedNodes({ flatten: true }).filter((node) => {
            return node.nodeType === Node.ELEMENT_NODE && (node.role === 'rowgroup' || node.role === 'listitem');
        });
        **const end = window.performance.now();
        console.log('GETITEMS ' + (end - start));**
        return items;
    }
```
Let me know if this is an incorrect way of testing this.

Before and after this change the load times of the list and list items also seem to average out to about the same. I used the perfMonitor helper to help perf testing with this.

Agreed about breakpoints - I can spin up a separate story for investigation of that? It would likely simplify the implementation to have that be at the list level rather than each `list item`.

svanherk: Not sure I completely followed this convo, but wanted to throw this here in case it's related: [US135486](https://rally1.rallydev.com/#/15545167705ud/search?detail=%2Fuserstory%2F622956168057&keywords=list%20item%20perf). It's a really old story about how expensive `resizedCallback` is and how much it slows down list item loading, and how we might be able to cut the initial number.  Or yeah, maybe remove the `breakpoints` piece altogether

components/list/list.js
```diff
@@ -11,6 +12,15 @@ const keyCodes = {
 
 export const listSelectionStates = SelectionInfo.states;
 
+const ro = new ResizeObserver(entries => {
+	entries.forEach(entry => {
+		if (!entry || !entry.target || !entry.target.resizedCallback) {
```
bearfriend, line 17:
```suggestion
		if (!entry?.target?.resizedCallback) {
```

components/list/list.js
```diff
@@ -11,6 +12,15 @@ const keyCodes = {
...
+			return;
+		}
+		entry.target.resizedCallback(entry.contentRect && entry.contentRect.width);
```
bearfriend, line 20:
```suggestion
		entry.target.resizedCallback(entry.contentRect?.width);
```

components/list/list.js
```diff
@@ -196,6 +208,13 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		return new SelectionInfo(keys, selectionInfo.state);
 	}
 
+	resizedCallback(width) {
+		const items = this.getItems();
+		items.forEach((item) => {
+			if (item.resizedCallback) item.resizedCallback(width);
```
bearfriend, line 214:
```suggestion
			item.resizedCallback?.(width);
```

bearfriend (commented):
Just nits:

margaree:
Closing for now - will be re-evaluating the strategy on this and doing further work on improvements

PR #3929, opened 2023-08-25
US152268 - Migrate alert vdiff tests

svanherk:
Pretty simple migration but one question below.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/16ce5c6a93796314f65af3219ce7ce7b2d7ebbbb/2023-08-28_16-26-34/report/


components/alert/test/alert-toast.vdiff.js
```diff
@@ -0,0 +1,36 @@
...
+
+	[
+		{ name: 'default', template: html`<d2l-alert-toast no-auto-close type="default" open>A default message.</d2l-alert-toast>` },
+		{ name: 'no-close', template: html`<d2l-alert-toast no-auto-close type="success" hide-close-button open>A message.</d2l-alert-toast>` },
+		{ name: 'button-close', template: html`<d2l-alert-toast no-auto-close type="warning" button-text="Do it!" open>A message.</d2l-alert-toast>` },
+		{ name: 'subtext-button-close', template: alertWithSubtextAndCloseButton }
+	].forEach(({ name, template }) => {
+		it(name, async() => {
+			const elem = await fixture(template, { viewport: { width: 700 } });
+			await expect(elem).to.be.golden({ margin: 50 });
```
svanherk, line 21: These tests had a very custom `rect` set ([see here](https://github.com/BrightspaceUI/core/blob/4b10bc2a21ce56f6cb6297607269172ce5031ac9/components/alert/test/alert-toast-helper.js#L6-L11)). We can set this `margin` to get it as close as possible to the old one (with more top margin, but still showing page position to the side/bottom).  Or we can just remove it and let the full page tests below handle testing position.
Removing `{ margin: 50 }`:
image
With `{ margin: 50 }`:
image
Old golden:
image

dlockhart: This feels like the kind of thing that should just be done with a full page test?

components/alert/test/alert.vdiff.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+describe('alert', () => {
+
+	[
+		{ name: 'type-default', template: html`<d2l-alert type="default">A default message.</d2l-alert>` },
+		{ name: 'type-success', template: html`<d2l-alert type="success">A success message.</d2l-alert>` },
+		{ name: 'type-critical', template: html`<d2l-alert type="critical">A critical message.</d2l-alert>` },
+		{ name: 'type-warning', template: html`<d2l-alert type="warning">A warning message.</d2l-alert>` },
+		{ name: 'type-error', template: html`<d2l-alert type="error">An error message.</d2l-alert>` },
```
dlockhart, line 21: Nit: could maybe refactor these "type" ones out.

```javascript
['default', 'success', 'critical', 'warning', 'error'].forEach(type => {
  it(`type-${type}`, async() => {
    const elem = await fixture(html`<d2l-alert type="${type}">A ${type} message</d2l-alert-type>`);
    await expect(elem).to.be.golden();
  });
});
```

dlockhart (approved):

components/alert/test/alert-toast.vdiff.js
svanherk: Yeah we can switch to that too, I'm not sure why that wasn't done originally.  I'll try to get it to line up

dbatiste (approved):

components/alert/alert-toast.js
```diff
@@ -144,7 +144,7 @@ class AlertToast extends LitElement {
 	render() {
 		return html`
 			<div
-				class="d2l-alert-toast-container"
+				class="d2l-alert-toast-container vdiff-target"
```
svanherk, line 147: Do we want to leave this here in case anyone ever wraps the toast and needs a way to target just that component? Or remove for now and add back if that happens?

components/alert/test/alert-toast.vdiff.js
```diff
@@ -0,0 +1,36 @@
...
+		{ name: 'default', height: 108, template: html`<d2l-alert-toast no-auto-close type="default" open>A default message.</d2l-alert-toast>` },
+		{ name: 'no-close', height: 106, template: html`<d2l-alert-toast no-auto-close type="success" hide-close-button open>A message.</d2l-alert-toast>` },
+		{ name: 'button-close', height: 108, template: html`<d2l-alert-toast no-auto-close type="warning" button-text="Do it!" open>A message.</d2l-alert-toast>` },
+		{ name: 'subtext-button-close', height: 188, template: alertWithSubtextAndCloseButton }
+	].forEach(({ name, template, height }) => {
+		it(name, async() => {
+			await fixture(template, { viewport: { width: 700, height } });
+			await expect(document).to.be.golden();
+		});
+	});
```
svanherk, line 23: Ok, I've updated the tests to take a screenshot of the full page and set the height of the viewport to match the old test.  But if we don't care about doing that, I can change it to just "190" for all of them

components/alert/alert-toast.js
dlockhart: I'm good either way!

components/alert/test/alert-toast.vdiff.js
dlockhart: Yeah this seems a little fragile, I'd lean towards just picking a single value that works.

PR #3932, opened 2023-08-25
list-item: Color Indicator Part 4 - Slim Style

margaree:
To do:
- [ ] Update breakpoint with value from design

This will work as expected once PR to move resize observer to the list is merged.


margaree:
Will open a fresh PR for this

PR #3935, opened 2023-08-28
US152268 - Migrate dialog-confirm and dialog-ifrau vdiff tests

svanherk:
Like the previously migrated `dialog` tests, I've turned off the `native` tests for now.

FYI @grant-cleary since you know more about the `ifrau` tests.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/5d76917d6b2e6eeec4a851c6bbbabe9f8e975f5a/2023-08-28_19-00-54/report/


grant-cleary (approved):
This brings me back. ifrau bits look good to me.

components/dialog/test/dialog-ifrau-contents.vdiff.html
```diff
@@ -18,24 +18,13 @@
 <body class="d2l-typography">
 	<d2l-button id="open">Open Dialog</d2l-button>
```
svanherk, line 19: I left this `d2l-button` here since this test requires a whole `html` file anyways, but I can remove it if we want to be consistent with the look of the other tests.  Left in, the `dialog-ifrau` tests result in no diffs 🎉

components/dialog/test/dialog-ifrau-contents.vdiff.html
```diff
@@ -18,24 +18,13 @@
...
 		<div>Line 3</div>
 		<div>Line 4</div>
 		<d2l-button slot="footer" primary>Yes</d2l-button>
 		<d2l-button slot="footer" id="cancel">No</d2l-button>
 	</d2l-dialog>
-
-	<script>
-		const dialog = document.getElementById('ifrau-dialog');
-		dialog.addEventListener('d2l-dialog-open', () => {
-			window.parent.postMessage('ifrau-dialog-open', '*');
```
dbatiste, line 33: Are these postMessage values no longer being used?

svanherk: Yeah, they were used to open the dialog (and then close it in `reset` for the next test).  Now that the tests are isolated I can just start with the dialog opened.  But if we _want_ to test this functionality, I could add it back, or maybe just a dedicated test for opening and one for closing?

components/dialog/test/dialog-confirm.vdiff.js
```diff
@@ -0,0 +1,75 @@
...
+import '../dialog-confirm.js';
+import { expect, fixture, html } from '@brightspace-ui/testing';
+
+const buttons = html`
+	<d2l-button slot="footer" primary>Yes</d2l-button>
+	<d2l-button slot="footer" id="cancel">No</d2l-button>
+`;
+
+const confirmDialog = html`
+	<d2l-dialog-confirm id="confirm" title-text="Title" text="Are you sure?" opened>
```
dbatiste, line 11: Setting up by using `opened` attribute is definitely much cleaner, which is great. I think the helper was implicitly also testing the `d2l-dialog-open` & `d2l-dialog-close` events. I think the open event may still be covered by some unit tests (also implicitly), but I'm not sure the close event is covered. I'm wondering if it would be a good idea for us to add an explicit unit test for those events, since consumers do rely on them. Or maybe you already checked and found they are covered?

svanherk: So I'd say that's _kinda_ covered by the [dialog-mixin.vdiff.js](https://github.com/BrightspaceUI/core/blob/main/components/dialog/test/dialog-mixin.vdiff.js) tests.  But I like the idea of adding non-vdiff, dedicated tests - I'll add some to this PR

dbatiste: If we do, I'm thinking those unit tests should be for the general mixin case.

dlockhart (approved):
Awesome, love how clean these migrations are.

components/dialog/test/dialog-confirm.vdiff.js
svanherk: Ok, just pushed a new file - it'll fail until we expose `oneDefaultPreventedEvent` in `@brightspace-ui/testing`, but is that what you had in mind?

svanherk: Ok, merged the change in `@brightspace-ui/testing` and now CI here is passing.

dbatiste: Yep, exactly. Looks great!

dbatiste (approved):

PR #3937, opened 2023-08-28
US152268 - Migrate async-container-mixin and typography tests

svanherk:
Nothing super interesting about the typography migration.

For the `async-container-mixin` migration, I modified the existing `demo/async-item.js` component so that `test/async-item.js` could be deleted.  The result is that two of the diffs are a slightly bigger version of themselves now, because the demo size was `100px` and the test size was `50px`. 

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/dfedfcf7f988a2343d7d800b25a04108255c65d0/2023-08-28_19-26-07/report/


.gitignore
```diff
@@ -7,8 +7,6 @@ components/*/test/screenshots/golden/
 helpers/test/screenshots/current/
 helpers/test/screenshots/golden/
 lang/fr-on.js
-mixins/*/test/screenshots/current/
-mixins/*/test/screenshots/golden/
```
svanherk, line 11: `async-container-mixin` is the only `vdiff` test in `mixin` so we can clean this up now. You'll need to clean up any lingering  local copies next time you pull after this merges.

components/typography/test/typography.vdiff.js
```diff
@@ -0,0 +1,54 @@
+import '../typography.js';
```
dlockhart, line 1: So including this is a little bit conflict-y with [what vdiff is already including](https://github.com/BrightspaceUI/testing/blob/main/src/server/wtr-config.js#L71). It's _probably_ fine, but we'll need to be careful to keep them in sync if we ever publish new versions of those assets to the CDN.

dlockhart (approved):

components/typography/test/typography.vdiff.js
svanherk: Yeah I can't really think of a way around it to still test the `typography` code, unless we isolated the typography tests into their own job and created a custom `testRunnerHtml` for them?

dlockhart: Yeah doesn't seem worth the trouble since this is working.

PR #3939, opened 2023-08-28
US152268 - Migrate prism vdiff tests

svanherk:
Ok, so this one is interesting.  The original vdiff test was not properly loading typography, _and_ we [add a few styles right into the vdiff page](https://github.com/BrightspaceUI/testing/blob/main/src/server/wtr-config.js#L89-L97).

So what I've added below is what we can do to minimize changes. But I think the commented code is what we actually want, so I'll open a branched PR off this one to switch them and get a different report to compare.

Final Reports:
- https://vdiff.d2l.dev/BrightspaceUI/core/d5a2f8f4258d8e57ac7c3917021ba83238e39997/2023-08-28_21-09-45/report/
- https://vdiff.d2l.dev/BrightspaceUI/core/74b96891e362e86e9f506ba7cccb480a2668b8bd/2023-09-15_20-18-44/report/


.gitignore
```diff
@@ -4,8 +4,6 @@ node_modules/
 .vdiff/
 components/*/test/screenshots/current/
 components/*/test/screenshots/golden/
-helpers/test/screenshots/current/
-helpers/test/screenshots/golden/
```
svanherk, line 8: This is the only `vdiff` test in `helpers`, so we can remove this. You'll need to clean up any lingering local copies next time you pull after this merges.

helpers/test/prism.vdiff.js
```diff
@@ -0,0 +1,102 @@
...
+					const elem = await fixture(createCodeWrapper({ content, dark: color === 'dark', language, lineNumbers, inline }));
+					window.codeFormatted = await formatCodeElement(elem.querySelector('.d2l-code'));
+					await expect(elem.querySelector('pre, p')).to.be.golden();
+				});
+			});
+		});
+	});
+
+	describe('tokens', () => {
+		[
```
svanherk, line 61: I left this in the original order to make it easier to compare to the old test, but I could also rearrange so languages are together.

grant-cleary (approved):
Looks good to me, though Dave is more familiar with these tests than I am, so I'll defer to any comments he may have.

svanherk:
Seeing different results here than locally, going to move this back to draft for now

dbatiste:
> Ok, so this one is interesting. The original vdiff test was not properly loading typography, _and_ we [add a few styles right into the vdiff page](https://github.com/BrightspaceUI/testing/blob/main/src/server/wtr-config.js#L89-L97).

Looks like the intent was there but I (?) failed at copy & paste. 😞 Is it the `letter-spacing`. Our prism styles don't enforce or override it, so it would be whatever the document/body is setting, which is now different.

svanherk:
> > Ok, so this one is interesting. The original vdiff test was not properly loading typography, _and_ we [add a few styles right into the vdiff page](https://github.com/BrightspaceUI/testing/blob/main/src/server/wtr-config.js#L89-L97).
> 
> Looks like the intent was there but I (?) failed at copy & paste. 😞 Is it the `letter-spacing`. Our prism styles don't enforce or override it, so it would be whatever the document/body is setting, which is now different.

Yeah exactly. There's a few things causing differences here:
- The `letter-spacing` just uses whatever the `body` uses because `prism` doesn't set it
- The inline code example is also getting its `color`, `font-size` and `font-family` from the `body` and that's now different, but that's less of a big deal since that's not testing the actual `prism` pieces.
- The language piece in the corner was trying to set Lato but couldn't because the font wasn't included (so that's fixed now)
- It looks like with Puppeteer we were getting the `Consolas` font, but it seems like Playwright's chromium doesn't have access to it. I don't know how to figure out what this new font is 🤔

dlockhart (approved):

PR #3944, opened 2023-08-29
Null check before adding event listener

jmurray-d2l:
Implementing [this suggestion](https://github.com/BrightspaceUI/core/pull/3870#discussion_r1300371107).  We are seeing this fail when using load more for our filters [here](https://github.com/Brightspace/nova/blob/main/src/app/shared/components/activities/activity-filter.js#L506)


svanherk (approved):
Thanks @jmurray-d2l!

PR #3946, opened 2023-08-29
US152268 - Migrate meter vdiff tests

svanherk:
This is another case where the new styles we are adding to the page are causing changes.  I'm going to run it once with those styles off to hopefully give a cleaner diff, and then once with them back on for the final diff.

Cleaner Report (with original typography): https://vdiff.d2l.dev/BrightspaceUI/core/b246776a82d1477ec9bade30dd86dab008d9119c/2023-08-29_15-01-05/report/
Final Report: https://vdiff.d2l.dev/BrightspaceUI/core/b75eb1fb75e327d6683beb7cee9c52764a4a5c05/2023-08-29_15-44-52/report/


dlockhart (approved):
Clean report looks good! Hopefully most vdiff HTML files were including our typography and that ones like these are the exception.

svanherk:
> Clean report looks good! Hopefully most vdiff HTML files were including our typography and that ones like these are the exception.

Yeah I hope so too, this definitely adds some time while deving and rerunning the CI. But I like the confidence I get with the diffs that put the old styles back.

PR #3959, opened 2023-09-01
Update `documentElement.lang` on demo language change

bearfriend:
Currently `documentElement.lang` is not updated on demo pages when the language is changed. This results in the `documentElement.lang` and the shared `DocumentLocaleSettings` being out of sync.

`DocumentLocaleSettings` observes `documentElement.lang` and will update itself.

One notable effect is that switching to languages that use different fonts now updates correctly, appearing as it would in production.


components/demo/test/demo-page.test.js
```diff
@@ -0,0 +1,27 @@
...
+		});
+	});
+
+	describe('language select', () => {
+		it('updates the documentElement language on change', async() => {
+			const elem = await fixture(html`<d2l-demo-page></d2l-demo-page>`);
+			const select = elem.shadowRoot.querySelector('select');
+			select.value = 'ko-kr';
+			setTimeout(() => select.dispatchEvent(new Event('change')));
+			await oneEvent(select, 'change');
```
dlockhart, line 20: Wonder if we should expose the [`selectOption` command](https://modern-web.dev/docs/test-runner/commands/#select-option) to make this kind of thing easier?

bearfriend: I almost did but I had too many things in both repos going simultaneously 😆

dlockhart (approved):

PR #3963, opened 2023-09-05
list-item: color indicator: extend-separators

margaree:


components/list/test/list.visual-diff.html
```diff
@@ -523,6 +523,7 @@
 			<div class="visual-diff" id="draggableSeparatorsExtended">
```
dbatiste, line 523: In comparing to the previous fixtures, is this fixture and the test that uses it named correctly? Also do you think we have the permutations adequately covered? I can't help but feel like we'll be leaning on the visual-diffs pretty heavily for any future tweaks we make.

margaree: I'll take a look at the naming - at a quick glance it's named the same as those around. Yeah I'm just looking at adding more permutations of visual diffs now. There's at least a few more scenarios I want to make sure are covered.

dbatiste (approved):

PR #3967, opened 2023-09-05
US152268 - Bump testing library, causing slight changes in rect size

svanherk:
The changes in https://github.com/BrightspaceUI/testing/pull/159 mean that the rect size is calculated slightly differently.  Specifically, the rounding [here](https://github.com/BrightspaceUI/testing/blob/57191051210921323b52338d091edee15005e513/src/browser/vdiff.js#L37-L40) is different than what `page.screenshot` was doing when we gave it fractions of pixels. This will result in some diffs changing size by a couple pixels.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/8253f016c5b2d2b6fdae6baf9cdf430666732bd4/2023-09-05_20-17-12/report/


dlockhart (approved):

components/inputs/test/input-number.test.js
```diff
@@ -739,8 +718,7 @@ describe('d2l-input-number', () => {
 		});
 
 		it('should handle different locales', async() => {
-			documentLocaleSettings.language = 'fr';
```
svanherk, line 742: @bearfriend @dlockhart Changes to the `testing` library made these tests start failing. I've updated them to set the language through `fixture` instead (and removed the extra waiting that shouldn't be needed anymore). But let me know if this was testing this method of setting the language specifically, and there's actually an issue in `testing`.

components/inputs/test/input-number.test.js
```diff
@@ -55,18 +54,6 @@ const inputWrapperTag = defineCE(
...
-	const elem = await fixture(f);
-	await waitUntil(() => {
-		const textInput = elem.shadowRoot.querySelector('d2l-input-text');
-		if (!textInput) return false;
-		const input = textInput.shadowRoot.querySelector('.d2l-input');
-		if (!input) return false;
-		return true;
-	}, { interval: 10 });
-	return elem;
-}
```
dlockhart, line 68: Sweet! 💥

components/inputs/test/input-number.test.js
dlockhart: Nope this is the perfect change!

dlockhart (approved):

PR #3973, opened 2023-09-06
feat: Icon button: add slot for custom icon

margaree:
[Rally US157379](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F721310111891)

Based on a Slack thread from yesterday, this adds a slot to icon button when no `icon` is present in order to allow for a custom icon.

To do:
- add icon-custom.js import to component-pages-data.js in documentation under button


components/button/demo/button-icon.html
```diff
@@ -7,6 +7,7 @@
 		<script type="module">
 			import '../../demo/demo-page.js';
 			import '../button-icon.js';
+ 			import '../../icons/icon-custom.js';
```
margaree, line 10: I can fix the order here

components/button/test/button-icon.vdiff.js
```diff
@@ -6,7 +6,17 @@ describe('button-icon', () => {
...
+		{ category: 'custom icon', template: html`
+			<d2l-button-icon text="Icon Button">
+				<d2l-icon-custom size="tier1">
+					<svg width="18" height="18" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18 18" mirror-in-rtl="true">
+						<path fill="#494c4e" d="M18 12v5a1 1 0 0 1-1 1H1a1 1 0 0 1-1-1v-5a1 1 0 0 1 2 0v4h14v-4a1 1 0 0 1 2 0z"/>
+						<path fill="#494c4e" d="M13.85 3.15l-2.99-3A.507.507 0 0 0 10.5 0H5.4A1.417 1.417 0 0 0 4 1.43v11.14A1.417 1.417 0 0 0 5.4 14h7.2a1.417 1.417 0 0 0 1.4-1.43V3.5a.47.47 0 0 0-.15-.35zM7 2h1a1 1 0 0 1 0 2H7a1 1 0 0 1 0-2zm4 10H7a1 1 0 0 1 0-2h4a1 1 0 0 1 0 2zm0-4H7a1 1 0 0 1 0-2h4a1 1 0 0 1 0 2z"/>
+					</svg>
+				</d2l-icon-custom>
+			</d2l-button-icon>`
+		}
```
margaree, line 19: Is this the recommended way to format such a large template? The others in this file are all 1 line but that seemed hard to read

components/button/button-icon.js
```diff
@@ -174,7 +176,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				name="${ifDefined(this.name)}"
 				title="${ifDefined(this.text)}"
 				type="${this._getType()}">
-				<d2l-icon icon="${ifDefined(this.icon)}" class="d2l-button-icon"></d2l-icon>
+				${icon}
```
dlockhart, line 179: Another option here would be to always render the slot and just provide the icon as the default content. The way slots work is it'll render the default content unless a slot is provided.

```html
<slot name="icon"><d2l-icon icon="${this.icon}" class="d2l-button-icon"></d2l-icon></slot>
```

We'd probably still want to render `nothing` though in the slot if no `icon` attribute is set.

components/button/button-icon.js
```diff
@@ -155,6 +156,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 	}
 
 	render() {
+		const icon = this.icon ? html`<d2l-icon icon="${this.icon}" class="d2l-button-icon"></d2l-icon>` : html`<slot></slot>`;
```
dlockhart, line 159: Wondering if we want to reserve the default slot for future and name this one `<slot name="icon">`?

components/button/test/button-icon.vdiff.js
dlockhart: Yeah it's more difficult when the tests are defined in a `forEach` like this. The alternative would be to define it up higher in the file and just reference it here, but then you're jumping back and forth to make sense of the test.

components/button/button-icon.js
```diff
@@ -174,7 +176,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 				name="${ifDefined(this.name)}"
 				title="${ifDefined(this.text)}"
 				type="${this._getType()}">
-				<d2l-icon icon="${ifDefined(this.icon)}" class="d2l-button-icon"></d2l-icon>
```
dlockhart, line 177: Hmm, the `.d2l-button-icon` CSS class is forcing a `0.9rem` width and height -- presumably because consumers can use tier2 and tier3 icons here and that would make the button huge. Maybe we want to do something similar? Like adjust the CSS rule to target `d2l-icon` and `d2l-icon-custom` inside the slotted content?

components/button/button-icon.js
margaree: Ah yup good call

components/button/test/button-icon.vdiff.js
svanherk: Yeah I've been using a very informal system of if it's more than a few lines, defining it at the top of the file or creating `createTemplate` helper functions. But personal preference will definitely affect what people decide to do 🤷

components/button/button-icon.js
```diff
@@ -174,7 +176,7 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
...
+				<slot name="icon"><d2l-icon icon="${ifDefined(this.icon)}"></d2l-icon></slot>
```
margaree, line 179: So I've removed the `d2l-button-icon` class. Based on a search this seemed safe but I can put it back if there are concerns.

components/button/button-icon.js
```diff
@@ -13,6 +13,7 @@ import { ThemeMixin } from '../../mixins/theme/theme-mixin.js';
 
 /**
  * A button component that can be used just like the native button for instances where only an icon is displayed.
+ * @slot - Optional slot for a custom icon
```
dlockhart, line 16:
```suggestion
 * @slot icon - Optional slot for a custom icon
```

components/button/button-icon.js
dlockhart: I'm good with that!

Thoughts about just conditionally rendering the `<d2l-icon>`? I'm not sure how the browser does this, but I'd hate for it to quickly render with no icon and then get replaced by the custom one.

Something like:

```javascript
const icon = this.icon ? html`<d2l-icon icon="${this.icon}"></d2l-icon>` : nothing;
return html`
  <button
    ...>
    <slot name="icon">${icon}</slot>
  </button>`;
```

components/button/button-icon.js
```diff
@@ -117,7 +118,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 					box-shadow: var(--d2l-button-icon-focus-box-shadow);
 				}
 
-				.d2l-button-icon {
+				d2l-icon,
+				slot[name="icon"]::slotted(d2l-icon-custom) {
```
dlockhart, line 122: This works eh? Like you don't need `slot[name="icon"]::slotted(d2l-icon)` after it gets re-slotted?

components/button/button-icon.js
margaree: Ah true, makes sense

components/button/button-icon.js
margaree: It so far seems fine but I can update it just in case

dlockhart: To test this out, you could do something like this:

```css
d2l-icon { width: 2em; }
slot[name="icon"]::slotted(d2l-icon) { width: 5em; }
```

And then see which one is getting applied in dev-tools, or see if maybe both are?

margaree: Ah yup only `d2l-icon` gets used. Checked on Chrome, Safari, and Firefox and all use the `d2l-icon` css, the other doesn't seem to get applied at all.

dlockhart: Ok cool so your original code was correct -- sorry for sending you down that path!

margaree: No worries!

dlockhart (approved):
Love how simple this ended up being, nice!

components/button/button-icon.js
```diff
@@ -117,7 +118,8 @@ class ButtonIcon extends ThemeMixin(ButtonMixin(VisibleOnAncestorMixin(RtlMixin(
 					box-shadow: var(--d2l-button-icon-focus-box-shadow);
 				}
 
-				.d2l-button-icon {
+				d2l-icon,
```
dbatiste, line 121: Optionally, we can lock the slot down to prevent people from stuffing things inside it we don't expect with some CSS along these lines:
```css
slot[name="icon"]::slotted(*) {
    display: none;
}
slot[name="icon"]::slotted(d2l-icon-custom) {
    display: inline-block; /* or whatever display value it would normally be */
    height: 0.9rem;
    width: 0.9rem;
}
```
Not sure if locking down was discussed. I think we've done it elsewhere a couple times.

margaree: That's a good idea. Then if there happen to be other things people want to put in there we would be able to evaluate on a case by case basis

dbatiste (approved):

PR #3975, opened 2023-09-06
feat: Subtle Button: Add slot for custom icon

margaree:
[Rally US157379](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F721310111891)

This follows the same general strategy as the icon button PR with some additional changes due to the more complex css.

Subtle button does already have a default slot. I added the visual diff test for the slot content before doing changes, and it stayed the same throughout changes. I think this is overall a safe change but let me know if I've missed something.


components/button/button-subtle.js
```diff
@@ -50,7 +50,9 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 			 * REQUIRED: Text for the button
 			 * @type {string}
 			 */
-			text: { type: String, reflect: true }
+			text: { type: String, reflect: true },
+			_hasCustomIcon: { state: true },
+			_hasIcon: { type: Boolean, reflect: true, attribute: '_has-icon' }
```
dbatiste, line 55: I always feel a bit uncomfortable when we sprout private properties as attributes on the host. I know there are cases where we've done it, particularly when the component doesn't have any intermediate element in the shadowDOM suitable for applying the class, such as when rendering is being done entirely by a mixin. I sort of view the host as being the public API to the consumer, as it's in the light DOM and all the other standard public attributes are on it. Do others agree? To avoid this, a `has-icon` class could be rendered internally on the `button`, thereby keeping this buried in the shadowDOM.

margaree: Agreed, I also don't love it. Good idea, I'll work on switching it out.

components/button/button-subtle.js
```diff
@@ -213,6 +220,19 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
 	}
 
+	willUpdate(changedProperties) {
+		if (changedProperties.has('_hasCustomIcon') || changedProperties.has('icon')) {
+			this._hasIcon = this._hasCustomIcon || this.icon;
+		}
+	}
+
+	_handleIconSlotChange(e) {
+		const icon = e && e.target && e.target.assignedNodes({ flatten: true }).filter((node) => {
```
dbatiste, line 230: Maybe one day we will be able to just use CSS  with the`:hasSlotted(d2l-icon-custom)` or `:has-assigned(d2l-icon-custom)` that's been proposd. https://github.com/w3c/csswg-drafts/issues/6867 😀

components/button/button-subtle.js
```diff
@@ -125,18 +127,19 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 				:host([active]:not([disabled])) button .d2l-button-subtle-content {
 					color: var(--d2l-color-celestine-minus-1);
 				}
-				:host([icon]) .d2l-button-subtle-content,
-				:host([dir="rtl"][icon][icon-right]) .d2l-button-subtle-content {
+				:host([_has-icon]) .d2l-button-subtle-content,
+				:host([dir="rtl"][_has-icon][icon-right]) .d2l-button-subtle-content {
 					padding-left: 1.2rem;
 					padding-right: 0;
```
dlockhart, line 133: We've been slowly adopting CSS logical properties on new things... since they would simplify these changes (and possibly remove the need for `RtlMixin`) it could be a good time to do that here. Thoughts?

components/button/button-subtle.js
```diff
@@ -204,7 +211,7 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 				id="${this._buttonId}"
 				name="${ifDefined(this.name)}"
 				type="${this._getType()}">
-				${icon}
+				<slot name="icon" @slotchange="${this._handleIconSlotChange}"><d2l-icon icon="${ifDefined(this.icon)}" class="d2l-button-subtle-icon"></d2l-icon></slot>
```
dlockhart, line 214: I still like conditionally rendering the `<d2l-icon>` vs. `nothing` if they don't provide an `icon` property. Technically I think subtle buttons could exist in the wild with no icon specified, and I'd hate to render a `<d2l-icon>` that did nothing in those cases.

components/button/button-subtle.js
```diff
@@ -213,6 +220,19 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
+		if (changedProperties.has('_hasCustomIcon') || changedProperties.has('icon')) {
+			this._hasIcon = this._hasCustomIcon || this.icon;
+		}
+	}
+
+	_handleIconSlotChange(e) {
+		const icon = e && e.target && e.target.assignedNodes({ flatten: true }).filter((node) => {
+			return node.nodeType === Node.ELEMENT_NODE && node.tagName === 'D2L-ICON-CUSTOM';
+		});
+		this._hasCustomIcon = icon.length === 1;
```
dlockhart, line 233: I _think_ everything can be simplified a bit to just require a single `_hasCustomIcon` state property? If you apply a CSS class in `render()` directly to the button as Dave suggested when `_hasCustomIcon || this.icon` is true, then I don't think you need the `willUpdate` code or the `_hasIcon` property. Could be missing something though.

components/button/button-subtle.js
margaree: Agreed! I'll update this to be done the same as button-icon

components/button/button-subtle.js
margaree: Yup I think that's right

components/button/button-subtle.js
dlockhart: The details in the note there says that Safari <15 only doesn't support the shorthand ones when the value is a CSS property. SO in this particular case, it would be fine. Safari 14 is 3 years old at this point as well.

components/button/button-subtle.js
```diff
@@ -72,36 +72,31 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
 					background-color: transparent;
 					border-color: transparent;
 					font-family: inherit;
-					padding: 0 var(--d2l-button-subtle-padding-inline-end) 0 var(--d2l-button-subtle-padding-inline-start);
+					padding-bottom: 0;
+					padding-inline-start: var(--d2l-button-subtle-padding-inline-start);
+					padding-inline-end: var(--d2l-button-subtle-padding-inline-end);
+					padding-top: 0;
```
dlockhart, line 78: Might as well go all-in!

```suggestion
					padding-block-end: 0;
					padding-block-start: 0;
					padding-inline-start: var(--d2l-button-subtle-padding-inline-start);
					padding-inline-end: var(--d2l-button-subtle-padding-inline-end);
```

components/button/button-subtle.js
```diff
@@ -72,36 +72,31 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
 					position: relative;
 				}
 
-				:host([dir="rtl"]) button {
-					padding: 0 var(--d2l-button-subtle-padding-inline-start) 0 var(--d2l-button-subtle-padding-inline-end);
-				}
```
dlockhart, line 79: 🎉

components/button/button-subtle.js
```diff
@@ -70,36 +72,31 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
 					--d2l-button-subtle-padding-inline-end: 0.4rem;
 				}
 
 				:host([h-align="text"]) button {
-					left: calc(var(--d2l-button-subtle-padding-inline-start) * -1);
-				}
-				:host([dir="rtl"][h-align="text"]) button {
-					left: 0;
-					right: calc(var(--d2l-button-subtle-padding-inline-start) * -1);
+					inset-inline: calc(var(--d2l-button-subtle-padding-inline-start) * -1) 0;
```
dlockhart, line 99: `inset-inline-start` is probably more analogous to `left` and would mean you don't need to have the `0` value? 🤷 

```suggestion
					inset-inline-start: calc(var(--d2l-button-subtle-padding-inline-start) * -1);
```

components/button/button-subtle.js
```diff
@@ -125,20 +122,21 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
-				:host([icon][icon-right]) .d2l-button-subtle-content {
-					padding-left: 0;
-					padding-right: 1.2rem;
+				:host([icon-right]) .d2l-button-subtle-has-icon .d2l-button-subtle-content {
+					padding-inline: 0 1.2rem;
 				}
 
-				d2l-icon.d2l-button-subtle-icon {
-					color: var(--d2l-color-celestine);
+				d2l-icon.d2l-button-subtle-icon,
```
dlockhart, line 132: I _think_ applying `display: none` to the `d2l-icon` isn't necessary since it is and was conditionally rendered based on the precense of the `icon` property. Further down it then gets `display: inline-block` which could have just been applied from the start?

components/button/button-subtle.js
```diff
@@ -148,19 +146,20 @@ class ButtonSubtle extends ButtonMixin(RtlMixin(LitElement)) {
...
 				}
-				:host([icon][icon-right]) d2l-icon.d2l-button-subtle-icon {
-					right: var(--d2l-button-subtle-padding-inline-end);
-				}
-				:host([dir="rtl"][icon][icon-right]) d2l-icon.d2l-button-subtle-icon {
-					left: var(--d2l-button-subtle-padding-inline-end);
-					right: auto;
+				:host([icon-right]) .d2l-button-subtle-has-icon d2l-icon.d2l-button-subtle-icon,
+				:host([icon-right]) slot[name="icon"]::slotted(d2l-icon-custom) {
+					inset-inline: auto var(--d2l-button-subtle-padding-inline-end);
```
dlockhart, line 162: Could this just be this or is the `auto` needed?

```suggestion
					inset-inline-end: var(--d2l-button-subtle-padding-inline-end);
```

margaree: Seems like it does just work fine without it - will remove

dlockhart (approved):

dbatiste (approved):

PR #3979, opened 2023-09-07
US152268 - Migrate collapsible-panel & expand-collapse vdiff tests

svanherk:
A few expected vdiff changes with this one, I'll comment on the goldens PR.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/e7c33afa902ab2e1752d035aa41fa6a3a43f39af/2023-09-07_12-53-11/report/


components/collapsible-panel/test/collapsible-panel-group.vdiff.js
```diff
@@ -0,0 +1,60 @@
...
+		].forEach(({ name, type }) => {
+			it(`${name}${rtl ? '-rtl' : ''}`, async() => {
+				const elem = await fixture(html`
+					<div style="padding: 1rem; width: 600px;">
+						<d2l-collapsible-panel-group>
+							${createCollapsiblePanel({ type })}
+							${createCollapsiblePanel({ type, summary: true })}
+							${createCollapsiblePanel({ type })}
+						</d2l-collapsible-panel-group>
+					</div>
```
dlockhart, line 33: Very nit, but curious if the surrounding `<div>` could be removed? The page should already have enough padding and we could tweak the viewport size to match the old `600px` value (although it would likely need to be `600px` - `2rem`).

components/collapsible-panel/test/collapsible-panel.vdiff.js
```diff
@@ -0,0 +1,163 @@
+import '../../button/button-icon.js';
+import '../../link/link.js';
+import '../../status-indicator/status-indicator.js';
+import '../../typography/typography.js';
```
dlockhart, line 4: The need for and application of the `d2l-typography` CSS class is a bit weird IMO. Wondering what that's doing...

dlockhart (approved):

components/collapsible-panel/test/collapsible-panel.vdiff.js
svanherk: It's for lines 61 and 89 (`<p class="d2l-body-small">...</p>`).  I could instead just set some styles directly to try to get "close enough"

dlockhart: I see. It's fine for the migration, but we probably want to avoid this kind of vdiff component cross-contamination in the future unless it's testing something very specific and common (e.g. different button components in overflow-group).

svanherk: Yeah agreed, my focus has been migrating exactly what's there but an audit wouldn't be a bad idea at some point

dlockhart (approved):

components/collapsible-panel/test/collapsible-panel-group.vdiff.js
svanherk: Got it working lol.  I need a viewport width of 676 and a margin of 30

dlockhart (approved):
Haha sorry that ended up being more fiddly that I was expecting -- looks good though. We could even follow up later and just set those back to the default viewport & margins since I don't think it'll affect too much.

PR #3983, opened 2023-09-07
fix: Subtle Button: Fix slot content rtl behaviour with icon

margaree:


components/button/button-subtle.js
```diff
@@ -202,8 +202,10 @@ class ButtonSubtle extends ButtonMixin(LitElement) {
 				name="${ifDefined(this.name)}"
 				type="${this._getType()}">
 				<slot name="icon" @slotchange="${this._handleIconSlotChange}">${icon}</slot>
-				<span class="d2l-button-subtle-content">${this.text}</span>
-				<slot></slot>
+				<span class="d2l-button-subtle-content-wrapper">
+					<span class="d2l-button-subtle-content">${this.text}</span>
+					<slot></slot>
+				</span>
```
margaree, line 208: `.d2l-button-subtle-content` also adds color to its content, which I didn't want to assume slot content would want, but it does make sense for it to also have the padding when there's an icon so I split that out. Let me know thoughts on this/alternative ideas.

There is the risk that someone could have already been handling this scenario and this would be adding additional padding.

dlockhart: I think what you have makes sense. It wouldn't be a very easy code search to find places using the slot...

dlockhart (approved):

PR #3985, opened 2023-09-07
list-item: color indicator: fix alignment when selectable

margaree:
For the case where a list item has a color, is selectable, and expandable we want to align the color indicator with the checkbox.

This also fixes the large amount of spacing between the expand/collapse icon and checkbox when extend-separators is true.


dbatiste (approved):
This looks good to me.

PR #3986, opened 2023-09-07
US152268 - Migrate calendar vdiff tests

svanherk:
Putting this up as a draft PR now, but switching of the language is still broken and needs to be fixed.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/fb38fc32295b9e57bf0b688111eb22f380269d4e/2023-09-08_17-55-03/report/


components/calendar/calendar.js
```diff
@@ -29,15 +29,13 @@ const keyCodes = {
...
-		calendarData = {};
-		calendarData.descriptor = getDateTimeDescriptorShared(refresh);
-		calendarData.firstDayOfWeek = calendarData.descriptor.calendar.firstDayOfWeek;
-		calendarData.daysOfWeekIndex = [];
-		for (let i = calendarData.firstDayOfWeek; i < calendarData.firstDayOfWeek + daysInWeek; i++) {
-			calendarData.daysOfWeekIndex.push(i % daysInWeek);
-		}
+function getCalendarData() {
+	calendarData = {};
+	calendarData.descriptor = getDateTimeDescriptor();
```
dlockhart, line 34: The culprit was actually two separate places where we were saving local caches -- `calendarData` here and in `getDateTimeDescriptorShared`.

Previously the caches were invalidated in calendar's event handler for `d2l-localize-resources-change`, but because of the isolated fixtures that event fires before any calendars are even put on the page.

I'm going to separately look into adding a cache inside the `@brighspace-ui/intl` library, since it's in a much better place to invalidate it.

helpers/dateTime.js
```diff
@@ -139,12 +139,6 @@ export function getDateNoConversion(value) {
 	return new Date(parsed.year, parsed.month - 1, parsed.date, parsed.hours, parsed.minutes, parsed.seconds);
 }
 
-let dateTimeDescriptor = null;
-export function getDateTimeDescriptorShared(refresh) {
```
dlockhart, line 143: I just noticed that [Nova is using this](https://github.com/Brightspace/nova/blob/main/src/app/main/views/activities/view-activity.js#L25) -- I'll switch that before merging this.

dlockhart (commented):
Should be all fixed up now -- added some explanations inline.

dlockhart (approved):

PR #4001, opened 2023-09-12
New Window / Tab Links

NicholasHallman:
Implementation Notes

Standardize the "opens in a new window" linking pattern by setting approriate defaults for
 - icon color
 - icon sizing
 - aria label
 - target

Quality Notes
 - Unit Tests
   - [x] new-window adds an icon
   - [x] new-window sets the target if it doesn't exist
   - [x] new-window adds an aria-label message for screen readers
 - Manual tests
   - [x] aria-hidden prevets screen reader from reading out "graphic"
   - [x] aria-label is read by screen reader "link, opens in a new window."
 - Vis Diffs
   - [x] standard link with icon
   - [x] small link with icon

new window links


components/link/link.js
```diff
@@ -139,13 +156,33 @@ class Link extends FocusMixin(LitElement) {
...
+	}
+
+	renderNewWindowIcon() {
+		const newWindowClasses = {
+			'd2l-new-window': true,
+			'd2l-icon-small': this.small,
+		};
+		return this.newWindow ? html`
+			<d2l-icon class="${classMap(newWindowClasses)}" aria-hidden="true" icon="tier1:new-window"></d2l-icon>
+		` : html``;
```
dlockhart, line 181: I think what we'll want to do instead of modifying `ariaLabel` is to just add "Opens in a new window" as offscreen text inside the link. That will essentially have the same effect.

Something like (you'll need to import the offscreen styles):
```suggestion
		return this.newWindow ? html`
			<d2l-icon class="${classMap(newWindowClasses)}" aria-hidden="true" icon="tier1:new-window"></d2l-icon>
			<span class="d2l-offscreen">this.localize('components.link.open-in-new-window')</span>
		` : nothing;
```

components/link/link.js
```diff
@@ -125,6 +141,7 @@ class Link extends FocusMixin(LitElement) {
 		this.main = false;
 		this.small = false;
 		this.lines = 0;
+		this.showOpenInNewWindow = false;
```
dlockhart, line 144: Doesn't appear to be used. Probably mean to do this?

```suggestion
		this.newWindow = false;
```

components/link/link.js
```diff
@@ -88,7 +99,12 @@ class Link extends FocusMixin(LitElement) {
 			 * Where to display the linked URL
 			 * @type {string}
 			 */
-			target: { type: String }
+			target: { type: String },
+			/**
+			 * Whether to display the open in new window icon
+			 * @type {boolean}
+			 */
+			newWindow: { type: Boolean, attribute: 'new-window' }
```
dlockhart, line 107: What are folks' thoughts about NOT making the opt-in? Like essentially just showing the indicator on all `<d2l-link>`s whenever `target="_blank">`. A quick code search only yields ~30 places using links with blank targets. We could do a quick audit to confirm that they'd all want this functionality.

It would dramatically simplify this code, but also the main benefit would be immediate UX/a11y gains for users.

components/link/link.js
```diff
@@ -44,13 +45,23 @@ export const linkStyles = css`
 			color: var(--d2l-color-ferrite);
 		}
 	}
+	d2l-icon.d2l-new-window {
+		color: var(--d2l-color-celestine);
+		margin-top: -7px;
```
dlockhart, line 50: I have a feeling these negative margin values are working nicely for our default font size, but may start to fall apart when users choose larger/smaller ones in account settings and/or use browser zoom.

Typically a better solution would be to apply flexbox to the container (the host in this case) and use that to vertically centre everything. That's not feasible here though as it'll break everyone's stuff. You could maybe try `rem`-based negative margins instead?

components/link/link.js
```diff
@@ -44,13 +45,23 @@ export const linkStyles = css`
 			color: var(--d2l-color-ferrite);
 		}
 	}
```
dlockhart, line 47: This CSS gets applied to native `<a class="d2l-link">` as well that consumers are using. Since they don't understand the "new window" stuff, any new-window-specific CSS can go down below in the CSS that's just used for the `<d2l-link>` web component.

components/link/link.js
NicholasHallman: Even if we don't show the icon, forcing the hidden is span whern target="_blank" is an immediate accessibility benefit

components/link/link.js
NicholasHallman: I fixed this by changing the vertical-align, setting it to inherti the text allignment worked perfectly for all the font size I tested with

components/link/link.js
```diff
@@ -115,6 +122,15 @@ class Link extends FocusMixin(LitElement) {
 					overflow: hidden;
 					overflow-wrap: anywhere;
 				}
+				d2l-icon.d2l-new-window {
+					color: var(--d2l-color-celestine);
+					vertical-align: inherit;
+				}
+
+				d2l-icon.d2l-new-window.d2l-icon-small {
```
dlockhart, line 130: CSS attribute selector here should remove the need for `classMap`:

```suggestion
				:host([small]) d2l-icon.d2l-new-window {
```

components/link/link.js
```diff
@@ -139,13 +156,31 @@ class Link extends FocusMixin(LitElement) {
 			'truncate': this.lines > 0
 		};
 		const styles = (this.lines > 0) ? { '-webkit-line-clamp': this.lines } : {};
+		if (this.newWindow && this.target === undefined) {
+			this.target = '_blank';
+		}
```
dlockhart, line 161: Let's not actually modify the target property, since that's mutating something that isn't really this code's responsibility and will also trigger another render.

```suggestion
		const target = (this.newWindow && this.target === undefined) ? '_blank' : this.target;
```

components/link/link.js
```diff
@@ -139,13 +156,31 @@ class Link extends FocusMixin(LitElement) {
...
+				target="${ifDefined(this.target)}"><slot></slot>${this.renderNewWindowIcon()}</a>`;
+	}
+
+	renderNewWindowIcon() {
+		const newWindowClasses = {
+			'd2l-new-window': true,
+			'd2l-icon-small': this.small,
+		};
+		return this.newWindow ? html`
+			<d2l-icon class="${classMap(newWindowClasses)}" aria-hidden="true" icon="tier1:new-window"></d2l-icon>
```
dlockhart, line 177: `aria-hidden` isn't needed since `<d2l-icon>` doesn't actually render anything that screen readers will notice.
```suggestion
			<d2l-icon class="${classMap(newWindowClasses)}" icon="tier1:new-window"></d2l-icon>
```

components/link/link.js
```diff
@@ -4,6 +4,8 @@ import { classMap } from 'lit/directives/class-map.js';
 import { FocusMixin } from '../../mixins/focus/focus-mixin.js';
 import { getFocusPseudoClass } from '../../helpers/focus.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
+import { offscreenStyles } from '../offscreen/offscreen.js';
```
dlockhart, line 8: Make sure `../icons/icon.js` is imported.

components/link/link.js
```diff
@@ -139,13 +156,31 @@ class Link extends FocusMixin(LitElement) {
...
 				aria-label="${ifDefined(this.ariaLabel)}"
 				class="${classMap(linkClasses)}"
 				style="${styleMap(styles)}"
 				?download="${this.download}"
 				href="${ifDefined(this.href)}"
-				target="${ifDefined(this.target)}"><slot></slot></a>`;
+				target="${ifDefined(this.target)}"><slot></slot>${this.renderNewWindowIcon()}</a>`;
+	}
+
+	renderNewWindowIcon() {
```
dlockhart, line 171: Once the `classMap` is removed, this should be simple enough that I'd suggest just moving it right into `render()`.

components/link/link.js
```diff
@@ -139,13 +156,31 @@ class Link extends FocusMixin(LitElement) {
...
+		const newWindowClasses = {
+			'd2l-new-window': true,
+			'd2l-icon-small': this.small,
+		};
+		return this.newWindow ? html`
+			<d2l-icon class="${classMap(newWindowClasses)}" aria-hidden="true" icon="tier1:new-window"></d2l-icon>
+			<span class="d2l-offscreen">${this._newWindowLangTerm}</span>
```
dlockhart, line 178: I don't think the extra getter for the lang term is necessary.
```suggestion
			<span class="d2l-offscreen">${this.localize('components.link.open-in-new-window')}</span>
```

lang/en.js
```diff
@@ -66,6 +66,7 @@ export default {
 	"components.input-time-range.errorBadInput": "{startLabel} must be before {endLabel}",
 	"components.input-time-range.startTime": "Start Time",
 	"components.interactive.instructions": "Press Enter to interact, Escape to exit",
+	"components.link.open-in-new-window": "Opens in a new window.",
```
dlockhart, line 69: This should be copied into all the other language files otherwise the validator is going to complain (see all its angry  comments below).

components/link/link.js
NicholasHallman: Narrator on windows says
"Link, Graphics, New Window Link, Opens in a new window." When the aria-hidden isn't included

With aira-hiddend it says
"Link, New Window Link" [Down Arrow Key] "Opens in a new window"

Personally I'd rather it say "graphic" and notify the user its a new window link all in one sentence than the aria-hidden version which breaks it up

components/link/link.js
dlockhart: Here's a list of all the places this would affect:

Definitely require action:
- [Insights Assessment Quality Dashboard](https://github.com/Brightspace/insights-assessment-quality-dashboard/blob/46bbb294bfc726e1bcd7e96a763416430aa145ad/pages/quiz-item-analysis-view.js#L481): has a "go to quiz" link and this is actually already rendering the "new window" icon. We'd need to ensure this got cleaned up to avoid 2 icons. It ALSO has an `aria-label` which has text indicating it opens in a new window, so that would need to be removed.
- [Wave](https://github.com/Brightspace/nova/blob/92a0fdb3f944fbc2092eca4ffc95ee10549223d9/src/app/main/components/applications/application-overview.js#L311): is also rendering the "new window" icon itself

We should confirm with teams that they're OK with this:
- [Parent Portal 1](https://github.com/Brightspace/parent-portal-ui/blob/c0ea044a9ece75c883f20afdb49bb89e3b7d2152/src/pages/no-children.js#L67): support text with "click here" link.
- [Parent Portal 2](https://github.com/Brightspace/parent-portal-ui/blob/c0ea044a9ece75c883f20afdb49bb89e3b7d2152/src/components/parent-teacher-interview-alert.js#L82): parent-teacher interview link
- [User Profile Card](https://github.com/BrightspaceHypermediaComponents/users/blob/6fedf39419f95be319b39590110215af217342d7/components/user-profile-card/user-profile-card.js#L505): for showing a user's website link
- [Import Quiz Questions 1](https://github.com/Brightspace/import-quiz-questions/blob/bb7e9c6cb788dc4dd0846117f21cdb389b781ec5/src/components/FileUpload/SupportedFormatsInfo.jsx#L10): link to download a CSV template. There's also a test doing this that'll need updating.
- [Import Quiz Questions 2](https://github.com/Brightspace/import-quiz-questions/blob/bb7e9c6cb788dc4dd0846117f21cdb389b781ec5/src/components/ErrorPage/ErrorPage.jsx#L64): wouldn't be impacted directly as it's using JavaScript to open the new window. However, it's manually rendering the "new window" icon so might benefit from being able to force it ON even if the target isn't set.
- [IPSIS Admin UI](https://github.com/Brightspace/ipsis-admin-ui/blob/3ffe3d024bc9657bc9a3889e1b37acbf8f0cc295/src/pages/dashboard-page.js#L302): "learn more" link to the Community -- probably the reason we're doing this?
- [Custom My Courses in Other Orgs](https://github.com/Brightspace/custom-my-courses-in-other-orgs/blob/b1f63178c6b6102caeaa68147ce2c1ea6a3cb044/custom-my-courses-in-other-orgs.js#L201): opens a link to an API endpoint
- [LCS Instructor Check-in Widget](https://github.com/Brightspace/lcs-aau-instructor-check-in-widget/blob/2f198cac337391ede2cf79fc85952c1c1386a568/src/shared/custom_widgets/instructor_check_in_widget/components/lcs-instructor-check-in-widget.js#L546): rendering a "get help" link and also a help icon
- [LCS core widgets](https://github.com/Brightspace/lcs-core-widgets/blob/079b08a208c8f7474765319e21be6eb70c16c430/code_le/src/lcs-cdn/widgets/profiles/X.X.X/components/admin-edit-student-view-profile.js#L500): link to student profile -- also rendering its own "opens in new window" offscreen text but no icon

Internal tools so we don't care:
- [AWS Jumpbox](https://github.com/Brightspace/aws-jumpbox/blob/e75417e29c12867a3e0cd2619d00d79f018525b0/web/src/index.html#L105)
- [Tenant Finder](https://github.com/Brightspace/tenant-finder.d2l.dev/blob/586774efe5c523a7e949b9a72a85c5bffd6fef7d/src/components/tenant-finder.js#L149)
- [CDS Catalogue](https://github.com/Brightspace/cds-catalogue/blob/f1a092632d4c444e2b74fc0fc4f5a6cc5ed476c3/src/components/cds-catalogue-modal.js#L38): innovation project
- [Instance Metadata Store](https://github.com/Brightspace/instance-metadata-store/blob/315e90ffcc0e3a05e885fc5c843e13ab671da059/ui/public/index.html#L40)
- [LCS webdev template](https://github.com/Brightspace/lcs-webdev-template/blob/f187d7968e9617637864ff384d3bd2571683ff78/src/components/d2l-component-collection.js#L170)
- [CDS Template UI Demo](https://github.com/Brightspace/cds-template-ui-demo/blob/4cd3c41a7cdc014453a45bba8322dc5a8b343b17/components/test-component.js#L40)
- [Test Reporting UI](https://github.com/Brightspace/test-reporting.d2l.dev/blob/ee3e92dbe1bdc2d9d5d2fc09efdb88544c290506/ui/src/components/test-reporting-github-actions-link.js#L32)

components/link/link.js
dlockhart: Interesting, I stand corrected. That's surprising though... I've never noticed this in other AT as we have `<d2l-icon>`s on their own a lot. I'd lean towards not putting `aria-hidden` on there just for consistency.

NicholasHallman:
I've asked Jaya and Abi if I should remove or change the color of the new window icon when the link is printed

components/link/link.js
```diff
@@ -139,13 +157,27 @@ class Link extends FocusMixin(LitElement) {
...
-				target="${ifDefined(this.target)}"><slot></slot></a>`;
+				target="${ifDefined(target)}"
+			>
+				<slot></slot>
+				${target === '_blank' ? html`
+					<span class="d2l-offscreen">${this.localize('components.link.open-in-new-window')}</span>
+				` : nothing}
+				${this.newWindow ? html`
+					<d2l-icon class="d2l-new-window" icon="tier1:new-window"></d2l-icon>
+				` : nothing}
```
bearfriend, line 179: This will have to be something like this (or some better less annoying method I don't know about), so that the icon can't wrap to the next line by itself.
```suggestion
				<span style="white-space: nowrap;">
					<span style="white-space: normal;"><slot></slot></span>${this.newWindow ? html`&nbsp;<d2l-icon class="d2l-new-window" icon="tier1:new-window"></d2l-icon>
					` : nothing}
					${target === '_blank' ? html`
						<span class="d2l-offscreen">${this.localize('components.link.open-in-new-window')}</span>
					` : nothing}
				</span>
```
So we get this:
Screenshot 2023-09-13 at 3 25 54 PM

Instead of this:
Screenshot 2023-09-13 at 3 26 17 PM

dlockhart: Oooh, good call. @NicholasHallman can you add a vdiff that covers that case to make sure we don't accidentally break it in the future?

bearfriend: Actually, I think this entire thing has to be one line, otherwise any trailing implicit space is included in the link.

Screenshot 2023-09-13 at 3 44 12 PM

vs

Screenshot 2023-09-13 at 3 43 57 PM

NicholasHallman: Your solution works and I couldn't find anything simpler so I've commited it, thanks!

NicholasHallman: > Actually, I think this entire thing has to be one line, otherwise any trailing implicit space is included in the link.

The only way I've found to fix this is to add `display: inline-block` to the `<a>` tag. Any other solution for this?

dlockhart:
Great, thanks! I'm just working on converting the link visual-diff tests to the new library today. Once I've merged that, I'll rebase this and fix it up.

components/link/link.js
```diff
@@ -139,13 +163,24 @@ class Link extends FocusMixin(LitElement) {
...
+
 		return html`<a
 				aria-label="${ifDefined(this.ariaLabel)}"
 				class="${classMap(linkClasses)}"
 				style="${styleMap(styles)}"
 				?download="${this.download}"
 				href="${ifDefined(this.href)}"
-				target="${ifDefined(this.target)}"><slot></slot></a>`;
+				target="${ifDefined(target)}"
+			><span style="white-space: nowrap;"><span style="white-space: normal;"><slot></slot></span>${newWindowIndicator}${newWindowMessage}</span></a>`;
```
dlockhart, line 183: I had to tighten up the whitespace here because it was introducing a lot of spaces every time there was a newline -- one of the visual diff tests caught it.

components/link/test/link.test.js
```diff
@@ -92,4 +92,11 @@ describe('d2l-link', () => {
 
 	});
 
+	describe('new-window', () => {
+		it('should add offscreen text', async() => {
+			const elem = await fixture(html`<d2l-link new-window>link text</d2l-link>`);
+			expect(elem.shadowRoot.querySelector('.d2l-offscreen').innerText).to.equal('Opens in a new window.');
+		});
```
dlockhart, line 99: I removed the other 2 tests as they were redundant with the vdiffs.

components/link/test/link.vdiff.js
```diff
@@ -21,6 +21,9 @@ describe('d2l-link', () => {
 		{ name: 'wc-standard', template: html`<d2l-link href="https://www.d2l.com">Standard Link</d2l-link>` },
 		{ name: 'wc-main', template: html`<d2l-link href="https://www.d2l.com" main>Main Link</d2l-link>` },
 		{ name: 'wc-small', template: html`<d2l-link href="https://www.d2l.com" small>Small Link</d2l-link>` },
+		{ name: 'wc-standard-new-window', template: html`<d2l-link new-window href="https://www.d2l.com">Standard New Window Link</d2l-link>` },
+		{ name: 'wc-small-new-window', template: html`<d2l-link new-window small href="https://www.d2l.com">Small New Window Link</d2l-link>` },
+		{ name: 'wc-standard-new-window-overflow', template: html`<div style="width: 240px"><d2l-link new-window href="https://www.d2l.com">Standard New Window Link</d2l-link></div>` },
```
dlockhart, line 26: Migrated these to the new library.

components/link/link.js
```diff
@@ -88,12 +91,12 @@ class Link extends FocusMixin(LitElement) {
 			 * Where to display the linked URL
 			 * @type {string}
 			 */
-			target: { type: String }
+			target: { type: String },
```
dlockhart, line 94:
```suggestion
			target: { type: String }
```

components/link/link.js
```diff
@@ -125,6 +145,7 @@ class Link extends FocusMixin(LitElement) {
 		this.main = false;
 		this.small = false;
 		this.lines = 0;
+		this.newWindow = false;
```
dlockhart, line 148:
```suggestion
```

components/link/link.js
```diff
@@ -115,6 +118,23 @@ class Link extends FocusMixin(LitElement) {
...
+
+				a:hover d2l-icon.d2l-new-window {
+					--d2l-icon-fill-color: var(--d2l-color-celestine-minus-1);
+				}
+
+				@media print {
+					d2l-icon.d2l-new-window {
+						display: none;
+					}
+				}
```
dlockhart, line 137: Now that this is conditionally rendered always, I think we can ditch the `d2l-new-window` CSS class?

NicholasHallman: I added the class because when other icons were placed in the slot it would get colored blue (which may not be wanted), maybe now that it's part of the scoped styles that won't happen though?

NicholasHallman: Tested this and it works fine now, other icons in the slot are still grey

dlockhart: Ok excellent... yeah that would surprise me -- I'd only expect them to be affected if `::slotted(d2l-icon)` were used.

components/tooltip/tooltip.js
```diff
@@ -991,6 +991,13 @@ class Tooltip extends RtlMixin(LitElement) {
 		target.appendChild(cloneContainer);
 		await this.updateComplete;
 
+		// if the clone is a web component it needs to update to fill in any slots
+		const customElm = customElements.get(clone.localName);
+		if (customElm !== undefined) {
+			clone.requestUpdate();
+			await clone.updateComplete;
+		}
+
```
svanherk, line 1000: @NicholasHallman Are you seeing that this is only a problem because we're focusing on the link too quickly in the test? Or doesn't matter how long we wait, this is still needed?  Just wondering if the new `fixture` the vdiffs use would fix it, which wait for all components inside to update.

NicholasHallman: I thought that might be the case but it wasn't working in the d2l-tooltip demo either. The element wouldn't contain anything inside the d2l-link so it had 0 width and never thought it needed to be truncated

dlockhart (approved):
When you merge, make sure you squash with `feat: link new window indicator` in the commit so it does a feature release.

PR #4007, opened 2023-09-13
Migrate table vdiffs

dlockhart:
[Link to the report](https://vdiff.d2l.dev/BrightspaceUI/core/0b7f70ba756daed6edd4aa26ab9c45498487f988/2023-09-13_21-15-24/report/).

Leaving as draft as there are a few outstanding issues.

Most of the diffs you'll see are correct but resized. This is because previously the entire viewport was captured whereas now it's limited to the table. These are smaller images which are better and IMO these diffs are fine.

Another class of resized diffs result from different margins around the table. Previously [there was code](https://github.com/BrightspaceUI/core/blob/main/components/table/test/table.visual-diff.js#L7) that would adjust the margins based on the `offsetLeft`/`offsetTop` of the table. I'm no longer doing that.

The main problem I'm seeing is the behaviour when scrolling, especially with sticky columns/rows. In these cases previously, the entire viewport was being captured successfully. Now, even if I use the new `viewport` feature (not shown in this code), things get clipped by the amount of scroll. So if it scrolls to the right by `100px`, `100px` gets clipped off on the right side.

This problem seems to get even worse in RTL, where things get both clipped AND a white overlay on top. If I disable sticky headers, the issue goes away.

Lastly, I'm seeing some concerning flake in some of the scroll-wrapper stuff where the placement of the dashed lines pixels can vary from run to run. I might try converting `<d2l-scroll-wrapper>` and seeing if the same issue comes up there. 😟


components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,1229 @@
...
+	['ltr', 'rtl'].forEach((dir) => {
+		describe(dir, () => {
+			['default', 'light'].forEach((type) => {
+
+				async function createFixture(markup, opts) {
+					const rtl = (dir === 'rtl');
+					const tag = defineCE(
+						class extends LitElement {
+							static get styles() {
+								return [tableStyles];
```
dlockhart, line 39: Table styles need to be applied from a Lit component, which is why there were so many test components before. Now we can just use `defineCE`, but it means these all need to go through this wrapper.

components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,1229 @@
...
+								const wrapper = html`
+									<d2l-table-wrapper
+										?no-column-border="${opts?.noColumnBorder}"
+										?sticky-headers="${opts?.stickyHeaders}"
+										style="--d2l-input-position: static;"
+										type="${type}">${markup}</d2l-table-wrapper>`;
+								if (!opts?.bottomMargin) {
+									return wrapper;
+								}
+								return html`<div style="margin-bottom: 1000px;">${wrapper}</div>`;
```
dlockhart, line 51: This is so that there's room at the bottom to scroll. I'm going to play around with this and look into applying right/left margin as well as maybe that's the source of some of the scrolling diff issues.

components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,1229 @@
...
+											<td>Cell 3-B</td>
+											<td>Cell 3-C</td>
+										</tr>
+									</tbody>
+								</table>
+							`);
+							await expect(elem).to.be.golden();
+						});
+
+						it('selected-top-bottom', async() => {
```
dlockhart, line 363: There's a lot of repetition between the fixtures, where only 1 or 2 things change between thing -- like in these "selected" tests. It was quickest/easiest to just copy these in for now, but there are definitely opportunities to build some functions that take parameters to create them.

components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,1229 @@
...
+														?sticky-headers="${condition1.stickyHeaders}"
+														type="${type}"
+														?visible-background="${condition1.visibleBackground}">
+													</d2l-test-table>
+												</div>
+											`,
+											{ rtl: dir === 'rtl', viewport: { width: 500 } }
+										);
+										elem.scrollTo(0, condition2.scroll);
+										await nextFrame();
```
dlockhart, line 1202: These are the ones where I'm seeing flake with the scroll wrapper dashed borders. The `nextFrame` was a failed attempt to fix them.

components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,1229 @@
...
+			</tr>
+		</tbody>
+	</table>
+`;
+
+describe('d2l-table', () => {
+
+	['ltr', 'rtl'].forEach((dir) => {
+		describe(dir, () => {
+			['default', 'light'].forEach((type) => {
```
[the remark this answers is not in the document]
dbatiste: Agreed. I think we should just do targeted tests for RTL or responsive cases that are "interesting".  I know that means busting out some of the tests so they aren't all neatly done in these nested loops though. Like maybe have a subset of tests that focuses on layout, and those are done with RTL, and then treat the light and dark as a separate set of tests. I

I'm not sure the best way to organize them, or whether we can choose a way to organize them the same across components. List has a mind boggling number of permutations and there's probably a dozen different ways the tests could be organized, and I struggle to decide which way is best.

svanherk:
Going to close this PR - replaced by https://github.com/BrightspaceUI/core/pull/4059 and https://github.com/BrightspaceUI/core/pull/4102

PR #4010, opened 2023-09-14
fix: list-item: color indicator: resize to 3px width when parent list width is less than 400px

margaree:
Also fix a spacing issue when extend separators when there is color indicator and checkbox but no expand/collapse

[Rally story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F721347694203)


components/list/list-item-mixin.js
```diff
@@ -512,6 +522,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 	}
 
 	resizedCallback(width) {
+		if (width < SLIM_COLOR_BREAKPOINT) this._slimColor = true;
+		else this._slimColor = false;
```
margaree, line 526: This does tap into the existing resize observer on the list item, which I know we want to change. I have a story for doing that work and I have time to do it October 2, so I'm wondering if it's okay to get this piece in now for color indicator and move this around later as we move the resize observer, or if we want to wait on this until that all is done?

dbatiste: These two lines could also be reduced to:
```javascript
this._slimColor = (width < SLIM_COLOR_BREAKPOINT);
```

dbatiste (approved):

PR #4017, opened 2023-09-14
updated deps

dlockhart:
Manually picked up the latest of everything so we have the latest testing with the Playwright fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,7 @@ jobs:
           node-version-file: .nvmrc
           cache: 'npm'
       - name: Install dependencies
-        run: |
-          npm ci
-          npx playwright install-deps
+        run: npm ci
       - name: Build
         run: npm run build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,9 +30,7 @@ jobs:
           node-version-file: .nvmrc
           cache: 'npm'
       - name: Install dependencies
-        run: |
-          npm ci
-          npx playwright install-deps
+        run: npm ci
```
dlockhart, line 33: This would have no longer worked, but it shouldn't be needed since `@brightspace-ui/testing` does it for us.

svanherk:
Lol I've been doing the same thing in https://github.com/BrightspaceUI/core/pull/4012

dlockhart:
Oops, didn't see that! I can close this one, not sure if you want to apply the change here to remove the `install-deps` call to yours.

PR #4020, opened 2023-09-14
Pass the new git HEAD to the publish site checkout

devpow112:
This ensures the version in the demos matches the latest release version. Right now it can be 1 behind.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,8 @@ jobs:
     name: Release
     timeout-minutes: 5
     runs-on: ubuntu-latest
+    outputs:
+      git-head: ${{steps.git.outputs.head}}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -34,6 +36,9 @@ jobs:
           NPM: true
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           RALLY_API_KEY: ${{ secrets.RALLY_API_KEY }}
+      - name: Get new git HEAD
+        id: git
+        run: echo "head=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
```
devpow112, line 41: This will be different then `GITHUG_REF` if semantic release generated and pushed a new commit

dlockhart (approved):
Cool, crazy!

PR #4021, opened 2023-09-14
US152268 - Migrate filter vdiff tests

svanherk:
Will do the other filter tests in another PR, this was getting big enough to do on its own. Expecting quite a few diffs, will comment on the goldens PR.

Interesting pieces:
- Some tests had filters that were opening up based on their location in the old HTML page. To minimize diffs, I kept them opening up, but we may decide it's not worth maintaining and just switch them all to open down.
- I had a hard time with the timing on some of these tests. They are all stable locally now, but we'll see about CI.  Will add a note inline.

Final reports: 

- https://vdiff.d2l.dev/BrightspaceUI/core/5fb0e0ac083004647810c9ccfe66981d609d07ff/2023-09-14_17-49-04/report/
- https://vdiff.d2l.dev/BrightspaceUI/core/ce8383ca940e59a321776776dcee78bb8e7e6925/2023-09-14_19-38-44/report/


components/filter/test/filter.vdiff.js
```diff
@@ -0,0 +1,269 @@
...
+
+	describe('multiple', () => {
+		async function enterDimension(filter, dimNum) {
+			const menu = filter.shadowRoot.querySelector('d2l-menu');
+			const initialMenuHeight = menu.clientHeight;
+			const dims = filter.shadowRoot.querySelectorAll('d2l-menu-item');
+			sendKeysElem(dims[dimNum - 1], 'press', 'Enter');
+			await oneEvent(filter, 'd2l-filter-dimension-first-open');
+			await waitUntil(() => menu.clientHeight !== initialMenuHeight);
+		}
```
svanherk, line 202: This was giving me a lot of trouble, but I think I've found a good combination of flake-resistant but flexible to future changes

dlockhart: Seems like it was painful for sure. This is the kind of thing we could put into a "menu test SDK" or something if it continues to crop up.

dlockhart (approved):
Nice!

PR #4023, opened 2023-09-14
Migrate link vdiffs

dlockhart:
Not expecting anything particularly remarkable with these. Locally, the ones that don't wrap onto multiple lines ended up `2px` higher and in some cases `2px` wider, but I'm not concerned with that, just more curious about it.

[Report Link](https://vdiff.d2l.dev/BrightspaceUI/core/9e051ba296f7043bb5fae333f655ad104fd83ba1/2023-09-14_18-27-49/report/)


components/link/test/link.vdiff.js
```diff
@@ -0,0 +1,61 @@
...
+
+	before(async() => {
+		return new Promise(resolve => {
+			const link = document.createElement('link');
+			link.id = 'link-sass';
+			link.type = 'text/css';
+			link.rel = 'stylesheet';
+			link.href = './test/sass.output.css';
+			link.onload = resolve;
+			document.getElementsByTagName('head')[0].appendChild(link);
```
dlockhart, line 14: This was the solution I landed on to get the Sass generated CSS into the page. It's not particularly amazing, but I'm also hoping doing something like this will be rare -- especially outside of core.

.github/workflows/ci.yml
```diff
@@ -30,9 +30,7 @@ jobs:
           node-version-file: .nvmrc
           cache: 'npm'
       - name: Install dependencies
-        run: |
-          npm ci
-          npx playwright install-deps
```
svanherk, line 35: Thanks!

svanherk (approved):

PR #4031, opened 2023-09-15
Migrate html-block vdiff tests

dlockhart:
Not expecting any crazy diffs here.

The math will render slightly smaller because we're no longer setting a `1.1` outputScale -- we can change this in the test framework but we decided initially against it.


components/html-block/html-block.js
```diff
@@ -242,6 +245,10 @@ class HtmlBlock extends LitElement {
 		}
 	}
 
+	async getLoadingComplete() {
```
dlockhart, line 248: This is to ensure that MathJax has loaded and done its typesetting before snapshotting. Previously, I suspect this was "just working" because all the tests were in the same page and by the time it got down to the math ones, a few seconds had passed.

I found that even waiting until after all the `render()` calls resolve wasn't enough, and that a further `100ms` was needed for the first one. The way I've buried this is by creating an optional `getLoadingComplete` on the renderer plugins. The MathJax one returns a Promise that's just a `100ms` timeout the _first_ time.

components/html-block/test/html-block.vdiff.js
```diff
@@ -0,0 +1,308 @@
+import '../html-block.js';
+import { expect, fixture, html } from '@brightspace-ui/testing';
+
+const content = `
```
dlockhart, line 4: Bonus is we can use string literals so the HTML/Math/Code is actually readable!

components/html-block/html-block.js
grant-cleary, line 248: Oh I like this! There are use cases in the LMS where it would be useful to know when HTML has finished rendering.

components/html-block/test/html-block.vdiff.js
```diff
@@ -0,0 +1,308 @@
...
+`;
+
+const code = `<pre class="d2l-code d2l-code-dark"><code class="language-javascript">function helloGrumpy(name) {
+	console.log(\`Hi there \${name}.\`);
+}
+helloGrumpy('Wizard');</code></pre>`;
+
+describe('d2l-html-block', () => {
+
+	const viewport = { width: 376 };
```
dlockhart, line 124: This weird value is to account for the different page padding/margins in the new framework.

grant-cleary (approved):
LGTM! Sorry this caused you some grief. 😞

helpers/mathjax.js
```diff
@@ -75,6 +77,12 @@ class HtmlBlockMathRenderer {
 		await window.D2L.renderingPromise;
 	}
 
+	async getLoadingComplete() {
+		if (this.#loadingCompletePromise !== undefined) return this.#loadingCompletePromise;
+		this.#loadingCompletePromise = new Promise(resolve => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
dlockhart: Yeah I spent a good hour poking at different things and ended up throwing my hands in the air. All their stuff resolves, the content has been replaced with `<mjx-doc>`, I waited `updateComplete`s, multiple animation frames... nothing.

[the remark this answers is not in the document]
grant-cleary: Yeah, I'm not sure what exactly is off about it, but _something_ about the shadowDOM workaround they suggested to us doesn't exactly play nicely with their core APIs. I've definitely found that a lot of their documented functionality just doesn't work for us. Blech.

svanherk (approved):

PR #4045, opened 2023-09-18
Alternative link new-window

dlockhart:
This shows an alternate approach to the "new-window" indicator that uses the `::after` pseudo-class. The improvement here is that you don't see a weird underline when you hover/focus.

Before:
Screenshot 2023-09-18 at 2 02 06 PM

After:
Screenshot 2023-09-18 at 1 34 52 PM

The icon's implementation is also constrained solely to CSS, except the offscreen text. Unfortunately that means embedding the SVG directly into the file, which means duplication of the SVG definition (once for each size actually).

The other downside to this approach is that I can't find a way to have the icon not wrap on its own onto a new line. The existing complex combination of `white-space: nowrap;` and `white-space: normal;` doesn't work. On the flip side, I noticed that code breaking some old visual diff tests where they were going `<d2l-link style="white-space: nowrap">` to prevent wrapping... so maybe it's not safe anyway.

Thoughts?


bearfriend:
I think using `white-space: inherit;` (instead of `normal`) on the inner `<span>` should fix the `<d2l-link style="white-space: nowrap">` issue, and forcing everything (including the offscreen stuff) onto a single line fixes the trailing underline.

components/link/link.js
```diff
@@ -123,19 +123,14 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
...
-					height: 14px;
-					width: 14px;
-				}
-
-				@media print {
-					d2l-icon.d2l-new-window {
-						display: none;
+				@media screen {
+					:host([new-window]) a::after {
+						content: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2218%22%20height%3D%2218%22%20viewBox%3D%220%200%2018%2018%22%20mirror-in-rtl%3D%22true%22%3E%0A%20%20%20%20%3Cpath%20fill%3D%22%23006fbf%22%20d%3D%22M12.1.6a.944.944%200%200%200%20.2%201.04l1.352%201.353L10.28%206.37a.956.956%200%200%200%201.35%201.35l3.382-3.38%201.352%201.352a.944.944%200%200%200%201.04.2.958.958%200%200%200%20.596-.875V.96a.964.964%200%200%200-.96-.96h-4.057a.958.958%200%200%200-.883.6z%22%2F%3E%0A%20%20%20%20%3Cpath%20fill%3D%22%23006fbf%22%20d%3D%22M14%2011v5a2.006%202.006%200%200%201-2%202H2a2.006%202.006%200%200%201-2-2V6a2.006%202.006%200%200%201%202-2h5a1%201%200%200%201%200%202H2v10h10v-5a1%201%200%200%201%202%200z%22%2F%3E%0A%3C%2Fsvg%3E%0A");
```
NicholasHallman, line 128: The icon size can be changed with css like we do in the LMS [here ](https://github.com/Brightspace/lms/blob/master/lp/framework/web/D2L.Web/ui/Link/Link.css#L34)

I also believe that since this icon exists in the LMS we can use it's link instead of the embedded svg

NicholasHallman: Ah yeah I fixed that with `position: absolute` which worked suprisingly even without a parent `relative`

NicholasHallman: > For the URL, we don't really know where it's going to end up on the CDN in all cases

Right but wouldn't the url used by the D2L MVC css point to an icon surved by IIS? Can't we use the icon at this address `/d2l/img/0/Framework.Link.OpenInNewWindow.svg?v=20.23.10.0-5644968`

NicholasHallman: Oh I guess this falls apart if the component is used outside the LMS, so I'm cool with using the embedded svg data

dlockhart: Yeah exactly -- core elements are used in a bunch of projects outside the LMS (Wave, the Daylight Site, etc.).

NicholasHallman:
> The improvement here is that you don't see a weird underline when you hover/focus.

I don't think there's a solution here where we keep the icon sticking to the last word and not have that weird spacing issue. Before the word-wrap change the link underline extended under the icon as well. If that's undesired then I believe this is the best solution.

dlockhart:
> I don't think there's a solution here where we keep the icon sticking to the last word and not have that weird spacing issue.

I agree, I haven't been able to find any solution to both issues either. I'll check with Jeff G to see if he prefers one trade-off or the other.

PR #4048, opened 2023-09-18
migrate skeleton visual-diff tests

dlockhart:
[Report](https://vdiff.d2l.dev/BrightspaceUI/core/26e8566cf0bccc75e6c8dee66018d18ea455df65/2023-09-18_20-13-36/report/?file=components%2Fskeleton%2Ftest%2Fskeleton.vdiff.js), with notes about the diffs in the [report PR](https://github.com/BrightspaceUI/core/pull/4049).


components/skeleton/demo/skeleton-group-test-wrapper.js
```diff
@@ -1,7 +1,16 @@
-import { html, LitElement } from 'lit';
+import { css, html, LitElement } from 'lit';
 import { SkeletonGroupMixin } from '../skeleton-group-mixin.js';
 
 class SkeletonGroupTestWrapper extends SkeletonGroupMixin(LitElement) {
+	static get styles() {
+		return css`
+			:host {
```
dlockhart, line 7: This was defined in the visual-diff HTML files, so just applied it to the demo component instead.

components/skeleton/test/skeleton-group.vdiff.js
```diff
@@ -0,0 +1,235 @@
...
+
+	it('make-one-skeleton', async() => {
+		const elem = await fixture(noSkeletonFixture);
+		const container = elem.querySelector('d2l-test-skeleton-container');
+		container.skeleton = true;
+		await container.updateComplete;
+		await expect(elem).to.be.golden();
+	});
+
+	it('make-one-not-skeleton', async() => {
```
dlockhart, line 69: @jameswklassen, I think this test wasn't set up correctly before? The previous fixture had no items with skeleton applied, the test wasn't actually doing anything. I changed it here to use the `oneSkeleton` fixture instead.

components/skeleton/test/skeleton.vdiff.js
```diff
@@ -0,0 +1,65 @@
...
+
+describe('d2l-skeleton', () => {
+
+	['ltr', 'rtl'].forEach(dir => {
+
+		const rtl = dir === 'rtl';
+
+		describe(dir, () => {
+
+			['standard', 'compact', 'small', 'label'].forEach((type) => {
```
dlockhart, line 18: This will add 3 more "lines" tests for the "label" type, but seemed worth it to get the nested grouping.

components/skeleton/test/skeleton-group.vdiff.js
```diff
@@ -0,0 +1,235 @@
...
+		const elem = await fixture(oneSkeletonFixture);
+		elem.querySelector('d2l-test-skeleton-container[skeleton]').remove();
+		await elem.updateComplete;
+		await expect(elem).to.be.golden();
+	});
+
+	[true, false].forEach((skeleton) => {
+		it(`mixed-elements${skeleton ? '-skeleton' : ''}`, async() => {
+			const elem = await fixture(html`
+				<d2l-skeleton-group-test-wrapper style="width: 330px">
```
jameswklassen, line 94: is hard coding `width` the new best practice?

dlockhart: It was what I needed to do here to get clean diffs. An alternate would be to set the viewport, but when I did that the responsive rules for typography (headings specifically) kicked in so all the headings were smaller.

Probably the "best" best practice would be to just not set a width or viewport and just roll with the default 800px width, unless that's totally unreasonable.

components/skeleton/test/skeleton-group.vdiff.js
```diff
@@ -0,0 +1,235 @@
...
+							<d2l-test-skeleton-container></d2l-test-skeleton-container>
+						</d2l-skeleton-group-test-wrapper>
+						<d2l-skeleton-group-test-wrapper>
+							<d2l-test-skeleton-box></d2l-test-skeleton-box>
+						</d2l-skeleton-group-test-wrapper>
+					</d2l-skeleton-group-test-wrapper>
+				</div>
+			`
+		}
+	].forEach(({ name, template }) => {
```
jameswklassen, line 153: wow, this is a great way to have tests with the same logic and different fixtures 😮

jameswklassen (approved):

components/skeleton/test/skeleton-group.vdiff.js
jameswklassen: yeah. 😬 I think I did this because of some weirdness with the `*-test-wrapper` functions that were baked in, and iirc, it wasn't trivial to change so I moved on (and forgot about it 😆).

I might take a look at improving that at some point

components/skeleton/test/skeleton-group.vdiff.js
```diff
@@ -0,0 +1,237 @@
...
+		const container = elem.querySelector('d2l-test-skeleton-container');
+		container.skeleton = false;
+		await container.updateComplete;
+		await expect(elem).to.be.golden();
+	});
+
+	it('add-element', async() => {
+		const elem = await fixture(noSkeletonFixture);
+		elem.insertAdjacentHTML('beforeend', '<d2l-test-skeleton-container skeleton></d2l-test-skeleton-container>');
+		await waitUntil(() => elem.skeleton);
```
dlockhart, line 80: `add-element` and `remove-element` flaked a bit I think because the communication of the new state isn't immediate. `waitUntil` should be pretty quick here.

components/skeleton/test/skeleton-group.vdiff.js
jameswklassen: oh good catch! yes this test should have a single element with `skeleton` on it, that gets removed and as a result the whole group un-skeletonizes

PR #4056, opened 2023-09-19
Link - new window fixes

bearfriend:
- Match the host `white-space` style
- Don't underline the space between the link text and icon

The only downside is that it doesn't update with changes to the host `white-space`, but that should basically not be an issue in practice.

Alternative to the alternatives #4050 and #4045

### DIFF
--- a/components/link/link.js
+++ b/components/link/link.js
@@ -125,13 +125,19 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 				}
 				d2l-icon.d2l-new-window {
 					color: var(--d2l-color-celestine);
-					margin-inline-start: 0.5ch;
+					height: 0.78em;
+					margin-inline-start: 0.3em;
 					vertical-align: inherit;
+					width: 0.78em;
 				}
 
 				:host([small]) d2l-icon.d2l-new-window {
-					height: 14px;
-					width: 14px;
+					height: 1em;
+					width: 1em;
+				}
+
+				:host(:hover) d2l-icon.d2l-new-window {
+					--d2l-icon-fill-color: var(--d2l-color-celestine-minus-1);
 				}
 
 				@media print {
--- a/components/link/link.js
+++ b/components/link/link.js
@@ -169,7 +169,7 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 			? '_blank'
 			: this.target;
 		const newWindowIndicator = this.newWindow
-			? html`<span style="white-space: nowrap;"><span style="font-size: 0;">&nbsp;</span><d2l-icon class="d2l-new-window" icon="tier1:new-window"></d2l-icon></span>`
+			? html`<span style="white-space: nowrap; line-height: 0;"><span style="font-size: 0;">&nbsp;</span><d2l-icon class="d2l-new-window" icon="tier1:new-window"></d2l-icon></span>`
 			: nothing;
 		const newWindowMessage = (target === '_blank')
 			? html`<span class="d2l-offscreen">${this.localize('components.link.open-in-new-window')}</span>`
--- a/components/link/link.js
+++ b/components/link/link.js
@@ -125,6 +125,7 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 				}
 				d2l-icon.d2l-new-window {
 					color: var(--d2l-color-celestine);
+					margin-inline-start: 0.5ch;
 					vertical-align: inherit;
 				}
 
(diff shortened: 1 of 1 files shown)


components/link/link.js
```diff
@@ -180,7 +181,7 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 				?download="${this.download}"
 				href="${ifDefined(this.href)}"
 				target="${ifDefined(target)}"
-			><span style="white-space: nowrap;"><span style="white-space: normal;"><slot></slot></span>${newWindowIndicator}${newWindowMessage}</span></a>`;
+			><span style="white-space: nowrap;"><span style="white-space: ${getComputedStyle(this).whiteSpace || 'normal'};"><slot></slot></span>${newWindowIndicator}${newWindowMessage}</span></a>`;
```
dlockhart, line 184: I think we can make a slight optimization and apply `white-space: nowrap;` to the `<a>`, removing one `<span>`?

components/link/link.js
```diff
@@ -125,6 +125,7 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 				}
 				d2l-icon.d2l-new-window {
 					color: var(--d2l-color-celestine);
+					margin-inline-start: 0.5ch;
```
dlockhart, line 128: When I asked Jeff how much space he'd like between the link text and the icon, he said:

```suggestion
					margin-inline-start: 6px;
```

dlockhart (approved):
Cool, that's a really creative solution that I definitely wouldn't have thought of! Seems like the best of all worlds?

components/link/link.js
bearfriend: Doesn't it seem weird for it to be fixed though?

dlockhart: Maybe. This ends up being `5.51px` @ `20px` default font size, so I'm good with it either way.

bearfriend: If we make the icon:
```
width: 0.78em;
height: 0.78em;
margin-inline-start: 0.33em; /* 6px default */
```

Then the icon scales with the link no what people do to it.

dlockhart: Chatted with Jeff -- he definitely likes the idea of scaling the icon with the text like we do with subtle buttons, so good call.

He'd like for it to be `18px` for "normal" links at default font size, so `0.9em` should achieve that.

For `small` links, they're a bit weird in that the font-size is `rem`-based and not `em`-based, so the font size never actually goes up/down with the cascaded font-size. So for them, I think we want `0.7rem` which would result in `14px`.

Does that make sense?

NicholasHallman (approved):
Nice! LGTM

components/link/link.js
bearfriend: It does except I still think `1em` for the small links makes more sense because it shouldn't really care about anything else at that point except the font-size of the thing it lives inside. So, whether the small link were to change its font-size to `0.8rem`, '16px', or `0.842105263156em`, the icon would still be fine.

dlockhart: Ah yep cool cool, carry on 😎

dlockhart: Could we just roll with `0.95em` for both then for ultimate consistency and less CSS?

components/link/link.js
```diff
@@ -164,7 +164,7 @@ class Link extends LocalizeCoreElement(FocusMixin(LitElement)) {
 			'd2l-link-small': this.small,
 			'truncate': this.lines > 0
 		};
-		const styles = (this.lines > 0) ? { '-webkit-line-clamp': this.lines } : {};
+		const styles = { whiteSpace: 'nowrap', webkitLineClamp: this.lines || null };
```
dlockhart, line 167: Since these get put inline, I think typically we'd only put dynamic stuff in these style maps? vs. just putting it up in the CSS?

bearfriend: lol, yes, of course.

dlockhart (approved):

bearfriend:
Finally figured out how to avoid the alternating/inherited `white-space` styles. No more `getComputedStyle`, which wasnt guaranteed to be correct on first render.

bearfriend:
This all looks god to me now. `vdiff` changes look good, but I'll let those run and merge on Nick's PR. The one `visual-diff` change is a flake, I'm pretty sure. I'll merge in the morning unless I see objections.

dlockhart:
Is it weird that the two "new window" print vdiffs grew in width by 8-10 pixels? Maybe not enough stuff is getting hidden in print mode?

bearfriend:
> Is it weird that the two "new window" print vdiffs grew in width by 8-10 pixels? Maybe not enough stuff is getting hidden in print mode?

They actually shrunk by 8-10px because the space/margin is being properly hidden in print mode now.

dlockhart (approved):

PR #4059, opened 2023-09-19
US152268 - Simple table vdiff migrations

svanherk:
This was created from https://github.com/BrightspaceUI/core/pull/4007 and then adjusted to try to reuse fixtures (which is probably both better and worse).  This only moves the "easy" diffs this first pass.  The next PR will deal with the tricky "sticky" ones.


components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,407 @@
...
+	const { selected } = { selected: false, ...opts };
+	return html`${keys.map(key => html`
+		<tr ?selected="${selected}">
+			<td>${`Cell ${key}-A`}</td>
+			<td>${`Cell ${key}-B`}</td>
+			<td>${`Cell ${key}-C`}</td>
+		</tr>
+	`)}`;
+}
+function createFruitRows(opts) {
```
svanherk, line 56: This function does a lot of things that aren't used yet, but will be used by the `sticky` tests

components/table/test/table.vdiff.js
```diff
@@ -0,0 +1,407 @@
...
+}
+
+describe('table', () => {
+
+	[
+		{ type: 'default', rtl: false  },
+		{ type: 'default', rtl: true  },
+		{ type: 'light', rtl: false  },
+		{ type: 'light', rtl: true }
+	].forEach(({ type, rtl }) => {
```
svanherk, line 96: I went with this pattern to try to reduce the amount of nesting, saves us a couple levels 🤷

dlockhart (approved):
Nice! Thanks, this pattern makes it much more readable and less repetitive.

svanherk:
> Nice! Thanks, this pattern makes it much more readable and less repetitive.

Yeah, I find it makes the `create*` functions look a bit like gibberish, but the individual tests are hopefully readable and small and you can see what is different between each one a bit easier. But still not 100% awesome, maybe because we're trying not to change the diffs and they'd be a bit cleaner writing them this way from the start

PR #4061, opened 2023-09-19
migrate switch vdiffs

dlockhart:
Not expecting anything surprising.


components/switch/test/switch.vdiff.js
```diff
@@ -0,0 +1,50 @@
...
+describe('d2l-switch', () => {
+
+	describe('ltr', () => {
+
+		describe('basic', () => {
+			[
+				{ name: 'off', template: offFixture },
+				{ name: 'off-focus', template: offFixture, action: async(elem) => await focusElem(elem) },
+				{ name: 'off-disabled', template: html`<d2l-switch text="Test Text" disabled></d2l-switch>` },
+				{ name: 'off-hover', template: offFixture, action: async(elem) => await hoverElem(elem.shadowRoot.querySelector('.d2l-switch-inner')) },
```
dlockhart, line 16: This was a little fun -- coincidentally `hoverElem(elem)` hovers right in the middle of the switch, which just happens to be between the toggle and the text in a place that _doesn't_ trigger the hover styles.

components/switch/test/switch-visibility.vdiff.js
```diff
@@ -0,0 +1,152 @@
...
+				{
+					name: 'on with only lots of whitespace conditions',
+					template: html`
+						<d2l-switch-visibility on>
+											
+						</d2l-switch-visibility> 
+					`
+				},
+				{
+					name: 'on with only loooots of whitespace conditions',
```
svanherk, line 98: Minor: Lol what is this testing, that the whitespace doesn't show up? Wondering if you could do something like https://github.com/BrightspaceUI/core/blob/15d79588871270849320a3928fa40cdeed95ede3/components/button/test/floating-buttons.vdiff.js#L5

components/switch/test/switch.vdiff.js
```diff
@@ -0,0 +1,50 @@
+import '../switch.js';
+import { expect, fixture, focusElem, hoverElem, html } from '@brightspace-ui/testing';
+
+const offFixture = html`<d2l-switch text="Test Text"></d2l-switch>`;
+const onFixture = html`<d2l-switch text="Test Text" on></d2l-switch>`;
+
+describe('d2l-switch', () => {
+
+	describe('ltr', () => {
```
svanherk, line 9: Minor: If you feel like it, I've been removing some places where we had the `ltr` (and then manually removing it from the new golden names). Anywhere where it doesn't really seem to be adding anything 🤷

svanherk (approved):

components/switch/test/switch-visibility.vdiff.js
dlockhart: Haha I think it's there because when there are "real" non-whitespace things in the slot it renders the help tooltip, so this is testing that they're ignoring empty text nodes in the slot. But yeah, I'll try changing it to something less verbose.

dlockhart: Alright, I've combined the "whitespace" tests into one and am using that `Array(blah).fill` method to fill in various types of whitespace.

components/switch/test/switch.vdiff.js
dlockhart: Cool -- I've done this for the switch tests but left the visibility-switch ones since the ltr/rtl suite is unfortunately outside the tests.

components/switch/test/switch-visibility.vdiff.js
svanherk: Did you mean to remove the off case too? If so probably need to delete that golden, or else it'll be deleted with the next failing vdiff run

dlockhart: I did mean to remove them but missed the goldens, thanks -- fixed.

svanherk (approved):

PR #4063, opened 2023-09-19
Fix flaky math vdiffs

dlockhart:
Just waiting `150ms` for MathJax to render math in `html-block`s is still flaking out sporadically. Unfortunately, this uses the nuclear option of waiting until the elements are the exact size we're expecting them to be when the math has rendered.

### DIFF
--- a/components/html-block/test/html-block.vdiff.js
+++ b/components/html-block/test/html-block.vdiff.js
@@ -196,11 +196,7 @@ describe('d2l-html-block', () => {
 			html`<d2l-html-block style="width: 650px;" html="${mathBlock}"></d2l-html-block>`,
 			{ mathjax: { renderLatex: true } }
 		);
-		await waitUntil(() => {
-			// eslint-disable-next-line no-console
-			console.log('math (block) height', elem.clientHeight);
-			return elem.clientHeight === 145;
-		});
+		await waitUntil(() => elem.clientHeight === 144);
 		await expect(elem).to.be.golden();
 	});
 
@@ -284,11 +280,7 @@ describe('d2l-html-block', () => {
 			html`<d2l-html-block style="width: 650px;" html="An equation...${mathInline} embedded inline with text, and showing placement of indicies for summations."></d2l-html-block>`,
 			{ mathjax: { renderLatex: true } }
 		);
-		await waitUntil(() => {
-			// eslint-disable-next-line no-console
-			console.log('math (inline) height', elem.clientHeight);
-			return elem.clientHeight === 61;
-		});
+		await waitUntil(() => elem.clientHeight === 61);
 		await expect(elem).to.be.golden();
 	});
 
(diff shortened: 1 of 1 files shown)


grant-cleary (approved):
Can we just delete MathJax? 😆

svanherk (approved):
I like @grant-cleary's idea too, either works for me 😏

dbatiste:
@grant-cleary , do you know if there is an API we can use to better know when things have fully rendered?

grant-cleary:
> @grant-cleary , do you know if there is an API we can use to better know when things have fully rendered?

For MathJax specifically, that was sort of the intent of setting `win.D2L.renderingPromise()`, though admittedly it seems that's not working as expected? I think part of the issue here is that MathJax does pretty much everything on a global level, so figuring out whether the math in a particular html-block has rendered is difficult.

It's possible that we could do something here to test: https://github.com/BrightspaceUI/core/blob/main/helpers/mathjax.js#L75, but from Dave's experience before I don't think that rendering promise proved sufficient. I would need to dig into why that is.

dbatiste:
Ok yeah, thanks. I see you guys had a bunch of conversation around this already.

dlockhart:
Yeah I tried all combinations of those promises, and then waited some more, and some more. Nothing resulted in consistent timings unfortunately.

grant-cleary:
> Yeah I tried all combinations of those promises, and then waited some more, and some more. Nothing resulted in consistent timings unfortunately.

Did you consider waiting for three "some more"s instead of just two? :trollface:

grant-cleary:
More seriously, there's something very odd about this custom MathJax stuff we have going on. MathJax's documentation suggests that it should just be plug-and-play... We're basically just adding a custom adapter, which is something that's supposed to be possible. But for some reason I find that I don't get the documented results I expect when using other MathJax APIs that should be untouched by the custom stuff we've done, so I feel like I need to take some time to do a deeper dive and figure out what's up with that.

I have a suspicion something similar might be biting us here, because we _should_ be able to rely on these promises to tell us if rendering is done. We know the queue does _something_ because the problem I put it in to fix (MathJax abandoning one html-block mid-render to go render another on the page) was definitely fixed. It's feasible that MathJax is respecting the queue, but resolving it before rendering is actually finished, for example. :shrug:

dbatiste:
The MathJax stuff that I did to make to work in a shadowRoot was cobbled together quite a long time ago based on a couple different GitHub conversations involving Davide, so it's possible there is now a better way now, with better support.

grant-cleary:
> The MathJax stuff that I did to make to work in a shadowRoot was cobbled together quite a long time ago based on a couple different GitHub conversations involving Davide, so it's possible there is now a better way, with better support.

I'm confident there is not (at least not a better supported way), because I've looked into it a number of times since then 😆. I've managed to make some changes with newer versions of MathJax that are a little bit better, but the core issue still remains. Unfortunately I think development on MathJax 3 has been dropped entirely in favour of MathJax 4, so I'm not expecting we're going to see much movement there. MathJax's APIs aren't well-documented, so it comes down to having to go through their code to try to figure out what the recommended paved paths are.

PR #4066, opened 2023-09-19
US157863: Alert-toast: Support multiple toasts

margaree:
[Rally story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F722730033259)

### DIFF
--- a/components/alert/alert-toast.js
+++ b/components/alert/alert-toast.js
@@ -1,6 +1,7 @@
 import './alert.js';
 import { css, html, LitElement } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { styleMap } from 'lit/directives/style-map.js';
 
 const reduceMotion = matchMedia('(prefers-reduced-motion: reduce)').matches;
 
@@ -58,6 +59,8 @@ class AlertToast extends LitElement {
 			 * @default "default"
 			 */
 			type: { type: String, reflect: true },
+			_bottomSpacing: { type: Number },
+			_closeClicked: { state: true },
 			_state: { type: String }
 		};
 	}
@@ -70,7 +73,6 @@ class AlertToast extends LitElement {
 
 			.d2l-alert-toast-container {
 				border-radius: 0.3rem;
-				bottom: 1.5rem;
 				box-shadow: 0 0.1rem 0.6rem 0 rgba(0, 0, 0, 0.1);
 				display: none;
 				left: 0;
@@ -104,6 +106,14 @@ class AlertToast extends LitElement {
 				transform: translateY(0);
 			}
 
+			.d2l-alert-toast-container[data-state="moving"] {
+				transition: bottom 600ms ease;
+			}
+
+			.d2l-alert-toast-container[data-state="moving-click"] {
+				transition: bottom 200ms ease;
+			}
+
 			d2l-alert {
 				animation: none;
 			}
@@ -123,6 +133,8 @@ class AlertToast extends LitElement {
 		this.noAutoClose = false;
 		this.open = false;
 
+		this._bottomSpacing = 0;
+		this._closeClicked = false;
 		this._hasFocus = false;
 		this._hasMouse = false;
 		this._state = states.CLOSED;
@@ -163,10 +175,14 @@ class AlertToast extends LitElement {
 	}
 
 	render() {
+		const containerStyles = {
+			bottom: `calc(1.5rem + ${this._bottomSpacing}px)`
+		};
 		return html`
 			<div
 				class="d2l-alert-toast-container vdiff-target"
 				data-state="${this._state}"
+				style="${styleMap(containerStyles)}"
 				@transitionend=${this._onTransitionEnd}>
 				<d2l-alert
 					@blur=${this._onBlur}
@@ -233,6 +249,7 @@ class AlertToast extends LitElement {
 
 	_onCloseClicked(e) {
 		e.preventDefault();
+		this._closeClicked = true;
 		this.open = false;
 	}
 
@@ -260,6 +277,8 @@ class AlertToast extends LitElement {
 	}
 
 	async _openChanged(newOpen) {
+		let computedStyles;
+		if (this._innerContainer) computedStyles = getComputedStyle(this._innerContainer);
 		if (newOpen) {
 			if (this._state === states.CLOSING) {
 				this._state = states.OPENING;
@@ -283,7 +302,7 @@ class AlertToast extends LitElement {
 				'd2l-alert-toast-open', {
 					bubbles: true,
 					composed: false,
-					detail: { height: height }
+					detail: { height }
 				}
 			));
 		} else {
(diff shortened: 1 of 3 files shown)


components/alert/alert-toast.js
```diff
@@ -256,19 +332,34 @@ class AlertToast extends LitElement {
...
 				this._state = states.CLOSED;
-			} else if (this._state === states.OPENING || this._state === states.OPEN) {
+			} else if (this._state === states.OPENING || this._state === states.OPEN || this._state === states.SLIDING) {
 				this._state = states.CLOSING;
 			}
 			this.dispatchEvent(new CustomEvent(
 				'd2l-alert-toast-close', {
 					bubbles: true,
 					composed: false,
-					detail: { action: this._action }
```
margaree, line 271: This doesn't seem to be hooked up to anything and doesn't seem to be used so I cleaned it up while here.

components/alert/alert-toast.js
```diff
@@ -58,6 +61,9 @@ class AlertToast extends LitElement {
 			 * @default "default"
 			 */
 			type: { type: String, reflect: true },
+			_bottomHeight: { type: Number },
+			_bottomMargin: { type: Number },
```
dlockhart, line 65: `_bottomHeight` and `_bottomMargin` feel like they could be state properties as well?

components/alert/alert-toast.js
```diff
@@ -88,29 +93,42 @@ class AlertToast extends LitElement {
...
-				transition-duration: 250ms;
+				transition-duration: 600ms;
 				transition-property: transform, opacity;
-				transition-timing-function: ease-in;
+				transition-timing-function: ease;
+			}
+
+			.d2l-alert-toast-container.d2l-alert-toast-container-not-bottom[data-state="closing"] {
+				transform: unset;
+				transition-property: opacity;
```
dlockhart, line 103: Just so I understand: this is preventing toasts that aren't at the bottom from transitioning vertically away as they close?

components/alert/alert-toast.js
```diff
@@ -200,6 +252,26 @@ class AlertToast extends LitElement {
 		clearTimeout(this._setTimeoutId);
 	}
 
+	_handleAlertClose(e) {
+		if (!e || e.target === this || !this.open) return;
```
dlockhart, line 256: Nit:
```suggestion
		if (e?.target === this || !this.open) return;
```

components/alert/alert-toast.js
```diff
@@ -200,6 +252,26 @@ class AlertToast extends LitElement {
...
+
+		const containerBottom = parseFloat(getComputedStyle(this._innerContainer).getPropertyValue('bottom'));
+		const closingContainerBottom = e.detail.bottom;
+		if (closingContainerBottom > containerBottom) return; // closing alert is above this alert, no need to adjust bottom spacing
+
+		this._bottomHeight -= e.detail.height;
+		this._bottomMargin -= 0.6;
```
dlockhart, line 263: The `0.6` that gets used as the space between each toast feels important and shows up a few times. Might be worth throwing it in a constant at the top of the file so we don't accidentally forget one of them if we adjust it in the future?

components/alert/alert-toast.js
```diff
@@ -256,19 +332,37 @@ class AlertToast extends LitElement {
...
 			this.setAttribute('role', 'alert');
+			await this.updateComplete;
+			const height = this._innerContainer.offsetHeight;
+			this.dispatchEvent(new CustomEvent(
+				'd2l-alert-toast-open', {
+					bubbles: true,
+					composed: false,
+					detail: { height }
+				}
+			));
```
dlockhart, line 343: This is going to force a browser repaint. Since you don't need to fire the event synchronously anyway, it might be slightly more performant to put this inside an rAF? Then `_openChanged` doesn't need to be `async` anymore and I don't think you need to wait for `updateComplete`.

```suggestion
			requestAnimationFrame(() => {
				const height = this._innerContainer.offsetHeight;
				this.dispatchEvent(new CustomEvent(
					'd2l-alert-toast-open', {
						bubbles: true,
						composed: false,
						detail: { height }
					}
				));
			});
```

If that works, the same could be applied to the `else` block.

components/alert/test/alert-toast.vdiff.js
```diff
@@ -33,4 +41,58 @@ describe('alert-toast', () => {
...
+			await expect(document).to.be.golden();
+		});
+
+		it('narrow', async() => {
+			const elem = await fixture(multipleAlerts, { viewport: { width: 400, height: 400 } });
+			await openAlerts(elem);
+			await expect(document).to.be.golden();
+		});
+
+		it.skip('resize', async() => {
```
dlockhart, line 94: TODO?

components/alert/test/alert-toast.vdiff.js
```diff
@@ -33,4 +41,58 @@ describe('alert-toast', () => {
...
+		});
+
+		it('open all then close bottom', async() => {
+			const elem = await fixture(multipleAlerts, { viewport: { width: 700, height: 300 } });
+			await openAlerts(elem);
+			const alert = elem.querySelector('#alert-bottom');
+			alert.open = false;
+			await alert.updateComplete;
+			await expect(document).to.be.golden();
+		});
```
dlockhart, line 86: For these 3, you could probably do something like:

```javascript
['top', 'middle', 'bottom'].forEach(position => {
  it(`open all then close ${position}`, async() => {
    const elem = await fixture(multipleAlerts, { viewport: { width: 700, height: 300 } });
    await openAlerts(elem);
    const alert = elem.querySelector(`#alert-${position}`);
    alert.open = false;
    await alert.updateComplete;
    await expect(document).to.be.golden();
  });
});
```

components/alert/alert-toast.js
margaree: Ah yup that was from earlier work and didn't end up being needed - will remove

components/alert/alert-toast.js
margaree: Correct, we just want them transitioning with opacity

components/alert/test/alert-toast.vdiff.js
margaree: Yup, I'll remove until the resize work happens.

components/alert/alert-toast.js
dlockhart: Cool. Not sure if it would be cleaner then to invert everything such that the CSS class is actually just applied to the bottom one and that's where the transform/transition-property is defined? Then you don't have to apply it to all of them and then remove it from the not-bottom ones. 🤷

components/alert/alert-toast.js
dlockhart: It'll work yeah, but any time you access `offsetHeight` it forces the browser to paint, so it's a good general practice to wrap those reads inside a `rAF` to minimize the repaints.

components/alert/alert-toast.js
```diff
@@ -256,21 +337,44 @@ class AlertToast extends LitElement {
...
+				this.dispatchEvent(new CustomEvent(
+					'd2l-alert-toast-open', {
+						bubbles: true,
+						composed: false,
+						detail: { height }
+					}
+				));
+			});
 		} else {
+			if (!this._innerContainer) return;
```
margaree, line 351: This was needed, otherwise there were issues with the visual diff tests and the `d2l-alert-toast-close` event would get fired.

It seems to me that it makes sense that the close functionality doesn't run on initial load but it's possible something was being relied upon here?

dlockhart: Cool. I noticed something similar in expand-collapse-content where it doesn't want to fire the event during the initialization. I think it was tracking `_isFirstUpdate` and using that -- `_innerContainer` gets set during `firstUpdated` so that's kind of accomplishing the same thing.

components/alert/test/alert-toast.vdiff.js
```diff
@@ -47,9 +47,9 @@ describe('alert-toast', () => {
 			const alert1 = elem.querySelector('#alert-middle');
 			const alert3 = elem.querySelector('#alert-bottom');
 			alert1.open = true;
-			await alert1.updateComplete;
+			await nextFrame();
```
dlockhart, line 50: Can these not `await oneEvent(alert1, 'd2l-alert-toast-open')`?

dlockhart (approved):
LGTM!

PR #4068, opened 2023-09-19
fix: `d2l-list-item` > add `mouseenter`/`mouseleave` events to content

jameswklassen:
## Description

There's an issue in Rubrics where the drag handle doesn't appear when hovering the rows (it only appears when hovering over the controls container).

I think this is what's required to fix.

After making this change, I played around with hover/focus on vanilla lists. I didn't see any issues, but someone more well-versed in `d2l-list` can put this through its paces more rigorously.

## Rally

[DE55064](https://rally1.rallydev.com/#/15545167705d/custom/21568985922?detail=%2Fdefect%2F723654980255)

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -735,7 +735,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				</div>` : nothing}
 				<div slot="content"
 					class="d2l-list-item-content"
-					id="${this._contentId}">
+					id="${this._contentId}"
+					@mouseenter="${this._onMouseEnter}"
+					@mouseleave="${this._onMouseLeave}">
 					<slot name="illustration" class="d2l-list-item-illustration">${illustration}</slot>
 					<slot>${content}</slot>
 				</div>


dlockhart (approved):
This _seems_ correct to me, although curious what Dave B thinks as he's much more familiar with the various slots.

It doesn't look like there's a common parent that we could add this listener to without also adding it to places we don't want it (like the selection controls or the nested lists).

jameswklassen:
> This _seems_ correct to me, although curious what Dave B thinks as he's much more familiar with the various slots.
> 
> It doesn't look like there's a common parent that we could add this listener to without also adding it to places we don't want it (like the selection controls or the nested lists).

Yeah, I thought to add it on the `d2l-list-item-generic-layout` but that's too broad, and it would capture all the nested list elements

https://github.com/BrightspaceUI/core/blob/7083ea76463e3795eb3cc646f8f41c550b65caf7/components/list/list-item-mixin.js#L699-L702

We already attach these functions to a number of elements -- they "cycle" through `mousein`/`mouseout` events as the mouse transitions between elements, so this feels like the way we _probably_ want to do it...but yeah I was thinking @dbatiste has the most context/opinion here.

dbatiste (approved):
Yep, this looks correct to me. 😄

PR #4081, opened 2023-09-20
Try to fix flaky filter vdiff test

svanherk:
Try to fix some flaky filter tests by waiting for the _exact_ menu height we expect.  Is this going to break if we adjust buttons, menu, list, etc. heights?  Yes, probably.

I could try to make them a bit more flexible (`waitUntil(() => menu.clientHeight > aBitLowerThanExpectedHeight);`) but that might still be flaky.

### DIFF
--- a/components/filter/test/filter.vdiff.js
+++ b/components/filter/test/filter.vdiff.js
@@ -195,13 +195,14 @@ describe('filter', () => {
 	});
 
 	describe('multiple', () => {
-		async function enterDimension(filter, dimNum) {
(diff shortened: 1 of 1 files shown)


dlockhart (approved):

svanherk:
Well, I've re-run vdiff 5 times here.  Filter has not flaked once, `overflow-group` has flaked every single time. Yay? 😬 🤦

dlockhart:
I call that a success!

PR #4083, opened 2023-09-20
Migrate input radio & checkbox vdiffs

dlockhart:
Nothing major here, although anything "disabled" because it's using opacity is perceived as a difference. I renamed some of the tests/goldens to not have `wc-` in them.


components/inputs/test/input-checkbox.vdiff.js
```diff
@@ -0,0 +1,108 @@
...
+describe('d2l-input-checkbox', () => {
+
+	before(loadSass);
+	after(unloadSass);
+
+	[false, true].forEach(skeleton => {
+		[false, true].forEach(disabled => {
+			const checkedStates = ['checked', 'unchecked'];
+			if (!skeleton) {
+				checkedStates.push('indeterminate');
```
dlockhart, line 15: The old HTML file actually has fixtures for skeleton + indeterminate but I've excluded them to be consistent. I can add them in though if we think they're valuable.

test/load-sass.js
dlockhart, line 1: Centralized this since a few of our vdiff tests need it now.

components/inputs/test/input-checkbox.visual-diff.js
```diff
@@ -1,57 +0,0 @@
...
-				states.push('indeterminate');
-			}
-			states.forEach((checked) => {
-				const id = `${type}-${state}-${checked}`;
-				it(id, async function() {
-					const rect = await visualDiff.getRect(page, `#${id}`);
-					await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-				});
-				if (state !== 'disabled') {
-					it(`${id}-focus`, async function() {
```
svanherk, line 32: Looks like these `focus` ones didn't get moved over, was that on purpose?

components/inputs/test/input-checkbox.vdiff.js
svanherk: Interesting, as in had fixtures for them but no actual tests/goldens? It would clean up three lines of code for some (hopefully) simple, non-flaky diffs, but not sure how _valuable_ they are.  I'd say I'm indifferent

components/inputs/test/input-checkbox.visual-diff.js
dlockhart: Oops yep thanks. Splitting `wc` out made it complicated and I forgot about those -- restored them now.

components/inputs/test/input-checkbox.vdiff.js
dlockhart: Yeah exactly. They're not testing anything interesting IMO so I'm going to leave them off.

svanherk (approved):

PR #4089, opened 2023-09-21
Try to fix flaky overflow-group test

svanherk:
Adding `getLoadingComplete` to `overflow-group` to wait until after the first chomp. This means the additional chomp menu that gets added is then included in the children `fixture` will wait for. If this isn't enough, Ill need to add something similar to `d2l-icon`, but at least now `d2l-icon` is being included in the `fixture` wait calculations at all.


svanherk:
Initial vdiff run passed, but I'll run it a few times to check

svanherk:
Ok, five runs with no flakes!

PR #4090, opened 2023-09-21
Toast Alerts: Multiple Toasts: Update Animations

margaree:
Closing toasts that aren't the bottom toast will now a 200ms opacity transition rather than 600ms to prevent ghost toast.

Opening and lowest closing keep the same animation as before.


dbatiste (approved):
Interesting. I didn't notice "ghost toast" when I was playing around with it before.

PR #4091, opened 2023-09-21
fix: show tooltips if initialized while target descendants have focus

dlockhart:
I ran into this case while migrating visual diffs.

This fixes an edge case where a tooltip is initially rendered and its target already has focus. This may not happen too often in practice, but in tests where things execute quickly it definitely can.

Previously this worked only if the target itself was the active element, but `<d2l-text-input>`'s validation tooltip has the `<d2l-text-input>` itself as the target (for complicated reasons but mostly because we wanted hovering over the whole thing to show the tooltip) but the actual element that gets focused is the underlying `<input type="text">`.

### DIFF
--- a/components/tooltip/test/tooltip.test.js
+++ b/components/tooltip/test/tooltip.test.js
@@ -1,5 +1,6 @@
 import '../tooltip.js';
-import { aTimeout, expect, fixture, focusElem, html, oneEvent, sendKeys } from '@brightspace-ui/testing';
+import { aTimeout, defineCE, expect, fixture, focusElem, html, oneEvent, sendKeys } from '@brightspace-ui/testing';
+import { LitElement } from 'lit';
 import { runConstructor } from '../../../tools/constructor-test-helper.js';
 
 const basicFixture = html`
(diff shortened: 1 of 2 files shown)


components/tooltip/test/tooltip.test.js
```diff
@@ -203,6 +204,26 @@ describe('d2l-tooltip', () => {
 			await oneEvent(tooltipFixture, 'd2l-tooltip-show');
 			expect(dynamicTooltip.showing).to.be.true;
 		});
+
```
dlockhart, line 207: The unit test above this one handles the other case where the target has focus.

components/tooltip/test/tooltip.test.js
```diff
@@ -203,6 +204,26 @@ describe('d2l-tooltip', () => {
...
+			const targetTag = defineCE(class extends LitElement {
+				render() {return html`<button>nested target</button>`; }
+			});
+			const elem = await fixture(`<div><${targetTag} id="target"></${targetTag}></div>`);
+			const target = elem.querySelector(targetTag);
+
+			await focusElem(target.shadowRoot.querySelector('button'));
+
+			const tooltip = document.createElement('d2l-tooltip');
+			tooltip.announced = true;
```
dlockhart, line 219: I used `announced` here to suppress the console warnings about the tooltip not being associated with an interactive element. I could have also set `for` to `target` and added `role` and `tabindex` to the target. 🤷

dbatiste (approved):

PR #4092, opened 2023-09-21
Migration input search and number vdiffs

dlockhart:
These ended up being quite complex and surfaced some underlying issues with the timing of the validation tooltips. I've put in a separate fix for that in #4091, ~but also including that here so these pass -- I'll resolve the conflict when #4091 merges~.

In terms of changes, I'm not expecting much except that validation tooltips are now captured and in some cases they are now opening down vs. up.


components/inputs/input-number.js
```diff
@@ -345,6 +345,7 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 				?autofocus="${this.autofocus}"
 				@blur="${this._handleBlur}"
 				@change="${this._handleChange}"
+				class="vdiff-target"
```
dlockhart, line 348: Allows input-text's tooltips to be included.

components/tooltip/tooltip.js
```diff
@@ -470,7 +470,9 @@ class Tooltip extends RtlMixin(LitElement) {
 		window.addEventListener('resize', this._onTargetResize);
 
 		requestAnimationFrame(() => {
-			this._updateTarget();
+			if (this.isConnected) {
+				this._updateTarget();
+			}
```
dlockhart, line 475: This came up due to the quickness that tooltips show/hide as elements receive focus. In this case, this rAF code was running after the tooltip had been removed from the DOM.

svanherk:
Looks like `input-number.visual-diff.html` isn't deleted

PR #4102, opened 2023-09-25
US152268 - Rest of table vdiff migrations

svanherk:
This PR migrates the rest of the `table` vdiff tests.  These were trickier and [required special logic added to `testing` to handle them](https://github.com/BrightspaceUI/testing/pull/190), but with that merged I'm actually expecting some pretty clean diffs.

Note, these tests were deemed not super useful and were not migrated over. Their goldens are just being deleted here:
- The `over` versions of `one-row-thead`, `one-row-no-thead-class`, `one-row-no-thead-attr`, `multi-row-thead`, `multi-row-no-thead-class`, `multi-row-no-thead-attr`, `selected-one-row`, `selected-top-bottom`, `selected-all`
- The `over` and `down` versions of `one-column`


dlockhart (approved):
Fantastic! Thanks for figuring out the crazy RTL stuff and slogging through these. 🥇

PR #4106, opened 2023-09-25
fix: List `drag-target-handle-only` option

bearfriend:
After #4100, the `drag-target-handle-only`option on `d2l-list-item` stopped working because that option does rely on the `outside-control` slot being interactable. This PR conditionally removes pointer events on that slot.

There is a bug in Safari that allows slots that are underneath others to receive pointer events in `d2l-list-item`. To prevent this, we strategically remove pointer events (and elsewhere control the slot size within the list-item grid). With `drag-target-handle-only` enabled, the `outside-control-action` slot (normally on top and interactable) is not used, and instead the `outside-control` slot is used for dragging. So, pointer events should only be removed if the `outside-control-action` slot is being used.

A CSS-only approach does not work here because the general sibling combinator can only select following siblings, and `outside-control` comes before `outside-control-action`.

See also: #3808


dbatiste (approved):

jameswklassen (approved):
Tested the vdiff tests without these changes in #4139 to verify they're testing the right thing.

bearfriend:
> Tested the vdiff tests without these changes in #4139 to verify they're testing the right thing.

Perfect. Everything looks good.

PR #4109, opened 2023-09-25
Migrate input text and textarea vdiffs

dlockhart:


components/inputs/test/input-text.vdiff.js
```diff
@@ -0,0 +1,173 @@
...
+			template: html`
+				<d2l-input-text label="Name" label-hidden value="Lorem ipsum dolor sit amet, consectetur adipiscing elit">
+					${createIcon('search', 'left')}
+				</d2l-input-text>
+			`
+		},
+		{
+			name: 'button-icon-right',
+			template: html`
+				<d2l-input-text label="Name" label-hidden value="Lorem ipsum dolor sit amet, consectetur adipiscing elit">
```
dlockhart, line 117: I added hidden labels to a lot of these as errors were getting thrown, causing subsequent tests to fail. Haha, so not _quite_ isolated.

components/inputs/test/input-text.visual-diff.js
```diff
@@ -1,96 +0,0 @@
...
-			it(id, async function() {
-				const rect = await visualDiff.getRect(page, `#${id}`);
-				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-			});
-			if (name.indexOf('disabled') === -1) {
-				it(`${id}-focus`, async function() {
-					await focusWithKeyboard(page, `#${id}`);
-					const rect = await visualDiff.getRect(page, `#${id}`);
-					await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-				});
```
svanherk, line 46: Looks like we removed all these focus tests, that was intentional?

svanherk: So what I do (for my PRs and all the ones I'm reviewing) is check the files here: https://github.com/BrightspaceUI/core/pull/4110/files
And see if any are being removed instead of replaced, and verify I expected that

dlockhart: Ah ok that makes sense, I'd just been relying on the report which is clearly not enough.

Why's there gotta be so many focus tests lol. I may remove some of these.

svanherk: Haha yeah I figured you were gonna say "Yup did it on purpose, they were basically all the same". But just wanted to check, I didn't look at the old pngs to see if they were useful

dlockhart: Alright, I've updated things to keep a couple of the focus tests but remove the redundant ones.

svanherk (approved):

PR #4116, opened 2023-09-26
Don't throw missing label error if host is no longer connected

bearfriend:
Because there is a hard [3-second delay](https://github.com/BrightspaceUI/core/blob/ef076269421ed0ceb07c28fed82367c5b9fed612/mixins/labelled/labelled-mixin.js#L204) in `lablledMixin` before throwing a missing label error, the error can be thrown between tests, or in other tests entirely, causing tests to fail in a confusing way. Checking that the host component is still connected before throwing prevents this.


dbatiste (approved):

dlockhart (approved):
Thanks for fixing that up.

PR #4117, opened 2023-09-26
US158015: Toast Alerts: Multiple Toasts: Hover and focus behaviour

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F723821091559)

The current behaviour is that when a toast is hovered over or focused on, its close timer resets. The desired behaviour is that when one of a group of visible toasts is hovered over or focused on then all those toasts have their timers stop.

### DIFF
--- a/components/alert/alert-toast.js
+++ b/components/alert/alert-toast.js
@@ -158,7 +158,9 @@ class AlertToast extends LitElement {
 		this._state = states.CLOSED;
 		this._totalSiblingHeightBelow = 0;
 
+		this._closeTimerStop = this._closeTimerStop.bind(this);
 		this._handlePageResize = this._handlePageResize.bind(this);
+		this._handleSiblingCloseTimerStart = this._handleSiblingCloseTimerStart.bind(this);
 		this._handleSiblingResize = this._handleSiblingResize.bind(this);
 		this._resizeObserver = null;
 	}
@@ -180,6 +182,8 @@ class AlertToast extends LitElement {
 		super.connectedCallback();
 		document.body.addEventListener('d2l-alert-toast-close', this._handleSiblingResize);
 		document.body.addEventListener('d2l-alert-toast-resize', this._handleSiblingResize);
+		document.body.addEventListener('d2l-alert-toast-timer-start', this._handleSiblingCloseTimerStart);
+		document.body.addEventListener('d2l-alert-toast-timer-stop', this._closeTimerStop);
 		if (mediaQueryList.addEventListener) mediaQueryList.addEventListener('change', this._handlePageResize);
 
 		await this.updateComplete;
@@ -191,6 +195,8 @@ class AlertToast extends LitElement {
 		super.disconnectedCallback();
 		document.body.removeEventListener('d2l-alert-toast-close', this._handleSiblingResize);
 		document.body.removeEventListener('d2l-alert-toast-resize', this._handleSiblingResize);
+		document.body.removeEventListener('d2l-alert-toast-timer-start', this._handleSiblingCloseTimerStart);
+		document.body.removeEventListener('d2l-alert-toast-timer-stop', this._closeTimerStop);
 		if (this._resizeObserver) this._resizeObserver.disconnect();
 		if (mediaQueryList.removeEventListener) mediaQueryList.removeEventListener('change', this._handlePageResize);
 	}
@@ -269,6 +275,7 @@ class AlertToast extends LitElement {
 	}
 
 	_closeTimerStop() {
+		if (!this.open) return;
 		clearTimeout(this._setTimeoutId);
 	}
 
@@ -300,6 +307,11 @@ class AlertToast extends LitElement {
 		));
 	}
 
+	_handleSiblingCloseTimerStart(e) {
+		if (!this.open || e?.detail.hasFocus || e?.detail.hasMouse) return;
+		this._closeTimerStart();
+	}
+
 	_handleSiblingResize(e) {
 		if (e?.target === this || !this.open) return;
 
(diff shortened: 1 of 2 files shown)


components/alert/test/alert-toast.vdiff.js
```diff
@@ -95,5 +118,40 @@ describe('alert-toast', () => {
 			await expect(document).to.be.golden();
 		});
 
+		it('hover then wait', async() => {
+			const elem = await fixture(multipleAlertsAutoClose, { viewport: { width: 700, height: 400 } });
+			await openAlerts(elem);
+			await hoverAlert(elem);
+			await aTimeout(4100);
```
margaree, line 125: Toasts pause for 4000ms before closing. This is currently non-configurable. If we wanted these vdiff tests to run faster than we could add a property for that (could not document it if we don't want it used, but there's always that risk).

margaree: If we want vdiff tests for this then something would have to be done because the tests are timing out. Options:
1. No vdiffs, test this another way?
2. Up the timeout only for these tests if possible
3. Property for close timer duration (https://github.com/BrightspaceUI/core/blob/main/components/alert/alert-toast.js#L264)

Let me know if you have any thoughts on this

components/alert/alert-toast.js
```diff
@@ -180,6 +182,8 @@ class AlertToast extends LitElement {
 		super.connectedCallback();
 		document.body.addEventListener('d2l-alert-toast-close', this._handleSiblingResize);
 		document.body.addEventListener('d2l-alert-toast-resize', this._handleSiblingResize);
+		document.body.addEventListener('d2l-alert-toast-timer-start', this._handleSiblingCloseTimerStart);
+		document.body.addEventListener('d2l-alert-toast-timer-stop', this._closeTimerStop);
```
margaree, line 186: Let me know if there are potentially better ways of doing this than event listeners. I feel like I just keep piling on the event listeners.

components/alert/test/alert-toast.vdiff.js
dlockhart: I'd recommend trying Sinon's `useFakeTimers` to advance the clock -- the calendar vdiffs show an example. Just make sure you only mock out `setTimeout` because otherwise the fixture won't mount.

margaree: Ah right good call!

components/alert/alert-toast.js
dlockhart: Events seem like an excellent choice to me!

components/alert/alert-toast.js
```diff
@@ -21,6 +21,9 @@ const TOAST_SPACING_SMALL = 0.3;
 
 const mediaQueryList = window.matchMedia('(max-width: 615px)');
 
+let ALERT_HAS_FOCUS = false; // if this alert or sibling alert is focused on
+let ALERT_HAS_HOVER = false; // if this alert or sibling alert is hovered on
```
margaree, line 25: An example usecase (added a visual diff for this) is:
- toast 1 has focus
- toast 2 is hovered on then hover is remove. 

Without being aware of the state of toast 1 (i.e., that some toast is either being focused or hovered on), toast 2 would just close.

The code changes below do assume that only one toast can be hovered or focused on at once.

components/alert/alert-toast.js
```diff
@@ -331,18 +355,29 @@ class AlertToast extends LitElement {
 	}
 
 	_onFocus() {
+		ALERT_HAS_FOCUS = true;
```
dlockhart, line 358: Presumably when you're tabbing from one alert to the next, the blur for the previous one fires first (setting `ALERT_HAS_FOCUS` to `false`) and _then_ the focus for the next one fires (setting it back to `true`)?

components/alert/alert-toast.js
dlockhart: At first I was confused as to how globals could work for this, but they do since at most one can be hovered and/or focused at once -- cool!

components/alert/test/alert-toast.vdiff.js
```diff
@@ -95,5 +105,68 @@ describe('alert-toast', () => {
...
+
+			beforeEach(() => {
+				clock = sinon.useFakeTimers({ toFake: ['clearTimeout', 'setTimeout'] });
+			});
+			afterEach(() => {
+				clock.restore();
+			});
+
+			it('hover then wait', async() => {
+				const elem = await fixture(multipleAlertsAutoClose, { viewport: { width: 700, height: 400 } });
```
dlockhart, line 135: Nit: I'd suggest pulling `{ width: 700, height: 400 }` into a `viewport` const to reuse across these.

```suggestion
				const elem = await fixture(multipleAlertsAutoClose, { viewport });
```

dbatiste:
Unrelated to your change - I noticed that `alert-toast.html` specifies `has-close-button` for a couple of toasts (one being for the multiple toast demo snippet), but `d2l-alert-toast` does not support a `has-close-button` attribute... it supports `hide-close-button`. If you're making any further changes, it would be good to fix that up.

In contrast, `d2l-alert` _does_ support the `has-close-button` attribute. I'm guessing the slight difference between them is the cause of this.

dlockhart (approved):
Looks good!

components/alert/alert-toast.js
margaree: Good question - tested this out (Chrome, Safari, Firefox) and it does seem to be the case that consistently blur will happen first and set `ALERT_HAS_FOCUS` to `false` and then focus will happen on the next alert

PR #4121, opened 2023-09-26
US152268 - Migrate selection vdiff tests

svanherk:
Nothing crazy here.  I broke the tests into two separate `describe`s to help with organizing, and I'm expecting some changes based on hover behavior.  Will call that out in the goldens PR.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/ca87a1d76ba9dc27dfffaf97a48985ca059bbe34/2023-09-26_19-53-47/report/


components/button/button-subtle.js
```diff
@@ -208,7 +208,7 @@ class ButtonSubtle extends ButtonMixin(LitElement) {
 				</span>
 			</button>
 			${this.description ? html`<span id="${this._describedById}" hidden>${this.description}</span>` : null}
-			${this.disabled && this.disabledTooltip ? html`<d2l-tooltip for="${this._buttonId}">${this.disabledTooltip}</d2l-tooltip>` : ''}
+			${this.disabled && this.disabledTooltip ? html`<d2l-tooltip class="vdiff-target" for="${this._buttonId}">${this.disabledTooltip}</d2l-tooltip>` : ''}
```
dlockhart, line 211: Aside: I love how easy it's been to "fix" this kind of thing -- awesome job with that feature!

components/selection/test/selection.vdiff.js
```diff
@@ -0,0 +1,313 @@
...
+			});
+
+			[
+				{ name: 'requires-selection-none', selectionInfo: { state: 'none', keys: [] } },
+				{ name: 'requires-selection-some', selectionInfo: { state: 'some', keys: [] } },
+				{ name: 'requires-selection-all', selectionInfo: { state: 'all', keys: [] } }
+			].forEach(({ name, selectionInfo }) => {
+				it(name, async() => {
+					const elem = await fixture(requiresSelectionTemplate);
+					elem.selectionInfo = selectionInfo;
```
dlockhart, line 72: I guess it doesn't need it, but surprised this doesn't require a `await elem.updateComplete` after setting the info.

components/selection/test/selection.vdiff.js
```diff
@@ -0,0 +1,313 @@
...
+
+	describe('radio', () => {
+		const defaultTemplate = html`<d2l-test-selection selection-single><d2l-selection-input label="item 1" key="key1"></d2l-selection-input></d2l-test-selection>`;
+		const selectedTemplate = html`<d2l-test-selection selection-single><d2l-selection-input label="item 1" key="key1" selected></d2l-selection-input></d2l-test-selection>`;
+
+		[
+			{ name: 'default', template: defaultTemplate },
+			{ name: 'focus', template: defaultTemplate, action: focusElem },
+			{ name: 'click', template: defaultTemplate, action: clickElem },
+			{ name: 'space', template: defaultTemplate, action: elem => sendKeysElem(elem, 'press', 'Space') },
```
dlockhart, line 153: Does this `await` the `sendKeysElem` in the action?

components/button/button-subtle.js
svanherk: Thanks, I've also been really happy with it!

components/selection/test/selection.vdiff.js
svanherk: Yeah, it _might_ be because calculating the rect forces a paint with the `getBoundingClientRect` calls? Although the old visual diffs also didn't need it

components/selection/test/selection.vdiff.js
svanherk: From what I can tell, _yes_. The above and this seem to be equivalent:
```
action: async(elem) => await sendKeysElem(elem, 'press', 'Space')
```

But if I add `{ }`, then no it does not wait:
```
action: elem => { sendKeysElem(elem, 'press', 'Space'); }
```

svanherk: If we'd rather be explicit we can add the `async` and `await`, although I'll probably need to go back and fix up a few other files, I've been using this single function passthrough strategy for a while now lol

dlockhart: Ah ok, that's good to know. I think it's OK to use the shorthand now that it's clear what it's doing.

PR #4129, opened 2023-09-27
US152268 - Migrate tag-list vdiff tests

svanherk:
So because the old test was using an `html` `font-size` of `16px`, and our new tests use `20px`, the diffs here are not going to be super helpful.  The tags have grown in size and we'll have to compare the old and the new images.

Since they weren't going to be helpful anyways, I made some changes I'll call out in the goldens PR.

Final report: https://vdiff.d2l.dev/BrightspaceUI/core/38a08ce141f2f4ed587df4f9fe2aff5926df39aa/2023-09-27_21-46-57/report/


dlockhart (approved):
Looks great!

margaree (approved):
LGTM!

PR #4132, opened 2023-09-28
US157857: List Item: Color Indicator: Skeletonize

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F722721154055)

We don't explicitly extend `SkeletonMixin` in `list-item-mixin`, but it's getting it through checkbox mixin and expand/collapse mixin. Let me know if it would be best to also extend it in `list-item-mixin`.


dbatiste:
> Let me know if it would be best to also extend it in list-item-mixin.

Yeah, if `ListItemMixin` depends on it then I think it should be added so that there wouldn't be a regression if it were removed from the others.

PR #4150, opened 2023-09-29
Date & date-range  input vdiffs

dlockhart:
These were... a grind. Notes in the vdiff PR.


components/inputs/test/input-date.visual-diff.js
```diff
@@ -1,576 +0,0 @@
...
-				const rect = await getRect(page, '#placeholder');
-				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-			});
-
-			it('open with click', async function() {
-				await openClick(page, '#placeholder');
-				const rect = await getRect(page, '#placeholder');
-				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
-			});
-		});
```
svanherk, line 382: Seems like these ones weren't moved, was that on purpose?

dlockhart: Oh yep sorry, I removed them purposefully because it didn't seem worth it to test what the opened dropdown looked like based on it having a placeholder or empty text... they were all the same.

svanherk (approved):

PR #4153, opened 2023-09-29
US157414: List item: Try to improve resize handling

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F721347677647)

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -57,10 +57,6 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			/**
 			 * @ignore
 			 */
-			breakpoint: { type: Number },
-			/**
-			 * @ignore
-			 */
 			breakpoints: { type: Array },
 			/**
 			 * A color indicator to appear at the beginning of a list item. Expected value is a valid 3, 4, 6, or 8 character CSS color hex code (e.g., #006fbf).
@@ -82,10 +78,6 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			 * @type {'normal'|'none'}
 			 */
 			paddingType: { type: String, attribute: 'padding-type' },
-			/**
-			 * @ignore
-			 */
-			slimColor: { type: Boolean, attribute: 'slim-color', reflect: true },
 			_displayKeyboardTooltip: { type: Boolean },
 			_hasColorSlot: { type: Boolean, reflect: true, attribute: '_has-color-slot' },
 			_hovering: { type: Boolean, reflect: true },
@@ -105,6 +97,11 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			:host {
 				display: block;
 				position: relative;
+				--d2l-list-item-color-border-radius: var(--d2l-list-color-border-radius, 6px);
+				--d2l-list-item-color-width: var(--d2l-list-color-width, 6px);
+				--d2l-list-item-illustration-margin-inline-end: var(--d2l-list-illustration-margin-inline-end, 0.9rem);
+				--d2l-list-item-illustration-max-height: var(--d2l-list-illustration-max-height, 2.6rem);
+				--d2l-list-item-illustration-max-width: var(--d2l-list-illustration-max-width, 4.5rem);
 			}
 			:host[hidden] {
 				display: none;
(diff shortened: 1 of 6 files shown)


components/list/list-item-mixin.js
```diff
@@ -483,15 +414,13 @@ export const ListItemMixin = superclass => class extends composeMixins(
...
-	}
+	/** TODO: remove this once usage of breakpoints on d2l-list-item has been moved to the list */
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
 
-	updated(changedProperties) {
-		super.updated(changedProperties);
-		if (changedProperties.has('breakpoints')) {
-			this.resizedCallback(this.offsetWidth);
+		if (changedProperties.has('breakpoints') && this.breakpoints) {
```
margaree, line 421: This is the temporary code to make this change backwards compatible until the one usage of breakpoints is updated. This code only handles breakpoints being set initially (can move this to `updated` to make it a bit safer there - the usage does not seem to update it ever) and then once `breakpoints` is changed, it sets it on the parent list. This isn't the most performant, since for each list item it would be checking for the parent having `breakpoints`.

components/list/list.js
```diff
@@ -47,7 +62,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 			 * @type {'all'|'between'|'none'}
 			 * @default "all"
 			 */
-			separators: { type: String, reflect: true }
+			separators: { type: String, reflect: true },
+			_breakpoint: { type: Number, reflect: true },
+			_slimColor: { type: Boolean, reflect: true, attribute: '_slim-color' }
```
margaree, line 67: Let me know if there's a better/preferred way to do this than internal properties

components/list/list.js
```diff
@@ -143,6 +197,13 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		`;
 	}
 
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('breakpoints') && changedProperties.get('breakpoints') !== undefined) {
+			this.resizedCallback(this.offsetWidth, true);
+		}
```
margaree, line 204: The intention of this is to only call `resizedCallback` not on the initial setting of `breakpoints` since it would be getting called anyways by the `ResizeObserver`

components/list/list-item-mixin.js
```diff
@@ -103,14 +88,18 @@ export const ListItemMixin = superclass => class extends composeMixins(
...
 
 	static get styles() {
 
 		const styles = [ css`
 			:host {
+				--d2l-list-item-color-border-radius: var(--d2l-list-color-border-radius, 6px);
+				--d2l-list-item-color-width: var(--d2l-list-color-width, 6px);
+				--d2l-list-item-illustration-margin-inline-end: var(--d2l-list-illustration-margin-inline-end, 0.9rem);
+				--d2l-list-item-illustration-max-height: var(--d2l-list-illustration-max-height, 2.6rem);
+				--d2l-list-item-illustration-max-width: var(--d2l-list-illustration-max-width, 4.5rem);
```
margaree, line 102: Let me know if I'm doing this totally wrong. It seems to work but I could see this not being done in a best practice way.

dbatiste: Do we need these `--d2l-list-item-*` variables? i.e. can the styles below just use the variables inherited from and defined by `d2l-list`?

components/list/list.js
dbatiste: If we want to avoid sprouting these private properties on the `host` then I think we'd have to render a class on the `div`, but then elements in the other slots would not inherit the variables. Will any of these slots need them?

components/list/list-item-mixin.js
margaree: Ah yup seems to work without - will fix that!

components/list/list.js
margaree: Hmm it's possible we'd want `_slimColor` for the `controls` in order to leave correct spacing at slim width. Currently I think it only leaves correct spacing at larger width.

dbatiste: In that case we'd either need to keep sprouting on the `host` like this, or introduce a new internal wrapper `div`. Not sure if introducing a new `div` will break anything.

dbatiste: Another way to avoid sprouting would be to set the variables through JavaScript using `this.style.setProperty('--d2l-list-color-width', '3px');`. I'm not sure I like that any better though... I'd probably just sprout if it came to that.

margaree: Quickly adding a wrapper div does break almost all the visual diffs but they're just all 1px taller, which isn't terrible but also scares me a bit

dbatiste (approved):
Looks good.

If you still have a test page with lots of items, it would be interesting to see if you notice a perf improvement.

margaree:
What I've noticed is just on the `list-nested` page with the `All Iterations` demo snippet the improvement is huge - on `main` I can visibly see all the list items load, but with these changes loading is instant.

PR #4158, opened 2023-10-02
US157414: List Item: Remove breakpoints

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F721347677647)

`breakpoints` has been moved from the `list-item` to `list` in `organizations` https://github.com/BrightspaceHypermediaComponents/organizations/pull/287 so we can cleanup the `breakpoints` implementation in `list-item`.

This is technically a breaking change but unless something super crazy was done in an implementation or someone external was using this, the only usage of `breakpoints` on `d2l-list-item` has been removed.


dbatiste (approved):
🎉

PR #4165, opened 2023-10-03
Allow native button in localizeHTML.

dbatiste:
The `d2l-file-input` component requires a special button rendered in the upload instruction. In order to facilitate localizing this text, we need to add native `button` to the allowed list.

image

### DIFF
--- a/mixins/localize/localize-mixin.js
+++ b/mixins/localize/localize-mixin.js
@@ -231,7 +231,7 @@ export const LocalizeMixin = superclass => class extends _LocalizeMixinBase(supe
 
 };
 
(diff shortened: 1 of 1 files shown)


bearfriend (approved):
I don't love this, but I can't think of a way around it. Also, I'm probably overthinking it. Just would be unfortunate to end up with lots of actual button-looking buttons in the middle of langterms.

dbatiste:
> I don't love this, but I can't think of a way around it. Also, I'm probably overthinking it. Just would be unfortunate to end up with lots of actual button-looking buttons in the middle of langterms.

Agreed. I initially began by adding support for `d2l-button-subtle`, then realized that this case is extra special.

PR #4180, opened 2023-10-06
US152268 - Cleanup old visual-diff workflow, files, and setup

svanherk:
Cleanup the old visual diff workflow and some random references, more info down below.


.gitattributes
svanherk, line 1: I don't actually know what this file does - don't we control this with `files` in `package.json`?

components/dialog/test/dialog-ifrau-contents.vdiff.html
```diff
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-	<link rel="stylesheet" href="../../../test/styles.css" type="text/css">
```
svanherk, line 4: This was the last remaining usage of this, so moved it right in here. This is a very specific case, needing to load an html file - all our other vdiff tests have this handled by the library automatically.

rollup/rollup.config.js
```diff
@@ -12,20 +12,19 @@ const buildDate = Intl.DateTimeFormat('en-CA', { timeZone: 'America/Toronto' }).
...
 	'!**/*.@(js|md|json)',
-	'!**/screenshots/**/*',
 	'!**/golden/**/*',
 ];
 
 export default {
 	input: glob.sync(jsGlob),
 	output: { dir: 'build', format: 'es', preserveModules: true },
-	external: ['puppeteer', '@brightspace-ui/visual-diff', '@brightspace-ui/testing', 'sinon'],
+	external: ['puppeteer', '@brightspace-ui/testing', 'sinon'],
```
svanherk, line 27: Do I still need `puppeteer` here?

.gitattributes
dlockhart: I think by "deployment" they mean "git archive" vs. NPM deployment which will use `package.json`.

rollup/rollup.config.js
dlockhart: Nothing should be using it anymore, so should be safe to remove.

dlockhart (approved):
🎉

.gitattributes
svanherk: Ah ok interesting, what have we used `git archive` for? Or it's just set up nicely in case we ever do?

dlockhart: lol I have no idea why it's there... it's been here since the repo was created in 2019. My vote would be to just delete the file!

svanherk: I can definitely do that too!

dlockhart (approved):

PR #4181, opened 2023-10-06
Fix firefox demo snippet fullscreen

bearfriend:
I'm not even sure why I added this selector originally. Seems quite unnecessary. Firefox doesn't support `:has()` and breaks the whole rule.


svanherk (approved):
Nice!

PR #4187, opened 2023-10-11
fix: Bump ifrau to latest version

EvanLemieux:


dlockhart (approved):
Evan, do you want me to pause the BSI update job while all these repos are updated? That way once they're all updated and released, it can resume and get the changes together.

EvanLemieux:
> Evan, do you want me to pause the BSI update job while all these repos are updated? That way once they're all updated and released, it can resume and get the changes together.

If that's an option and it makes things break less then perhaps that's a good path forward.

PR #4192, opened 2023-10-12
Force offscreen `direction` to match the document

bearfriend:
[DE55375](https://rally1.rallydev.com/#/?detail=/defect/728711712205&fdp=true): New link offscreen text causes RTL issues in user profile card

Instead of using javascript/lit to set a static `direction` style (or attempting to listen and force updates), we can simply set a custom CSS property (variable) with the correct direction and use that keep all `offscreen` elements updated.

Requires https://github.com/BrightspaceUI/testing/pull/216 for `vdiff` to pass


components/html-block/test/html-block.vdiff.js
```diff
@@ -280,7 +280,8 @@ describe('d2l-html-block', () => {
 			html`<d2l-html-block style="width: 650px;" html="An equation...${mathInline} embedded inline with text, and showing placement of indicies for summations."></d2l-html-block>`,
 			{ mathjax: { renderLatex: true } }
 		);
-		await waitUntil(() => elem.clientHeight === 61);
+		console.log(elem.clientHeight, ', expected 61');
+		//await waitUntil(() => elem.clientHeight === 61);
```
dlockhart, line 284: You may want to put the console inside the `waitUntil` -- that way you see the different values it's getting over time.

bearfriend: Yeah I wasn't expecting them to change after being non-zero, but I'll make some changes to see what the final result is.

components/offscreen/offscreen.js
```diff
@@ -1,15 +1,20 @@
 import { css, html, LitElement } from 'lit';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
-export const offscreenStyles = css`
-	.d2l-offscreen {
+const offscreenStyleDeclarations = css`
+		direction: var(--d2l-document-direction, ${document.dir === 'rtl' ? css`rtl` : css`ltr`});
```
dlockhart, line 5: This is a cool solution! Just so I make sure I have this correct:
- `direction` will use `--d2l-document-direction` and fall back to `document.dir` if that isn't set... which would only be true if typography isn't included. What happens if typography is included _after_ `offscreen.js`, does that still work and use the variable?
- `--d2l-document-direction` dynamically changes when the `dir` attribute on the `html` element changes. I'm assuming that means any changes MUST modify the attribute only, and can't set the property or `document.dir` directly.
- `direction` in this style block controls `inset-inline-start` but has no effect on `left` which is only used as a fallback anyway for browsers who don't support logical properties.

components/offscreen/offscreen.js
```diff
@@ -1,15 +1,20 @@
...
 		position: absolute !important;
 		white-space: nowrap;
 		width: 1px;
+`;
+
+export const offscreenStyles = css`
+	.d2l-offscreen {
+		${offscreenStyleDeclarations}
 	}
 	:host([dir="rtl"]) .d2l-offscreen {
```
svanherk, line 19: We can just remove this now right (and the one below)?

dlockhart: Assuming `direction` only affects the logical properties, I think it's still acting as a fallback for older browsers.

bearfriend: Maybe, but if anyone is using `dir` (like not from the mixin) on a component, then removing this could break stuff I guess.

svanherk: Yeah, based on the slack thread I thought we were ok with moving to "If your browser doesn't support logical properties and you're using RTL mode, you will now see the big scrollbar"

dlockhart: Yeah I'm OK with making that change.

components/offscreen/offscreen.js
```diff
@@ -5,7 +5,7 @@ const offscreenStyleDeclarations = css`
 		direction: var(--d2l-document-direction, ${document.dir === 'rtl' ? css`rtl` : css`ltr`});
 		height: 1px;
 		inset-inline-start: -10000px;
-		left: -10000px;
```
dlockhart, line 8: Wondering if we should at least leave `left: -10000px`... removing the RTL stuff will have minimal impact for old browsers, but removing this will cause the text to become visible.

bearfriend: The line below is just making it `left` or `right` based on the initial `document.dir`

dlockhart: Ah nice, missed that!

dlockhart (approved):

svanherk (approved):

PR #4195, opened 2023-10-12
fix: alias runConstructor to @brightspace-ui/testing version

dlockhart:
Now that `runConstructor` exists in `@brightspace-ui/testing`, it doesn't need to be in core anymore. But lots of places are referencing it, so this creates an alias.

Also switches everything in core over to use the testing version which was quite tedious!

### DIFF
--- a/tools/constructor-test-helper.js
+++ b/tools/constructor-test-helper.js
@@ -1,7 +1,1 @@
-import { expect } from '@open-wc/testing';
-
-export function runConstructor(nodeName) {
-	const ctor = customElements.get(nodeName);
-	expect(ctor).to.not.be.undefined;
-	document.createElement(nodeName);
-}
+export { runConstructor } from '@brightspace-ui/testing';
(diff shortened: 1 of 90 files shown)


tools/constructor-test-helper.js
```diff
@@ -1,7 +1 @@
-import { expect } from '@open-wc/testing';
-
-export function runConstructor(nodeName) {
-	const ctor = customElements.get(nodeName);
-	expect(ctor).to.not.be.undefined;
-	document.createElement(nodeName);
-}
+export { runConstructor } from '@brightspace-ui/testing';
```
dlockhart, line 1: Hmm, I think I have to make `@brightspace-ui/testing` a non-dev dependency if we want to do this actually. Thoughts? The other option is to leave the copy here.

svanherk: Hmm, I don't really want it in BSI... I think I'd vote to just keep the copy

dlockhart: Yeah agreed. Ok I've restored the copy -- technically there's was a dependency on `@open-wc/testing` in the same way... but chances are consumers have it already.

svanherk: Was just gonna ask how that was working lol, but right, they'll have that. But likely not our new library yet 👍

svanherk (approved):

PR #4199, opened 2023-10-13
fix: wait for hidden items to appear before focusing

dlockhart:
I definitely don't love this solution, so interested in other ideas as well.

With Chrome 118, when filter tags are chomped (or unchomped) and focus is moved from the "show more"/"show less" button to either the last item before showing more or the "show more" button the focus now fails.

I _think_ it's because the chomping is triggering the resize observer which in turn does a bunch of recalculation and in the process actually sets `visibility: hidden` on everything. And browsers don't like to focus on hidden things and if a hidden thing has focus, they remove it. What I can't explain is why this just started failing or why it's not failing in other browsers. I also can't find a good way to test if it's about to happen, and no amount of `requestAnimationFrame`s waited long enough -- I needed the longer timeout.

I also played around with just keeping focus on the "show more" / "show less" button and it experienced the same issue because of the `visibility: hidden` thing as well.

### DIFF
--- a/components/filter/test/filter-tags.vdiff.js
+++ b/components/filter/test/filter-tags.vdiff.js
@@ -2,7 +2,7 @@ import '../filter.js';
 import '../filter-dimension-set.js';
 import '../filter-dimension-set-value.js';
 import '../filter-tags.js';
-import { aTimeout, clickElem, expect, fixture, html, sendKeys, sendKeysElem } from '@brightspace-ui/testing';
+import { aTimeout, clickElem, expect, fixture, html, oneEvent, sendKeys, sendKeysElem } from '@brightspace-ui/testing';
 
 const singleFilter = html`
 	<d2l-filter id="filter">
@@ -97,7 +97,10 @@ describe('filter-tags', () => {
 				const tagList = filterTags.shadowRoot.querySelector('d2l-tag-list');
 				const showMoreButton = tagList.shadowRoot.querySelector('.d2l-tag-list-button');
 
-				if (showMoreButton) await clickElem(showMoreButton);
+				if (showMoreButton) {
+					clickElem(showMoreButton);
+					await oneEvent(tagList, 'd2l-tag-list-focus');
+				}
 				await expect(elem).to.be.golden();
 			});
 		});
--- a/components/tag-list/tag-list.js
+++ b/components/tag-list/tag-list.js
@@ -407,11 +407,13 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 
 		if (!this.shadowRoot) return;
 
+		const isMoreButton = e.target.classList.contains('d2l-tag-list-button-show-more');
 		await this.updateComplete;
-		if (e.target.classList.contains('d2l-tag-list-button-show-more')) this._items[this._chompIndex].focus();
-		else {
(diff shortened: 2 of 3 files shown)


margaree:
Bleh yeah your analysis looks right - it's running `_toggleHiddenTagVisibility` first which is triggering the focus but then a resize and a slot change are happening which set visibility to hidden back and forth and also recalculate the `_chompIndex` (I think because there's potentially edge cases where it would change between `_toggleHiddenTagVisibility` and the `resize`, though then the focus would be on the wrong item?). It looks like the same sequence is happening on Firefox but it keeps the focus state (feels like maybe it shouldn't?), whereas chrome no longer does.

It feels like this could use a rewrite to simplify it, but also it was super complicated so I don't know much that would be possible.

Maybe we could keep track of the fact that a `tag-list-item` should have focus, as set in `_toggleHiddenTagVisibility`, then when `_contentReady` is true we focus on `this._items[this._chompIndex], and when blur happens we set it to false. 

It looks like a `blur` is being triggered very first thing when the "Show More" button is clicked so this strategy could work (also it looks like this is happening on chrome but not firefox, so maybe related to the problem?) but also feels like it adds another layer of complexity.

I also don't love the timeout but it does feel like the correct behaviour is waiting for everything else to be done before focusing, I'm just also not sure what the best way is of doing that.

dlockhart:
Thanks for checking @margaree -- that's exactly what I was seeing. If we're wanting to do some of those larger efforts, we may want to merge a temporary fix like this and then maybe @dbatiste you could add it to the Fixer queue?

Haha and now the vdiffs are failing because it's waiting so long that for that special tag consumer the item becomes visible and the mouse is over it (because it clicked "show more" which was in the same spot) so the dropdown appears, then the focus happens which hides the dropdown. Are we OK with this change to the vdiffs?

svanherk:
> Haha and now the vdiffs are failing because it's waiting so long that for that special tag consumer the item becomes visible and the mouse is over it (because it clicked "show more" which was in the same spot) so the dropdown appears, then the focus happens which hides the dropdown. Are we OK with this change to the vdiffs?

Grrr, can we move the mouse right after click?

dlockhart:
> Can we move the mouse right after click?

Haha it looks like yes -- so the diff will be the same but the behaviour is slightly different, since I'm expecting it to be very common for the mouse to stay over the item that's about to receive focus.

svanherk:
> Haha it looks like yes -- so the diff will be the same but the behaviour is slightly different, since I'm expecting it to be very common for the mouse to stay over the item that's about to receive focus.

Hmm, right, which is kinda weird behaviour.  As the user, I wouldn't really expect the focus change to close my tooltip

dbatiste:
Yeah I'm also not in love with waiting an arbitrary timeout. I was also thinking that maybe we could wait (in a more definitive way) until all this chomping business is complete before trying to apply focus.

dbatiste:
Just spitballing here but I wonder if we had some internal promise or event that we could wait for chomping to complete. You two have obviously stepped through it so maybe you've already considered that.

Anyway, I'm fine with this as a temporary solution and we can add it to the fixer queue. I might be able to spend some time on it depending on which issues Nathan prioritizes.

dlockhart:
Ok I've created a Fixer story in Rally to come up with a better long-term solution. If folks are OK with this temporarily, can I get a ✅ ?

PR #4203, opened 2023-10-17
Add option to specify a provider delegate.

dbatiste:
This PR updates the `provideInstance` helper method to optionally accept a `ProviderDelegate` that will be invoked when a value is requested.

### DIFF
--- a/mixins/provider/provider-mixin.js
+++ b/mixins/provider/provider-mixin.js
@@ -1,9 +1,20 @@
+export class ProviderDelegate {
+	constructor(delegate) {
+		this.delegate = delegate;
+	}
+}
+
 export function provideInstance(node, key, obj) {
 	if (!node._providerInstances) {
 		node._providerInstances = new Map();
 		node.addEventListener('d2l-request-instance', e => {
 			if (node._providerInstances.has(e.detail.key)) {
-				e.detail.instance = node._providerInstances.get(e.detail.key);
+				const instance = node._providerInstances.get(e.detail.key);
+				if (instance instanceof ProviderDelegate) {
+					e.detail.instance = instance.delegate();
+				} else {
+					e.detail.instance = instance;
+				}
 				e.stopPropagation();
 			}
 		});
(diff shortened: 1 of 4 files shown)


mixins/provider/provider-mixin.js
```diff
@@ -1,9 +1,20 @@
...
+}
+
 export function provideInstance(node, key, obj) {
 	if (!node._providerInstances) {
 		node._providerInstances = new Map();
 		node.addEventListener('d2l-request-instance', e => {
 			if (node._providerInstances.has(e.detail.key)) {
-				e.detail.instance = node._providerInstances.get(e.detail.key);
+				const instance = node._providerInstances.get(e.detail.key);
+				if (instance instanceof ProviderDelegate) {
```
dbatiste, line 13: I initially implemented this to just check for `Function`, but changed to accept `ProviderDelegate` because there a couple of existing usages of `provideInstance` that already specify a function. I think it is safer to accept this specific class, leaving the option to consumers to provide an actual function to requestors that doesn't get invoked by us here.

dlockhart: Yeah I was thinking the same... in retrospect this should have just always taken a function (which could have in turn returned a function), but we're kind of stuck with this now.

mixins/provider/test/provider-mixin.test.js
```diff
@@ -0,0 +1,54 @@
...
+		}));
+		expect(requestInstance(document.body, 'instance-delegate')).to.equal(42);
+	});
+
+	it('should provide async delegate result', async() => {
+		provideInstance(document, 'instance-async-delegate', new ProviderDelegate(() => {
+			return new Promise(resolve => {
+				setTimeout(() => resolve(42), 200);
+			});
+		}));
```
dlockhart, line 50: To avoid this test taking > 200ms, you could use Sinon's `useFakeTimers` to advance the clock.

dbatiste: Cool yeah, I removed the `setTimeout` altogether. I think simply returning a `Promise` is sufficient to test this cas.

PR #4216, opened 2023-10-24
Simplify d2l-test-runner.config.js

bearfriend:
Now that `d2l-test-runner` supports implicit groups and excludes files in `node_modules`, we can simplify the `pattern` and get rid of the explicit `axe` group.

### DIFF
--- a/d2l-test-runner.config.js
+++ b/d2l-test-runner.config.js
@@ -1,10 +1,3 @@
-const pattern = type => `+(components|controllers|directives|helpers|mixins|templates)/**/*.${type}.js`;
-
 export default {
-	pattern,
-	groups: [{
-		name: 'aXe',
-		files: pattern('axe'),
-		browsers: ['chrome']
-	}]
+	pattern: type => `**/test/*.${type}.js`
 };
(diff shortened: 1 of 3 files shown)


d2l-test-runner.config.js
```diff
@@ -1,10 +1,3 @@
-const pattern = type => `+(components|controllers|directives|helpers|mixins|templates)/**/*.${type}.js`;
-
 export default {
-	pattern,
-	groups: [{
-		name: 'aXe',
-		files: pattern('axe'),
-		browsers: ['chrome']
-	}]
+	pattern: type => `**/test/*.${type}.js`
```
svanherk, line 2: Are we sure there's no nested test files?

bearfriend: Yeah, I compared the lists from `npm run test:* -- --watch`. We could still expand it, though.

svanherk: I just know we've accidentally not run all our tests before, but probably not an issue if you're adding _new_ tests. But maybe

svanherk (approved):

PR #4219, opened 2023-10-25
Add Maori

bearfriend:
[US156660](https://rally1.rallydev.com/#/?detail=/userstory/720165179691&fdp=true): Core

Adds Maori support by importing the new shared list from `intl`


dlockhart (approved):
🎉

PR #4224, opened 2023-10-31
Only trap focus if last activated trap

dbatiste:
This PR updates `d2l-focus-trap` to only trap focus if it's the last activated focus trap. This prevents focus traps that are not nested from stealing focus and causing an infinite loop. 

While nested focus traps are already supported and handled (for nested `d2l-dialog-*` components), it is possible for consumers to use a `d2l-dialog` in their app component, which subsequently uses an LMS/legacy helper to render a `d2l-dialog-fullscreen` at the root. In such a case, the dialog components are not nested, and therefore they compete for focus in an endless loop.

### DIFF
--- a/components/focus-trap/focus-trap.js
+++ b/components/focus-trap/focus-trap.js
@@ -4,6 +4,8 @@ import { FocusMixin } from '../../mixins/focus/focus-mixin.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { isComposedAncestor } from '../../helpers/dom.js';
 
+const traps = [];
+
 /**
  * A generic container component to trap user focus.
  * @fires d2l-focus-trap-enter - Dispatched when focus enters the trap. May be used to override initial focus placement when focus enters the trap.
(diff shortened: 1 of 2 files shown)


components/focus-trap/test/focus-trap.test.js
dbatiste, line 1: You'll likely want to hide whitespace for this change. All that was added was the siblings fixture and a test for it. The rest of the white-space is just nesting one level deeper to keep the creation of the two different fixtures separate.

components/focus-trap/focus-trap.js
```diff
@@ -68,6 +70,17 @@ class FocusTrap extends FocusMixin(LitElement) {
...
 	}
 
+	willUpdate(changedProperties) {
+		if (!changedProperties.has('trap')) return;
+		if (this.trap) {
+			traps.push(this);
+		} else {
+			const trapIndex = traps.findIndex(trap => trap === this);
+			if (trapIndex === -1) return;
+			traps.splice(trapIndex, 1);
```
dbatiste, line 80: A `pop()` could be used here instead, but I felt the look-up and `splice()` was safer in case multiple dialogs are closed at once and their traps are removed in an order that doesn't match our `traps` array.

dlockhart (approved):

PR #4226, opened 2023-10-31
fix: tag-list-item: use focus-visible

margaree:
[Rally defect](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F720939316847)

### DIFF
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -81,12 +81,12 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 				padding: 0.25rem 0.6rem;
 				text-overflow: ellipsis;
 			}
-			:host(:focus-within) .tag-list-item-container,
-			:host(:focus-within:hover) .tag-list-item-container {
+			.tag-list-item-container:has(:focus-visible),
+			:host(:hover) .tag-list-item-container:has(:focus-visible) {
 				box-shadow: inset 0 0 0 2px var(--d2l-color-celestine), 0 2px 4px rgba(0, 0, 0, 0.03);
 			}
 			:host(:hover) .tag-list-item-container,
-			:host(:focus-within) .tag-list-item-container {
+			.tag-list-item-container:has(:focus-visible) {
 				background-color: var(--d2l-color-sylvite);
 			}
 			:host(:hover) .tag-list-item-container {
--- a/components/tag-list/test/tag-list.vdiff.js
+++ b/components/tag-list/test/tag-list.vdiff.js
@@ -43,6 +43,13 @@ describe('tag-list', () => {
 			await expect(elem).to.be.golden();
 		});
 
+		it('click focus', async() => {
+			const elem = await fixture(createTagList(), { viewport });
+			const firstChild = elem.querySelector('d2l-tag-list-item');
+			clickElem(firstChild);
+			await expect(elem).to.be.golden();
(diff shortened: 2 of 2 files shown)


components/tag-list/tag-list-item-mixin.js
```diff
@@ -81,12 +81,12 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 				padding: 0.25rem 0.6rem;
 				text-overflow: ellipsis;
 			}
-			:host(:focus-within) .tag-list-item-container,
-			:host(:focus-within:hover) .tag-list-item-container {
+			.tag-list-item-container:has(:focus-visible),
```
dbatiste, line 84: The `:has(:focus-visible)` is really awesome for dealing with this scenario. If we're going to use it though, we should couple it with `@supports` for backwards compatibility.

```css
foo1:has(:focus-visible) {
	...
}
@supports not selector(:has(a, b)) {
	foo1:focus-within {
		...
	}
}
```
The above is concise, but for the best backwards compatibility we probably would reverse it because I think the `selector` option is a bit newer. Unfortunately that makes the CSS a little less tidy though...
```css
foo1:focus-within {
	... // "fallback" styles
}
@supports selector(:has(a, b)) {
	foo1:focus-within {
		... // undo the above since we don't really want those focus styles if `:has` is supported
	}
	foo1:has(:focus-visible) {
		...
	}
}
```

margaree: Thanks! That's really helpful.

One thing I'm seeing with `:focus-visible` is that on Safari, if I click one of the tag list items, giving it focus, then press the right or left arrow key, those items do not get the focus ring. On Chrome they do, and they do if I tab into them on Safari then they do. From some searching it looks like Safari handles keyboard focus on click differently than Chrome, but I'm not sure. Have you seen this before? Is this just expected?

dbatiste: Np!  No, I haven't seen that before. I just did a quick check and it looks like  `d2l-file-input`'s usage of `:has(:focus-visible)` is working in Safari as expected. I also did a quick check with `d2l-tabs` since it uses arrow keys to manage focus along with `:focus-visible`, and it also appears to be working as expected as well. 🤔

margaree: Oh interesting, I just tested Safari with tabs and I'm seeing the same issue as with tag-list. Maybe it's my browser version? I'm on Safari 16.6

dbatiste: Oh I was doing something slightly different - I was tabbing into the list from outside (rather than clicking a tab), and then using the arrows. 

I think the browser tracks whether the last interaction was keyboard for the purpose of `:focus-visible`, and it would seem that Safari doesn't regard arrow keys as keyboard interactions, so its state would just be whatever it was when focus entered into the list of tabs. So if tabbing in and then using arrow keys, then it works, but if we click in and then use arrow keys, it doesn't. Whereas, it would seem Chrome regards a broader set of keys as keyboard interactions.

I wonder if there is a way we can trick Safari into being in keyboard mode. For example, before we move focus in response to the arrow keys, perhaps we can send a key event of some type that will force Safari's state. I think we do something similar in the testing library. Like a shift/ctrl/cmd key or something.

margaree:
Closing for now. This PR is referenced in the Rally defect.

PR #4238, opened 2023-11-03
Fix d2l-list-item-drag-handle mouse interaction.

dbatiste:
[DE55479](https://rally1.rallydev.com/#/?detail=/defect/729253956433&fdp=true)

This PR fixes an issue where the `d2l-list-item-drag-handle`'s up/down arrows are not clickable.

These arrows should be clickable with the mouse... but they are no longer. Clicking them causes the drag handle to revert its rendering back to drag mode (the dots icon).
image

For Chrome and Firefox, this appears to be a regression from this recent change: https://github.com/BrightspaceUI/core/pull/4100.

For Safari, there's more - it also suffers from [this long-standing terrible issue that Apple refuses to fix](https://bugs.webkit.org/show_bug.cgi?id=22261), where clicking a button that has focus will actually cause it to lose focus, which in this case, causes the drag-handle to toggle out of the `d2l-button-move` interaction.

### DIFF
--- a/components/list/list-item-drag-drop-mixin.js
+++ b/components/list/list-item-drag-drop-mixin.js
@@ -335,6 +335,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 					opacity: 0;
 				}
 				:host([selected]) d2l-list-item-drag-handle,
+				d2l-list-item-drag-handle:hover,
 				d2l-list-item-drag-handle.d2l-hovering,
 				d2l-list-item-drag-handle.d2l-focusing {
 					opacity: 1;
--- a/components/list/list-item-drag-handle.js
+++ b/components/list/list-item-drag-handle.js
@@ -72,7 +72,7 @@ class ListItemDragHandle extends LocalizeCoreElement(FocusMixin(RtlMixin(LitElem
 			:host {
 				display: flex;
 				margin: 0.25rem;
-				pointer-events: all; /* required since its parent may set point-events: none; (see generic layout) */
+				pointer-events: auto; /* required since its parent may set point-events: none; (see generic layout) */
 			}
 			:host([hidden]) {
 				display: none;
--- a/components/list/list-item-drag-handle.js
+++ b/components/list/list-item-drag-handle.js
@@ -72,6 +72,7 @@ class ListItemDragHandle extends LocalizeCoreElement(FocusMixin(RtlMixin(LitElem
 			:host {
 				display: flex;
 				margin: 0.25rem;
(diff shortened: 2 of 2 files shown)


components/list/list-item-drag-drop-mixin.js
```diff
@@ -335,6 +335,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 					opacity: 0;
 				}
 				:host([selected]) d2l-list-item-drag-handle,
+				d2l-list-item-drag-handle:hover,
```
dbatiste, line 338: I almost wonder if the `pointer-events: none` change in [this PR](https://github.com/BrightspaceUI/core/pull/4100/files) is necessary with this change. I am a little worried about causing a regression on that fix if it's removed however.

components/list/list-item-drag-handle.js
```diff
@@ -72,6 +72,7 @@ class ListItemDragHandle extends LocalizeCoreElement(FocusMixin(RtlMixin(LitElem
 			:host {
 				display: flex;
 				margin: 0.25rem;
+				pointer-events: auto; /* required since its parent may set point-events: none; (see generic layout) */
```
dbatiste, line 75: This makes it so `d2l-list-item-drag-handle`'s child `d2l-button-move` is clickable event though its parent sets `pointer-events: none`. Unfortunately this CSS could not be placed in the same spot next to it, since `::slotted()` does not enable targeting elements deeper than their immediate children.

components/list/list-item-drag-handle.js
```diff
@@ -181,25 +182,6 @@ class ListItemDragHandle extends LocalizeCoreElement(FocusMixin(RtlMixin(LitElem
 		this.activateKeyboardMode();
 	}
 
-	_onKeyboardButtonFocusIn() {
```
dbatiste, line 184: I just renamed these handlers for consistency. I vaguely recall not doing it previously in effort to keep a previous PR smaller.

components/list/list-item-drag-handle.js
```diff
@@ -274,6 +275,10 @@ class ListItemDragHandle extends LocalizeCoreElement(FocusMixin(RtlMixin(LitElem
 
 	}
 
+	_onMoveButtonMouseDown(e) {
+		e.preventDefault();
```
dbatiste, line 279: This is needed to stop Safari from moving focus away from the button when it is clicked due to the frustrating position that Apple has taken.

bearfriend (approved):
Ugh. I hate how Safari has made this all so complicated again. I guess it's still better than the z-index stuff, though.

PR #4261, opened 2023-11-16
fix: validate plainText option asynchronously

dlockhart:
With the recent fix to `@brightspace-ui/testing` that listens for nodes to be removed, it seems just slower enough that this taglist test is failing. It [was introduced here](https://github.com/BrightspaceUI/core/pull/3772).

I'm honestly not sure how this test was actually working before, since `fixture()` is asynchronous and so is `render()` and the whole render cycle. I've never been able to `try`-`catch` exceptions thrown inside `render()`. Also I'd fully expect `this.name` to be `undefined` potentially during first render anyway, and especially so if its set via an async API call.

This adjust the validation of the `plainText` option to be more similar to other "required" attributes, where we give the consumer 3 seconds to set the attribute.


components/tag-list/test/tag-list.test.js
```diff
@@ -133,16 +133,5 @@ describe('d2l-tag-list-item-mixin-consumer', () => {
...
-
-		it('should error if not provided', async() => {
-			let error;
-			try {
-				await fixture(html`<d2l-tag-list-item-mixin-consumer></d2l-tag-list-item-mixin-consumer>`);
-			} catch (e) {
-				error = e;
-			}
-			expect(error?.name).to.equal('TypeError');
-		});
```
dlockhart, line 145: Unfortunately this means we can't test for it any more.

I'm going to see if I can come up with a more reusable/consistent way to validate required things like this, and in doing so maybe make it easier to test for them as well.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -276,4 +277,16 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		`;
 	}
 
+	_validatePlainText() {
+		clearTimeout(this._validatingPlainTextTimeout);
+		// don't error immediately in case it doesn't get set immediately
+		this._validatingPlainTextTimeout = setTimeout(() => {
+			this._validatingPlainTextTimeout = null;
+			const hasPlainText = (typeof this._plainText === 'string') && this._plainText.length > 0;
```
bearfriend, line 285: Just want to note that `typeof new String()` is `'object'`. Unlikely anyone is using it, but 🤷

dlockhart: Fair, and that's ridiculous JavaScript.

dlockhart: Ok updated.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -282,7 +282,7 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		// don't error immediately in case it doesn't get set immediately
 		this._validatingPlainTextTimeout = setTimeout(() => {
 			this._validatingPlainTextTimeout = null;
-			const hasPlainText = (typeof this._plainText === 'string') && this._plainText.length > 0;
+			const hasPlainText = (this._plainText?.constructor !== String) && this._plainText?.length > 0;
```
bearfriend, line 286:
```suggestion
			const hasPlainText = (this._plainText?.constructor === String) && this._plainText?.length > 0;
```

dlockhart: 🤦

bearfriend (approved):

PR #4262, opened 2023-11-16
use a central test eslint config

dlockhart:
Been meaning to do this for a while. This switches to defining the eslint config for tests at the root of the project, instead of needing to copy it into every `test` directory.

This change actually found a misconfiguration as the `async-container` test config was incorrect.


bearfriend (approved):
🎉

PR #4268, opened 2023-11-17
Update note message about site

devpow112:
GitHub removed the legacy way of doing this. [New method](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts).

Example before when note lost all it's fancy styling
image


svanherk (approved):
Thanks!

PR #4270, opened 2023-11-20
Inline Add Button Component

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F730605354321)

This is the initial work on the inline add button component. Its original purpose was for list but it now has the requirement of being a standalone component. I'll add comments throughout, but most of this is up for debate especially naming.

This does not include more detailed documentation on the README (followup story) and also doesn't include the ability to hide when not hovering (will do separately, didn't want to clutter this PR too much).

### DIFF
--- a/components/button/button-add.js
+++ b/components/button/button-add.js
@@ -1,4 +1,5 @@
 import '../tooltip/tooltip.js';
+import '../icons/icon.js';
 import { css, html, LitElement } from 'lit';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { labelStyles } from '../typography/styles.js';
@@ -54,8 +55,11 @@ class ButtonAdd extends LocalizeCoreElement(LitElement) {
 				width: 100%;
 			}
 			:host(:hover) .line,
-			:host(:focus-visible) .line {
+			button:focus .line {
 				border-top-color: var(--d2l-color-celestine);
+			}
+			button:focus d2l-icon {
+				color: var(--d2l-color-celestine);
 			}
 			d2l-icon.icon-no-text {
 				background-color: white;
@@ -87,8 +91,9 @@ class ButtonAdd extends LocalizeCoreElement(LitElement) {
 	constructor() {
 		super();
 
-		this._iconId = getUniqueId();
 		this.visibleText = false;
+		this._iconId = getUniqueId();
+
 		this.addEventListener('click', this._onClick);
 	}
 
@@ -96,7 +101,7 @@ class ButtonAdd extends LocalizeCoreElement(LitElement) {
 		const text = this.text || this.localize('components.button-add.addItem');
 		const label = this.label || this.localize('components.button-add.addItem');
 		return html`
-			<button class="d2l-label-text">
+			<button class="d2l-label-text" aria-label="Add Button">
 				<div class="line"></div>
 				${this.visibleText
 					? html`
@@ -105,8 +110,8 @@ class ButtonAdd extends LocalizeCoreElement(LitElement) {
 							<span>${text}</span>
 						</div>`
 					: html`
-						<d2l-icon icon="tier1:plus-default" id="${this._iconId}" class="icon-no-text"></d2l-icon>
-						<d2l-tooltip for="${this._iconId}">${label}</d2l-tooltip>
+						<d2l-icon icon="tier1:plus-default"  class="icon-no-text"></d2l-icon>
+						<d2l-tooltip class="vdiff-target" offset="15">${label}</d2l-tooltip>
 					`
 				}
 			</button>
--- a/components/button/button-add.js
+++ b/components/button/button-add.js
@@ -53,7 +53,7 @@ class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
 
 			.line {
 				border-top: 1px var(--d2l-button-add-line-style) var(--d2l-color-mica);
-				margin: 3px 0;
+				margin: 3px 0; /** hover/click target */
 				width: 100%;
 			}
 			button:hover .line,
@@ -62,16 +62,15 @@ class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
 			}
 
 			.content {
-				background-color: white;
-				padding: 3px;
-				position: absolute;
 				align-items: center;
+				background-color: white;
 				display: flex;
+				position: absolute;
 			}
 			:host([visible-text]) .content {
 				color: var(--d2l-color-celestine);
 				height: 1.5rem;
-				padding: 0 0.6rem;
+				padding: 0 0.3rem;
 			}
 
 			:host([visible-text]) d2l-icon,
@@ -81,6 +80,8 @@ class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
 			}
 			:host(:not([visible-text])) d2l-icon {
 				color: var(--d2l-color-galena);
+				margin: -3px; /** hover/click target */
+				padding: 3px; /** hover/click target */
 			}
 			:host([visible-text]) d2l-icon {
 				padding-right: 0.2rem;
--- a/components/button/button-add.js
+++ b/components/button/button-add.js
@@ -112,7 +112,7 @@ class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
 					<d2l-icon icon="tier1:plus-default"></d2l-icon>
 					${this.visibleText
 		? html`<span>${text}</span>`
-		: html`<d2l-tooltip class="vdiff-target" offset="18" for="${this._buttonId}">${label}</d2l-tooltip>`}
+		: html`<d2l-tooltip class="vdiff-target" offset="18" for="${this._buttonId}" for-type="label">${label}</d2l-tooltip>`}
 				</div>
 			</button>
 		`;
--- a/components/button/button-add.js
+++ b/components/button/button-add.js
@@ -1,18 +1,16 @@
-import '../tooltip/tooltip.js';
+import '../colors/colors.js';
 import '../icons/icon.js';
+import '../tooltip/tooltip.js';
 import { css, html, LitElement } from 'lit';
+import { FocusMixin } from '../../mixins/focus/focus-mixin.js';
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { labelStyles } from '../typography/styles.js';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
 
 /**
  * A component for quickly adding items to a specific locaiton.
- * TODO:
- * - smarter margin and padding and top
- * - focus
- * - simplify html and css
  */
-class ButtonAdd extends LocalizeCoreElement(LitElement) {
+class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
 	static get properties() {
 		return {
 			/**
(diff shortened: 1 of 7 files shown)


components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+			/**
+			 * The text to show in the button when visible-text is true
+			 * @type {string}
+			 */
+			text: { type: String },
+			/**
+			 * When true, show the button with icon and visible text. When false, only show icon.
+			 * @type {boolean}
+			 */
+			visibleText: { type: Boolean, attribute: 'visible-text' }
```
margaree, line 30: Most basic component states (demo page is also probably more useful for this):

1. Default -> `visible-text` is false, component shows the divider line, an add icon, and a tooltip which contains whatever is in `label` (design refers to this as "hint") OR "Add Item" as a fallback
2. Visible Text -> when `visible-text` is true, the component shows the divider, and in the centre is the plus icon and the text specified in `text` OR "Add Item" as a fallback.

`text` and `label` will not be used at the same time (the tooltip will not be shown if the text is visible and vice versa).

An aria description is potentially going to be used as well, I'll be discussing that more with design.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+		};
+	}
+
+	static get styles() {
+		return [labelStyles, css`
```
margaree, line 35: I didn't use `buttonStyles` here as it felt like it complicated things more than made them easier. Let me know if I should be using that.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+import { css, html, LitElement } from 'lit';
+import { FocusMixin } from '../../mixins/focus/focus-mixin.js';
+import { getUniqueId } from '../../helpers/uniqueId.js';
+import { labelStyles } from '../typography/styles.js';
+import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
+
+/**
+ * A component for quickly adding items to a specific locaiton.
+ */
+class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
```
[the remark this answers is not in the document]
margaree: Also, I didn't use `ButtonMixin`. Let me know if there are other pieces of that which are important to a button that I'm missing.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+			<button class="d2l-label-text" id="${this._buttonId}">
+				<div class="line"></div>
+				<div class="content">
+					<d2l-icon icon="tier1:plus-default"></d2l-icon>
+					${this.visibleText
+		? html`<span>${text}</span>`
+		: html`<d2l-tooltip class="vdiff-target" offset="18" for="${this._buttonId}" for-type="label">${label}</d2l-tooltip>`}
+				</div>
+			</button>
+		`;
```
margaree, line 119: This implements its own button within rather than using subtle and/or icon button because we don't seem to want the focus/hover states that those provide.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	_onClick() {
+		/** Dispatched when click happens */
+		this.dispatchEvent(new CustomEvent('d2l-button-add-click', {
```
margaree, line 124: I made this a component specific click event rather than just firing up `click` since `list` consumers could be listening for this and just having a generic `click` seemed like it could cause confusion.

lang/en.js
```diff
@@ -1,6 +1,7 @@
 export default {
 	"components.alert.close": "Close Alert",
 	"components.breadcrumbs.breadcrumb": "Breadcrumb",
+	"components.button-add.addItem": "Add Item",
```
margaree, line 4: I only did `en` for now in case the component name changes.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+			/**
+			 * The text to show in the button when visible-text is true
+			 * @type {string}
+			 */
+			text: { type: String },
+			/**
+			 * When true, show the button with icon and visible text. When false, only show icon.
+			 * @type {boolean}
+			 */
+			visibleText: { type: Boolean, reflect: true, attribute: 'visible-text' }
```
dbatiste, line 30: Nit: I'd probably name this `text-visible` or `text-hidden` since that sounds more like a boolean than a string. That is also inline with a bunch of other components that have `label-hidden`.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+ * A component for quickly adding items to a specific locaiton.
+ */
+class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
+	static get properties() {
+		return {
+			/**
+			 * The text to be shown in the tooltip as a hint when visible-text is false
+			 * @type {string}
+			 */
+			label: { type: String },
```
dbatiste, line 20: It seems like we might not need this `label` property. The `text-visible` essentially controls whether or not the button is labelled by text within the button or the tooltip contents. It seems like we only need `text` and `text-visible`. Thoughts?

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+			visibleText: { type: Boolean, reflect: true, attribute: 'visible-text' }
+		};
+	}
+
+	static get styles() {
+		return [labelStyles, css`
+			:host {
+				--d2l-button-add-line-style: solid;
+			}
+			button {
```
dbatiste, line 39: I _think_ we probably want `cursor: pointer;` and the `user-select: none;` styles that would normally be applied from `buttonStyles`.

Note: I feel like we can omit the vendor prefixed `*-user-select`. That would only affect super ancient FF, and the consequence of omitting is pretty negligible.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+				outline: none;
+				padding: 0;
+				position: relative;
+				white-space: nowrap;
+				width: 100%;
+			}
+
+			.line {
+				border-top: 1px var(--d2l-button-add-line-style) var(--d2l-color-mica);
+				margin: 3px 0; /** hover/click target */
```
dbatiste, line 56: Is this needed if the line is within the button? I would have thought that the click target would effectively be a rectangle of the button, including the whitespace inside the button above and below the line.

components/button/button-add.js
dbatiste: That makes sense. In that case, we probably want to call `e.stopPropagation();`. I'm kinda surprised we aren't doing that [here](https://github.com/BrightspaceUI/core/blob/main/components/list/list-item-button-mixin.js#L66).

components/button/button-add.js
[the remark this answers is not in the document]
dlockhart: I agree with not using `ButtonMixin` -- it just includes too many `<button>` attributes that would be confusing to handle here.

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+		`];
+	}
+
+	constructor() {
+		super();
+
+		this.visibleText = false;
+		this._buttonId = getUniqueId();
+
+		this.addEventListener('click', this._onClick);
```
dlockhart, line 98: Could this not be a `@click` handler on `<button>` element?

components/button/button-add.js
```diff
@@ -0,0 +1,131 @@
...
+	static get focusElementSelector() {
+		return 'button';
+	}
+
+	render() {
+		const text = this.text || this.localize('components.button-add.addItem');
+		const label = this.label || this.localize('components.button-add.addItem');
+
+		return html`
+			<button class="d2l-label-text" id="${this._buttonId}">
```
dlockhart, line 110: I don't know how I feel about using label styles for non-label things... I could see us want to change these in the future and regret applying them to places like this.

dlockhart, line 110: Technically the `id` is only needed when `visibleText` is `false`. 🤷

components/button/button-add.js
dlockhart: I'd like to explore this a little bit... there are lots of components that fire click events, including just clicking on a `<span>` or whatever, so hoping we don't need to worry about this.

Either way, does it need to bubble and cross shadow root boundaries?

components/button/test/button-add.axe.js
```diff
@@ -0,0 +1,40 @@
...
+	it('visible text', async() => {
+		const el = await fixture(html`<d2l-button-add visible-text></d2l-button-add>`);
+		await expect(el).to.be.accessible();
+	});
+
+	it('focused', async() => {
+		const el = await fixture(html`<d2l-button-add></d2l-button-add>`);
+		setTimeout(() => el.focus());
+		await oneEvent(el, 'focus');
+		await expect(el).to.be.accessible();
```
dlockhart, line 30: This uses the actual keyboard instead of the JS API which won't trigger `:focus-visible`.

```suggestion
		const el = await fixture(html`<d2l-button-add></d2l-button-add>`);
		await focusElem(el);
		await expect(el).to.be.accessible();
```

components/button/test/button-add.test.js
```diff
@@ -0,0 +1,24 @@
...
+			runConstructor('d2l-button-add');
+		});
+
+	});
+
+	describe('events', () => {
+
+		it('dispatches d2l-button-add-click event when clicked', async() => {
+			const el = await fixture(html`<d2l-button-add></d2l-button-add>`);
+			setTimeout(() => el.click());
```
dlockhart, line 18: Same here for the mouse.

```suggestion
			setTimeout(() => clickElem(el));
```

components/button/button-add.js
margaree: I'm good with that since we don't use both text and label at the same time.

components/button/button-add.js
margaree: It would need to bubble and cross shadow root boundaries. The plan is that to use this consumers of list would have something like:
`<d2l-list add-button>...</d2l-list>`
and then the `button-add` component would appear after each list item. We don't want consumers to have to manage adding this component themselves after each item, and behaviour is designed to be consistent (i.e., wouldn't have some list items with `button-add` and others that don't have it, etc.).

All that to say, for the list case, consumers of list would want to listen for this event and then handle the click of the add button however they wish to.

dlockhart: Gotcha. In that case, another option: since list is going to be the one rendering `<d2l-button-add>`, it could wire up to its click event and then fire its own `d2l-list-add-button-click` event.

dlockhart: So consumers would go `<d2l-list add-button @d2l-list-add-button-click="{...}">`.

margaree: I like that idea - I think that would be more straightforward for consumers.

components/button/button-add.js
margaree: It is - the click target for the line does end up being just the line and so we want that buffered a bit (3 px on each side). The design also specifies that we don't want the click target for the line to be the same height as the "+" / "+ Add Item" part.

components/button/button-add.js
```diff
@@ -0,0 +1,125 @@
...
+				width: 100%;
+			}
+
+			.line {
+				border-top: 1px var(--d2l-button-add-line-style) var(--d2l-color-mica);
+				margin: 3px 0; /** hover/click target */
+				width: 100%;
+			}
+			button:hover .line,
+			button:focus .line {
```
dbatiste, line 57: Should this be:
```css
button:hover .line,
button:${unsafeCSS(getFocusPseudoClass())} .line {
	border-top-color: var(--d2l-color-celestine);
}
```

components/button/button-add.js
```diff
@@ -14,25 +15,20 @@ class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
...
-			 * The text to be shown in the tooltip as a hint when visible-text is false
-			 * @type {string}
-			 */
-			label: { type: String },
-			/**
-			 * The text to show in the button when visible-text is true
+			 * When text-visible is true, the text to show in the button. When text-visible is false, the text to show in the tooltip.
 			 * @type {string}
 			 */
 			text: { type: String },
```
dlockhart, line 21: Once the `PropertyRequiredMixin` stuff merges, this should likely use it.

components/button/test/button-add.vdiff.js
```diff
@@ -2,29 +2,29 @@ import '../button-add.js';
...
+				describe(category, () => {
+					[
+						{ name: 'normal' },
+						{ name: 'hover', action: hoverElem },
+						{ name: 'focus', action: focusElem },
+						{ name: 'click', action: clickElem }
+					].forEach(({ action, name }) => {
+						it(name, async() => {
+							let elem = await fixture(template);
+							if (elem.tagName !== 'D2L-BUTTON-ADD') elem = elem.querySelector('d2l-button-add');
```
dlockhart, line 22: Is this ever not the case?

lang/en.js
dlockhart: The linter isn't going to let you merge as-is. Typically we'll just use `mvf` to copy the English value everywhere.

margaree: I thought I was going to manually have to update them so I was waiting to make sure the component name was good to go, but yay for automation! 🎉 Will fix

components/button/button-add.js
```diff
@@ -7,18 +7,19 @@ import { getFocusPseudoClass } from '../../helpers/focus.js';
...
-class ButtonAdd extends FocusMixin(LocalizeCoreElement(LitElement)) {
+class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(LitElement))) {
 	static get properties() {
 		return {
 			/**
 			 * When text-visible is true, the text to show in the button. When text-visible is false, the text to show in the tooltip.
 			 * @type {string}
 			 */
-			text: { type: String },
+			text: { type: String, required: true },
```
dlockhart, line 22: Yay!

dlockhart (approved):
Cool!

components/button/button-add.js
```diff
@@ -0,0 +1,127 @@
...
+	static get focusElementSelector() {
+		return 'button';
+	}
+
+	render() {
+		const text = this.text || this.localize('components.button-add.addItem');
+		const id = !this.textVisible ? this._buttonId : undefined;
+
+		return html`
+			<button class="d2l-label-text" id="${ifDefined(id)}">
```
dlockhart, line 114: Just remembered randomly: make sure you set `type="button"` on this, otherwise it defaults to `submit` and will submit forms!

margaree: Good catch, thanks! Will fix

PR #4272, opened 2023-11-21
PropertyRequiredMixin

dlockhart:
This is a mixin that can be leveraged whenever a component has a property/attribute that they want to treat as "required" -- something the developer consuming the component must set. This typically comes up on the accessibility side where we want to ensure labels are provided.

I've only implemented this for `String` properties, because I wasn't sure what it would mean for a `Boolean` (present or not?) or a `Number` to be "required". If we have actual use cases, I'm happy to add support for that.


components/inputs/input-color.js
```diff
@@ -277,8 +276,6 @@ class InputColor extends FocusMixin(FormElementMixin(LocalizeCoreElement(LitElem
 
 		super.updated(changedProperties);
 
-		if (changedProperties.has('label') || changedProperties.has('type')) this._validateLabel();
```
dlockhart, line 280: This use case forced me to add support for "dependent properties" -- essentially other properties to watch and re-validate when they change.

components/inputs/test/input-color.test.js
```diff
@@ -128,4 +129,35 @@ describe('d2l-input-color', () => {
 
 	});
 
+	describe('validation', () => {
```
dlockhart, line 132: Example of how a component can now test that the validation is set up properly.

mixins/labelled/labelled-mixin.js
```diff
@@ -96,20 +97,23 @@ export const LabelledMixin = superclass => class extends superclass {
...
+			message: defaultMessage => {
+				if (!this.labelledBy) return defaultMessage;
+				return `LabelledMixin: "${this.tagName.toLowerCase()}" is labelled-by="${this.labelledBy}", but its label is empty`;
+			},
+			validator: hasValue => {
+				if (!this.labelRequired || hasValue) return true;
+				if (!this.labelledBy) return false;
+				return this._labelElem !== null;
+			}
+		});
```
dlockhart, line 110: This is probably the most complex use-case imaginable with a custom validator and custom message (sometimes).

mixins/property-required/property-required-mixin.js
```diff
@@ -0,0 +1,108 @@
...
+			dependentProps: opts.dependentProps,
+			message: opts.message,
+			thrown: false,
+			timeout: null,
+			validator: opts.validator
+		});
+
+	}
+
+	flushRequiredPropertyErrors() {
```
dlockhart, line 69: This method is just here for testing, allowing them to bypass the 3s delay and have the exceptions be thrown in their context.

mixins/property-required/test/property-required-mixin.test.js
```diff
@@ -0,0 +1,219 @@
...
+);
+
+const tagString = defineCE(
+	class extends PropertyRequiredMixin(LitElement) {
+		static properties = {
+			attr: { type: String }
+		};
+		constructor() {
+			super();
+			this.addRequiredProperty('attr');
```
dlockhart, line 16: An alternate API I considered would be to make it part of the property defintion, like this:

```javascript
static properties = {
  attr: { type: String, required: true }
};
```

It's a nice API, but doesn't leave any room for custom validators or messages. We _could_ optionally make it an object:

```javascript
static properties = {
  attr1: { type: String, required: true },
  attr2: {
    type: String,
    required: {
      validator: () => {...},
      message: () => {...}
    }
};
```

Thoughts?

mixins/property-required/test/property-required-mixin.test.js
```diff
@@ -0,0 +1,219 @@
...
+		);
+		const elem = await fixture(`<${tag} attr1="value" attr2="valid"></${tag}>`);
+		expect(() => elem.flushRequiredPropertyErrors()).to.not.throw();
+		elem.removeAttribute('attr2');
+		await elem.updateComplete;
+		expect(() => elem.flushRequiredPropertyErrors())
+			.to.throw(TypeError, createDefaultMessage(tag, 'attr1'));
+	});
+
+	it('should work in a subclass/mixin', async() => {
```
dlockhart, line 197: This was a crazy case. When mixins or inheritence are present, the static `properties` need to be collected by traversing the inheritence tree. Loosly inspired by @eKoopmans' [provider stuff](https://github.com/Brightspace/unified-content-experience/blob/main/src/utilities/provider-framework/provider-mixin/provider-mixin.js#L41).

mixins/property-required/property-required-mixin.js
```diff
@@ -0,0 +1,108 @@
...
+			info.thrown = true;
+			const defaultMessage = createDefaultMessage(this.tagName.toLowerCase(), info.attrName);
+			throw new TypeError(info.message(defaultMessage));
+		}
+
+	}
+
+	_initProperties(base) {
+		if (base === null) return;
+		this._initProperties(Object.getPrototypeOf(base));
```
eKoopmans, line 96: Neat, I figured there had to be a better/cleaner option than `__proto__`.

dlockhart: Yeah was going to mention this option to you, as when I was looking up `__proto__` the Mozilla docs had all kinds of scary deprecation warnings!

mixins/property-required/test/property-required-mixin.test.js
dbatiste: I don't have strong feeling one way or the other. I tend to cringe a bit when property definitions get super bloated, but having it all in one place is nice, and it needs to get defined _somewhere_. Mixing with the Lit-specific API feels sorta uncomfortable, but the mixin is Lit-specific anyway, so it's not like a non-Lit component would be able to use this, if such a thing existed in our system. 🤷‍♂️

dbatiste: I wonder if Lit would ever introduce `required` as a property of properties. 🤔

dbatiste: Yeah. Similarly, the properties could be prefixed with `d2l` but I'm not in love with that either.

My feeling is that is were worried about it, we should just avoid mixing with their API altogether. I'm not particularly worried, but it could happen. Or unlikely as it might be, Lit could one day do something internally to mangle these.

dlockhart: Agreed. OK I'll put up a PR those shows what leveraging the existing properties might look like... if it's feasible technically I think I'd prefer that.

dlockhart: We're going to go with the alternate approach, which has now been merged into this branch.

eKoopmans: BTW I know I set a precedent on overloading this `properties` data, but I've been planning to leave it again - if we move the "providers" thing to a controller instead of a mixin, then it'll make sense to declare the "fromProvider" properties when creating the controller (i.e. in the constructor).

dlockhart: Cool, yeah I think controllers make a lot of sense for your case since provider properties are by their nature not something consumers of the component should/can be setting anyway.

mixins/property-required/property-required-mixin.js
```diff
@@ -0,0 +1,107 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+
+const TIMEOUT_DURATION = 3000;
+export const ERROR_CODE = '(@brightspace-ui/core:PropertyRequiredMixin)';
```
dlockhart, line 4: This is the first time we've done this, but what do folks think about including an error code in the exceptions that we throw? This would allow us to search JavaScript errors in the logs for `@brightspace-ui/core:` or for errors thrown from this mixin specifically. Because the error messages contain tag and property names, it would be difficult to search for them otherwise.

dlockhart:
This is ready for review. As I mentioned in standup, I'm going to do a follow-up that converts more things to use this.

helpers/error.js
dlockhart, line 1: After chatting with @dbatiste, we decided that it would be useful to centralize the formatting of errors originating from custom elements. That way we can ensure they're formatted consistently, always include the tag name in the same way and can optionally include the entire composed path.

This error now looks like:
> TypeError: <d2l-input-color>: "label" attribute is required. Path: "d2l-demo-page > main > div > slot > d2l-demo-snippet > div > div > div > slot > div > d2l-input-color". (@brightspace-ui/core:PropertyRequiredMixin)

dbatiste: I'm guessing there isn't actually a double `: :`... or is there?

Anyway, I think this path will be really helpful to locate the source of some issues. 🎉

dlockhart: Haha there is not a double `:`, not sure where that copy-paste mistake came from!

dbatiste (approved):
Looks great!

PR #4273, opened 2023-11-22
PropertyRequiredMixin Option 2

dlockhart:
This is another flavour of #4272.

Instead of consumers calling `addRequiredProperty('name, opts)` in their constructor, this lets them define that a property is required right in the `static properties` Lit definition.

```javascript
static properties = {
  someProp: { type: String, required: true }
};
```

If options are needed, then `required` can be an object:

```javascript
static properties = {
  someProp: {
    type: String,
    required: {
      validator: () => ...
    }
  }
};
```

### DIFF
--- a/components/inputs/input-color.js
+++ b/components/inputs/input-color.js
@@ -104,7 +104,13 @@ class InputColor extends PropertyRequiredMixin(FocusMixin(FormElementMixin(Local
 			 * REQUIRED: Label for the input, comes with a default value for background & foreground types.
 			 * @type {string}
 			 */
-			label: { type: String },
+			label: {
+				type: String,
+				required: {
+					dependentProps: ['type'],
+					validator: (_value, elem, hasValue) => elem.type !== 'custom' || hasValue
+				}
+			},
 			/**
 			 * Hides the label visually
 			 * @type {boolean}
@@ -233,10 +239,6 @@ class InputColor extends PropertyRequiredMixin(FocusMixin(FormElementMixin(Local
 		this._missingLabelErrorHasBeenThrown = false;
 		this._opened = false;
 		this._value = undefined;
-		this.addRequiredProperty('label', {
-			dependentProps: ['type'],
-			validator: hasValue => this.type !== 'custom' || hasValue
-		});
 	}
 
 	get associatedValue() { return this._associatedValue; }
--- a/mixins/property-required/property-required-mixin.js
+++ b/mixins/property-required/property-required-mixin.js
@@ -6,19 +6,14 @@ export function createDefaultMessage(tagName, propertyName) {
 	return `${tagName}: "${propertyName}" attribute is required.`;
 }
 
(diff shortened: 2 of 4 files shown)


components/inputs/input-color.js
```diff
@@ -104,7 +104,13 @@ class InputColor extends PropertyRequiredMixin(FocusMixin(FormElementMixin(Local
 			 * REQUIRED: Label for the input, comes with a default value for background & foreground types.
 			 * @type {string}
 			 */
-			label: { type: String },
+			label: {
+				type: String,
+				required: {
+					dependentProps: ['type'],
+					validator: (_value, elem, hasValue) => elem.type !== 'custom' || hasValue
```
dlockhart, line 111: We now need to pass `elem` to the validator and message options since `properties` is `static` and doesn't have access to `this`.

mixins/property-required/property-required-mixin.js
```diff
@@ -6,19 +6,14 @@ export function createDefaultMessage(tagName, propertyName) {
 	return `${tagName}: "${propertyName}" attribute is required.`;
 }
 
-export function createUndefinedPropertyMessage(tagName, propertyName) {
-	return `PropertyRequiredMixin: "${tagName.toLowerCase()}" has no property "${propertyName}".`;
-}
```
dlockhart, line 11: Bonus: requiring a property that isn't defined is no longer a possibility, so this error case goes away.

mixins/property-required/property-required-mixin.js
```diff
@@ -80,13 +71,18 @@ export const PropertyRequiredMixin = dedupeMixin(superclass => class extends sup
 		clearTimeout(info.timeout);
 		info.timeout = null;
 
-		const hasValue = this[name]?.constructor === String && this[name]?.length > 0;
-		const success = info.validator(hasValue);
+		if (info.type !== undefined && info.type !== String) {
```
dlockhart, line 74: I moved the type checking here since otherwise there was no way to test it (custom elements constructors that throw errors can't be caught).

mixins/property-required/test/property-required-mixin.test.js
```diff
@@ -81,88 +73,70 @@ describe('PropertyRequiredMixin', () => {
...
 		);
-		const elem = await fixture(`<${tag}></${tag}>`);
-		expect(() => elem.addRequiredProperty('attr'))
+		const elem = await fixture(`<${tag} attr="value"></${tag}>`);
+		expect(() => elem.flushRequiredPropertyErrors())
 			.to.throw(createInvalidPropertyTypeMessage(tag, 'attr'));
 	});
 
-	it('should use custom message', async() => {
+	it('should not throw if type is undefined (defaults to String)', async() => {
```
dlockhart, line 98: Added a test case for if they don't include a `type`, which the Lit docs say defaults to `String`.

dlockhart:
We're going to go with this solution -- so will merge this into the other branch.

PR #4274, opened 2023-11-22
Plugins helpers

dbatiste:
[US160427](https://rally1.rallydev.com/#/?detail=/userstory/731548187981&fdp=true)

This PR adds plugins helpers - `registerPlugin`, `getPlugins`, and `getPlugin` to enable implementors to define plugins without consumers of those plugins being required to import them directly. BSI or a higher level module is responsible for importing the registrations.

Other options were considered, including deferring the registrations or by accepting an async provider method for the plugin, however we do not want plugin implementors assuming that if the registration is deferred, that it would be safe to synchronously import dependencies within the registration module. Typically, we expect projects registrations to be aggregated into a single public export, which means that unrelated plugins could be registered (and synchronous imports loaded) as soon as the registration module is requested. The implementor mindset should be geared towards deferring its dependencies until needed, as shown in the readme.

Required for: https://github.com/BrightspaceUI/lms-core/pull/61

### DIFF
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -210,7 +210,7 @@ registerPlugin('foo-plugins', { prop1: 'other value' }, { key: 'key-2', sort: 2 
 
 // Defer loading dependencies until needed
 registerPlugin('foo-plugins', { getRenderer: async () => {
-    return (await import('./some-module.js').catch(() => {}))?.renderer
+    return (await import('./some-module.js')).renderer
 }});
 ```
 
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -214,16 +214,16 @@ registerPlugin('foo-plugins', { getRenderer: async () => {
 }});
 ```
 
-The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request the plugin by using `getPlugin` and specifying the plugin set key and plugin key.
+The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request the plugin by using `tryGetPluginByKey` and specifying the plugin set key and plugin key.
 
 ```js
-import { getPlugin, getPlugins } from '@brightspace-ui/core/helpers/plugins.js';
+import { getPlugins, tryGetPluginByKey } from '@brightspace-ui/core/helpers/plugins.js';
 
 // Call getPlugins to get plugins
 const plugins = getPlugins('foo-plugins');
 
-// Call getPlugin to get a specific plugin by key
-const plugin = getPlugin('foo-plugins', 'key-1');
+// Call tryGetPluginByKey to get a specific plugin by key
+const plugin = tryGetPluginByKey('foo-plugins', 'key-1');
 ```
 
 ## queueMicrotask
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -191,31 +191,35 @@ const offsetParent = getLegacyOffsetParent(element);
 
 Plugin helpers provide a way for modules to implement and register objects that can be plugged into other project's modules without requiring the plugin consumer to import those modules or objects directly. A higher order module (ex. in BSI) is responsible for importing the plugin registrations.
 
-The plugin implementor uses the `registerPlugin` helper method to make its implementation available to interested consumers. The implementor provides a key for the set of plugins in which to register.
+The plugin implementor uses the `registerPlugin` helper method to make its implementation available to interested consumers. The implementor provides a key for the set of plugins in which to register, and the plugin.
 
-Optionally, an object to specify a `key` for the plugin and/or the `sort` value may be provided. They `key` is useful if consumers intend to request a specific plugin, while the `sort` is useful in cases where order of plugins is important to consumers.
+Optionally, an object to specify a `key` for the plugin and/or the `sort` value may be provided. The `key` is useful if consumers intend to request a specific plugin, while the `sort` is useful in cases where the order of plugins is important to consumers. If `sort` is not specified for at least one plugin, they will be provided to consumers in registration order.
 
 **Important!** plugin registrations should defer loading their dependencies using dynamic imports. They should **not** be synchronously imported in the registration module.
 
 ```js
 import { registerPlugin } from '@brightspace-ui/core/helpers/plugins.js';
 
-// Provide plugin set key, plugin, and optionally a key and/or sort value
+// Provide plugin set key, plugin
+registerPlugin('foo-plugins', { prop1: 'some value' });
+registerPlugin('foo-plugins', { prop1: 'other value' });
+
+// Optionally provide key and/or value
 registerPlugin('foo-plugins', { prop1: 'some value' }, { key: 'key-1', sort: 1 });
 registerPlugin('foo-plugins', { prop1: 'other value' }, { key: 'key-2', sort: 2 });
 
 // Defer loading dependencies until needed
-registerPlugin('foo-plugins', { getRenderer: () => {
+registerPlugin('foo-plugins', { getRenderer: async () => {
     return (await import('./some-module.js').catch(() => {}))?.renderer
 }});
 ```
 
-The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request it by using `getPlugin` and specifying the plugin set key and plugin key.
+The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request the plugin by using `getPlugin` and specifying the plugin set key and plugin key.
 
 ```js
 import { getPlugin, getPlugins } from '@brightspace-ui/core/helpers/plugins.js';
 
-// Call getPlugins to get a sorted array of plugins
+// Call getPlugins to get plugins
 const plugins = getPlugins('foo-plugins');
 
 // Call getPlugin to get a specific plugin by key
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -204,7 +204,7 @@ import { registerPlugin } from '@brightspace-ui/core/helpers/plugins.js';
 registerPlugin('foo-plugins', { prop1: 'some value' });
 registerPlugin('foo-plugins', { prop1: 'other value' });
 
-// Optionally provide key and/or value
+// Optionally provide key and/or sort value
 registerPlugin('foo-plugins', { prop1: 'some value' }, { key: 'key-1', sort: 1 });
 registerPlugin('foo-plugins', { prop1: 'other value' }, { key: 'key-2', sort: 2 });
 
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -187,6 +187,41 @@ import { getLegacyOffsetParent } from '@brightspace-ui/core/helpers/offsetParent
 const offsetParent = getLegacyOffsetParent(element);
 ```
 
+## Plugins
+
+Plugin helpers provide a way for modules to implement and register objects that can be plugged into other project's modules without requiring the plugin consumer to import those modules or objects directly. A higher order module (ex. in BSI) is responsible for importing the plugin registrations.
+
+The plugin implementor uses the `registerPlugin` helper method to make its implementation available to interested consumers. The implementor provides a key for the set of plugins in which to register.
+
+Optionally, an object to specify a `key` for the plugin and/or the `sort` value may be provided. They `key` is useful if consumers intend to request a specific plugin, while the `sort` is useful in cases where order of plugins is important to consumers.
+
+**Important!** plugin registrations should defer loading their dependencies using dynamic imports. They should **not** be synchronously imported in the registration module.
+
+```js
+import { registerPlugin } from '@brightspace-ui/core/helpers/plugins.js';
+
+// Provide plugin set key, plugin, and optionally a key and/or sort value
+registerPlugin('foo-plugins', { prop1: 'some value' }, { key: 'key-1', sort: 1 });
+registerPlugin('foo-plugins', { prop1: 'other value' }, { key: 'key-2', sort: 2 });
+
+// Defer loading dependencies until needed
+registerPlugin('foo-plugins', { getRenderer: () => {
+    return (await import('./some-module.js').catch(() => {}))?.renderer
+}});
+```
+
+The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request it by using `getPlugin` and specifying the plugin set key and plugin key.
+
+```js
+import { getPlugin, getPlugins } from '@brightspace-ui/core/helpers/plugins.js';
+
+// Call getPlugins to get a sorted array of plugins
+const plugins = getPlugins('foo-plugins');
+
+// Call getPlugin to get a specific plugin by key
+const plugin = getPlugin('foo-plugins', 'key-1');
+```
+
 ## queueMicrotask
 
 A polyfill for [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask). For more information on microtasks, read [this article from Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide).
--- a/helpers/plugins.js
+++ b/helpers/plugins.js
@@ -1,6 +1,8 @@
 const pluginSets = new Map();
+const pluginsRequested = new Set();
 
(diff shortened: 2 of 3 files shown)


helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
+const pluginSets = new Map();
+
+export function getPlugin(setKey, pluginKey) {
```
dbatiste, line 3: In hindsight, `getPlugins` could have been implemented to accept a plugin key. In such a case, it would still return an array, but it would only contain plugins within the set that have a certain plugin key. This would effectively mean that different plugins could legit have the same key, which is _interesting_. We have no use case for this, and we could always add it later.

helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
...
+		pluginSet.plugins.sort((item1, item2) => item1.options.sort - item2.options.sort);
+		pluginSet.requiresSorting = false;
+	}
+	return pluginSet.plugins.map(item => item.plugin);
+}
+
+export function registerPlugin(setKey, plugin, options) {
+	let pluginSet = pluginSets.get(setKey);
+	if (!pluginSet) {
+		pluginSet = { plugins: [], requiresSorting: false };
```
dbatiste, line 21: When I originally wrote this, this wrapper object around the `plugins` array and `requiresSorting` didn't exist, and I was blindly slapping `requiresSorting` onto the native array, which felt a little wrong even though it wasn't causing any issues. 😬 An interesting behaviour that that resulted in was - if another plugin was registered after the consumer called `getPlugins`, it would magically show up in their array because they are referencing the same array.

So a consequence of this wrapper (and the `map(...)` above) is that consumers get their own array, and if plugins are registered after they called `getPlugins`, they won't see it unless they call `getPlugins` again. Feature or bug? Consumers getting their own array seems safer, and all registrations _should_ technically be done by the time anyone requests plugins. (is this assumption true?)

helpers/README.md
```diff
@@ -187,6 +187,45 @@ import { getLegacyOffsetParent } from '@brightspace-ui/core/helpers/offsetParent
 const offsetParent = getLegacyOffsetParent(element);
 ```
 
+## Plugins
+
+Plugin helpers provide a way for modules to implement and register objects that can be plugged into other project's modules without requiring the plugin consumer to import those modules or objects directly. A higher order module (ex. in BSI) is responsible for importing the plugin registrations.
+
+The plugin implementor uses the `registerPlugin` helper method to make its implementation available to interested consumers. The implementor provides a key for the set of plugins in which to register, and the plugin.
```
dlockhart, line 194: Haha, I thought you forgot to end the sentence... "and the plugin..." the plugin does what?

```suggestion
The plugin implementor uses the `registerPlugin` helper method to make its implementation available to interested consumers. The implementor provides a key for the set of plugins in which to register and the plugin implementation.
```

helpers/README.md
```diff
@@ -187,6 +187,45 @@ import { getLegacyOffsetParent } from '@brightspace-ui/core/helpers/offsetParent
...
+registerPlugin('foo-plugins', { prop1: 'some value' });
+registerPlugin('foo-plugins', { prop1: 'other value' });
+
+// Optionally provide key and/or value
+registerPlugin('foo-plugins', { prop1: 'some value' }, { key: 'key-1', sort: 1 });
+registerPlugin('foo-plugins', { prop1: 'other value' }, { key: 'key-2', sort: 2 });
+
+// Defer loading dependencies until needed
+registerPlugin('foo-plugins', { getRenderer: async () => {
+    return (await import('./some-module.js').catch(() => {}))?.renderer
```
dlockhart, line 213: What does the `.catch()` and the `?` accomplish? Wouldn't this swallow a missing file or export and make it more difficult to troubleshoot?

helpers/README.md
```diff
@@ -187,6 +187,45 @@ import { getLegacyOffsetParent } from '@brightspace-ui/core/helpers/offsetParent
...
+The plugin consumer uses the `getPlugins` helper method to get references to the registered plugins by providing a key for the set of plugins. If the consumer knows the key of the plugin it needs, it can request the plugin by using `getPlugin` and specifying the plugin set key and plugin key.
+
+```js
+import { getPlugin, getPlugins } from '@brightspace-ui/core/helpers/plugins.js';
+
+// Call getPlugins to get plugins
+const plugins = getPlugins('foo-plugins');
+
+// Call getPlugin to get a specific plugin by key
+const plugin = getPlugin('foo-plugins', 'key-1');
```
dlockhart, line 226: Thoughts about `tryGetPluginByKey()` as a name?

helpers/plugins.js
dlockhart: Yeah... wondering if for now we should be throwing if two plugins get registered with the same key? It was likely a mistake and would cause a lot of confusion as to why it's not working... and even worse someone could accidentally overwrite someone else's plugin without knowing it.

helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
+const pluginSets = new Map();
+
+export function getPlugin(setKey, pluginKey) {
+	const pluginSet = pluginSets.get(setKey);
```
dlockhart, line 4: Should this `throw` if the set doesn't exist instead of returning `undefined`?

If this does end up being a `tryGet`, I'd suggest returning `null` if the plugin can't be found.

helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
...
+	return pluginSet?.plugins.find(plugin => plugin.options.key === pluginKey)?.plugin;
+}
+
+export function getPlugins(setKey) {
+	const pluginSet = pluginSets.get(setKey);
+	if (!pluginSet) return [];
```
dlockhart, line 10: Again, wondering if this should just `throw`, or do you think the possibility of a plugin set having no implementations is plausible?

helpers/plugins.js
dlockhart: I think consumers having their own copy is good, since theoretically they could sort it themselves in a different way (e.g. alphabetically) by something only they understand and we wouldn't want that impacting others.

helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
...
+	}
+	return pluginSet.plugins.map(item => item.plugin);
+}
+
+export function registerPlugin(setKey, plugin, options) {
+	let pluginSet = pluginSets.get(setKey);
+	if (!pluginSet) {
+		pluginSet = { plugins: [], requiresSorting: false };
+		pluginSets.set(setKey, pluginSet);
+	}
```
dlockhart, line 23: Nit: not sure if this is better or not, but another option that's 1 less line!

```javascript
if (!pluginSets.has(setKey)) {
  pluginSets.set({ plugins: [], requiresSorting: false });
}
const pluginSet = pluginSets.get(setKey);
```

helpers/plugins.js
```diff
@@ -0,0 +1,26 @@
...
+	pluginSet.plugins.push({ plugin, options: Object.assign({ key: undefined, sort: 0 }, options) });
+	pluginSet.requiresSorting = pluginSet.requiresSorting || (options?.sort !== undefined);
```
dlockhart, line 25: Won't `options.sort` never be `undefined`? I guess if they explicitly pass in `{ sort: undefined }` as an option, but that seems weird to do.

helpers/test/plugins.test.js
```diff
@@ -0,0 +1,67 @@
...
+describe('plugins', () => {
+
+	before(() => {
+		registerPlugin('test-plugins', { prop1: 'beer' });
+		registerPlugin('test-plugins', { prop1: 'donuts' });
+		registerPlugin('test-plugins-sorted', { prop1: 'beer' }, { sort: 2 });
+		registerPlugin('test-plugins-sorted', { prop1: 'donuts' }, { sort: 1 });
+		registerPlugin('test-plugins-keyed', { prop1: 'beer' }, { key: 'plugin1' });
+		registerPlugin('test-plugins-keyed', { prop1: 'donuts' }, { key: 'plugin2' });
+	});
```
dlockhart, line 13: Would it make sense to export a `reset` method that empties everything out? That way you could call it in `afterEach` and then each test could be more isolated.

helpers/test/plugins.test.js
```diff
@@ -0,0 +1,67 @@
...
+	it('getPlugins should return array of plugins in new sort order if additional plugins are registered', () => {
+		registerPlugin('test-plugins-sorted-add', { prop1: 'beer' }, { sort: 3 });
+		registerPlugin('test-plugins-sorted-add', { prop1: 'donuts' }, { sort: 1 });
+
+		let plugins = getPlugins('test-plugins-sorted-add');
+		expect(plugins.length).to.equal(2);
+		expect(plugins[0].prop1).to.equal('donuts');
+		expect(plugins[1].prop1).to.equal('beer');
+
+		registerPlugin('test-plugins-sorted-add', { prop1: 'candy apple' }, { sort: 2 });
```
dlockhart, line 43: Question: wondering if registering a plugin _after_ someone has asked for the plugins should be an error state and `throw`? The way we want this working is for BSI/whatever to call into all the registrations first thing before any other code runs, so if someone registers later that to me implies they're doing something naughty.

helpers/test/plugins.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+	it('getPlugin should return undefined for invalid plugin key', () => {
+		const plugin = getPlugin('test-plugins-keyed', 'pluginx');
+		expect(plugin).to.equal(undefined);
+	});
+
+	it('getPlugin should return plugin for specified keys', () => {
+		const plugin = getPlugin('test-plugins-keyed', 'plugin1');
+		expect(plugin.prop1).to.equal('beer');
+	});
```
dlockhart, line 65: Worth adding a test to make sure each place that asks for the plugins gets a copy of the array?

helpers/README.md
dbatiste: Ah, I think I can remove this.

I had copied this from elsewhere, where this was being done:
```javascript
Promise.all([
    import(...).catch(() => {}),
    import(...).catch(() => {})
]);
```
The rationale where it was copied from: `Promise.all` rejects if _any_ of the promises rejects, and in the context of html-block renderer plugins, we felt that ignoring faulty plugin configuration was preferable to not rendering the HTML at all. The cleaner is to use `Promise.allSettled`, but at the time, I think browser support was iffy.

helpers/plugins.js
dbatiste: Yeah, that's a good idea. If a some functionality requires something akin to plugins with the same key, they can just not use the `key` and they can define something on their plugin API.

helpers/README.md
dbatiste: I'm cool with that. It makes it a bit clearer that it may or nay not return a plugin, and it's retrieving it based on key.

dlockhart: Yeah either that, or if we think that the list of plugins is essentially static and plugins won't come and go based on external conditions, then we could keep it as-is and `throw` if a plugin with the requested key isn't there.

helpers/plugins.js
dbatiste: Hmmm, the cases I've been considering so far would always have a respective plugin unless misconfigured. 

But let's say we have some file-type plugins for something, and we're given a file that we don't have a plugin for. In that case, a consumer might try to get the plugin keyed by extension or mime-type, and fallback if necessary. Alternatively, the consumer could get all the plugins, and do the `find` themselves, or iterate over them calling `tryDoSomething()`.

With that in mind, I sort of lean towards `tryGetPluginByKey`, returning `null` if not found. But I could be convinced to go the error route, because as mentioned, the consumer has other means to the same end - it's really just sugar.

helpers/plugins.js
dbatiste: I could go either way on this, but my current feeling is that it's the consumer's decision whether no plugins is an error. Getting no plugins back in my case would certainly be unexpected. But maybe there's a scenario where some behaviours or features are enabled through user configuration, which in turn controls whether certain plugins are registered. In such case, maybe we'd expect no plugins. 🤷‍♂️

helpers/plugins.js
dbatiste: That's a good point.

helpers/plugins.js
dbatiste: Oh yeah... I think I might had that initially, but then switched it to only do the single look-up. The difference is probably negligible for the number of plugin sets we'll have though.

helpers/plugins.js
dbatiste: This `options` variable is the object they passed in, not the result of `Object.assign(target, source)`. The source options remain unchanged,  so I think the only time `options?.sort !== undefined` is when they've passed `{ sort: something }`.

helpers/test/plugins.test.js
dbatiste: Haha, I actually had that method while I was writing the tests, but then I thought to myself... if someone imported that and called it in the app, it would be very bad. But yeah, I'm happy to add that is we prefer. 😄

helpers/test/plugins.test.js
dbatiste: Yep, that's a good test to have since we've established that we don't really want them sharing arrays.

helpers/test/plugins.test.js
dbatiste: Haha, you just reminded me of Austin Powers. If we're gonna throw an error for being naughty, we should at least give them an error message in "English English". 😆 

Anyway, good question. The only scenario I can think of is if someone tried to register new plugin as part of dynamically loading a bunch more stuff after the initial page load. But given that consumers would not be notified that their plugin array is stale, it seems like a muddled situation to say the least.

helpers/plugins.js
dbatiste: Added this along with a couple tests for it.

helpers/test/plugins.test.js
dbatiste: Added this back, and organized tests a bit.

helpers/test/plugins.test.js
dbatiste: Added test for this case.

helpers/plugins.js
dlockhart: Definitely sugar yeah. As is this whole method and key thing anyway. Yeah I'm good with whatever!

helpers/plugins.js
dlockhart: Should be `O(1)` lookup on the `Map` anyway.

helpers/plugins.js
dlockhart: Yep my bad!

helpers/test/plugins.test.js
dlockhart: Cool. Yeah I think it's fine -- I'd love to switch core over to the `exports` model and then we could actually hide it from consumers.

helpers/test/plugins.test.js
```diff
@@ -1,67 +1,104 @@
...
+	describe('keyed', () => {
+
+		beforeEach(() => {
+			registerPlugin('test-plugins', { prop1: 'beer' }, { key: 'plugin1' });
+			registerPlugin('test-plugins', { prop1: 'donuts' }, { key: 'plugin2' });
+		});
+
+		it('getPlugin should return undefined for invalid plugin set key', () => {
+			const plugin = getPlugin('invalid-plugin-set-key', 'plugin1');
+			expect(plugin).to.equal(undefined);
```
dlockhart, line 77: Nit alternate:

```suggestion
			expect(plugin).to.not.be.defined;
```

dlockhart (approved):

helpers/README.md
dbatiste: Renamed method and updated to return null.

helpers/plugins.js
dbatiste: Resolving since we chose to rename method.

helpers/test/plugins.test.js
dbatiste: Updated to throw error if `registerPlugin` is called after `getPlugins` for the same plugin Set key.

helpers/plugins.js
```diff
@@ -1,11 +1,8 @@
 const pluginSets = new Map();
-
-export function getPlugin(setKey, pluginKey) {
-	const pluginSet = pluginSets.get(setKey);
-	return pluginSet?.plugins.find(plugin => plugin.options.key === pluginKey)?.plugin;
-}
+const pluginsRequested = new Set();
 
 export function getPlugins(setKey) {
+	if (!pluginsRequested.has(setKey)) pluginsRequested.add(setKey);
```
dlockhart, line 5: Another option would be to set something like a `requested = true` on the set, similar to `requiresSorting`.

dbatiste: Yeah, I originally considered that but felt wrong because `getPlugins` would need to add to the `pluginSets` for the case where there are no registrations. But now that I think about it more, tracking the bit separately isn't really saving anything, whether the plugin Set key is valid or not. I think I will switch it.

dlockhart (approved):

PR #4275, opened 2023-11-23
use PropertyRequiredMixin in empty state and text input

dlockhart:
These ended up being the only additional places -- very satisfying to delete some code though!


dbatiste (approved):
🎉

PR #4276, opened 2023-11-23
removing a11y-test-helpers

dlockhart:
The only place that was using this [was just Graveyarded](https://github.com/BrightspaceGraveyard/learning-paths), so cleaning up this old way to run aXe tests.


svanherk (approved):
🎉

PR #4277, opened 2023-11-24
suggest langterm for errorMaxDateOnly and errorMinDateOnly

lideng928:
## Context

- [issue in regards to the tooltip error message](https://github.com/Brightspace/nova/issues/5730) raised from the wave team

## Implementation Details

- Update the `errorMaxDateOnly` message to `Date must be {maxDate} or before`
- Update the `errorMinDateOnly` message to `Date must be {minDate} or later`


margaree:
Thanks for the PR! I agree this is definitely unclear. @geurts do you have any thoughts on these error messages?

geurts:
Nice find, glad we're updating this! Definitely agree it's wrong the way it is.

After collaborating with @nicolebezaire, here's what we'd like to go with:

- `Date must be before or on [date]`
- `Date must be on or after [date]`

This orders the prepositions (on/before/etc) chronologically, and keeps the date at the end, which should make it more easily scannable / parsable.

dlockhart:
Closing in favour of #4351.

PR #4278, opened 2023-11-24
[DE55422] Input text has click event which propagates upwards. Stop p…

jesseag:
…ropagation of this click event so the click handler in rubric feedback no longer receives this event.

Rally ticket: [DE55422](https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F728905531923)

Please refer to the following PR/discussion.
https://github.com/BrightspaceUI/htmleditor/pull/1914
https://github.com/BrightspaceUI/htmleditor/pull/1914#discussion_r1403460071

A click event is bubbling upwards from the `d2l-input-text`. After discussion with @grant-cleary in the PR linked above I am trying to determine whether we should stop propagation of this click event in `d2l-input-text` or if this is intended behaviour to have the click event propagate upwards? If this is intended behaviour then I will go ahead and merge my changes in the above htmleditor PR. However, if this is not the intended behaviour then I would like to stop propagation of this click event directly at the source. Thanks!


jesseag:
@BrightspaceUI/gaudi-dev
Should I be concerned about the test/vdiff failures?

dlockhart:
As you discussed in that thread, I don't think we can stop propagating this event -- there's no telling if existing consumers could be relying on it, making this a breaking change. Native `click` events in general always propagate up through the DOM no matter what element was clicked (so this isn't a `d2l-input-text` specific problem)... I think it would be best to just handle this from the editor code.

jesseag:
> As you discussed in that thread, I don't think we can stop propagating this event -- there's no telling if existing consumers could be relying on it, making this a breaking change. Native `click` events in general always propagate up through the DOM no matter what element was clicked (so this isn't a `d2l-input-text` specific problem)... I think it would be best to just handle this from the editor code.

Okay yeah this tracks for me. I don't want to be making any breaking changes. 
I'll close this PR down then and merge in my fix in the `htmleditor` repo instead. Thanks!

jesseag:
> > As you discussed in that thread, I don't think we can stop propagating this event -- there's no telling if existing consumers could be relying on it, making this a breaking change. Native `click` events in general always propagate up through the DOM no matter what element was clicked (so this isn't a `d2l-input-text` specific problem)... I think it would be best to just handle this from the editor code.
> 
> Sure this resonates with me. I don't want to be making any breaking changes. I'll close this PR down then and merge in my fix in the `htmleditor` repo instead. Thanks!

Sure this resonates with me. I don't want to be making any breaking changes.
I'll close this PR down then and merge in my fix in the htmleditor repo instead. Thanks!

PR #4280, opened 2023-11-24
feat: removed unused dom-test-helpers

dlockhart:
The few remaining uses of these have been cleaned up.


svanherk (approved):
Thank you! 🎉

PR #4281, opened 2023-11-27
[1] Add Button: modes

margaree:
This makes use of `VisibleOnAncestorMixin` so I divided out the "+" icon and the "+ text" in order to make use of that. I have kept this in the same file since another file doesn't really seem necessary since it's small and we don't really want to advertise this as another component, but let me know if I should divide it out.

**Updated Description:**
Adds modes for `icon`, `icon-and-text`, and `icon-when-interacted`.
- `icon`: + icon is always visible. `text` is shown in tooltip on hover/focus.
- `icon-and-text`: + icon and text are always visible. No tooltip.
- `icon-when-interacted`: + icon is only visible when hover/focus. `text` is shown in tooltip also on hover/focus.

Note that animations will be done in a followup PR.

**No Longer Applicable Description:**
How this works:
- When `text-visible` is `true`: button is always visible, these new hint properties have no impact
- When `text-visible` is `false`: new state for `iconOnlyVisibleOnHoverFocus`:
     - "true": the icon is only visible when shown by `VisibleOnAncestorMixin`
     - "false" (default): the icon is always visible
- ~When `text-visible` is `false`: 3 new states for "hint timing" (could also call it `visibility timing` or something else?) which impact the "+" icon but NOT the separator line:~
     - ~"always": the "+" icon is always visible~
     - ~"nearby": the "+" icon is visible when user is hovering within a certain distance (default is 37px, can be defined in the `hint-nearby-distance` attribute) or hover/focus~
     - ~"never": the "+" icon is only visible on direct hover and focus~


components/button/button-add.js
```diff
@@ -1,20 +1,39 @@
...
  * A component for quickly adding items to a specific locaiton.
  */
 class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(LitElement))) {
 	static get properties() {
 		return {
+			/**
+			 * ONLY used when text-visible is FALSE and hint-timing is "nearby". Distance in px where when the user is hovering within that distance the icon will appears. Default is 37.
+			 * @type {number}
+			 */
+			hintNearbyDistance: { type: Number, reflect: true, attribute: 'hint-nearby-distance' },
```
margaree, line 31: I don't love the names of these new properties - let me know other thoughts. Maybe "visibility" instead of "hint"?

components/button/button-add.js
```diff
@@ -24,7 +43,8 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
 			 * When true, show the button with icon and visible text. When false, only show icon.
 			 * @type {boolean}
 			 */
-			textVisible: { type: Boolean, reflect: true, attribute: 'text-visible' }
+			textVisible: { type: Boolean, reflect: true, attribute: 'text-visible' },
+			_hasFocusOrHover: { state: true }
```
margaree, line 47: This was added to pass the hover/focus state down to the +. I think it seems to be working as expected for focus-visible.

components/button/button-add.js
```diff
@@ -59,30 +79,131 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+	}
+
+	static get focusElementSelector() {
+		return 'button';
+	}
+
+	render() {
+		const buttonSpacing = this._getButtonSpacing();
+		const text = this.text || this.localize('components.button-add.addItem');
+		const content = this.textVisible ? this._renderWithTextVisible(text) : this._renderWithTextHidden(text);
```
margaree, line 109: I divided render out into text visible vs hidden since they are fairly different in what they do

components/button/button-add.js
```diff
@@ -59,30 +79,131 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+				<div class="line"></div>
+				${content}
+			</button>
+		`;
+	}
+
+	_getButtonSpacing() {
+		if (this.textVisible || this.hintTiming !== HINT_TIMING.NEARBY) return {};
+
+		this._visibleDistance = this.hintNearbyDistance || DEFAULT_HINT_DISTANCE;
```
margaree, line 127: `this._visibleDistance` is used for calculating the tooltip offset

components/button/button-add.js
dbatiste: Ug, yeah, this seems awkward to name. Maybe `iconHintDistance`, `iconShownDistance`, `hoverDistance`, `iconShownHoverDistance`? 

Do we foresee consumers needing to control this? We might get a bunch of inconsistency as a result... if so, do we think that's a concern?

components/button/button-add.js
```diff
@@ -59,30 +79,131 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+		this.hintTiming = HINT_TIMING.ALWAYS;
+		this.textVisible = false;
+
+		this._buttonId = getUniqueId();
+		this._hasFocusOrHover = false;
+		this._visibleDistance = 0;
+
+		this.addEventListener('focus', this._onFocusHover);
+		this.addEventListener('mouseover', this._onFocusHover);
+		this.addEventListener('blur', this._onBlur);
```
dbatiste, line 98: Nit: if we want to share handlers for focus and mouse events, maybe we should name them... `_onFocusMouseOver` and `_onBlurMouseOut`?

components/button/button-add.js
dbatiste: Does this need to be a variable stored on the object? A couple other ways that might be nicer:

* custom `get/set` for `hintNearbyDistance` reactive property (naming to be determined), or
* calculating the value in `render` and then passing into `_getButtonSpacing` and `_renderWithTextHidden`, or
* moving this to small helper function `_getHintDistance()`, or 
* if it _does_ need to be separate variable stored on the object, then calculating this in [willUpdate](https://lit.dev/docs/v2/components/lifecycle/#willupdate) since it would only really need to be recalculated when  the `hintNearbyDistance` reactive property changes, or

Thoughts?

components/button/button-add.js
```diff
@@ -59,30 +79,131 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+			</button>
+		`;
+	}
+
+	_getButtonSpacing() {
+		if (this.textVisible || this.hintTiming !== HINT_TIMING.NEARBY) return {};
+
+		this._visibleDistance = this.hintNearbyDistance || DEFAULT_HINT_DISTANCE;
+		return {
+			marginBottom: `-${this._visibleDistance}px`,
```
dbatiste, line 129: I think we briefly talked about this when we met with Jeff, but I'll raise it again since I think it might bite us. Overlapping with other elements can be problematic. In this case, the link is entirely non-operable with a mouse.

image

components/button/button-add.js
```diff
@@ -59,30 +79,131 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+			text: { type: String },
+			showFocus: { type: Boolean, reflect: true, attribute: 'show-focus' }
+		};
+	}
+
+	static get styles() {
+		return [visibleOnAncestorStyles, css`
+			:host {
 				align-items: center;
 				background-color: white;
```
dbatiste, line 187: There's no guarantee that the consumer's background color will be white. We might end up needing this to be `transparent` if it looks weird for selected list-items, but I know the reason is to "break" the line. We might end up needing more than one line, for the left & right, unless we can find another clever way to do it.

dbatiste: One option would be to figure out the background color at runtime. I would favour two lines if necessary though, because calculating the background color can be costly and imperfect.

components/button/button-add.js
margaree: Yeah agreed. I ran into this when trying to add it into list - it basically seems (which makes sense) like you can get either the interactive element (e.g., list item) or the link but not both. We likely want to rethink having `hintNearbyDistance` at all and just have the hover distance be whatever `button-add`'s padding is. I'll raise this with Sravan.

components/button/button-add.js
```diff
@@ -1,20 +1,39 @@
...
+			/**
+			 * ONLY used when text-visible is FALSE and hint-timing is "nearby". Distance in px where when the user is hovering within that distance the icon will appears. Default is 37.
+			 * @type {number}
+			 */
+			hintNearbyDistance: { type: Number, reflect: true, attribute: 'hint-nearby-distance' },
+			/**
+			 * ONLY used when text-visible is FALSE. When to show the icon/icon text with user interaction. "always" - always visible, "nearby" - when mouse within hint-nearby-distance, "never" - do not show until hover/focus. Default is "always".
+			 * @type {'always'|'nearby'|'never'}
+			 */
+			hintTiming: { type: String, reflect: true, attribute: 'hint-timing' },
```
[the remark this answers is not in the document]
margaree: `buttonInnerContentVisibilityState`? ``buttonInnerContentVisibilityCondition`? I feel like that's what it is but it's way too wordy.

components/button/button-add.js
```diff
@@ -56,49 +74,28 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
 				width: 100%;
 			}
 			button:hover .line,
-			button:${unsafeCSS(getFocusPseudoClass())} .line {
+			button:focus .line {
```
margaree, line 77: I switched this to use `focus` rather than `focus-visible` because VisibleOnAncestorMixin uses `focus` and having some parts use `focus` and others use `focus-visible` resulted in some strange focus behaviour.

components/button/button-add.js
```diff
@@ -56,49 +74,28 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+				--d2l-button-add-icon-text-icon-color: var(--d2l-color-celestine);
 			}
-
-			span {
-				font-size: 0.7rem;
-				font-weight: 700;
-				letter-spacing: 0.2px;
-				line-height: 1rem;
+			:host([visibility-condition="nearby"]) button:not(:focus):not(:hover) d2l-button-add-icon-text,
+			:host([visibility-condition="hover-focus"]) button:not(:focus):not(:hover) d2l-button-add-icon-text {
```
margaree, line 86: In order to not have the background color issue, there are now 2 lines. When `button-add` is not hovered on or focused on and the visibility condition is nearby or hover (i.e., inner content should not be visible), positioning the inner content absolute allows the 2 lines to combine

components/button/test/button-add.vdiff.js
```diff
@@ -2,30 +2,56 @@ import '../button-add.js';
...
+								if ((name === 'hover' || name === 'focus') && !elem.textVisible) await oneEvent(elem, 'd2l-tooltip-show');
+								await expect(elem).to.be.golden();
+							});
 						});
 					});
 				});
 			});
 		});
 	});
+	describe('text-visible true', () => {
```
margaree, line 33: I separated these visual diffs out since they aren't affected by `visibility-condition`

components/button/button-add.js
```diff
@@ -15,6 +15,10 @@ import { PropertyRequiredMixin } from '../../mixins/property-required/property-r
 class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(LitElement))) {
 	static get properties() {
 		return {
+			/**
+			 * ONLY used when text-visible is FALSE. When true, icon is only visible when component receives hover or focus. When false (default), icon is always visible.
+			 */
+			iconOnlyVisibleOnHoverFocus: { type: Boolean, reflect: true, attribute: 'icon-only-visible-on-hover-focus' },
```
margaree, line 21: I've removed the 2 new variables for the always/never/hover-focus states and replaced with a bool since there are now only the 2 possible states for icon.

components/button/button-add.js
```diff
@@ -15,6 +15,10 @@ import { PropertyRequiredMixin } from '../../mixins/property-required/property-r
 class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(LitElement))) {
```
dbatiste, line 15: Should this be using the `PropertyRequiredMixin`? I notice the error messages showing in the console on the demo-page, but the component appears to always have text.

components/button/button-add.js
dbatiste: Neat!

dbatiste:
I'm not sure if this is an issue or not... when I move my mouse across the page when `icon-only-visible-on-hover-focus` is set, the icon and line flickers. It's perhaps most noticeable when moving the pointer across the line. It can be a little distracting, but maybe I'm trying too hard to make it happen. 🤷‍♂️  

It's basically the result of showing and then hiding quickly, due to the interactive height being so short. This isn't a noticeable issue with things like `d2l-card` because the `d2l-visible-on-ancestor-target` area is suitably large in both the x & y directions. 

Finding a way to only show after some brief period of hovering might improve the user experience. At first glance, the `VisibleOnAncestorMixin` (which is giving us this behaviour) doesn't provide any delay options. 

Maybe get Jeff and/or Sravan to play with it and see what they think.

components/button/button-add.js
margaree: I'm torn on it - using `PropertyRequiredMixin` hopefully forces someone to consider the value while having a valid fallback in case it's missed, but it's not totally necessary since it does have a fallback.

margaree:
> I'm not sure if this is an issue or not... when I move my mouse across the page when `icon-only-visible-on-hover-focus` is set, the icon and line flickers. It's perhaps most noticeable when moving the pointer across the line. It can be a little distracting, but maybe I'm trying too hard to make it happen. 🤷‍♂️
> 
> It's basically the result of showing and then hiding quickly, due to the interactive height being so short. This isn't a noticeable issue with things like `d2l-card` because the `d2l-visible-on-ancestor-target` area is suitably large in both the x & y directions.
> 
> Finding a way to only show after some brief period of hovering might improve the user experience. At first glance, the `VisibleOnAncestorMixin` (which is giving us this behaviour) doesn't provide any delay options.
> 
> Maybe get Jeff and/or Sravan to play with it and see what they think.

I've noticed that sometimes as well. I'll dig at it a bit and see if I can improve it.

components/button/button-add.js
dbatiste: If they are legit allowed to just use the fallback text, then I think we should avoid the error. Otherwise we'll be adding errors to the error db even though they aren't real errors. On the other hand, if they really are required to provide their own text, then the error is fair game.

margaree: Make sense. I'll clarify with design but my current understanding is that the fallback text is allowed.

components/button/button-add.js
```diff
@@ -109,19 +93,103 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+		this._hasFocus = true;
+	}
+
+	_onMouseEnter() {
+		this._hasHover = true;
+	}
+
+	_onMouseLeave() {
+		this._hasHover = false;
+	}
```
margaree, line 132: Separated these out to cover the case of focus + hover then remove focus should still show the hover/focus icon

components/button/button-add.js
```diff
@@ -109,19 +93,103 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+				font-size: 0.7rem;
+				font-weight: 700;
+				letter-spacing: 0.2px;
+				line-height: 1rem;
+			}`
+		];
+	}
+
+	render() {
+		const icon = this.hoverFocusIcon ? 'tier1:add-filled' : 'tier1:add';
```
margaree, line 188: The other option I played with was having the icon completely set by the main component, but this way seemed cleaner.

components/icons/catalogue.md
```diff
@@ -38,9 +38,10 @@ Size: `18px` x `18`px
...
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/add-file.svg?sanitize=true) | add-file | | ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/event-log.svg?sanitize=true) | event-log | | ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/plus-large.svg?sanitize=true) | plus-large |
+| ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/add-filled.svg?sanitize=true) | add-filled | | ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/exemption-add.svg?sanitize=true) | exemption-add | | ![](https://raw.githubusercontent.com/BrightspaceUI/core/main/components/icons/images/tier1/preview.svg?sanitize=true) | preview |
```
margaree, line 44: This adds the `add-filled` tier1 icon and modifies the `add` icon.
There are also `tier2` versions but they needed tweaks so I'll be updating that later.

margaree (commented):
This is ready for re-review.

There are still some to dos to get to matching the design spec but I'm thinking I'll do those separate so as not to add more to the design. The major changes are in here.

In response to Dave B's comment here https://github.com/BrightspaceUI/core/pull/4281#issuecomment-1834711661, we'll be adding a delay to avoid that flicker but I was thinking of doing that all separately from this.

components/icons/catalogue.md
margaree: Update on this: we don't yet want to replace the existing `add` icon so I'm going to make this a new file.

components/button/button-add.js
dlockhart: I think I'd lean towards no fallback and making it required. That's assuming that we don't expect consumers to ever want / be satisfied with the default "add item" text.

components/button/button-add.js
dlockhart: Hmm... would a single tri-state attribute make more sense? Something like `mode` which has the possible values of: `icon` (default), `icon-and-text` (equivalent to the current `text-visible`) or `icon-when-interacted`? You could still reflect it to target the different states with CSS attribute selectors.

It's always a bit of a code smell to me when we have attributes that only apply under certain circumstances and if ignored can cause the thing to be in a weird state.

components/button/button-add.js
```diff
@@ -110,18 +92,91 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+					<path fill-rule="evenodd" clip-rule="evenodd" d="M10 5C10 4.44772 9.55228 4 9 4C8.44772 4 8 4.44772 8 5V8H5C4.44772 8 4 8.44772 4 9C4 9.55228 4.44772 10 5 10H8V13C8 13.5523 8.44772 14 9 14C9.55228 14 10 13.5523 10 13V10H13C13.5523 10 14 9.55228 14 9C14 8.44772 13.5523 8 13 8H10V5Z" />
+				</g>
+			</svg>
+			${this.text ? html`<span class="d2l-label-text">${this.text}</span>` : nothing}
+		`;
+	}
+}
+customElements.define('d2l-button-add-icon-text', ButtonAddIconText);
```
dlockhart, line 182: Do we intend to use this elsewhere? If not, just wondering if we should be exposing it publicly like this vs. just baking it into `d2l-button-add`.

components/button/button-add.js
margaree: Good call, I like that better. It seems more clear.

components/button/button-add.js
margaree: We don't intend to use this elsewhere. I separate this piece out in order to get `visible-on-ancestor-mixin` to work. Is there a way to use `visible-on-ancestor-mixin` within the same component? There are some animation tweaks I need to make to visible-on-ancestor in a follow up PR so I'm not opposed to pulling the functionality into `button-add` and modifying as needed to get it to work for a single component case if needed/possible.

dbatiste: Currently, it looks like it starts its search (upwards) beginning with the host of the element extending the mixin ([here](https://github.com/BrightspaceUI/core/blob/main/mixins/visible-on-ancestor/visible-on-ancestor-mixin.js#L113)). Maybe it doesn't search within it's own DOM context because in theory, their could be a nested VOA element and target. That case might be pretty edgy. Maybe we could make VOA a little more flexible.

dbatiste (approved):
Looks great!

PR #4285, opened 2023-12-04
[3] Add Button: List: Initial work

margaree:
[Jira Ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?assignee=557058%3A87a56381-8a81-483e-91b9-fabf668aa40c&selectedIssue=GAUD-282)

Initial work to use the "button-add" component within list. Adds ~3~2 new properties related to this to `list` and then adds the component after each `list-item` when applicable.

~I've added `add-button` property to the list for every vdiff case just for now in order to easily be able to see what's changed while this is in review. I'll be removing that and adding only applicable cases once this looks acceptable.~ I added vdiff tests to test applicable cases. Let me know if there are any I missed. To see how lists were affected when `add-button` was on every list, see [this PR](https://github.com/BrightspaceUI/core/pull/4286).

Note that this does not currently handle the nested case specifically, though it seems to generally be working well and I'll be working on it further in a follow up PR.

Ignore the crazy number of commits - many of them are from an earlier branch that this branch was originally off of that has since merged but I haven't been able to get GitHub to not display all the other branch's commits..


components/list/list-item-drag-drop-mixin.js
```diff
@@ -632,6 +634,7 @@ export const ListItemDragDropMixin = superclass => class extends superclass {
 		} else {
 			if (this.shadowRoot) {
 				const nodeImage = this.shadowRoot.querySelector('.d2l-list-item-drag-image') || this;
+				nodeImage.classList.add('dragging');
```
margaree, line 637: I added a `dragging` class and used it in `list-item-mixin` in order to set `display: none` on the add button slot when dragging. This is because the tooltip in add button is causing additional height on the drag preview. This seems to be a long existing issue in Chrome and Safari, where non-visible children can affect the drag preview heights.

I think that even if we do have the add button always visible, it's still okay to hide it on drag and drop, since we don't show the outer borders anyways in the drag preview, and drag and drop is about moving the item itself.

I did try to use the `dragging` property but because it's set to true in `setTimeout` it doesn't trigger the css rule.

dbatiste: Seems reasonable. Do we always get the extra height, or is it only when the tooltip is shown?

components/list/list-item-generic-layout.js
```diff
@@ -69,7 +69,10 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 					[control-end content-start] minmax(0, auto)
 					[content-end actions-start] minmax(0, min-content)
 					[end actions-end];
-				grid-template-rows: [main-start] [main-end nested-start] [nested-end];
+				grid-template-rows:
+					[main] minmax(0, min-content)
+					[add-start add-end] minmax(0, min-content)
+					[nested-start nested-end] minmax(0, min-content);
```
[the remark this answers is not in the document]
dbatiste: So we're adding a new grid-line, named by both `add-start` and `add-end`. It's a little peculiar that a gridline is named as both the start and end line, with nothing between them. I'm not 100% certain the new grid-lines are doing what we intend, but maybe we don't notice because the content is so short.

components/list/list-item-generic-layout.js
```diff
@@ -171,6 +174,10 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 				grid-column: content-start / end;
 				grid-row: nested-start / nested-end;
 			}
+			::slotted([slot="add"]) {
+				grid-column: color-start / end;
+				grid-row: add-start / add-end;
```
dbatiste, line 179: It's not necessary to change this, but when the names are `add-start` and `add-end`, the "start" and "end" will be assumed when writing `grid-row: add;`. Just a little short-hand tip! 

I started using this short-hand [here](https://github.com/BrightspaceUI/lms-core/blob/main/src/components/attachments/attachment.js#L54) if you're curious to see it in action.

components/list/list-item-generic-layout.js
```diff
@@ -218,6 +225,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			<slot name="actions" class="d2l-cell" data-cell-num="7"></slot>
 
 			<slot name="drop-target"></slot>
+			<slot name="add" class="d2l-cell" data-cell-num="9"></slot>
```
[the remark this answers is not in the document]
dbatiste: Seems reasonable. It's pretty specific, which is perhaps what we want. Food for thought though, do you think this slot could be more generic, and if so, would we want it to be or not?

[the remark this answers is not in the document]
margaree: It definitely could be - perhaps there is some other case where we'd want something in the padding of list items, so could be something like `padding-content` or `border-content`; however I don't have any other use case ideas.

components/list/list-item-mixin.js
```diff
@@ -366,6 +378,18 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				margin-left: 0;
 				margin-right: -6px;
 			}
+
+			[slot="add"] {
+				margin-bottom: -4px;
+				margin-top: -3px;
+				z-index: 10;
```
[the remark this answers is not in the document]
margaree: When I was initially adding `button-add` in here I believe it was showing up under the list items or hover wasn't working as expected, however this seems to no longer be an issue so I'll remove this🎉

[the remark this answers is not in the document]
margaree: Ah nevermind - so the case is if approaching using hovering from below the `button-add`, it wont show up until you reach the list item it's a part of without this `z-index` (we'd want it to show up when we hover slightly below or at the + icon or slightly below the line)

components/list/list-item-mixin.js
```diff
@@ -523,6 +544,11 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		}
 	}
 
+	_handleButtonAddClick() {
+		/** Dispatched when the add button directly after the item is clicked */
+		this.dispatchEvent(new CustomEvent('d2l-list-item-add-button-click', { bubbles: true, composed: true }));
```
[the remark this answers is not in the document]
dbatiste: Hmmm, I'm sort of undecided on this at the moment, but these are my initial thoughts...

The consumer is going to need to know which add-button was clicked. If the event is dispatched from the item, then they will know the reference item via `e.target`. If we dispatch the event from the list, how will they know? - I think we'd need to provide them with event detail such as the item `key`, which is currently at least defined and used by `ListItemCheckboxMixin` and `ListItemDragDropMixin`.

From an ergonomics perspective, if the event is defined to bubble from the item (which it currently is), technically consumers could wire-up a single handler to the list.

But should it bubble? We should consider nested lists where a use clicks an add-button in a nested list. Currently, it looks like the event will bubble up and event handlers on ancestor lists would be invoked. Is that what we'd want? It seems like that might make the consumer's ancestor handler more error prone since it would have to consider that the event might not be coming from its own items.

If the event handler was wired up to the item, and that handler is defined in their component with references to their list state, ids, etc., but then the item is moved to a different list. The original handler might be holding reference to stale state. However, I think our drag and drop creates brand new items when items are moved, and the original DOM not is removed, so maybe that would not be a problem in practice. Nevertheless, something to think about, and kinda leads me more to thinking the event belongs more to the list itself, rather than the items.

The event is currently set to `composed: true`. Do we need it to traverse DOM scopes?

components/list/list-item-mixin.js
```diff
@@ -595,6 +624,11 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		const alignNested = ((this.draggable && this.selectable) || (this.expandable && this.selectable && this.color)) ? 'control' : undefined;
 		const primaryAction = ((!this.noPrimaryAction && this._renderPrimaryAction) ? this._renderPrimaryAction(this._contentId) : null);
 		const tooltipForId = (primaryAction ? this._primaryActionId : (this.selectable ? this._checkboxId : null));
+
+		const rootList = this.getRootList();
```
dbatiste, line 628: A couple of different thoughts here...

Is it possible to have a scenario where a nested list has the add button, but the ancestor list does not? Just wondering whether we should be checking the list that the item belongs to vs the root list.

Ideally if the list's `add-button` property changes, the list-items update themselves to show/hide the add-buttons as needed. It kinda seems like the items would add/remove the add-button as a result of the item's `render` method running for other reasons. Is that correct? Do we need something more to handle this case?

Kinda related to the above, I'm curious how often this runs, and whether we can avoid calling `getRootList` every time the item's `render` is called.  I'm not certain but I suspect that the render method may get called often, as the user interacts (hovers, focuses, selects, expands, drags, etc).

components/list/list-item-generic-layout.js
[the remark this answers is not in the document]
dbatiste: Cool. I'm fine with it being specific. Just figured it'd be worth mulling over. 😄

components/list/list-item-mixin.js
margaree: Great points
1 - I'll confirm with Jeff but my understanding is that we want this to be set at the root list level and would not want consumers just enabling `add-button` for certain nested lists
2 - You are correct on how it should behave and how it currently is. I'll look into if there's a better way to handle this so if `add-button` changes then list items update
3 - I also think it runs fairly often. I'll see if there's a better way to avoid calling `getRootList` so often.

margaree: Checked with Jeff - we actually want this to just be a property on each list (similar to `grid`) so nested lists would have to opt-in as well. I think that drastically simplifies all of this.

margaree:
I'm pushing my latest changes in order to see if there's anything in the PR I need to cleanup, but haven't yet address the event so this can be ignored for now.

components/list/list.js
```diff
@@ -207,6 +218,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 		if (changedProperties.has('breakpoints') && changedProperties.get('breakpoints') !== undefined) {
 			this.resizedCallback(this.offsetWidth, true);
 		}
+		if (changedProperties.has('add-button') || changedProperties.has('add-button-text')) {
+			this._listChildrenUpdatedSubscribers.updateSubscribers();
```
margaree, line 222: I tabbed into the updating subscribers code in order to update the list children when `add-button` or `add-button-text` get changed. Not sure if this is totally ideal, as it also runs the other update functions. If we don't think this is going to change post-initial-render, it could also be done like `grid` and `extend-separators` where those get set for the child list items in `list-item-role-mixin`

components/button/button-add.js
```diff
@@ -54,13 +58,18 @@ class ButtonAdd extends RtlMixin(PropertyRequiredMixin(FocusMixin(LocalizeCoreEl
 				justify-content: center;
 				outline: none;
 				padding: 0;
+				position: relative;
 				user-select: none;
 				white-space: nowrap;
 				width: 100%;
+				z-index: 10;
```
margaree, line 65: The only way I found to get the `button-add` to be able to be interacted with from below (i.e., from a sibling list item) and also for the tooltip to be above neighbouring `button-add`s was to do this. I think this is relatively safe, as `button-add` would generally usually be above nearby content, but it could cause problems

components/list/list-controls.js
```diff
@@ -52,6 +52,10 @@ export class ListControls extends SelectionControls {
 		if (this._extendSeparator) this.style.setProperty('--d2l-selection-controls-padding', '0px');
 	}
 
+	updateParentHasAddButon() {
+		// N/A
```
margaree, line 56: Since `list-controls` is a subscriber, it's affected by the `list` change below to use the subscriber code

components/list/list-item-mixin.js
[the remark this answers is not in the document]
margaree: Great points.

I would say I agree the in the nested list case we actually would not want it to bubble. It does sound safest to add the listener to the list itself and use the keys for identifying the list item whose button was clicked. Then everything `add-button` would be added in the same place as well. I don't think we need the event to traverse DOM scopes.

margaree:
This should be ready for another looking when you have time. There are a couple of new outstanding questions on implementation. I added comments in the [vdiff PR ](https://github.com/BrightspaceUI/core/pull/4286) on the interesting changes and messaged Jeff about them as well.

components/list/list.js
dbatiste: I think the way `ListItemRoleMixin` works, it only checks for `grid` and `extend-separators` in the `connectedCallback`. I think there is an assumption (which is probably fairly safe), that those won't change once the item is connected to a list.

For this case though, might we ever expect the consumer to change `addButton` or `addButtonText` after the items have rendered, perhaps in response to some user action? I'm honestly not sure, but your current solution would appear to handle that case.

components/list/list-item-mixin.js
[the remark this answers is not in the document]
dbatiste: Ok. Sounds good. I notice the code still dispatches `d2l-list-item-add-button-click` from the list-item. Are you planning to change it so that the event is dispatched from `d2l-list`, and configured to not bubble?

components/button/button-add.js
dbatiste: This is the sticky header case with `z-index` > 5:
image

Without the `z-index`, it's possible for tooltip to float over the sticky header. That might have been an existing issue we were just "living with", as it probably would not come up often.
image

Is there a reason `10` was chosen? It seems like `1` might be sufficient to stack it on top of the adjacent next sibling, while allowing it to float under the sticky header. But it looks like drag & drop uses `z-index` a bit, so maybe `1` _isn't_ sufficient to avoid issues.

dbatiste: This is kinda icky...

image

But like the 2nd screenshot above, it's not due to your `z-index` usage, but rather, I think it's probably just due to `d2l-tooltip`. I think it was possible to make this happen without your changes, it just would have been less likely because it would require users to put tooltips (or components using them) into their items, and many list-items do not contain interactive content.

margaree: Ah gross - I'll play around with smaller z-indexes and see if I can get something else working. Agreed with your analysis about tooltip - it has a large z-index on it that's causing this to happen and this just makes it a bit more likely (though hopefully this is still rare in practice)

margaree: z-index of 1 does seem to give the functionality we want and does not have that sticky header issue. I didn't see any drag and drop issue (we do hide the add button during dragging so it hopefully wouldn't have an effect) but let me know if there are any you know of.

components/list/list-item-mixin.js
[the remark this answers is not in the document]
margaree: So I set it up so that the event chain goes: `click` event (`d2l-button-add`) > `d2l-list-item-add-button-click` event (`list-item-mixin`) > `d2l-list-add-button-click` event (`list`). I noticed that if I didn't have `bubbles` on the list item event then the list event listener for `d2l-list-item-add-button-click` didn't actually get triggered. Is there something wrong with my setup for this?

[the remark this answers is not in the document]
dbatiste: Just that in the nested case, higher level lists would also get the event. If we don't want those higher level lists to see the events coming from nested lists, we'd either have to:

a) dispatch the `d2l-list-add-button-click` event directly from the `d2l-list` element and use `bubbles: false`, or
b) `d2l-list` would need to implement an event handler for it that calls `e.stopPropagation();` so that the event doesn't bubble any further than the immediate list parent. In this case, hopefully `stopPropagation` would not prevent _all_ the handlers (ours as well as the consumers) on `d2l-list` from running.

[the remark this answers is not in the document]
margaree: Ah gotcha - I did add `e.stopPropogation` to the `list` handler for the list item event but it sounds like that might be riskier so I can switch to method (a) if that's preferred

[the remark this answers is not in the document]
dbatiste: Ah sorry. I didn't notice that. I think either way if fine. If we run into any problem with the consumer's handler not running, we can switch.

components/list/list.js
margaree: It's totally possible it could change in response to a user action, so it sounds like this way might be the better option.

components/list/list-item-mixin.js
```diff
@@ -384,7 +384,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				margin-bottom: -4px;
 				margin-top: -3px;
 			}
-			:host([draggable][selectable]) [slot="add"] {
+			:host([draggable][selectable][_hovering]) [slot="add"],
+			:host([draggable][selectable][_focusing]) [slot="add"] {
 				padding-inline-end: 6px;
 			}
```
margaree: My concern was that if hovering/focusing on a sibling then the line extended a bit too far right. However, this is actually consistent with current behaviour.

components/list/list-item-mixin.js
```diff
@@ -478,7 +479,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 	}
 
 	updateParentHasAddButon(addButton, addButtonText) {
-		this._addButtonText = addButtonText || this.localize('components.list-item.addItem');
+		this._addButtonText = addButtonText;
```
margaree: It turned out that `this.localize` running here was returning an empty string, so I moved it to within the rendering

components/button/button-add.js
```diff
@@ -54,13 +58,18 @@ class ButtonAdd extends RtlMixin(PropertyRequiredMixin(FocusMixin(LocalizeCoreEl
 				justify-content: center;
 				outline: none;
 				padding: 0;
+				position: relative;
 				user-select: none;
 				white-space: nowrap;
 				width: 100%;
+				z-index: 1;
```
dbatiste, line 65: Nit: maybe we can add a comment here indicating what this `z-index` is for.

components/list/list-item-mixin.js
```diff
@@ -74,6 +75,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			 * @type {'normal'|'none'}
 			 */
 			paddingType: { type: String, attribute: 'padding-type' },
+			_addButtonText: { type: String },
```
dbatiste, line 78: I think this could be `{ state: true }`, ya?

dbatiste (approved):
Looks great!

PR #4287, opened 2023-12-05
[ignore for now] Filter: Overflow group: Fix change event not firing

margaree:
[Rally Story](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F732021201387)

The issue is that when chomped the filters are [kept up to date through the dimension sets](https://github.com/BrightspaceUI/core/blob/main/components/filter/filter-overflow-group.js#L98), which does not trigger a `d2l-filter-change` event on the original filter. I've added dispatching the change event to the `_handleDimensionDataChange` function which is the handler for `d2l-filter-dimension-data-change` events. This is likely the simplest way to fix this, but I don't know if there are side affects that I'm ignoring with this or if there is a reason this was not done initially (other than perhaps it just wasn't necessary).

This does not handle having `cleared` or `allCleared` be accurate in the results, but the event is triggered and the other data within it seems accurate.

`this._activeFiltersSubscribers.updateSubscribers();` happens in `_dispatchChangeEvent` so I removed it.

### DIFF
--- a/components/filter/filter.js
+++ b/components/filter/filter.js
@@ -701,7 +701,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 					dimension.appliedCount--;
 					this._totalAppliedCount--;
 				}
-				this._activeFiltersSubscribers.updateSubscribers();
+				this._dispatchChangeEvent(dimension, { valueKey: e.detail.valueKey, selected: newValue });
 			} else if (prop === 'values') {
 				if (dimension.searchValue || dimension.searchType === 'manual') shouldSearch = true;
 				shouldRecount = true;


svanherk:
> which does not trigger a d2l-filter-change event on the original filter. I've added dispatching the change event to the _handleDimensionDataChange function which is the handler for d2l-filter-dimension-data-change events. This is likely the simplest way to fix this, but I don't know if there are side affects that I'm ignoring with this or if there is a reason this was not done initially (other than perhaps it just wasn't necessary).

So from what I recall, I think this was done because it causes a loop in the HM case. A user changes a value, which triggers an event. A HM-driven filter will then update it's copy of the data, setting the selected values accordingly.  If those _also_ trigger an event, it causes at minimum an unnecessary second pass-through on the HM side, or at worst it keeps looping.

One thing that might be similar but safer is to expose a new function in `filter` like `requestFilterValueChange` that mimics firing the event instead, and then the consumer will catch it and set it and trigger the regular update flow. Basically an alternative to the "just in JS" method if we _know_ we want an event?

margaree:
Ah darn, that was my fear.

So you're thinking something like `filter-overflow-group` would call filter's (new) `requestFilterValueChange` when the dimension set values change in `filter-overflow-group` and that would fire `d2l-filter-change`?

svanherk:
Yeah, that was my thinking, and then the consumer must keep the selected value up-to-date based on the events received, and that will set the hidden version of the filter so the hidden and non-hidden match. I think it will look basically like `requestFilterValueClear`, but without the hardcoded `false`, so maybe worth generalizing that?

For clear button presses, we have `requestFilterClearAll` but not `requestFilterDimensionClear` - maybe it's not worth worrying about, and if your filter is clamped you get individual messages and can't utilize the performance gain of the `cleared: true` event?

And then not sure what to do with `first-open`... 🤔

svanherk:
Or `load-more`, or `search`, or `empty-state` events... 😟 

One POC that might be worth it after this one is creating a demo that puts all the events on the `d2l-filter-overflow-group`, and then reacts accordingly based on the `target`. Messier for the consumer, but otherwise we need to map each event over.  Maybe we can do that generally though? We catch the event in the overflow filter, find the corresponding original filter, and re-raise the same event?

margaree:
So sadly putting the events on `d2l-filter-overflow-group` doesn't work out-of-the-boxy for the overflowed filter dimensions.

I did get the process of getting the overflowed dimension set's parent filter then dispatching the change event from that original filter working. That seems like it might be the path we have to go down. I just didn't like it because it felt like it was giving filter overflow group too much knowledge about the inner workings of filter, and then if things change in filter we have to remember to change them in filter overflow group too. I might be able to somehow put some of that code back in filter and then call functions in filter overflow group - I'll see what I can do with that.

PR #4292, opened 2023-12-12
fix: US160944: Date and Time Ranges: Fix bug where wrapping does not work correctly when component is direct child of custom element

margaree:
[Rally](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fuserstory%2F732652754923)

This was triggering a console error due to the fact that if the date/time range component was the first child in a custom element, `this.parentNode` would return `ShadowRoot` which would then throw in the resize observer.

This only happens when the date/time range component is the first child and would likely only be an issue in the UI if `child-labels-hidden` is true.

I'm seeing if I can get unit tests working to just confirm that the error throwing is no longer happening. The visual diff tests do capture this though, as the tests fail if the error happens.

### DIFF
--- a/components/inputs/input-date-range.js
+++ b/components/inputs/input-date-range.js
@@ -163,8 +163,7 @@ class InputDateRange extends FocusMixin(SkeletonMixin(FormElementMixin(RtlMixin(
 		if (!this.label) {
 			console.warn('d2l-input-date-range component requires label text');
 		}
-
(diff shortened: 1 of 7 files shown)


dlockhart (approved):
Good catch. Yeah getting the non-vdiff tests to "catch" this error is going to be really tricky / maybe impossible, seeing as it gets thrown asynchronously. I haven't had success with this kind of thing in the past, hence why I took a different approach for the `PropertyRequiredMixin` errors.

PR #4295, opened 2023-12-14
DE55986: Nested list: Bug with nested list disabled selection

margaree:
[Rally](https://rally1.rallydev.com/#/15545167705ud/custom/21568985922?detail=%2Fdefect%2F731764811011)

This is improved but the states are still a bit weird, namely that previously when something was in `indeterminate` state its `selected` state would be `false` but now it is `true` if it's the direct parent and I believe it's usually `false` if it's a later grant+parent. The grand+parents will also appear as `selected` rather than `indeterminate` when all children other than `selection-disabled` ones are checked, so that's a bit wonky. But the ability is uncheck now works! So basically it's kind of better and also kind of confusing. Looking for input! I updated the demo page with a couple scenarios. I also haven't added tests for this since it's likely very up-in-the-air.

### DIFF
--- a/components/list/demo/list-nested.html
+++ b/components/list/demo/list-nested.html
@@ -29,94 +29,6 @@
 
 		<d2l-demo-page page-title="d2l-list (nested)">
 
-			<h2>Nested</h2>
-
-			<d2l-demo-snippet>
-				<template>
-					<d2l-list grid>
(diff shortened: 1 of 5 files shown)


margaree:
Closing for now. This PR is referenced in the Rally DE.

PR #4297, opened 2023-12-14
fix: lint errors

dlockhart:
Ironically (coincidentally?) we've fallen behind on the `eslint-config-brightspace`, so I wanted to upgrade to the latest `0.x` release and fix these issues before testing out the `1.0` changes with core.

Most of these are the "no two spaces" lint issue.

### DIFF
--- a/components/alert/alert.js
+++ b/components/alert/alert.js
@@ -172,7 +172,7 @@ class Alert extends LocalizeCoreElement(RtlMixin(LitElement)) {
 	}
 
 	render() {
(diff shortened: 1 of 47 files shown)


dbatiste (approved):

dlockhart:
Looks like we've got the "byte changed" vdiff issue for almost every image. :(

dlockhart:
> Looks like we've got the "byte changed" vdiff issue for almost every image. :(

Ended up being because I was too zealous with the `package-lock` update.

PR #4302, opened 2023-12-18
Fix lint errors for upcoming changes

dlockhart:
This updates core to adhere to the changes in an upcoming `1.0.0` release of `eslint-config-brightspace`. There were two new rules that this specifically addresses:

[prefer-nothing](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/prefer-nothing.md): pretty straightforward -- use `nothing` whenever possible.

[no-native-attributes](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/no-native-attributes.md): ideally, we'd never be duplicating native attribute names on custom elements. In practice, we've done this in a few "legitimate" cases (like `role`), a few mistakes (like `draggable`) and a few places we thought were OK but in reality we shouldn't have (`autofocus`, `title`).

For the most part, I've added ignore comments. For number/percent inputs though, I was able to remove the `title` attribute and clean up the [few places](https://github.com/Brightspace/nova/pull/6218) that [were using it](https://github.com/BrightspaceHypermediaComponents/activities/pull/4373).

### DIFF
--- a/components/inputs/input-text.js
+++ b/components/inputs/input-text.js
@@ -45,9 +45,9 @@ class InputText extends PropertyRequiredMixin(FocusMixin(LabelledMixin(FormEleme
 			 */
 			autocomplete: { type: String },
 			/**
-			 * When set, will automatically place focus on the input
-			 * @type {boolean}
+			 * @ignore
 			 */
+			// eslint-disable-next-line lit/no-native-attributes
 			autofocus: { type: Boolean },
 			/**
 			 * Additional information communicated in the aria-describedby on the input
@@ -135,9 +135,9 @@ class InputText extends PropertyRequiredMixin(FocusMixin(LabelledMixin(FormEleme
 			 */
 			step: { type: String },
 			/**
-			 * Text for additional screenreader and mouseover context
-			 * @type {string}
+			 * @ignore
 			 */
+			// eslint-disable-next-line lit/no-native-attributes
 			title: { type: String },
 			/**
 			 * The type of the text input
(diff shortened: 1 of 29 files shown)


components/inputs/input-text.js
```diff
@@ -135,9 +135,9 @@ class InputText extends PropertyRequiredMixin(FocusMixin(LabelledMixin(FormEleme
 			 */
 			step: { type: String },
 			/**
-			 * Text for additional screenreader and mouseover context
-			 * @type {string}
+			 * @ignore
 			 */
+			// eslint-disable-next-line lit/no-native-attributes
```
dlockhart, line 140: I would have liked to remove this like number/percent but there were just too many search results to comb through. For now it'll be hidden from the docs and maybe we do a more concerted search effort later.

components/menu/menu-item-mixin.js
```diff
@@ -18,6 +18,7 @@ export const MenuItemMixin = superclass => class extends superclass {
 			/**
 			 * @ignore
 			 */
+			// eslint-disable-next-line lit/no-native-attributes
 			hidden: { type: Boolean, reflect: true },
```
dlockhart, line 22: Ignoring this one for now because there's logic wired up in `updated`. That should probably switch to a `MutationObserver`.

dbatiste (approved):
Would almost be nice if we could configure eslint to not throw errors for these some of these (ex. `role`).

dlockhart:
> Would almost be nice if we could configure eslint to not throw errors for these some of these (ex. role).

Yeah, although I kind of like them being warnings at the same time because we know we're doing something wrong here even if there's no good alternative until the AOM exist.

PR #4309, opened 2024-01-03
[2] Add Button: Animations

margaree:
This PR is on top of https://github.com/BrightspaceUI/core/pull/4281

Adds animations according to the spec. I'll add comments throughout the PR.


components/button/button-add.js
```diff
@@ -53,13 +54,26 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
 			}
 
 			.line {
-				border-top: 1px solid var(--d2l-color-mica);
+				background: var(--d2l-color-mica);
```
margaree, line 57: This was switched from border to background in order to do a linear gradient animation (below)

components/button/button-add.js
```diff
@@ -69,6 +83,34 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+			}
+
+			@media (prefers-reduced-motion: reduce) {
+				button:hover .line,
+				button:focus .line {
+					animation: none !important;
+					background: var(--d2l-color-celestine-minus-1);
+				}
+				:host([mode="icon-when-interacted"]) button:hover .line {
+					transition: background 100ms step-end;
```
margaree, line 94: When `prefers-reduced-motion` is enabled, we don't want the line animation but we do want a bit of a delay in changing the line colour to avoid flicker when user is moving their mouse around a page.

components/button/button-add.js
```diff
@@ -69,6 +83,34 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
...
+				}
+			}
+
+			@keyframes line-start-animation {
+				0% { background: linear-gradient(to right, var(--d2l-color-mica) 0%, var(--d2l-color-mica) 11%, var(--d2l-color-celestine-minus-1) 11%) left center / 135%; }
```
margaree, line 99: This is supposed to be a gradient animation that starts at around 85% length of the line outwards on each side and then fills to 100%.

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
```diff
@@ -34,6 +45,10 @@ export const VisibleOnAncestorMixin = superclass => class extends superclass {
 
 	static get properties() {
 		return {
+			/**
+			 * @ignore
+			 */
+			animationType: { type: String, reflect: true, attribute: 'animation-type' },
```
margaree, line 51: For the usage of `visible-on-ancestor-mixin` in `button-add`, we only want an opacity animation. The animation is also slightly different than the existing one (300ms ease-in vs 200ms ease-out). This does start to feel a bit custom so let me know if there are alternative ideas for this.

I added the property `animation-type` with the idea there can be different options, since the current default animation is quite specific.

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
```diff
@@ -134,7 +150,7 @@ export const VisibleOnAncestorMixin = superclass => class extends superclass {
 	}
 
 	__voaShow() {
-		if (reduceMotion) {
+		if (reduceMotion && this.animationType !== 'opacity') {
```
margaree, line 153: For our usecase, we want the opacity transition to still run when prefers-reduce-motion (otherwise it gets too flicker-y). I realize that this isn't standard, so let me know other thoughts around this.

dbatiste: Is this flickeryness an issue in other places we use VOA and animations are turned off?

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
```diff
@@ -15,6 +15,17 @@ export const visibleOnAncestorStyles = css`
 		transition: transform 200ms ease-out, opacity 200ms ease-out !important;
 	}
 
+	:host([__voa-state="hidden"][animation-type="opacity"]),
+	:host([__voa-state="hiding"][animation-type="opacity"]) {
+		transform: none !important;
```
dbatiste, line 20: It's too bad we need all these `!important` declarations. I know that the original code is basically forcing you to follow suit with this. I have a sneaking suspicion that the original reason for them was to trump the mixin consumer if they tried to set these properties in their CSS. 🤷‍♂️ Do you see another reason why we would have used them?

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
dbatiste: Yeah, I have a hard time rationalizing the need for it to be different, especially the timing function. Like some other use-case could come along and want something else. But I'm sure Jeff has weighed the options and wouldn't ask for it if it wasn't needed.

Another option would be to expose CSS variables that consumers could override. But I think I like keeping it under wraps as much as possible, and not exposing specific CSS details to consumers that would make it difficult for us to change the animations later.

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
dbatiste: Hmmm, in the `handleTransitionEnd` below, it looks like there may be an assumption made, that `transform` will always be a property that transitions to end. If you haven't already, it might be good to step through this handler or add console logs to make sure it's running as it should.

If we are still transitioning `opacity` (but not `transform`) when reduced motion is turned on, then that check might need to be a little different. Maybe checking `transform` or `opacity` would be appropriate.

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
margaree: That was my guess as well. I don't know of/didn't encounter another reason.

mixins/visible-on-ancestor/visible-on-ancestor-mixin.js
margaree: The main other use case seems to be in cards. In that case it seems fine, as the cards are spacious enough that going quickly between them the dropdown appearing and disappearing doesn't seem too jarring. It's more of an issue here because of the list usage where the add buttons would be close together. However, it is totally possible for other places to be flickery too.

components/button/button-add.js
```diff
@@ -35,6 +37,11 @@ class ButtonAdd extends PropertyRequiredMixin(FocusMixin(LocalizeCoreElement(Lit
 
 	static get styles() {
 		return css`
+			:host {
+				--d2l-button-add-animation-delay: 50ms;
+				--d2l-button-add-animation-duration: 200ms;
+				--d2l-button-add-hover-focus-color: var(--d2l-color-celestine-minus-1);
```
dbatiste, line 43: Looking at the diffs... the focus color, at least of the border around the text/icon seems darker than our normal buttons. Is this expected?

margaree: This is expected - since we are using celestine already for the icon + text case, we wanted to use a different colour for the hover/focus state

dbatiste (approved):

PR #4328, opened 2024-01-10
Add delay and ease-out animation when radio button in dropdown menu is pressed

saifaldin14:
JIRA Ticket: https://desire2learn.atlassian.net/browse/GAUD-144

Description: When the user opens a dropdown menu and clicks a radio button the menu will close after a short delay with an ease-out animation. This is the animation in action:

https://github.com/BrightspaceUI/core/assets/24593380/7873a8f3-4a3a-42a8-822d-29b458151fff


components/dropdown/dropdown-content-styles.js
```diff
@@ -42,6 +44,10 @@ export const dropdownContentStyles = css`
 		top: auto;
 	}
 
+	:host([close-action]) {
+		animation: var(--d2l-dropdown-close-animation-name) 300ms ease-out;
+	}
+
```
margaree, line 50: We would also need to handle the `prefers-reduced-motion` case (see [below in the file](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-content-styles.js#L182) and [here](https://desire2learn.atlassian.net/wiki/spaces/Accessibility/pages/1435995795/Motion) for more details) for users that have that setting enabled

components/dropdown/dropdown-content-styles.js
```diff
@@ -214,6 +220,16 @@ export const dropdownContentStyles = css`
...
+	@keyframes d2l-dropdown-close-animation {
+		0% { display: inline-block; opacity: 1; transform: translate(0, 0); }
+		100% { display: none; opacity: 0; transform: translate(0, -10px); }
+	}
+
+	@keyframes d2l-dropdown-close-animation-dark {
+		0% { opacity: 0.9; transform: translate(0, 0); }
+		100% { opacity: 0; transform: translate(0, -10px); }
+	}
+
```
margaree, line 232: Is the display needed in the one but not the other?

components/dropdown/dropdown-menu.js
```diff
@@ -65,7 +65,13 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 		if (e.target.getAttribute('role') !== 'menuitemradio') {
 			return;
 		}
-		this.close();
+
+		// Add a small delay before closing the menu
+		const duration = 50;
```
margaree, line 70: This can be declared as a global const at the top of the file (e.g., https://github.com/BrightspaceUI/core/blob/main/components/calendar/calendar.js#L15)

components/filter/test/filter-tags.vdiff.js
```diff
@@ -124,7 +124,7 @@ describe('filter-tags', () => {
 			await sendKeys('press', 'ArrowRight');
 			await sendKeys('press', 'Delete');
 
-			await aTimeout('300'); // clearing has a 200ms delay
+			await aTimeout('400'); // clearing has a 200ms delay
```
margaree, line 127: Is this change still needed?

components/dropdown/dropdown-content-styles.js
margaree: Do you know if we see this being used in other close cases at any point? If it's going to stay local to `dropdown-menu` then it might make sense to just add a property and applicable styles to `dropdown-menu`

components/filter/test/filter-tags.vdiff.js
nathankoop: what does the comment mean in this context? It seems to be suggesting that it's clearing an existing 200ms delay

saifaldin14: The comment was already here, I changed the test before but now I've resolved Margaree's previous comments and I don't need to mess with the tests anymore

components/dropdown/dropdown-content-styles.js
```diff
@@ -192,6 +194,12 @@ export const dropdownContentStyles = css`
 		}
 	}
 
+	@media (prefers-reduced-motion: reduce) {
+		:host([close-radio]) {
+			animation: var(--d2l-dropdown-close-animation-name) 300ms ease-out;
+		}
+	}
+
```
margaree, line 202: `prefers-reduced-motion: reduce` actually means that someone would not want to see the animation, rather than they would want to see it. In this case what your code would be doing is that the animation would only be seen by someone who wants that reduced motion.

Another example of this being used in the file is [here](https://github.com/BrightspaceUI/core/blob/5f94fe7938ca697207bdcb0027fd3d8a2d6257b1/components/dropdown/dropdown-content-styles.js#L191) where the animations that were previously set are disabled for the `prefers-reduced-motion` case.

Given that you've moved `close-radio` to be local to `dropdown-menu`, these animation styles that are specific to `dropdown-menu` could also move in there (not sure if others on Gaudi have thoughts on that though - perhaps we do just want all dropdown-content related styles in this file)

components/dropdown/dropdown-menu.js
```diff
@@ -65,7 +81,15 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
...
-		this.close();
+
+		// Add a small delay before closing the menu
+		setTimeout(() => {
+			this.closeRadio = true;
+			this.close();
+		}, closeTimeoutDuration);
+
+		// Wait until animation is complete before toggling value
+		setTimeout(() => { this.closeRadio = false; }, 100);
```
margaree, line 92: We likely just want `this.closeRadio = false` in `_onClose` (called after the `d2l-dropdown-close` event) in this file, though you likely would want to test that that works out okay timing-wise to allow the animation to run.

saifaldin14: I moved it to _onClose but I still need to wait for the animation to complete before toggling it off. If I have it without the timeout the animation doesn't play at all or gets cut off

components/dropdown/dropdown-menu.js
```diff
@@ -3,6 +3,9 @@ import { DropdownContentMixin } from './dropdown-content-mixin.js';
 import { dropdownContentStyles } from './dropdown-content-styles.js';
 import { ThemeMixin } from '../../mixins/theme/theme-mixin.js';
 
+// The duration of the delay needs to be atleast as long as the animation (300ms in this case)
+const closeTimeoutDuration = 300;
```
margaree, line 7: I had thought this delay was for the time between when the user clicks an item and when the close animation happens, and for that case we wanted 50. Has this changed?

components/dropdown/dropdown-menu.js
```diff
@@ -12,6 +15,19 @@ import { ThemeMixin } from '../../mixins/theme/theme-mixin.js';
  */
 class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 
+	static get properties() {
+		return {
+			/**
+			 * Close dropdown when radio is pressed.
+			 * @type {boolean}
+			 */
+			closeRadio: {
```
margaree, line 24: Since the styles were moved into this file then this can be treated as an internal property (e.g., [Daylight docs](https://daylight.d2l.dev/developing/best-practices/attributes-properties/#do%3A-use-a-leading-underscore-for-private-properties)). Would end up more as:
```
_closeRadio: { type: Boolean, reflect: true, attribute: '_close-radio' }
```
and you would remove the description.

Just as an FYI for future work - the comments above the property description are used to generate the docs for the Daylight site. For properties where we wouldn't want consumers to actually use them, such as this one, we would instead set the description to `@ignore`. If you do add an underscore in front of the property then you don't have to worry about that, as those properties get ignored by Daylight.

components/dropdown/dropdown-menu.js
saifaldin14: Yeah, when I was testing it out further the 50ms delay didn't allow the animation to actually finish it would cut out abruptly. When I tried values that were equal to greater than the animation length the animation would have enough time to play

Using 50ms delay:

https://github.com/BrightspaceUI/core/assets/24593380/a3fe546c-0ade-4b34-a5a5-26ffae6767c3

Using 300ms delay:

https://github.com/BrightspaceUI/core/assets/24593380/3c9fa756-ee28-46b9-b99a-fc196a78f9cb

components/dropdown/dropdown-menu.js
```diff
@@ -12,14 +15,58 @@ import { ThemeMixin } from '../../mixins/theme/theme-mixin.js';
...
+					}
+				}
+				@keyframes d2l-dropdown-close-animation {
+					0% { display: inline-block; opacity: 1; transform: translate(0, 0); }
+					100% { display: none; opacity: 0; transform: translate(0, -10px); }
+				}
+
+				@keyframes d2l-dropdown-close-animation-dark {
+					0% { display: inline-block; opacity: 0.9; transform: translate(0, 0); }
+					100% { display: none; opacity: 0; transform: translate(0, -10px); }
```
margaree, line 59: I'm wondering if it's possible to change this such that we don't need to modify `display` within these animations?

margaree (commented):
It's coming together! It could be useful to add a visual diff for a radio dropdown menu to confirm it closes properly on selection, since that would have caught the one issue with `prefers-reduced-motion` I mention in the comments.

components/dropdown/dropdown-menu.js
```diff
@@ -36,6 +81,7 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 			this.maxHeight = this._maxHeightNonTray;
 		}
 
+		this.addEventListener('animationend', this._closeRadioAction);
```
margaree, line 84:
```suggestion
		this.addEventListener('animationend', this._onAnimationEnd);
```

components/dropdown/dropdown-menu.js
```diff
@@ -61,11 +107,24 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
...
 		if (e.target.getAttribute('role') !== 'menuitemradio') {
 			return;
 		}
-		this.close();
+
+		if (reduceMotion) {
+			// Don't trigger the animation but still wait before closing the dropdown
+			setTimeout(() => {
+				this.close();
+			}, dropdownDelay);
```
margaree, line 123: Should this one also be 300ms or should it be the 50ms?

components/dropdown/dropdown-menu.js
```diff
@@ -12,14 +14,57 @@ import { ThemeMixin } from '../../mixins/theme/theme-mixin.js';
...
+				}
+
+				:host([theme="dark"]) {
+					--d2l-dropdown-close-animation-name: d2l-dropdown-close-animation-dark;
+				}
+
+				:host([_close-radio]) {
+					animation: var(--d2l-dropdown-close-animation-name) ${dropdownDelay}ms ease-out;
+					animation-delay: 50ms;
+					display: inline-block;
```
margaree, line 42: Something to consider is that if you move the `this.close()` to your animationend event handler (for non-reduced motion), then you can get rid of this and not have style duplication (e.g., if for some reason we were to want to change dropdowns to be something other than `inline-block` in the future, we'd have to know to update here as well).

margaree (commented):
Were you able to get the vdiff test working? I thought I saw it at one point but don't anymore. Let me know if you need help with that.

components/dropdown/test/dropdown-menu.vdiff.js
```diff
@@ -45,6 +45,14 @@ const menuWithHeaderFooter = html`
 	<div slot="footer">Available 2020</div>
 `;
 
+const dropdownRadioMenu = html`
```
margaree, line 48: If this is only used the once you can just put it inline with the rest of the dropdown

components/dropdown/test/dropdown-menu.vdiff.js
```diff
@@ -135,4 +143,22 @@ describe('dropdown-menu', () => {
...
+					<d2l-dropdown-menu theme="dark" class="vdiff-include">
+						${dropdownRadioMenu}
+					</d2l-dropdown-menu>
+				</d2l-dropdown>
+			</div>
+		`);
+		const dropdown = elem.querySelector('d2l-dropdown');
+		dropdown.toggleOpen();
+		await oneEvent(dropdown, 'd2l-dropdown-open');
+		clickElem(elem.querySelector('#select-radio'));
```
margaree, line 161: I think you may also need to wait for the `d2l-dropdown-close` event - looking at the resulting image it looks like it's not fully closed.

margaree (commented):
The dropdown menu code is looking good!

components/dropdown/test/dropdown-menu.vdiff.js
```diff
@@ -135,4 +136,27 @@ describe('dropdown-menu', () => {
 		await oneEvent(dropdown, 'd2l-dropdown-open');
 		await expect(dropdown).to.be.golden();
 	});
+
+	it('radio-button-close', async() => {
+		const elem = await fixture(html`
+			<div style="background-color: #000000; padding: 250px;">
```
margaree, line 142: Super nit - this is running the test in dark mode, but it doesn't really matter either way since it's just testing functionality

margaree (approved):

PR #4342, opened 2024-01-15
feat: Added inline help slot for self-contained components; input-checkbox, input-color, input-time

saifaldin14:
Added inline help slots for the self-contained components before moving on to the connected components.

input-checkbox:
Screenshot (9)

input-checkbox (multiline):
Screenshot (16)

input-color:
Screenshot (10)

input-time:
Screenshot (11)

input-time (multiline):
Screenshot (19)


components/inputs/input-styles.js
```diff
@@ -114,3 +114,8 @@ export const inputStyles = css`
 		background-image: none;
 	}
 `;
+
+export const inlineHelpStyles = {
```
dbatiste, line 118: I think we should rely on CSS class instead. We generally don't export styles for inline.
```javascript
export const inlineHelpStyles = css`
    .d2l-input-inline-help {
        marginTop: '0.5rem'
    }
`;
```

Also, I think we should not be specifying `fontSize: '15px'` here. At least for the checkbox case, the font-size will come from `d2l-body-small`, which I think is correct.

components/inputs/input-color.js
dbatiste, line 1: I think we still need to add `super.styles` to `d2l-input-color`'s styles block. Currently we have:
image

Once we add it, we will see the correct styles:
image

components/inputs/demo/input-time.html
```diff
@@ -53,6 +53,29 @@ <h2>Custom Default Time</h2>
 					<d2l-input-time label="Start Time" default-value="09:00:00"></d2l-input-time>
 				</template>
 			</d2l-demo-snippet>
+
+			<h2>Inline help text</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-input-time label="Start Time" default-value="09:00:00">
```
dbatiste, line 60: Is this what we expect for `d2l-input-time`? Initially I'd expected it to look like the others, but then discovered that we're actively limiting the width, so maybe this is ok. @margaree might have more context here.
image

margaree: This PR https://github.com/BrightspaceUI/core/pull/1089/files is context on why we are limiting the width. We could possibly move that calculation into `input-date-time` since that's where that problem was but I'm not sure if there are places now relying on its size being as small as possible by default.

components/inputs/input-checkbox.js
```diff
@@ -151,6 +157,9 @@ class InputCheckbox extends FocusMixin(SkeletonMixin(RtlMixin(LitElement))) {
 					bottom: 0.3rem;
 					top: 0.3rem;
 				}
+				.d2l-input-inline-help {
+					margin-left: ${cssSizes.inputBoxSize + cssSizes.checkboxMargin}rem;
```
dbatiste, line 161: We can change this to `margin-inline-start` to make this work for both LTR and RTL.

Our alternative (old) approach is:
```css
.d2l-input-inline-help {
    margin-left: ${cssSizes.inputBoxSize + cssSizes.checkboxMargin}rem;
}
:host([dir="rtl"]) .d2l-input-inline-help {
    margin-left: 0;
    margin-right: ${cssSizes.inputBoxSize + cssSizes.checkboxMargin}rem;
}
```

components/inputs/demo/input-time.html
saifaldin14: When I moved the max-width to the dropdown or altered it the component looped around itself. The label text was in the same line as the input text

components/inputs/input-inline-help-mixin.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	static get styles() {
+		const styles = [ bodySmallStyles, css`
+			:host([_has-inline-help]) .d2l-input-inline-help {
+				display: block;
+			}
+			.d2l-input-inline-help {
+				display: none;
+				margin-top: 0.5rem !important;
```
dbatiste, line 19: Given that we potentially have cases where the inline-help could be limited in width, I think it would be worthwhile adding `overflow-wrap: anywhere;` to handle unbreakable text cases. 

In case you are curious, despite what it sounds like, [anywhere](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere) is actually pretty smart.

dbatiste (approved):

PR #4351, opened 2024-01-16
fix: adjust date and date-time min/max validation language to include min/max value (GAUD-698)

dlockhart:
Slight variation based on @lideng928's #4277 (which was done in a fork so I can't modify it directly) with suggested language from @geurts.

Since this is a change to an existing language term I've added "2" to the end of the terms, essentially creating new terms. I _think_ this is the first time we've done this, or at least it's the first time we've done it this way.

Resolve https://github.com/Brightspace/nova/issues/5730.


margaree (approved):

dlockhart:
Apparently we have unit tests for this, neat! Will update.

svanherk:
Why is the `2` necessary? We're not hooked up to OSLO, and I'd _expect_ the langterms to be re-sent because the English term changed? Or does the database only look at term name, not contents?

dlockhart:
> Why is the 2 necessary?

So normally we'd worry about existing client overrides, but yeah it's a good point about there being no OSLO and therefore no client overrides so maybe we don't care? I think that's correct. I'll switch this back to remove the `2`.

PR #4358, opened 2024-01-17
fix: add correct padding

saifaldin14:
Eliminate uneven vertical padding for empty dropdown:

Screenshot (15)


components/dropdown/dropdown-content-styles.js
```diff
@@ -128,7 +128,7 @@ export const dropdownContentStyles = css`
 	.d2l-dropdown-content-bottom {
 		box-sizing: border-box;
 		max-width: 100%;
-		min-height: 5px;
+		min-height: 0.3rem;
```
margaree, line 131: Did the design call for this change or was this more to simplify the calculation for the bottom padding? If the latter, you could leave this as `5px` and then in `filter` use a `calc` to get the padding-bottom value.

saifaldin14: Yeah, I changed it to use calc instead

margaree (approved):
Looks good!

PR #4361, opened 2024-01-18
GAUD-329: Add Button: List: button-add at the top of the list

margaree:
Related to getting the nested lists to work, we want a `button-add` at the top of a list as well. 

In nested lists this top one will be displayed and the bottom one of the parent list item will not (or if possible the parent one will show when hovering left/right (rtl) of the nested button-add. That will be done in a subsequent PR.

I'll add comments/questions throughout.


components/list/list-item-generic-layout.js
```diff
@@ -69,7 +69,12 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 					[control-end content-start] minmax(0, auto)
 					[content-end actions-start] minmax(0, min-content)
 					[end actions-end];
-				grid-template-rows: [main-start] [main-end add-start] [add-end nested-start] [nested-end];
+				grid-template-rows:
+					[start add-top-start] minmax(0, min-content)
+					[add-top-end main-start] minmax(0, min-content)
+					[main-end add-start] minmax(0, min-content)
+					[add-end nested-start] minmax(0, min-content)
+					[nested-end end];
```
margaree, line 77: This seemed to work okay to get the top add-button slot. Looking in dev tools it looked like this wasn't actually being applied before.

Let me know if there's a preferred way.

components/list/list-item-mixin.js
```diff
@@ -60,6 +60,10 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			 * @type {string}
 			 */
 			color: { type: String },
+			/**
+			 * @ignore
+			 */
+			first: { type: Boolean, reflect: true },
```
margaree, line 66: I added a `first` property to be set by the parent list on the first list item of a given list

components/list/list.js
```diff
@@ -320,7 +320,7 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 	_handleListItemAddButtonClick(e) {
 		e.stopPropagation();
 		/** Dispatched when the add button directly after the item is clicked. Event detail includes the key of the item directly above where the add button was clicked. */
-		this.dispatchEvent(new CustomEvent('d2l-list-add-button-click', { detail: { key: e.target.key } }));
+		this.dispatchEvent(new CustomEvent('d2l-list-add-button-click', { detail: { key: e.target.key, isFirstItem: e.detail.isFirstItem } }));
```
margaree, line 323: This might be a bit weird - we are using the same event and key from list for clicking `button-add` above or below the first item (and then only for below for subsequent items). I wasn't sure/couldn't find if there was precedent for a different way of doing this but let me know any suggestions.

margaree (commented):
I ran all the visual diffs for list with `add-button` enabled and there wasn't anything remarkable that changed - same as previous with things like if separators are between there is now a top and bottom one too.

components/list/list-controls.js
```diff
@@ -19,11 +19,13 @@ export class ListControls extends SelectionControls {
...
 				--d2l-selection-controls-background-color: var(--d2l-list-controls-background-color);
 				--d2l-selection-controls-padding: var(--d2l-list-controls-padding, 18px);
-				z-index: 6; /* must be greater than d2l-list-item-active-border */
 			}
 			:host([no-sticky]) {
 				z-index: auto;
 			}
+			:host([_scrolled]) {
+				z-index: 6;  /* must be greater than d2l-list-item-active-border and d2l-button-add */
+			}
```
margaree, line 28: Otherwise when hover/focused the add-button gets cut off by the controls. I think the scrolled case is the biggest issue here but let me know if there are other cases.

components/list/list-item-mixin.js
```diff
@@ -682,7 +695,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 					class="d2l-list-item-actions-container">
 					<slot name="actions" class="d2l-list-item-actions">${actions}</slot>
 				</div>
-				${this._showAddButton ? html`
+				${this._showAddButton && !this._hasNestedListAddButton ? html`
```
margaree, line 698: We don't want the add button for the last item of the list showing if there is a nested list below it.
We would want it showing if the nested list is collapsed though, so I'll do a bit more playing with this.

components/list/list.js
```diff
@@ -368,6 +371,8 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 
 	_handleSlotChange() {
 		this._updateItemShowingCount();
+		const items = this.getItems();
+		if (items.length > 0) items[0].first = true;
```
dbatiste, line 375: Do we need to iterate over all the items to ensure that all other items have `first = false`?  For example, are there scenarios where a new item could be added to the top of the list (by whatever means, possibly drag & drop or an action performed in the list-controls header, etc), such that the initial item with `first = true` becomes the second item and should be updated as such?

components/list/list-item-generic-layout.js
dbatiste: Seems reasonable.

components/list/list-item-generic-layout.js
```diff
@@ -92,7 +97,7 @@ class ListItemGenericLayout extends RtlMixin(LitElement) {
 			::slotted([slot="outside-control-container"]),
 			::slotted([slot="control-container"]),
 			::slotted([slot="drop-target"]) {
-				grid-row: 1 / 2;
+				grid-row: 2 / 3;
```
dbatiste, line 100: 👍 . This is the bit I was worried about.

components/list/list.js
```diff
@@ -120,6 +120,9 @@ class List extends PageableMixin(SelectionMixin(LitElement)) {
 				--d2l-list-item-color-border-radius: 3px;
 				--d2l-list-item-color-width: 3px;
 			}
+			:host([add-button]) ::slotted([slot="controls"]) {
+				margin-bottom: calc(6px + 0.4rem); /* controls section margin-bottom + spacing for add-button */
```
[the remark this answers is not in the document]
dbatiste: > I didn't have success getting a css sibling selector working (something like when controls and list item are siblings add top margin or padding) but I definitely could have been doing that incorrectly.

`::slotted` supports compound selectors ([spec](https://drafts.csswg.org/css-scoping/#slotted-pseudo)), but it does _not_ support complex selectors (i.e. selectors comprised of more than one compound selector separated by a combinator, like a sibling selector).

components/list/list.js
dbatiste: So the consumer will listen for the one event (which seems good). But for the first item, they need to know whether an item is being added before or after. I suppose we could have `e.detail.position: 'before|after'`, and we default to `after`. 🤷‍♂️

margaree: I like the position idea - it makes the consumer have to be aware that position is used.

components/list/list-item-mixin.js
margaree: I've dealt with the expand/collapse state piece in the follow up nested PR.

components/list/list.js
margaree: Good call!

dbatiste (approved):

PR #4364, opened 2024-01-19
fix: resolve animation issue

saifaldin14:
Resolve the animation issue by setting closeRadio to be false on open


components/dropdown/dropdown-menu.js
```diff
@@ -176,6 +176,8 @@ class DropdownMenu extends ThemeMixin(DropdownContentMixin(LitElement)) {
 
 		menu.resize();
 
+		this._closeRadio = false;
```
margaree, line 179: super nit: I'd probably put this with `this._initializingHeight = true;` to have all the property changes grouped

PR #4366, opened 2024-01-22
feat: Add inline-help for input-text, input-search and input-number

saifaldin14:
Added inline help functionality to input-text, input-search and input-number

input-text (normal)
Screenshot (21)

input-text (multiline)
Screenshot (22)

input-search (normal)
Screenshot (23)

input-search (multiline)
Screenshot (24)

input-number (normal)
Screenshot (25)

input-number (multiline)
Screenshot (26)


components/inputs/input-number.js
```diff
@@ -363,6 +365,7 @@ class InputNumber extends FocusMixin(LabelledMixin(SkeletonMixin(FormElementMixi
 					<slot slot="left" name="left"></slot>
 					<slot slot="right" name="right"></slot>
 					<slot slot="after" name="after"></slot>
+					${this._renderInlineHelpNested(this._inlineHelpId)}
```
dbatiste, line 368: Is `_renderInlineHelpNested` needed? I was thinking that all we would need here is:
```javascript
<slot slot="inline-help" name="inline-help"></slot>
```
In such case, I don't think these components need to extend `InputInlineHelpMixin`. The slot change handler and styles will already be taken care of by `d2l-input-text`. Generating the id is already handled too.

components/inputs/input-search.js
```diff
@@ -67,7 +69,7 @@ class InputSearch extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement)))
 	}
 
 	static get styles() {
-		return [inputStyles, css`
+		return [ super.styles, inputStyles, css`
```
dbatiste, line 72: Per previous comment, I think pretty much everything will be handled by `d2l-input-text`, in which case this `super.styles` should not be needed here. I think we just need to expose the slot with `<slot slot="inline-help" name="inline-help"></slot>`.

components/inputs/test/input-number.vdiff.js
```diff
@@ -4,6 +4,28 @@ import { clickAt, expect, fixture, focusElem, html, oneEvent, sendKeysElem } fro
 
 const requiredFixture = html`<d2l-input-number label="Number" required></d2l-input-number>`;
 const simpleFixture = html`<d2l-input-number label="Number"></d2l-input-number>`;
+const inlineHelpComponents = {
```
dbatiste, line 7: Nit: let's be consistent with the variable naming... let's call this `inlineHelpFixtures`.

dbatiste (approved):
Looks good!

PR #4369, opened 2024-01-22
GAUD-329: Add Button: List: Nested

margaree:
This handles miscellaneous issues (will comment) remaining to do with `button-add` in nested lists and also adds demos and vdiff tests.

~This PR comes after this PR https://github.com/BrightspaceUI/core/pull/4361 (merged). This PR https://github.com/BrightspaceUI/core/pull/4382 should probably go in first.~


components/list/list-item-mixin.js
```diff
@@ -146,7 +146,6 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			:host([_focusing-primary-action]) [slot="control-container"]::after,
 			:host([selected]:not([selection-disabled]):not([skeleton])) [slot="control-container"]::before,
 			:host([selected]:not([selection-disabled]):not([skeleton])) [slot="control-container"]::after,
-			:host([_show-add-button]) [slot="control-container"]::after,
```
margaree, line 149: This keeps the border color when there is an add button rather than setting it to transparent. This is due to the fact we do still want the full border for nested lists.

I'm going to see if I can target this more specifically to at least just those cases. The impact of this is that you see a tiny bit of border behind the + icon in the add button.

components/list/list-item-mixin.js
```diff
@@ -695,7 +696,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 					class="d2l-list-item-actions-container">
 					<slot name="actions" class="d2l-list-item-actions">${actions}</slot>
 				</div>
-				${this._showAddButton && !this._hasNestedListAddButton ? html`
+				${this._showAddButton && (!this._hasNestedListAddButton || (this.expandable && !this.expanded)) ? html`
```
margaree, line 699: We want to show the bottom add button only when it's not expandable, or if it is expandable then it must not be expanded.

dbatiste (approved):

PR #4377, opened 2024-01-23
feat: Add inline functionality for input-percent, input-date and input-textarea

saifaldin14:
Added inline functionality for these components:

input-percent (normal):
Screenshot (28)

input-percent (multiline):
Screenshot (29)

input-date (normal):
Screenshot (30)

input-date (multiline):
Screenshot (31)

input-textarea (normal):
Screenshot (32)

input-textarea (multiline):
Screenshot (33)


components/inputs/input-textarea.js
```diff
@@ -252,6 +256,7 @@ class InputTextArea extends FocusMixin(LabelledMixin(FormElementMixin(SkeletonMi
 					.value="${this.value}">${this.value}</textarea>
 				${this.validationError ? html`<d2l-tooltip for=${this._textareaId} state="error" align="start">${this.validationError}</d2l-tooltip>` : null}
 			</div>
+			${this._renderInlineHelp(this._inlineHelpId)}
```
dbatiste, line 259: I think we need to look at supporting skeletons. There are two different approaches that we can take, can we should discuss with @geurts.

1. Update `InputInlineHelpMixin` to include the `d2l-skeletize` class. With this approach, the inline help would only be skeletized if the input component supports skeletons. As such, components that do not support skeletons (like `d2l-input-color`), would never have their inline help skeletized; or
2. Same as above except also update `InputInlineHelpMixin` to extend `SkeletonMixin`. With this approach, the inline help could be skeletized even if the input component doesn't support skeletons. Note that `SkeletonMixin` is deduped, so there is no concern with it being extended by both the input component and `InputInlineHelpMixin`.

Once we decide which approach we want to take, we should confirm this is working for all the input components, and we also want a visual-diff for each in skeleton mode. To put the component into skeleton mode, simply apply the `skeleton` attribute to the component. 

If you want to address skeletons for inline help as a separate PR, that's totally fine, since I know you'll need to go back and add visual-diff tests for other input components.

Regardless of the approach, these screenshots show the current code vs with the `d2l-skeletize` class.

`d2l-input-textarea` **without** `d2l-skeletize` (incorrect):
image

`d2l-input-textarea` **with** `d2l-skeletize` (correct):
image

dbatiste (approved):
Looks good. There's just the skeleton case we need to address. As discussed offline, the skeletons will be addressed in a separate PR.

bearfriend:
Is that `filter-tags` test just being flaky?

dbatiste:
> Is that `filter-tags` test just being flaky?

Yeah seems to be. Definitely not expecting that change. 🤷‍♂️

svanherk:
> > Is that `filter-tags` test just being flaky?
> 
> Yeah seems to be. Definitely not expecting that change. 🤷‍♂️

It's been really flaky for a little while now, even after https://github.com/BrightspaceUI/core/pull/4199. I think the focused state is the correct one, so the open PR should actually be merged in to remove the change to it in this PR

bearfriend:
> It's been really flaky for a little while now, even after #4199. I think the focused state is the correct one, so the open PR should actually be merged in to remove the change to it in this PR

And then 🤞 it doesn't flake again

PR #4380, opened 2024-01-24
fix: Fix lopsided padding on the right

saifaldin14:
Ticket: https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-710

Fix the extra padding right for empty state container. Moved the padding to the button, so that if it's not there there's no extra padding

Original Issue:
image-20240117-171633

Padding now:
Screenshot (34)

When a button exists the padding is still here:
Screenshot (38)

When the language is changed to Arabic (rtl):

No padding without button:
Screenshot (37)

Padding in correct place with button:
Screenshot (39)


components/empty-state/empty-state-styles.js
```diff
@@ -30,16 +36,10 @@ export const emptyStateSimpleStyles = css`
...
 
-	:host([dir="rtl"]) .d2l-empty-state-description {
-		padding-left: 0.5rem;
-		padding-right: 0;
-	}
-
 	.d2l-empty-state-description {
 		display: inline;
-		padding-right: 0.5rem;
+		margin: 0;
```
margaree, line 41: I played around with this a bit. I think in order to get the height right when there's just the description, you also need `line-height: 28px`. I believe you also want to only have the `column-gap` when there's a description, i.e., `:host([description]) .empty-state-container`.

Also, looking at the `normal-link-rtl` (and other link rtl vdiffs) I'm not totally sure what's going on - they now look consistent with the button vdiffs in rtl but I don't know if what's there before was actually considered correct.

components/empty-state/empty-state-simple.js
```diff
@@ -18,20 +18,34 @@ class EmptyStateSimple extends PropertyRequiredMixin(RtlMixin(LitElement)) {
...
-			<slot class="action-slot"></slot>
+			<div class="empty-state-container">
+				<p class="d2l-body-compact d2l-empty-state-description">${this.description}</p>
+				<slot class="action-slot" @slotchange="${this._handleActionSlotChange}"></slot>
+			</div>
 		`;
 	}
 
+	_handleActionSlotChange() {
+		this._hasDescription = this.description.trim().length > 0;
```
margaree, line 46: Were you able to get the vdiffs running locally? It looks like the ones on the PR weren't running because if `this.description` is undefined then we can't use `trim`.

It's actually possible to do this all in the css though by checking if `description` itself is defined and then you can avoid an extra property and slotchange listener:
```
:host([description]) .empty-state-container
```
which could live in the styles file

PR #4382, opened 2024-01-24
Add Button: List: Fixes

margaree:
Fixes the following:
- Border was sometimes visible under the add-button icon
- Increases the click target of the line

Note on the border:
This was complicated for 2 cases:
- For nested lists we do want to see border to the left past the add button for the first item (to show that when clicking the add button it would correspond to the nested list)
- For selectable items we want the rounded border, but then to hide the top and bottom. This doesn't work with the rounded corners.
For both of these cases I used gradients. The selectable one should be safe, but the nested one is a bit of an estimate and if it were super wide I could see there being cases where the 20% "border" wouldn't be enough. Let me know any alternative suggestions.


components/list/list-item-mixin.js
```diff
@@ -125,6 +125,14 @@ export const ListItemMixin = superclass => class extends composeMixins(
...
 				width: 100%;
 			}
+            :host([_show-add-button][_has-nested-list]) [slot="control-container"]::after {
+                background-clip: content-box, border-box;
+                background-image: linear-gradient(white, white), linear-gradient(to right, var(--d2l-color-mica) 20%, transparent 20%, transparent 80%, var(--d2l-color-mica) 80%);
+                background-origin: border-box;
+            }
+			:host([_show-add-button][_has-nested-list]:not([selection-disabled]):not([skeleton])[selected]) [slot="control-container"]::after {
+                background-image: linear-gradient(white, white), linear-gradient(to right, #b6cbe8 20%, transparent 20%, transparent 80%, #b6cbe8 80%);
+			}
```
margaree, line 135: This is for the case where there's a nested list and we want to show the first part of the border.
We could actually remove the end part being mica as well, that was just leftover from the gradient below.

components/list/list-item-mixin.js
```diff
@@ -265,9 +273,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				margin: 0 -12px;
 			}
 			.d2l-list-item-content-extend-separators [slot="outside-control-container"] {
-				border-left: none;
-				border-radius: 0;
-				border-right: none;
+				border-left: none !important;
+				border-radius: 0 !important;
+				border-right: none !important;
```
margaree, line 278: The `!important` was used because otherwise the `border` from the new css rules below take priority.

components/list/list-item-mixin.js
```diff
@@ -387,8 +410,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 
 			[slot="add"],
 			[slot="add-top"] {
-				margin-bottom: -4px;
-				margin-top: -3px;
+				margin-bottom: -6px;
+				margin-top: -5px;
```
margaree, line 414: Related to the increase in line click target

components/list/demo/list-item-actions.html
```diff
@@ -296,7 +296,7 @@ <h2>Grid Actions with Navigation (href) Primary Action</h2>
 
 			<d2l-demo-snippet id="grid-link-demo">
 				<template>
-					<d2l-list grid>
+					<d2l-list grid add-button>
```
margaree, line 299: Jeff was interested in this example scenario

components/button/button-add.js
```diff
@@ -54,8 +54,9 @@ class ButtonAdd extends RtlMixin(PropertyRequiredMixin(FocusMixin(LocalizeCoreEl
 				cursor: pointer;
 				display: flex;
 				font-family: inherit;
-				height: 0.9rem;
+				height: 11px;
```
margaree, line 57: setting height to be 1px (line) + 5 px on each side of the line.
The margins below are to prevent nearby content from overlapping with the + / + and text

components/list/list-item-mixin.js
```diff
@@ -125,6 +125,13 @@ export const ListItemMixin = superclass => class extends composeMixins(
 				position: absolute;
 				width: 100%;
 			}
+			:host([_show-add-button][_has-nested-list]) [slot="before-content"] {
+				border-bottom: 1px solid var(--d2l-color-mica);
+				margin-bottom: -1px;
+			}
+			:host([_show-add-button][_has-nested-list]:not([selection-disabled]):not([skeleton])[selected]) [slot="before-content"] {
+				border-bottom-color: #b6cbe8;
```
dbatiste, line 133: I guess we couldn't use one of our palette colors here eh?

dbatiste: Oh I see this is already used. `/* celestine alpha 0.3 */`

components/list/list-item-mixin.js
```diff
@@ -387,8 +409,8 @@ export const ListItemMixin = superclass => class extends composeMixins(
 
 			[slot="add"],
 			[slot="add-top"] {
-				margin-bottom: -4px;
-				margin-top: -3px;
+				margin-bottom: -12.5px;
```
dbatiste, line 412: How does the browser handle the half pixels? Is this just the same as `-12px` or `-13px`?

dbatiste: On a separate note, hopefully these negative margins don't cause us any issues. I think we had an issue in the past where a `-1px` margin was causing scrollbars to appear somewhere, possibly in a scrollable container.

dbatiste (approved):

components/list/list-item-mixin.js
margaree: Yup! I'll add a comment to indicate that as in the other places

components/list/list-item-mixin.js
margaree: I'll re-test these numbers. Eek I also hope not! Will keep an eye on that.

PR #4386, opened 2024-01-25
feat: [DE50021] Course Merge - UI - Filter - Multiple selections in quick succession do not filter correctly

JayArora1997:
[DE50021](https://rally1.rallydev.com/#/?detail=/defect/646778496077&fdp=true): Course Merge - UI - Filter - Multiple selections in quick succession do not filter correctly

Change timeout to de-bouncing which reduces the numbers of events triggered when items selected in quick succession.

** Will help to reduce load times for [hm-filter](https://github.com/BrightspaceHypermediaComponents/common/blob/main/components/d2l-hm-filter/d2l-hm-filter.js) as it further makes requests for each change.


components/filter/filter.js
```diff
@@ -594,8 +593,6 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			detail: { allCleared: allCleared, dimensions: dimensions }
 		}));
 		this._changeEventsToDispatch = new Map();
-		this._changeEventTimeout = null;
```
svanherk, line 597: I'm thinking we still want `clearTimeout(this._changeEventTimeout);` here, so in cases where `_dispatchChangeEventNow` is called directly (when the "Clear" or "Clear All" button is pressed), if a timeout has already started it won't run again needlessly in a few milliseconds?

components/filter/filter.js
```diff
@@ -573,13 +573,12 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 
 	_dispatchChangeEvent(dimension, change) {
 		this._setDimensionChangeEvent(dimension, change, false);
+		clearTimeout(this._changeEventTimeout);
```
svanherk, line 576: So just to make sure I'm understanding the change correctly - instead of simply batching the events and running every 200ms, we now wait until there's been 300ms of inactivity before we send them. Is this necessary for the changes in https://github.com/BrightspaceHypermediaComponents/common/pull/52, or just helpful to reduce the HM calls over there?

Overall it seems like a good idea that even non-HM filter uses could benefit from, but want to make sure I understand the context.

components/filter/filter-tags.js
```diff
@@ -8,7 +8,7 @@ import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
  * A tag-list allowing the user to see (and remove) the currently applied filters.
  */
 
-const CLEAR_TIMEOUT = 210; /** Corresponds to timeout in _dispatchChangeEvent in filter + 10 ms */
+const CLEAR_TIMEOUT = 310; /** Corresponds to timeout in _dispatchChangeEvent in filter + 10 ms */
```
svanherk, line 11: Any reason for the jump from 200 to 300? Did the extra amount cut down on HM calls even more than adding just the debouncing?

components/filter/filter.js
JayArora1997: Yes, now we will fire event after 300ms of inactivity. I changed to 300s as normally when selecting too fast, it seemed to be good enough time to select multiple filters, that will reduce the triggered events. It would reduce the the number of fired events for both HM and non-HM filters

components/filter/filter-tags.js
JayArora1997: I changed to 300s as normally when selecting too fast, it seemed to be good enough time to select multiple filters, that will reduce the triggered events. This would just cut the calls to update the list in non-HM. In case of HM, lot of calls are associated with toggling the filters and then updating the list which can be reduced!

components/filter/filter.js
JayArora1997: I think when "clear" or "clear All" button is pressed, we don't want to reset the timeout as we would want it to just go through and trigger the event immediately. When someone press clear and then apply a filter in a quick succession, it would be useful in that case but clearing it in "_dispatchChangeEventNow" would lead to calling it immediately instead of waiting for the user to stop interacting which would end up firing more events which we are trying to reduce.

svanherk: Not sure I follow - it's been a while since I was in this code, but it seems like:
1. A user clicks a bunch of filters and then clicks "Clear"
    - This results in a timeout being set for the first filter clicks, but then when "Clear" is pressed `_dispatchChangeEventNow` will be called and all pending changes will be immediately handled and the event immediately sent. If we don't clear the timeout, `_dispatchChangeEventNow` will run again later with no events left to process.
2. A user clicks "Clear" and then clicks a bunch of filters
    - When "Clear" is pressed, `_dispatchChangeEventNow` will be called and the event immediately sent. It'll try to clear a timeout, but there won't be any anyways. Then the following filter clicks will set a timeout to be handled.
3. A user clicks a bunch of filters, then clicks "Clear", and then clicks a bunch of filters
    - Same as 1 - this results in a timeout being set for the first filter clicks, but then when "Clear" is pressed `_dispatchChangeEventNow` will be called and all pending changes will be immediately handled and the event immediately sent. We clear the timeout because we just handled all the events, but then the next round of clicks would reset that timeout.

Unless you're seeing a race condition in there? Worst case, calling `_dispatchChangeEventNow` again with no events left to process shouldn't be a huge issue

JayArora1997: I got what you are trying to say, its very unlikely of this to occur as I had to be very fast with mouse to reproduce this and that to only with first filter as its hard to press clear under 300s when selecting filters. But yes, I tested with your suggested change and it would help to reduce the case that your mentioned without causing any issue. Thank you

zwilson2 (approved):

svanherk (approved):
Approving now since I'm off Monday - this looks good to me, ideally https://github.com/BrightspaceUI/core/pull/4392 gets in as well to avoid the vdiff flake

JeremyWinkler (approved):

JayArora1997:
### [Dev Testing](https://github.com/BrightspaceHypermediaComponents/common/pull/52#issuecomment-1919415240)

PR #4390, opened 2024-01-26
Don't try to set up resize observer if there's no content to observe

grant-cleary:
Curious about thoughts on this one, because I'm not super sure if there are ramifications here that I'm not thinking of. To keep a long story short, it looks like there can be some interesting timing shenanigans that go on when something extends both `HierarchicalViewMixin` and `LocalizeDynamicMixin`, in which we end up trying to hook up the resize observer before the element has actually rendered (all that exists is the shadowroot).

This cropped up in the HTML editor ages ago and we worked around it at the time (https://github.com/BrightspaceUI/htmleditor/blob/main/components/toolbar/table-create.js#L88), but in some testing I did today, I noticed a couple things:
- When this fails (`content` is null), it seems to always be caused by the Lit update triggered by the lang terms being fetched.
- The failures happen in the middle of rendering, but always seem to be followed by successful hookups, so I think in reality the observer is still being connected properly; it's just failing on _some_ updates.

Independent of the above, I'm also just not sure what else we could or should try to do if we find that the element is connected but hasn't rendered yet. There's nothing to observe at that point and it feels like a bit of a violation of Lit's lifecycle rules to wait until an update has finished during `connectedCallback`. Maybe I'm off base there though.

EDIT: This was really just making a long story longer... So many words for one early exit. 😅

### DIFF
--- a/components/hierarchical-view/hierarchical-view-mixin.js
+++ b/components/hierarchical-view/hierarchical-view-mixin.js
@@ -554,6 +554,8 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 
 	__startResizeObserver() {
 		const content = this.shadowRoot.querySelector('.d2l-hierarchical-view-content');
+		if (!content) return;
+
 		this.__bound_dispatchViewResize = this.__bound_dispatchViewResize || this.__dispatchViewResize.bind(this);
 		this.__resizeObserver = this.__resizeObserver || new ResizeObserver(this.__bound_dispatchViewResize);
 		this.__resizeObserver.disconnect();


components/hierarchical-view/hierarchical-view-mixin.js
```diff
@@ -554,6 +554,8 @@ export const HierarchicalViewMixin = superclass => class extends superclass {
 
 	__startResizeObserver() {
 		const content = this.shadowRoot.querySelector('.d2l-hierarchical-view-content');
+		if (!content) return;
```
dbatiste, line 557: So this guards against wiring up the observer to `null`/`undefined`. But it still seems like there would still be an issue, because `__startResizeObserver` is only called from a rAF callback within `connectedCallback`. In effect, the observer would never get wired up.

It would seem that the rAF is used at least in part to wait until `render` has completed before trying to do the wire-up. But in certain cases that doesn't seem to be enough? Perhaps the wire-up should be done from `firstUpdated`, or it should be wired up to the `host` instead, both of which could have adverse effects on the behaviour of hierarchical-view and menus.

grant-cleary: I think it's definitely possible that there are cases where this wouldn't get wired up; however, in the one case I've seen this go wrong (using both the dynamic localize mixin and this), this always seems to reconnect later after render has succeeded. So in some ways, this perhaps solves a very specific occurrence of this problem, but maybe not in the general case.

The big downside I can see to wiring up during `firstUpdated` is that if the host is ever disconnected from the DOM, the observer gets wiped out and never reconnected, which doesn't seem ideal. I'm personally not _super_ sure about the impact of wiring it up on the host. I don't have a solid enough understanding of the order of operations there.

dbatiste: I think there's no harm in doing this check.

Still seems really strange, since I'm not sure how it would "reconnect" later unless the element was removed and then re-added to the DOM. Maybe that's what is happening in your case, with all the fang-dangled toolbar chomping/fullscreen logic.

Yeah, when we wire up in `firstUpdated`, we need to make sure that we re-wire/un-wire in the connected/disconnected callbacks.

As for the `host`, in contrast to the elements created as a result of `render()`, it exists right away. But I would be pretty nervous about changing the element that we're observing, at least in this case due to the complexity of hierarchical-view and limitations with the old `ResizeObserver` polyfill.

dbatiste (approved):

components/hierarchical-view/hierarchical-view-mixin.js
grant-cleary: Yeah, I think that's very likely true. Because of the way we construct the toolbar, there's a lot of churn in those elements. It's entirely possible that this incompatibility _only_ exists in this case for the same reason; I haven't tried mixing these two mixins in other conditions.

Right, yeah, that's a fair point. Observing the host seems dangerous. Wiring up on `firstUpdated` feels safer, and I'm willing to follow up with that if you'd prefer. Not super sure how likely that would be to break other stuff though. 🤔

PR #4394, opened 2024-01-29
fix: Add skeleton support for inline-help components + update tests

saifaldin14:
Add skeleton support for inline-help components as well as update vdiff tests for existing components


components/inputs/input-inline-help-mixin.js
```diff
@@ -37,7 +37,7 @@ export const InputInlineHelpMixin = superclass => class extends superclass {
 
 	_renderInlineHelp(id) {
 		return html`
-			<div id="${id}" class="d2l-body-small d2l-input-inline-help">
+			<div id="${id}" class="d2l-body-small d2l-input-inline-help d2l-skeletize">
```
margaree, line 40: This is working for the inputs because they already import `SkeletonMixin` and the `super.styles`. However if this was added to something that didn't then it wouldn't actually do anything. I believe we'd want to import that here in order to have everything needed for the code to work as expected with the code.

Other than this everything else LGTM.

margaree (approved):

PR #4396, opened 2024-01-29
feat: Add inline help support for input-date-range, input-date-time-range, input-date-time and input-time-range

saifaldin14:
Add inline help support for input-date-range, input-date-time-range, input-date-time and input-time-range with accessibility and vdiff tests included

input-date-range (normal):
Screenshot (44)

input-date-range (multiline):
Screenshot (45)

input-date-time (normal):
Screenshot (46)

input-date-time (multiline):
Screenshot (47)

input-date-time-range (normal):
Screenshot (48)

input-date-time-range (multiline):
Screenshot (49)

input-time-range (normal):
Screenshot (50)

input-time-range (multiline):
Screenshot (51)


dbatiste (commented):
There are a [couple errors](https://vdiff.d2l.dev/BrightspaceUI/core/3ec9b8a12feaf56cf6355e33fba3543e2e4cff28/2024-01-30_16-08-34/report/?file=components%2Finputs%2Ftest%2Finput-date-time-range.vdiff.js&test=d2l-input-date-time-range+%3E+inline-help) for the `d2l-input-date-time-range` visual-diffs that are unexpected. Both normal and the multiline got the `Cannot read properties of null (reading 'querySelectorAll')` error, so they could not produce screenshots.

Otherwise changes look good. I highly recommend that we test this solution with different screenreader + browser combinations.

components/inputs/test/golden/input-date/chromium/inline-help.png
dbatiste, line 1: Is this what we're expecting or this one?

saifaldin14: I don't understand how this looks different than the last one?

components/inputs/test/golden/input-checkbox/chromium/default-indeterminate-inline-help.png
dbatiste, line 1: Note that the visual-diff tests are kind of expensive. We probably could get away with not capturing inline-help for every permutation of the selected state.

components/inputs/test/golden/input-color/chromium/inline-help-focus.png
dbatiste, line 1: Is this what we're expecting for this one?

components/inputs/test/golden/input-checkbox/chromium/default-indeterminate-inline-help.png
saifaldin14: I see, in that case I'll get rid of a few

components/inputs/test/golden/input-date/chromium/inline-help-multiline.png
dbatiste, line 1: Are we expecting to see the tooltip in this screenshot?

components/inputs/test/golden/input-date/chromium/inline-help.png
dbatiste: We'd need to look at the report to see which pixels are different. But this is an inline-help visual-diff test, but the tooltip appears to be covering it, so it doesn't seem effective at capturing what we're trying to cover.

components/inputs/test/golden/input-checkbox/chromium/disabled-checked-inline-help.png
dbatiste, line 1: I think capturing inline-help + disabled would be a worthwhile case to capture. It is an interesting case because we do not expect the inline help to become "greyed out".

dbatiste (approved):

PR #4398, opened 2024-01-29
Add Button: List: Documentation [GAUD-330]

margaree:
The adds an example to Daylight on how to use the new `add-button` attribute on `d2l-list`. ~This should wait until after the 2 nesting-related PRs are merged, as there's some functionality that wont look quite right until then.~


components/list/README.md
```diff
@@ -163,6 +163,99 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
     - Image: max dimensions: `width: 216px` and `height: 120px` and has `20px margin` from the main content;
     - default break: `843px < x`  where `x` is the width of the component.
 
+### add-button Property
+
+The `add-button` attribute inserts the `d2l-button-add` component above and below each item in a list. Each nested list that wants to have this functionality will need to have this attribute on the nested `d2l-list` as well. Each list item needs to have its own `key` in order to be able to communicate where a new item should be positioned.
```
dbatiste, line 168:
```suggestion
The `add-button` attribute inserts the `d2l-button-add` component above and below each item in a list. Each nested list that wants to have this functionality will need to have this attribute on the nested `d2l-list` as well. Each list item needs to have its own `key` in order to communicate where a new item should be positioned.
```

components/list/README.md
```diff
@@ -163,6 +163,99 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
...
+
+The adding of a new item into the list is to be entirely handled by consumers through listening for the `d2l-list-add-button-click` event. A simple example scenario is below.
```
dbatiste, line 170:
```suggestion
Adding new items to the list is handled entirely by consumers by listening for the `d2l-list-add-button-click` event. A simple example scenario is below.
```

components/list/README.md
```diff
@@ -163,6 +163,99 @@ The `d2l-list` is the container to create a styled list of items using `d2l-list
...
+  import '@brightspace-ui/core/components/list/list-controls.js';
+  import '@brightspace-ui/core/components/list/list-item.js';
+  import '@brightspace-ui/core/components/list/list-item-content.js';
+  import '@brightspace-ui/core/components/selection/selection-action.js';
+  import { css, html, LitElement } from 'lit';
+  import { getUniqueId } from '@brightspace-ui/core/helpers/uniqueId.js';
+
+  class ListDemoAddButton extends LitElement {
+
+    _handleListAddButtonClick(e) {
```
dbatiste, line 185: Note: this would not pass linting since they are private methods and would normally be required to be placed after `render`. I'm guessing that these are placed first just to help draw attention to them in the docs...?

dbatiste (approved):
Just a couple minor wording suggestions.

components/list/README.md
margaree: Yup that's exactly it. I debated over that but figured it was worth it to put them first to call attention to them because the rest of the code doesn't matter as much. Open to the other way too though.

PR #4408, opened 2024-01-31
If d2l-dialog-confirm text value is not a string, just render it.

dbatiste:
This PR updates `d2l-dialog-confirm` to render the `text` property value "as-is" if the `split()` function is not present. 

The MVC Inline Dialog's secondary text can contain SML, [rendered as HTML here](https://github.com/Brightspace/lms/blob/7e119c1870e327cf9d3be2ac8a77720c8c14ecd5/lp/framework/web/D2L.LP.Web.UI/Desktop/MasterPages/Dialog/Inline/InlineDialogFactory.cs#L66). That HTML would get encoded as block of text, which isn't what we'd want.

A potential solution to this is to quietly allow the LMS MVC Inline Dialog JavaScript to set `d2l-dialog-confirm`'s `text` property to a DOM node. To allow that, we simply need to not assume the `text` property value is a string.

Background: the purpose of the `split` method is to convert a string containing line-breaks into paragraph elements. Whether intentional or not, the code would currently explode if a consumer tried to set the property to a DOM node.

As a separate option, we can comb through all the lang-terms to find the ones that contain SML, but I figured I would test the water with this first.

### DIFF
--- a/components/dialog/dialog-confirm.js
+++ b/components/dialog/dialog-confirm.js
@@ -77,7 +77,7 @@ class DialogConfirm extends DialogMixin(LitElement) {
 						<div><h2 id="${this._titleId}" class="d2l-heading-3">${this.titleText}</h2></div>
 					</div>` : null}
 				<div id="${this._textId}" class="d2l-dialog-content">
-					<div>${this.text ? this.text.split('\n').map(line => html`<p>${line}</p>`) : null}</div>
+					<div>${this?.text?.split ? this.text.split('\n').map(line => html`<p>${line}</p>`) : this.text}</div>
 				</div>
 				<div class="d2l-dialog-footer">
 					<slot name="footer" class="d2l-dialog-footer-slot"></slot>


components/dialog/dialog-confirm.js
```diff
@@ -77,7 +77,7 @@ class DialogConfirm extends DialogMixin(LitElement) {
 						<div><h2 id="${this._titleId}" class="d2l-heading-3">${this.titleText}</h2></div>
 					</div>` : null}
 				<div id="${this._textId}" class="d2l-dialog-content">
-					<div>${this.text ? this.text.split('\n').map(line => html`<p>${line}</p>`) : null}</div>
+					<div>${this?.text?.split ? this.text.split('\n').map(line => html`<p>${line}</p>`) : this.text}</div>
```
margaree, line 80: Previously when the check for `this.text` returned false we set the div contents to `null`. With this change we set it to `this.text` instead if there either is not `this.text` or there isn't `this.text.split`, which is potentially different. Are there any concerns around that?

Otherwise this makes sense to me.

dlockhart: Agree with Margaree here that this is definitely different behaviour. Another option:

```javascript
let content = nothing;
if (typeof this.text === 'string') {
  content = this.text.split('\n').map(line => html`<p>${line}</p>`);
} else if( this.text instanceof HTMLElement ){
  content = this.text;
}
```

But I still have some concerns -- will message in Slack.

dbatiste: Exactly yeah. Previously if a consumer tried to set the property to a DOM node, they would have gotten an error. With this change, they could do that, even though we don't really want them to. We'd effectively be making an exception to enable migrating our MVC inline dialog rendering while maintaining their original lang-terms.

dbatiste:
Closing this since we dealt with the HTML cases in the LMS.

PR #4431, opened 2024-02-02
Fix tags flake and focus issues

bearfriend:
[GAUD-300](https://desire2learn.atlassian.net/browse/GAUD-300): Fixer: proper solution for filter-tags focus issue

Fixes flake in `tag-list` vdiff tests and focus issues in `tag-list` and `filter-tags`

Largely reverts the quick fix in #4199

[GAUD-300]: https://desire2learn.atlassian.net/browse/GAUD-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/tag-list/test/tag-list.vdiff.js
+++ b/components/tag-list/test/tag-list.vdiff.js
@@ -1,7 +1,7 @@
 import '../tag-list.js';
 import '../tag-list-item.js';
 import './tag-list-item-mixin-consumer.js';
-import { clickElem, expect, fixture, focusElem, hoverAt, hoverElem, html, nextFrame, oneEvent, sendKeys, waitUntil } from '@brightspace-ui/testing';
+import { clickElem, expect, fixture, focusElem, hoverElem, html, nextFrame, oneEvent, sendKeys, waitUntil } from '@brightspace-ui/testing';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { nothing } from 'lit';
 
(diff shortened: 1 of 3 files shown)


dlockhart (approved):
Looks like it was a fun journey to get here!

components/tag-list/test/tag-list.vdiff.js
```diff
@@ -120,8 +116,10 @@ describe('tag-list', () => {
 		});
 
 		it('delete first item', async() => {
-			await sendKeys('press', 'Tab');
-			await sendKeys('press', 'Delete');
+			elem.focus();
+			sendKeys('press', 'Delete');
```
margaree, line 120: Is the `await` no longer needed for `sendKeys`? Otherwise LGTM!

margaree (approved):

components/tag-list/test/tag-list.vdiff.js
dlockhart: I think the idea is that you don't need to `await` it if you're expecting it to trigger another event that you can wait for. Otherwise there's the possibility that the event you're waiting for could fire while you're still waiting for `sendKeys` to return.

dlockhart, line 120: Could these 2 lines be combined into `sendKeysElem(elem, 'press', 'Delete')`?

PR #4437, opened 2024-02-05
fix: fix lopsided padding on the right

saifaldin14:


components/empty-state/demo/empty-state.html
```diff
@@ -43,6 +43,15 @@ <h5>Empty State Action Link</h5>
 				</template>
 			</d2l-demo-snippet>
 
+			<h5>Empty State Only Button</h5>
```
margaree, line 46:
```suggestion
			<h5>Empty State Only Link</h5>
```
Did you want to add one for the button too?

components/filter/filter.js
```diff
@@ -439,6 +445,7 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 		if (emptyState.actionText && emptyState.actionHref) {
 			emptyStateAction = html`
 				<d2l-empty-state-action-link
+					class="d2l-empty-state-action"
```
margaree, line 448: What did this end up being needed for?

margaree: Just looking at the test site and the alignment of the text and link looks off on the regular empty-state case (not just in filter). We'd probably like to fix this more generally within the empty-state css rather than just in filter

margaree (approved):
LGTM!

PR #4439, opened 2024-02-05
Align `.scss` files to stylelint rules

bearfriend:
I was curious how bad the `.scss` files might be since we don't/can't lint them with the current setup, so I ran them through the linter and fixed up what was fixable.

Spoiler, they weren't bad at all.


dbatiste (approved):

dlockhart (approved):
It'd be great to not need these one day eh.

PR #4442, opened 2024-02-07
Add-Button: List: Fix bug when child list nested

margaree:
The issues can be seen [here](https://live.d2l.dev/BrightspaceUI/core/components/list/demo/list-nested.html) in the "Nested - Selectable" demo, where the first nested list has `add-button` but the remaining do not:
1. If you select all the list, there's a missing border on the top of "Flow and Transport..."
2. Checked with design and we'd want the add button for adding a sibling to "Introductory Earth Sciences" to appear above "Flow and Transport..." whereas what it's currently doing is showing directly under "Introductory Earth Sciences", which is above the nested list and makes where it would get added a bit confusing.

In practice, this is likely a rare situation where only the nested list has the add-button, but it could happen, and we should straighten out the style logic for it.

To fix this, I moved the bottom add slot to be below nested, since that's where we'd actually want the add button. We still hide this add button if there is an add button for the nested list and show it if there's expand/collapse when collapsed. I then added a `_hideBottomBorder` property which calculates when we'd want to actually hide the bottom border because the `add-button` would appear.

Note that with the case of mixed `add-button` in a list (i.e., some nested and some not) it makes hiding the border behind not as straightforward.


components/list/list-item-mixin.js
```diff
@@ -659,6 +661,9 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			'd2l-list-item-color-inner': true,
 			'd2l-skeletize': this.color
 		};
+		const bottomBorderClasses = {
+			'hide-bottom-border': this._showAddButton && (!this._hasNestedList || this._hasNestedListAddButton)
```
margaree, line 665: I added a class for this rather than exposing another property. I can switch to that if desired since it would be more consistent with this particular code.

dbatiste (approved):

PR #4446, opened 2024-02-09
Add Button: List: Fix drag and drop border bug

margaree:
**Problem**:
Can be seen in the [drag and drop demo](https://live.d2l.dev/BrightspaceUI/core/components/list/demo/list-drag-and-drop.html) "Add Button" demo at the bottom of the page. Drag the two nested items out of the bottom nested list and note that the border isn't full length and there is a console error.
The issue was that it actually no longer had a nested list and was failing when checking for `add-button` attribute.


dbatiste (approved):
Cool! That turned out to be a simpler fix than what I'd imagined. 😄

PR #4452, opened 2024-02-13
Use proper stacking context creation in skeleton mixin

bearfriend:
I added this line a while back to create a stacking context and only just learned today that there is a [dedicated property](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) to do this.


dlockhart (approved):
And it has excellent browser support! Good thing IE11 is gone. 🎉

PR #4455, opened 2024-02-15
Expand more/less on focusin only if necessary

bearfriend:
https://d2l.slack.com/archives/C0PHG3QB0/p1707927916567479

[GAUDI-5063](https://desire2learn.atlassian.net/browse/GAUD-5063) more/less component conditional autoexpand

- Auto-expands `d2l-more-less` only if the focused element isn't already on screen
- Scrolls focused elements into view after expand and collapse transitions
- Fixes the auto-expand transition (animation) direction


components/more-less/demo/more-less.html
```diff
@@ -13,6 +13,16 @@
...
+						<h2>More-less with custom blur color</h2>
+
+						<d2l-demo-snippet>
+							<template>
+								<d2l-more-less blur-color="#f00">
+									<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. <a href="">Vestibulum</a> elementum venenatis arcu sit amet varius. Maecenas posuere magna arcu, quis maximus odio fringilla ac. Integer ligula lorem, faucibus sit amet cursus vel, pellentesque a justo. Aliquam urna metus, molestie at tempor eget, vestibulum a purus. Donec aliquet rutrum mi. Duis ornare congue tempor. Nullam sed massa fermentum, tincidunt leo eu, vestibulum orci. Sed ultrices est in lacus venenatis, posuere suscipit arcu scelerisque. In aliquam ipsum rhoncus, lobortis ligula ut, molestie orci. Proin scelerisque tempor posuere. Phasellus consequat, lorem quis hendrerit tempor, sem lectus sagittis nunc, in tristique dui arcu non arcu. Nunc aliquam nisi et sapien commodo lacinia. <a href="">Quisque</a> iaculis orci vel odio varius porta. Fusce tincidunt dolor enim, vitae sollicitudin purus suscipit eu.</p>
+								</d2l-more-less>
+							</template>
+						</d2l-demo-snippet>
+
```
bearfriend, line 25: This is just for testing at the top of the page/scrolling up. I'll remove it before merging.

components/more-less/more-less.js
```diff
@@ -234,13 +237,27 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
...
 			return;
 		}
 
-		this.__expand();
-		this.__autoExpanded = true;
+		const target = e.composedPath()[0];
+
+		if (isSafari) {
+			target.scrollIntoViewIfNeeded();
+		}
```
bearfriend, line 249: Safari doesn't immediately scroll the focused element into view within the component content div, so we force it to. This still seems better than doing a bunch of custom measurement to determine if the target is on screen etc.

components/more-less/more-less.js
```diff
@@ -234,13 +237,27 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
...
+
+		if (this.__content.scrollTop) {
+			this.__content.scrollTo({ top: 0, behavior: 'instant' });
```
bearfriend, line 252: We scroll the content div back to the top instantly. This fixes the expand transition when using the keyboard to focus. It currently jumps to pull the active element into view within the collapsed content area, then expands it, reversing the intended animation from revealing the bottom to revealing the top.

components/more-less/more-less.js
```diff
@@ -234,13 +237,27 @@ class MoreLess extends LocalizeCoreElement(LitElement) {
...
+			this.__expand();
+			if (reduceMotion) {
+				target.scrollIntoView({ behavior: 'instant' });
+			} else {
+				setTimeout(() => target.scrollIntoView({ behavior: 'smooth' }, 100));
+			}
```
bearfriend, line 258: ...then we can expand, then make sure the target is in view.


Amazingly, `smooth` scroll is still used with reduced motion, so we need to disable it manually. The delay ensures that the smooth scroll doesn't miscalculate as the expand is still happening, and roughly times them together for less jumpiness.

dlockhart (approved):
Always wary of UA-checks but in this case it seems necessary.

PR #4465, opened 2024-02-22
fix: Remove reference to BrightspaceUI/tabs

svanherk:
Removing the reference to https://github.com/BrightspaceUI/tabs. Also fixing the menu link, which doesn't work at all on the Daylight site.


dlockhart (approved):
Good catch!

PR #4466, opened 2024-02-22
Introduce HTML embeds framework and implement it in the html-block component

grant-cleary:
As implied in the title, this is a PoC and shouldn't be merged in this state. I'm just putting this in draft PR form so it's easier for us to collaborate on this and make sure we're all on the same page. :)

### DIFF
--- a/components/html-block/html-block.js
+++ b/components/html-block/html-block.js
@@ -1,12 +1,13 @@
 import '../colors/colors.js';
 import { codeStyles, createHtmlBlockRenderer as createCodeRenderer } from '../../helpers/prism.js';
-import { css, html, LitElement, unsafeCSS } from 'lit';
+import { css, html, LitElement, nothing, unsafeCSS } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { createHtmlBlockRenderer as createMathRenderer } from '../../helpers/mathjax.js';
 import { getFocusPseudoClass } from '../../helpers/focus.js';
 import { HtmlAttributeObserverController } from '../../controllers/attributeObserver/htmlAttributeObserverController.js';
-
+import { renderEmbeds } from '../../helpers/embeds.js';
 import { requestInstance } from '../../mixins/provider/provider-mixin.js';
+import { until } from 'lit/directives/until.js';
 
 export const htmlBlockContentStyles = css`
 	.d2l-html-block-rendered {
@@ -231,7 +232,9 @@ class HtmlBlock extends LitElement {
 		};
 
 		return html`
-			<div class="${classMap(renderContainerClasses)}"></div>
+			<div class="${classMap(renderContainerClasses)}">
+				${(this._embedsFeatureEnabled() && !this.noDeferredRendering) ? until(this._processEmbeds(), nothing) : nothing}
+			</div>
 			${this.noDeferredRendering ? html`<slot @slotchange="${this._handleSlotChange}"></slot>` : ''}
 		`;
 	}
@@ -270,9 +273,19 @@ class HtmlBlock extends LitElement {
 		return false;
 	}
 
+	_embedsFeatureEnabled() {
+		return window.D2L?.LP?.Web?.UI?.Flags.Flag('shield-7574-enable-embed-rendering-framework', true);
+	}
+
 	async _handleSlotChange(e) {
 		if (!e.target || !this.shadowRoot || !this.noDeferredRendering) return;
 		await this._renderInline(e.target);
+	}
+
+	async _processEmbeds() {
+		const htmlFragment = document.createRange().createContextualFragment(this.html);
+		await renderEmbeds(htmlFragment);
+		return htmlFragment;
 	}
 
 	async _processRenderers(elem) {
@@ -324,7 +337,7 @@ class HtmlBlock extends LitElement {
 
 	async _updateRenderContainer() {
 		const renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
-		renderContainer.innerHTML = this.html;
+		if (!this._embedsFeatureEnabled()) renderContainer.innerHTML = this.html;
 		await this._processRenderers(renderContainer);
 	}
 
--- a/helpers/embeds.js
+++ b/helpers/embeds.js
@@ -1,0 +1,102 @@
+import { html, nothing, render } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { map } from 'lit/directives/map.js';
+import { tryGetPluginByKey } from './plugins.js';
+
+const embedTypeAttributeName = 'data-d2l-embed-type';
+
+function tryGetEmbedRendererPlugin(embedType) {
+	const embedRendererPlugin = tryGetPluginByKey('d2l-html-embed-renderer', embedType);
+	if (!embedRendererPlugin) {
+		console.warn(`d2l-html-embed-renderer: Can't find plugin for ${embedType} embed type.`);
+	}
+
+	return embedRendererPlugin;
+}
+
+export async function createEmbedViewPlaceholder(embedType, props) {
+	const embedRendererPlugin = tryGetEmbedRendererPlugin(embedType);
+	if (!embedRendererPlugin) return;
+
+	const placeholderData = await embedRendererPlugin.getViewPlaceholderData(props);
+	const contents = placeholderData.contents
+		? map(Object.entries(placeholderData.contents), ([id, content]) => html`<template data-d2l-embed-template-id=${id}>${content}</template>`)
+		: nothing;
+
+	const properties = placeholderData.properties ? JSON.stringify(placeholderData.properties) : undefined;
+
+	return placeholderData.inline
+		? html`
+			<span data-d2l-embed-type="${embedType}" data-d2l-embed-props="${ifDefined(properties)}" style="${placeholderData.style}">
+				${contents}
+			</span>
+		` : html`
+			<div data-d2l-embed-type="${embedType}" data-d2l-embed-props="${ifDefined(properties)}" style="${placeholderData.style}">
+				${contents}
+			</div>
+		`;
+}
+
+export async function createEmbedEditorPlaceholder(embedType, props) {
+	const embedRendererPlugin = tryGetEmbedRendererPlugin(embedType);
+	if (!embedRendererPlugin) return;
+
+	const placeholderData = await embedRendererPlugin.getEditorPlaceholderData(props);
+	const text = placeholderData.text || nothing;
+
+	const properties = placeholderData.properties ? JSON.stringify(placeholderData.properties) : undefined;
+
+	return placeholderData.inline
+		? html`
+			<span
+				contentEditable="${ifDefined(placeholderData.contentEditable)}"
+				data-d2l-embed-type="${embedType}"
+				data-d2l-embed-props="${ifDefined(properties)}"
+				style="${placeholderData.style}">
+				${text}
+			</span>
+		` : html`
+			<div
+				contentEditable="${ifDefined(placeholderData.contentEditable)}"
+				data-d2l-embed-type="${embedType}"
+				data-d2l-embed-props="${ifDefined(properties)}"
+				style="${placeholderData.style}">
+				${text}
+			</div>
+		`;
+}
+
+export async function renderEmbeds(node) {
+	const placeholders = [...node.querySelectorAll(`div[${embedTypeAttributeName}], span[${embedTypeAttributeName}]`)];
+	if (placeholders.length === 0) return;
+
+	const processPlaceholder = async placeholder => {
+		const embedRendererPlugin = tryGetEmbedRendererPlugin(placeholder.dataset.d2lEmbedType);
+		if (!embedRendererPlugin) return;
+
+		const templates = [...placeholder.querySelectorAll('template[data-d2l-embed-template-id]')];
+		const processedTemplates = await Promise.all(templates.map(async template => {
+			const templateNode = template.content.cloneNode(true);
+			await renderEmbeds(templateNode);
+			return html`${templateNode}`;
+		}));
+
+		const processedTemplateContents = templates.reduce((acc, template, index) => {
+			acc[template.dataset.d2lEmbedTemplateId] = processedTemplates[index];
+			return acc;
+		}, {});
+
+		const props = JSON.parse(placeholder.dataset.d2lEmbedProps);
+		return embedRendererPlugin.renderView(processedTemplateContents, props);
+	};
+
+	const embeds = await Promise.all(placeholders.map(processPlaceholder));
+	placeholders.forEach((placeholder, index) => {
+		if (!embeds[index]) {
+			console.warn(`d2l-html-embed-renderer: Can't find valid embed for placeholder with ${placeholder.dataset.d2lEmbedType} embed type`);
+		}
+
+		render(embeds[index], placeholder.parentNode, { renderBefore: placeholder });
+		placeholder.remove();
+	});
+}
--- a/components/html-block/test/html-block.test.js
+++ b/components/html-block/test/html-block.test.js
@@ -128,8 +128,11 @@ describe('d2l-html-block', () => {
 		const htmlBlock = await fixture(emptyReplacementFixture);
 		// Wait a frame for rendering to finish, since we temporarily add elements to the DOM
 		requestAnimationFrame(() => {
-			expect(htmlBlock.shadowRoot.querySelector('.d2l-html-block-rendered').innerHTML)
-				.to.equal('');
+			const renderContainer = htmlBlock.shadowRoot.querySelector('.d2l-html-block-rendered');
+
+			// This will contain Lit's default template marker, but there should be no non-whitespace text and no element nodes.
+			expect(renderContainer.innerText.trim()).to.equal('');
+			expect(renderContainer.children).to.be.empty;
 		});
 	});
 


helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
...
+		this.#childNodes.push(childNode);
+	}
+
+	async renderNode(options) {
+		for await (const childNode of this.#childNodes) {
+			await childNode.renderNode(options);
+		}
+
+		if (this.#node.id === '3') {
+			console.log('wee');
```
grant-cleary, line 24: I was just using this for debugging. Probably expect commits to show up with debugging code still in it while this is a PoC. I'm not exactly being careful about making sure my code is up to snuff. 😆

helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
...
+
+	const renderNodes = new Map();
+	const rootNodes = [];
+	const constructRenderNode = elem => {
+		if (renderNodes.has(elem)) return;
+
+		const renderNode = new RenderNode(elem);
+		renderNodes.set(elem, renderNode);
+
+		const nearestEmbedAncestor = findComposedAncestor(elem, ancestor => ancestor !== elem && elems.includes(ancestor));
```
grant-cleary, line 67: Possibly overkill since we won't actually be dealing with web components up the chain, but it was a quick and dirty way to find ancestors.

We're basically making use of `querySelectorAll` returning in DOM order here. If we don't find an ancestor that's in our list of embeds, then we're dealing with a root node. If we do, then we know it must have been processed already (and thus must be in the list of `renderNodes`) because it must have come earlier in the node list.

helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
...
+export async function createEmbedPlaceholder(embedType, options) {
+	const embedRendererPlugin = tryGetPluginByKey('d2l-html-embed-renderer', embedType);
+	if (!embedRendererPlugin) return;
+
+	const placeholderData = await embedRendererPlugin.getPlaceholder(options);
+	/* NOT IMPLEMENTED */
+}
+
+export async function renderEmbeds(node, options) {
+	const elems = [...node.querySelectorAll('div[data-d2l-embed-type], span[data-d2l-embed-type]')];
```
grant-cleary, line 56: I can get rid of the array construction here. It was present because I was using other array functions that aren't also present on `NodeList`. Now I'm not. Since this list will always be static, no need to massage it into an array.

helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
+import { findComposedAncestor } from '../helpers/dom.js';
+import { html } from 'lit';
+import { tryGetPluginByKey } from '@brightspace-ui/core/helpers/plugins.js';
+
+class RenderNode {
```
grant-cleary, line 5: This is just a little helper to encapsulate the rendering logic and make sure we're tracking/following child relationships properly. It's probably not needed, and there may be cleaner ways to handle this, but for now I at least like that it makes debugging easier.

helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
...
+		if (!nearestEmbedAncestor) {
+			rootNodes.push(renderNode);
+			return;
+		}
+
+		renderNodes.get(nearestEmbedAncestor).addChildNode(renderNode);
+	};
+
+	elems.forEach(constructRenderNode);
+	rootNodes.forEach(async rootNode => await rootNode.renderNode(options));
```
grant-cleary, line 77: It should always be safe to let the root nodes process in parallel. Though I should change this to a `map` and wrap it in a `Promise.all` to make sure we know all renderers have run before the MathJax and Prism renderers kick in. Right now the `await` in the html-block code won't do much here.

helpers/embeds.js
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	addChildNode(childNode) {
+		this.#childNodes.push(childNode);
+	}
+
+	async renderNode(options) {
+		for await (const childNode of this.#childNodes) {
+			await childNode.renderNode(options);
+		}
```
grant-cleary, line 21: Ideally this shouldn't need to queue like this, since siblings should all be able to render "in parallel", but there's some timing wonkiness going on here with fetching the OAuth2 tokens that causes this promise not to resolve properly. It's a bit of a headache, so for now I'm just leaving it like this while I poke around.

helpers/embeds.js
grant-cleary: Oop, scratch that. Forgot that I'm still using `includes`.

helpers/embeds.js
```diff
@@ -0,0 +1,75 @@
...
+		});
+
+		let props = this.#node.dataset.d2lEmbedData;
+		try {
+			props = JSON.parse(this.#node.dataset.d2lEmbedData);
+		} catch (e) { /* empty */ }
+
+		const embed = await embedRendererPlugin.renderView(fragment, props || {}, options || {});
+		if (embed) {
+			render(embed, this.#node.parentNode);
```
grant-cleary, line 37: I just jammed this in as a quick "Would Lit work here!?" test, and yeah, this works fine. In practice, if we wanted to do this using Lit, I'm sure there's a more clever way we could manage this to avoid direct DOM manipulation, at least for the html-block. In Content I think this is always going to be our lot in life.

helpers/embeds.js
grant-cleary: Yeah, I'm a bit surprised that more array functions aren't implemented on `NodeList`. Maybe `includes` is a bit of a weird one, but no `map`!? Sad.

dbatiste: I'm not opposed to querying `div` and `span` like this, but we could perhaps just query `[data-d2l-embed-type]`. Is there a reason to limit it?

grant-cleary: So the idea (not currently implemented in this PR) is that we have a helper function that builds the div/span for them. That gives an option for the placeholder to be either inline or not, allows us to make some assumption when building those elements for the consumer, and also makes it clear that the placeholder shouldn't really be doing anything super complex.

helpers/embeds.js
dbatiste: Could this alternatively be:
```javascript
findComposedAncestor(elem, ancestor => ancestor !== elem && elem.hasAttribute('data-d2l-embed-type'));
```
The benefit being avoiding an array search for every element in the composed path.

helpers/embeds.js
```diff
@@ -0,0 +1,82 @@
...
+	const constructRenderNode = elem => {
+		if (renderNodes.has(elem)) return;
+
+		const renderNode = new RenderNode(elem);
+		renderNodes.set(elem, renderNode);
+
+		const isEmbedAncestor = ancestor => {
+			if (ancestor === elem || ancestor.nodeType !== Node.ELEMENT_NODE) return;
+			return ancestor.hasAttribute(embedTypeAttributeName);
+		};
```
grant-cleary, line 69: Required an extra check here to make sure `hasAttribute` is valid, so I split it out into its own helper.

helpers/embeds.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export async function renderEmbeds(node, options) {
+	const elems = node.querySelectorAll(`div[${embedTypeAttributeName}], span[${embedTypeAttributeName}]`);
+	if (elems.length === 0) return;
+
+	const renderNodes = new Map();
+	const rootNodes = [];
+	const constructRenderNode = elem => {
+		if (renderNodes.has(elem)) return;
```
dbatiste, line 61: Is there a scenario where this would ever be `true`? `renderNodes` is initialized in this method, so it seems like even if this were to run a second time, it would be impossible case. Is this just being extra careful?

helpers/embeds.js
```diff
@@ -0,0 +1,82 @@
...
+		const nearestEmbedAncestor = findComposedAncestor(elem, isEmbedAncestor);
+		if (!nearestEmbedAncestor) {
+			rootNodes.push(renderNode);
+			return;
+		}
+
+		renderNodes.get(nearestEmbedAncestor).addChildNode(renderNode);
+	};
+
+	elems.forEach(constructRenderNode);
```
dbatiste, line 80: Perhaps more of a stylistic thing, but it may be possible to write this in such a way that the helper method is more stateless, such that it doesn't mutate things outside itself:
```javascript
const getRenderNodeContext = elem => {
    ...
    const renderNode = new RenderNode(elem);
    ...
    return {
        ancestorElem: nearestEmbedAncestor,
        renderNode: renderNode
    };
};

elems.forEach(elem => {
    const context = getRenderNodeContext(elem);
    renderNodes.set(elem, context.renderNode);
    context.ancestorElem ? 
        renderNodes.get(context.ancestorElem).addChildNode(context.renderNode) : 
        rootNodes.push(context.renderNode);
});
```

In your current implementation, `renderNode` is always added to `renderNodes` (via `set()`), and then it would potentially also be added to a `renderNode`'s child array (via `addChildNode`). As such, it would potentially be referenced twice in the `renderNodes` "tree". Is that intentional?

I think the naming of things is kinda confusing right now. Whenever I read "node"/"Node", I think of an actual DOM node. Like `renderNode` sounds like a method that renders an actual DOM node. I think I would consider renaming to avoid that confusion. It would also be good to make it clearer that the elems are the embed placeholders - future readers trying to understand this code might need to read a bunch of the code to discern that they aren't the replacement elements. I'd probably also be intentional about "node" vs "elem", but that's less confusing. I don't want to bike shed on naming, but straightening that out might help.

helpers/embeds.js
grant-cleary: It's just being extra careful, yeah. This shouldn't be possible for the same reason this algorithm (hopefully) works in the first place; since we're iterating over a static NodeList that's guaranteed to be in DOM order, we can never get a duplicate.

helpers/embeds.js
grant-cleary: Yeah, so as written right now, this duplication exists because we're using the `renderNodes` map solely to build up the array in `rootNodes`. Basically, `rootNodes` ends up being an array of placeholders we need to process, each with children (with their own children, etc.). i.e. Those are the trees we care about. `renderNodes` is just a temporary map to make sure we don't try to process a placeholder multiple times, so if an element has, say, 10 children, we'll add it to the tree when we find the first child and then know to skip it for the next nine. It's possible there are cleaner ways to do this, but a map stuck out to me at first because of it's quick insertion/lookup time.

RE: the nomenclature and general structure, this is just something quick and dirty I threw together to prove out that this is possible, so I haven't put much thought into structure or where state is being tracked. I fully expect that most of this will be heavily rewritten to be cleaner when we have a firm idea that this (or something else) is the direction we want to take. Unfortunately that does mean a lot of repeated `elem` and `elems` names in various scopes. If it makes it easier to parse for now, I can change up the names to be more obvious, but I don't want to spend a ton of time thinking about it until we know for sure we're not going to swing in some wildly different direction, if that makes sense.

helpers/embeds.js
```diff
@@ -0,0 +1,82 @@
...
+			const temp = node.cloneNode(true);
+			fragment.appendChild(temp);
+		});
+
+		let props = this.#node.dataset.d2lEmbedData;
+		try {
+			props = JSON.parse(this.#node.dataset.d2lEmbedData);
+		} catch (e) { /* empty */ }
+
+		const embed = await embedRendererPlugin.renderView(fragment, props || {}, options || {});
```
dbatiste, line 37: Curious about the nested case with a named `slot`. For example, where would the `slot="header"` be done? I don't think it would be appropriate for a renderer to necessarily assume that the component it is rendering is being placed into a particular slot, since that's a property of the context in which it is being used. If the slot name is a "standard"/"reserved" property in `props`, maybe we could apply it to the element returned here. But would it be appropriate to expect the slot name there?

helpers/embeds.js
dbatiste: Interesting. What value does using Lit's `render` provide here? Just making the rendering `async`, with the hope all the replacements would pop in at the same time? Note that we're already directly/synchronously updating the DOM with the following `node.remove()` statement.

helpers/embeds.js
```diff
@@ -0,0 +1,82 @@
...
+	constructor(node) {
+		this.#node = node;
+		this.#childNodes = [];
+	}
+
+	addChildNode(childNode) {
+		this.#childNodes.push(childNode);
+	}
+
+	async renderNode(options) {
```
dbatiste, line 20: What might these `options` be?

helpers/embeds.js
grant-cleary: For Content files in particular, nothing really. I see it being a bigger advantage in the case where this gets used within the html-block, since then this code could return a template that the html-block could insert in its own `render` method. Of course a potential benefit on the plugin side of things is that more complex structures could be more easily constructed via Lit templates. Otherwise plugins would need to build out their own DOM elements, which I imagine could get complex and annoying fast for complex components.

helpers/embeds.js
grant-cleary: What we talked about doing is encoding that in the placeholder in some way, so the placeholder's data attribute could track whether its children should be slotted or not. So in that sense it becomes the parent component's responsibility to indicate whether its children should be slotted (and what slot name that should have) and providing that information so the framework can slap the slot name on the child element.

Exactly what that looks like is TBD I think. I haven't written any code for it, but because the user profile card uses a slotted button, I was thinking about making a second silly toy renderer for the button and playing around with what that looks like.

helpers/embeds.js
grant-cleary: I'm guessing in practice, nothing. Right now it's just in there because I needed the user profile renderer to (temporarily) know whether it's iframed or not for my PoC to be able to fetch OAuth2 tokens. In reality, the plan is to pull the token stuff out into its own package that can make this determination itself.

helpers/embeds.js
dbatiste: Sounds like a good thing to try.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+
+	const processPlaceholder = async placeholder => {
+		const embedRendererPlugin = tryGetPluginByKey('d2l-html-embed-renderer', placeholder.dataset.d2lEmbedType);
+		if (!embedRendererPlugin) return;
+
+		const templates = [...placeholder.querySelectorAll('template[data-d2l-embed-template-id]')];
+		const processedTemplates = await Promise.all(templates.map(async template => {
+			const templateNode = template.content.cloneNode(true);
+			await renderEmbeds(templateNode, options);
+			return html`${templateNode}`;
```
grant-cleary, line 43: While `templateNode` is a document fragment, Lit nonetheless seems to handle this completely fine. I put together a test with a template that included a text node, some vanilla HTML, and a couple placeholders, and they all rendered into the DOM correctly.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+
+		return embedRendererPlugin.renderView(processedTemplateContents, props, options || {});
+	};
+
+	const embeds = await Promise.all(placeholders.map(processPlaceholder));
+	placeholders.forEach((placeholder, index) => {
+		if (!embeds[index]) return;
+		render(embeds[index], placeholder.parentNode, { renderBefore: placeholder });
+		placeholder.remove();
+	});
```
grant-cleary, line 64: The way this is structured now, we're converting between Lit template literals and DOM nodes a fair bit here. I'm not really sure of a great way to avoid that, since we need the template literals to interface between the plugins and the framework, but we also need the raw DOM nodes to do things like extract the template contents and query for placeholders.

This works out of the box in a Content file because we're working purely with raw HTML nodes to begin with, but for the html-block use case where we're starting with a Lit template, this maybe gets more complicated. It's possible that we'll need to render the Lit template into a document fragment in order to pass it down through the framework, then do something like ``return html`${documentFragment}`;`` in order to extract it out again.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+		? html`
+			<span data-d2l-embed-type="${embedType}" data-d2l-embed-props="${ifDefined(attributes)}">
+				${contents}
+			</span>
+		` : html`
+			<div data-d2l-embed-type="${embedType}" data-d2l-embed-props="${ifDefined(attributes)}">
+				${contents}
+			</div>
+		`;
+}
```
grant-cleary, line 29: I haven't tested this placeholder code because it requires wiring stuff up in the HTML editor that's (frankly) too much of a pain for me to want to deal with right now. So conceptually this is what I want to do, but the code itself might be bad or broken.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+
+		let props = placeholder.dataset.d2lEmbedProps;
+		try {
+			props = JSON.parse(placeholder.dataset.d2lEmbedProps);
+		} catch (e) { /* empty */ }
+
+		return embedRendererPlugin.renderView(processedTemplateContents, props, options || {});
+	};
+
+	const embeds = await Promise.all(placeholders.map(processPlaceholder));
```
grant-cleary, line 59: We let placeholders in a given subtree render as they wish, but we wait for them all to finish before we start any DOM replacements. It would definitely be possible to swap in the elements as part of the `processPlaceholder` function, but I'm still of the opinion that it's best to keep any asyncness out of there. Consider that if this is at the top level (and thus we're actually replacing real DOM elements), the more frames we take to do this work, the more likely we are to have client scripts start running before we finish.

I would also be amenable to treating the top level as a special case, since any placeholders lower in the tree are rendering into cloned template document fragments and thus aren't attached to the DOM yet. I'm just not sure we gain much by doing so.

helpers/embeds.js
grant-cleary: I've now confirmed that this works as expected, at least for the profile card case.

components/html-block/html-block.js
```diff
@@ -231,7 +233,9 @@ class HtmlBlock extends LitElement {
 		};
 
 		return html`
-			<div class="${classMap(renderContainerClasses)}"></div>
+			<div class="${classMap(renderContainerClasses)}">
+				${!this.noDeferredRendering ? until(this._processEmbeds(), unsafeHTML(this.html)) : nothing}
```
grant-cleary, line 237: Because embed processing doesn't require us to be attached to the DOM (and in fact, depends on working on nodes that _aren't_), we can handle it as part of the render method.

components/html-block/html-block.js
```diff
@@ -275,6 +279,12 @@ class HtmlBlock extends LitElement {
 		await this._renderInline(e.target);
 	}
 
+	async _processEmbeds() {
+		const htmlFragment = document.createRange().createContextualFragment(this.html);
+		await renderEmbeds(htmlFragment);
+		return htmlFragment;
```
grant-cleary, line 285: Right now, `renderEmbeds` doesn't return anything; instead, it works on an existing DOM node. This makes it easy to adapt both to the full HTML document case and the html-block case, though that comes at the cost of relying on side effects in this case.

components/html-block/html-block.js
```diff
@@ -324,7 +334,6 @@ class HtmlBlock extends LitElement {
 
 	async _updateRenderContainer() {
 		const renderContainer = this.shadowRoot.querySelector('.d2l-html-block-rendered');
-		renderContainer.innerHTML = this.html;
 		await this._processRenderers(renderContainer);
```
grant-cleary, line 337: We still need to process MathJax and Prism updates post-render, so this needs to stay. I suspect this could introduce a race condition with the initial render, since these renderers will kick off as soon as `this.html` is rendered into the container (i.e. we won't wait for the `until` to resolve). So it may be necessary to make use of a promise here to ensure we don't try to process math or code snippets until our embeds have finished rendering.

grant-cleary:
Test failure is expected now that we're rendering the HTML with Lit and will have Lit markup in the html-block's shadowRoot even when it's otherwise empty. If we decide this is the direction we want to go, I'll just exclude comment nodes from the test's verification.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+export async function createEmbedPlaceholder(embedType, props) {
+	const embedRendererPlugin = tryGetPluginByKey('d2l-html-embed-renderer', embedType);
+	if (!embedRendererPlugin) return;
+
+	const placeholderData = await embedRendererPlugin.getPlaceholder(props);
+	const contents = placeholderData.contents
+		? map(Object.entries(placeholderData.contents), ([id, content]) => html`<template data-d2l-embed-template-id=${id}>${content}</template>`)
+		: nothing;
+
+	const attributes = placeholderData.attributes ? JSON.stringify(placeholderData.attributes) : undefined;
```
dlockhart, line 17: Nit: I think we should call this `properties` instead of `attributes`, since they're more inline with the concept of camelCase strongly-typed "properties" vs. kebab-cased everything-is-a-string "attributes".

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+		`;
+}
+
+export async function renderEmbeds(node, options) {
+	const placeholders = [...node.querySelectorAll(`div[${embedTypeAttributeName}], span[${embedTypeAttributeName}]`)];
+	if (placeholders.length === 0) return;
+
+	const processPlaceholder = async placeholder => {
+		const embedRendererPlugin = tryGetPluginByKey('d2l-html-embed-renderer', placeholder.dataset.d2lEmbedType);
+		if (!embedRendererPlugin) return;
```
dlockhart, line 37: There's a few of these `return` cases where we're _likely_ not expecting them to actually happen... and maybe want to be logging it. But we can defer that.

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
...
+
+		const processedTemplateContents = templates.reduce((acc, template, index) => {
+			acc[template.dataset.d2lEmbedTemplateId] = processedTemplates[index];
+			return acc;
+		}, {});
+
+		let props = placeholder.dataset.d2lEmbedProps;
+		try {
+			props = JSON.parse(placeholder.dataset.d2lEmbedProps);
+		} catch (e) { /* empty */ }
```
dlockhart, line 54: This also feels like at the very least we'd want to `console.warn` it... otherwise the dev may be scratching their head for a while.

We could also `throw` in some of these cases that we're really not expecting, which would have the added bonus of being logged by the JS logging stuff (outside of content).

helpers/embeds.js
grant-cleary: Good thought. I can change the verbiage here.

helpers/embeds.js
grant-cleary: Yeah, I was thinking the same thing, but I'm also not super certain what's generally considered ideal for logging in our front-end stuff (beyond our ESLint plugin being grumbly about console statements). 🤔

helpers/embeds.js
grant-cleary: Yeah, I wasn't 100% sure about this case in particular. It kind of depends on what we want to support, since the JSON parser is kind of picky. If we want to say "this must be JSON parse-able", then that seems fine to me, though it does force that on consumers (i.e. they couldn't set `data-d2l-embed-props="someStringProperty"`, for example).

helpers/embeds.js
dlockhart: We've been `throw`-ing more and more in our components when developers don't provide things like labels, etc. In the web components do it in a timeout to avoid render-blocking.

helpers/embeds.js
dlockhart: I'd be fine with a JSON-parse-able requirement. It kind of forces consumers to future-proof their data for a time when they might want more than 1 thing.

grant-cleary: Good point. In which case I think it's reasonable to just remove the try-catch and let the parser throw?

helpers/embeds.js
```diff
@@ -0,0 +1,65 @@
+import { html, nothing, render } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { map } from 'lit/directives/map.js';
+import { tryGetPluginByKey } from './plugins.js';
+
+const embedTypeAttributeName = 'data-d2l-embed-type';
+
+export async function createEmbedPlaceholder(embedType, props) {
```
dbatiste, line 8: Not to bike shed, but I am wondering is we want to replace "Placeholder" term with "Template" here. Whenever I think of the placeholder, I think if the thing that the html-editor displays in its design-view, which I expect is different, ya? Might help avoid confusion when we're talking about these things.

grant-cleary: I don't personally have strong opinions here, so I'm fine with whatever we think is ideal. I don't really think it's bikeshedding either, since this is going to be a public (at least to D2L) API. 😆

grant-cleary: Does that make pie filling a placeholder? It fills a place and it's temporary because it's destined to be eaten. 🤔

components/html-block/html-block.js
grant-cleary: This is for sure a behavioural change though, so it's plausible we'll want to feature flag this. I'm just not sure there's a fantastic way to do so. I don't think we've really flagged stuff directly in the html-block before.

helpers/embeds.js
dbatiste: At least for myself, it would be helpful to somehow differentiate. If it's a matter of calling it DesignViewPlaceholder in the editor context, and Placeholder here, that would be sufficient as along as we're are consistent.

grant-cleary: Before you commented, I did update it to `createEmbedViewPlaceholder` (with the implied  assumption that the editor one would be `createEmbedEditPlaceholder`), though I'm not fixed to that either. I do think that a distinction is necessary if we're switching to having the edit one be centralized as well (which I think is where we want to go). I'm content with whatever naming we all feel like, though I wonder how prevalent the "Design view" parlance is w.r.t. the editor.

grant-cleary: I know "design mode" or "design view" is something we've used before to draw that distinction, but we've never really had need to expose it externally. The C+ stuff still remains the only set of plugins that weren't written by either you or me, so I suspect there's no prevalent terminology for this stuff. 😆

helpers/embeds.js
dbatiste: Yeah, the `html` template literal can handle a DOM node. In fact, I _think_ it handles arrays of nodes `${[nodes]}` just fine too.

Is it necessary to wrap `templateNode` in a Lit template here? It appears that it will result in `processedTemplateContents` being an array of Lit templates, which gets passed to `renderView(...)`. Does `renderView` need those to be Lit templates?

I checked out your user profile plugin to see how it would be used, but it doesn't appear to make use of this. I guess it's only for the nested case?

grant-cleary: Yeah, I deliberately want plugins to _only_ work with Lit templates. So they get Lit templates passed in via the plugin when there's nested content. The problem with passing raw DOM nodes around is that plugins can start walking the DOM to edit things they shouldn't.

While that can technically still happening (nothing stops a plugin from rendering these into a temporary DOM node and having at it), I'm hopeful providing what's effectively an opaque string will be a hint that they _shouldn't_ do this.

helpers/embeds.js
```diff
@@ -0,0 +1,102 @@
...
+	const properties = placeholderData.properties ? JSON.stringify(placeholderData.properties) : undefined;
+
+	return placeholderData.inline
+		? html`
+			<span
+				contentEditable="${ifDefined(placeholderData.contentEditable)}"
+				data-d2l-embed-type="${embedType}"
+				data-d2l-embed-props="${ifDefined(properties)}"
+				style="${placeholderData.style}">
+				${text}
```
grant-cleary, line 56: This supports the current @mentions case, but I didn't want to prototype anything further. It doesn't necessarily matter if we get this one right immediately, since the in-editor placeholder will never get saved anywhere.

helpers/embeds.js
```diff
@@ -0,0 +1,102 @@
...
+	return embedRendererPlugin;
+}
+
+export async function createEmbedViewPlaceholder(embedType, props) {
+	const embedRendererPlugin = tryGetEmbedRendererPlugin(embedType);
+	if (!embedRendererPlugin) return;
+
+	const placeholderData = await embedRendererPlugin.getViewPlaceholderData(props);
+	const contents = placeholderData.contents
+		? map(Object.entries(placeholderData.contents), ([id, content]) => html`<template data-d2l-embed-template-id=${id}>${content}</template>`)
```
grant-cleary, line 23: I think this is safe to do in the view placeholder because (1) we're templating the content anyways, and (2) if there's user-authored HTML in here, we're expecting it to be pre-filtered. This doesn't have a means to support more complex structures in terms of the placeholders, but honestly, I'm not sure it's worth spending too much time exploring that before we need it. We can always add something like that later.

components/html-block/html-block.js
grant-cleary: True. Probably still fine though. If we do some kind of context thing it will still only work inside the LMS. So if you're fine with that in `core`, I can do that for this piece.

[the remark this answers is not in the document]
grant-cleary: That's right. The alternative I guess would be to render nothing. I'm not entirely sure which is preferable. My gut feeling was that rendering the actual HTML would be less jittery than just rendering nothing, but I don't have strong feelings on that. Rendering nothing would probably look a bit more like it just took a longer time to render in the first place.

helpers/embeds.js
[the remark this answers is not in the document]
grant-cleary: That's right, yeah. And probably inside the HTML editor's iframe.

components/html-block/html-block.js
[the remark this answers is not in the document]
dlockhart: We may have to test it in practice to find out. I'm worried about "slow" things like images, videos, etc. that start downloading, then get removed, re-added and download again.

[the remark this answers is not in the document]
grant-cleary: That's a good point. To make things worse, this is already going to happen once because MathJax needs to remove everything from the DOM and re-add it here: https://github.com/BrightspaceUI/core/blob/main/helpers/mathjax.js#L69

So minimizing how often we need to completely re-render content is definitely ideal.

grant-cleary:
I'm not going to merge any vdiff changes in here until we're happy with this and I can do a rebase.

components/html-block/html-block.js
```diff
@@ -270,11 +273,21 @@ class HtmlBlock extends LitElement {
 		return false;
 	}
 
+	_embedsFeatureEnabled() {
+		return window.D2L?.LP?.Web?.UI?.Flags.Flag('SHIELD-7574-enable-embed-rendering-framework', true);
```
grant-cleary, line 277: This flag doesn't exist yet, but if we're okay with flagging this way, I'll create it.

dlockhart: Yeah I'm OK with this as I assume it'll be short-lived anyway.

grant-cleary: Sounds good. We generally let these sorts of flags soak for ~3 releases before getting rid of them, so it certainly won't be in here for long.

dlockhart (approved):

PR #4469, opened 2024-02-23
Dialog: Add "critical" property [GAUD-6059]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/browse/GAUD-6059)

Adds `critical` property to `d2l-dialog` and `d2l-dialog-confirm`


components/dialog/dialog-confirm.js
```diff
@@ -66,12 +73,19 @@ class DialogConfirm extends DialogMixin(LitElement) {
 		`];
 	}
 
-	render() {
-		if (!this._titleId) this._titleId = getUniqueId();
-		if (!this._textId) this._textId = getUniqueId();
```
margaree, line 71: I think we generally put these in the `constructor` so I moved them when adding `offscreenId`, but let me know if there was a reason they were in `render` instead

components/dialog/dialog-confirm.js
```diff
@@ -66,12 +73,19 @@ class DialogConfirm extends DialogMixin(LitElement) {
...
+		super();
+		this.critical = false;
+		this._offscreenId = getUniqueId();
+		this._textId = getUniqueId();
+		this._titleId = getUniqueId();
+	}
 
+	render() {
 		const inner = html`
+			${this.critical ? html`<d2l-offscreen id="${this._offscreenId}">${this.localize('components.dialog.critical')}</d2l-offscreen>` : nothing}
```
margaree, line 86: The new code for `critical` is very duplicated between `d2l-dialog` and `d2l-dialog-confirm`. It all could be in the mixin but I left it out since we don't want this in the fullscreen dialog. Should I put it there anyways and somehow block usage of `critical` in fullscreen?

components/dialog/dialog-confirm.js
```diff
@@ -85,13 +99,14 @@ class DialogConfirm extends DialogMixin(LitElement) {
 			</div>`;
 
 		const labelId = (this.titleText && this.text) ? this._titleId : this._textId;
+		const fullLabelId = this.critical ? `${this._offscreenId} ${labelId}` : labelId;
```
dlockhart, line 102: Instead of hooking into `aria-labelledby` like this, could the `<d2l-offscreen>` stuff literally go _inside_ the `<h2>` above?

components/dialog/dialog-confirm.js
dlockhart: Let's see where everything ends up, but we could also just create a little `renderCritical` helper in one of the files that can be exported and imported into the places that need it.

components/dialog/dialog-styles.js
```diff
@@ -94,13 +94,27 @@ export const dialogStyles = css`
 		height: 100%;
 	}
 
+	:host([critical]) .d2l-dialog-highlight {
+		background-color: var(--d2l-color-cinnabar);
+		border-radius: 0.4rem 0.4rem 0 0;
```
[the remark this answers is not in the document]
bearfriend: Could we use `.d2l-dialog-inner::before`?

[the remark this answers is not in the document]
margaree: Ooo this might work

components/dialog/dialog-confirm.js
margaree: It could, but the `h2` is only rendered if there's a title. We could do some conditionals where if there's a title then it goes there and if not then it goes with the text, that just gets a bit more complex but it's do-able.

components/dialog/dialog-styles.js
```diff
@@ -94,13 +94,27 @@ export const dialogStyles = css`
...
+		height: 0.4rem;
+		margin-inline-start: -1px;
+		margin-top: -1px;
+		position: absolute;
+		top: 0;
+		width: calc(100% + 1px);
```
bearfriend, line 105: Is this a little less awkward?
```suggestion
		position: absolute;
		inset: -1px;
```

margaree: Definitely!

components/dialog/dialog-confirm.js
dbatiste: Btw, it's also valid to reference `visibility: hidden`/`display: none` elements when using `aria-labelledby` so this is valid and avoids the extra custom element. 
```html
<div id="${this._criticalLabelId}" hidden>${this.localize('components.dialog.critical')}</div>
```

components/dialog/dialog-styles.js
```diff
@@ -94,6 +94,14 @@ export const dialogStyles = css`
 		height: 100%;
 	}
 
+	:host([critical]) .d2l-dialog-header {
+		margin-inline-start: -1px;
+		margin-top: -1px;
+		border-radius: 0.4rem;
+		border-top: 0.4rem solid var(--d2l-color-cinnabar);
+		padding-top: 1rem;
```
dlockhart, line 102: I think we should be consistent with our use of logical properties?

```suggestion
		margin-block-start: -1px;
		margin-inline-start: -1px;
		border-block-start: 0.4rem solid var(--d2l-color-cinnabar);
		border-start-start-radius: 0.4rem;
		border-start-end-radius: 0.4rem;
		padding-block-start: 1rem;
```

bearfriend: I don't think this all works because it's moving the header text around and is still 1px short on the right.

bearfriend: Hah, obviously it works. I just mean it has other effects.

bearfriend: I think this is what we want, including all consistent logical properties?
```suggestion
		margin-inline: -1px;
		margin-block: -1px 1px;
		border-radius: 0.4rem;
		border-block-start: 0.4rem solid var(--d2l-color-cinnabar);
		border-start-start-radius: 0.4rem;
		border-start-end-radius: 0.4rem;
		padding-inline: 31px;
```

margaree: That looks great, thanks!

bearfriend: That still wasn't right. Edited.

components/dialog/dialog-styles.js
```diff
@@ -94,6 +94,15 @@ export const dialogStyles = css`
 		height: 100%;
 	}
 
+	:host([critical]) .d2l-dialog-header {
+		border-block-start: 0.4rem solid var(--d2l-color-cinnabar);
+		border-start-start-radius: 0.4rem;
+		border-start-end-radius: 0.4rem;
+		margin-block: -1px 0;
+		margin-inline: -1px;
+		padding: 1rem 31px 23px 31px;
```
margaree, line 103: I combined all the padding and put the 23px up here as well. I felt it was a bit more readable with the disadvantage that if the 23px changes in the regular `.d2l-dialog-header` css then it would have to get changed here as well.

bearfriend (approved):

dlockhart (approved):
Looks great!

dbatiste (approved):

PR #4475, opened 2024-02-27
Fix bad tooltip positioning when used in other components

bearfriend:
When we use `d2l-tooltip` inside other components, the tooltip can be positioned incorrectly if we don't give the target element a non-`static` `position`.

Reported [here](https://grow.d2l.com/d2l/le/enhancedSequenceViewer/6942?url=https%3A%2F%2Fe16bd763-9b2f-4f8d-b9da-2c008c2b19ff.sequences.api.brightspace.com%2F6942%2Factivity%2F11371%3FfilterOnDatesAndDepth%3D1).

### DIFF
--- a/components/button/button.js
+++ b/components/button/button.js
@@ -42,6 +42,7 @@ class Button extends ButtonMixin(LitElement) {
 				button {
 					font-family: inherit;
 					padding: 0.55rem 1.5rem;
+					position: relative;
 					width: 100%;
 				}
 
--- a/components/count-badge/count-badge-mixin.js
+++ b/components/count-badge/count-badge-mixin.js
@@ -198,7 +198,8 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
 		if (!numberStyles || numberStyles.visibility !== 'hidden') {
 			numberStyles = {
 				...numberStyles,
-				visibility: hideNumber ? 'hidden' : 'visible'
+				visibility: hideNumber ? 'hidden' : 'visible',
+				zIndex: 1
 			};
 		}
 		return html`
@@ -213,7 +214,8 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
 		<div id="${this._labelId}"
 			aria-label="${this.text}"
 			aria-atomic="true"
-			aria-live="${this.announceChanges ? 'polite' : 'off'}">
+			aria-live="${this.announceChanges ? 'polite' : 'off'}"
+			style="isolation: isolate">
 			${innerHtml}
 		</div>
 		${this.hasTooltip && !this.skeleton ? html`<d2l-tooltip class="vdiff-target" for="${badgeId}" for-type="label">${this.text}</d2l-tooltip>` : null}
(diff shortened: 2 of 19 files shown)


components/form/test/form-helper.test.js
```diff
@@ -142,7 +142,7 @@ describe('form-helper', () => {
 
 		const implicitLabelWithTooltipFixture = html`
 			<label>Do you like peas?
-				<input id='target' type="checkbox" name="peas">
+				<input id='target' type="checkbox" name="peas" style="position: relative;">
```
bearfriend, line 145: The usages in test files like this don't _need_ to have this added, but I was just going through and silencing all the new warnings.

components/tooltip/tooltip.js
```diff
@@ -725,15 +725,23 @@ class Tooltip extends RtlMixin(LitElement) {
...
 		} else {
+			console.warn('<d2l-tooltip>: missing required attribute "for"');
 			const parentNode = this.parentNode;
 			target = parentNode.nodeType === Node.DOCUMENT_FRAGMENT_NODE ? ownerRoot.host : parentNode;
 		}
+
+		if (window.getComputedStyle(target).position === 'static') {
+			const host = this.getRootNode().host?.tagName.toLowerCase() ?? 'body';
+			const selectorText = targetSelector ? `(${targetSelector}) ` : '';
+			console.warn(`In <${host}>, <d2l-tooltip> target element ${selectorText}does not have a CSS position. The tooltip may not position itself correctly on the page.`);
```
bearfriend, line 742: I added the `host` here mostly for visibility while fixing tests. I don't think it's necessary to always include, though.

components/button/button.js
```diff
@@ -42,6 +42,7 @@ class Button extends ButtonMixin(LitElement) {
 				button {
 					font-family: inherit;
 					padding: 0.55rem 1.5rem;
+					position: relative;
```
dbatiste, line 45: I have to admit, requiring consumers to do this doesn't feel great. If we change the tooltip positioning logic in the future, these `position: relative` styles may become unnecessary, and it will be a little risky to remove them if we aren't certain they aren't needed for some other purpose as well.

I'm not really clear on what it is in the tooltip positioning logic that requires the target to be positioned, but I wonder if there is a fix we can make to that logic that doesn't require consumers to do this.

components/count-badge/count-badge-mixin.js
```diff
@@ -198,7 +198,8 @@ export const CountBadgeMixin = superclass => class extends LocalizeCoreElement(S
 		if (!numberStyles || numberStyles.visibility !== 'hidden') {
 			numberStyles = {
 				...numberStyles,
-				visibility: hideNumber ? 'hidden' : 'visible'
+				visibility: hideNumber ? 'hidden' : 'visible',
+				zIndex: 1
```
dbatiste, line 202: Is this `z-index` and the `isolation` below to address the same issue, or something different?

components/inputs/input-text.js
```diff
@@ -197,6 +197,10 @@ class InputText extends InputInlineHelpMixin(PropertyRequiredMixin(FocusMixin(La
 				:host([value-align="end"]) {
 					--d2l-input-text-align: end;
 				}
+				#d2l-input-text-container {
+					display: inline;
```
dbatiste, line 201: I'm assuming the `display: inline` became necessary as a result of changing the position. Ya? Though I'm not sure why.

PR #4481, opened 2024-02-28
fix: NIM1872 - add vdiff tests for cases where table-wrapper has stickyHeadersScrollWrapper enabled

derekmoran:
Add vdiff tests for [d2l-table-wrapper](https://daylight.d2l.dev/components/table/#d2l-table-wrapper) for the case of its _stickyHeadersScrollWrapper_ property being enabled.

The [current state](https://d2l.slack.com/archives/C0PHG3QB0/p1709128112399439) of some these are not ideal. That said I think it would be useful to have them here, and I _**may**_ be attempting to improve some of these in the near future in which case it will be invaluable to see what affect such attempts are having.


components/table/test/table.vdiff.js
```diff
@@ -424,181 +425,183 @@ describe('table', () => {
...
-						<td>Cell 2-C</td>
-					</tr>
-				`;
-				[
-					{
-						name: 'one-row-thead',
-						template: html`
-								<thead>${createHeaderRow({ trClass: 'top' })}</thead>
+			[true, false].forEach((hasStickyHeadersScrollWrapper) => {
+				describe(`sticky${hasStickyHeadersScrollWrapper ? '-scroll-wrapper' : ''}`, () => {
```
eKoopmans, line 429: So this reproduces all the sticky tests in two forms now, with and without the scroll-wrapper. I'm not sure whether we need all of these, or whether we want more targeted vdiffs for the specific behaviour of `sticky-headers-scroll-wrapper` - but I'll leave that for Gaudi to decide.

derekmoran: hah yeah, there were already quite lot of sticky cases with _default vs light_ and _ltr vs rtl_.  I was concerned that any one those might behave uniquely, so I added the _scroll-wrapper_ scenario for that entire suite. If we're thinking that is overkill I can certainly scale it back a bit 😄

dbatiste: Yeah that was my first reaction. The visual-diffs certainly aren't "free" so we don't necessarily always capture every permutation of every property, state, etc. If they are worthwhile, then it's fine, but it's also a lot. 

It would be nice to have a callout in the ones that are showing broken behaviour.

derekmoran: Thank you both for the great feedback!

I do think there will be value in having both _ltr and rtl,_ but _default vs light_ don't seem as likely to make a difference 🤔  I'll reflect further and see what I can cut down on. 

Also really like the idea of calling out the ones that seem busted. I'll update this thread when I've got something that takes these considerations into account ...

PR #4487, opened 2024-02-29
fix: added description to get rid of console log error

saifaldin14:


components/empty-state/demo/empty-state.html
```diff
@@ -46,7 +46,7 @@ <h5>Empty State Action Link</h5>
 			<h5>Empty State Only Button</h5>
 			<d2l-demo-snippet>
 				<template>
-					<d2l-empty-state-simple>
+					<d2l-empty-state-simple description="Temp">
```
margaree, line 49: We tend to have our examples use semi-realistic text (e.g., the descriptions above). Can these descriptions also contain more realistic text? Otherwise LGTM!

margaree (approved):

PR #4490, opened 2024-02-29
fix: Add Button in List: Grid mode tooltip

margaree:
When in grid mode, there should be an instructional tooltip displayed when the user first tabs in. This did not work as expected with button-add.

Worth noting is that button-add does normally have its own tooltip that says "Add Item" by default. This one does seem like it should take priority for this first focus. I also think the position start is okay here because otherwise it seems too much related specifically to the add button itself if it's centred but adding in @geurts to see if there's any thoughts on that.

Screenshot 2024-02-29 at 3 44 19 PM

### DIFF
--- a/components/list/list-item-mixin.js
+++ b/components/list/list-item-mixin.js
@@ -434,6 +434,7 @@ export const ListItemMixin = superclass => class extends composeMixins(
 		this.first = false;
 		this.noPrimaryAction = false;
 		this.paddingType = 'normal';
+		this._addButtonTopId = getUniqueId();
 		this._contentId = getUniqueId();
 		this._displayKeyboardTooltip = false;
 		this._hasColorSlot = false;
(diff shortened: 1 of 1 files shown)


dbatiste:
Yep, definitely interested in @geurts thoughts here, but otherwise LGTM!

dbatiste (approved):
LGTM from my perspective.

saifaldin14 (approved):

geurts:
Sorry I missed this thread, but I've tested this and it works well. This is in keeping with our established compromise where we allow the keyboard instructions tooltip to supplant the tooltip that would normally appear (if there was one). 

**Note**: So far we have allowed this approach since the keyboard tooltip only appears when first focused, and users can still get the original tooltip content upon subsequent focusing. We may revisit this one day to find a better solution, but the fix in this PR aligns to our established best practice for now.

PR #4492, opened 2024-03-01
Fix bad tooltip positioning

bearfriend:
Attempt 2

If the `offsetParent` itself _is_ the `boundingContainer`, we need to account for scroll when we position the tooltip.

The first vdiff run will produce the current broken behavior. Then I'll uncomment the fix.


components/tooltip/tooltip.js
```diff
@@ -582,18 +582,17 @@ class Tooltip extends RtlMixin(LitElement) {
 		this._openDir = space.dir;
 
 		// Compute the x and y position of the tooltip relative to its target
-		let parentTop;
-		let parentLeft;
+		let offsetTop, offsetLeft;
```
bearfriend, line 585: `parent` was just confusing me here, but if it makes more sense to others I can revert this.

components/tooltip/tooltip.js
```diff
@@ -729,8 +728,9 @@ class Tooltip extends RtlMixin(LitElement) {
 		if (this.for) {
 			const targetSelector = `#${cssEscape(this.for)}`;
 			target = ownerRoot.querySelector(targetSelector);
-			target = target || (ownerRoot && ownerRoot.host && ownerRoot.host.querySelector(targetSelector));
+			target = target || ownerRoot?.host?.querySelector(targetSelector);
 		} else {
+			console.warn('<d2l-tooltip>: missing required attribute "for"');
```
bearfriend, line 733: This seems like a good thing to me, but willing to revert this too if there's disagreement.

components/tooltip/test/golden/tooltip/chromium/scrolling-boundingcontainer:-50,-100.png
bearfriend, line 1: This is a slightly larger, but also probably extremely rare, issue that I'll leave for another time. I think it has to do with how the available space is calculated before/without looking at the required shift.

components/tooltip/tooltip.js
dbatiste: I don't feel strongly one way or another. It's nice to avoid the inconsistency, but on the other hand, it's also nice not to force consumers to create a non-interactive wrapper element just for the `for`.

dbatiste (approved):

components/tooltip/tooltip.js
bearfriend: Ok, I'm going to merge this. If it proves more annoying than helpful we can remove it and indicate `for` is required _if_ the target isn't the parent/host.

PR #4498, opened 2024-03-04
Fix various tag focus issues

bearfriend:
For `d2l-tag-list` _and_ `d2l-filter-tags`:
- Activating `show more` should move focus to the first previously hidden tag
- Activating `show less` should move focus to `show more`
- Deleting first tag should focuses the next tag (now first)
- Pressing tab from any tag should move focus to the first non-tag item (more, less, clear, next focusable element)
- Pressing shift + tab from any tag should move focus to the first focusable element immediately preceding the list

While testing #4431, it was discovered that focus could be lost before being able to read the active element to restore later. There were also a few other focus issues that should all be fixed here.

[GAUD-300](https://desire2learn.atlassian.net/browse/GAUD-300) proper solution for filter-tags focus issue

[GAUD-300]: https://desire2learn.atlassian.net/browse/GAUD-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/tag-list/tag-list-item-mixin.js
```diff
@@ -172,7 +172,7 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			}
 
 			await this.updateComplete;
-
+			this.setAttribute('tabindex', '-1');
```
dbatiste, line 175: It looks like the `tabindex` is managed [here in d2l-tag-list](https://github.com/BrightspaceUI/core/blob/main/components/tag-list/tag-list.js#L399). Does this new assignment cover a case that wasn't being handled?

components/tag-list/tag-list.js
```diff
@@ -219,9 +220,10 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 					<d2l-button-subtle aria-hidden="true" slim text="${this.localize('components.tag-list.num-hidden', { count: '##' })}" class="d2l-tag-list-hidden-button"></d2l-button-subtle>
 					${this.arrowKeysContainer(list)}
 				</div>
-			`, this.localize('components.tag-list.interactive-label', { count: this._items ? this._items.length : 0 }),
```
dbatiste, line 222: Looks like no changes to this... just formatting, ya?

components/tag-list/tag-list.js
```diff
@@ -350,9 +352,11 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 		const itemIndex = children.indexOf(rootTarget);
 
 		if (children.length <= 1) return;
-		const focusableElem = children[itemIndex - 1] || children[itemIndex + 1];
```
dbatiste, line 353: With the previous implementation, it looks like it was assuming that the consumer would not have removed the item yet, and then trying to either focus the previous item if it exists, otherwise the next item. I gather that was not working as intended. Was that assumption not valid, or was something else going wrong?

With the new implementation, it seems like the `children[itemIndex] !== e.target` is meant to handle both the case where the item has been removed by the time the callback runs, and the case where it hasn't been removed yet. Am I understanding that correctly?

components/tag-list/tag-list.js
```diff
@@ -406,7 +410,22 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
...
+		if (this._refocus?.classList.contains('d2l-tag-list-button')) {
+			this._refocus = this.shadowRoot.querySelector('.d2l-tag-list-button');
+		}
+		await this._refocus?.updateComplete;
+		(this._refocus || refocus).focus?.();
+		this._refocus = null;
+	}
+
+	_handleSlotFocusIn() {
+		this._items[0].setAttribute('tabindex', '-1');
```
dbatiste, line 422: If the `slotchange` event runs while focus is inside the slot, this block of code (above) would undo this assignment. It looks like the slot's `focusin`/`focusout` handlers run when an item is removed via keyboard. As far as I know, there wouldn't be any other way for the `slotchange` event to be triggered while focus is already inside the slot, so maybe we're covered. It's not like a tag would be added while focus is inside... that would require the user to something with focus outside the slot.

```javascript
this._items.forEach((item, index) => {
	item.setAttribute('tabIndex', index === 0 ? 0 : -1);
});
```

components/tag-list/tag-list.js
bearfriend: Yeah that's exactly right. In `tag-list` the item is not yet removed by the time it gets to this method, but in `filter-tags` it is. However, I realize now that this is simply because `children` is outdated. I think it'd be simpler to move everything inside the `setTimeout` and then do:
```
const focusableElem = children[itemIndex - 1] || children[0];
```

bearfriend: Yeah that's exactly right. In `tag-list` the item is not removed until after the timeout, but in `filter-tags` it's removed before the method is called.

bearfriend: lol, I keep writing comments and deleting them because I keep finding more weird stuff. It looks like this might be down to a lit DOM updating issue between `tag-list` and `filter-tags`. It's basically removing a different child from the DOM than what was deleted, so the references change, but of course the indexes don't.

components/tag-list/tag-list-item-mixin.js
bearfriend: Yes, this handles the case where you focus the first tag and then press `Shift+Tab` to go backward. At that point focus is on the tag content, but because `tabindex` is `0` on the host, it focuses the host, but this handler doesn't run again (which would just trap it) I guess because it never lost focus in the light dom. Then you can move backward from there.

components/tag-list/tag-list.js
bearfriend: I think we do want this to run in that case, so if you press `Shift+Tab` from the middle of the list you don't go to the first item before exiting the list. Maybe I'm wrong about that though?

dbatiste (approved):

PR #4505, opened 2024-03-06
Add LoadingCompleteMixin

semibran:
I've been noticing there's some boilerplate that has to be copied between components in order for `getLoadingComplete` to work in stuff that uses the new test framework, the idea here is to just abstract as much of this away as possible to make refactoring a bit easier

Having this in core (or elsewhere) will make it easier to share this logic between repos. Lmk if there's a more suitable place for this to live.


mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -0,0 +1,21 @@
...
+	constructor() {
+		super();
+		this._loadingCompletePromise = new Promise(resolve => {
+			this._loadingCompleteResolve = resolve;
+		});
+	}
+
+	getLoadingComplete() {
+		return this._loadingCompletePromise;
+	}
```
semibran, line 12: The original example has this marked as async which seems a bit redundant --

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -0,0 +1,21 @@
...
+
+	loadingComplete() {
```
semibran, line 14: in the interest of bikeshedding over semantics is there a more descriptive name we could be using for this? Some other ideas I was workshopping:
- `completeLoad`
- `resolveLoadingComplete`
- `setLoadingComplete`
- `markLoadingComplete`

dlockhart: Yeah my only hesitation with `loadingComplete` is in Lit, it has `updateComplete` which is the `Promise` that resolves and `getUpdateComplete()` which is the method you can override to customize it. We kind of borrowed those conventions with `getLoadingComplete()`, so having `loadingComplete` be a method deviates a bit. I kind of like `resolveLoadingComplete()` but curious what others think...

devpow112: Same. I like `resolveLoadingComplete` feels like it avoids the possible confusion with the `updateComplete`

mixins/loading-complete/README.md
```diff
@@ -0,0 +1,22 @@
...
+Apply the mixin and call `loadingComplete()` once your component has finished loading to indicate it is ready for validation.
+
+```js
+import { LoadingCompleteMixin } from '@brightspace-ui/core/mixins/loading-complete/loading-complete-mixin.js';
+
+class MyComponent extends LoadingCompleteMixin(LitElement) {
+
+	connectedCallback() {
+		this._fetchMyData().then(() => {
+			this.loadingComplete();
```
devpow112, line 15: Need to update the docs as well

bearfriend:
This is a great idea. I think if we're going to make this a mixin we need to take a couple more steps to make it more resilient, though. Do you mind if I take a stab at this and open to PR to yours?

semibran:
> Do you mind if I take a stab at this and open a PR to yours?

by all means please do! Curious what you might have in mind wrt resiliency

dlockhart (approved):
Looks great all!

devpow112 (approved):
So happy to see this in `core`. Thanks @semibran and @bearfriend.

mixins/loading-complete/loading-complete-mixin.js
dlockhart, line 1: We should maybe add a few unit tests for this to cover some of the weird inheritance cases?

mixins/loading-complete/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+In some cases, instead of finding one spot to call `resolveLoadingComplete`, you may find it easier to `await` a set of promises in a custom `getLoadingComplete` method.
+
+You'll also need to use this method if you're working in a general-use mixin rather than directly in a component.
+
+```js
+class MyComponent extends LoadingCompleteMixin(LitElement) {
+
+	async getLoadingComplete() {
+		await super.getLoadingComplete();
```
semibran, line 41: running into a bit of a logic snag here while writing tests:
As it stands this way of extending `LoadingCompleteMixin.getLoadingComplete` will never resolve because this entire execution path is just blocked. Naturally it works fine if we call `resolveLoadingComplete` before `getLoadingComplete` but there is a redundancy element there.... curious what folks think

semibran: or wait yeah, this is the intention behind overriding loadingCompletePromise, let me double check my setup

semibran: Ok I think this is just related to order of execution..... chalk it up to user error

PR #4507, opened 2024-03-07
Rework loading complete mixin

bearfriend:
This expands the mixin for more general use. We already have [a number of uses](https://github.com/search?q=saved%3AD2L+%2F%5Cw*%5C.getloadingcomplete%2F&type=code&saved_searches=%5B%7B%22name%22%3A%22D2L%22%2C%22query%22%3A%22org%3ABrightspace+OR+org%3ABrightspaceUI+OR+org%3ABrightspaceUILabs+OR+org%3ABrightspaceHypermediaComponents%22%7D%5D&expanded_query=org%3ABrightspace+org%3ABrightspaceUI+org%3ABrightspaceUILabs+org%3ABrightspaceHypermediaComponents+%2F%5Cw*%5C.getloadingcomplete%2F) of `getLoadingComplete` outside of tests, so we can't rely on it only being used there.

### DIFF
--- a/mixins/loading-complete/README.md
+++ b/mixins/loading-complete/README.md
@@ -17,7 +17,7 @@ class MyComponent extends LoadingCompleteMixin(LitElement) {
 }
 ```
 
-To make use of the `loadingComplete` promise, simply `await` it where needed
+To make use of the `loadingComplete` promise, simply `await` it where needed:
 
 ```js
 async removeSkeleton() {
@@ -26,13 +26,13 @@ async removeSkeleton() {
 }
 ```
 
-If for any reason `resolveLoadingComplete` is never called, `loadingComplete` won't resolve and any consumers will hang
+If for any reason `resolveLoadingComplete` is never called, `loadingComplete` won't resolve and any consumers will hang.
 
 ### `getLoadingComplete`
 
-In some cases, instead of finding one spot to call `resolveLoadingComplete`, you may find it easier to `await` a set of promises in a custom `getLoadingComplete` method
+In some cases, instead of finding one spot to call `resolveLoadingComplete`, you may find it easier to `await` a set of promises in a custom `getLoadingComplete` method.
 
-You'll also need to use this method if you're working in a general-use mixin rather than directly in a component
+You'll also need to use this method if you're working in a general-use mixin rather than directly in a component.
 
 ```js
 class MyComponent extends LoadingCompleteMixin(LitElement) {
@@ -44,4 +44,4 @@ class MyComponent extends LoadingCompleteMixin(LitElement) {
 	}
 }
 ```
-Make sure all your promises are set before `firstUpdated`
+Note that the work to generate these promises should have already started, before `firstUpdated`, and we simply `await` them here.
--- a/mixins/loading-complete/README.md
+++ b/mixins/loading-complete/README.md
@@ -1,9 +1,9 @@
 # LoadingCompleteMixin
 
-The `LoadingCompleteMixin` contains boilerplate code for [`getLoadingComplete`](https://github.com/BrightspaceUI/testing#waiting-for-asynchronous-components). It simplifies fixture setup for custom components in automated tests.
+The `LoadingCompleteMixin` contains boilerplate code for [`getLoadingComplete`](https://github.com/BrightspaceUI/testing#waiting-for-asynchronous-components). It simplifies fixture setup for custom components in automated tests, and provides a `loadingComplete` promise that can be used internally.
 
 ## Usage
-Apply the mixin and call `resolveLoadingComplete()` once your component has finished loading to indicate it is ready for validation.
+Apply the mixin and call `resolveLoadingComplete()` once your component has finished loading
 
 ```js
 import { LoadingCompleteMixin } from '@brightspace-ui/core/mixins/loading-complete/loading-complete-mixin.js';
@@ -11,12 +11,37 @@ import { LoadingCompleteMixin } from '@brightspace-ui/core/mixins/loading-comple
 class MyComponent extends LoadingCompleteMixin(LitElement) {
 
 	connectedCallback() {
-		this._fetchMyData().then(() => {
-			this.resolveLoadingComplete();
-		});
+		this._fetchMyData().then(this.resolveLoadingComplete);
 	}
 
 }
 ```
 
-If for whatever reason `resolveLoadingComplete` is never called, the `getLoadingComplete` promise will never resolve. In such cases any consumers (e.g. fixtures) will hang. If this behavior is not desired, ensure all code paths eventually call `resolveLoadingComplete`, or specify `{ awaitLoadingComplete: false }` in the `fixture` call.
+To make use of the `loadingComplete` promise, simply `await` it where needed
+
+```js
+async removeSkeleton() {
+	await this.loadingComplete;
+	this.skeleton = false;
+}
+```
+
+If for any reason `resolveLoadingComplete` is never called, `loadingComplete` won't resolve and any consumers will hang
+
+### `getLoadingComplete`
+
+In some cases, instead of finding one spot to call `resolveLoadingComplete`, you may find it easier to `await` a set of promises in a custom `getLoadingComplete` method
+
+You'll also need to use this method if you're working in a general-use mixin rather than directly in a component
+
+```js
+class MyComponent extends LoadingCompleteMixin(LitElement) {
+
+	async getLoadingComplete() {
+		await super.getLoadingComplete();
+		await this._myCustomIconImport;
+		await Promise.all(this._allMyDataPromises);
+	}
+}
+```
+Make sure all your promises are set before `firstUpdated`
--- a/mixins/loading-complete/loading-complete-mixin.js
+++ b/mixins/loading-complete/loading-complete-mixin.js
@@ -5,7 +5,7 @@ export const LoadingCompleteMixin = dedupeMixin((superclass) => class extends su
 	#loadingCompleteResolve;
 
 	// eslint-disable-next-line sort-class-members/sort-class-members
-	#loadingCompletePromise = Object.hasOwn(this.constructor.prototype, 'getLoadingComplete')
+	#loadingCompletePromise = !Object.hasOwn(this.constructor.prototype, 'getLoadingComplete')
 		? new Promise(resolve => this.#loadingCompleteResolve = resolve)
 		: Promise.resolve();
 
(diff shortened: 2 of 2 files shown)


mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,36 @@
...
+	get getLoadingComplete() {
+		return this.#getLoadingComplete;
+	}
+
+	get loadingComplete() {
+		return this.#getLoadingComplete();
+	}
+
+	get resolveLoadingComplete() {
+		return this.#resolveLoadingComplete.bind(this);
```
bearfriend, line 21: This makes it unnecessary to wrap before calling.

```js
this.getData.then(this.resolveLoadingComplete);
```

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,36 @@
...
-			this._loadingCompleteResolve = resolve;
-		});
+export const LoadingCompleteMixin = dedupeMixin((superclass) => class extends superclass {
+
+	#loadingCompleteResolve;
+
+	// eslint-disable-next-line sort-class-members/sort-class-members
+	#loadingCompletePromise = this.getLoadingComplete === this.#getLoadingComplete
+		? new Promise(resolve => this.#loadingCompleteResolve = resolve)
+		: Promise.resolve();
```
bearfriend, line 10: If there is already a `getLoadingComplete` method, don't wait for `resolveLoadingComplete` to be called.

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,36 @@
...
+
+	get getLoadingComplete() {
+		return this.#getLoadingComplete;
+	}
+
+	get loadingComplete() {
+		return this.#getLoadingComplete();
+	}
```
bearfriend, line 18: Allows `await this.loadingComplete`, just like `updateComplete`

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,36 @@
...
+	get resolveLoadingComplete() {
+		return this.#resolveLoadingComplete.bind(this);
 	}
 
-	getLoadingComplete() {
-		return this._loadingCompletePromise;
+	async #getLoadingComplete() {
+		await super.getLoadingComplete?.();
+		return this.#loadingCompletePromise;
 	}
```
bearfriend, line 27: Other mixins may already be extending this one, or may just have their own `getLoadingComplete` methods already, so we need to make sure those are called.

mixins/loading-complete/README.md
```diff
@@ -1,22 +1,47 @@
...
-			this.resolveLoadingComplete();
-		});
+		this._fetchMyData().then(this.resolveLoadingComplete);
 	}
 
 }
 ```
 
-If for whatever reason `resolveLoadingComplete` is never called, the `getLoadingComplete` promise will never resolve. In such cases any consumers (e.g. fixtures) will hang. If this behavior is not desired, ensure all code paths eventually call `resolveLoadingComplete`, or specify `{ awaitLoadingComplete: false }` in the `fixture` call.
+To make use of the `loadingComplete` promise, simply `await` it where needed
```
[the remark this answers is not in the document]
bearfriend: It's not me! https://daylight.d2l.dev/style-elements/writing/#periods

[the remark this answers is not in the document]
bearfriend: But I'm fine if we don't want to follow that here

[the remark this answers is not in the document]
dlockhart: Hmm... I think that's intended more in the context of an application interface? Like if there's a label that says "Due tomorrow", you don't need a period.

@nicolebezaire could you clarify what the recommendation would be here where we're just writing documentation?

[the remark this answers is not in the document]
bearfriend: I also personally like omitting the period because the text is leading into a relevant code example but I admit that may be weird.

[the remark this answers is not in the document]
svanherk: Then I usually do a `:`, but I agree this is all probably personal preference. I like the periods but 🤷

mixins/loading-complete/loading-complete-mixin.js
devpow112: Why not use an arrow capture here?
```js
return ()=> this.#resolveLoadingComplete();
```
Not sure if there's a draw back/better reason to using bind or `() =>`

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,36 @@
...
+export const LoadingCompleteMixin = dedupeMixin((superclass) => class extends superclass {
+
+	#loadingCompleteResolve;
+
+	// eslint-disable-next-line sort-class-members/sort-class-members
+	#loadingCompletePromise = this.getLoadingComplete === this.#getLoadingComplete
+		? new Promise(resolve => this.#loadingCompleteResolve = resolve)
+		: Promise.resolve();
+
+	get getLoadingComplete() {
```
dlockhart, line 12: Haha I think I'm getting confused by all the different things called `getLoadingComplete`. For my own sanity, is this correct?
- This is a getter for the `getLoadingComplete()` method that consumers can optionally override to implement custom logic. If they do override it, they're still expected to wait for `super.getLoadingComplete()`.
- If they don't override it, the default behaviour is to return a reference to the _private_ `#getLoadingComplete()` method
- That private method waits for `super.getLoadingComplete` in case other mixins/etc. have things to wait for and then finally waits for the private `#loadingComplete` Promise

What's the purpose of the private method in all this? Couldn't its code just move into here?

mixins/loading-complete/loading-complete-mixin.js
bearfriend: In this case, I don't think there's a difference. Just an old habit to use `bind` in this scenario.

semibran: what's the difference (functional/semantic/otherwise) between this vs. just keeping `#resolveLoadingComplete` public?

mixins/loading-complete/loading-complete-mixin.js
semibran: Quick double check around order of execution here, this is essentially sugar for constructor-level definitions right? i.e. do we expect this to evaluate to `Promise.resolve` in cases where the consumer component overrides `getLoadingComplete`, and not just in cases where mixins override it?

mixins/loading-complete/loading-complete-mixin.js
bearfriend: I don't think it makes a difference since these are just the promises. The actual work has already been kicked off.

mixins/loading-complete/loading-complete-mixin.js
semibran: ok right yea that makes sense, fresh eyes help me read better

mixins/loading-complete/loading-complete-mixin.js
bearfriend: Yes, it's the same thing as a constructor assignment, but I'll have to double-check the order and if this actually works with mixed use between component and mixin. I think it could fail to resolve the default promise as-is.

mixins/loading-complete/README.md
```diff
@@ -1,22 +1,47 @@
...
+
+In some cases, instead of finding one spot to call `resolveLoadingComplete`, you may find it easier to `await` a set of promises in a custom `getLoadingComplete` method
+
+You'll also need to use this method if you're working in a general-use mixin rather than directly in a component
+
+```js
+class MyComponent extends LoadingCompleteMixin(LitElement) {
+
+	async getLoadingComplete() {
+		await super.getLoadingComplete();
```
dbatiste, line 41: Hopefully people who use this would recognize that this would be sequential, and if they want them all in parallel then they would do this instead:
```javascript
getLoadingComplete() {
    return Promise.all([
        super.getLoadingComplete(),
        this._myCustomIconImport,
        ...this._allMyDataPromises
    ]);
}
```

bearfriend: These are just the promises, though, not the things that start the work and return the promises. However, maybe that's unclear here and we should just document it this way to avoid confusion.

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,32 @@
...
-		super();
-		this._loadingCompletePromise = new Promise(resolve => {
-			this._loadingCompleteResolve = resolve;
-		});
+export const LoadingCompleteMixin = dedupeMixin((superclass) => class extends superclass {
+
+	#loadingCompleteResolve;
+
+	// eslint-disable-next-line sort-class-members/sort-class-members
+	#loadingCompletePromise = Object.hasOwn(this.constructor.prototype, 'getLoadingComplete')
```
bearfriend, line 8: @dbatiste This is safe now, right?! 😆

mixins/loading-complete/README.md
dbatiste: I see yeah. I wonder if anyone will ever be tempted to call a method here that actually kicks off work...?

mixins/loading-complete/loading-complete-mixin.js
bearfriend: I got rid of the private method by using `hasOwn` instead of comparing the two methods for equality, which didn't work if the mixin chain was swapped around. 2-for-1 there.

mixins/loading-complete/loading-complete-mixin.js
bearfriend: Ok, that is exactly what was happening if I swapped the mixin chain around. Fixed by using `hasOwn` instead of comparing the methods.

mixins/loading-complete/README.md
bearfriend: Yeah, we should definitely call that out better.

mixins/loading-complete/loading-complete-mixin.js
```diff
@@ -1,21 +1,32 @@
...
+
+	get loadingComplete() {
+		return this.getLoadingComplete();
 	}
 
-	getLoadingComplete() {
-		return this._loadingCompletePromise;
+	get resolveLoadingComplete() {
+		return () => this.#resolveLoadingComplete();
 	}
```
semibran, line 18: Without the bind this has me tempted to suggest dropping the `#resolveLoadingComplete` body directly into the arrow definition just so we can strip out that extra layer of indirection, it's not being used elsewhere

bearfriend: Yes! +1 arrow functions

mixins/loading-complete/loading-complete-mixin.js
dbatiste: In BSI context, yeah. In a FRA context, it would require their build process to include the Babel plugin for transpiling. So if this would ever make its way into a FRA, that would be a risk.

dbatiste: Yeah, I would probably avoid `hasOwn` if there is any worry it will be pulled into other things outside of BSI.

semibran (approved):
LGTM, thanks for putting this all together! Big improvement over the original

devpow112 (approved):

PR #4512, opened 2024-03-11
feat: start new dropdown feature for sortable list

saifaldin14:
So far, I've added a new property to the table sort for sortOptions and the intention it to render this in a dropdown where the user can select which sort they want to go by

### DIFF
--- a/components/table/table-col-sort-button.js
+++ b/components/table/table-col-sort-button.js
@@ -27,6 +27,14 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 			nosort: {
 				reflect: true,
 				type: Boolean
+			},
+			/**
+			 * Show the different sort options available.
+			 * @type {String}
+			 */
+			sortOptions: {
+				reflect: true,
+				type: String
 			}
 		};
 	}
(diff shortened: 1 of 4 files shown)


components/table/table-col-sort-button.js
```diff
@@ -27,6 +27,14 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				type: Boolean
+			},
+			/**
+			 * Show the different sort options available.
+			 * @type {String}
+			 */
+			sortOptions: {
+				reflect: true,
+				type: String
 			}
```
margaree, line 38: I think this strategy makes sense. However, you'd likely want to use the `Array` type (there are [some examples](https://search.d2l.dev/search?project=BrightspaceUI&full=%22type%3A+Array%22&defs=&refs=&path=%22BrightspaceUI%2Fcore%22&hist=&type=&xrd=&nn=1&si=path&si=path)) in core). Also I'd recommend taking a look at [the Daylight properties and attributes documentation](https://daylight.d2l.dev/developing/best-practices/attributes-properties/#do%3A-use-correct-casing-for-attributes-and-properties).

PR #4516, opened 2024-03-11
Fix test names for Windows

bearfriend:
Windows is mad about `:` being in file paths.


svanherk:
I'd probably rename the goldens to what you expect, then we can verify there's no `vdiff` changes to them?

dlockhart (approved):

components/tooltip/test/tooltip.vdiff.js
```diff
@@ -146,7 +146,7 @@ describe('tooltip', () => {
 				{ x: 50, y: 100 }
 			].forEach(({ x, y }) => {
```
dlockhart, line 148: Should we maybe be doing something more on the `@brightspace-ui/testing` in the conversion from test suite/name -> filename? Like just remove/replace certain characters so that nobody can accidentally cause this?

dlockhart: Ah looks like you're already on it!

PR #4525, opened 2024-03-18
fix: NIM1872- add some vdiff tests for cases where table-wrapper has stickyHeadersScrollWrapper enabled

derekmoran:
Add vdiff tests for [d2l-table-wrapper](https://daylight.d2l.dev/components/table/#d2l-table-wrapper) for the case of its _stickyHeadersScrollWrapper_ property being enabled.

The [current state](https://d2l.slack.com/archives/C0PHG3QB0/p1709128112399439) of some these are not ideal. That said I think it would be useful to have them here, and I _**may**_ be attempting to improve some of these in the near future in which case it will be invaluable to see what affect such attempts are having.


components/table/test/table.vdiff.js
```diff
@@ -424,181 +425,223 @@ describe('table', () => {
...
-						name: 'one-row-thead',
-						template: html`
-								<thead>${createHeaderRow({ trClass: 'top' })}</thead>
+			// only test the default type to minimize permutations (hasStickyHeadersScrollWrapper isn't expected to have a different impact on the light vs default type)
+			(type === 'default' ? [true, false] : [false]).forEach((hasStickyHeadersScrollWrapper) => {
+				describe(`sticky${hasStickyHeadersScrollWrapper ? '-scroll-wrapper' : ''}`, () => {
+
+					// Currently some of the scenarios with 'hasStickyHeadersScrollWrapper' setting enabled have unexpected outputs
+					// This is used to call those out by wrapping a warning grouping around them
+					function itWithWarnings(testName, testFunc) {
```
derekmoran, line 434: I'm aiming to call out the specific tests that I believe currently have unexpected outputs, as mentioned in an earlier review [here](https://github.com/BrightspaceUI/core/pull/4481#discussion_r1516912332). 

This might be a touch fragile, but I liked how clearly it calls them out in the report  🤷   I welcome any and all alternative suggestions  :)

derekmoran:
FWIW I've come up with a [solution that fixes our specific case in assessments](https://github.com/BrightspaceHypermediaComponents/activities/pull/4498) without having to touch core.

I think these new vdiffs might still be useful, but I'm no longer planning to attempt a fix from core in the near future, so happy to back away from this slowly too 🤷

dlockhart (approved):
Thanks for adding these. I agree, they're definitely still useful!

TristanDevos (approved):

PR #4537, opened 2024-03-21
feat: Add styles for sortable table button

saifaldin14:
[Jira Ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-222)
Screenshot (104)
Screenshot (105)


components/table/demo/table-test.js
```diff
@@ -140,20 +143,26 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
...
 		return html`
-			<th scope="col">
+			<th class="sortableCell" scope="col">
 				<d2l-table-col-sort-button
 					@click="${this._handleSort}"
 					?desc="${this._sortDesc}"
-					?nosort="${noSort}">${fruit}</d2l-table-col-sort-button>
+					?nosort="${noSort}">
+					${fruit}
+				</d2l-table-col-sort-button>
```
margaree, line 165: Are these changes needed for this PR?

components/table/table-wrapper.js
```diff
@@ -248,6 +248,7 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-header-background-color: var(--d2l-color-regolith);
 				--d2l-table-row-border-color-selected: var(--d2l-color-celestine);
 				--d2l-table-row-background-color-selected: var(--d2l-color-celestine-plus-2);
+				--d2l-sortable-button-border-radius: 0;
```
margaree, line 251: This seems like a good way to do this change👍

components/table/table-col-sort-button.js
```diff
@@ -77,9 +95,14 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 
 	render() {
 		const iconView = !this.nosort ?
-			html`<d2l-icon icon="${this.desc ? 'tier1:arrow-toggle-down' : 'tier1:arrow-toggle-up'}"></d2l-icon>` :
+			html`<d2l-icon class="d2l-sortable-button-icon" icon="${this.desc ? 'tier1:arrow-toggle-down' : 'tier1:arrow-toggle-up'}"></d2l-icon>` :
 			null;
-		return html`<button type="button"><slot></slot>${iconView}</button>`;
+
+		return html`
+			<div class="d2l-sortable-button-container">
```
margaree, line 102: Is this container actually needed or would we be able to use `:host` in the css?

margaree (commented):
I haven't reviewed the css changes in the component in a huge amount of detail because I think the change I mentioned in the demo file will cause changes to the component css, so I'll take another look at that once that piece gets figured out.

components/table/table-col-sort-button.js
margaree: Did you look into why that would be?

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,40 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				font-family: inherit;
 				font-size: inherit;
+				gap: 12px;
+				height: calc(var(--d2l-table-cell-height) + var(--d2l-sortable-button-height));
 				letter-spacing: inherit;
 				margin: 0;
-				padding: 0;
+				padding-inline-start: var(--d2l-sortable-button-padding);
 				text-decoration: none;
+				text-wrap: nowrap;
```
margaree, line 52: You should confirm with Nicole what the desired overflow behaviour would be. We'd likely still want it to wrap if that's what it was doing before - it was more that the text in the visual diff now wrapping was a concern because it indicated something big changed.

Also it looks like [text-wrap](https://caniuse.com/?search=text-wrap) is pretty new and might not work as expected for all our users.

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,40 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				background-color: transparent;
 				border: none;
+				box-sizing: border-box;
 				color: inherit;
 				cursor: pointer;
 				display: inline-flex;
 				font-family: inherit;
 				font-size: inherit;
+				gap: 12px;
+				height: calc(var(--d2l-table-cell-height) + var(--d2l-sortable-button-height));
```
margaree, line 47: I find this a bit confusing because `--d2l-sortable-button-height` also itself depends on `--d2l-table-cell-height`. Is this something to do with the dropdown button that we should be leaving out of this PR or is unrelated? If not, can this be done a bit more clearly?

I also noticed that `--d2l-table-cell-height` isn't actually used in connection with the header (it's overwritten with `27px`) so it might be best to not involve it.

components/table/table-wrapper.js
```diff
@@ -248,6 +248,12 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
+				--d2l-sortable-button-padding: 1rem;
```
margaree, line 252: The spec shows the proposed left padding for both default and light as 15 px. Did this get changed?

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,40 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 			button:focus-visible,
 			button:${unsafeCSS(getFocusPseudoClass())} {
-				border-radius: 0.2rem;
+				border-radius: var(--d2l-sortable-button-border-radius);
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine);
+				height: var(--d2l-sortable-button-focus-height);
+				margin-inline-start: calc(var(--d2l-sortable-button-padding) * var(--d2l-sortable-button-focus-margin));
 				outline-style: none;
+				padding-inline-start: calc(var(--d2l-sortable-button-padding) * (1 - var(--d2l-sortable-button-focus-margin)));
+				width: var(--d2l-sortable-button-focus-width);
```
margaree, line 69: Does this need to use a `box-shadow`? I think a lot of it might be able to be simplified with a `border` instead?

components/table/table-wrapper.js
```diff
@@ -368,6 +380,10 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 			r.classList.toggle('d2l-table-selected-first', firstNonHeaderRow && isSelected);
 
 			Array.from(r.cells).forEach((c, index) => {
+				const isSortableCell = Array.from(c.childNodes).find((element) => element.localName === 'd2l-table-col-sort-button');
+				if (isSortableCell) {
+					c.style.padding = 0;
+				}
```
margaree, line 386: I believe this should only be run if `isHeader` is true. We can also likely follow the same pattern here and toggle a class (e.g., `d2l-table-header-col-sortable`) and then add the style in the css. If we do that, we can put setting the padding in the css next to or near the place where the padding was originally set, making it more readable as to where it's being set and why. Using a class will also allow us to add other css that depends on that class if the need comes up.

components/table/table-wrapper.js
```diff
@@ -262,6 +268,12 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-cell-padding-alt: calc(0.6rem - 1px) 0.6rem 0.6rem 0.6rem;
 				--d2l-table-border-color: var(--d2l-color-gypsum);
 				--d2l-table-header-background-color: #ffffff;
+				--d2l-sortable-button-border-radius: 0.2rem;
+				--d2l-sortable-button-padding: 0.6rem;
+				--d2l-sortable-button-height: calc(var(--d2l-table-cell-height) * 0.9);
```
margaree, line 273: The height of the hovered light items looks quite tall as compared to the spec. Did the spec get changed?

components/table/demo/table-test.js
```diff
@@ -153,7 +153,9 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
...
```
margaree, line 158: Super nit - you can put this back to what it was before so there isn't an extra change to it in the commit history where nothing really actually changed.

components/table/table-col-sort-button.js
saifaldin14: With a border instead it looks like this:

Screenshot (119)

I think that might be why the original code used box-shadow

components/table/table-col-sort-button.js
saifaldin14: I think last time the wrapping occured because of the added space between the text and icon

components/table/table-wrapper.js
saifaldin14: I modified it to use the spec height of 40px

components/table/table-col-sort-button.js
margaree: If all the new properties are removed and the box-shadow gets replaced with a border then the sizing does look correct but there is a slight flickering in the position of the text, so that's likely why a box-shadow was used instead. I'm still thinking it would be possible to simplify this a bit though.

saifaldin14: Ok, I was able to find a way to use border instead

components/table/table-wrapper.js
```diff
@@ -243,11 +245,16 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-border-radius: 0.3rem;
 				--d2l-table-border-radius-sticky-offset: calc(1px - var(--d2l-table-border-radius));
 				--d2l-table-cell-height: 41px; /* min-height to be 62px including border */
-				--d2l-table-cell-padding: 0.5rem 1rem;
+				--d2l-table-cell-padding: 0.5rem 15px;
```
margaree, line 248: Would we not want to use this padding within the table-col-sort component as well? 

Also we'd likely also want to use rem for this as well (`0.75 rem` for this case; it's generally spec px value / 20 px)

components/table/table-wrapper.js
```diff
@@ -243,11 +245,16 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
 				--d2l-table-cell-padding-alt: calc(0.5rem - 1px) 1rem 0.5rem 1rem;
 				--d2l-table-header-background-color: var(--d2l-color-regolith);
 				--d2l-table-row-border-color-selected: var(--d2l-color-celestine);
 				--d2l-table-row-background-color-selected: var(--d2l-color-celestine-plus-2);
+				--d2l-sortable-button-border-radius: 0;
+				--d2l-sortable-button-vertical-padding: 1rem;
+				--d2l-sortable-button-height: 50px;
```
margaree, line 255: Is this in the spec? I didn't see it. From the spec it looks like we'd just want to have top and bottom padding of 1.5rem. If you haven't yet, I'd recommend testing the table without the other types of cells in the header row to see what the height looks like if there were only the sort buttons (i.e., just delete the header cells that are the `fruits.map`)

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,39 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine);
+				height: var(--d2l-sortable-button-focus-height);
+				margin-inline-start: 0.2rem;
 				outline-style: none;
+				padding-inline-start: 0.55rem;
+				width: 95%;
+			}
+			button:hover {
+				background-color: var(--d2l-color-gypsum);
+				border-radius: var(--d2l-sortable-button-border-radius);
```
margaree, line 72: In the spec, the border-radius is actually different for light mode in hover vs focus states.

components/table/table-wrapper.js
```diff
@@ -262,6 +267,9 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-cell-padding-alt: calc(0.6rem - 1px) 0.6rem 0.6rem 0.6rem;
 				--d2l-table-border-color: var(--d2l-color-gypsum);
 				--d2l-table-header-background-color: #ffffff;
+				--d2l-sortable-button-border-radius: 0.2rem;
+				--d2l-sortable-button-height: 90%;
+				--d2l-sortable-button-focus-height: 80%;
```
margaree, line 272: In the spec, light mode takes up the full height (goes down to the border) when focused, so this might need some tweaking.

components/table/table-wrapper.js
```diff
@@ -243,11 +245,14 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
 				--d2l-table-cell-height: 41px; /* min-height to be 62px including border */
-				--d2l-table-cell-padding: 0.5rem 1rem;
+				--d2l-table-cell-padding: 0.5rem 0.75rem;
 				--d2l-table-cell-padding-alt: calc(0.5rem - 1px) 1rem 0.5rem 1rem;
 				--d2l-table-header-background-color: var(--d2l-color-regolith);
 				--d2l-table-row-border-color-selected: var(--d2l-color-celestine);
 				--d2l-table-row-background-color-selected: var(--d2l-color-celestine-plus-2);
+				--d2l-sortable-button-border-radius: 0;
+				--d2l-sortable-button-height: 100%;
+				--d2l-sortable-button-focus-height: 85%;
```
margaree, line 255: I'm wondering if it's possible to use a calculation for this rather than a semi-mysterious percentage.

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,39 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
-			}
 			button:focus-visible,
 			button:${unsafeCSS(getFocusPseudoClass())} {
-				border-radius: 0.2rem;
+				border-radius: var(--d2l-sortable-button-border-radius);
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine);
+				height: var(--d2l-sortable-button-focus-height);
+				margin-inline-start: 0.2rem;
 				outline-style: none;
+				padding-inline-start: 0.55rem;
```
margaree, line 67: Where did these numbers come from for this and `margin-inline-start`? Is it possible to have a calculation which makes their values more clear, or maybe a small comment that specifies how you came to them?

margaree (commented):
I noticed that when testing, if I clicked the "select all" checkbox then uncheck it, all the cell contents shift. This didn't previously happen so it's something to look into.

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,39 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 			button:disabled {
 				opacity: 0.5;
 			}
-			button:hover {
-				text-decoration: underline;
-			}
 			button:focus-visible,
 			button:${unsafeCSS(getFocusPseudoClass())} {
-				border-radius: 0.2rem;
+				border-radius: calc(var(--d2l-sortable-button-border-radius) * 1.5); /* To get 6px (4px * 1.5) in light mode, will remain 0 in normal case */
```
margaree, line 62: As far as I know these values aren't really dependant on one another and should be separately specified.

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,39 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine);
+				height: calc(100% - 0.4rem);
+				margin-inline-start: 0.2rem; /* The margin moves the focus box shadow to the left and makes it stay within the cell lines */
```
margaree, line 65: Sorry I meant more just define why 0.2rem not what the margin is meant to be doing

components/table/table-wrapper.js
```diff
@@ -262,6 +267,9 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-cell-padding-alt: calc(0.6rem - 1px) 0.6rem 0.6rem 0.6rem;
 				--d2l-table-border-color: var(--d2l-color-gypsum);
 				--d2l-table-header-background-color: #ffffff;
+				--d2l-sortable-button-border-radius: 4px;
```
margaree, line 270: We would likely still want this to be rem

components/table/table-wrapper.js
```diff
@@ -243,11 +245,14 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-border-radius: 0.3rem;
 				--d2l-table-border-radius-sticky-offset: calc(1px - var(--d2l-table-border-radius));
 				--d2l-table-cell-height: 41px; /* min-height to be 62px including border */
-				--d2l-table-cell-padding: 0.5rem 1rem;
-				--d2l-table-cell-padding-alt: calc(0.5rem - 1px) 1rem 0.5rem 1rem;
+				--d2l-table-cell-padding: 0.5rem 0.75rem;
+				--d2l-table-cell-padding-alt: calc(0.5rem - 1px) 0.75rem 0.5rem 0.75rem;
```
margaree, line 249: One other thing to confirm with Nicole is that this also changes the end padding so we'd just want to confirm that that should now also be decreased everywhere.

margaree (commented):
A few more things but it's definitely almost there. The biggest thing is just confirming with Nicole if the padding changes are meant to be in all table cells.

I'd also recommend checking with her on the hover + focus state, since they have different border radiuses.

saifaldin14:
Yeah she confirmed that padding should be applied to all. As well as the border radiuses being different, that was intentional

components/table/table-wrapper.js
```diff
@@ -258,10 +264,14 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 				--d2l-table-border-radius: 0rem; /* stylelint-disable-line length-zero-no-unit */
 				--d2l-table-border-radius-sticky-offset: 0rem; /* stylelint-disable-line length-zero-no-unit */
 				--d2l-table-cell-height: 1.15rem; /* min-height to be 48px including border */
-				--d2l-table-cell-padding: 0.6rem;
-				--d2l-table-cell-padding-alt: calc(0.6rem - 1px) 0.6rem 0.6rem 0.6rem;
+				--d2l-table-cell-padding: 0.75rem;
+				--d2l-table-cell-padding-alt: calc(0.75rem - 1px) 0.75rem 0.75rem 0.75rem;
```
margaree, line 268: Did you confirm all this with Nicole? This modifies the padding on all sides for light style for the regular table cells, whereas for default you only modified the left/right.

components/table/table-wrapper.js
```diff
@@ -258,10 +264,14 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
 				--d2l-table-border-color: var(--d2l-color-gypsum);
 				--d2l-table-header-background-color: #ffffff;
+				--d2l-sortable-button-border-radius: 0.2rem;
+				--d2l-sortable-button-border-focus-radius: 0.3rem;
+				--d2l-sortable-button-height: calc(100% - 0.4rem);
+				--d2l-sortable-button-width: calc(100% - 0.4rem);
```
margaree, line 274: These should also use px instead of rem

margaree (commented):
It looks really nice testing it out. Just the one little tweak and then the question to Nicole - it would be useful to know how much we should be modifying the other cell padding vs just the sort header cell.

components/table/table-wrapper.js
margaree: The spec is specifically for the sort header though, so it would be good to confirm if it's also all cells.
If that is the case then the default padding would also need updating.

saifaldin14: I sent her another message about this specifically but this is what she told me before:

Yes, the adjustment to the cell padding was intended across all header cells - and I suppose that might make sense to apply across all cells so that the header content lines up with the cell content below it

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,47 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				cursor: pointer;
 				display: inline-flex;
 				font-family: inherit;
 				font-size: inherit;
+				height: var(--d2l-sortable-button-height);
 				letter-spacing: inherit;
+				line-height: 0.9rem;
 				margin: 0;
-				padding: 0;
+				padding: 0.75rem;
```
margaree, line 50: I would still recommend using `--d2l-table-cell-padding` here. Since we want the padding of cells to now all be consistent, then if it gets overridden this location also stays consistent.

components/table/table-col-sort-button.js
```diff
@@ -37,30 +37,47 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
-			}
 			button:focus-visible,
 			button:${unsafeCSS(getFocusPseudoClass())} {
-				border-radius: 0.2rem;
+				border-radius: var(--d2l-sortable-button-border-focus-radius);
 				box-shadow: 0 0 0 2px #ffffff, 0 0 0 4px var(--d2l-color-celestine);
+				height: calc(100% - 8px);
+				margin-inline-start: 4px; /* Used to offset the outer box shadow */
 				outline-style: none;
+				padding-inline-start: calc(0.75rem - 4px);
```
margaree, line 67: If you use the `--d2l-table-cell-padding` var, you should also use it here for this calculation.

components/table/table-wrapper.js
```diff
@@ -32,6 +32,9 @@ export const tableStyles = css`
 	d2l-table-wrapper[dir="rtl"] .d2l-table > * > tr > * {
 		text-align: right;
 	}
+	.d2l-table-header-col-sortable {
+		--d2l-table-cell-padding: 0;
```
margaree, line 36: If you use the var for padding, this might just need to be `padding: 0`

margaree (commented):
Other than the mentioned usage of the var, it LGTM. You'll want to review the vdiff PR carefully, since it looks like this impacted a lot of places, so just make sure there isn't anything unexpected in there.

margaree (approved):

PR #4543, opened 2024-03-26
fix: Tooltip: Safari: show-truncated-only not working [GAUD-6196]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6196)

**Problem**:
In the tag-list the truncation tooltip is always showing (other than on first focus). For example:
Screenshot 2024-03-26 at 1 11 31 PM

This was also causing the keyboard instructional tooltip to not show.

Our expectation seems to be that generally `scrollWidth` of the clone is equal to `offsetWidth` of the target. However, Safari seems to be adding 1px to the `scrollWidth`. I couldn't find a why for this, but let me know if there's something I missed there.

### DIFF
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -998,7 +998,7 @@ class Tooltip extends RtlMixin(LitElement) {
 			await clone.updateComplete;
 		}
 
-		this._truncating = clone.scrollWidth > target.offsetWidth;
+		this._truncating = (clone.scrollWidth - target.offsetWidth) > 2;
 		this._resizeRunSinceTruncationCheck = false;
 		target.removeChild(cloneContainer);
 	}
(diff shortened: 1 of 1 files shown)


components/tooltip/tooltip.js
```diff
@@ -998,7 +998,7 @@ class Tooltip extends RtlMixin(LitElement) {
 			await clone.updateComplete;
 		}
 
-		this._truncating = clone.scrollWidth > target.offsetWidth;
+		this._truncating = (clone.scrollWidth - target.offsetWidth) > 2;
```
dbatiste, line 1001: Could it be related to a border somewhere? Granted, I think browsers are pretty consistent with measurement nowadays?

It might be worth adding a comment in the code for future us.

dbatiste (approved):

PR #4552, opened 2024-03-28
Validate output of web-component-analyzer

dlockhart:
Here's an alternative to the full-page vdiffs for the Daylight Site. This generates `custome-elements.json` as part of the build, and then runs some validation logic in the CI workflow.


components/selection/selection-controls.js
```diff
@@ -14,6 +14,7 @@ import { SelectionObserverMixin } from './selection-observer-mixin.js';
 /**
  * Controls for selection components (e.g. list, table-wrapper) containing select-all, etc.
  * @slot - Responsive container using `d2l-overflow-group` for `d2l-selection-action` elements
+ * @fires d2l-selection-observer-subscribe - Internal event
```
dlockhart, line 17: Caught a missing event description!
Screenshot 2024-03-28 at 3 01 40 PM

dlockhart:
I should mention that I first looked at [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc/tree/main), but it doesn't have rules for the things we need, even if they sound like it does.

cli/validate-wca.js
```diff
@@ -0,0 +1,66 @@
+import chalk from 'chalk';
+import { fileURLToPath } from 'url';
+import path from 'path';
+import { readFile } from 'fs/promises';
```
bearfriend, line 4:
```suggestion
import { fileURLToPath } from 'node:url';
import path from 'node:path';
import { readFile } from 'node:fs/promises';
```

cli/validate-wca.js
```diff
@@ -0,0 +1,66 @@
...
+
+	// assume an element with no description is to be ignored
+	if (!tag.description) {
+		return 0;
+	}
+
+	let errors = 0;
+
+	tag.attributes?.forEach(a => {
+		if (!a.description) {
```
dbatiste, line 19: I'm guessing there's no way that the description could be an empty string, ya?

dlockhart: This will error when description is `undefined` or an empty string.

cli/validate-wca.js
```diff
@@ -0,0 +1,66 @@
...
+			console.log(`Attribute "${a.name}" on tag "${tag.name}" is missing a description.`);
+			errors++;
+		}
+	});
+
+	tag.events?.forEach(e => {
```
dbatiste, line 25: Should we be checking that all slots have a description too?

dlockhart: I was going to say "no, the list of slots come from the comments" but even though that _is_ true, they could technically still go `@slot foo` and not provide a description so I'll add a check for it!

cli/validate-wca.js
dbatiste: Oh right. I don't know why I always thing empty string needs it's own check. 🤷‍♂️

cli/validate-wca.js
dlockhart: Done!

dbatiste (approved):
This looks great to me!

margaree (approved):

PR #4556, opened 2024-04-01
fix: Make buttons with siblings not use new styles

saifaldin14:
[Jira](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6211)

If a button has any siblings the sort button will use the old styles and not the new table header:

Screenshot (132)
Screenshot (133)
Screenshot (139)
Screenshot (134)
Screenshot (135)
Screenshot (136)
Screenshot (137)
Screenshot (138)


components/table/table-col-sort-button.js
```diff
@@ -27,13 +27,21 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 			nosort: {
 				reflect: true,
 				type: Boolean
+			},
+			/**
+			 * Check to see if the button has siblings.
+			 * @type {boolean}
+			 */
+			hasSiblings: {
```
margaree, line 35: I'd recommend `hasSibling` instead since it only matters if they have at least one.
Also we'd want the doc comments for this to just be `@ignore` since we wouldn't want internal consumers setting this.
See [Daylight > Attributes & Properties](https://daylight.d2l.dev/developing/best-practices/attributes-properties/#do%3A-use-correct-casing-for-attributes-and-properties) for details on defining this property and corresponding attribute.

components/table/table-wrapper.js
```diff
@@ -381,6 +391,25 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		});
 	}
 
+	_checkSiblingSortableCells(c) {
+		const nodes = Array.from(c.childNodes);
+		const isSortableCell = nodes.some((element) => element.localName === 'd2l-table-col-sort-button');
```
margaree, line 396: I'd recommend an early return instead, since there isn't anything that happens if it's not a sortable cell, e.g., 
```suggestion
		const isSortableCell = nodes.some((element) => element.localName === 'd2l-table-col-sort-button');
		if (!isSortableCell) return;
```

components/table/table-wrapper.js
```diff
@@ -381,6 +391,25 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
+
+		if (isSortableCell) {
+			let siblings = 0;
+			nodes.forEach((element) => {
+				if (element.localName) siblings++;
+			});
```
margaree, line 402: Is there a way to just get the count of number of items in the cell? Then we'd know if it was more than 1 then there are siblings.

components/table/table-wrapper.js
```diff
@@ -381,6 +391,25 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
+
+			c.classList.toggle('d2l-table-header-col-sortable', siblings === 1);
```
margaree, line 404: Is this saying if there is only one item in the cell then toggle this class? If so, I'd recommend a bit of restructuring with this as `siblings` is a bit misleading here.

components/table/table-wrapper.js
```diff
@@ -381,6 +391,25 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
+			if (siblings > 1) {
+				nodes.forEach((element) => {
+					element.hasSiblings = element.localName === 'd2l-table-col-sort-button';
```
margaree, line 407: nit for readability
```suggestion
					if (element.localName === 'd2l-table-col-sort-button') element.hasSibling = true;
```

components/table/test/table.vdiff.js
```diff
@@ -27,6 +27,19 @@ function createSortableHeaderRow() {
 		</tr>
 	`;
 }
+
+function createSortableHeaderRowWithDoubleButton() {
```
margaree, line 31: Another case that's worth adding is when it's a `d2l-table-col-sort-button` with a different sibling (e.g., the case with the col-sort-button next to the context menu).

margaree (commented):
Something else to look into is if `d2l-table-col-sort-button` is used outside of the header. I would think it shouldn't be, but it's worth a quick look into.

components/table/test/table.vdiff.js
```diff
@@ -415,6 +459,50 @@ describe('table', () => {
...
+				describe('col-sort-button-double', () => {
+					it('another-button', async() => {
+						const elem = await createTableFixture(html`
+							<thead>${createSortableHeaderRowWithTwoButtons()}</thead>
+							<tbody>${createRows([1])}</tbody>
+						`);
+						await expect(elem).to.be.golden();
+					});
+
+					it('another-button-focus', async() => {
```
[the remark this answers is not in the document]
margaree: It looks like there aren't any visual diffs for testing hover. If it's not too complicated, I'd recommend adding those. If it seems like a lot of extra work, you can spin up a new ticket for that.

components/table/test/table.vdiff.js
```diff
@@ -27,6 +27,50 @@ function createSortableHeaderRow() {
...
+		</tr>
+	`;
+}
+
+function createSortableHeaderRowWithTwoElements() {
+	return html`
+		<tr>
+			<th>
+				<d2l-table-col-sort-button>Double 1</d2l-table-col-sort-button>
+				<div>Dummy text</div>
```
margaree, line 49: I noticed that "Dummy text" is wrapping under the button. Is this because it's long or for another reason? Perhaps take a look at what's going on in the DOM in the context menu case in grades and add a visual diff for that case so that we can be sure that one will behave as expected.

components/table/test/table.vdiff.js
```diff
@@ -27,6 +27,48 @@ function createSortableHeaderRow() {
...
+}
+
+function createSortableHeaderRowWithIconElement() {
+	return html`
+		<tr>
+			<th scope="col">
+				<d2l-table-col-sort-button>
+					Status
+				</d2l-table-col-sort-button>
+				<d2l-button-icon icon="tier1:help" type="button"></d2l-button-icon>
```
margaree, line 64: You'll want to add an import for `button-icon` - something else is importing it, which is why this is working, but if whatever that is got removed then this test would stop working so it's good practice to have explicit imports for everything.

PR #4567, opened 2024-04-03
fix: Tag list item: Long items were not showing any tooltip on first focus [GAUD-6139]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6139)
This functionality was broken when working on this ticket (keyboard tooltip should not be shown on hover).

**Problem:**
On first tab focus on a truncated tag-list-item, no tooltip would be displayed (neither keyboard nor truncation tooltip).

**Expected Behaviour:**
On first tab focus, the keyboard instructions tooltip displays. On second focus, the truncation tooltip displays.

**Solution Notes:**
It turned out to be a bit of a timing issue. The truncation tooltip would be rendered on initial render. On focus, the keyboard tooltip would be rendered, but then it seems it would be quickly closed and the truncation tooltip would be re-rendered.


components/tag-list/tag-list-item-mixin.js
```diff
@@ -191,11 +191,11 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 
 				announce(message);
 			}
+			this._onFocusIn();
 
 			await this.updateComplete;
 			// delay the focus to allow focusin to fire
 			setTimeout(() => {
-				this._onFocusIn();
```
bearfriend, line 198: This might break something else I recently worked on but I'll have to refresh my memory on what it was exactly.

margaree: Sounds good - let me know if you think of it. I only recently added `_onFocusIn` so it might post-date the problems you worked on.

bearfriend: All clear. I was mixing this up with the [parent](https://github.com/BrightspaceUI/core/pull/4498/files#diff-2589d7434564f559d1ad28a1e94699c08b1d183ad460f77275d0d6fa1bd903ddR422) `focusIn` I added, which is what the comment is referring too. Maybe we should make that clearer.

bearfriend (approved):

PR #4578, opened 2024-04-08
Code snippet comment colour should pass 4.5:1 contrast requirements in light mode

grant-cleary:
The light mode colour for code comments within our code snippets currently fails AA contrast requirements when put up against the background. I chatted with Jeff, and we're going to darken it from chromite to galena to meet contrast requirements.

Technically in dark mode, we already meet contrast requirements, but Jeff suggested we use the opportunity to bring the comment colour there on palette, so I'm also lightening that up a bit to chromite.


dbatiste (approved):
Thanks!

PR #4580, opened 2024-04-08
GAUD-6124 Add new interim styles for table header

saifaldin14:
When there are multiple elements inside a table header it will apply the interim styles outlined by Nicole and doesn't apply the existing table header styles. It will remove the existing styles with the underline and replace it with a small hover background instead.

Interim styles for button + another element:
Screenshot (146)


components/table/table-wrapper.js
```diff
@@ -392,6 +396,12 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		});
 	}
 
+	_checkSelectIconButton(r) {
+		const header = Array.from(r.childNodes).find((element) => element.localName === 'th');
+		const selectAllButton = Array.from(header.childNodes).find((element) => element.localName === 'd2l-selection-select-all');
+		r.classList.toggle('d2l-table-select-icon', selectAllButton);
+	}
+
```
margaree, line 404: Since this is a separate story/change, you'll likely want to make it in a separate PR for potentially at least a couple reasons:
- If there are issues with it but not with the other work, it ends up blocking
- If both changes result in visual diff changes, we can't isolate which change caused the change and if the changes that change specifically caused were what we intended

Related to this change:
- You should look into if we can just handle this with negative margins instead rather than adding another javascript loop-through
- I'd recommend thinking through the other cases with table that would also need this change

margaree (commented):
I'd recommend doing additional testing here. I took a look at the test page and the table row heights seem very inconsistent, when we'd want them all (including light) to be 48px (ignoring the checkbox piece for now)

components/table/table-col-sort-button.js
```diff
@@ -67,13 +67,24 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 			button::-moz-focus-inner {
 				border: 0;
 			}
+			button:hover {
+				background-color: var(--d2l-color-gypsum);
+			}
+			:host([has-sibling]) button {
+				--d2l-table-cell-padding: 0;
```
margaree, line 74: Does this do anything?

components/table/demo/table-test.js
```diff
@@ -147,6 +147,19 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
...
+			<th scope="col">
+				<d2l-table-col-sort-button nosort>
+					${fruit1}
+				</d2l-table-col-sort-button>
+				<d2l-table-col-sort-button nosort>
+					${fruit2}
+				</d2l-table-col-sort-button>
+			</th>
+		`;
+	}
```
margaree, line 161: It would be useful if this were more functional in order to be able to test the impact of clicking the button.

components/table/table-col-sort-button.js
```diff
@@ -67,13 +67,24 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+				margin-block: -0.3rem -0.3rem;
+				margin-inline: -0.3rem 0.3rem;
+				padding: 0.3rem;
+			}
 			:host([has-sibling]) button:hover {
-				text-decoration: underline;
+				border-radius: 0.2rem;
 			}
 			:host(:not([has-sibling])) button {
+				border-radius: var(--d2l-sortable-button-border-radius);
```
margaree, line 84: Was there a case that caused this to need to move here?

saifaldin14: The spec for the interim styles have a border radius of 0.2rem but in default table the border radius should be 0

PR #4581, opened 2024-04-09
Restore skipped vdiffs due to Chrome 125 bug (GAUD-6232)

dlockhart:
Reverts the tests that were skipped in #4570 due to [this defect in Chrome 124](https://issues.chromium.org/issues/333414300). When & if that defect is resolved, this PR should start passing again and we can merge it, restoring the tests.


dlockhart:
To pass this is going to need Chromium >= `124.0.6367.49`, which will need `puppeteer-core` > `1.43.1`.

dlockhart:
Alright, we finally have a new version of Playwright that includes new Chrome with the bug fixed, so these tests can come back!

dbatiste (approved):

PR #4587, opened 2024-04-10
[d2l-meter] property for hiding text 

abcotter:
## Context

- circle meter took in a `text` property like the radial meter, however it only assigned it to the aria text. We'd like the circle meter to be able to behave similarly to the radial meter
- https://d2l.slack.com/archives/C0PHG3QB0/p1712758381367159

## Implementation Details

- add `text-hidden` prop to meter-mixin
- begin using the new prop to dynamically inject text beneath the circle meter

### DIFF
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -13,15 +13,12 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 		return [ bodySmallStyles, bodyStandardStyles, css`
 		:host {
 			display: inline-block;
-			width: auto;
+			width: 2.4rem;
 		}
 		.d2l-meter-circle {
 			display: flex;
 			flex-direction: column;
 			justify-content: center;
-		}
-		:host(:not([style*=width])) .base-width {
-			width: 2.4rem;
 		}
 		.d2l-meter-circle-full-bar,
 		.d2l-meter-circle-progress-bar {
@@ -63,6 +60,9 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 	}
 
 	render() {
+		if (this.diameter) {
+			this.style.setProperty('width', `${this.diameter}${this.diameterUnits}`);
+		}
 		const lengthOfLine = 21 * Math.PI * 2; // approximation perimeter of circle
 		const percent = this.max > 0 ? (this.value / this.max) : 0;
 		const visibility = (percent < 0.005) ? 'hidden' : 'visible';
@@ -79,18 +79,12 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 			'd2l-meter-circle-text': true
 		};
 
-		// For circle meters that will not have text displayed we need to add a basic width to the svg meter (2.4rem)
-		// In the css, if the host level has has the width declared for the meter then we will stop the 2.4rem width from being applied to the svg
-		const svgClasses = {
-			'base-width': (!this.text || !!this.textHidden)
-		};
-
 		return html`
 			<div
 				class="d2l-meter-circle"
 				aria-label="${this._ariaLabel(primary, secondary)}"
 				role="img">
-				<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision" class="${classMap(svgClasses)}">
+				<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision">
 					<circle class="d2l-meter-circle-full-bar" cx="24" cy="24" r="21"></circle>
 					<circle
 						class="d2l-meter-circle-progress-bar"
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -83,7 +83,7 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 		// In the css, if the host level has has the width declared for the meter then we will stop the 2.4rem width from being applied to the svg
 		const svgClasses = {
 			'base-width': (!this.text || !!this.textHidden)
-		}
+		};
 
 		return html`
 			<div
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -13,12 +13,15 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 		return [ bodySmallStyles, bodyStandardStyles, css`
 		:host {
 			display: inline-block;
-			width: 2.4rem;
+			width: auto;
 		}
 		.d2l-meter-circle {
 			display: flex;
 			flex-direction: column;
 			justify-content: center;
+		}
+		:host(:not([style*=width])) .base-width {
+			width: 2.4rem;
 		}
 		.d2l-meter-circle-full-bar,
 		.d2l-meter-circle-progress-bar {
@@ -76,12 +79,18 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 			'd2l-meter-circle-text': true
 		};
 
+		// For circle meters that will not have text displayed we need to add a basic width to the svg meter (2.4rem)
+		// In the css, if the host level has has the width declared for the meter then we will stop the 2.4rem width from being applied to the svg
+		const svgClasses = {
+			'base-width': (!this.text || !!this.textHidden)
+		}
+
 		return html`
 			<div
 				class="d2l-meter-circle"
 				aria-label="${this._ariaLabel(primary, secondary)}"
 				role="img">
-				<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision">
+				<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision" class="${classMap(svgClasses)}">
 					<circle class="d2l-meter-circle-full-bar" cx="24" cy="24" r="21"></circle>
 					<circle
 						class="d2l-meter-circle-progress-bar"
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -1,7 +1,6 @@
 import '../colors/colors.js';
-import { bodySmallStyles } from '../typography/styles.js';
+import { bodySmallStyles, bodyStandardStyles } from '../typography/styles.js';
 import { css, html, LitElement, nothing } from 'lit';
-import { bodyStandardStyles } from '../typography/styles.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { MeterMixin } from './meter-mixin.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
--- a/components/meter/meter-circle.js
+++ b/components/meter/meter-circle.js
@@ -81,8 +81,7 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 
 		return html`
 			<div
-				class="${(this.text && !this.textHidden) ? 'fit-content ' : ''}d2l-meter-circle"
-				role="img"
+				class="d2l-meter-circle"
 				aria-label="${this._ariaLabel(primary, secondary)}">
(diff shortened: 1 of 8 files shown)


components/meter/meter-radial.js
```diff
@@ -61,7 +61,7 @@ class MeterRadial extends MeterMixin(RtlMixin(LitElement)) {
 		const progressFill = percent * lengthOfLine;
 		const primary = this._primary(this.value, this.max);
 		const secondary = this._secondary(this.value, this.max, this.text);
-		const secondaryTextElement = this.text ? html`<div class="d2l-body-small d2l-meter-radial-text">${secondary}</div>` : nothing;
+		const secondaryTextElement = (this.text && !this.textHidden) ? html`<div class="d2l-body-small d2l-meter-radial-text">${secondary}</div>` : nothing;
```
abcotter, line 64: we shouldn't need to update any uses of the radial meter

components/meter/meter-circle.js
```diff
@@ -41,6 +50,14 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 		:host([foreground-light]) .d2l-meter-circle-text {
 			fill: white;
 		}
+		.d2l-meter-circle-text-secondary {
+			color: var(--d2l-color-ferrite);
+			font-size: 0.6rem;
```
dbatiste, line 54: Looks like `font-size: 0.6rem` should already be applied by the `d2l-body-small` on this element.

components/meter/meter-circle.js
```diff
@@ -57,26 +74,33 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 
 		const primary = this._primary(this.value, this.max) || '';
 		const secondary = this._secondary(this.value, this.max, this.text);
+		const secondaryTextElement = (this.text && !this.textHidden) ? html`<div class="d2l-body-small d2l-meter-circle-text-secondary">${secondary}</div>` : nothing;
```
abcotter, line 77: there are 3 uses where we need to add text-hidden before we merge this https://d2l.slack.com/archives/C0PHG3QB0/p1712758960144389?thread_ts=1712758381.367159&cid=C0PHG3QB0

components/meter/meter-mixin.js
```diff
@@ -21,6 +21,11 @@ export const MeterMixin = superclass => class extends LocalizeCoreElement(superc
 			 * @type {string}
 			 */
 			text: { type: String },
+			/**
+			 * Displays the text as a label beneath the meter
+			 * @type {boolean}
+			 */
+			textHidden: { type: Boolean, attribute: 'text-hidden' },
```
dbatiste, line 28: It looks like `d2l-meter-circle` makes use of this attribute in the CSS, so we'd want to specify `reflect: true` here.

components/meter/meter-circle.js
```diff
@@ -57,26 +74,33 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
...
-			<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision" role="img" aria-label="${this._ariaLabel(primary, secondary)}">
-				<circle class="d2l-meter-circle-full-bar" cx="24" cy="24" r="21"></circle>
-				<circle
-					class="d2l-meter-circle-progress-bar"
-					cx="24" cy="24" r="21"
-					stroke-dasharray="${progressFill} ${space}"
-					stroke-dashoffset="${dashOffset}"
-					visibility="${visibility}"></circle>
+			<div
+				class="${(this.text && !this.textHidden) ? 'fit-content ' : ''}d2l-meter-circle"
```
dbatiste, line 85: This class seems kinda weird. I don't see it specified anywhere. 🤔 

The semantics here seem a little weird - we have an `img` inside of an `img`. Is there a way to specify the aria-label on the actual image (svg), so that the div is simply used for layout?

components/meter/meter-mixin.js
```diff
@@ -21,6 +21,11 @@ export const MeterMixin = superclass => class extends LocalizeCoreElement(superc
 			 * @type {string}
 			 */
 			text: { type: String },
+			/**
+			 * Displays the text as a label beneath the meter
```
dlockhart, line 25: Actually, the inverse? You'll probably want to adjust the wording for `text` to be "Displayed beneath the meter. If the text contains..." and then this one can be something like `Hides the text`.

components/meter/meter-circle.js
dlockhart: Well... this is just copying what `meter-radial` is already doing so they're consistent. But yes, your concerns still seem valid.

abcotter: The reason we need the "fit-content" set for when there is text to display is because the text may riun over the 2.4rem width set int he host

dbatiste: Looks like in the `meter-radial` case, it doesn't specify `role="img"` on the `svg`. Not sure if there is a meaningful difference there or not.

dbatiste: >The reason we need the "fit-content" set for when there is text to display is because the text may riun over the 2.4rem width set int he host

I don't see where that `fit-content` class is defined in the CSS. Maybe that was missed from checkin?

dlockhart: I don't think `fit-content` is doing what you think here -- it's putting it as a `class` but there's no `fit-content` CSS class defined. You have a separate CSS rule that's being applied and is likely accomplishign this for you:

```css
:host([text]:not([text-hidden])) {
  width: fit-content;
}
```

dlockhart: I agree with Dave B here -- I think the `role=img` can likely be removed from the `svg` here so that they match. But I'd want to do some accessibility testing here to make sure it's reading everything out as you'd expect.

abcotter: oh sorry my bas - yes that class assignment should have been removed for the dynamic css selector stuff

components/meter/meter-circle.js
```diff
@@ -10,11 +10,19 @@ import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
...
 class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
 	static get styles() {
-		return [ bodyStandardStyles, css`
+		return [ bodySmallStyles, bodyStandardStyles, css`
 		:host {
 			display: inline-block;
 			width: 2.4rem;
 		}
+		:host([text]:not([text-hidden])) {
+			width: fit-content;
```
dbatiste, line 19: One more thing to highlight re `fit-content`: it wasn't added to FF until v94 ([CanIUse](https://caniuse.com/?search=fit-content)), which is in our "legacy" browser support band. Maybe we're "ok" with the visual not being ideal for those users, but figured I would call it out is case we're not.

components/meter/meter-circle.js
```diff
@@ -57,26 +72,32 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
...
-				<circle
-					class="d2l-meter-circle-progress-bar"
-					cx="24" cy="24" r="21"
-					stroke-dasharray="${progressFill} ${space}"
-					stroke-dashoffset="${dashOffset}"
-					visibility="${visibility}"></circle>
+			<div
+				class="d2l-meter-circle"
+				aria-label="${this._ariaLabel(primary, secondary)}">
+				<svg viewBox="0 0 48 48" shape-rendering="geometricPrecision" role="img">
```
dlockhart, line 86: As Dave B suggested in the now-outdated comments, I think this `role="img"` needs to move to the `div` to align with how `meter-circle` is doing it. Otherwise the `aria-label` on the `div` isn't doing anything.

components/meter/meter-circle.js
dlockhart: Abi, what's the visual result of this not working? Just curious how bad it would look.

components/meter/meter-mixin.js
```diff
@@ -21,6 +21,11 @@ export const MeterMixin = superclass => class extends LocalizeCoreElement(superc
 			 * @type {string}
 			 */
 			text: { type: String },
+			/**
+			 * Hides the text from being displayed beneath the meter
+			 * @type {boolean}
+			 */
+			textHidden: { type: Boolean, attribute: 'text-hidden', reflect: true },
```
dlockhart, line 28: Looks like `meter-linear` also renders a secondary text element and extends this mixin -- can you apply the same treatment there?

dlockhart, line 28: Also, can you add new vdiffs to the 3 meters that exercise this new property?

components/meter/meter-circle.js
abcotter: Without the fit-content the text isn't nicely centred under the meter - maybe there is another way to resolve this that I'm unaware of? 
Screenshot 2024-04-10 at 3 53 33 PM

dlockhart: Haha what's going on with the other circles in that screenshot? If they're unrelated, then this seems like a fine trade-off for older browsers. As long as it's not totally broken, it's fine.

dbatiste: > Haha what's going on with the other circles in that screenshot?

Haha, I think that's just the adjacent meter components included in the screenshot of the demo page.

> Without the fit-content the text isn't nicely centred under the meter - maybe there is another way to resolve this that I'm unaware of?

`d2l-meter-radial` has the same issue. The text simply overflows.

What I observed with `fit-content`: `svg` scales based on the text size. As a result, the same meter would have different size for different languages, or adjacent meters with different text would be different size. That's probably not the behaviour we want.  
image


I managed to get something sorta working by setting `width: auto` on the host, moving the baked `width: 2.4rem` to the `svg` (from the host), and setting `margin: 0 auto` on the `svg`. But then the text potentially pushes them further apart.
image


But currently, the consumer can set the size by specifying `<d2l-meter-circle style="width: 25%;">`. Moving the `width` to the internal `svg` breaks that behaviour, and so we'd need to do something else to allow them to control the size of the `svg`. Maybe a CSS variable or a property. Nevertheless, it things seem possible, but it's a change in behaviour.

abcotter: used a combo of `classMap` and attribute selectors to get the `width: 2.4rem` appearing on meters without text/ with text-hidden AND don't have a style for width set on the host level :) 

looks like it's working well - I'll add some inline notes for future devs to make it clear what we're doing

Screenshot 2024-04-11 at 8 49 10 AM

components/meter/meter-circle.js
```diff
@@ -57,26 +72,39 @@ class MeterCircle extends MeterMixin(RtlMixin(LitElement)) {
...
 			'd2l-meter-circle-text-ltr': !this.percent,
 			'd2l-body-standard': true,
 			'd2l-meter-circle-text': true
 		};
 
+		// For circle meters that will not have text displayed we need to add a basic width to the svg meter (2.4rem)
+		// In the css, if the host level has has the width declared for the meter then we will stop the 2.4rem width from being applied to the svg
+		const svgClasses = {
+			'base-width': (!this.text || !!this.textHidden)
+		}
```
abcotter, line 86:
```suggestion
		};
```

components/meter/meter-circle.js
abcotter: https://github.com/BrightspaceUI/core/pull/4587/files#diff-4df82a82ac6364e1408ebd4ce8b7a8ede6a5f0c0e4bb96d0e4f005ebf9f50f33R82-R86

dbatiste: It looks like the size of the `svg` would vary based on the size of the text. We can check with Jeff, but my current feeling is that we don't want that behaviour as mentioned above.
image

dbatiste:
Any objection if we close this PR?

dbatiste:
Closing. We can revisit this in the future if required.

PR #4590, opened 2024-04-11
GAUD 5465: use "slider" ARIA role and only render divider when resizable

dlockhart:
This addresses two problems with the primary/secondary template:

The first is that when the panels aren't resizable, the divider was still there and discoverable to assistive technology. In this scenario, I've updated it to render a "readonly" version of the divider, also known as "a grey line".

While fixing the first thing, I discovered that the "separator" ARIA role we were using for the divider simply wasn't working as expected. It seems like support for it was never really ironed out or implemented widely. To address this, we've converted it to a "slider" role. We've also:

- Added a "secondary panel" label to the `<aside>` element
- Removed the keyboard instructions for using the divider since it's a standard ARIA thing and simplified its label to "Secondary panel divider" which matches the label on the `<aside>`
- Changed the scale of the divider max/min/value to be percents instead of arbitrary pixel values
- Set `aria-valuetext` to `<value> %` so it reads out as a percent instead of a number

### DIFF
--- a/templates/primary-secondary/primary-secondary.js
+++ b/templates/primary-secondary/primary-secondary.js
@@ -4,9 +4,8 @@ import '../../components/icons/icon.js';
 import '../../components/offscreen/offscreen.js';
 import { css, html, LitElement, unsafeCSS } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { formatPercent } from '@brightspace-ui/intl';
 import { getFocusPseudoClass } from '../../helpers/focus.js';
-import { getUniqueId } from '../../helpers/uniqueId.js';
-import { ifDefined } from 'lit/directives/if-defined.js';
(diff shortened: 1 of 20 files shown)


templates/primary-secondary/primary-secondary.js
```diff
@@ -591,7 +590,8 @@ class TemplatePrimarySecondary extends RtlMixin(LocalizeCoreElement(LitElement))
 			_isCollapsed: { type: Boolean, attribute: false },
 			_isExpanded: { type: Boolean, attribute: false },
 			_isMobile: { type: Boolean, attribute: false },
-			_size: { type: Number, attribute: false }
+			_size: { type: Number, attribute: false },
+			_sizeAsPercent: { state: true }
```
dlockhart, line 594: I needed to make this a reactive state property so that when it gets changed when the window resizes it triggers a re-render.

dbatiste (approved):

PR #4597, opened 2024-04-12
GAUD-6276 Udate table header style height when icon is inside

saifaldin14:
[Jira](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6276)

When the select all or input checkbox icon is in the first row of the table, the dimensions of the element won't affect the 48px height of the header cell.

The height of the first row will be 48.6px in default because of the added 0.8px border on the first row. In the table light mode the height will remain unchanged at 48px.

Screenshot (155)


margaree (commented):
I'm wondering if it would be possible to this more as a `calc` that relies on variables rather than just a single number - the single number makes it a bit hard to tell where it comes from, since it's not in the spec, and if something changes this wont necessarily.

I'm also wondering about other cases - what's the expectation if there's a button icon in the cell? Are there cases that use a regular checkbox input rather than this select all component?

Also, did you end up creating that separate story to handle the heights of the regular table rows/cells that aren't headers? The story title itself for this PR is "Update icon button padding" with no further detail so I'm not totally sure if that should encompass other rows as well.

components/table/table-wrapper.js
```diff
@@ -62,6 +63,15 @@ export const tableStyles = css`
 	th.d2l-table-header-col-sortable {
 		padding: 0;
 	}
+	.d2l-checkbox,
+	d2l-input-checkbox,
+	d2l-selection-select-all,
+	input[type='checkbox'] {
+		margin-block: calc(-1 * ${cssSizes.inputBoxSize}rem); /* The height of an input checkbox */
```
margaree, line 70: This would be setting the top and bottom margin to each be `-24px`. Is that what we want?
We can likely make a smarter equation here. We have the following information:
- Height we want (48px)
- Checkbox height (24px)
- Padding (0.75rem)
With this information we can calculate how much the top and bottom margins would need to change.

components/table/demo/table-test.js
```diff
@@ -98,7 +99,7 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
 					</thead>
 					<tbody>
 						<tr class="d2l-table-header">
-							<th scope="col" sticky></th>
+							<th scope="col" sticky><d2l-button-icon icon="tier1:gear"></d2l-button-icon></th>
```
margaree, line 102: We generally want to create examples that look like "real world" usages. e.g., for this one, we would want it within one of the header cells with content rather than the checkbox cell.
Also, this doesn't actually have bottom spacing as it is now, only top.

components/table/demo/table-test.js
```diff
@@ -99,7 +100,7 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
 					<tbody>
 						<tr class="d2l-table-header">
 							<th scope="col" sticky></th>
-							${this._renderDoubleSortButton('Avocado', 'Prune')}
+							${this._renderDoubleSortButton('Avocado')}
```
margaree, line 103: We would also still want to keep this scenario. It's an important one to be able to test easily.

components/table/table-wrapper.js
```diff
@@ -62,6 +63,15 @@ export const tableStyles = css`
 	th.d2l-table-header-col-sortable {
 		padding: 0;
 	}
+	.d2l-checkbox,
+	d2l-input-checkbox,
+	d2l-selection-select-all,
+	input[type="checkbox"] {
+		margin-block: calc(-0.5 * (${cssSizes.inputBoxSize}rem - var(--d2l-table-cell-padding)));
```
margaree, line 70: Not sure I'm understanding the reasoning behind this calculation, and the results don't quite seem correct (e.g., on the test page the button icon spacing isn't quite right).

Think more about what you're looking to correct for. 
Without any change (i.e., prior to this PR), the cell height = checkbox height + padding (with medium font-size this is 24px + 15px + 15px = 54px)
We want the cell height to be 48px. So how much less does the padding need to be in order for this to be the case? Is there a way we can consistently calculate that using variables?

components/table/table-wrapper.js
saifaldin14, line 1: I built the equation based on this:

48px = 24px + 2* (15px - x)
0.5(48px - 24px - 2 * 15px) = x
x = -3px

When I did this however, there was an extra pixel added so I have to remove it

margaree: So the extra pixel is likely the border. If you haven't already, confirm with Nicole if she wants the 48px to be interior or including the border.

components/table/table-wrapper.js
```diff
@@ -62,6 +63,16 @@ export const tableStyles = css`
 	th.d2l-table-header-col-sortable {
 		padding: 0;
 	}
+	.d2l-checkbox,
+	d2l-input-checkbox,
+	d2l-selection-select-all,
+	input[type="checkbox"] {
```
margaree, line 69: Looking at the `input[type="checkbox"]` in the vdiffs, it's already quite short so I think we can leave it out at this point.

components/table/table-wrapper.js
```diff
@@ -62,6 +63,16 @@ export const tableStyles = css`
...
+		/* Get the margin required to make the height 48px based on the checkbox height */
+		margin-block: calc(0.5 * (48px - ${cssSizes.inputBoxSize}rem - 2 * var(--d2l-table-cell-padding)) - 1px);
```
margaree, line 71: It would probably be helpful to use a css var for the `48px`. Then we can also use that when we update the heights of the regular cells.

margaree (commented):
After the few comments have been addressed, this LGTM

components/table/table-wrapper.js
saifaldin14: Nicole said 48px should be internal and not include the border

margaree (approved):

PR #4599, opened 2024-04-12
fix: don't validate disconnected nested forms (GAUD-6271)

dlockhart:
@rileyodonnelld2l noticed this issue. The scenario: they had a form which contained a HTML Editor, which could launch the "insert link" dialog which contained a nested form that "joined" the outer form (aside: it probably should have had `no-nesting` on it since joining the parent form is weird, but regardless). Even after the insert link dialog was closed and it was totally removed from the DOM, the outer form would still include it in its validation.

The reason for this is that the logic that handles disconnected nested forms never worked, and this fell into a special edge case that required it to work.

The main reason for that was a mismatch between the event being dispatched (`d2l-form-disconnected`) and the event being listened for (`d2l-form-disconnect`). Beyond that though, the logic to remove the nested form was still flawed, because it was assuming the target of the event would match the nested form itself, but if the nested form was inside a custom element, the event would be re-targeted to the custom element.

And then on top of all that, if a custom element contained multiple nested forms it couldn't handle that and would always just remember the last one that connected.

### DIFF
--- a/components/form/form.js
+++ b/components/form/form.js
@@ -139,10 +139,14 @@ class Form extends FormMixin(LitElement) {
 		this._nestedForms.set(e.target, form);
 
 		const onFormDisconnect = () => {
-			form.removeEventListener('d2l-form-disconnect', onFormDisconnect);
-			this._nestedForms.delete(e.target);
+			this._nestedForms.forEach((value, key) => {
+				if (value === form) {
+					form.removeEventListener('d2l-form-disconnected', onFormDisconnect);
+					this._nestedForms.delete(key);
+				}
+			});
 		};
-		form.addEventListener('d2l-form-disconnect', onFormDisconnect);
+		form.addEventListener('d2l-form-disconnected', onFormDisconnect);
 	}
 
 	async _submitData(submitter) {
--- a/components/form/form.js
+++ b/components/form/form.js
@@ -51,7 +51,7 @@ class Form extends FormMixin(LitElement) {
 	disconnectedCallback() {
 		super.disconnectedCallback();
 		/** @ignore */
-		this.dispatchEvent(new CustomEvent('d2l-form-disconnected'));
+		this.dispatchEvent(new CustomEvent('d2l-form-disconnect'));
 		this._isSubForm = false;
 	}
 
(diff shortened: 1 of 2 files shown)


components/form/form.js
```diff
@@ -139,10 +139,14 @@ class Form extends FormMixin(LitElement) {
 		this._nestedForms.set(e.target, form);
 
 		const onFormDisconnect = () => {
-			form.removeEventListener('d2l-form-disconnect', onFormDisconnect);
-			this._nestedForms.delete(e.target);
+			this._nestedForms.forEach((value, key) => {
```
dlockhart, line 142: Hmm... writing out a description here, I think this still might be slightly flawed. Kicking this back to draft while I investigate.

dbatiste: Cool... I was just trying to follow this.

Is it true that the `const form = e.composedPath()[0];` above is the actual form element? And that `e.composedPath()[0]` of the `d2l-form-disconnected` event would be the same element?

It seems like the original code would have failed if there was more than one nested form too.

dlockhart: Yeah exactly -- if there were more than 1 nested form inside a custom element, then the existing code would only know about the last one, since its `Map` is just from `e.target` (in this case the custom element) to form.

I'm going to rework this so that the `Map` contains an array of nested forms, and I think I can just have a single disconnect listener.

dbatiste: Ok, sounds good. Yeah I was wondering if we really wanted multiple handlers wired up.

The alternative I was thinking before you mentioned that was something roughly along these lines...
```javascript
const onFormDisconnect = disconnectedEvent => {
    const disconnectedForm = disconnectedEvent.composedPath()[0];
    if (disconnectedForm === form) {
        ...
    }
}
```

components/form/form.js
```diff
@@ -136,13 +140,27 @@ class Form extends FormMixin(LitElement) {
 		e.stopPropagation();
 		e.preventDefault();
 		const form = e.composedPath()[0];
-		this._nestedForms.set(e.target, form);
```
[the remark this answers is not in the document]
dbatiste: Would we still need to use an array if the key was `form`, rather than `e.target`?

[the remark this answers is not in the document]
dlockhart: It would definitely affect how the validation & submission happens, since `_findFormElements` uses the presence of an element in this DOM scope (`e.target` here) to add it to the list of things that needs validation/submitting.

dbatiste (approved):

PR #4601, opened 2024-04-15
fix: When all items of a filter set are deleted, move focus to the next filter set

saifaldin14:
[Jira](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6132)

Problem:
When all items of a certain filter set are deleted the focus breaks away from the filter tags and goes to the top of the page. This is happening because the active filter registries contain all of the tags of a certain set and when they become 0 we still keep the empty set in the active registries.

Proposed solution:
In the `_tagListItemDeleted` method when a filter tag is deleted, if the filter set in which it belongs to is length 1 (that means that the current item to be deleted is the last tag in the set) then we delete that registry all together and not keep the empty registry. This correctly shifts the focus to the next set without breaking away.

Demo:

https://github.com/BrightspaceUI/core/assets/24593380/1b7b987f-6465-4a26-9a87-3104125eb04e


margaree (commented):
I don't totally know enough about filters to know if this is a non-ideal change/I can't remember if this was done on purpose, however in testing I noticed that previously order was preserved of filters (i.e., if I deleted all of filter 1's filters, then selected any of them, they would come before filter 2's) but this is no longer the case with this change.

I did some digging into this and noticed that in [`tag-list`](https://github.com/BrightspaceUI/core/blob/main/components/tag-list/tag-list.js#L354) if I printed all the children's text after the `setTimeout` there would be a duplicate of the last item of the first filter. I'd recommend taking a look at that - it seems unexpected.

I'd also recommend adding a simple visual diff test for this if possible.

components/tag-list/tag-list.js
```diff
@@ -346,13 +346,10 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
...
 		const rootTarget = e.composedPath()[0];
-		const children = this._getVisibleEffectiveChildren();
-		const itemIndex = children.indexOf(rootTarget);
-
-		if (children.length <= 1) return;
+		const itemIndex = this._items.slice().indexOf(rootTarget);
 
 		setTimeout(() => {
-			const focusableElem = children[itemIndex - 1] || (children[itemIndex] === e.target ? children[itemIndex + 1] : children[itemIndex]);
+			const focusableElem = this._items[itemIndex - 1] || this._items[0];
```
margaree, line 352: I don't know if this can be simplified this much - see https://github.com/BrightspaceUI/core/pull/4498#discussion_r1517732635 for more context. @bearfriend wondering if you have thoughts on this?

components/tag-list/tag-list.js
```diff
@@ -346,13 +346,10 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
 
 	_handleItemDeleted(e) {
 		const rootTarget = e.composedPath()[0];
-		const children = this._getVisibleEffectiveChildren();
-		const itemIndex = children.indexOf(rootTarget);
-
-		if (children.length <= 1) return;
+		const itemIndex = this._items.slice().indexOf(rootTarget);
```
margaree, line 349: What would be the advantage of using `slice` here?

margaree (commented):
I would still recommend adding a visual diff test for this case if possible.

components/filter/test/filter-tags.vdiff.js
```diff
@@ -136,6 +136,19 @@ describe('filter-tags', () => {
...
 		});
 
+		it('clear-one-filter-set', async() => {
+			const elem = await fixture(tagsTwoFiltersMoreSelected, { viewport: { width: 1700 } });
+
+			for (let i = 0; i < 2; i++) {
+				await sendKeysElem(elem.querySelector('d2l-filter-tags'), 'press', 'Tab');
+				await sendKeys('press', 'ArrowRight');
+				sendKeys('press', 'Delete');
+			}
```
margaree, line 146: I don't know that this test is necessarily testing what we want. Based on looking at `tagsTwoFiltersMoreSelected` it looks like only one filter value is selected from `filter-1` so all we'd want to do is delete the first item and confirm that focus goes to the second item.

components/tag-list/tag-list.js
```diff
@@ -346,13 +346,10 @@ class TagList extends LocalizeCoreElement(InteractiveMixin(ArrowKeysMixin(LitEle
...
 
 		setTimeout(() => {
-			const focusableElem = children[itemIndex - 1] || (children[itemIndex] === e.target ? children[itemIndex + 1] : children[itemIndex]);
+			const focusableElem = this._items[itemIndex - 1] || this._items[0] || (this._items[itemIndex] === e.target ? this._items[itemIndex + 1] : this._items[itemIndex]);
```
margaree, line 352: This logic looks a bit different than it was previously. Is there a different case that's trying to tackle?

margaree (approved):

PR #4602, opened 2024-04-15
Add experimental support for MML3 features within MathJax (including mlongdiv)

grant-cleary:
MathJax offers experimental support for some MathML elements via an mml3 extension. This gives us at least some support for symbols like long division that we wouldn't get otherwise (stock MathJax yields a math input error if you try to use, say, `<mlongdiv>`. It's likely ideal for us to support this rather than not.

### DIFF
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -47,6 +47,7 @@ class HtmlBlockMathRenderer {
 
 		const mathJaxConfig = {
 			deferTypeset: true,
+			enableMML3Support: context.enableMML3Support,
 			renderLatex: context.renderLatex,
 			outputScale: context.outputScale || 1,
 			window: window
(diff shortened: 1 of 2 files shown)


helpers/mathjax.js
```diff
@@ -47,6 +47,7 @@ class HtmlBlockMathRenderer {
 
 		const mathJaxConfig = {
 			deferTypeset: true,
+			enableMML3Support: context.enableMML3Support,
```
grant-cleary, line 50: This doesn't exist yet, but I'll be creating a feature flag in the LMS to provide it. I'm uncomfortable enabling anything experimental without at least having the ability to turn it off quickly.

helpers/mathjax.js
```diff
@@ -102,7 +103,15 @@ export function loadMathJax(mathJaxConfig) {
...
 		},
-		loader: { load: ['ui/menu'] },
+		loader: {
+			load: mathJaxConfig && mathJaxConfig.enableMML3Support
+				? [
+					'[mml]/mml3',
+					'ui/menu'
+				] : [
+					'ui/menu'
+				]
```
grant-cleary, line 113: I don't really know how MathJax will behave if some of these array elements are undefined or some default value. Given that this feature flag should be short lived (a few releases at most), I'm not really concerned about the boilerplate of reproducing this array.

dbatiste (approved):

PR #4604, opened 2024-04-16
fix: further handle nested forms inside custom elements

dlockhart:
This is a follow-up to #4599, as the original fix didn't fully cover the case in Creator. Explanation in line in the code.

### DIFF
--- a/components/form/form.js
+++ b/components/form/form.js
@@ -139,23 +139,25 @@ class Form extends FormMixin(LitElement) {
 		}
 		e.stopPropagation();
 		e.preventDefault();
-		const form = e.composedPath()[0];
 
-		if (!this._nestedForms.has(e.target)) {
-			this._nestedForms.set(e.target, []);
+		const form = e.composedPath()[0];
+		const target = e.target;
+
+		if (!this._nestedForms.has(target)) {
+			this._nestedForms.set(target, []);
 		}
-		this._nestedForms.get(e.target).push(form);
+		this._nestedForms.get(target).push(form);
 
 		const onFormDisconnect = () => {
 			form.removeEventListener('d2l-form-disconnect', onFormDisconnect);
-			if (this._nestedForms.has(e.target)) {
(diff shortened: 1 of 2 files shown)


components/form/form.js
```diff
@@ -139,23 +139,25 @@ class Form extends FormMixin(LitElement) {
 		}
 		e.stopPropagation();
 		e.preventDefault();
+
 		const form = e.composedPath()[0];
+		const target = e.target;
```
dlockhart, line 144: The only change here is the pull out a copy of `e.target` into a local variable, instead of referring to it as `e.target` everywhere. Why? The answer is interesting.

Normally it wouldn't matter, but as the original event bubbles up through the DOM, as soon as it crosses a shadow root boundary, the target changes to the new element to "hide" its internals. So after this initial code runs, `e.target` can be an element further up the DOM, which means looking for it later in the `this._nestedForms` Map isn't going to find anything.

With my original test code this wasn't an issue, since there was only 1 level of custom elements. But the Creator case had multiple layers. I've added a new test case to handle this.

components/form/test/form.test.js
```diff
@@ -5,12 +5,29 @@ import './nested-form.js';
...
+		return html`
+			<d2l-form id="nested-form-1">
+				<input type="text" aria-label="Input 1" name="input1" required>
+			</d2l-form>
+			<d2l-form id="nested-form-2">
+				<input type="text" aria-label="Input 2" name="input2" required>
+			</d2l-form>
+		`;
+	}
+}
```
dlockhart, line 19: I had to pull this up here instead of using `defineCE` because you can't embed things that were also created using `defineCE` in a thing using `defineCE` -- at least I couldn't get it to work.

components/form/form.js
dbatiste: TIL. I think that explain som weird behaviour I've seen before.

dbatiste (approved):

PR #4606, opened 2024-04-16
Fix tooltips used on targets with delegated focus

bearfriend:
Currently, if a tooltip targets an element with delegated focus (i.e. basically everything: `d2l-button`, `d2l-link`, `d2l-input-text`, etc.), that target will never be properly described by the tooltip because once it delegates focus internally, the newly focused element is not described by the tooltip.

Non-interactable target warnings are currently disabled on the constructor after a single warning, so most tooltips are never communicating this issue.

Once that restriction is removed, all focus-delegated targets begin to warn because they are not in `interactiveElements` and do not have a `role` as the delegate should handle both of these.

- Detect elements using `FocusMixin`, use their delegate as the tooltip target (which also prevents the warning)
- Since delegates are always across shadow DOM boundaries, set an `aria-label/description` directly on the target
- Manage `label`, `description`, `labelledby`, and `describedby` appropriately
- Move the old "I don't know, maybe?" warning to a definitely-worded error
- Disable logging the error only once
- Prevent disconnected tooltips from hiding the `activeTooltip`
- Prevents firing `d2l-tooltip-show` on tooltips that were never shown before being hidden by another tooltip
- Read `tag-list-item` truncated text as a label, instead of duplicating it as a description

[GAUD-6266](https://desire2learn.atlassian.net/browse/GAUD-6266): Investigate "d2l-tooltip" non-accessible warnings

[GAUD-6266]: https://desire2learn.atlassian.net/browse/GAUD-6266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/tag-list/tag-list-item-mixin.js
+++ b/components/tag-list/tag-list-item-mixin.js
@@ -264,7 +264,7 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		} else if (options.hasTruncationTooltip || hasDescription) {
 			const tooltipHeader = hasDescription ? html`<div class="d2l-heading-4">${tagContent}</div>` : tagContent;
 			tooltip = html`
-				<d2l-tooltip class="vdiff-target" for="${this._id}" ?show-truncated-only="${!hasDescription}">
+				<d2l-tooltip class="vdiff-target" for="${this._id}" ?show-truncated-only="${!hasDescription}" for-type="${hasDescription ? 'descriptor' : 'label'}">
 					${tooltipHeader}
 					${hasDescription ? options.description : nothing}
 				</d2l-tooltip>`;
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -737,6 +737,8 @@ class Tooltip extends RtlMixin(LitElement) {
 				await customElements.whenDefined(target.localName);
 				await target.updateComplete;
 				target = target?.shadowRoot?.querySelector(target?.constructor.focusElementSelector) ?? target;
+			} else {
+				this.#targetDelegated = false;
 			}
 		} else {
 			console.warn('<d2l-tooltip>: missing required attribute "for"');
@@ -920,9 +922,11 @@ class Tooltip extends RtlMixin(LitElement) {
 			await this.updateComplete;
 			await this.updatePosition();
 			if (dispatch) {
-				this.dispatchEvent(new CustomEvent(
-					'd2l-tooltip-show', { bubbles: true, composed: true }
-				));
+				requestAnimationFrame(() => {
+					this.dispatchEvent(new CustomEvent(
+						'd2l-tooltip-show', { bubbles: true, composed: true }
+					));
+				});
 			}
 
 			if (this.announced && !this._isInteractive(this._target)) announce(this.innerText);
@@ -966,10 +970,10 @@ class Tooltip extends RtlMixin(LitElement) {
 
 			if (this.#targetDelegated) {
 				if (this.forType === 'label') {
-					this.removeAttribute('aria-labelledby');
+					this._target.removeAttribute('aria-labelledby');
 					this._target.ariaLabel = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
 				} else if (!this.announced || isInteractive) {
-					this.removeAttribute('aria-describedby');
+					this._target.removeAttribute('aria-describedby');
 					this._target.ariaDescription = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
 				}
 			}
--- a/components/tooltip/tooltip.js
+++ b/components/tooltip/tooltip.js
@@ -10,8 +10,6 @@ import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
-let logAccessibilityWarning = true;
-
 /* only one tooltip is to be shown at once - track the active tooltip so it can be hidden if necessary */
 let activeTooltip = null;
 
@@ -96,7 +94,7 @@ const computeTooltipShift = (centerDelta, spaceLeft, spaceRight) => {
 
 /**
  * A component used to display additional information when users focus or hover on a point of interest.
- * @slot - Default content placed inside of the tooltip
+ * @slot - Default content placed inside of the tooltip. This content is also used as the description for the target element, and may override existing descriptors.
  * @fires d2l-tooltip-show - Dispatched when the tooltip is opened
  * @fires d2l-tooltip-hide - Dispatched when the tooltip is closed
  */
@@ -422,6 +420,8 @@ class Tooltip extends RtlMixin(LitElement) {
 		`];
 	}
 
+	#targetDelegated = false;
+
 	constructor() {
 		super();
 
@@ -487,6 +487,10 @@ class Tooltip extends RtlMixin(LitElement) {
 		if (this._target) {
 			elemIdListRemove(this._target, 'aria-labelledby', this.id);
 			elemIdListRemove(this._target, 'aria-describedby', this.id);
+			if (this.#targetDelegated) {
+				this._target.removeAttribute('aria-label');
+				this._target.removeAttribute('aria-description');
+			}
 		}
 	}
 
@@ -525,7 +529,7 @@ class Tooltip extends RtlMixin(LitElement) {
 				<div class="d2l-tooltip-position" style=${styleMap(tooltipPositionStyle)}>
 					<div class="${classMap(contentClasses)}">
 						<div role="text">
-							<slot></slot>
+							<slot @slotchange="${this._handleSlotChange}"></slot>
 						</div>
 					</div>
 				</div>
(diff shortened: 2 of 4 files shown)


components/tooltip/tooltip.js
```diff
@@ -96,7 +94,7 @@ const computeTooltipShift = (centerDelta, spaceLeft, spaceRight) => {
 
 /**
  * A component used to display additional information when users focus or hover on a point of interest.
- * @slot - Default content placed inside of the tooltip
+ * @slot - Default content placed inside of the tooltip. This content is also used as the description for the target element, and may override existing descriptors.
```
bearfriend, line 97: It will override if the target focus is delegated

bearfriend, line 97: It will override if the target focus is delegated

components/tooltip/tooltip.js
```diff
@@ -931,25 +946,49 @@ class Tooltip extends RtlMixin(LitElement) {
...
-		this._target = this._findTarget();
+
+		if (this._target && this.id) {
+			elemIdListRemove(this._target, 'aria-labelledby', this.id);
+			elemIdListRemove(this._target, 'aria-describedby', this.id);
+			if (this.#targetDelegated) {
+				this._target.removeAttribute('aria-label');
+				this._target.removeAttribute('aria-description');
+			}
+		}
```
bearfriend, line 959: If we already have a target, we're likely updating it, so clean up the existing one.

components/tooltip/tooltip.js
```diff
@@ -931,25 +946,49 @@ class Tooltip extends RtlMixin(LitElement) {
...
-					'\'input\'', '\'select\', \'textarea\' or static / custom elements if a role has been set and the element is focusable.',
+
+			if (this.#targetDelegated) {
+				if (this.forType === 'label') {
+					this.removeAttribute('aria-labelledby');
+					this._target.ariaLabel = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
+				} else if (!this.announced || isInteractive) {
+					this.removeAttribute('aria-describedby');
+					this._target.ariaDescription = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
+				}
```
bearfriend, line 974: When the target has been delegated, we can't use the `...by` properties, which take precedence, so remove them.

bearfriend: Whoops.
```suggestion
				if (this.forType === 'label') {
					this._target.removeAttribute('aria-labelledby');
					this._target.ariaLabel = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
				} else if (!this.announced || isInteractive) {
					this._target.removeAttribute('aria-describedby');
					this._target.ariaDescription = this.shadowRoot.querySelector('slot').assignedNodes().map(n => n.textContent).join('\n');
				}
```

components/tooltip/tooltip.js
```diff
@@ -931,25 +946,49 @@ class Tooltip extends RtlMixin(LitElement) {
...
+			}
+			else {
+				if (this.forType === 'label') {
+					this._target.removeAttribute('aria-label');
+					elemIdListAdd(this._target, 'aria-labelledby', this.id);
+				} else if (!this.announced || isInteractive) {
+					this._target.removeAttribute('aria-description');
+					elemIdListAdd(this._target, 'aria-describedby', this.id);
+				}
```
bearfriend, line 983: Removing the `label`/`description` isn't strictly necessary here, but they won't be used since the `...by` property we're setting takes precedence, and cleaning them up seems good.

components/tag-list/tag-list-item-mixin.js
```diff
@@ -264,7 +264,7 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 		} else if (options.hasTruncationTooltip || hasDescription) {
 			const tooltipHeader = hasDescription ? html`<div class="d2l-heading-4">${tagContent}</div>` : tagContent;
 			tooltip = html`
-				<d2l-tooltip class="vdiff-target" for="${this._id}" ?show-truncated-only="${!hasDescription}">
+				<d2l-tooltip class="vdiff-target" for="${this._id}" ?show-truncated-only="${!hasDescription}" for-type="${hasDescription ? 'descriptor' : 'label'}">
```
bearfriend, line 267: Prevents reading out the truncated contents twice.

components/tooltip/tooltip.js
```diff
@@ -892,19 +909,22 @@ class Tooltip extends RtlMixin(LitElement) {
 	}
 
 	async _showingChanged(newValue, dispatch) {
+		if (!this.isConnected) return;
```
bearfriend, line 912: I'm not sure why this only started being necessary now, but `input-date-range` has two layers of validation tooltips, and the upper layer (which is suppressed if there are child validation errors), was showing itself and hiding the child error before then removing itself from the DOM to allow the child tooltip to show, but that one had already been hidden, so we ended up with none. This simply bails if the tooltip that called it is no longer connected to the DOM.

components/tooltip/tooltip.js
```diff
@@ -892,19 +909,22 @@ class Tooltip extends RtlMixin(LitElement) {
...
-
 			this._dismissibleId = setDismissible(() => this.hide());
 			this.setAttribute('aria-hidden', 'false');
 			await this.updateComplete;
-			await this.updatePosition();
-			if (dispatch) {
+
+			if (this._showing && dispatch) {
+				await this.updatePosition();
 				this.dispatchEvent(new CustomEvent(
```
bearfriend, line 928: If, after the update, another tooltip has become active (and hidden this one), don't fire the `d2l-tooltip-show` event since it never actually showed. Also, don't both updating the position since it can and should be recalculated on the next `show`

components/tooltip/tooltip.js
dlockhart: So if I'm reading this correctly, there's no way for the downstream code (e.g. `d2l-button`) to leverage `aria-labelledby`/`aria-decribedby` to set a label/description themselves on their `<button>` and then have a consumer point a tooltip at them... since that would fall into the `targetDelegated` case which would then remove their label/description?

The only way it could work would be if `<d2l-button>` itself rendered the tooltip, in which case their in the same scope and therefore _not_ delegated so the tooltip `id` would get added to the list.

bearfriend: Right, I couldn't find a convenient way to get the computed `label`/`description`, besides by just getting the element via the id and grabbing the text content, which seems like it's just as well to handle once in the tooltip code rather than duplicating that in every delegated-focus component.

Obviously we will someday be able to use the ElementInternals/AOM stuff, but until then this seems better than having nothing.

> The only way it could work would be if <d2l-button> itself rendered the tooltip, in which case they're in the same scope and therefore not delegated so the tooltip id would get added to the list.

Yes, that's exactly right. I think there could be some much more complicated solutions, but I'm not sure it's worth the effort.

dlockhart: Cool, makes sense! I thought of an alternate approach last night... so as a thought experiment:

What if the components themselves opted into this using something like a `TooltipTargetMixin`? That mixin could come with some reactive state properties like `tooltipValue` and `tooltipType`, and the tooltip code would set those whenever the tooltip value changed or it was shown/hidden.

That would then put the responsibility on the target to figure out what to actually do with the value and what element to put it on -- using `aria-label/description`, maybe an offscreen element hooked up with `aria-labelledby` if there are multiple labels, etc.

The advantage to this approach is that it empowers the target element to "do what's best" and hooks into Lit's lifecycle to do the rendering. Tooltip could even `throw` if the target didn't implement the mixin. It's obviously more work though for each one to opt in.

My worry with the more "automatic" approach is that consumers can accidentally mess up a core component's accessibility by unknowingly completely overriding its label. But maybe that's what we want anyway... if you're pointing a tooltip at a button or whatever, you probably want its value to become the label. 🤷

components/tooltip/tooltip.js
```diff
@@ -931,25 +951,49 @@ class Tooltip extends RtlMixin(LitElement) {
...
-				elemIdListAdd(this._target, 'aria-describedby', this.id);
-			}
-			if (logAccessibilityWarning && !isInteractive && !this.announced) {
-				console.warn(
-					'd2l-tooltip may be being used in a non-accessible manner; it should be attached to interactive elements like \'a\', \'button\',' +
-					'\'input\'', '\'select\', \'textarea\' or static / custom elements if a role has been set and the element is focusable.',
+
+			if (this.#targetDelegated) {
+				if (this.forType === 'label') {
+					this._target.removeAttribute('aria-labelledby');
```
dbatiste, line 974: Is it safe to remove `aria-labelledby` and `aria-describedby` from the target? i.e. what if the target custom element has set this attribute... this code would be wiping it out. I've always felt like modifying the target's DOM was sketchy here, violating the principle that the custom element's DOM should strictly be a function of its `render` method and reactive properties.

components/tooltip/tooltip.js
dbatiste: This overlaps with the comments I just made.
https://github.com/BrightspaceUI/core/pull/4606/files#r1571060598
https://github.com/BrightspaceUI/core/pull/4606/files#r1571072507

Dave, if I follow you correctly, that would address my concern about modifying the DOM outside of the target custom element's `render` method. That's where my mind was going too, but you put a bit more thought into the sol'n. 😄

dlockhart: Yeah, that was my intent.

bearfriend: Yeah, I agree with all of this. I'll investigate a little more, and try to get some numbers on how many places would need to be updated.

bearfriend:
Closing for now. We can re-open or re-work it when it's better prioritized.

PR #4609, opened 2024-04-17
Allow exemptions to focus trapping

bearfriend:
[GAUD-6323](https://desire2learn.atlassian.net/browse/GAUD-6323): equatio chrome extension not functioning in quizzing tool

To allow specific elements outside a focus trap container to be focused, we can add an exemption list. This allows for mouse clicks and roundabout screenreader navigation to focus the exempted elements, but not normal keyboard navigation (tabbing).

[GAUD-6323]: https://desire2learn.atlassian.net/browse/GAUD-6323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/focus-trap/focus-trap.js
+++ b/components/focus-trap/focus-trap.js
@@ -40,7 +40,7 @@ class FocusTrap extends FocusMixin(LitElement) {
 	].join(', ');
 
 	static #isExempt(target) {
-		return !!target.closest(this.#exemptSelectors);
+		return !!target?.closest(this.#exemptSelectors);
 	}
 
 	constructor() {
--- a/components/focus-trap/focus-trap.js
+++ b/components/focus-trap/focus-trap.js
@@ -34,6 +34,15 @@ class FocusTrap extends FocusMixin(LitElement) {
 		`;
 	}
 
+	static #exemptSelectors = [
+		'.equatio-toolbar-wrapper',
+		'.equatio-toolbar-shadow-root-container'
+	].join(', ');
+
+	static #isExempt(target) {
+		return !!target.closest(this.#exemptSelectors);
+	}
+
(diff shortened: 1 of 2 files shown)


components/focus-trap/focus-trap.js
```diff
@@ -34,6 +34,15 @@ class FocusTrap extends FocusMixin(LitElement) {
 		`;
 	}
 
+	static #exemptSelectors = [
```
dlockhart, line 37: Is it worth adding a built-in one like `.d2l-focus-trap-exempt` that Equatio (or some other future thing) could possibly put on their stuff, allowing us to not have references to their classes in our code?

components/focus-trap/focus-trap.js
dlockhart, line 1: Wondering if the legacy prompt stuff in here could be switched over to using this somehow...

bearfriend: Right now it looks like it just disables trapping altogether when a legacy prompt is present, which means users could tab through everything to get to that prompt. If we added an exemption for the prompts instead it would currently still trap when tabbing, but not when clicking or navigating with screen reader.

I did a bunch of investigating whether I could make tabbing work too, but it just gets really ugly trying to essentially reinvent the trap as you tab through. You'd need to inject more `d2l-focus-trap-start`/`d2l-focus-trap-end` elements on the fly.

dlockhart: Gotcha -- ok thanks for taking a look!

dlockhart (approved):

dbatiste (approved):

PR #4619, opened 2024-04-23
Try removing eslint-parser depdendency

dlockhart:
`@babel/eslint-parser` is a peer dependency of `eslint-config-brightspace` so shouldn't be required here. Likely a holdover from ancient times.


lweber-d2l (approved):
Thanks for discovering this!

bearfriend (approved):

PR #4626, opened 2024-04-25
Add the dropdown with accessibility

saifaldin14:


components/table/table-col-sort-button.js
```diff
@@ -83,19 +96,31 @@ export class TableColSortButton extends LocalizeCoreElement(RtlMixin(FocusMixin(
 				background-color: var(--d2l-color-gypsum);
 			}
 			:host([has-sibling]) button {
+				--d2l-table-cell-padding: 0;
```
margaree, line 99: Does setting this here actually do anything?

components/table/table-col-sort-button.js
```diff
@@ -129,6 +166,7 @@ export class TableColSortButton extends LocalizeCoreElement(RtlMixin(FocusMixin(
 		super();
 		this.dataType = 'words';
 		this.desc = false;
+		this._hasDropdownItems = false;
```
margaree, line 173: We generally set internal properties alphabetically below the external ones (i.e., it would go after `this.nosort`)

components/table/table-col-sort-button.js
```diff
@@ -40,13 +40,26 @@ export class TableColSortButton extends LocalizeCoreElement(RtlMixin(FocusMixin(
...
 				type: String
 			},
+			/**
+			 * Custom message to define the type of sorting that's occuring.
+			 * @type {String}
+			 */
+			description: {
+				reflect: true,
+				type: String
+			},
```
margaree, line 50: So I believe we want to pull the text for the `aria-description` for the dropdown button directly from the selected item in the menu and don't want to have the consumer have to specify it.

margaree (commented):
As mentioned in our call feel free to ignore this, but just getting this review in in case you continue working on this today!

PR #4628, opened 2024-04-25
Address Accessbility no dropdown

saifaldin14:


components/table/table-col-sort-button.js
```diff
@@ -122,8 +131,24 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 		const iconView = !this.nosort ?
 			html`<d2l-icon icon="${this.desc ? 'tier1:arrow-toggle-down' : 'tier1:arrow-toggle-up'}"></d2l-icon>` :
 			null;
+		const buttonTitle = this._getSortButtonTitle();
+		const buttonDescription = !this.nosort ? 'click to change sort order' : 'click to add sort order';
```
bearfriend, line 135: All the rendered text like this needs to be run through our [`LocalizeMixin`](https://github.com/BrightspaceUI/core/tree/main/mixins/localize) so that we render in the correct language. So you need to set up the mixin, add these strings to the `en.js` lang file, then do something like this:
```suggestion
		const buttonDescription = !this.nosort ? this.localize('components.table.clickChangeSortOrder') : this.localize('components.table.clickAddSortOrder');
```

components/table/table-col-sort-button.js
```diff
@@ -122,8 +131,24 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+
+		return html`<button aria-description="${buttonDescription}" aria-label="${buttonTitle}" title="${buttonTitle}" type="button"><slot></slot>${iconView}</button>`;
+	}
+
+	_getSortButtonTitle() {
+		const columnDataTypeWords = {
+			'words': ['A to Z', 'Z to A'],
+			'numbers': ['low to high', 'high to low'],
+			'dates': ['old to new', 'new to old']
+		};
```
bearfriend, line 145: Same deal here, but you can store this data as a static private like:
```javascript
export class TableColSortButton extends FocusMixin(LitElement) {

static #columnDataTypeKeys = {
	words: ['components.table.aToZ', 'components.table.zToA'],
	numbers: ['components.table.lowToHigh', 'components.table.highToLow'],
	dates: ['components.table.oldToNew', 'components.table.newToOld']
}
```

All these langterms that get put into `en.js` should be the complete sentence/phrase, like, `Sorted high to low`.

Then back here, call `localize`:
```javascript
if (!this.nosort) {
	sortKey = TableColSortButton.#columnDataTypeKeys[this.columnDataType][Number(!this.desc)]
	return this.localize(sortKey);
	
}
return '';
```

components/table/table-col-sort-button.js
```diff
@@ -107,11 +117,20 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 		`;
 	}
 
+	static #columnDataTypeKeys = {
+		words: ['components.table.a-to-z', 'components.table.z-to-a'],
+		numbers: ['components.table.low-to-high', 'components.table.high-to-low'],
+		dates: ['components.table.old-to-new', 'components.table.new-to-old']
+	};
+
+	static #buttonDescription = ['components.table.change-sort-order', 'components.table.add-sort-order'];
```
margaree, line 126: It might be clearer to entirely use objects instead of arrays for these. For example:
```
static #columnDataTypeKeys = {
words: { asc: 'components.table.a-to-z', desc: 'components.table.z-to-a' }
};
```
etc.

components/table/table-col-sort-button.js
```diff
@@ -28,6 +30,14 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 				type: Boolean
 			},
+			/**
+			 * The type of data in the column.
+			 *  @type {'words'|'numbers'|'dates'}
+			 */
+			columnDataType: {
+				reflect: true,
+				type: String
+			},
```
margaree, line 40: This could probably just be `dataType` since the component is already column sort.
Reminder to add the `attribute`.

components/table/table-col-sort-button.js
```diff
@@ -107,11 +117,20 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+	};
+
+	static #buttonDescription = ['components.table.change-sort-order', 'components.table.add-sort-order'];
+
 	constructor() {
 		super();
 		this.nosort = false;
 		this.desc = false;
 		this.hasSibling = false;
+		this.columnDataType = 'words';
```
margaree, line 133: nit: we usually alphabetize these (though I see that hasn't really been followed at this point)

components/table/table-col-sort-button.js
```diff
@@ -122,8 +141,19 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+	}
+
+	_getSortButtonTitle() {
+		if (!this.nosort) {
+			const sortKey = TableColSortButton.#columnDataTypeKeys[this.columnDataType][Number(!this.desc)];
+			return this.localize(sortKey);
+		}
 
-		return html`<button type="button"><slot></slot>${iconView}</button>`;
+		return '';
```
margaree, line 156: With how this is currently done, I think it would still render with an empty `aria-label`. What we'd generally want to do is set the value to `undefined` in the case where there wouldn't be a label. Then in the html we'd use `aria-label="${ifDefined(buttonTitle)}"`.

Also with this function, you could early exit with an `if (this.nosort) return undefined` and then you don't have to indent the rest.

components/table/table-col-sort-button.js
```diff
@@ -122,8 +141,18 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 		const iconView = !this.nosort ?
 			html`<d2l-icon icon="${this.desc ? 'tier1:arrow-toggle-down' : 'tier1:arrow-toggle-up'}"></d2l-icon>` :
 			null;
+		const buttonTitle = this._getSortButtonTitle();
+		const description = this.localize(!this.nosort ? 'components.table.change-sort-order' : 'components.table.add-sort-order');
+
+		return html`<button aria-description="${description}" aria-label="${ifDefined(buttonTitle)}" title="${buttonTitle}" type="button"><slot></slot>${iconView}</button>`;
```
margaree, line 147: In the spec is there both `aria-label` and `title`? I had thought we just wanted `title`

PR #4631, opened 2024-04-26
Fix backdrop body scrolling methods

bearfriend:
Currently, backdrop methods to control body scrolling can register and track the same host multiple times, which can be disconnected from the DOM before all `scrollKeys` for that host can be removed, leaving an orphaned `scrollKey` that prevents the body scrolling from returning to its original value.

This change instead uses a `Set` to store references to host instances themselves rather than generating unique string ids for each `preventBodyScroll` call.

There are two uses of these methods outside core. Namely, `htmleditor` and `nova`, which I have updated and will open PRs for once this is approved.

[GAUD-6349](https://desire2learn.atlassian.net/browse/GAUD-6349): Fix d2l-backdrop `scrollKey` registration

[GAUD-6349]: https://desire2learn.atlassian.net/browse/GAUD-6349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -1,3 +1,4 @@
+import '../backdrop/backdrop.js';
 import '../focus-trap/focus-trap.js';
 import '../../helpers/viewport-size.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
--- a/components/dialog/dialog-mixin.js
+++ b/components/dialog/dialog-mixin.js
@@ -1,6 +1,5 @@
 import '../focus-trap/focus-trap.js';
 import '../../helpers/viewport-size.js';
-import { allowBodyScroll, preventBodyScroll } from '../backdrop/backdrop.js';
 import { clearDismissible, setDismissible } from '../../helpers/dismissible.js';
 import { findComposedAncestor, isComposedAncestor } from '../../helpers/dom.js';
 import { getComposedActiveElement, getFirstFocusableDescendant, getNextFocusable, isFocusable } from '../../helpers/focus.js';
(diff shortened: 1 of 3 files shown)


components/dialog/dialog-mixin.js
bearfriend, line 1: Body scrolling is already managed by the `d2l-backdrop` element this mixin renders, so these calls were unnecessary (and what was causing double registration).

dbatiste: The `d2l-backdrop` would not be rendered in the case of native `dialog` elements. (they have their own pseudo-element `::backdrop`, which is rendered immediately under the element on the top-layer). That might have been why the `d2l-backdrop` & `allowBodyScroll`/`preventBodyScroll` where originally decoupled.

I think we have native `dialog` disabled at the moment due to one or two browser bugs (per comments at the top of `dialog-mixin.js`), but hopefully we will be able to re-enable someday.

bearfriend: Ah, ok. I'll add this back conditionally, then. Thanks.

components/dialog/dialog-mixin.js
```diff
@@ -416,8 +415,7 @@ export const DialogMixin = superclass => class extends RtlMixin(superclass) {
 			return node.classList && node.classList.contains('d2l-dialog-outer');
 		});
 
-		// native dialog backdrop does not prevent body scrolling
```
bearfriend, line 419: @dbatiste I guess this comment, which I read many times, should have tipped me off 😆

dbatiste (approved):
Looks good!

PR #4636, opened 2024-04-30
Update tag-list to use focus-visible when supported.

dbatiste:
[GAUD-5473](https://desire2learn.atlassian.net/browse/GAUD-5473)

This PR updates the `TagListItemMixin` to use `:focus-visible` when supported by the browser. If not supported, it falls back to the regular focus style behaviour.

[GAUD-5473]: https://desire2learn.atlassian.net/browse/GAUD-5473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/tag-list/tag-list-item-mixin.js
```diff
@@ -111,6 +111,16 @@ export const TagListItemMixin = superclass => class extends LocalizeCoreElement(
 			:host(:focus-within:hover) .tag-list-item-container {
 				box-shadow: inset 0 0 0 2px var(--d2l-color-celestine), 0 2px 4px rgba(0, 0, 0, 0.03);
 			}
+			@supports selector(:has(a, b)) {
```
dbatiste, line 114: Note that this is not the most concise way to write this. It is written so that the fallback will work properly in older browsers. The most concise would be to write it as the following, but `selector(...)` must be supported in order to rely on writing fallback in this way. ([canIUse @supports selector](https://caniuse.com/mdn-css_at-rules_supports_selector))
```css
.some-class:has(:focus-visible) {
     ...
}
@supports not selector(:has(a, b)) {
     :some-class:focus-within {
          ...
     }
}
```

dlockhart (approved):

PR #4645, opened 2024-05-01
PoC - HTML elem context provider

grant-cleary:
Quick proof of concept around what a centralized HTML elem context provider might look like. Note that there's no reliance on the actual `lms-context-provider` here, as `core` can't reference it. Consumers (i.e. BSI) would need to hook that stuff up themselves.

### DIFF
--- a/helpers/html-elem-context-provider.js
+++ b/helpers/html-elem-context-provider.js
@@ -1,0 +1,32 @@
+const attributeCallbacks = new Map();
+
+const contextObserver = new MutationObserver(mutations => {
+	mutations.forEach(mutation => {
+		const callbacks = attributeCallbacks.get(mutation.attributeName);
+		if (callbacks !== undefined && callbacks.length > 0) {
+			callbacks.forEach(callback => callback());
+		}
+	});
+});
+
+let observerInitialized = false;
+
+export function getContext(attributeName) {
+	const contextString = document.documentElement.hasAttribute(attributeName)
+		? document.documentElement.getAttribute(attributeName)
+		: undefined;
+
+	return JSON.parse(contextString);
+}
+
+export function registerOnChangeHandler(attributeName, onChange, sendImmediate) {
+	if (!observerInitialized) {
+		contextObserver.observe(document.documentElement, { attributes: true });
+		observerInitialized = true;
+	}
+
+	if (sendImmediate) onChange(getContext(attributeName));
+
+	if (!attributeCallbacks.has(attributeName)) attributeCallbacks.set(attributeName, []);
+	attributeCallbacks.get(attributeName).push(() => onChange(getContext(attributeName)));
+}


helpers/html-elem-context-provider.js
```diff
@@ -0,0 +1,32 @@
...
+		const callbacks = attributeCallbacks.get(mutation.attributeName);
+		if (callbacks !== undefined && callbacks.length > 0) {
+			callbacks.forEach(callback => callback());
+		}
+	});
+});
+
+let observerInitialized = false;
+
+export function getContext(attributeName) {
```
grant-cleary, line 14: As I mentioned in the BSI PR, I think whether we even want this in `core` depends on whether we think there's value in other consumers being able to import this to set up their own fake hosts using the HTML element. I think it might more sense to have the `lms-context-provider` provide a `test-host` export that can do this without relying on the document, however, in which case this stuff should probably just live in (and be managed by) BSI. 🤷

helpers/html-elem-context-provider.js
```diff
@@ -0,0 +1,32 @@
+const attributeCallbacks = new Map();
+
+const contextObserver = new MutationObserver(mutations => {
+	mutations.forEach(mutation => {
+		const callbacks = attributeCallbacks.get(mutation.attributeName);
+		if (callbacks !== undefined && callbacks.length > 0) {
+			callbacks.forEach(callback => callback());
+		}
```
dlockhart, line 8: Nit: could probably be:

```suggestion
		attributeCallbacks.get(mutation.attributeName)?.forEach(callback => callback());
```

helpers/html-elem-context-provider.js
```diff
@@ -0,0 +1,32 @@
...
+});
+
+let observerInitialized = false;
+
+export function getContext(attributeName) {
+	const contextString = document.documentElement.hasAttribute(attributeName)
+		? document.documentElement.getAttribute(attributeName)
+		: undefined;
+
+	return JSON.parse(contextString);
```
dlockhart, line 19: `JSON.parse(undefined)` throws... is that what we want or should it just return `undefined`?

helpers/html-elem-context-provider.js
dlockhart: I'm personally leaning towards trying to find a nicer "automatic" solution for tests & demos. This doesn't feel very core-y... like _maybe_ if it were made more generic as a way to observe attributes on an arbitrary element but even that feels very specific to this case.

grant-cleary: Yeah, I think I agree. I'm inclined to move this logic into BSI (we need this behaviour one way or the other in the LMS) and expose something like that `test-host` in the provider repo itself. `core` stuff like the `html-block` can keep using the attribute controller as it currently does, and if that's ever something we want to change, we can cross that bridge as we come to it.

helpers/html-elem-context-provider.js
grant-cleary: I guess in theory this could be any arbitrary string, which could throw. I can wrap it in a try-catch (though I'll do that in the BSI move).

helpers/html-elem-context-provider.js
grant-cleary: I'm never going to get used to being able to use chaining. 😆

grant-cleary:
As I've percolated a bit more, I more and more prefer the idea of moving this into BSI and allowing for tests/demos with some kind of dedicated `test-host` exported by the context provider package. I'm going to close this out and shift all of this stuff to the BSI PR.

PR #4651, opened 2024-05-02
Table updates: table-col-sort-button: Accessibility improvements [GAUD-6125]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6125)

We want to add a title and aria-description to the button based on the state of the component. `data-type` attribute was added in order to use proper phrasing. If this seems like a good pattern, usages of `d2l-table-col-sort-button` will be updated to use `data-type` and in the meantime will fallback to the `numbers` phrasing (will check with Nicole as it's possible we'd want to just fallback to "Sorted").


margaree:
will open a fresh PR

PR #4656, opened 2024-05-03
Update inline-help and radio-spacer to make styling more fool proof.

dbatiste:
[GAUD-6341](https://desire2learn.atlassian.net/browse/GAUD-6341)

This PR makes styling and usage of the `d2l-input-radio-spacer` clearer based on conversation with @geurts . It makes it easier for us to update inline-help styles in the future, rather than having consumers using the less semantic `d2l-body-small`.

Specifically:
* update example to show typical wire-up using `aria-describedby`
* factor out the inline help styles into `inlineHelpStyles`
* add `_generateBodySmallStyles` to support `inlineHelpStyles` implementation
* update to show usage of `inlineHelpStyles` & `.d2l-input-inline-help` CSS class

Although not strictly required, this is the expected typical setup:
```html
<label class="d2l-input-radio-label">
	<input type="radio" aria-describedby="desc1" name="myGroup" value="normal">
	Option 1
</label>
<d2l-input-radio-spacer id="desc1" class="d2l-input-inline-help">
	Additional content can go here and will line up nicely with the edge of the radio.
</d2l-input-radio-spacer>
```

image

[GAUD-6341]: https://desire2learn.atlassian.net/browse/GAUD-6341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/inputs/input-inline-help.js
+++ b/components/inputs/input-inline-help.js
@@ -1,6 +1,16 @@
 import { css, html } from 'lit';
-import { bodySmallStyles } from '../typography/styles.js';
+import { _generateBodySmallStyles } from '../typography/styles.js';
 import { SkeletonMixin } from '../skeleton/skeleton-mixin.js';
+
+export const inlineHelpStyles = [
+	_generateBodySmallStyles('.d2l-input-inline-help'),
+	css`
+		.d2l-input-inline-help {
+			margin-top: 0.3rem !important;
+			overflow-wrap: anywhere;
+		}
+	`
+];
 
 export const InputInlineHelpMixin = superclass => class extends SkeletonMixin(superclass) {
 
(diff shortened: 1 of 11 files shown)


components/typography/styles.js
dbatiste, line 1: Nothing changed with these styles - they were just moved into the generate method so that they can be used by the inline-help styles.

components/inputs/input-inline-help.js
dbatiste, line 1: I decided to rename this file because it now exports more than just the mixin.

dlockhart (approved):

PR #4657, opened 2024-05-03
Table updates: Support radio menu items in a slot for additional sorting [GAUD-6122]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6122)

This allows consumers to put `d2l-menu-item-radio` items in the `items` slot in order to have cases with more than asc and desc (e.g., first name/last name combinations).
Note that this does not work with sticky headers.
I'll be adding documentation with further examples in a followup story.


margaree:
Will open a fresh PR for this.

PR #4671, opened 2024-05-10
Table Updates: Styles

margaree:
Due to the potential downstream impact of table style changes, I wanted to get hopefully all of them in one PR, so it is a bit large.

This PR covers the following stories:
- [Tables > Update table-col-sort-button styles](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6123)
- [Tables > Handle table-col-sort-button styles when multiple items in a cell](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6211)
- [Table > Handle cell height when checkbox, icon button, or text input is in cell](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6276)
- [Table > Update cell padding of normal cell](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6281)

In summary, we want to do the following:
- Update the look and feel of the `d2l-table-col-sort-button` to take up the whole cell if it can (nothing else in the cell) and have fallback styles if not (10px padding on each side in the shared cell case)
- All cells should have 0.7rem of padding


components/table/table-col-sort-button.js
dlockhart, line 1: Playing around with the first demo, I noticed that because the sort indicator can cause the clickable area to become larger when it's applied, in some scenarios it can cause wrapping:

Screenshot 2024-05-10 at 9 07 16 AM

Wondering if it might make sense to "reserve" some space for it using a bit of additional padding to the left and right (which adds up to the exact size of the indicator) that then gets removed when the indicator is present?

margaree: I'll raise that with Jeff and Nicole but that would seem like a good idea to me

dlockhart: Awesome, thanks. Even for columns that don't wrap (like the "Population" one in the demo), I'm also seeing them grow slightly and cause some jitter/reflow. That same solution should solve that as well.

components/table/table-wrapper.js
```diff
@@ -32,6 +33,38 @@ export const tableStyles = css`
...
+		.d2l-table > * > tr > :has(d2l-selection-input) {
+			padding-block: calc(0.5 * (var(--d2l-table-cell-overall-height) - ${cssSizes.inputBoxSize}rem));
+		}
+		.d2l-table > * > tr.d2l-table-selected-first > :has(.d2l-checkbox),
+		.d2l-table > * > tr.d2l-table-selected-first > :has(d2l-input-checkbox),
+		.d2l-table > * > tr.d2l-table-selected-first > :has(d2l-selection-input) {
+			padding-bottom: calc(0.5 * (var(--d2l-table-cell-overall-height) - ${cssSizes.inputBoxSize}rem));
+			padding-top: calc(0.5 * (var(--d2l-table-cell-overall-height) - ${cssSizes.inputBoxSize}rem) - 1px);
+		}
+	}
```
margaree, line 67: The checkbox inputs add extra height that we don't want so we shrink the padding accordingly.

components/table/table-wrapper.js
```diff
@@ -54,10 +87,46 @@ export const tableStyles = css`
...
-		height: 27px; /* min-height to be 48px including border */
+
+	d2l-table-col-sort-button {
+		vertical-align: middle;
+	}
+	d2l-table-col-sort-button:not(:only-child) {
+		--d2l-table-col-sort-button-additional-padding-inline-end: 0px; /* stylelint-disable-line length-zero-no-unit */
+	}
+	d2l-table-col-sort-button:not(:only-child):first-child {
+		--d2l-table-col-sort-button-margin-inline-end: 1px;
```
margaree, line 99: Minimize space between `d2l-table-col-sort-button`s on siblings (default is 4px)

components/table/table-wrapper.js
```diff
@@ -257,9 +328,6 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 			:host([type="light"]) {
 				--d2l-table-border-radius: 0rem; /* stylelint-disable-line length-zero-no-unit */
 				--d2l-table-border-radius-sticky-offset: 0rem; /* stylelint-disable-line length-zero-no-unit */
-				--d2l-table-cell-height: 1.15rem; /* min-height to be 48px including border */
-				--d2l-table-cell-padding: 0.6rem;
-				--d2l-table-cell-padding-alt: calc(0.6rem - 1px) 0.6rem 0.6rem 0.6rem;
```
margaree, line 262: Light type padding and height is now to be consistent with default

components/table/table-col-sort-button.js
```diff
@@ -33,19 +33,32 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 
 	static get styles() {
 		return css`
+			:host {
+				--d2l-table-col-sort-button-additional-padding-inline-end: 0px; /* stylelint-disable-line length-zero-no-unit */
+				--d2l-table-col-sort-button-margin-inline-end: var(--d2l-table-cell-col-sort-button-size-offset);
+				--d2l-table-col-sort-button-width: calc(100% - 2 * var(--d2l-table-cell-col-sort-button-size-offset));
+			}
+			:host([nosort]) {
+				--d2l-table-col-sort-button-additional-padding-inline-end: calc(0.6rem + 18px);
```
margaree, line 42: This is to account for extra space for the icon + icon's margin when the button is in `nosort` state. It's a cssvar in order to be able to set it for different cases when there are siblings.

components/table/table-wrapper.js
```diff
@@ -54,10 +87,47 @@ export const tableStyles = css`
...
+		padding: 0;
+	}
+	.d2l-table th.d2l-table-header-col-sortable-siblings d2l-table-col-sort-button {
+		--d2l-table-col-sort-button-width: unset;
+	}
+	.d2l-table-header-col-sortable-no-sorted.d2l-table-header-col-sortable-siblings :last-child {
+		padding-inline-end: calc(0.6rem + 18px);
+	}
+	.d2l-table-header-col-sortable :not(d2l-table-col-sort-button) {
+		margin: var(--d2l-table-cell-col-sort-button-size-offset);
```
margaree, line 114: This adds a margin to non-table-col-sort-buttons that are in a cell with one. In our demos it looks weird without this, however consumers can be setting their own spacing on their cell contents so I don't know that it's right for us to be dictating that. Let me know any thoughts. I'm open to removing it from here and moving it elsewhere for the demo.

components/table/table-wrapper.js
```diff
@@ -368,6 +437,7 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 			r.classList.toggle('d2l-table-selected-first', firstNonHeaderRow && isSelected);
 
 			Array.from(r.cells).forEach((c, index) => {
+				if (isHeader && !CSS.supports('selector(:has(a, b))')) this._checkSiblingSortableCells(c);
```
margaree, line 440: With `:has` we are able to do everything without js, so that's why there is a separate `:has` section in the css. This can be removed once we only support browsers with `:has`

components/table/table-wrapper.js
```diff
@@ -242,9 +312,11 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
 				--d2l-table-border-radius: 0.3rem;
 				--d2l-table-border-radius-sticky-offset: calc(1px - var(--d2l-table-border-radius));
-				--d2l-table-cell-height: 41px; /* min-height to be 62px including border */
-				--d2l-table-cell-padding: 0.5rem 1rem;
-				--d2l-table-cell-padding-alt: calc(0.5rem - 1px) 1rem 0.5rem 1rem;
+				--d2l-table-cell-overall-height: 46px;
+				--d2l-table-cell-height: calc(var(--d2l-table-cell-overall-height) - 2 * var(--d2l-table-cell-padding));
+				--d2l-table-cell-padding: 0.7rem;
+				--d2l-table-cell-padding-alt: calc(0.7rem - 1px) 0.7rem 0.7rem 0.7rem;
+				--d2l-table-cell-col-sort-button-size-offset: 4px;
```
margaree, line 319: This corresponds to the 4px of buffer we want on all sides of the table-col-sort-button

margaree (commented):
Nicole is interested in having it so that if there are multiple things in the cell (e.g., multiple table-col-sort-button, table-col-sort-button + button-icon) that when they wrap, if one is focused and the other is hovered then those two styles just touch in order to reduce height. I have found trying to do this to be difficult (perhaps I'm missing something totally obvious) so let me know if you have suggestions on that.

components/table/table-wrapper.js
dlockhart: Yeah this seems a bit invasive to me, and likely best handled by consumer code wherever it comes up.

dlockhart (approved):
Crazy how complex all the positioning ended up being, nice job!

PR #4674, opened 2024-05-14
feat: Collapsible Panel: Remove title line clamping [GAUD-6356]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6356)


dbatiste (approved):

components/collapsible-panel/collapsible-panel.js
```diff
@@ -171,7 +171,6 @@ class CollapsiblePanel extends SkeletonMixin(FocusMixin(RtlMixin(LitElement))) {
 				-webkit-box-orient: vertical;
 				display: -webkit-box;
 				flex: 1;
-				-webkit-line-clamp: 2;
```
dlockhart, line 174: There were a [few other changes](https://github.com/BrightspaceUI/core/pull/3299) made when Dave added this... some I think we still need (like `overflow-wrap: anywhere`, but I'm wondering about `overflow-x/y` and a couple others...

dbatiste: Oh yeah, good call. I think `-webkit-box-orient: vertical;` and `display: -webkit-box;` are probably to make the line-clamping work too.

dlockhart (approved):

components/collapsible-panel/collapsible-panel.js
bearfriend: What is the `overflow-wrap: anywhere` getting us here?

dbatiste: Wrapping unbreakable titles.

PR #4679, opened 2024-05-14
Document collapsible panel events

semibran:
Noticed this stuff is present in [daylight](https://daylight.d2l.dev/components/collapsible-panel/) but not here, adding for clarity


dlockhart (approved):
Thanks!

PR #4685, opened 2024-05-16
fix: Dropdown: Align value is not properly respected when opener is wider than dropdown [GAUD-6375]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6375)


components/dropdown/dropdown-content-mixin.js
```diff
@@ -320,6 +320,15 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
 	}
 
+	async getUpdateComplete() {
+		const fontsPromise = document.fonts ? document.fonts.ready : Promise.resolve();
+		await super.getUpdateComplete();
+		/* wait for the fonts to load because browsers have a font block period
+		where they will render an invisible fallback font face that may result in
+		improper width calculations before the real font is loaded */
+		await fontsPromise;
+	}
```
margaree, line 330: I found that if the page loaded in an `opened` state then the positioning wasn't right. This sometimes affected visual diffs as well. I found that waiting for the fonts to be ready fixed this, and this is the pattern that tooltip uses so I copied this here.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -973,13 +981,13 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
-
+	_getPosition(spaceAround, targetWidth, contentWidth) {
+		const centerDelta = contentWidth - targetWidth;
 		const contentXAdjustment = centerDelta / 2;
-		if (centerDelta <= 0) {
+		if (!this.align && centerDelta <= 0) {
 			return contentXAdjustment * -1;
 		}
-		if (spaceAround.left > contentXAdjustment && spaceAround.right > contentXAdjustment) {
+		if (!this.align && spaceAround.left > contentXAdjustment && spaceAround.right > contentXAdjustment) {
```
margaree, line 990: If align has a value and wider opener then `_getPosition` ends up in here which is not the result we want.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1001,6 +1009,14 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
 			}
 		}
+		if (this.align === 'start' || this.align === 'end') {
+			const shift = Math.min((targetWidth / 2) - (20 + 16 / 2), 0);
+			if (this.align === 'start') {
+				return shift;
+			} else {
+				return targetWidth - contentWidth - shift;
+			}
+		}
```
margaree, line 1019: This is at the bottom because the above cases still apply for the cases where there isn't enough space to align right at the end or at the start.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1001,6 +1009,14 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 				return spaceAround.right * -1;
 			}
 		}
+		if (this.align === 'start' || this.align === 'end') {
+			const shift = Math.min((targetWidth / 2) - (20 + 16 / 2), 0);
```
margaree, line 1013: The 20 corresponds approximately to 1rem, and the 16 is the pointer size. I'll add a comment to note this.
This is the same logic that tooltip uses for positioning.

components/dropdown/dropdown-content-styles.js
```diff
@@ -61,16 +64,26 @@ export const dropdownContentStyles = css`
...
+	:host([align="start"]) .d2l-dropdown-content-pointer,
+	:host([align="end"][dir="rtl"]) .d2l-dropdown-content-pointer {
+		left: min(calc(1rem + ${(pointerRotatedLength - pointerLength) / 2}px), calc(50% - ${pointerLength / 2}px)); /* 1rem corresponds to .d2l-dropdown-content-container padding */
+		right: auto;
+	}
+	:host([align="end"]) .d2l-dropdown-content-pointer,
+	:host([align="start"][dir="rtl"]) .d2l-dropdown-content-pointer {
+		left: auto;
+		right: min(calc(1rem + ${(pointerRotatedLength - pointerLength) / 2}px), calc(50% - ${pointerLength / 2}px)); /* 1rem corresponds to .d2l-dropdown-content-container padding */
+	}
```
margaree, line 76: The pointer positioning previously was always approximately centre with the opener, which is not correct for the cases with align values and wide openers.
I checked with Jeff and he wanted it to be the same behaviour as tooltip, which is what this is.

Note that in tooltip this has a backup for browser that don't support `min`. I checked our updated supported browsers chart and compared that with caniuse and all cases where browsers don't support it correspond with our unsupported browsers. Let me know if I should still add a backup just in case.

components/dropdown/dropdown-content-mixin.js
nathankoop: Is there a case for this to be moved up to a parent class?

helpers/visualReady.js
```diff
@@ -0,0 +1,3 @@
+[...document.fonts].forEach(font => font.load());
+
+export const visualReady = Promise.all([document.fonts.ready]);
```
margaree, line 3: @bearfriend let me know thoughts on this and if it's inline with what you were thinking

dlockhart (approved):
This looks excellent to me!

helpers/visualReady.js
bearfriend: Yep this looks good 👍

PR #4687, opened 2024-05-17
chore: Add lifecycle info to localize mixin

eKoopmans:
I want to clarify the Lit lifecycle implications for `LocalizeMixin`, to make it clear when we are guaranteed to have valid terms.

This came up when we were trying to troubleshoot where to update `document.title`, which some of our components were doing too soon in `connectedCallback` - some Slack conversation here:
https://d2l.slack.com/archives/C0PHG3QB0/p1715891946771899

### DIFF
--- a/mixins/localize/README.md
+++ b/mixins/localize/README.md
@@ -61,9 +61,15 @@ static get localizeConfig() {
 }
 ```
 
+### Lifecycle
+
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Thanks!

PR #4695, opened 2024-05-24
feat: Filters > Count Badge to show sum of selected filters [GAUD-96]

KearseTrevor:
WIP: Failing axe tests and filter > opener vdiffs currently.

https://desire2learn.atlassian.net/browse/GAUD-96

Addresses consistency issue between filter sum count display and dimension count display.

Note: Added noise suppression jsdoc comments because my lit-plugin complained that 
[`@d2l-hierarchical-view-hide`, `@d2l-hierarchical-view-show-complete`, `@d2l-hierarchical-view-show-start`] were unknown events from `filter.js`'s perspective. Can revert these if desired but should probably extend the jsdoc comments to be proper descriptions. I just didn't feel like I had the background to write adequate comments on the fly.

Additional note: I am unaware of the workflow/auditing tools that Gaudi makes use of for accessibility - and as such would like the reviewer of this PR to kindly take a moment to discuss with me offline. Thank you!

Before Changes:
image

After Changes:
image


components/filter/filter.js
```diff
@@ -2,6 +2,7 @@ import '../colors/colors.js';
 import '../count-badge/count-badge.js';
 import '../button/button-icon.js';
 import '../button/button-subtle.js';
+import '../dropdown/dropdown.js';
 import '../dropdown/dropdown-button-subtle.js';
```
dbatiste, line 6: Looks like the `dropdown-button-subtle.js` import can be removed.

components/filter/filter.js
```diff
@@ -266,17 +266,32 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				</d2l-dropdown-menu>
 			`;
 
+		const countBadgeTemplate = filterCount ? html`
+				<d2l-count-badge
+					type="count"
+					max-digits=5
```
dbatiste, line 272: The [default max-digits for d2l-count-badge](https://github.com/BrightspaceUI/core/blob/main/components/count-badge/count-badge-mixin.js#L12) is currently 5. We discussed it very briefly but it wasn't super clear whether she was ok with the default (which I mentioned was 5) vs really wanting 5 regardless of the default.

dbatiste:
> Note: Added noise suppression jsdoc comments because my lit-plugin complained that
> [`@d2l-hierarchical-view-hide`, `@d2l-hierarchical-view-show-complete`, `@d2l-hierarchical-view-show-start`] were unknown events from `filter.js`'s perspective. Can revert these if desired but should probably extend the jsdoc comments to be proper descriptions. I just didn't feel like I had the background to write adequate comments on the fly.

It looks like there are 5 events in total for the hierarchical-view:
```
d2l-hierarchical-view-hide-start
d2l-hierarchical-view-show-start
d2l-hierarchical-view-hide-complete
d2l-hierarchical-view-show-complete
d2l-hierarchical-view-resize
```

We have descriptions for 4 of the events [in the hierarchical-view readme](https://github.com/BrightspaceUI/core/tree/main/components/hierarchical-view#events) (the resize event might have been added later and we didn't document it). My suggestion is that we update the JSDoc comments with these descriptions (rephrasing/reformatting to be consistent with other components), and remove them from the readme. That way, they'll be formatted using our Daylight documentation template (ex. [Alert events](https://daylight.d2l.dev/components/alert/#d2l-alert-events)).

Note: I think the `HierarchicalViewMixin` has them set to `/** @ignore */` so that we don't display them in the docs for `d2l-menu`, since `Menu` extends `HierarchicalViewMixin` and we didn't want menu consumers trying to use those events, since the underlying implementation was/is always subject to change.

components/count-badge/demo/count-badge.html
```diff
@@ -41,6 +41,19 @@ <h2>Large Count Badge</h2>
 				</template>
 			</d2l-demo-snippet>
 
+			<h2>Override Colorway of Count Badge</h2>
+			<d2l-demo-snippet>
+				<template>
+					<d2l-count-badge id="colorway-override" size="large" text="1 new notification." type="count" number="1"></d2l-count-badge>
+
+					<script>
```
dbatiste, line 49: I think this would also work:
```html
<d2l-count-badge style="--d2l-count-badge-background-color: purple; --d2l-count-badge-foreground-color: white;" ...></d2l-count-badge>
```

margaree:
> > Note: Added noise suppression jsdoc comments because my lit-plugin complained that
> > [`@d2l-hierarchical-view-hide`, `@d2l-hierarchical-view-show-complete`, `@d2l-hierarchical-view-show-start`] were unknown events from `filter.js`'s perspective. Can revert these if desired but should probably extend the jsdoc comments to be proper descriptions. I just didn't feel like I had the background to write adequate comments on the fly.
> 
> It looks like there are 5 events in total for the hierarchical-view:
> 
> ```
> d2l-hierarchical-view-hide-start
> d2l-hierarchical-view-show-start
> d2l-hierarchical-view-hide-complete
> d2l-hierarchical-view-show-complete
> d2l-hierarchical-view-resize
> ```
> 
> We have descriptions for 4 of the events [in the hierarchical-view readme](https://github.com/BrightspaceUI/core/tree/main/components/hierarchical-view#events) (the resize event might have been added later and we didn't document it). My suggestion is that we update the JSDoc comments with these descriptions (rephrasing/reformatting to be consistent with other components), and remove them from the readme. That way, they'll be formatted using our Daylight documentation template (ex. [Alert events](https://daylight.d2l.dev/components/alert/#d2l-alert-events)).
> 
> Note: I think the `HierarchicalViewMixin` has them set to `/** @ignore */` so that we don't display them in the docs for `d2l-menu`, since `Menu` extends `HierarchicalViewMixin` and we didn't want menu consumers trying to use those events, since the underlying implementation was/is always subject to change.

Just a note on this - [the hierarchical view mixin's page on Daylight](https://daylight.d2l.dev/components/hierarchical-view/#hierarchical-view---mixin) doesn't follow the component page format where it auto-magically generates the events table from the jsdoc comments (I think it just didn't seem to make sense for this case/demo), so if we wanted to do this then the whole page/demo would need an overhaul.

components/filter/filter.js
KearseTrevor: Yah wasn't sure if either solution here mattered too much. Feels like we can just let it follow whatever the default for countBadge `max-digits` ends up being?
I'll remove this explicit set then

components/count-badge/demo/count-badge.html
KearseTrevor: This can definitely also work and was the original way that I had implimented it but it caused issues with stylelint that I couldn't get around without an igore statement or file bypass being added in `.stylelintrc.json`.
I'm fine to revert it back to that solution as the demo was cleaner / more intuitive but this way relays the same information without causing stylelint issue

dbatiste: I see. I'd probably just add an ignore statement, but I'm ok either way. I probably wouldn't add a one-off file exception to `.stylelintrc.json` though.

KearseTrevor:
> > > Note: Added noise suppression jsdoc comments because my lit-plugin complained that
> > > [`@d2l-hierarchical-view-hide`, `@d2l-hierarchical-view-show-complete`, `@d2l-hierarchical-view-show-start`] were unknown events from `filter.js`'s perspective. Can revert these if desired but should probably extend the jsdoc comments to be proper descriptions. I just didn't feel like I had the background to write adequate comments on the fly.
> > 
> > 
> > It looks like there are 5 events in total for the hierarchical-view:
> > ```
> > d2l-hierarchical-view-hide-start
> > d2l-hierarchical-view-show-start
> > d2l-hierarchical-view-hide-complete
> > d2l-hierarchical-view-show-complete
> > d2l-hierarchical-view-resize
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We have descriptions for 4 of the events [in the hierarchical-view readme](https://github.com/BrightspaceUI/core/tree/main/components/hierarchical-view#events) (the resize event might have been added later and we didn't document it). My suggestion is that we update the JSDoc comments with these descriptions (rephrasing/reformatting to be consistent with other components), and remove them from the readme. That way, they'll be formatted using our Daylight documentation template (ex. [Alert events](https://daylight.d2l.dev/components/alert/#d2l-alert-events)).
> > Note: I think the `HierarchicalViewMixin` has them set to `/** @ignore */` so that we don't display them in the docs for `d2l-menu`, since `Menu` extends `HierarchicalViewMixin` and we didn't want menu consumers trying to use those events, since the underlying implementation was/is always subject to change.
> 
> Just a note on this - [the hierarchical view mixin's page on Daylight](https://daylight.d2l.dev/components/hierarchical-view/#hierarchical-view---mixin) doesn't follow the component page format where it auto-magically generates the events table from the jsdoc comments (I think it just didn't seem to make sense for this case/demo), so if we wanted to do this then the whole page/demo would need an overhaul.

Okay so for now I'll leave the page/demo overhaul as it's out of scope but I'll update the `README` and the jsdoc for `hierarchical-view.js` so that they are both in sync and have descriptions for all 5 events

dbatiste (approved):

components/count-badge/demo/count-badge.html
KearseTrevor: Stylelint doesn't seem to be playing nicely with parsing html inline styles when it comes to rules exceptions. 

I've swapped the current example to at least make use of a <style> tag rather than js style override.
https://github.com/BrightspaceUI/core/pull/4695/commits/035d517f2060ba8393346f28797991a831c7dc9d

KearseTrevor:
@dbatiste ready for re-review - only main note I have with the current state is
- Should we remove the [demo for colorway override](https://github.com/BrightspaceUI/core/pull/4695/files#diff-6d947c9320b49367eb2581af57f1616d9f47acf3184d14e81b7591b516f8b76fR44)?

dbatiste (approved):
Looks great!

bearfriend (approved):
This looks so much better

PR #4699, opened 2024-05-27
feat: Table updates: Accessibility improvements [GAUD-6125]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6125)

We want to add a title and description to the button based on the state of the component. The `data-type` attribute was added in order to use proper phrasing depending on type of data. The fallback is "Sorted ascending" and "Sorted descending".


components/table/table-col-sort-button.js
```diff
@@ -84,17 +95,57 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
 	static get focusElementSelector() {
 		return 'button';
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		if (!this.dataType) {
+			console.warn('d2l-table-col-sort-button: data-type attribute does not have a value.');
+		}
```
margaree, line 111: I added warnings for cases where there wasn't a `data-type`, but it's possible this would get too spammy, since nowhere will have it yet, and there also maybe could be cases where we would just want to use the fallback.

components/table/table-col-sort-button.js
```diff
@@ -84,17 +95,57 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+	}
+
 	render() {
+		const buttonDescription = this.nosort ? this.localize('components.table-col-sort-button.addSortOrder') : this.localize('components.table-col-sort-button.changeSortOrder');
+		const buttonTitle = this.nosort
+			? undefined
+			: this.localize('components.table-col-sort-button.title', {
+				dataType: this.dataType,
+				direction: this.desc ? 'desc' : undefined
+			});
```
margaree, line 121: I used the nested template on the Daylight localization sandbox to generate the lang term. It's a bit debatable on if we should use that one vs just having the code specify which lang term to use based on `data-type` and direction, but we would end up with 8 additional lang terms instead of one. Let me know thoughts on this.

lang/en.js
```diff
@@ -105,6 +105,9 @@ export default {
 	"components.switch.visibleWithPeriod": "Visible.",
 	"components.switch.hidden": "Hidden",
 	"components.switch.conditions": "Conditions must be met",
+	"components.table-col-sort-button.addSortOrder": "Click to add sort order",
+	"components.table-col-sort-button.changeSortOrder": "Click to change sort order",
```
margaree, line 109: The spec has "Click" but I'm wondering if it should be something like "Select" instead

components/table/table-col-sort-button.js
```diff
@@ -81,19 +92,54 @@ export class TableColSortButton extends FocusMixin(LitElement) {
...
+			</button><span id="${this._describedById}" hidden>${buttonDescription}</span>`;
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		if (!changedProperties.has('sourceType')) return;
+
+		if (!this.sourceType || this.sourceType === 'unknown') {
+			console.warn('d2l-table-col-sort-button: source-type attribute does not have a value specified.');
```
dlockhart, line 141: I'd be tempted to remove this and the warning in `firstUpdated` as well. It's going to be spammy, clients are going to see it, and since we have a decent (good enough for the past several years) fallback, maybe it's OK to just leave them as-is and treat setting this as a strong recommendation but not warning-worthy?

margaree: I'm also good with that

dlockhart (approved):
This LGTM. Probably just the README that needs updating?

dlockhart (approved):
🎉

PR #4701, opened 2024-05-29
fix: Table styles: handle when no spacing between table-col-sort-button components

margaree:
When working on removing the comma separator in the LMS tables with multiple table-col-sort-buttons side-by-side I noticed overlap:
Screenshot 2024-05-29 at 11 06 28 AM

The problem turned out to be that it looks like the LMS removes spacing between the components in the HTML markup, whereas in the demo I was working on I have them on separate lines. Because they are `inline-flex`, the browser adds space between them when they're on separate lines.

I simplified the code to remove the negative margins used. This does mean that in the case like in the demo site there will be extra space between 2 items, which we ideally don't want but I don't know if that's possible to reliably remove.

I looked into ways to remove the space but nothing seemed great so let me know if there are other ideas.


components/table/demo/table-test.js
```diff
@@ -60,7 +60,6 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
 				padding-block: 0;
 			}
 			.d2l-table > * > tr > :has(d2l-table-col-sort-button) d2l-button-icon {
-				margin-bottom: 4px;
```
margaree, line 63: This was no longer needed

components/table/table-col-sort-button.js
```diff
@@ -53,8 +52,8 @@ export class TableColSortButton extends FocusMixin(LitElement) {
 				font-size: inherit;
 				letter-spacing: inherit;
 				line-height: 0.9rem;
-				margin: var(--d2l-table-cell-col-sort-button-size-offset, 4px);
-				margin-inline-end: var(--d2l-table-col-sort-button-margin-inline-end);
+				margin-block: 0 var(--d2l-table-cell-col-sort-button-size-offset, 4px);
+				margin-inline: 0 var(--d2l-table-cell-col-sort-button-size-offset, 4px);
```
margaree, line 56: The cell handles adding padding on the top and left, and the component adds margins to the end and bottom in order to handle wrapping.

components/table/test/table.vdiff.js
```diff
@@ -552,6 +552,26 @@ describe('table', () => {
 					await expect(elem).to.be.golden();
 				});
 
+				it('two-col-sort-button-focus-hover', async() => {
```
margaree, line 555: These tests would have caught the issue

dbatiste (approved):

PR #4703, opened 2024-05-29
fix: add missing docs for inline-help slots

dlockhart:
Noticed that we weren't consistent documenting the `inline-help` slots, either in the code or READMEs. I think this is all of them now.


margaree (approved):
Thanks!

PR #4706, opened 2024-05-30
Strip out mo/mi elements when rendering MathML if they only contain non-breaking space characters or other whitespace

grant-cleary:
There's a bug in MathJax's experimental `[mml]/mml3` package that can cause equations to fail to render if extra spaces are added to equations via `<mo>` and `<mi>` elements containing non-breaking spaces. We don't really want this behaviour anyways, since adding extra spaces just messes up MathJax's rendering, so the simplest option here is to just remove these elements if we find them.

It does look like there's a fix for this, but only in MathJax 4, which is still in beta, so not something we'll be updating quite yet.

### DIFF
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -66,11 +66,19 @@ class HtmlBlockMathRenderer {
 
 		if (context.enableMML3Support) {
 			// There's a bug in the experimental MML3 plugin that causes mi and mo elements containing non-breaking
-			// spaces to break MathJax's math processing. Unfortunately, WIRIS tends to add a lot of these in chemistry
-			// equations. Since they're not really intended in the MathML spec anyways, we'll just remove them.
-			// See https://github.com/mathjax/MathJax/issues/3030 for more details.
+			// spaces to break MathJax's math processing (e.g. <mo>&nbsp;</mo>, <mi>Some&nbsp;Identifier</mi>).
+			// Unfortunately, WIRIS tends to add a lot of these in chemistry equations, so they break math processing.
+			//
+			// In order to address this, we can just remove any non-breaking spaces entirely, replacing them with
+			// empty strings. MathJax will ignore any empty elements as a result, and while this may mean intended
+			// whitespace is occasionally removed, it's necessary for MathJax to render anything at all.
+			//
+			// NOTE: MathJax evidently has a fix for this in MathJax 4, so we should consider trying to remove this when
+			// the update comes out of beta and we decide to take it on.
+			//
+			// See https://github.com/mathjax/MathJax/issues/3030 for some related discussion.
 			elem.querySelectorAll('mo, mi').forEach(elm => {
-				if (elm.innerHTML.replace(/&nbsp;/g, '').trim().length === 0) elm.remove();
+				elm.innerHTML = elm.innerHTML.replace(/&nbsp;/g, '');
 			});
 		}
 
--- a/helpers/mathjax.js
+++ b/helpers/mathjax.js
@@ -64,6 +64,16 @@ class HtmlBlockMathRenderer {
 			elm.style.height = '0.5rem';
 		});
 
+		if (context.enableMML3Support) {
+			// There's a bug in the experimental MML3 plugin that causes mi and mo elements containing non-breaking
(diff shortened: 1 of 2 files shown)


helpers/mathjax.js
```diff
@@ -64,6 +64,16 @@ class HtmlBlockMathRenderer {
 			elm.style.height = '0.5rem';
 		});
 
+		if (context.enableMML3Support) {
+			// There's a bug in the experimental MML3 plugin that causes mi and mo elements containing non-breaking
+			// spaces to break MathJax's math processing. Unfortunately, WIRIS tends to add a lot of these in chemistry
+			// equations. Since they're not really intended in the MathML spec anyways, we'll just remove them.
+			// See https://github.com/mathjax/MathJax/issues/3030 for more details.
+			elem.querySelectorAll('mo, mi').forEach(elm => {
+				if (elm.innerHTML.replace(/&nbsp;/g, '').trim().length === 0) elm.remove();
```
bearfriend, line 73:
```suggestion
				elm.innerText.trim() || elm.remove();
```

grant-cleary: Unfortunately that won't work. The `mo` and `mi` elements don't support `innerText`. 😞

grant-cleary: From the spec, `mo` or `mi` elements can't have children, so it's plausible that `textContent` would work in a similar fashion, but I'm also a bit nervous about making that assumption because the MathML standard is so shaky.

dbatiste: Based on the fix, we remove the element if there's **only** non-breaking spaces, but the comment indicates that MathJax processing breaks if there are non-breaking spaces in general. But we don't strip the non-breaking spaces, we remove the actual element. If the element is empty, is that an error too, vs just stripping the non-breaking spaces?

grant-cleary: Yeah, so this proposed solution is based on Davide's suggestion further down in that issue thread, in that further versions of MathJax are going to wipe out these elements entirely. Whether we care to be so heavy-handed is another question entirely, though an empty `mo` or `mi` element also doesn't really make conceptual sense, since "nothing" can't really be either an operator or an identifier.

As for the mixed `&nbsp;`/symbols case. Yeah, it's possible that could also pose an issue, though I'm a bit more nervous about stripping in that case because non-breaking spaces actually mean something. While I don't think whitespace in an `mo` element make a ton of sense (what kind of operator has spaces in it?), it could certainly be legitimate in a `mi` element. But the spec is also loosey-goosey about what a valid operator looks like, so I'm concerned that removing mixed spaces could be over-zealous.

That said, it does look like MathJax blows up in the mixed case, which feels unfortunate. Just stripping out the non-breaking spaces seems to work, but I'm at least a little bit worried that this will prove problematic in the future. Why does MathJax have to make everything more complicated...

dbatiste: Curious how do non-breaking spaces even get in there? Blindly pasting, or do users actually put them in there on purpose?

I vaguely recall running into something like this a long time ago. If the MathML spec indicated that they were invalid, I'd have no problem trimming them. If it's just MathJax's opinion, then I'm a little on the fence, but still leaning towards trimming because I'd rather have consumers' math working than not.

Either way, I think I would update the comment with a little more detail re which cases produce errors, and how/why we dealt with it the way we did.

grant-cleary: WIRIS likes dropping them in to properly space its chemistry equations, so using the chemistry editor basically always reproduces this. That's not great on WIRIS' end because the whole _point_ of rendering MathML is to take that presentation aspect out of the mark-up, but it's also not something we can fix on that end.

I have a suspicion that this is mostly MathJax being opinionated, and based on that comment thread, I think in MathJax 4 the renderer will remove these automatically within MathJax. For now though, I'm also of the opinion that rendering math is better than not doing so, even if it means following MathJax's whims.

I'll update the comment. I'll also drop in a line about making sure we check whether we can remove it if/when we get to MathJax 4.

dbatiste (approved):

grant-cleary:
Damn, I forgot to update the commit message on merge... Might be a bit of confusion when looking at the commit history now.

PR #4709, opened 2024-05-30
feat: Table: Support multi-faceted sorting with a dropdown [GAUD-6122]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6122)

This creates a new `d2l-col-sort-button-item` (radio menu item) and allows consumers to put these items in the `items` slot in order to have cases with more than asc and desc (e.g., first name/last name combinations).
Note that this does not work with sticky headers when there are multiple header rows.
I'll be adding documentation with further examples in a followup story.


components/table/demo/table-test.js
```diff
@@ -153,38 +154,40 @@ class TestTable extends RtlMixin(DemoPassthroughMixin(TableWrapper, 'd2l-table-w
...
-					?desc="${this._sortDesc}"
-					?nosort="${this._sortField !== item1.toLowerCase()}">${item1}</d2l-table-col-sort-button>
-				<d2l-table-col-sort-button
-					@click="${this._handleSort}"
-					source-type="words"
 					?desc="${this._sortDesc}"
-					?nosort="${this._sortField !== item2.toLowerCase()}">${item2}</d2l-table-col-sort-button>
+					?nosort="${noSort}">${name}
+					<d2l-menu-item-radio slot="items" text="City, A to Z" data-field="city" @d2l-menu-item-select="${this._handleSortComplex}" value="1"></d2l-menu-item-radio>
+					<d2l-menu-item-radio slot="items" text="City, Z to A" data-field="city" data-desc @d2l-menu-item-select="${this._handleSortComplex}" value="2"></d2l-menu-item-radio>
```
margaree, line 187: This would be my suggested usage:
- `data-field` (follows how the LMS handles this (uses `data-fieldname`)) to get the field to sort on
- `data-desc` to get whether that should trigger adding `desc` or not to the parent `d2l-table-col-sort-button`

This would be decided by the consumer since they have to handle the select event etc. Another option would be to create something like `d2l-table-col-sort-button-item` that extends `MenuItemMixin` and could then have these as properties. Any preference?

components/table/table-col-sort-button.js
```diff
@@ -108,20 +119,57 @@ export class TableColSortButton extends LocalizeCoreElement(FocusMixin(LitElemen
...
+	}
+
+	updated(changedProperties) {
+		super.updated(changedProperties);
+
+		// de-select any selected dropdown menu item
+		if (changedProperties.has('nosort') && this.nosort && this._hasDropdownItems) {
+			const selectedItem = this.querySelector('[selected]');
+			if (selectedItem) selectedItem.selected = false;
+		}
```
margaree, line 159: I'm torn on this bit as we've generally been having the consumer do everything with this (e.g., setting `nosort` and `desc`) so I could see arguing that the consumer should handle this as well.

components/table/table-col-sort-button.js
```diff
@@ -108,20 +119,57 @@ export class TableColSortButton extends LocalizeCoreElement(FocusMixin(LitElemen
...
+
+	_handleMenuItemChange(e) {
+		this._selectedMenuItemText = e.target?.text;
+	}
+
+	_handleSlotChange(e) {
+		const items = e.target?.assignedNodes({ flatten: true }).filter((node) => node.nodeType === Node.ELEMENT_NODE);
+		const filteredItems = items.filter((item) => {
+			const role = item.getAttribute('role');
+			return (role === 'menuitem' || role === 'menuitemcheckbox' || role === 'menuitemradio');
```
margaree, line 170: We do only want this to get `menu-item-radio` so perhaps we should only return those

components/table/demo/table-test.js
dbatiste: I have mixed feelings about having consumers place `d2l-menu-item-radio` into `d2l-table-col-sort-button`'s slot. It's effectively surfacing the internal `d2l-menu` usage as part of `d2l-table-col-sort-button`'s public API. I don't expect that we'd change that, but I think I prefer it if the consumer provides all the menu bits (inc `d2l-menu`), or none of it (using `d2l-table-col-sort-button-item` like you mentioned). If we provide a special item component, we'd also want to dispatch a corresponding event, keeping the menu-item even under wraps.

margaree: That's a good point. I think in that case my personal preference would be to have a special item component that gets used. We'd want the consumer to implement something very specific (I'm especially thinking of the accessibility behaviour here) but we'd be giving the option to do basically anything.

components/table/table-col-sort-button.js
dbatiste: I was just going to ask about that. With the current slot API, where the consumer renders `d2l-menu-item-radio` elements, it almost invites them to try and put other types of menu-items into the slot. And then I was wondering, is that valid, do we want/need to prevent that?

margaree: We only want `menu-item-radio` since we'd only want one multi-faceted sort option selected at a time.

dbatiste: If we really want to lock it down, sometimes we do something like:
```
::slotted(*) {
	display: none;
}
::slotted(d2l-table-col-sort-button-item) {
	display: ...;
}
```
Technically I think they'd still be assigned in the slot and we'd want to still filter here, but developers would figure out pretty quick there stuff's not going to work.

dbatiste (approved):
Looks good!

We should probably update the PR description with the name of the item component.

PR #4712, opened 2024-05-31
feat: Table Styles: Support other text alignments

margaree:
Related to [backport](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6126)
This is one option for this issue.

**Problem:**
Some LMS tables use `text-align: center` and `text-align: right` (this one is more in theory) on table headers. With the `table-col-sort-button` changes, it's currently only start aligned and the `text-align: center` on the `th` does nothing. We do want to support the centering.

**Solution Notes:**
Adds a new `position` property to `d2l-table-col-sort-button` which then sets `justify-content` and the padding.

**Previously considered options:**
**_Option 1:_** Two new css vars are exposed and then consumers can override them to get the desired alignment. The problem with this is the override isn't totally simple.

**_Option 1b:_** We could also potentially just set the padding based on if `--d2l-table-col-sort-button-justify-content` is set and its value so that would simplify this.

**_Option 2:_** A a new `position` property to `d2l-table-col-sort-button` which then sets `justify-content` and the padding. Con of this is in the LMS getting the alignment through from `grid` to `gridHeaderSortField` in order to set `position` on this component might not be straightforward. Pro is that consumers don't have to know the padding values, and if they change then we don't have to update them in consumers.

**_Option 3:_** In `d2l-table-col-sort-button` reach out to the parent and see if `text-align` is set then update padding and such within the component based on that. Pros are we don't have to do anything in the consumer code. Cons is that this is a bit hacky and if there are other ways centering is being done we'd have to account for them.


dbatiste:
Option 1a and 1b expose the internal (`inline-flex`) layout implementation detail, so I'm not a super huge fan of those. Will it always be `inline-flex`? Maybe, but if it changes, then the consumer's alignment stops working.

Option 2 keep the internal implementation hidden, and consumers don't need to do anything special. I think I prefer this approach. Which LMS Grid control are you concerned about?

Option 3 requires reaching out to the parent and calling `getComputedStyle(...)`, and then potentially re-rendering the component depending on what it computes. It also forces the user to specify a CSS property on the `<th>` that doesn't really work the way they think. Not really in love with this.

Other option - we specify `justify-content: inherit` on the internal `button`, and then the consumer can specify `justify-content: ...` on `d2l-table-col-sort-button` to control the alignment. But maybe there'd still be something to be done with padding? I dislike this option for the same reason as 1a & 1b though.

dbatiste:
> Which LMS Grid control are you concerned about?

If it is legacy grid [here](https://github.com/Brightspace/lms/blob/5463bc4660d52d63af4dcd23bc687e9952d1ac45/lp/framework/web/D2L.Web/ui/grid/gridHeaderSortField.cs#L112), then I think we would just check `gc.CurrentHeaderCell.Align`, but I am not 100% without running it.

margaree:
> > Which LMS Grid control are you concerned about?
> 
> If it is legacy grid [here](https://github.com/Brightspace/lms/blob/5463bc4660d52d63af4dcd23bc687e9952d1ac45/lp/framework/web/D2L.Web/ui/grid/gridHeaderSortField.cs#L112), then I think we would just check `gc.CurrentHeaderCell.Align`, but I am not 100% without running it.


It is legacy grid. I was worried we didn't have the header cell info but yeah I also just came to the same conclusion - that should work. Okay yeah I agree with your assessment. I'll pursue option 2 with a bit more testing LMS-side to make sure it will work.

dbatiste (approved):
Looks great!

PR #4716, opened 2024-06-03
MAN-2993: Promote z-index for scroll-wrapper-actions

semibran:
[MAN-2993](https://desire2learn.atlassian.net/browse/MAN-2993)
Related to https://d2l.slack.com/archives/C04511G83KR/p1717183794167289

The `sticky-headers-scroll-wrapper` attribute is causing issues with tooltip display in mastery view.
One possible workaround is to instead [manually wrap the d2l-table in a scroll-wrapper](https://github.com/Brightspace/d2l-outcomes/pull/1287). This appears to work for our particular use case.

However since both the table sticky headers and the scroll-wrapper actions both have [z-index: 4](https://github.com/BrightspaceUI/core/pull/3635), the sticky headers take precedence in the render order, blocking out the left-side chevron:

image

Promoting the z-index for the scroll-wrapper-actions is the most straightforward (naive?) way to solve the issue.

DISCLAIMER: This is not an ideal solution. If we think it's risky/unnecessary for these changes to apply globally we can just apply some of these z-index overrides at the consumer-level (though this does end up being a little less pretty)

[MAN-2993]: https://desire2learn.atlassian.net/browse/MAN-2993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/table/table-controls.js
+++ b/components/table/table-controls.js
@@ -20,7 +20,7 @@ class TableControls extends SelectionControls {
 			:host {
 				--d2l-selection-controls-background-color: var(--d2l-table-controls-background-color);
 				--d2l-selection-controls-shadow-display: var(--d2l-table-controls-shadow-display);
-				z-index: 5; /* Must be greater than d2l-table-wrapper and d2l-scroll-wrapper */
+				z-index: 6; /* Must be greater than d2l-table-wrapper and d2l-scroll-wrapper */
 			}
 			:host([no-sticky]) {
 				z-index: auto;
--- a/components/scroll-wrapper/scroll-wrapper.js
+++ b/components/scroll-wrapper/scroll-wrapper.js
@@ -124,7 +124,7 @@ class ScrollWrapper extends RtlMixin(LitElement) {
 				position: -webkit-sticky;
 				position: sticky;
 				top: var(--d2l-table-sticky-top, 0);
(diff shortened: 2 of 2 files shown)


components/table/table-controls.js
```diff
@@ -20,7 +20,7 @@ class TableControls extends SelectionControls {
 			:host {
 				--d2l-selection-controls-background-color: var(--d2l-table-controls-background-color);
 				--d2l-selection-controls-shadow-display: var(--d2l-table-controls-shadow-display);
-				z-index: 5; /* Must be greater than d2l-table-wrapper and d2l-scroll-wrapper */
+				z-index: 6; /* Must be greater than d2l-table-wrapper and d2l-scroll-wrapper */
```
semibran, line 23: https://search.d2l.dev/xref/BrightspaceUI/core/components/list/list-controls.js?r=e3fee725#22
Potential conflict

dbatiste:
This seems reasonable to me.

Yeah, rather not have work-arounds in the app code, since they'll just make it harder for us later when we want to make changes.

dbatiste (approved):

PR #4717, opened 2024-06-04
fix: Table Updates: Document recent changes [GAUD-6129]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6129)

This documents the usage of multi-faceted table sorting. The code ends up a bit verbose but I think overall that would be helpful to consumers. Let me know thoughts on this.


components/table/table-col-sort-button-item.js
```diff
@@ -6,6 +6,9 @@ import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
 /**
  * A radio menu item to be used within the d2l-table-col-sort-button component for a multi-faceted sort.
+ * @fires d2l-menu-item-change - Internal event
+ * @fires d2l-menu-item-select - Internal event
+ * @fires d2l-menu-item-visibility-change - Internal event
```
margaree, line 11: These get pulled in through the `MenuItemRadioMixin` but I believe we only really want the `d2l-table-col-sort-button-item-change` event to be documented, so this is the way to hide them.

components/table/README.md
```diff
@@ -228,7 +232,127 @@ When tabular data can be sorted, the `<d2l-table-col-sort-button>` can be used t
...
+
+This is a radio menu item to be used within the `d2l-table-col-sort-button` component for a [multi-faceted sort](#multi-faceted-sort-button).
+
+<!-- docs: demo code properties autoSize:false align:start name:d2l-table-col-sort-button-item -->
+```html
+<script type="module">
+  import '@brightspace-ui/core/components/table/table-col-sort-button.js';
+  import '@brightspace-ui/core/components/table/table-col-sort-button-item.js';
+</script>
+<d2l-table-col-sort-button style="--d2l-table-cell-padding: 15px;" desc>
```
dbatiste, line 253: Is the custom padding needed here?

dbatiste (approved):

components/table/README.md
margaree: It is, otherwise there is no padding on the col-sort-button component. Thoughts on if the component should have a default fallback as well?

dbatiste: I don't think I fully understand the padding usage scenarios. My hope would be that they can just use `d2l-table-col-sort-button` out of the box. If they have to provide their own padding, are we setting ourselves up for inconsistency or maintenance headaches down the road?

margaree: It's currently just dependant on `--d2l-table-cell-padding` as there is the assumption that this is used within a table that is using the table styles. It can be used out of the box for a regular consumer who puts it in a table, I just didn't do that in this example.

PR #4719, opened 2024-06-05
fix: List: Disable radio selection default behaviour when in grid mode [GAUD-6514]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6514)

When a `d2l-list` has `selection-single` and is in `grid` mode, the arrow key action between radio buttons does not work particularly well. This is due to the fact that the `SelectionMixin` is also trying to control arrow key interaction. This code change adds a property to not run that code in `SelectionMixin`. Another option would be to check if `d2l-list-item` is an ancestor within `SelectionMixin` but that felt too coupled with list.

If this seems like a good approach I'll see about adding some tests and also probably rename the property to something like `selectionIgnoreRadioKeyUp`.

Note that the general default browser behaviour for radio button groups is to auto-select on arrow key interaction, and this change completely disables that in list in grid mode (but it also was not working particularly well before).


components/selection/selection-mixin.js
```diff
@@ -42,6 +42,11 @@ export const SelectionMixin = superclass => class extends RtlMixin(CollectionMix
 
 	static get properties() {
 		return {
+			/**
```
dbatiste, line 45: I think we'd probably want to keep this hidden, since we would not want this showing up in `d2l-list`'s or `d2l-table-wrapper`'s documentation. 

Maybe name it something more semantic (maybe `selectionNoInputArrowKeyBehaviour` or something along those lines), in case the `keydown` handler ever needed to do something besides what it currently does.

dbatiste (approved):

PR #4725, opened 2024-06-07
Add basic fixed positioning (no edge cases) for dropdown components.

dbatiste:
[GAUD-6452](https://desire2learn.atlassian.net/browse/GAUD-6452)

This PR adds basic fixed positioning of dropdown content. This only handles the initial positioning. It does not handle any of the edge cases (scrolling, DOM mutations, etc) - there will be subsequent PRs to address those cases.

I need to do some more testing with this in different containing blocks to see how it behaves. Mostly just want to get 👀 on the initial positioning in case there's something major I am missing.

[GAUD-6452]: https://desire2learn.atlassian.net/browse/GAUD-6452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/dropdown/demo/dropdown-fixed.html
dbatiste, line 1: I am going to expand on what's in this page and eventually clean it up. This file can be ignored for the time being.

components/dropdown/demo/dropdown-fixed.html
```diff
@@ -0,0 +1,171 @@
...
+
+		<h2>Dropdown (in a dialog)</h2>
+
+		<d2l-demo-snippet>
+			<template>
+				<d2l-button id="openDialog1">Show Dialog</d2l-button>
+				<d2l-dialog id="dialog1" title-text="Dialog Title">
+					<div>
+						<p>Deadlights jack lad schooner scallywag dance the hempen jig carouser broadside cable strike colors. Bring a spring upon her cable holystone blow the man down spanker</p>
+						<d2l-dropdown prefer-fixed-positioning>
```
dbatiste, line 144: Currently, the consumer would be required to provide `prefer-fixed-positioning` on both the opener and the content element. We can look at reducing this to just one by doing some DOM checking of the corresponding opener/content if need be.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -7,6 +7,7 @@ import { getComposedActiveElement, getFirstFocusableDescendant, getPreviousFocus
 import { classMap } from 'lit/directives/class-map.js';
 import { html } from 'lit';
 import { LocalizeCoreElement } from '../../helpers/localize-core-element.js';
+import ResizeObserver from 'resize-observer-polyfill/dist/ResizeObserver.es.js';
```
dbatiste, line 10: I think this was a missing import.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -678,20 +711,20 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
 
 			if (!ignoreVertical) {
 				this.openedAbove = this._getOpenedAbove(spaceAround, spaceAroundScroll, spaceRequired);
 			}
 
-			const position = this._getPosition(spaceAround, targetRect.width, contentRect.width);
-			if (position !== null) {
-				this._position = position;
-			}
+			this._position = this._getPosition(spaceAround, targetRect, contentRect);
```
dbatiste, line 726: `_getPosition()` has been updated to return an object with the position properties. Previously it was just returning the relative `left` value, but for fixed positioning, it will provide `top`, etc.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -978,8 +1011,60 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		return false;
 	}
 
-	_getPosition(spaceAround, targetWidth, contentWidth) {
-		const centerDelta = contentWidth - targetWidth;
+	_getPointerPosition(targetRect) {
```
dbatiste, line 1014: This is completely new, since with relative positioning, calculating the position of the pointer was simply CSS.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -978,8 +1011,60 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+		if (this.openedAbove) {
+			position.bottom = window.innerHeight - targetRect.top + 8;
+		} else {
+			position.top = targetRect.top + targetRect.height + this._verticalOffset - 7;
+		}
+
+		return position;
+	}
+
+	_getPosition(spaceAround, targetRect, contentRect) {
```
dbatiste, line 1037: As mentioned, this has been updated to return an object, rather than just the left adjustment value. The non-fixed positioning value remains the same.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1064,15 +1149,6 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 	}
 
 	_renderContent() {
-		const positionStyle = {};
```
dbatiste, line 1067: This was moved below and adjusted to be more dynamic, since we may be applying more than just the left/right position.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1131,16 +1207,34 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+		`;
+
+		const pointerPositionStyle = {};
+		if (this._pointerPosition) {
+			for (const prop in this._pointerPosition) {
+				pointerPositionStyle[prop] = `${this._pointerPosition[prop]}px`;
+			}
+		}
+
+		const pointer = html`
```
dbatiste, line 1235: The pointer was moved into this mixin from the subclass `render` methods since the pointer position needs to be calculated and rendered in the same way.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -220,6 +234,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			_contentHeight: {
 				type: Number
 			},
+			_pointerPosition: {
+				state: true
+			},
 			_position: {
 				type: Number
```
margaree, line 241: `_position` should probably have its type updated

margaree (approved):
This approach generally seems fine to me.
Have you tried getting the vdiff goldens without `prefer-fixed-positioning` and then adding it to all the tests to see what happens? I've found that helpful for changes like this in the past.

components/dropdown/dropdown-content.js
```diff
@@ -18,9 +18,6 @@ class DropdownContent extends DropdownContentMixin(LitElement) {
 	render() {
 		return html`
 			${this._renderContent()}
-			<div class="d2l-dropdown-content-pointer">
-				<div></div>
-			</div>
```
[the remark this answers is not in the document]
dbatiste: Oh good point... Thx!

dbatiste:
>Have you tried getting the vdiff goldens without prefer-fixed-positioning and then adding it to all the tests to see what happens?

Not yet, but I was thinking about doing that. Obviously I'll also eventually want to add some diffs for the use cases that these changes aim to improve as well, so that's not represented here in this PR either.

dbatiste:
So I did run the the diffs with the new positioning against the existing goldens.

The good news is that the content is positioned as we expect, and overflows in the constrained cases. The bad news is that my logic for positioning the pointer needs some work to handle all the cases. I am going to try and improve that logic for this PR.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -1007,11 +1103,11 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 			}
 		}
 		if (this.align === 'start' || this.align === 'end') {
-			const shift = Math.min((targetWidth / 2) - (20 + 16 / 2), 0); // 20 ~= 1rem, 16 = pointer size
+			const shift = Math.min((targetRect.width / 2) - (20 + 16 / 2), 0); // 20 ~= 1rem, 16 = pointer size
```
margaree, line 1106: super nit but since you added `pointerLength` that could be used for this `16` here (and the 20 could also be a variable as well since it's used here and now in `_getPointerPosition`)

margaree (approved):

components/dropdown/dropdown-content-mixin.js
dbatiste: Sounds good. I will tidy that up with one of my next PRs.

dbatiste: So I decided to at least update to use `pointerLength`. There appears to be other usage of `20` in the code, so I think I want to look at those as well - will do that separatly.

PR #4726, opened 2024-06-10
Date Filters: New component for Pre-defined ranges [GAUD-6470]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6470)

This creates a new component for filter options that are pre-defined date ranges (e.g., "last 30 days"). The accepted values are based on commonly used options seen elsewhere and are also meant to remove ambiguity in certain cases (e.g., "Last 24 hours" rather than "Last day").

There are still some questions around the properties so I'll add those as needed in the PR comments.

To do:
- [x] Update README and Daylight documentation after approach is settled on


helpers/dateTime.js
```diff
@@ -161,6 +161,44 @@ export function getUTCDateTimeFromLocalDateTime(date, time) {
 	return formatDateTimeInISO(utcDateTime);
 }
 
+export function getUTCDateTimeRange(type, diff) {
```
margaree, line 164: I added a new helper since this functionality was used in the new -value component as well as in filter.

helpers/dateTime.js
```diff
@@ -161,6 +161,44 @@ export function getUTCDateTimeFromLocalDateTime(date, time) {
...
+		const today = formatDateInISO(getToday());
+		const startValue = getUTCDateTimeFromLocalDateTime(today, '0:0:0');
+		const endValue = getUTCDateTimeFromLocalDateTime(today, '23:59:59');
+		return { startValue, endValue };
+	}
+
+	/**
+	 * endValue = now in UTC string
+	 * startValue = now minus number of hours/days/weeks/months in diff then converted to UTC string
+	 */
```
margaree, line 179: Please confirm my assumption/logic with this:
If we get `new Date()` in ISO string, that would be the UTC time for now and timezone wouldn't matter (the cases where this usually gets complicated is when the user has a different LMS timezone set from their browser's timezone, which was what we handled quite a bit with date-time inputs). We therefore are safe to just use that as the end value, then get the start value, without having to consider the user's timezone.

helpers/dateTime.js
```diff
@@ -161,6 +161,44 @@ export function getUTCDateTimeFromLocalDateTime(date, time) {
...
+export function getUTCDateTimeRange(type, diff) {
+	if (!type || (!diff && diff !== 0)) return {};
+	if (type !== 'hours' && type !== 'days' && type !== 'weeks' && type !== 'months') return {};
+
+	if (type === 'days' && diff === 0) {
+		// assume "today" (midnight in user's time to 23:59:59 in user's time)
+		const today = formatDateInISO(getToday());
+		const startValue = getUTCDateTimeFromLocalDateTime(today, '0:0:0');
+		const endValue = getUTCDateTimeFromLocalDateTime(today, '23:59:59');
+		return { startValue, endValue };
```
margaree, line 173: This uses `getToday` which gets the local date based on the user's LMS timezone, then gets the UTC value of that date and midnight for start value and at 23:59:59 for end value. Confirmed with Jeff that these are the values we want even though this is the only preset that would involve future times.

components/filter/filter-dimension-set-date-text-value.js
```diff
@@ -0,0 +1,150 @@
...
+			endValue: { type: String },
+			/**
+			 * REQUIRED: Unique key to represent this value in the dimension
+			 * @type {string}
+			 */
+			key: { type: String },
+			/**
+			 * @type {'today'|'lastHour'|'24hours'|'48hours'|'7days'|'14days'|'30days'|'6months'}
+			 */
+			range: { type: String },
```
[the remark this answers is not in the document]
dbatiste: Or how about `yesterday`? Also, if we provide two properties, wouldn't they be able to provide all sorts of permutations that we don't support? I'm uncertain at this point.

components/filter/filter-dimension-set-date-text-value.js
```diff
@@ -0,0 +1,149 @@
...
+			disabled: { type: Boolean, reflect: true },
+			/**
+			 * @ignore
+			 */
+			endValue: { type: String },
+			/**
+			 * REQUIRED: Unique key to represent this value in the dimension
+			 * @type {string}
+			 */
+			key: { type: String },
```
dbatiste, line 27: Does the key serve any purpose to us or the consumer if they are using `rangeNum` & `rangeType`?

components/filter/filter-dimension-set.js
```diff
@@ -142,7 +142,10 @@ class FilterDimensionSet extends LitElement {
 				disabled: value.disabled,
 				key: value.key,
 				selected: value.selected,
-				text: value.text
+				text: value.text,
+				type: value.type,
+				rangeNum: value.rangeNum,
```
dbatiste, line 147: Can we have the consumer specify `<d2l-filter-dimension-set-date-text-value key="lastHour">`? Then the filter dimension set doesn't need to know about these special properties. Or, maybe we can have the item components implement a common method that could be used here, so that this doesn't need to be aware of the different item types.

margaree: I'd be concerned that if they just used `key` they might not realize that `key` has to be specific values. With implementing the custom range there become more things that need to be passed through, so I think I favour the common method idea.

components/filter/filter-dimension-set-date-text-value.js
```diff
@@ -0,0 +1,149 @@
...
+
+		if (this.selected) {
+			// if the value is initially selected, startValue and endValue should be set in case used by consumer
+			const dateTimeRange = getUTCDateTimeRange(this.rangeType, this.rangeNum);
+			this.startValue = dateTimeRange.startValue;
+			this.endValue = dateTimeRange.endValue;
+		}
+	}
+
+	updated(changedProperties) {
```
dbatiste, line 80: It seems like these computed properties could be set earlier in the life cycle with: 
```javascript
	...
}
```
I'm not sure if it really provides much benefit for filter though, given that filter does the actual rendering vs this component. Nevertheless, I'd opt to move those earlier if possible.

components/filter/filter-dimension-set-date-text-value.js
```diff
@@ -0,0 +1,149 @@
...
+			/** @ignore */
+			this.dispatchEvent(new CustomEvent('d2l-filter-dimension-set-value-data-change', {
+				detail: { valueKey: this.key, changes: changes },
+				bubbles: true,
+				composed: false
+			}));
+		}
+	}
+
+	_handleRange() {
```
dbatiste, line 109: Nit: maybe name this `_handleRangeUpdated`?

helpers/dateTime.js
```diff
@@ -161,6 +161,44 @@ export function getUTCDateTimeFromLocalDateTime(date, time) {
 	return formatDateTimeInISO(utcDateTime);
 }
 
+export function getUTCDateTimeRange(type, diff) {
+	if (!type || (!diff && diff !== 0)) return {};
```
dbatiste, line 165: Would positive and negative values be valid?

helpers/dateTime.js
```diff
@@ -161,6 +161,44 @@ export function getUTCDateTimeFromLocalDateTime(date, time) {
...
+	if (type !== 'hours' && type !== 'days' && type !== 'weeks' && type !== 'months') return {};
```
dbatiste, line 166: Should this helper support more timeframes from the get-go? Thinking of `years`, `minutes`, `seconds`.

helpers/dateTime.js
dbatiste: I _think_ so.

helpers/dateTime.js
margaree: With the current implementation no, since negative values would mix up which date would be `startValue` and which would be `endValue`. I'll look into improving this though or at least making it more clear how to use this.

dbatiste: Currently for `lastHour`, filter calls: 
```javascript
const dateTimeRange = getUTCDateTimeRange('hours', 1);
```
But maybe the consumer should use negative for ranges in the past, and positive for ranges in the future.
```javascript
// lastHour
const dateTimeRange = getUTCDateTimeRange('hours', -1);

// nextHour
const dateTimeRange = getUTCDateTimeRange('hours', 1);
```
I know filter doesn't have the `nextHour` case, but perhaps we should set this helper to be more generic in that sense, since it has potential to be used by anything. That's a decision we'd want to make now.

This helper assumes the reference date is _now_. Maybe that's ok. In the future we would accept option for the consumer to specify the reference date if necessary.

Thoughts?

margaree: I agree - that seems much more flexible and future proof. Will update

components/filter/filter-dimension-set-date-text-value.js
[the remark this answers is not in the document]
margaree: My thought (ran this by Jeff as well) is that for now we just expose `range` and do not document any ability to use `rangeType` or `rangeNum`. This way we're not allowing semi-unlimited possibilities, and we can see if there is any use case for consumers to use these properties (or perhaps if there's another common range we can just add that into our `range` options).
  
I can either leave `rangeType` and `rangeNum` in the code but not document them, or if we really want them not to be used I can remove them from properties and just have the necessary things happen when `range` changes.

lang/ar.js
```diff
@@ -25,6 +25,9 @@ export default {
 	"components.filter.searchResults": "{number, plural, =0 {ما من نتائج بحث} one {{number} نتيجة بحث}‏ other {{number} من نتائج البحث}}",
 	"components.filter.selectedFirstListLabel": "{headerText}. تظهر عوامل التصفية المحددة أولاً.",
 	"components.filter.singleDimensionDescription": "التصفية حسب: {filterName}",
+	"components.filter-dimension-set-date-text-value.textHours": "{num, plural, one {Last hour} other {Last {num} hours}}",
+	"components.filter-dimension-set-date-text-value.textDays": "{num, plural, =0 {Today} one {Last {num} days} other {Last {num} days}}",
+	"components.filter-dimension-set-date-text-value.textMonths": "Last {num} months",
```
margaree, line 30: I had initially used the nested type and just one lang term, but there wasn't really an "other" case so I divided them.

lang/en.js
```diff
@@ -25,6 +25,9 @@ export default {
 	"components.filter.searchResults": "{number, plural, =0 {No search results} one {{number} search result} other {{number} search results}}",
 	"components.filter.selectedFirstListLabel": "{headerText}. Selected filters appear first.",
 	"components.filter.singleDimensionDescription": "Filter by: {filterName}",
+	"components.filter-dimension-set-date-text-value.textHours": "{num, plural, one {Last hour} other {Last {num} hours}}",
+	"components.filter-dimension-set-date-text-value.textDays": "{num, plural, =0 {Today} one {Last {num} days} other {Last {num} days}}",
```
margaree, line 29: The LocalizeMixin sandbox auto-adds the `one` case for several languages (I just had `=0` and `other`). We don't currently have a case for one, though it would probably be "yesterday" if there was anything. Not sure if I should leave as-is or change.

components/filter/filter-dimension-set-date-text-value.js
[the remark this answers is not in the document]
dbatiste: I'm good either way, but yeah, I probably lean that way too.

components/filter/filter-dimension-set.js
```diff
@@ -173,7 +167,16 @@ class FilterDimensionSet extends LitElement {
 	_getSlottedNodes() {
 		if (!this._slot) return [];
 		const nodes = this._slot.assignedNodes({ flatten: true });
-		return nodes.filter((node) => node.nodeType === Node.ELEMENT_NODE && node.tagName.toLowerCase() === 'd2l-filter-dimension-set-value');
+		return nodes.filter((node) => {
+			if (node.nodeType !== Node.ELEMENT_NODE) return false;
+
+			const hasDateFilter = node.tagName.toLowerCase() === 'd2l-filter-dimension-set-date-text-value';
```
dbatiste, line 173: We have these types tightly under wraps - it's not like anyone can just create their own value types. Nevertheless, it feels a little off to have `d2l-filter-dimension-set` and `d2l-filter` be aware of the different value types. To avoid that, at least for this bit of code, we'd need something to check on the node to identify it as a value item for filtering, and we'd also need a way for them to indicate whether they support searching and multiple selection. Thoughts?

components/filter/filter.js
```diff
@@ -860,7 +861,15 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 			this._totalAppliedCount--;
 		}
 
-		this._dispatchChangeEvent(dimension, { valueKey: valueKey, selected: selected });
+		const details = { valueKey: valueKey, selected: selected };
+
+		if (value.type === 'date' && selected) {
+			const dateTimeRange = getUTCDateTimeRange(value.rangeType, value.rangeNum);
+			details.startValue = dateTimeRange.startValue;
+			details.endValue = dateTimeRange.endValue;
```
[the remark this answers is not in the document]
dbatiste: Same thing with my comment above about this code being aware of the different value types. As mentioned above though, maybe this isn _ok_ given that we only have the two types and they are controlled by us.

dbatiste (approved):
This is looking good. 

As discussed separately, if you decide to make further changes to extract the date-specific bits out of the dimension set and filter, let me know and I will take another look.

components/filter/filter-dimension-set-date-text-value.js
```diff
@@ -0,0 +1,156 @@
...
+
+	getValueDetails() {
+		this._handleRangeUpdated();
+
+		return {
+			disabled: this.disabled,
+			key: this.key,
+			selected: this.selected,
+			text: this.text,
+			getAdditionalEventDetails: this._getAdditionalEventDetails.bind(this)
```
dbatiste, line 108: I guess we can't just provide `startValue` and `endValue` here eh (i.e. instead of a method)?

margaree: We want `startValue` and `endValue` to be up to date as of when the user selects them. If we set them here, they would only be up to date when `getValueDetails` runs, which currently is only on slot change. These in theory could be fairly different times depending on the length of time between the slot change and selection.

dbatiste (approved):

PR #4729, opened 2024-06-10
Update html-block to use new context provider framework

grant-cleary:
**NOTE:** ~This can't merge until https://github.com/Brightspace/brightspace-integration/pull/12033 is merged.~ It should also be merged in tandem with https://github.com/BrightspaceHypermediaComponents/users/pull/389.

This PR looks to switch over the html-block to the new context provider framework for its renderers. In terms of the changes themselves, there's not _too_ much going on here; mostly just removing the use of the attribute controller and updating to use the new provider.

This can't merge until 20.24.8, as it's going to break [this copy of the component](https://github.com/BrightspaceUI/creator-plus-authoring-tools/blob/main/components/insert-element/learner/lit/core/components/html-block/html-block.js). I figure saving this for 20.24.8 gives a good amount of time for that to be removed first.


grant-cleary:
... I definitely forgot to commit the Prism renderer changes. One moment please.

grant-cleary:
Hmm, I'm not sure if I broke some dependencies when I updated the package-lock file. Maybe I brought something along that I shouldn't have.

grant-cleary:
Weird. I just added the new dependency and `npm i`'d, but yeah, I see a bunch of rollup changes in there, which would match with what the test errors are complaining about. I'll undo the package-lock changes and just surgically add the new dependency.

dlockhart (approved):
Nice!

components/html-block/demo/html-block.html
```diff
@@ -46,9 +47,6 @@
 			});
 
 		</script>
-		<script type="module">
-			import '../../../tools/mathjax-test-context.js';
```
grant-cleary, line 50: This was happening too late in the process, so context was being added _after_ first render. The old attribute controller didn't care about that mis-ordering, but the context provider absolutely does.

helpers/mathjax.js
```diff
@@ -32,16 +32,14 @@ const mathjaxFontMappings = new Map([
...
+	get contextKeys() {
+		return [mathjaxContextKey];
 	}
 
 	async render(elem, options) {
 		if (!options.contextValues) return elem;
-		const contextVal = options.contextValues.get(mathjaxContextAttribute);
-		if (contextVal === undefined) return elem;
-
-		const context = JSON.parse(contextVal) || {};
```
grant-cleary, line 44: No more strings means no more sketchy JSON parsing. :)

grant-cleary:
Hmm, the vdiff failures are unexpected. The html-block vdiffs don't bother setting any of the context stuff so I expected them to remain unchanged. I suspect they must be timing out waiting for the context to resolve, though I'm unsure why. I'll take a peek locally and see if I can make sense of it.

grant-cleary:
Locally, one vdiff test was still reliably failing after these changes. I _think_ this may just be a local vs. CI or a browser thing, so I'm going to see if CI fails again before I poke any further.

PR #4733, opened 2024-06-11
Reposition the fixed position dropdown when scrolling.

dbatiste:
[GAUD-6453](https://desire2learn.atlassian.net/browse/GAUD-6453)

Reposition the fixed position dropdown content when scrolling. Scrolling could occur either on the window/document, or a nested element.

[GAUD-6453]: https://desire2learn.atlassian.net/browse/GAUD-6453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/dropdown/dropdown-content-mixin.js
```diff
@@ -316,6 +317,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this.mediaQueryList = window.matchMedia(`(max-width: ${this.mobileBreakpointOverride - 1}px)`);
 		this._useMobileStyling = this.mediaQueryList.matches;
 		if (this.mediaQueryList.addEventListener) this.mediaQueryList.addEventListener('change', this._handleMobileResize);
+		if (this.opened) this.__addRepositionHandlers();
```
dbatiste, line 320: We only really want the reposition handlers wired up while the dropdown is open. These `__addRepositionHandlers` and `__removeRepositionHandlers` in the connected callbacks serve to handle the cases where the dropdown is added/removed to/from the DOM while open.

components/dropdown/demo/dropdown-fixed.html
dbatiste, line 1: I replaced this demo page with a new one that is cleaner and better supports side-by-side comparison with the old positioning logic.

components/dropdown/demo/dropdown-positioning.html
dbatiste, line 1: To view this demo page using the new fixed positioning logic, simply add the `?position=fixed` to the query-string.

margaree (commented):
This generally LGTM. I did play around with it a bit and noticed that when scrolling the dropdown will entirely reposition itself such that, for example, if it was initially opened above but then because of scrolling there becomes more room below then it will move below, which I don't think is the case with the non-fixed positioning, and it can be a bit jarring. Is it possible to preserve the direction the dropdown opened in?

dbatiste:
> Is it possible to preserve the direction the dropdown opened in?

I think so. I'll try to make that work.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -755,6 +789,27 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
...
+			node.removeEventListener('scroll', this.__reposition);
+		});
+		this._scrollablesObserved = null;
+	}
+
+	__reposition() {
+		// throttle repositioning (https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event#scroll_event_throttling)
+		if (!this.__repositioning) {
+			requestAnimationFrame(() => {
+				this.__position(true);
```
dbatiste, line 806: This `true` tells it to ignore vertical when repositioning the content, so that it's not flipping as the user scrolls.

If we later decide we actually _do_ want it to flip, we can simply remove this `true` from here.

margaree (approved):

PR #4736, opened 2024-06-12
Reposition fixed position dropdown when DOM mutations occur in the ancestor path

dbatiste:
[GAUD-6454](https://desire2learn.atlassian.net/browse/GAUD-6454)

This PR updates the fixed position dropdown to reposition itself when it observes DOM mutations in its ancestor path. It wires up the observer in each DOM scope, however it _does not_ wire-up to sibling scopes.

[GAUD-6454]: https://desire2learn.atlassian.net/browse/GAUD-6454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -285,6 +285,7 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this._verticalOffset = defaultVerticalOffset;
 
 		this.__reposition = this.__reposition.bind(this);
+		this.__onAncestorMutation = this.__onAncestorMutation.bind(this);
 		this.__onResize = this.__onResize.bind(this);
 		this.__onAutoCloseFocus = this.__onAutoCloseFocus.bind(this);
 		this.__onAutoCloseClick = this.__onAutoCloseClick.bind(this);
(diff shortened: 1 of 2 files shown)


dbatiste:
I was thinking about this last night... we probably want to ignore mutations that come from within the dropdown that's observing. Gonna see if I can add that, otherwise we could be running the positioning logic excessively as the user interacts with the dropdown contents.

components/dropdown/dropdown-content-mixin.js
```diff
@@ -528,6 +541,13 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		this._hasHeader = e.target.assignedNodes().length !== 0;
 	}
 
+	__onAncestorMutation(mutations) {
+		const opener = this.__getOpener();
+		// ignore mutations that are within this dropdown
+		const reposition = !!mutations.find(mutation => !isComposedAncestor(opener, mutation.target));
```
margaree, line 547: Potentially a silly question - is it possible to not observe these cases initially rather than having to filter them here?

dbatiste: I was thinking about that a bit... 

We're using the `{ attributes: true, childList: true, subtree: true }` config so that we'll be notified of any changes in that DOM scope, including the addition/removal of siblings (and/or their sub-tree), or attribute changes that might cause their styles to change. 

We might be able to avoid the filtering here if we use `subtree: false`, but then we'd have to walk this scope's tree and wire-up the mutation observer to each node, excluding the ones in the dropdown's subtree. I think there would also be the added complexity that if any of those nodes were removed, we'd want to stop observing them.

I think it becomes a trade off between a bit of DOM walking only when a mutation occurs (for the single observer per scope) vs DOM walking any time the dropdown opens + observing a lot more nodes.

margaree (approved):

PR #4739, opened 2024-06-13
Hide/show the fixed position dropdown depending on whether the opener is in view.

dbatiste:
[GAUD-6455](https://desire2learn.atlassian.net/browse/GAUD-6455)

This PR updates the fixed position dropdown to move offscreen when the opener is not in view. For instance, when the opener is scrolled out of view.

[GAUD-6455]: https://desire2learn.atlassian.net/browse/GAUD-6455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -5,7 +5,7 @@ const intersectionObserver = new IntersectionObserver(entries => {
 	entries.forEach(entry => {
 		entry.target.__updateContentVisibility(entry.isIntersecting);
 	});
-}, { threshold: 0 });
+}, { threshold: 0 }); // 0-1 (0 -> intersection requires any pixel visible, 1 -> intersection requires all pixels visible)
 
 export const DropdownOpenerMixin = superclass => class extends superclass {
 
--- a/components/dropdown/dropdown-opener-mixin.js
+++ b/components/dropdown/dropdown-opener-mixin.js
@@ -97,7 +97,7 @@ export const DropdownOpenerMixin = superclass => class extends superclass {
 		this.addEventListener('mouseenter', this.__onMouseEnter);
 		this.addEventListener('mouseleave', this.__onMouseLeave);
 
-		if (this.dropdownOpened) {
+		if (this._fixedPositioning && this.dropdownOpened) {
 			intersectionObserver.observe(this);
 		}
 		if (this.openOnHover) {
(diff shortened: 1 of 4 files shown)


components/dropdown/dropdown-opener-mixin.js
```diff
@@ -1,6 +1,12 @@
 import { getUniqueId } from '../../helpers/uniqueId.js';
 import { isComposedAncestor } from '../../helpers/dom.js';
 
+const intersectionObserver = new IntersectionObserver(entries => {
+	entries.forEach(entry => {
+		entry.target.__updateContentVisibility(entry.isIntersecting);
+	});
+}, { threshold: 0 });
```
dbatiste, line 8: `threshold: 0` means the opener is considered intersecting if at least 1px is in view. `threshold: 1` would mean the entire opener would need to be in view.

components/offscreen/offscreen.js
```diff
@@ -1,7 +1,10 @@
 import { css, html, LitElement } from 'lit';
 import { RtlMixin } from '../../mixins/rtl/rtl-mixin.js';
 
-const offscreenStyleDeclarations = css`
+/**
```
dbatiste, line 4: I don't love exporting these like this, and I considered moving them into a generator method, like we've done for [typography styles](https://github.com/BrightspaceUI/core/blob/main/components/typography/styles.js#L245). It would be more flexible, but we need these applied to the dropdown content host element, and our CSS selector Regex validation used by those generated methods doesn't allow `:host` [for this reason](https://github.com/BrightspaceUI/core/pull/2698#discussion_r931527209). I could be convinced to go that route, but it opens up the possibility for more invalid scenarios, so I decided to take the simpler approach here, given that we keep these under wraps either way.

margaree (approved):
Code-wise this LGTM. Playing around with it, I find the sudden disappearing fairly abrupt. Not sure if you've run that by Jeff? I'm also good with this going in and having a future polish happen.

There's also a couple of these new pieces, like this one and the DOM mutation situation, that could potentially benefit from some visual diffs, if it's possible to enable the flag for just a few tests at this point.

dbatiste:
> Code-wise this LGTM. Playing around with it, I find the sudden disappearing fairly abrupt. Not sure if you've run that by Jeff? I'm also good with this going in and having a future polish happen.

Sounds good. I'll touch base with Jeff when he's back and if he has ideas we can tweak things.

> There's also a couple of these new pieces, like this one and the DOM mutation situation, that could potentially benefit from some visual diffs, if it's possible to enable the flag for just a few tests at this point.

Sound good. Maybe I'll put those together in a new PR that covers these cases (overflow cases, mutation, opener out of view).

PR #4744, opened 2024-06-17
feat: InteractiveMixin: Change how escape and tab are handled [GAUD-6474]

margaree:
Part of [Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6474)

Open to suggestions on this.
We want `d2l-input-date-time-range` and `d2l-input-date-range` to be within `InteractiveMixin` for usage in a filter list item. In order for the keyboard behaviour to work as expected, changes are needed to how escape and tab are handled.

Problems:
- Escape: When the date/time input dropdown is opened, close would exit the dropdown but also leave interactive mode, which we wouldn't want. To mitigate this I added a check for `e.target?.opened` but let me know if that's getting too specific with knowledge of our dropdown being involved.
- Tab: in date-time the tab key would normally take the user between inputs. The current behaviour is that it goes to the next tab stop. Note that previously if the user was within an interactive area and then pressed tab, they would go to the next tab stop, but with this change they exit the interactive area and focus goes back to the interactive area (i.e., focus ring is around the interactive area)

If you wish to test this out with the date-time range, it can be seen [https://live.d2l.dev/prs/BrightspaceUI/core/pr-4742/components/filter/demo/filter.html] in the last demo on the page ("Date Filter")

### DIFF
--- a/mixins/interactive/interactive-mixin.js
+++ b/mixins/interactive/interactive-mixin.js
@@ -9,7 +9,8 @@ import { RtlMixin } from '../rtl/rtl-mixin.js';
 
 const keyCodes = {
 	ENTER: 13,
-	ESCAPE: 27
+	ESCAPE: 27,
+	TAB: 9
 };
 
 export function isInteractiveDescendant(node) {
@@ -106,10 +107,15 @@ export const InteractiveMixin = superclass => class extends LocalizeCoreElement(
 	}
 
 	async _handleInteractiveKeyDown(e) {
-		if (this._interactive && e.keyCode === keyCodes.ESCAPE) {
+		if (!this._interactive) return;
+
+		if (e.keyCode === keyCodes.ESCAPE && !e.target?.opened) {
+			e.stopPropagation();
 			this._interactive = false;
 			await this.updateComplete;
 			this.shadowRoot.querySelector('.interactive-toggle').focus();
+		} else if (e.keyCode === keyCodes.TAB) {
+			e.stopPropagation();
 		}
 	}
 


mixins/interactive/interactive-mixin.js
```diff
@@ -106,10 +107,15 @@ export const InteractiveMixin = superclass => class extends LocalizeCoreElement(
 	}
 
 	async _handleInteractiveKeyDown(e) {
-		if (this._interactive && e.keyCode === keyCodes.ESCAPE) {
+		if (!this._interactive) return;
+
+		if (e.keyCode === keyCodes.ESCAPE && !e.target?.opened) {
```
dbatiste, line 112: The [DropdownContentMixin](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-content-mixin.js#L663) relies on [dismissible.js](https://github.com/BrightspaceUI/core/blob/main/helpers/dismissible.js) to help manage the dismiss behaviour. It works best when everything uses it because it manages a simple stack of callbacks - invoking the top callback when the user presses the `[Esc]` key. 

The `InteractiveMixin` currently doesn't use `dismissible.js`, but perhaps it could. It would be set up the same as dropdowns... `setDismissible` would be called when the use enters interactive mode (like we do when a dropdown is opened), and `clearDismissible` would be called when the user leaves interactive mode. I think that would avoid the need for `opened` stuff creeping in here.

Thoughts? I think `dismissible.js` existed before `InteractiveMixin`, and so I'd be surprised if this wasn't considered at the time. There might be a reason using it wasn't possible.

mixins/interactive/interactive-mixin.js
```diff
@@ -106,10 +107,15 @@ export const InteractiveMixin = superclass => class extends LocalizeCoreElement(
...
+			e.stopPropagation();
 			this._interactive = false;
 			await this.updateComplete;
 			this.shadowRoot.querySelector('.interactive-toggle').focus();
+		} else if (e.keyCode === keyCodes.TAB) {
```
dbatiste, line 117: I tried this out and it seems to work well.

mixins/interactive/interactive-mixin.js
margaree: Good idea! So far this looks to work well

mixins/interactive/interactive-mixin.js
```diff
@@ -98,25 +102,29 @@ export const InteractiveMixin = superclass => class extends LocalizeCoreElement(
...
-
-		if (e.keyCode === keyCodes.ESCAPE && !e.target?.opened) {
-			e.stopPropagation();
-			this._interactive = false;
-			await this.updateComplete;
-			this.shadowRoot.querySelector('.interactive-toggle').focus();
-		} else if (e.keyCode === keyCodes.TAB) {
-			e.stopPropagation();
-		}
+		if (this._interactive && e.keyCode === 9) e.stopPropagation();
```
dbatiste: Nit: maybe we should add a comment here to indicate which key this is... `// TAB`

dbatiste (approved):
LGTM!

PR #4747, opened 2024-06-17
feat: Date and time inputs: Add prefer-fixed-positioning attribute

margaree:
Part of [Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6474)

This adds a `prefer-fixed-positioning` attribute to all the date and time range components and passes it down to the base `input-date` and `input-time` components to be added to the dropdown-content and dropdown components in order to enable that behaviour when the flag is also true.

I added this to `input-time-range` as well even though it's not in scope for the filter work since it seems like it would be needed eventually.

Let me know if there would be a preferred way of doing this rather than passing it all the way through the components.

### DIFF
--- a/components/inputs/input-date-range.js
+++ b/components/inputs/input-date-range.js
@@ -93,6 +93,11 @@ class InputDateRange extends FocusMixin(SkeletonMixin(FormElementMixin(RtlMixin(
 			 */
 			minValue: { attribute: 'min-value', reflect: true, type: String },
 			/**
+			 * Temporary.
+			 * @ignore
+			 */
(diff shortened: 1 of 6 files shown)


dbatiste (approved):
Haha yeah, hopefully these attributes won't be around for very long.

An alternative would be to add some ikky code to `DropdownContentMixin` that would use `findComposedAncestor` to detect if it's within a `d2l-filter`, rather than using these temporary attributes. However, that precludes anything else from opting into this new positioning algo.

PR #4752, opened 2024-06-18
feat: Date Filter: Custom date-time range [GAUD-6474]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6474)

This adds the new `d2l-filter-dimension-set-date-time-range-value` component.

Will be done separately:
- `date-range`
- documentation
- backup solution for when flag is off


components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -0,0 +1,139 @@
...
+			}
+		});
+		if (changes.size > 0) {
+			/** @ignore */
+			this.dispatchEvent(new CustomEvent('d2l-filter-dimension-set-value-data-change', {
+				detail: { valueKey: this.key, changes: changes, dispatchChangeEvent: shouldDispatchChangeEvent },
+				bubbles: true,
+				composed: false
+			}));
+			this._dispatchFilterChangeEvent = false;
```
margaree, line 85: Generally the same as the other -value components with the addition of dispatchChangeEvent for start and end value

components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	constructor() {
+		super();
+		this.disabled = false;
+		this.selected = false;
+		this._dispatchFilterChangeEvent = false;
+		this._enforceSingleSelection = true;
+		this._filterSetValue = true;
+		this._minWidth = 375;
```
dbatiste, line 53: I'm always kinda torn about how these type members should be named... where we reference it internally by other components, but we don't want to expose it to other consumers.

dbatiste (approved):

PR #4754, opened 2024-06-18
Update fixed position dropdown so that the height is not updated as the user scrolls

dbatiste:
[GAUD-6558](https://desire2learn.atlassian.net/browse/GAUD-6558)

This PR updates the fixed position dropdown contents so that their container height is not updated as the user scrolls.

[GAUD-6558]: https://desire2learn.atlassian.net/browse/GAUD-6558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/dropdown/dropdown-content-mixin.js
+++ b/components/dropdown/dropdown-content-mixin.js
@@ -698,7 +698,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		}
 	}
 
-	async __position(ignoreVertical, contentRect) {
+	async __position(contentRect, options) {
+
+		options = Object.assign({ updateAboveBelow: true, updateHeight: true }, options);
 
 		const opener = this.__getOpener();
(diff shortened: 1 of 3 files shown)


components/dropdown/dropdown-content-mixin.js
```diff
@@ -698,7 +698,9 @@ export const DropdownContentMixin = superclass => class extends LocalizeCoreElem
 		}
 	}
 
-	async __position(ignoreVertical, contentRect) {
```
dbatiste, line 701: This has been updated to take an `options` argument to keep the API tidy. 

The `ignoreVertical` arg was renamed to `updateAboveBelow` for clarity since we're adding new option to ignore height, which also relates to the vertical dimension.

margaree (approved):

PR #4755, opened 2024-06-18
fix: prune redundant semantic-dom-diff

KearseTrevor:
https://desire2learn.atlassian.net/browse/GAUD-6570


bearfriend (approved):
:shipit:

dlockhart (approved):
Well that was an easy solution...

PR #4757, opened 2024-06-19
Update the table header z-index when a contained dropdown is opened/closed

dbatiste:
[GAUD-6559](https://desire2learn.atlassian.net/browse/GAUD-6559)

This PR updates the sticky table header (ex. `<th>`) `z-indxex` such that header elements containing open dropdowns will stack on top of those that follow it in the DOM. Without this change, they all have the same `z-index`, so the sticky positioned elements that follow will incorrectly stack on top of the dropdown. (see screenshots)

**Note:** the new browser Popover API will address this in a more durable way since it promotes the element to the `top-layer`. We have a separate story for that, since we'd like to adopt it if possible.

**Before change:**
image

**After change:**
image

[GAUD-6559]: https://desire2learn.atlassian.net/browse/GAUD-6559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/table/table-wrapper.js
```diff
@@ -370,9 +375,19 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		this._tableScrollers = {};
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+
+		this.addEventListener('d2l-dropdown-open', this._handlePopoverOpen);
```
dbatiste, line 381: I chose this `Popover` naming because eventually we want to have a Popover element or mixin that both dropdowns and tooltips can use, in which case the event handling here will accommodate both. 

I chose not to introduce the new `d2l-popover-open` and `d2l-popover-close` events yet. When that work happens, we'll com back in here and update the events its listening for.

components/table/table-wrapper.js
```diff
@@ -589,6 +612,30 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
 		}
 	}
 
+	_updateStickyAncestor(node, popoverOpened) {
+		if (!this.stickyHeaders) return;
+
+		node = getComposedParent(node);
```
dbatiste, line 618: This code basically walks up the tree to find the sticky element and keeps count of the popovers within, which is targeted by the CSS to elevate the `z-index`.

For now, I think we can only expect one dropdown within a table header cell. However, more than one is entirely possible in the future, for example if menu-items supported tooltips (something Jeff and I chatted about recently).

dbatiste:
I need to add a visual-diff test for this.

components/table/test/golden/table/chromium/ltr-default-sticky-scroll-wrapper-multi-row-col-sort-button-dropdown-open.png
dbatiste, line 1: This dropdown is being constrained by the scroll-wrapper containing block. Our change to fixed positioning algorithm will address this limitation, but it is not enabled in visual-diff yet.

components/table/test/golden/table/chromium/rtl-default-sticky-scroll-wrapper-multi-row-col-sort-button-dropdown-open.png
dbatiste, line 1: Same with this one - our fixed position algorithm will address this containment limitation.

margaree (approved):

components/table/table-wrapper.js
```diff
@@ -589,6 +612,30 @@ export class TableWrapper extends RtlMixin(PageableMixin(SelectionMixin(LitEleme
...
+					let popoverCount = Number.parseInt(node.getAttribute('data-popover-count'));
+					if (Number.isNaN(popoverCount)) popoverCount = 0;
+					popoverOpened ? popoverCount++ : popoverCount--;
+
+					if (popoverCount > 0) node.setAttribute('data-popover-count', popoverCount);
+					else node.removeAttribute('data-popover-count');
+
+					return;
+				}
+			}
```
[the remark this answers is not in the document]
dbatiste: Oh yeah, I should be able to add something for this.

PR #4760, opened 2024-06-20
[GAUD-6551] list > hover state

KearseTrevor:
https://desire2learn.atlassian.net/browse/GAUD-6551


components/list/test/list.vdiff.js
```diff
@@ -3,6 +3,7 @@ import '../../dropdown/dropdown.js';
 import '../../dropdown/dropdown-content.js';
 import '../../link/link.js';
 import '../../paging/pager-load-more.js';
+import '../../selection/selection-action.js';
```
KearseTrevor, line 6: noticed missing import

KearseTrevor: https://github.com/BrightspaceUI/core/blob/8f92920c87615e428bde514ce0ad50fbde5cf3b8/components/list/test/list.vdiff.js#L869

components/list/test/list.vdiff.js
```diff
@@ -285,11 +286,11 @@ describe('list', () => {
 			</d2l-list>
 		`;
 		function createSelectableList(opts) {
-			const { selected, addButton } = { selected: false, addButton: false, ...opts };
+			const { selected, addButton, selectionDisabled } = { selected: false, addButton: false, selectionDisabled: false, ...opts };
 			return html`
 				<d2l-list style="width: 400px;" ?add-button="${addButton}">
 					<d2l-list-item label="Item 1" selectable key="1" ?selected="${selected}" color="${ifDefined(!selected ? '#00ff00' : undefined)}">Item 1</d2l-list-item>
-					<d2l-list-item label="Item 2" selection-disabled selectable key="2" ?selected="${selected}" color="${ifDefined(selected ? '#00ff00' : undefined)}">Item 2</d2l-list-item>
+					<d2l-list-item label="Item 2" ?selection-disabled="${selectionDisabled}" selectable key="2" color="${ifDefined(selected ? '#00ff00' : undefined)}">Item 2</d2l-list-item>
```
KearseTrevor, line 293: wanted to be able to dynamically set this so that hover/focus of sibling was possible for new tests

components/list/test/golden/list/chromium/selectable-not-selected.png
dbatiste, line 1: Are we expecting the second checkbox to change in this one?

KearseTrevor: Yeah sorry I brain farted this morning and decided to call this out in the vdiff PR just as I closed and merged it

KearseTrevor: https://github.com/BrightspaceUI/core/pull/4761#issuecomment-2189183470

KearseTrevor: I changed the structure on a groupings of these tests so that I could dynamically change them for the new tests/tests that actually care about them. If we want to toggle the value on/off for specific old tests for backport/equivalency let me know and I can go and make parity with the old.

dbatiste: Ah I see. Yep, thanks. Seems fine.

components/list/list-item-mixin.js
```diff
@@ -302,7 +302,6 @@ export const ListItemMixin = superclass => class extends composeMixins(
 			:host([_focusing-primary-action]) [slot="outside-control-container"],
 			:host(:not([selection-disabled]):not([skeleton])[selected][_hovering-selection]) [slot="outside-control-container"],
 			:host(:not([selection-disabled]):not([skeleton])[selectable][_focusing]) [slot="outside-control-container"] {
-				background-color: white;
```
dbatiste, line 305: Do you know if we have any visual-diff tests that capture cases where the page/parent background is non-white? I didn't dig into this, but I have a nagging concern that by removing the background, we could cause a regression for that case.

KearseTrevor: I unfortunately don't know if we do - when I checked through a number of other tests that included list none of them seemed to contain anything similar. I share the worry of regression - I can spin up a sandbox to display what the behaviour may look like in order to reduce our concerns wrt regression

dbatiste: I took a quick peek on [this demo page](https://live.d2l.dev/prs/BrightspaceUI/core/pr-4760/templates/primary-secondary/demo/integration.html), and it _seems_ ok.

dbatiste: It would be nice if we had a select subset of visual-diff tests that covered that case. It requires a bit of thought though, to figure out which cases are "interesting". We probably don't want to simply double our tests. 

If we were to do that, I'd suggest adding those in a separate PR first, then updating this PR after that's been merged to see how this change affects them. Unless there is a rush to get this change in, that's probably the safest approach, which also helps protect us in the future too.

KearseTrevor: List of tests I find interesting to have an offcolor background on:
- button-hover, button-focus
- all of the selectable tests
  - selectable - href
- draggable > color > selectable > hover
- separators - none

KearseTrevor: We can expand on this but it's a decent starting point I think - I'll work on that other PR to add these tests

KearseTrevor:
This PR is blocked until #4775 is merged

KearseTrevor:
New offcolor vdiff results are in
https://vdiff.d2l.dev/BrightspaceUI/core/1ab6a14285598aa133bbe5714103b1a8f2adf7c3/2024-06-27_01-05-12/report/

I think it would be safest if we consider these at least a partial regression. Let's have a discuss with design to see the following:
- Do they want us to workshop this stop-gap so we are regression proof
  - What is the desired behaviour for these cases then
  - We can potentially do this ourselves but would still need to circle back with them on it
- Do we side table this until the updated design patterns come up with more opinion on these states?
- Is there some alt solve they'd have us go with for the time being if the long term pattern solution is a ways away still

KearseTrevor:
@dbatiste looks like we've got the official go ahead from Jeff and David on this so full throttle ahead?

Only other thing I thought to do was to potentially switch the solid background color on these new tests to a checkerboard but that's obviously outside of the scope of this PR.

Image for reference/conversation sake - mainly came as a hold over from camera/vision system calibrations I used to do back in the day
image

dbatiste (approved):
Looks good.

Re the choice of background... I am good either way. Generally I try to avoid loading images in the visual-diffs if they might result in some delay that could cause test flakiness. But I know we've done it in a couple places.

KearseTrevor:
> Looks good.
> 
> Re the choice of background... I am good either way. Generally I try to avoid loading images in the visual-diffs if they might result in some delay that could cause test flakiness. But I know we've done it in a couple places.

Oh this is just css - though I suppose it could still cause flakiness. 
I'm going to be punting this till inspiration but I'll be looking to add this as an optional on `fixture` most likely
https://codepen.io/KearseTrevor/pen/rNgboLR

PR #4764, opened 2024-06-24
feat: Filter: Add date-range to custom date range filter value component [GAUD-6578]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6578)

Adds `type` property which defaults to `date-time` but can be set to `date` in order to use a `d2l-input-date-range` component inside the custom date filter value component.


components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -39,14 +41,20 @@ class FilterDimensionSetDateTimeRangeValue extends LocalizeCoreElement(LitElemen
 			 * Defaults to "Custom Date Range" (localized). Can be overridden if desired.
 			 * @type {string}
 			 */
-			text: { type: String, reflect: true }
+			text: { type: String, reflect: true },
+			/**
+			 * Date/time range input type
+			 * @type {'date'|'date-time'}
+			 */
+			type: { type: String }
```
margaree, line 49: Another option would be to have a bool like `hide-time`. The advantage of `type` is in case in the future we need to do some sort of time-only range or some other state (though there are currently no plans for those)

components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -124,8 +140,13 @@ class FilterDimensionSetDateTimeRangeValue extends LocalizeCoreElement(LitElemen
 	}
 
 	async _handleDateChange(e) {
-		this.startValue = e.target.startValue;
-		this.endValue = e.target.endValue;
+		if (this.type === 'date') {
+			this.startValue = e.target.startValue ? getUTCDateTimeFromLocalDateTime(e.target.startValue, '0:0') : undefined;
+			this.endValue = e.target.endValue ? getUTCDateTimeFromLocalDateTime(e.target.endValue, '0:0') : undefined;
```
margaree, line 145: `d2l-input-date-range` values are only localized so this gets the UTC value for midnight on the selected date.

components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -104,13 +112,21 @@ class FilterDimensionSetDateTimeRangeValue extends LocalizeCoreElement(LitElemen
...
+				prefer-fixed-positioning
+				start-value="${ifDefined(this.startValue ? getLocalDateTimeFromUTCDateTime(this.startValue) : undefined)}"
+			></d2l-input-date-range>`
+			: html`<d2l-input-date-time-range
 				@change="${this._handleDateChange}"
 				child-labels-hidden
-				data-dimensionvaluekey="${this.key}"
 				end-value="${ifDefined(this.endValue)}"
-				label="Custom Range"
+				label="${this.localize('components.filter-dimension-set-date-time-range-value.text')}"
```
margaree, line 129: This would (potentially) be the same as the list item text if not set by the consumer. I'll be looking at accessibility in a follow up story so I'll do some testing at that point to see if that's too repetitive.

components/filter/filter-dimension-set-date-time-range-value.js
dbatiste: Of the two, I prefer the `type` approach for the reason you mentioned.

Another alternative we can contemplate is busting it into two different components, which might be overkill if the API is essentially the same and there isn't much in the way of conditional logic.

margaree: Agreed - at this point (nor going forward) I don't see it as necessary to split them out.

dbatiste (approved):

PR #4768, opened 2024-06-24
[a11y docn] Update switch component's readme for accessibility project

ChrisLabattD2L:
One of the current initiatives of the Accessibility Llamas is to improve the Daylight site's documentation when it comes to accessibility, which includes adding a new [accessibility icon](https://github.com/BrightspaceUI/core/blob/main/components/icons/images/tier3/accessibility.svg) whenever `ACCESSIBILITY:` is added and decided to move the Accessibility Properties table into the main table. In addition, we've decided to create a new Accessibility section below everything else and include some helpful information about the component. This PR is obviously just for the Switch component, there will be changes to other components when we find more time to do other ones as well.

Currently this is waiting for https://github.com/BrightspaceUI/documentation/pull/1697 to be merged in so that it actually shows the Accessibility Icon instead of just `ACCESSIBILITY:`.

### DIFF
--- a/components/switch/README.md
+++ b/components/switch/README.md
@@ -43,7 +43,7 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 
 | Property | Type | Description |
 |---|---|---|
-| `text` | String, required | Acts as the [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) for the switch. Visible unless text-position is `hidden`. |
+| `text` | String, required | Acts as the primary label for the switch. Visible unless text-position is `hidden`. |
 | `disabled` | Boolean | Disables the switch |
 | `on` | Boolean | Whether the switch is "on" or "off" |
 | `text-position` | String | Valid values are: `start`, `end` (default), and `hidden` |
@@ -106,3 +106,4 @@ If an activity is set to `Visible` but also has other conditions affecting its v
 ## Accessbility
 
 - When `text-position` is set to `hidden`, the required `text` will be used for the switch's hidden label so that it is [accessible](https://www.w3.org/WAI/WCAG22/quickref/#name-role-value) to non-sighted users.
+- `d2l-switch-visibility` is a notable exception to the best practice of having a static term as the label, this was done to provide more clarity in situations where there are external conditions that could influence whether or not the content in question is visible.
--- a/components/switch/README.md
+++ b/components/switch/README.md
@@ -43,7 +43,7 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 
 | Property | Type | Description |
 |---|---|---|
-| `text` | String, required | Accessible text for the switch |
+| `text` | String, required | Acts as a  [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) on the switch. Visible unless text-position is `hidden`. |
 | `disabled` | Boolean | Disables the switch |
 | `on` | Boolean | Whether the switch is "on" or "off" |
 | `text-position` | String | Valid values are: `start`, `end` (default), and `hidden` |
@@ -51,14 +51,6 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 
 - `change`: dispatched when the `on` property is updated
 <!-- docs: end hidden content -->
-
-### Accessibility Properties
-
-To make your usage of `d2l-switch` accessible, use the following property:
-
-| Attribute | Description |
-|---|---|
-| `text` | **REQUIRED** [Acts as a primary label on the switch](https://www.w3.org/WAI/tutorials/forms/labels/). Visible unless text-position is `hidden`. |
 
 ## Visibility Switch [d2l-switch-visibility]
 
@@ -110,3 +102,7 @@ If an activity is set to `Visible` but also has other conditions affecting its v
   </ul>
 </d2l-switch-visibility>
 ```
+
+## Accessbility
+
+- When `text-position` is set to `hidden`, the switch will still make use of the `text` property as it is used to set the [`aria-label`](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA6) of the component, so that screen readers can still pick up what the text of the component is.
--- a/components/switch/README.md
+++ b/components/switch/README.md
@@ -105,4 +105,4 @@ If an activity is set to `Visible` but also has other conditions affecting its v
 
 ## Accessbility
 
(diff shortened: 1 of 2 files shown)


components/switch/README.md
```diff
@@ -43,7 +43,7 @@ The `d2l-switch` element is a generic switch with on/off semantics.
 
 | Property | Type | Description |
 |---|---|---|
-| `text` | String, required | Accessible text for the switch |
+| `text` | String, required | Acts as a  [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) on the switch. Visible unless text-position is `hidden`. |
```
dbatiste, line 46:
```suggestion
| `text` | String, required | Acts as the [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) for the switch. Visible unless text-position is `hidden`. |
```

components/switch/switch.js
```diff
@@ -6,7 +6,7 @@ import { SwitchMixin } from './switch-mixin.js';
 
 /**
  * A generic switch with on/off semantics.
- * @attr {string} text - REQUIRED: The text that is displayed for the switch label.
+ * @attr {string} text - ACCESSIBILITY: REQUIRED: Acts as a  [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) on the switch. Visible unless text-position is `hidden`.
```
dbatiste, line 9:
```suggestion
 * @attr {string} text - ACCESSIBILITY: REQUIRED: Acts as the [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) for the switch. Visible unless text-position is `hidden`.
```

components/switch/README.md
```diff
@@ -110,3 +102,7 @@ If an activity is set to `Visible` but also has other conditions affecting its v
   </ul>
 </d2l-switch-visibility>
 ```
+
+## Accessbility
+
+- When `text-position` is set to `hidden`, the switch will still make use of the `text` property as it is used to set the [`aria-label`](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA6) of the component, so that screen readers can still pick up what the text of the component is.
```
dbatiste, line 108: Perhaps we should avoid referring to the internal implementation (i.e. Aria usage), since it's always subject to change.

```suggestion
- When `text-position` is set to `hidden`, the required `text` will be used for the switch's hidden label so that it is accessible to non-sighted users.
```

dbatiste (commented):
I've include a couple suggestions here. Not sure if we're intending to make sweeping changes to how accessibility notes are communicated. Happy to discuss. 😄

components/switch/README.md
ChrisLabattD2L: So I was thinking of either adding a short sentence after this one that will include a link to https://www.w3.org/WAI/WCAG22/quickref/#name-role-value, mentioning that this is what it's helping us meet or alternatively add that link to the word `accessible` in this sentence instead. Do you think it's too extra and should just leave things as-is?

dbatiste:
@ChrisLabattD2L , this looks fine to me. Are there more changes you are planning for this PR, given that it's in Draft?

ChrisLabattD2L:
> @ChrisLabattD2L , this looks fine to me. Are there more changes you are planning for this PR, given that it's in Draft?

@dbatiste the only thing I was waiting for was a reply to my reply in your last comment about adding a link or not

dbatiste:
Oh yeah. I'm fine with that.

components/switch/README.md
```diff
@@ -110,3 +102,7 @@ If an activity is set to `Visible` but also has other conditions affecting its v
   </ul>
 </d2l-switch-visibility>
 ```
+
+## Accessbility
+
+- When `text-position` is set to `hidden`, the required `text` will be used for the switch's hidden label so that it is [accessible](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA14) to non-sighted users.
```
dbatiste, line 108: Did you mean to use (from your comment) https://www.w3.org/WAI/WCAG22/quickref/#name-role-value instead? I think referencing the WCAG criteria itself is more valuable since it highlights the importance of the value, rather than the specific technique we've chosen, which could change any time.

ChrisLabattD2L: Yep, I copied the link from the wrong tab 😅 I'll replace it with that one 😄

components/switch/switch.js
```diff
@@ -6,7 +6,7 @@ import { SwitchMixin } from './switch-mixin.js';
 
 /**
  * A generic switch with on/off semantics.
- * @attr {string} text - REQUIRED: The text that is displayed for the switch label.
+ * @attr {string} text - ACCESSIBILITY: REQUIRED: Acts as the [primary label](https://www.w3.org/WAI/tutorials/forms/labels/) for the switch. Visible unless text-position is `hidden`.
```
dbatiste, line 9: Hopefully it's obvious that by providing `text`, we handle the label for them, and that they don't need to wrap their switch inside of a `<label>`, add `aria-label` or one of the techniques listed in the W3C forms doc. If we really want to link to something here, should we be limiting to the WCAG criteria instead?

ChrisLabattD2L: I feel like it's straightforward since it's saying the text provided will act as the label. We could specify that they shouldn't need to apply `aria-label` to the switch, but I'm not entirely sure how beneficial that information would be. In terms of changing the link, would this be the preferred one? https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html Or did you have a particular one in mind?

components/switch/README.md
```diff
@@ -110,3 +102,8 @@ If an activity is set to `Visible` but also has other conditions affecting its v
   </ul>
 </d2l-switch-visibility>
 ```
+
+## Accessbility
+
+- When `text-position` is set to `hidden`, the required `text` will be used for the switch's hidden label so that it is [accessible](https://www.w3.org/WAI/WCAG22/quickref/#name-role-value) to non-sighted users.
+- `d2l-switch-visibility` is a notable exception to the best practice of having a static term as the label, this was done to provide more clarity in situations where there are external conditions that could influence whether or not the content in question is visible.
```
margaree, line 109: nit:
```suggestion
- `d2l-switch-visibility` is a notable exception to the best practice of having a static term as the label; this was done to provide more clarity in situations where there are external conditions that could influence whether or not the content in question is visible.
```
Or divide into two separate sentences.

margaree (approved):
LGTM!

PR #4775, opened 2024-06-26
[GAUD-6591] lists > vdiff > add offcolor background tests

KearseTrevor:
https://desire2learn.atlassian.net/browse/GAUD-6591

Related to conversation here https://github.com/BrightspaceUI/core/pull/4760#discussion_r1653354188

- The only pre-existing tests that should be modified are the ones that are updated due to the import correction on `import '../../selection/selection-action.js';`
  - aka, `components/list/test/golden/list/chromium/color-extend-separators-nested-selectable-hover.png` and siblings
 
 Note to reviewer:
 - Feel free to request different background colors if you don't like the pink default
 - If you know of any other novel/unique tests that you would like to have a background color test added for please let me know and I'll include
- If we ever find ourselves wanting to do these background color changes in other tests we should just extract this function into a common space and use via import


components/list/test/golden/list/chromium/selectable-button-selection-disabled-button-disabled-hover-off-color-background.png
dbatiste, line 1: This isn't something you need to worry about for this PR, but this struck me as odd. The background color is bleeding through on disabled checkboxes, I think due to our `opacity: 0.5;` [this CSS](https://github.com/BrightspaceUI/core/blob/main/components/inputs/input-checkbox.js#L57). 🤔

dbatiste (approved):
That's great. Thanks for adding these Trevor!

PR #4779, opened 2024-06-27
Update table wrapper to update z-index of sticky header when tooltip opens/closes

dbatiste:
[GAUD-6592](https://desire2learn.atlassian.net/browse/GAUD-6592)

This PR updates the z-index of sticky table header cells in the same way as we've done for dropdowns, so that tooltips will stack on top of sticky header cells that follow in the DOM.

**Note:** this does _not_ address the sticky + scroll-wrapper + tooltip case. That requires other more significant changes in order to work.

**Before:**
image

**After:**
image

[GAUD-6592]: https://desire2learn.atlassian.net/browse/GAUD-6592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/table/test/golden/table/chromium/ltr-default-sticky-scroll-wrapper-multi-row-col-sort-button-tooltip-open.png
dbatiste, line 1: This isn't desired, but it's what we currently have for the sticky + scroll-wrapper case. I hadn't intended to capture this, but now that I see it, I think it's valuable to keep it since we eventually want this case to work.

margaree (approved):

PR #4780, opened 2024-06-27
Date Filters: Support filter tags [GAUD-6484]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6484)

Looking to get feedback on this approach as it feels a bit hacky in order to get filter-tags to do what we want. Will add comments.


components/filter/demo/filter-tags.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-timezone='{"name":"Canada - Toronto", "identifier":"Australia/Eucla"}'>
```
margaree, line 2: I'll fix this to have identifier match

components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -24,6 +25,10 @@ class FilterDimensionSetDateTimeRangeValue extends LocalizeCoreElement(LitElemen
 			 * @type {string}
 			 */
 			endValue: { type: String, attribute: 'end-value' },
+			/**
+			 * @ignore
+			 */
+			inactive: { type: Boolean },
```
margaree, line 31: Currently `filter-tags` just show the `text` of a filter whenever that filter is selected. However, the usage of filter tags for the custom range has the following customizations:
1. We don't want to show the tag if the custom range filter is selected but no date has yet been selected
2. We don't want the `text` to be the `text` of the filter item (by default "Custom date range") but rather something that actually shows the selected date(s).

In order to do this I added two new properties:
1. `inactive` is used in filter in order to filter out this selected filter when we don't want it shown in filter tags (this also has raised the question of if it should trigger a `filter-change` event when selected but no value; my feeling is that it should since there was a change and then consumers can handle this however they wish but I'm open to thoughts on this as it's a bit of a weird situation)
2. `descriptiveText` contains the text to be used in the tag

components/filter/filter-dimension-set-date-time-range-value.js
```diff
@@ -144,14 +161,46 @@ class FilterDimensionSetDateTimeRangeValue extends LocalizeCoreElement(LitElemen
 	async _handleDateChange(e) {
 		if (this.type === 'date') {
 			this.startValue = e.target.startValue ? getUTCDateTimeFromLocalDateTime(e.target.startValue, '0:0') : undefined;
-			this.endValue = e.target.endValue ? getUTCDateTimeFromLocalDateTime(e.target.endValue, '0:0') : undefined;
+			this.endValue = e.target.endValue ? getUTCDateTimeFromLocalDateTime(e.target.endValue, '23:59:59') : undefined;
```
margaree, line 164: This isn't necessarily related to tags specifically but this is what our end value time should be for the date-only case.

components/filter/filter.js
```diff
@@ -1026,9 +1028,14 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
...
-						if (value.selected) {
+						if (value.selected && !value.inactive) {
 							const keyObject = { dimension: dimension.key, value: value.key };
-							const text = dimension.valueOnlyActiveFilterText ? value.text : `${dimension.text}: ${value.text}`;
+							let text;
+							if (value.descriptiveText) {
+								text = dimension.valueOnlyActiveFilterText ? value.descriptiveText : `${dimension.text}: ${value.descriptiveText}`;
+							} else {
+								text = dimension.valueOnlyActiveFilterText ? value.text : `${dimension.text}: ${value.text}`;
+							}
```
margaree, line 1038: The `SubscriberRegistryController` for `active-filters` is currently only used by `filter-tags`, so this change wouldn't impact anywhere else and also these new properties are only used by the new date-range-value filter, but I still tried to keep this not filter-tags specific.

Let me know any alternative ideas.

lang/en.js
```diff
@@ -28,6 +28,9 @@ export default {
 	"components.filter-dimension-set-date-text-value.textHours": "{num, plural, =1 {Last hour} other {Last {num} hours}}",
 	"components.filter-dimension-set-date-text-value.textDays": "{num, plural, =0 {Today} one {Last {num} day} other {Last {num} days}}",
 	"components.filter-dimension-set-date-text-value.textMonths": "Last {num} months",
+	"components.filter-dimension-set-date-time-range-value.descriptiveTextRange": "{startValue} to {endValue}",
```
margaree, line 31: I'll add these lang terms to the other lang files but I am waiting on Jeff for more input there.

components/filter/filter.js
```diff
@@ -750,6 +750,8 @@ class Filter extends FocusMixin(LocalizeCoreElement(RtlMixin(LitElement))) {
 				this._activeFiltersSubscribers.updateSubscribers();
 			} else if (prop === 'loading') {
 				shouldResizeDropdown = true;
+			} else if (prop === 'text') {
+				this._activeFiltersSubscribers.updateSubscribers();
```
margaree, line 754: The `text` on a `filter-tag` previously wasn't updating if it changed, so this updates that. This potentially normally wouldn't be an issue, however with the new filter date value components the lang terms were sometimes later to come in than when `text` was actually set on the filter-tags so this keeps that in sync.

components/filter/filter-dimension-set-date-time-range-value.js
dbatiste: I think this makes sense.

Maybe an alternative to adding the inactive property would be to set the `descriptiveText = null` and the filter tags could be configured to ignore it. But I think I prefer being explicit with the state property. It's a weird state... where the option has been "selected", but from a filtering perspective, it's not _really_ selected. 🤔 

I'm not opposed to `descriptiveText`, but are we confusing the label text with the value text? I hesitate to introduce a property on the value components that's tag-specific... hopefully we don't end up having to do something like that.

margaree: Currently the list item for the filter uses `value.text` for both the `d2l-list-item` `label` as well as for the displayed list item content. We could use `descriptiveText` in the list item `label` as well, maybe renaming it to something like `description`, in which case it wouldn't be tag specific.

dbatiste: I was curious if we should have `valueText`? Does that make sense or no?

margaree: Maybe - it depends what you were thinking it would be used for? Like would it be the same as what `descriptiveText` is doing here or would it have a larger scope?

dbatiste: Yeah, just wondering if that would be more accurate than `descriptiveText`. I'm not against `descriptiveText`... just wondering if there is a better name for it. 🤷‍♂️

margaree: Yup we can do that

dbatiste (approved):
Took a quick peek at the diffs... I think they all look as expected.

LGTM!

PR #4783, opened 2024-06-27
fix: Filter dates: Change custom range demo text

margaree:
There is a "Last 7 days" option for the precanned date filter ranges, so this changes the text to be something that isn't already available for the custom option.


KearseTrevor (approved):
👍

PR #4784, opened 2024-06-27
fix: Dropdown: When using mobile tray if minWidthOverride > maxWidthOverride use maxWidthOverride [GAUD-6588]

margaree:
[Jira ticket](https://desire2learn.atlassian.net/jira/software/c/projects/GAUD/boards/330?selectedIssue=GAUD-6588)

This is technically an issue with date filters, where we override dropdown `min-width` with 375px but when width is less than that we actually want `min-width` to be less than that. I could fix this more at the filter level with a `matchMedia` change listener but it seemed to make sense to prevent this at the dropdown level. Let me know thoughts on this.


dbatiste:
Seem reasonable to me. Is that visual-diff expected then?

margaree:
> Seem reasonable to me. Is that visual-diff expected then?

It would be, yes. Since in that case the `minWidthOverride` [would be 285px](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-content-mixin.js#L988) but the `maxWidthOverride` would end up [less than that as the screen width gets smaller](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-content-mixin.js#L978). With some quick testing on chrome with the emulator, at 320px screen width it ends up with min: 285px and max: 281px.

I'll also add a visual diff to test this specific case.

PR #4787, opened 2024-06-28
feat: Button > A11y docs improvements

geurts:
Accessibility documentation improvements in collaboration with Llamas. Adds a new top-level section to replace all the individual sections, and flags relevant properties as accessibility.

Also updated the Floating Buttons link (was pointing to GitHub) and improved the Add Button description while I was in there.

### DIFF
--- a/components/button/README.md
+++ b/components/button/README.md
@@ -53,16 +53,6 @@ The `d2l-button` element can be used just like the native button element, but al
 | `primary` | Boolean | Styles the button as a primary button |
 <!-- docs: end hidden content -->
 
-### Accessibility Properties
-
-To make your `d2l-button` accessible, use the following properties when applicable:
-
-| Attribute | Description |
-|---|---|
-| `aria-expanded` | [Indicate expansion state of a collapsible element](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-expanded). Example: [d2l-more-less](https://github.com/BrightspaceUI/core/blob/f9f30d0975ee5a8479263a84541fc3b781e8830f/components/more-less/more-less.js#L158). |
-| `aria-haspopup` | [Indicate clicking the button opens a menu](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-haspopup). Example: [d2l-dropdown](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-opener-mixin.js#L46). |
-| `description` | Use when text on button does not provide enough context. |
-
 ## Subtle Button [d2l-button-subtle]
 
 The `d2l-button-subtle` element can be used just like the native `button`, but for advanced or de-emphasized actions.
@@ -90,16 +80,6 @@ The `d2l-button-subtle` element can be used just like the native `button`, but f
 | `icon` | String | [Preset icon key](../../components/icons#preset-icons) (e.g. `tier1:gear`) |
 | `icon-right` | Boolean | Render the icon on the right of the button |
 <!-- docs: end hidden content -->
-
-### Accessibility Properties
-
-To make your `d2l-button-subtle` accessible, use the following properties when applicable:
-
-| Attribute | Description |
-|---|---|
-| `aria-expanded` | [Indicate expansion state of a collapsible element](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-expanded). Example: [d2l-more-less](https://github.com/BrightspaceUI/core/blob/f9f30d0975ee5a8479263a84541fc3b781e8830f/components/more-less/more-less.js#L158). |
-| `aria-haspopup` | [Indicate clicking the button opens a menu](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-haspopup). Example: [d2l-dropdown](https://github.com/BrightspaceUI/core/blob/main/components/dropdown/dropdown-opener-mixin.js#L46). |
-| `description` | Use when text on button does not provide enough context. |
 
 ### Subtle Button with Custom Icon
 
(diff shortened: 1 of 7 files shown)


components/button/README.md
```diff
@@ -198,7 +166,21 @@ The `d2l-button-add` element is for quickly adding new items inline (for example
 
 ## Floating Buttons [d2l-floating-buttons]
 
-See [floating buttons](https://github.com/BrightspaceUI/core/tree/main/components/button/floating-buttons.md).
+See [floating buttons](../../components/floating-buttons).
```
margaree, line 169:
```suggestion
See [floating buttons](./floating-buttons.md).
```

You'll want to test this but I think it will work here and in Daylight. The tooltip one below looks to be fine.

components/button/README.md
```diff
@@ -198,7 +166,21 @@ The `d2l-button-add` element is for quickly adding new items inline (for example
...
+
+## Accessibility
+
+Daylight buttons rely on standard button semantics to ensure a smooth experience for all assistive technologies, but there are a few interesting details to note:
+
+* If the button's context is implied by visual layout, then `description` can be used to add missing context
+  * Example: if multiple page sections have an Edit button relying on visual layout to indicate which section it edits, there could be extra information in the `description` to help differentiate the Edit buttons for non-sighted users
+
+* For icon buttons where there is no visible, `text` will be displayed in tooltip
```
dlockhart, line 178: Is this what you meant?

```suggestion
* For icon buttons where there is no text visible, `text` will be displayed in tooltip
```

components/button/floating-buttons.md
```diff
@@ -58,3 +58,9 @@ Floating workflow buttons behavior can be added by using the `<d2l-floating-butt
 |--|--|--|
 | `always-float` | Boolean | Indicates to display buttons as always floating |
 <!-- docs: end hidden content -->
+
+## Accessibility
+
+Despite visually floating or sticking to the bottom of the viewport, floating buttons maintain their position in the document's structure, typically at the bottom of the content. This means the tab order is unaffected when the buttons float and the effect is invisible to screen reader users.
```
dlockhart, line 64: Nit:

```suggestion
Despite visually floating or sticking to the bottom of the viewport, floating buttons maintain their position in the document's structure, typically at the bottom of the content. This means the tab order is unaffected when the buttons float and the effect is imperceptible to screen reader users.
```

components/button/floating-buttons.md
```diff
@@ -58,3 +58,9 @@ Floating workflow buttons behavior can be added by using the `<d2l-floating-butt
...
+
+Be cautious if using `always-float`.   since screen magnifier users may find it difficult to locate the buttons at the bottom of a large viewport, especially when the content is short enough to expect to find the buttons inline.
```
dlockhart, line 66: Did you mean to have a comma here?

```suggestion
Be cautious if using `always-float`, since screen magnifier users may find it difficult to locate the buttons at the bottom of a large viewport, especially when the content is short enough to expect to find the buttons inline.
```

components/button/README.md
geurts: Ahh, I mean "no visible label" but must have removed it in one of my edits 🤦‍♂️

components/button/README.md
geurts: Okay will test it out! Thx

components/button/floating-buttons.md
geurts: Nice, definitely more accurate, thanks 🙏

dlockhart (approved):

components/button/README.md
```diff
@@ -198,7 +166,21 @@ The `d2l-button-add` element is for quickly adding new items inline (for example
 
 ## Floating Buttons [d2l-floating-buttons]
 
-See [floating buttons](https://github.com/BrightspaceUI/core/tree/main/components/button/floating-buttons.md).
+See [floating buttons](../floating-buttons).
```
ChrisLabattD2L, line 169: This link seems to send to a 404 page when clicking it on GitHub, I think you need the `.md` file extension at the end since there are multiple files with that name

components/button/floating-buttons.md
```diff
@@ -58,3 +58,9 @@ Floating workflow buttons behavior can be added by using the `<d2l-floating-butt
 |--|--|--|
 | `always-float` | Boolean | Indicates to display buttons as always floating |
 <!-- docs: end hidden content -->
+
+## Accessibility
+
+Despite visually floating or sticking to the bottom of the viewport, floating buttons maintain their position in the document's structure, typically at the bottom of the content. This means the tab order is unaffected when the buttons float and the effect is imperceptible to screen reader users.
+
+Be cautious if using `always-float`, since screen magnifier users may find it difficult to locate the buttons at the bottom of a large viewport, especially when the content is short enough to expect to find the buttons inline.
```
ChrisLabattD2L, line 66: For `Be cautious if using...`, I just wanted to check if you meant to put `of` instead of `if`? It might also sound better if it's changed to `Be cautious when using...`

components/button/README.md
```diff
@@ -198,7 +166,21 @@ The `d2l-button-add` element is for quickly adding new items inline (for example
...
+
+Daylight buttons rely on standard button semantics to ensure a smooth experience for all assistive technologies, but there are a few interesting details to note:
+
+* If the button's context is implied by visual layout, then `description` can be used to add missing context
+  * Example: if multiple page sections have an Edit button relying on visual layout to indicate which section it edits, there could be extra information in the `description` to help differentiate the Edit buttons for non-sighted users
+
+* For icon buttons where there is no visible label, `text` will be displayed in a tooltip
+  * If both `text` and `aria-label` are used, then `aria-label` will be used as the primary label while `text` will be used in a [tooltip](../../components/tooltip)
+
+* When buttons are used to expand more content or to display a menu or dropdown, [aria-expanded](https://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded) and [aria-haspopup](https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup) attributes should be used in accordance with best practices — follow the links to learn more
```
ChrisLabattD2L, line 181: You should be able to have both a link and code formatting (at least in GitHub) so you can do something like:
[`aria-expanded`](https://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded) and [`aria-haspopup`](https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup)

components/button/floating-buttons.md
geurts: I meant `if` but `when` is definitely better -- thanks!

components/button/README.md
geurts: Oh wow, had no idea you could do that, nice :)

PR #4788, opened 2024-06-28
Initial popover

dbatiste:
[GAUD-6457](https://desire2learn.atlassian.net/browse/GAUD-6457)

This PR introduces the initial, low level `PopoverMixin` that will eventually be used by (but not limited to) dropdown and tooltip rendering. This is a work in progress... currently it:

* simply detects whether the native popover API is supported and uses it if possible
* exposes an `opened` property that controls whether the element is showing or not (named this based on existing dropdown and dialog `opened`, and aligns with terminology in the native `:popover-open` pseudo-class 
* sets up custom CSS variables that will be used by tooltips and dropdowns, since they have different style requirements

Future PRs will introduce focus management, light-dismiss/soft-close behaviour, dismiss (escape key) behaviour, sizing & positioning, pointer rendering, etc. Internal layout will be handled by the consumer (dropdown & tooltip).

The `top-layer`:
image

[GAUD-6457]: https://desire2learn.atlassian.net/browse/GAUD-6457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
+import '../colors/colors.js';
+import { css, html } from 'lit';
+
+const isSupported = ('popover' in HTMLElement.prototype);
+
+// eslint-disable-next-line no-console
+console.log('Popover', isSupported);
```
dbatiste, line 7: This will eventually be removed.

components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+export const PopoverMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			/**
+			 * Whether the popover is open or not
+			 * @type {boolean}
+			 */
+			opened: { type: Boolean, reflect: true },
+			_useNativePopover: { type: Boolean, reflect: true, attribute: 'popover' }
```
dbatiste, line 18: Calling the native `showPopover` or `hidePopover` requires the native `popover` attribute, otherwise an error will be thrown by the browser.

components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+
+	static get styles() {
+		return css`
+			:host {
+				--d2l-popover-default-background-color: #ffffff;
+				--d2l-popover-default-border-color: var(--d2l-color-mica);
+				--d2l-popover-default-border-radius: 0.3rem;
+				--d2l-popover-default-foreground-color: var(--d2l-color-ferrite);
+				--d2l-popover-default-shadow: 0 2px 12px 0 rgba(0, 0, 0, 0.15);
+				background-color: transparent;
```
dbatiste, line 30: We need to override many of these properties because the browser applies them when `popover` is set. We want to internalize these styles because we have not only the content container to deal with, but eventually, also the pointer.

components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+		`;
+	}
+
+	constructor() {
+		super();
+		this.opened = false;
+		this._useNativePopover = isSupported;
+	}
+
+	connectedCallback() {
```
dbatiste, line 69: These aren't needed at the moment. Not sure why I created them so early. I can remove them but eventually they will be needed.

components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+		super.updated(changedProperties);
+		if (changedProperties.has('opened')) {
+
+			if (this.opened) {
+				this.dispatchEvent(new CustomEvent('d2l-popover-open', { bubbles: true, composed: true }));
+			} else if (changedProperties.get('opened') !== undefined) {
+				this.dispatchEvent(new CustomEvent('d2l-popover-close', { bubbles: true, composed: true }));
+			}
+
+			if (this._useNativePopover) {
```
dbatiste, line 87: The interesting part. Calling `showPopover` is what causes the browser the promote the element to the "top-layer". Likewise, calling `hidePopover` is what removes it from the "top-layer".

components/popover/popover-mixin.js
```diff
@@ -0,0 +1,99 @@
...
+
+export const PopoverMixin = superclass => class extends superclass {
+
+	static get properties() {
+		return {
+			/**
+			 * Whether the popover is open or not
+			 * @type {boolean}
+			 */
+			opened: { type: Boolean, reflect: true },
```
dbatiste, line 17: I could be convinced to go with either `opened` or `showing`. The native API uses `show/hidePopover` for the methods, but then uses `:popover-open` for the CSS pseudo-class. Our dialog and dropdown use `opened`, while our tooltips use `showing`. 🤷‍♂️

margaree: Oh interesting - should we get these consistent between dropdown and tooltip? If we did want to do that, `d2l-tooltip` + `showing` has far less usages (but overall it's probably not super necessary)

I'm fine with `opened` here.

components/popover/test/popover-mixin.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+	describe('events', () => {
+
+		it('should fire the open event when the opened attribute is set to true', async() => {
+			const elem = await fixture('<d2l-test-popover></d2l-test-popover>');
+			elem.opened = true;
+			await oneEvent(elem, 'd2l-popover-open');
+		});
+
+		it('should fire the close event when the opened attribute is set to true', async() => {
```
margaree, line 22:
```suggestion
		it('should fire the close event when the opened attribute is set to false', async() => {
```

margaree: nit: Since it's also in the component logic, it could be good to have a test for if `<d2l-test-popover></d2l-test-popover>` has `opened` assigned to false (from undefined) then it does not fire the close event.

components/popover/popover-mixin.js
dbatiste: It's a bikeshed moment. 😄  I'm good either way, and we can easily switch with subsequent changes.

components/popover/test/popover-mixin.test.js
dbatiste: I added event tests for both the opened and closed initial states.

margaree (approved):

